PR #2, opened 2020-06-22
Add name mangling for method and type names to fit foreign language naming conventions

jhugman:
This PR may not be useful, but starts as a way of exploring the codebase. Absenting some issues ;)

* The IDL may specify method names in camel case or snake case.
* The generated kotlin method names are in camel case. 
* The generated python method names are in snake case.
* The rust method names are in snake case.

Is this useful? 

* If the rust devs are delivering generated APIs to App developers, then yes. 
* If the rust devs are delivering handmade wrappers around the generated APIs, then no. 

This should probably be a candidate for project specific/language specific `uniffi` config.

### DIFF
--- a/examples/naming-conventions/src/naming-conventions.idl
+++ b/examples/naming-conventions/src/naming-conventions.idl
@@ -1,7 +1,16 @@
+enum Case {
+  "snake_case",
+  "UpperCamelCase",
+  "camelCase",
+  "SHOUTY_SNAKE_CASE",
+};
 
 namespace naming_conventions {
-  snake_case_object snake_case_method(u32 id);
-  CamelCaseObject camelCaseMethod(u32 id);
+  snake_case_object snake_case_method(u32 id, Case src);
+  CamelCaseObject camelCaseMethod(u32 id, Case src);
+
+  Case get_snake_case();
+  Case getCamelCase();
 };
 
 dictionary snake_case_object {
--- a/examples/naming-conventions/src/naming-conventions.idl
+++ b/examples/naming-conventions/src/naming-conventions.idl
@@ -1,0 +1,13 @@
+
+namespace naming_conventions {
+  snake_case_object snake_case_method(u32 id);
+  CamelCaseObject camelCaseMethod(u32 id);
+};
+
+dictionary snake_case_object {
+  u32 id;
+};
+
+dictionary CamelCaseObject {
+  u32 id;
+};
--- a/src/bindings/kotlin/gen_kotlin.rs
+++ b/src/bindings/kotlin/gen_kotlin.rs
@@ -4,7 +4,7 @@
 
 use anyhow::Result;
 use askama::Template;
-use heck::{ CamelCase, MixedCase };
+use heck::{ CamelCase, MixedCase, ShoutySnakeCase };
 
 use crate::interface::*;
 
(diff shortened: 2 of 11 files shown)


rfk:
Great, thanks for diving in!! I had `heck` in the list of dependencies here with the intention of experimenting with this sort of thing at some point, so very glad to have you take a look.

I think it makes sense to pursue the " rust devs are delivering generated APIs to App developers" philosophy, at least to start. So being able to generate bindings that feel like they fit in the target language will be important.

I'll try to carve out some time to take a detailed look later today.

examples/naming-conventions/src/naming-conventions.idl
```diff
@@ -0,0 +1,13 @@
+
+namespace naming_conventions {
+  snake_case_object snake_case_method(u32 id);
+  CamelCaseObject camelCaseMethod(u32 id);
```
rfk, line 4: Interesting philosophical point here about what to allow in the IDL, and I'm kind of in two minds about it. On one hand, it seems wrong to outright error out if something declared in this file does not match a particular naming convention, but on the other hand, allowing mixing conventions like this might lead to confusion. Perhaps it's the sort of thing that might one day be a candidate for a warning.

src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -283,13 +284,13 @@ internal interface _UniFFILib : Library {
 // Public interface members begin here.
 
 {% for e in ci.iter_enum_definitions() %}
-    enum class {{ e.name() }} {
+    enum class {{ e.name()|decl_name_kt }} {
```
rfk, line 287: It's probably not clear from the code, but in my head I'd originally intended the "decl" prefix on these filter function to be read as "declare type". Something like "class_name_kt" might be a clearer name for this filter (and the existing "decl_kt" might be better as something like "type_kt" for clarity).

src/scaffolding.rs
```diff
@@ -34,7 +35,7 @@ ffi_support::define_bytebuffer_destructor!({{ ci.ffi_bytebuffer_free().name() }}
 // of items as declared in the rust code, but no harm will come from it.
 
 {% for e in ci.iter_enum_definitions() %}
-    unsafe impl uniffi::support::ViaFfi for {{ e.name() }} {
+    unsafe impl uniffi::support::ViaFfi for {{ e.name()|decl_name_rs }} {
```
rfk, line 38: Rather than doing this conversion when generating the Rust scaffolding, what do you think about the following approach:

* Decree that the naming convention in the IDL should be the same as the naming convention in the Rust code (but allow users to write it differently if they want, maybe with a warning in future).
* When parsing the IDL, canonicalize all names to the rust convention as part of building the `ComponentInterface` struct (this would be the place where we warn about it in future).
* Allow each foreign language binding to choose whether and how it wants to deviate from that convention to fit in with the norms of that language.

Alternately, we could just insist that the names in Rust and IDL must match exactly, however you've written them, and not try to do any conversion.

Keeping the IDL and the Rust more tightly bound makes sense to me in a vague kind of way, especially if we imagine that one day we might general definitions using inline rust macros rather than a separate IDL file. It would also mean there are no "control knobs" for callers to fiddle with when interfacing between the IDL and the Rust code, while allowing the possibility of such knobs when generating the bindings (which there is also a placeholder `Config` class for in the kotlin and python bindings generators).

Thoughts?

rfk:
What I also need to do, is figure out some basic test infrastructure here. I'd like to be able to turn the existing `example.py` and `example.kts` files into actual tests with assertions that can be executed to make sure the generated bindings are working as expected. So far I've just been running them by hand and eyeballing the output, but that's clearly not going to scale.

src/scaffolding.rs
jhugman: If I understand correctly, I agree with your: 

> insist that the names in Rust and IDL must match
> while allowing the possibility of such knobs when generating the bindings

There will be some back pressure from the hand written rust from the linter or rustc so that the rust and WebIDL will naturally tend to conform to a rust naming convention— because the rust compiler will insist, rather than `uniffi`.  If users really want to have funky rust identifiers, they'll have to add their own annotations in rust.

The foreign language bindings are a view onto the WebIDL, so I think they should have a little bit of leeway to convert/mangle names to suite the language/project. I don't think we should do it without configuration in the long run because:

 * collisions are still possible, (the webIDL could define `foo_bar` and `fooBar`, which would both convert to the same methods). This might correspond to a normalize + check step before generating any bindings.
 * generating source-backward-compatible bindings for existing projects may still be an aim for the project.

If all this is desirable, then we should also consider doing the same for the property names.

src/bindings/kotlin/gen_kotlin.rs
jhugman: Ok, I'll change functions as you suggest.

examples/naming-conventions/src/naming-conventions.idl
jhugman: See discussion below. I think the chance of collision is too high here.

src/scaffolding.rs
rfk: > collisions are still possible, (the webIDL could define foo_bar and fooBar, which would both convert to the same methods).
> This might correspond to a normalize + check step before generating any bindings.

This is a good point; I think I'd be interested in trying out a normalize+check step when processing the IDL. In the same way that we'd error out if you tried to define two methods with exactly the same name, we could error out if you try to define two methods that would be ambiguous under common case conversions. (And it seems easier to error out now and loosen the rules later, than to start with loose rules and try to tighten them up later).

src/bindings/kotlin/gen_kotlin.rs
jhugman: :+1:

rfk:
(heads-up that I edited this PR to target the new `main` branch rather than `master`)

src/scaffolding.rs
jhugman: Filed #13

rfk:
Thanks for pushing on this! Just wanted to comment to set expectations, I should be able to find time to look at it by end-of-week.

src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -273,66 +274,53 @@ internal interface _UniFFILib : Library {
...
-            internal fun lift(n: Int): {{ e.name() }} {
-                return when (n) {
-                  {% for value in e.values() %}
-                  {{ loop.index }} -> {{ value }}
-                  {% endfor %}
-                  else -> {
-                      throw RuntimeException("invalid enum value, something is very wrong!!")
-                  }
+            internal fun lift(n: Int) =
+                try { values()[n - 1] }
```
rfk, line 294: oh, neat!

PR #6, opened 2020-06-24
Generate Swift bindings

linacambridge:
This is very much a work-in-progress, but we've got some Hot (Swift) Garbage being generated now! 🔥🗑️ Try running `cargo run -- generate -l swift` to generate a bridging header and the Swift library.

I haven't figured out the right incantations to build this with `swiftc` yet. Locally, I made a Swift Package (using `swift package init --type library`), with C and Swift sources (and without using Xcode), so I know it's possible. SPM relies on a lot of conventions (if you have a `Sources/CTarget` folder with a bunch of `.c` files, and a `Sources/CTarget/include` folder with your `.h` files, SPM is smart enough to build the C library and link in the headers for bridging)—but I'm not sure yet how to make it link with an already-built library.


src/bindings/swift/gen_swift.rs
```diff
@@ -0,0 +1,498 @@
...
+    }
+}
+
+#[derive(Template)]
+#[template(
+    ext = "h",
+    escape = "none",
+    source = r#"
+// This file was autogenerated by some hot garbage in the `uniffi` crate.
+// Trust me, you don't want to mess with it!
```
rfk, line 31: I'm very happy that you chose to copy this comment across verbatim 😆

src/bindings/swift/gen_swift.rs
```diff
@@ -0,0 +1,498 @@
...
+    /// ...
+    pub fn lift_swift(
+        name: &dyn fmt::Display,
+        type_: &TypeReference,
+    ) -> Result<String, askama::Error> {
+        Ok(format!("{}.lift({})", decl_swift(type_)?, name))
+    }
+
+    /// ...
+    pub fn decl_enum_variant_swift(name: &str) -> Result<String, askama::Error> {
```
linacambridge, line 494: I like what @jhugman is doing with naming conventions over in #2! This was a pretty specific hack to turn `SCREAMING_CASE` enum variant names into `camelCase`; it would be amazing to have something more generic!

src/bindings/swift/gen_swift.rs
```diff
@@ -0,0 +1,498 @@
...
+    ext = "swift",
+    escape = "none",
+    source = r#"
+// This file was autogenerated by some hot garbage in the `uniffi` crate.
+// Trust me, you don't want to mess with it!
+
+extension Data {
+    init(rustBuffer: RustBuffer) {
+        // TODO: This copies the buffer. Can we read directly from a
+        // Rust buffer?
```
linacambridge, line 78: Copying on the way in and out simplifies things—I mainly did that to get it working, but we should see if we can refactor to avoid it.

src/bindings/swift/gen_swift.rs
```diff
@@ -0,0 +1,498 @@
...
+    let value = T.lift(from: reader)
+    if reader.hasRemaining() {
+        fatalError()
+    }
+    buf.deallocate()
+    return value
+}
+
+extension Bool : Lowerable, Liftable {
+    func lower() -> UInt8 {
```
linacambridge, line 206: Add comments explaining the difference between `lower()` and `lower(into:)` (and `lift(_)` and `lift(from:)`).

src/bindings/swift/gen_swift.rs
```diff
@@ -0,0 +1,498 @@
...
+    }
+
+    static func lift(from buf: DataReader) -> Self {
+        switch buf.readUInt8() {
+        case 0:
+            return nil
+        case 1:
+            return Wrapped.lift(from: buf)
+        default:
+            fatalError()
```
linacambridge, line 304: Better errors, let's turn this into a `throws` function instead of crashing! 😁

src/bindings/swift/gen_swift.rs
```diff
@@ -0,0 +1,498 @@
...
+        guard let value = self else {
+            buf.writeUInt8(0)
+            return
+        }
+        buf.writeUInt8(1)
+        value.lower(into: buf)
+    }
+}
+
+extension Optional: Liftable where Wrapped: Liftable {
```
rfk, line 292: I don't have any deep swift knowledge, and this protocol/extension stuff feels surprisingly nice, and a good mental map back to how it's done on the Rust side. I like it!

linacambridge: Yeah, this was super fun to write! I haven't worked much with Swift before, but its generics feel really familiar coming from Rust, which is exciting! 🦀

linacambridge:
Also if we could get `xctest` working and running on CI, that would be cool!

src/bindings/swift/gen_swift.rs
linacambridge: https://twitter.com/jckarter/status/1141752551185866753 🤣

linacambridge:
I think I'm happy with this! 😁 All the examples work now, and I cleaned up the shared helpers a little.

rfk:
Looking forward top taking a more detailed look at this! Just wanted to pop in to say, I probably won't get a chance to look at this for a day or two.

src/lib.rs
```diff
@@ -311,23 +318,24 @@ fn run_bindings_exec_subcommand(
...
             }
             let script_file_buf = PathBuf::from(script_file.unwrap());
-            let ext = script_file_buf.extension().unwrap_or_default();
-            if ext == "kts" {
-                "kotlin"
-            } else if ext == "py" {
-                "python"
-            } else {
-                bail!("Cannot guess language of script file, please specify it explicitly")
+            match script_file_buf.extension().unwrap_or_default().to_str() {
```
rfk, line 321: Oooooh, yes I see, `to_str()`; I spent like half an hour the other day trying to figure out whether I could `match` on an `OsStr` in a nice way, but this makes a lot of sense :-)

src/bindings/swift/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+        let out_path = PathBuf::from(out_dir);
+        let cc_options: Vec<OsString> = MODULE_MAPS
+            .iter()
+            .flat_map(|module_map| {
+                let mut module_map_file = out_path.clone();
+                module_map_file.push(module_map);
+                if !module_map_file.exists() {
+                    // Missing module maps (for example, `uniffi_arithmetic` when we're
+                    // running the `geometry` example) will fail, so ignore maps that
+                    // don't exist. Gross!
```
rfk, line 147: Hrm, this does seem fiddly. I wonder if you could do the reverse, and iterate over all files in `out_path` looking for any with the right extension, then deciding what to use based on that.

linacambridge: That’s a great idea—I could have it search `out_path` for `.dylib`s and Swift module files (or, better yet, output the generated stuff into `swift`, `kotlin`, `python`, etc. subdirectories in `out_path`—so we’d avoid, say, linking the libs for `example`—or just add a filter for those). A special target subdirectory per language would also give you a good idea of what to include.

TBH, the only reason for this jiggery-pokery is so that you can run the Swift examples from the command line, or in a REPL. If you’re building with Xcode or SPM, you only need the bridging header and generated `.swift` file (SPM will make a `modulemap` for you, and also automatically make a dylib for the REPL—looking at that was how I figured out the right set of flags to pass to `swift` and `swiftc`). We only need the `swiftmodule`, `modulemap`, and `dylib` for `cargo run — exec`.

examples/arithmetic/example.swift
```diff
@@ -0,0 +1,4 @@
+import arithmetic
+
+print("Alrighty, let's do some arithmetic FROM RUST, IN SWIFT!")
+print("2 + 3 = \(add(a: 2, b: 3, overflow: .saturating))")
```
rfk, line 4: Huh. I don't know much about swift so I want to double-check what I read a bit about online: is this thing where you have to name all the arguments just an expected convention in the swift world, or are we doing something that means you have to explicitly name them in cases where the caller might not expect to need to?

src/bindings/swift/gen_swift.rs
rfk: +1 to "get it working" FWIW!

src/bindings/swift/gen_swift.rs
```diff
@@ -0,0 +1,615 @@
...
+
+// Types conforming to `Liftable` know how to deserialize ("lift") themselves
+// from a byte buffer. This is equivalent to the `Liftable` trait on the Rust
+// side.
+protocol Liftable {
+    static func lift(from: Reader) throws -> Self
+}
+
+// Types conforming to `Lowerable` know how to serialize ("lower") themselves
+// into a byte buffer. Equivalent to the `Lowerable` trait on the Rust side.
```
rfk, line 224: These "lift" and "lower" names, I pinched from the Wasm Interface Types proposal, but TBH we're not really using them in a way that's very similar to how they're used there. Without wanting to rush into changing things right now, I'm curious if you found they helped or hindered understanding of how the code works? (As opposed to, say, a name like "Deserializble" which is what you've used in the comment here to explain that "Liftable" actually means).

src/bindings/swift/gen_swift.rs
```diff
@@ -0,0 +1,615 @@
...
+}
+{% endfor %}
+
+{% for func in ci.iter_function_definitions() %}
+{%- match func.return_type() -%}
+{%- when Some with (return_type) %}
+
+public func {{ func.name() }}(
+    // TODO: More considered handling of labels (don't emit them
+    // for single-argument functions; others?)
```
rfk, line 433: Aha, I guess this is related to my question about argument labels above.

src/bindings/swift/gen_swift.rs
```diff
@@ -0,0 +1,615 @@
...
+
+{%- endmatch %}
+{% endfor %}
+
+{% for obj in ci.iter_object_definitions() %}
+public class {{ obj.name() }} {
+    private let handle: UInt64
+
+    {%- for cons in obj.constructors() %}
+    public init(
```
rfk, line 468: Hrm, interesting name-collision case here. What should we do if the Rust side declares a normal member method (or maybe a staticmethod?) named `init`?

src/bindings/swift/mod.rs
rfk: Something else that might be useful here is, when each example component runs the command-line helper, it passes in the name of the component [here](https://github.com/rfk/uniffi-rs/blob/f28adde220adfcb01c2732b23d7c857010a0d268/src/lib.rs#L185). You might be able to thread that name through down to the call to `run_swift_script` and use it to help select which files to include here.

examples/arithmetic/example.swift
linacambridge: I'm glad you pointed this out!

I think, as of [this RFC](https://github.com/apple/swift-evolution/blob/master/proposals/0046-first-label.md), all parameters need a label by default, unless you use `_`. If you omit the label in the declaration, like the generated code does, it'll be the same as the argument name. But I'm not sure if we can come up with a generic convention. There are some [guidelines](https://swift.org/documentation/api-design-guidelines/#argument-labels) for labels (and a great post with some more [rules of thumb](https://forums.swift.org/t/when-to-use-argument-labels-a-new-approach/1289/7)), but it doesn't seem like something we can infer in the general case. It only takes a few cases for "good enough" to produce bizarro results—like `intersection(a, ln2: b)`.

I was thinking "no label if only one argument, labels if more than one" would be a general enough rule, but even that's not the case. [This thread](https://forums.swift.org/t/when-to-use-argument-labels-a-new-approach/1289) has good counterexamples (`a.dismiss(animated: b)`; also, something like `sort(stable: Bool)`). On the other hand, with a function like `zip(a, b)` or `intersection(a, b)`, the names are redundant.

I think the answer might be in adding smarts to the IDL syntax. What if we added a special annotation for labels, or let you specify labels in declarations with a special syntax (like `Point? intersection(Line ln1 labeled _, Line ln2 labeled _)`, as an example)...or let you to omit argument names (so `Point? intersection(Line, Line)` would imply no labels, while `Point? intersection(Line ln1, Line ln2)` would emit labels)?

On the other hand, none of the other languages we target support this. Python and Kotlin have keyword/named arguments, but they're the same as the argument names. I guess we could slurp up labeled arguments into `**kwargs` in Python, or options objects in JS, but _why_, when you can pass a `record` instead? That's a whole lotta work to support a syntax that's only special in Swift.

Another idea is, what if no labels at all? They do help with clarity, but...no labels would match our other generated bindings.

jhugman: I have seen adding a naming conventions in the IDL work okayish.

e.g. count the number of suffix words in the function name, and use those to name the arguments; if you run out of suffix words, use `_`.

```idl
Point? intersection_with(Line line1, Line line2)
```

generating

```swift
func intersection(_ line1: Line, with line2: Line) -> Point?
```

But there're a tons of ways this will produce rubbish results, particularly if we're mixing other naming conventions.

I like your annotation idea but I have no preference of syntax.

```idl
@argumentLabels("_", "with")
Point? intersection(Line line1, Line line2)
```

But I think the reason for this comment is around collisions. 

There may be nothing to do for now, but it's adjacent to this problem: swift and kotlin both have different ways of defining function signatures (method name + argument labels vs method name + argument types).

src/bindings/swift/gen_swift.rs
jhugman: If this project goes anywhere, I'd like a t-shirt with "hot garbage" on it.

src/bindings/swift/gen_swift.rs
```diff
@@ -0,0 +1,615 @@
...
+        {{ arg.name()|lower_swift(arg.type_()) }}{% if loop.last %}{% else %},{% endif %}
+        {%- endfor %}
+    )
+    return try! {{ "_retval"|lift_swift(return_type) }}
+}
+
+{% when None -%}
+
+public func {{ func.name() }}(
+    {%- for arg in func.arguments() %}
```
jhugman, line 449: I wonder if we can define a function arguments template, and call it from this one.

src/bindings/swift/gen_swift.rs
jhugman: My vote: error/crash.

src/bindings/swift/mod.rs
```diff
@@ -0,0 +1,179 @@
...
+    library: String,
+}
+
+/// Generate Swift bindings for the given ComponentInterface, as a string.
+pub fn generate_swift_bindings(ci: &ComponentInterface) -> Result<Bindings> {
+    let config = Config::from(&ci);
+    use askama::Template;
+    let header = BridgingHeader::new(&config, &ci)
+        .render()
+        .map_err(|_| anyhow!("failed to render Swift bridging header"))?;
```
jhugman, line 31: OMG. Bridging headers.

src/bindings/swift/gen_swift.rs
linacambridge: Yeah, let's add `init` and `deinit` to our future reserved word list (#10) for Swift.

rfk:
(heads-up that I edited this PR to target the new `main` branch rather than `master`)

linacambridge:
@rfk Haha, I was planning to circle back to this next week, and see if I could figure out a better way to build the examples without hardcoding library names! 😅 Thanks for having a look and the review feedback! Let me get an issue on file, and we can figure it out as part of getting automated tests running!

rfk:
I should also say, I really don't want to only be able to develop this thing on a mac, so I experimented a little bit with building the swift command-line toolchain under WSL. It seemed to work, my next step is to see if I can compile the `Foundation` lib and make it available. We'll see how it goes, but I'm hopeful.

src/bindings/swift/gen_swift.rs
linacambridge: I found them pretty helpful! The thing that confused me at first was having `lift`, `liftFrom`, `lower`, and `lowerInto` methods when I was reading through the Kotlin generator. At first, I wondered why we needed `lift` and `lower`, and why we couldn't use a byte buffer for everything? Then I had a closer look at `ViaFfi` in the Rust support, and it clicked that `lift` and `lower` were equivalent to `try_from_ffi_value` and `into_ffi_value`, because allocating buffers when you only wanted to pass primitives would be super inefficient.

The other thing is, `Serializable` in Swift (the same as `ViaFfiUsingByteBuffer` in Rusr) feels like a bad name, since its constraint is `Liftable & Lowerable`, but `Lowerable` is for "types that know how to serialize themselves". 🤦‍♀️

Also, even though Swift really likes protocol names to be nouns and adjectives, I wonder if we should name our protocols and methods the same as Rust, for consistency's sake? Whatever we decide on, have it match in all our bindings so that it's easy to follow.

src/bindings/swift/gen_swift.rs
linacambridge: Let's do this as part of #15!

linacambridge:
Hmmm, I think the only thing we need `Foundation` for is `Data`! I wonder if we could remove that dependency if we fixed th TODO about copying and just read directly from and into the Rust buffer... 🤔

jhugman:
@linacambridge I'd love to start using this! I really want to start #15 on your Swift code; BUT: I don't have it building yet.

What magic did you need to get it building `cargo run -- generate -l swift` with? Is it documented someplace and I haven't found it? Am I too stupid to work it out?

jhugman:
Specifically: 

```sh
Generating Swift bindings...
Compiling generated Swift bindings into module...
<unknown>:0: error: unable to load standard library for target 'x86_64-apple-macosx10.15'
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: running `swiftc` failed', src/main.rs:6:5
```

linacambridge:
Hey @jhugman! Oh, wild—for me, just running `cargo run -- generate -l swift` (inside any of the examples) and then, say, `cargo run -- exec example.swift` builds and runs without errors. I have the latest Xcode 11.6 installed on macOS 10.15.4; haven't installed the 10.15.5 update yet.

```
$ swiftc --version
Apple Swift version 5.2.4 (swiftlang-1103.0.32.9 clang-1103.0.32.53)
Target: x86_64-apple-darwin19.4.0
```

From Googling around, I found [this ticket](https://gitlab.kitware.com/cmake/cmake/-/issues/19880) and [this comment](https://github.com/vapor/vapor/issues/177#issuecomment-225352975); do either of those help?

jhugman:
Ok, I think I got it.

I had an old version of `swiftc` on my `PATH`.

```sh
$ swiftc -version
Apple Swift version 5.1.3 (swiftlang-1100.0.282.1 clang-1100.0.33.15)
Target: x86_64-apple-darwin19.6.0

$ realpath swiftc
/Library/Developer/CommandLineTools/usr/bin
```

```sh
$ xcrun swiftc -version
Apple Swift version 5.2.4 (swiftlang-1103.0.32.9 clang-1103.0.32.53)
Target: x86_64-apple-darwin19.6.0

$ xcrun --find swiftc
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swiftc
```

The fixes I can suggest is: 

 (1) For me to fix my path. ✅ 
 (2) Use `xcrun swiftc` instead of `swiftc` ✅
 (3) Detect we're not on a macos or we haven't got `xcrun` installed and do something appropriate.

Would you be happy for me to do (2) in as a side benefit of #15, and I'll file an issue for 3. Alternatively, I can file an issue and fix it separately.

rfk:
Just a note that over in https://github.com/rfk/uniffi-rs/pull/18 I'm using swiftc on ubuntu for a bit of simple CI. It would be nice not to break that if possible (but we should optimize for develop experience where it's in conflict with CI, if necessary)

jhugman:
>  I'm using swiftc on ubuntu for a bit of simple CI.

That's helpful to know! Thanks.

Issue #3, opened 2020-06-23
Add support for strings in arguments, return values, and record fields


rfk:
The example code so far only really exercises support for numbers. In practice we have a lot of code that uses strings, so we need to be able to deal with those conveniently and safely.

Adding support for strings would be a good opportunity for someone new to this codebase to get involved. They would need to:

* Think up a new [example](https://github.com/rfk/uniffi-rs/tree/master/examples) for some small code that uses strings, and try writing the IDL and the corresponding rust code for such a component. (It won't work yet, but having a specific example to get working is a good place to start).
  * Try to `cargo build` your example component, and confirm that it doesn't compile. If it *does* compile despite not having support for strings, that's a separate bug...
* Decide on how Strings should be passed across the FFI layer, both as individual arguments and when part of a compount record. Implement the rust half of that via the traits in [`uniffi::support`](https://github.com/rfk/uniffi-rs/blob/master/src/support/mod.rs).
  * My suggestion (which isn't necessarily the best way!) would be to
    * Implement `Lowerable` by writing out a fixed-size length field, followed by the body of the string.
    * Implement `Liftable` by doing the reverse, reading the fixed-size length followed by the string data.
    * Implement `ViaFfi` using the [`ffi_support::FfiStr`](https://docs.rs/ffi-support/0.4.0/ffi_support/struct.FfiStr.html) helper, which is what we use for strings in our hand-written rust component today.
* Add support for string types in [`uniffi:scaffolding`]. There's some placeholder stuff in there using `&str` and `FfiStr<>` but it'll probably need adjusting.
  * Try to get to a point where you can `cargo build` your example component and have it work successfully.
* Add support for strings in the [kotlin](https://github.com/rfk/uniffi-rs/blob/master/src/bindings/kotlin/gen_kotlin.rs) and [python](https://github.com/rfk/uniffi-rs/blob/master/src/bindings/python/gen_python.rs) bindings generators. The details here will depend on how strings are represented when crossing the FFI.
    * If you used `FfiStr` above, you might find that passing strings are arguments just works, because the native-code-binding layers of each language know how to pass a string object as a `const char*` in their FFI layers.
    * You'll need to add support for lifting and lowering strings as part of compound records; copy the approach taken for other primitive data types in these files.
    * Receiving strings as a return value and making sure they show up as native language string objects, is left as an exercise to the reader.

rfk:
[...]

PR #9, opened 2020-06-24
WIP experiment with a custom IDL

rfk:
I mentioned this in another issue recently, so figured I may as well push it for visibility.

So far, this project has very explicitly avoided trying to invent a custom DSL for describing component interfaces, and instead used whatever is available from WebIDL using an existing parser. This has been very helpful in getting things off the ground to quickly show progress, but I've come to realize two things:

* WebIDL is at best an "ok" fit for our needs.
* The [`nom`](https://docs.rs/nom) crate can be used to make surprisingly legible parser code

So I wanted to see what it would look like to use an IDL that's more directly suited to our needs, and whether the increase in legibility of the `.idl` files would offset the costs of maintaining a custom language. So far I'm a big fan of `nom`, but undecided on the cost/benefit tradeoff. I plan to keep poking at this branch to get it fully working on the examples and then we can see what we reckon.

(Also, I wanted to use this as an excuse to get more familiar with using rustdoc effectively).


src/interface/idl.rs
```diff
@@ -0,0 +1,594 @@
...
+//!   defined using the keywords below.
+//! * At the top level of the file we expect a single [`module`](ModuleDefinition)
+//!   definition, which provides metadata about the component as a whole and serves
+//!   as a namespace for the functionality to be exported.
+//! * The [`enum`](EnumDefinition) keyword defines a C-style enum as a new named type,
+//!   which can take on one of a fixed set of named values.
+//! * The [`record`](RecordDefinition) keyword defines a compound data type made up
+//!   of one or more named [fields](FieldDefinition), each with a particular type.
+//! * The [`class`](ClassDefinition) keyword defines a new opaque data type that
+//!   encapsulates state and behvaiour, and provides named [methods](MethodDefinition)
```
linacambridge, line 90:
```suggestion
//!   encapsulates state and behaviour, and provides named [methods](MethodDefinition)
```

src/interface/idl.rs
```diff
@@ -0,0 +1,594 @@
...
+
+/// The [`enum`](EnumDefinition) keyword defines a C-style enum as a new named type.
+/// Each enum has a fixed number of options which can be specified by name. It is not
+/// possible to set or to example the internal representation of the enum.
+///
+/// ```text
+/// enum exampleNameOfEnum {
+///      /* An identifier for each option, separated by commas */
+///     OPTION_ONE,
+///     OPTION_TWO,
```
linacambridge, line 185: Oooh, I spy a trailing comma, nice!

src/interface/idl.rs
```diff
@@ -0,0 +1,594 @@
...
+            map(kwd("u16"), |_| TypeReference::U16),
+            map(kwd("i16"), |_| TypeReference::I16),
+            map(kwd("u32"), |_| TypeReference::U32),
+            map(kwd("i32"), |_| TypeReference::I32),
+            map(kwd("u64"), |_| TypeReference::U64),
+            map(kwd("i64"), |_| TypeReference::I64),
+            map(kwd("float"), |_| TypeReference::Float),
+            map(kwd("double"), |_| TypeReference::Double),
+            map(
+                tuple((
```
linacambridge, line 276: It’s kind of a gratuitous change, but how do you feel about copying Swift’s `[Type]` syntax (or `Type[]`) for sequences, instead of `sequence<Type>`? I could make up a reason about being able to better represent fixed-size arrays as an optimization (`[Type; 12]`), but really it’s because I’ve never been very fond of angle brackets 😆

On the other hand, `sequence<Type>` is nice because it’s consistent with WebIDL...using another syntax would be a divergence for no meaningful reason.

src/interface/idl.rs
```diff
@@ -0,0 +1,594 @@
...
+    Double,
+    Nullable(Box<TypeReference<'src>>),
+    Sequence(Box<TypeReference<'src>>),
+    Named(&'src str),
+}
+
+impl<'src> TypeReference<'src> {
+    fn parse_from(input: &'src str) -> ParseResult<'src, Self> {
+        let (input, type_) = alt((
+            map(kwd("boolean"), |_| TypeReference::Boolean),
```
linacambridge, line 264: Another gratuitous suggestion: how do you feel about `bool`, `f32`, and `f64` to match Rust?

(And thanks for not calling `u64` `unsigned long long`! 😆❤️)

src/interface/idl.rs
```diff
@@ -0,0 +1,594 @@
...
+                arguments,
+                return_type,
+            },
+        )(input)
+    }
+}
+
+/// Parses a comma-separated list of items from a given parser.
+///
+/// With optional trailing comma, because optional trailing commas are cool.
```
linacambridge, line 442: Yassssss! 😍

src/interface/idl.rs
```diff
@@ -0,0 +1,594 @@
...
+}
+
+/// Check whether a character can be part of an identifier.
+fn is_identifier_char(c: char) -> bool {
+    c.is_alphanumeric() || c == '_'
+}
+
+/// Parses an identifier: a sequence of characters from [A-Za-z0-9_].
+///
+/// XXX TODO: should not be allowed to start with a number, or match a keyword.
```
linacambridge, line 492: I was just going to comment about that, nice!

src/interface/idl.rs
```diff
@@ -0,0 +1,594 @@
...
+    I64,
+    Float,
+    Double,
+    Nullable(Box<TypeReference<'src>>),
+    Sequence(Box<TypeReference<'src>>),
+    Named(&'src str),
+}
+
+impl<'src> TypeReference<'src> {
+    fn parse_from(input: &'src str) -> ParseResult<'src, Self> {
```
linacambridge, line 262: Teaching `TypeReference` about strings would be fun, as part of #3!

thomcc:
Just nipping in to say: An issue I've had in the past with nom-based parsers is that it's hard to teach them to generate good error messages on failure. You generally can't get error line/col info, let alone a span. That might be fixed now, since the last time I used it was when it was heavily macro based.

rfk:
> it's hard to teach them to generate good error messages on failure

Ack, thanks, good to know. I was pleased to discover that the non-macro-based version of `non` seemed a lot more approachable in general, I'll see what I can find out about error generation specifically.

Another thought I have here is: if we were going to do all this work on parsing stuff, should we instead be investing it in a proc-macro that can generate direct from annotations in the rust code? I have some philosophical reasons why I don't think we should do that *at this stage* but it's a question worth asking.

thomcc:
> if we were going to do all this work on parsing stuff, should we instead be investing it in a proc-macro that can generate direct from annotations in the rust code

This was my initial thought about this.

I think the answer is a strong yes, for tooling reasons. Using proc-macros allows us to reuse the existing rust tooling, and allows rust-analyzer, rls, intellij-rust, etc. to help with this sort of code

None of those tools currently run build.rs scripts (my understanding is that this is unlikely to be a situation that changes any time soon), and even proc macros they're often very lazy about running, but by and large they'd be able to "ignore" these sorts of proc macros and get Rust code they can understand.

rfk:
My weakly-held opinions on proc-macros vs IDL file:

* Using a separate IDL file means we're more likely to think about the interface exposed to consumer first, and treat rust as an implementation detail; using macros means we're more likely to make a nice rust API and then work out how to expose it to consumers. (IOW, the difference between "implementing cross-language components in rust" and "exposing rust components to multiple languages", and it's my opinion that we should be doing the former).
* Proc-macros feel scary to write and maintain (but maybe they're not that much scarier than a custom parser in practice).

But I am highly open to being convinced otherwise! Your point about tooling is one that I had not considered in any detail.

thomcc:
> Using a separate IDL file means we're more likely to think about the interface exposed to consumer first, and treat rust as an implementation detail

I don't really disagree, but I think in practice this isn't going to be a huge issue. Specifically, I'm not sure how concretely the two things you're distinguishing will look much different in practice.

> Your point about tooling is one that I had not considered in any detail.

It's one of my largest concerns with uniffi, to be honest. IMO It's pretty painful working without IDE support.

> Proc-macros feel scary to write and maintain (but maybe they're not that much scarier than a custom parser in practice).

They're probably less scary in practice compared to a parser that does proper error span reporting. That stuff comes more-or-less for free with proc macros.

linacambridge:
Oooh, that's a really cool idea! It does feel like pulling the interface from the Rust component would remove a layer of duplication.

Another thought, almost completely the other way—what if we generated the Rust library from the IDL, too? Kind of like how WebIDL bindings are generated in Firefox...

thomcc:
> what if we generated the Rust library from the IDL, too? Kind of like how WebIDL bindings are generated in Firefox...

This is the way that leads to the worst support for IDE integration, sort of the way that it is working with this stuff inside desktop firefox (which IMO is much more painful for this reason than working in app-services)

rfk:
> Another thought, almost completely the other way—what if we generated the Rust library from the IDL, too?

Earlier versions of the prototype tried this, e.g. if you declared an `enum` in the WebIDL then it would just generate all the rust code for that enum as part of the scaffolding, saving you from having to specify it again in rust.

I found the developer experience on the Rust side kind of awkward/confusing in practice - you end up writing rust code for the functions and classes, that references a bunch of types that aren't visible in your code at all, and are instead magicked into existence by some other part of the build. It worked OK, it just felt a bit weird. Some brief commentary on the commit that changd it to its current form here https://github.com/rfk/uniffi-rs/commit/8d9c203d33900417d5b97e03a6093734e468fd8b.

YMMV of course.

linacambridge:
Ooof, that sounds like the worst of all worlds! Let's not do the clunky "magical" experience with bad IDE support.

rfk:
@thomcc I think you have convinced me that it's worth trying a timeboxed experiment with using a proc-macro for this, to get a feel for what the tradeoffs of the different approaches would be in practice.

Not sure when I'll find time for that myself, and I would welcome other folks experimenting with it if it seems interesting to you. We don't have to commit to an approach with any particular urgency just yet, it feels worthwhile to explore the design space here.

thomcc:
I'll play around with it when I get a chance (although I'm unsure when that will be). That said, I just wanted to chime in that I had concerns here. I think the current approach (including using a custom IDL) is probably fine in the short term, so long as we avoid the half-generated-rust case which would really *remove* IDE support we already have.

I think there is a lot of stuff we should do around making sure uniffi is easy to debug, though.

jhugman:
Just popping in with my drive by, and largely disconnected thoughts: 

 * even just working on toy examples, duplicating method signatures and properties in rust and IDL feels like mindless duplication. As a library developer, I'd strongly favour the source of truth to be in Rust.
 * Lots of error detection, making of nice error messages, line numbers etc just go away if you use someone else's compiler. While writing parsers is fun and feels productive, supporting and extending it over time is a pain.
 * As a library consumer, I'd like the library developers to notice when the API breaks, before they ship breaking changes. Being able to diff an interface definition file (whether in IDL or rust) is useful too.
 * A `proc_macro` approach means we'd have to split up the foreign language compilation step from the rust compilation step. This may not be a bad thing— it is already, but foreign language source generation and compilation being done in two different stages would be different.
 * This may not be actually be a concern: covering an existing rust library with FFIs needs the upstream dev to use uniffi.rs' `proc_macro`s, or a further rust wrapper around the unmodified library.
 * Generating a hand-writable IDL file with a `proc_macro`, and then using uniffi.rs to compile the FFI artifacts may be the way forward here.

I'm actually quite in favour of a `proc_macro` approach; and that it should be the place we end up at, so am surprised to come up with all the reasons not to use them.

I think there are going to be impedance mismatches all over this project, because the foreign languages and rust are all different in both subtle and large ways. Adding another one, by using a separate definition language means either crappy tooling (already mentioned) or needlessly low lowest-common-denominator interfaces.

rfk:
FYI I took a bit of time to mess around with proc macros, and pushed the (very incomplete!) results over in https://github.com/rfk/uniffi-rs/pull/23 for visibility, along with a bit of commentary.

rfk:
This generated some good discussions and was a fun experiment, but it feels like we're all more-or-less agreed that we should stick with WebIDL in the short term, and pursue a macro-based approach in the medium- to long-term. So, I'm going to close this PR. Thanks all!

Issue #15, opened 2020-07-09
Refactor templates to increase template re-use and make them easier to navigate


jhugman:
There are quite a few places where we're doing the same thing, but with minor differences in the header or footer.

e.g. 

 * unpacking arguments, calling method calls with and without a return value
 * methods in enums, namespaces, interfaces, etc etc.
 * iterating through properties to generate lift and lower methods.
 * copyright & similar comments

We should investigate ways our templates could be split up and parameterized so working with templates easier for language experts and more readable.

1. Decorator `stucts` and [rendering templates in place][rip].
2. Using the [`path` sub-attribute][path] of the `template` attribute.
3. [`include` directive][incl]. e.g. `{% include "hotmess.txt" %}`
4. [Template inheritance][inh].
5. [Template importing][imp].

This is likely a starting point / meta for smaller bugs rather than a specific piece all on its own.

[rip]: https://github.com/djc/askama/blob/master/testing/templates/render_in_place.html
[path]: https://github.com/djc/askama/blob/master/book/src/creating_templates.md
[incl]: https://github.com/djc/askama/blob/master/testing/templates/include.html
[imp]: https://github.com/djc/askama/blob/master/testing/templates/import.html
[inh]: https://github.com/djc/askama/blob/master/testing/tests/inheritance.rs

Issue #7, opened 2020-06-24
Add support for returning and handling errors


linacambridge:
We have lots of `TODO`s for adding error handling. A few things we’ll want to think about:

* How do we represent errors in WebIDL? As `Result` return values? Following how Firefox does it, using a `Throws` annotation on fallible methods? Do we want to have structured errors that can be inspected, or are error strings good enough?
* In the FFI, we’ll need to have fallible methods take an error out parameter.
* Swift and Kotlin specify fallible functions differently—Swift has the `throws` keyword to indicate a function is fallible, and Kotlin and Python functions can throw exceptions. We will need to teach our code generation layer to reflect error results as exceptions.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1566)

rfk:
> Following how Firefox does it, using a Throws annotation on fallible methods?

This was my first instinct as well.

But I also have to be honest, I've been spending my evenings exploring what it would look like to use our own custom variant of an IDL rather than just relying on what's available in WebIDL...which might come to nothing or might change how we want to proceed here. We should not block on that though :-)

> Do we want to have structured errors that can be inspected, or are error strings good enough?

I think we need some structure, e.g. in the fxa-client the kotlin code needs to be able to distinguish between a network error and a hard failure.

Strawman: `[Throws(EnumName)]` would annotate a function/method that can throw, and the resulting error would have both a string and an integer errcode that maps to the specified enum. This is basically how our existing `ExternError` class works in the hand-written bindings IIUC, except you have to know the possible range of error codes via out-of-band means.

tarikeshaq:
I'm thinking of taking a look at this, but before I go in over my head, I wanted to dump my thought process in case I have something off (or maybe there are blockers I haven't thought about or something)

### The WebIDL would look something like this:
```idl
enum Error {
   "NETWORK_ERROR", "TYPE_ERROR"
};

interface InterfaceName {
   constructor();
   [Throws(Error)]
   string func();
};
```
- From that, we can add a `throws` `Option<T>` in here: https://github.com/rfk/uniffi-rs/blob/6a399f8a2e42893fa99d3894ee1d75afb042e73d/uniffi/src/interface/mod.rs#L590-L594 That gets populated with the throws attribute (Haven't looked at the details here)

This should give us all the information needed to form the scaffolding and the bindings (right?)

### From the scaffolding side
- Add the `error: &mut ExternError` argument to the FFI functions
[...]

Issue #4, opened 2020-06-23
Add support for reference arguments in the rust code


rfk:
Currently, every rust function exposed over the FFI is expected to take ownership of its arguments. This is most obvious in the existing "sprites" example, where the code for [`translate`](https://github.com/rfk/uniffi-rs/blob/f28adde220adfcb01c2732b23d7c857010a0d268/examples/sprites/src/lib.rs#L21) has to be declared as:

* `fn translate(p: Point, v: Vector) -> Point`

Which means that when the `move_to` method of the `Sprite` class calls it, it has to [pass in a clone](https://github.com/rfk/uniffi-rs/blob/f28adde220adfcb01c2732b23d7c857010a0d268/examples/sprites/src/lib.rs#L51):

* `        self.current_position = translate(self.current_position.clone(), direction)`

Making this more egonomic could be a good opportunity for somone new to the codebase to get familiar with it. The challenge would be to update the sprites example so that the `translate` function is declared as:

* `fn translate(p: &Point, v: &Vector) -> Point`

And the `.clone()` mentioned above can be removed.

In order to achieve this I think you would need to:

* Dig in to the [code that generates the ffi wrappers](https://github.com/rfk/uniffi-rs/blob/f28adde220adfcb01c2732b23d7c857010a0d268/src/scaffolding.rs#L103) that call these functions, to see where to pass the argument by reference instead of by value.
* Observe that you need some flag in the IDL to know whether to pass each argument by reference or by value, and that this will probably need to be added to the [`Argument`](https://github.com/rfk/uniffi-rs/blob/f28adde220adfcb01c2732b23d7c857010a0d268/src/interface/mod.rs#L360) struct so that it's available when you're generating the code.
* Observe that there is [some placeholder code](https://github.com/rfk/uniffi-rs/blob/f28adde220adfcb01c2732b23d7c857010a0d268/src/interface/mod.rs#L438) on the `Argument` struct that currently says "argument attributes are not yet supported", which corresponds to the [extended attributes](https://heycam.github.io/webidl/#idl-extended-attributes) feature of WebIDL.
* Experiment with adding an extended attribute to the arguments of the `translate` function in [`sprites.idl`](https://github.com/rfk/uniffi-rs/blob/master/examples/sprites/src/sprites.idl), something like `Point translate([ByRef] Point position, [ByRef] Vector direction)`.
[...]

PR #17, opened 2020-07-10
Breakup code-generation templates in to smaller files

jhugman:
This is the beginning of a refactor of code-generation templates in to smaller, more re-usable files, suggested by #15.

It occurred to me that it's quite a big change to anyone working on the backend, and we should merge it quickly so unrelated backend work can proceed in parallel with minimal discomfort (e.g. #7, #3, #4 ).

On the whole, I've resisted the urge to do anything other than move templates around and split them up. This would include: making the generated code more idiomatic, finding template re-use opportunities, making the filters more uniform etc. I'll file separate issues as followups.


src/bindings/kotlin/templates/wrapper.kt
```diff
@@ -0,0 +1,50 @@
...
+{% for e in ci.iter_enum_definitions() %}
+{% include "EnumTemplate.kt" %}
+{%- endfor -%}
+
+{%- for rec in ci.iter_record_definitions() %}
+{% include "RecordTemplate.kt" %}
+{% endfor %}
+
+{% for func in ci.iter_function_definitions() %}
+{% include "TopLevelFunctionTemplate.kt" %}
```
rfk, line 45: IIUC this is more-or-less a literal text inclusion like `#include` or similar, and the including template gets access to all the visible names at this point. Accurate?

jhugman: Yes, that was my understanding too.

jhugman:
Work left to do immediately:

 * [ ] break up `scaffolding.rs` in the same way,

I don't think waiting on this work should delay landing this, given that it's so disruptive to other PRs. 

Other things I'm noodling on, which should definitely wait for another issue: 

 * structs/templates for language specific `FunctionArgsDecl`, `FunctionCallArgs`, for declaring and calling constructors, methods and functions.

src/templates/RecordTemplate.rs
```diff
@@ -0,0 +1,27 @@
+{#
```
linacambridge, line 1: OOC, how come this (and inside `TopLevelFunctionTemplate`) is in a special `{#` comment block, but the others aren’t? Because we don’t want to include this in the generated output, it’s for the benefit of whoever’s editing the template?

rfk: That's my interpretation FWIW; I think it makes sense considering that this comment was moved from outside of the top-level `for rec in ci.iter_record_definitions()` to inside it, so if we left it as a plain comment we'd get multiple copies of it, one per record.

src/templates/scaffolding_template.rs
```diff
@@ -0,0 +1,52 @@
...
+// order of items *as declared in the IDL file*. This might be different to the order
+// of items as declared in the rust code, but no harm will come from it.
+{% for e in ci.iter_enum_definitions() %}
+{% include "EnumTemplate.rs" %}
+{% endfor %}
+
+// For each record declared in the IDL, we assume the caller has provided a corresponding
+// rust `struct` with the declared fields. We provide the traits for sending it across the FFI.
+// If the caller's struct does not match the shape and types declared in the IDL then the rust
+// compiler will complain with a type error.
```
rfk, line 20: Although the comment seems to be duplicate here as well, might be worth removing from here.

rfk:
> Merge and iterate would be my preference.

Let's do it!

Issue #5, opened 2020-06-23
Add testing infrastructure for running examples as fixtures.


jhugman:
From #2 

> I'd like to be able to turn the existing example.py and example.kts files into actual tests with assertions that can be executed to make sure the generated bindings are working as expected. So far I've just been running them by hand and eyeballing the output, but that's clearly not going to scale.

rfk:
This was done in #18.

PR #18, opened 2020-07-11
Hook up a basic foreign-language test harness

rfk:
This is a pitch for how we might do #5, turning our little suite of example foreign-language scripts into actual testcases. It works well enough for me to show it for some feedback, but if we like it, it still needs a bit of polish before merge.

Broad strokes of the PR:

* The `uniffi` crate has been moved into a subdirectory, and it is now part of a single cargo workspace along with the example crates.
* There is a new module `uniffi::support::tests` which builds on the existing ability to run foreign language scripts via an interpreter, and exposes some methods for executing a given script file as a testcase. There is some refactoring of existing modules to make this work nicely, including a significant cleanup of the `uniffi::bindings::*` modules for consistency.
* There is a new crate `uniffi_macros` that exports a single macro, `build_foreign_language_testcases!()`. The idea is that you give it a list of files and it turns them into a suite of `#[test]` functions, one per file.
  * To be honest this probably doesn't need to be a macro, or at least not a procedural macro. But I wanted a small excuse to play around with procedural macros, and I quite like the result.
* The example scripts from `./examples/arithmetic` have been converted into concrete testcases. Not that they really test very much, but they should give you an idea of the resulting consumer experience.
  * If we like the shape of this, I'll have a go at converting the other example crates as well.
* As a side-effect of all this shuffling around, I've also addressed some clippy lints and general code cleanup (to give that satisfying clean `cargo build`-without-warnings feeling).

All of this is geared around a single high-level goal: make `cargo test` in the top-level directory behave in a useful and sensible manner. Doing `cargo test` runs the full test suite for all examples in all target languages. Doing `cargo test py` runs only the python-related tests. And so-on.

Thoughts @linacambridge @jhugman @thomcc?

---

TODO before merge:

* [x] convert the remaining example scripts into testcases.
* [x] update various readmes/docs to mention the new structure.


uniffi/src/bindings/kotlin/mod.rs
```diff
@@ -56,10 +48,8 @@ pub fn compile_kotlin_bindings(ci: &ComponentInterface, out_dir: &OsStr) -> Resu
         .arg(&kt_file)
         .arg("-d")
         .arg(jar_file)
-        .spawn()
-        .map_err(|_| anyhow::anyhow!("failed to spawn `kotlinc`"))?
-        .wait()
-        .map_err(|_| anyhow::anyhow!("failed to wait for `kotlinc` subprocess"))?;
+        .spawn()?
+        .wait()?;
```
rfk, line 52: Turns out I had a while bunch of spurious `map_err` throughout the codebase; `std::io::Error` is `Into<anyhow::Error>` so they're not needed.

uniffi/src/bindings/swift/mod.rs
```diff
@@ -112,68 +106,43 @@ pub fn compile_swift_module(ci: &ComponentInterface, out_dir: &OsStr) -> Result<
...
-            .flat_map(|module_map| {
-                let mut module_map_file = out_path.clone();
-                module_map_file.push(module_map);
-                if !module_map_file.exists() {
-                    // Missing module maps (for example, `uniffi_arithmetic` when we're
-                    // running the `geometry` example) will fail, so ignore maps that
-                    // don't exist. Gross!
-                    return Vec::new();
+        cmd.arg("-I").arg(out_dir).arg("-L").arg(out_dir);
+        for entry in PathBuf::from(out_dir).read_dir()? {
```
rfk, line 123: @linacambridge here I've tried to address https://github.com/rfk/uniffi-rs/issues/16 by looking for files in the target directory, instead of hard-coding them. It seems to work in my local testing. WDYT?

uniffi/src/lib.rs
```diff
@@ -27,7 +27,7 @@
 //! in a `.idl` file, where it can be processed by the tools from this crate.
 //! For example you might define an interface like this:
 //!
-//! ```
+//! ```text
```
rfk, line 30: Otherwise `cargo test` tries to run these as doctests by default, which doesn't end well.

uniffi/src/lib.rs
```diff
@@ -258,110 +262,64 @@ fn run_bindings_generate_subcommand(default_lib_file: Option<std::ffi::OsString>
...
+    let target_dir = match default_lib_file {
+        None => None,
+        Some(lib_file) => {
+            let lib_file = PathBuf::from(lib_file);
+            match lib_file.parent() {
+                None => None,
+                Some(target_dir) => {
+                    let ci = bindings::get_component_interface_from_cdylib(&lib_file)?;
+                    bindings::write_bindings(&ci, target_dir, lang)?;
+                    bindings::compile_bindings(&ci, target_dir, lang)?;
```
rfk, line 316: This now re-generates and re-compiles the bindings whenever you `cargo run -- exec` a script, but I think that's an improvement in ergonomics overall, as you don't have to remember to do that by hand.

uniffi/src/bindings/kotlin/mod.rs
linacambridge: I’m wondering if this will still give us nice context for errors—“failed to wait for `kotlinc` subprocess” pinpoints the problem much better than some generic “process exited with status code 1337”. Maybe Anyhow’s [`context` and `with_context`](https://docs.rs/anyhow/1.0.13/anyhow/trait.Context.html) are the ticket there. Not saying we need to sprinkle those everywhere, just that I liked your old error messages too! 😸

uniffi/src/bindings/kotlin/mod.rs
```diff
@@ -69,40 +59,32 @@ pub fn compile_kotlin_bindings(ci: &ComponentInterface, out_dir: &OsStr) -> Resu
...
-                    if ext == "jar" {
-                        classpath.push(":");
-                        classpath.push(entry.path().to_str().unwrap());
-                    }
+        for entry in PathBuf::from(out_dir).read_dir()? {
+            let entry = entry?;
+            if let Some(ext) = entry.path().extension() {
+                if ext == "jar" {
+                    classpath.push(":");
+                    classpath.push(entry.path().to_str().unwrap());
```
linacambridge, line 74: I think `entry.path().as_os_str()` will give you an `OsStr` that you can append without converting and unwrapping.

uniffi/src/bindings/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+pub mod kotlin;
+pub mod python;
+pub mod swift;
+
+/// Enumeration of all foreign language targets currently supported by this crate.
+///
+/// The functions in this module will delegate to a language-specific backend based
+/// on the provided `TargetLanguage`. For convenience of calling code we also provide
+/// a few `TryFrom` implementations to help guess the correct target language from
+/// e.g. a file extension of command-line argument.
```
linacambridge, line 26: Nice refactoring! 🚀

uniffi/src/lib.rs
linacambridge: 👍 When I was working on Swift, I’d forget to do `cargo run -- generate` before `exec` all the time. Definitely an improvement!

uniffi/src/support/tests.rs
```diff
@@ -0,0 +1,135 @@
...
+    P1: AsRef<Path>,
+    P2: AsRef<Path>,
+    L: TryInto<bindings::TargetLanguage, Error = anyhow::Error>,
+{
+    let cdylib_file: PathBuf = cdylib_file.as_ref().to_path_buf();
+    let out_dir: PathBuf = out_dir.as_ref().to_path_buf();
+    let language: bindings::TargetLanguage = language.try_into()?;
+    // Have we already generated these bindings?
+    let mut generated_bindings = GENERATED_BINDINGS.lock().unwrap();
+    // XXX TODO: figure out correct type of `HashSet` so that I don't need to `clone()` here...
```
linacambridge, line 125: Ooh, good question! I wonder if this will need a tuple struct with a custom `Borrow<&(&PathBuf, &PathBuf, &TargetLanguage)>` implementation, or if you can sprinkle the right combination of `&`s  that’ll make the compiler understand this...

uniffi_macros/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+                .file_name()
+                .expect("Test file has no name, cannot build tests for generated bindings")
+                .to_string_lossy();
+            let test_name = format_ident!(
+                "uniffi_foreign_language_testcase_{}",
+                test_file_name.replace(|c: char| !c.is_alphanumeric(), "_")
+            );
+            quote! {
+                #[test]
+                fn #test_name () {
```
linacambridge, line 50: Could this return a `Result` instead of unwrapping to retain the stack trace if the test case throws?

uniffi_macros/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+use std::env;
+use std::path::PathBuf;
+use syn::{punctuated::Punctuated, LitStr, Token};
+
+/// A macro to build testcases for a component's generated bindings.
+///
+/// This macro provides some plumbing to write automated tests for the generated
+/// foreign language bindings of a component. As a component author, you can write
+/// script files in the target foreign language(s) that exercise you component API,
+/// and then call this macro to produce a `cargo test` testcase from each one.
```
linacambridge, line 20: Oooh, this is exciting!

uniffi/src/bindings/swift/mod.rs
```diff
@@ -112,68 +106,43 @@ pub fn compile_swift_module(ci: &ComponentInterface, out_dir: &OsStr) -> Result<
...
+            let entry = entry?;
+            if let Some(ext) = entry.path().extension() {
+                if ext == "modulemap" {
+                    let mut option = OsString::from("-fmodule-map-file=");
+                    option.push(entry.path());
+                    cmd.arg("-Xcc");
+                    cmd.arg(option);
+                } else if ext == "dylib" || ext == "so" {
+                    let mut option = OsString::from("-l");
+                    option.push(entry.path());
```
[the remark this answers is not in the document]
rfk: I didn't realize it either, I accidentally tried it and it appeared to work so... 🤷‍♂️

rfk:
Note to self: the change to make this all compile in a single workspace, doesn't play well with swift bridging headers; building multiple components give me this error: 

```
/Users/rfk/repos/rfk/uniffi-rs/target/debug/uniffi_sprites.modulemap:4:21: error: umbrella for module 'uniffi_geometry' already covers this directory
    umbrella header "/Users/rfk/repos/rfk/uniffi-rs/target/debug/sprites-Bridging-Header.h"
```

rfk:
> the change to make this all compile in a single workspace, doesn't play well with swift bridging headers

I fixed this with a bit of subdirectory chicanery, it's not pretty but it seems to work.

uniffi/src/bindings/kotlin/mod.rs
rfk: Thanks; turns out plain old `entry.path()` also works

.circleci/config.yml
```diff
@@ -22,7 +24,9 @@ jobs:
   Rust and Foreign Language tests:
     docker:
       - image: rfkelly/uniffi-ci:latest
+    resource_class: small
     steps:
+      - run: cat ~/.profile >> $BASH_ENV
```
rfk, line 29: I added a `.circleci/config.yaml` directly on master so that circle would start building things, but I'd welcome any feedback on it as part of this PR.

docker/Dockerfile-build
```diff
@@ -57,4 +57,14 @@ RUN mkdir -p /tmp/setup-kotlin \
...
+
+RUN mkdir -p /tmp/setup-jna \
+    && cd /tmp/setup-jna \
+    && curl -o jna.jar https://repo1.maven.org/maven2/net/java/dev/jna/jna/5.6.0/jna-5.6.0.jar \
+    # XXX TODO: should check a sha256sum or something here...
+    && sudo mv jna.jar /opt \
+    && echo "export CLASSPATH=\"\$CLASSPATH:/opt/jna.jar\"" >> /home/circleci/.bashrc \
+    && echo "export CLASSPATH=\"\$CLASSPATH:/opt/jna.jar\"" >> /home/circleci/.profile \
+    && cd ../ \
+    && rm -rf ./setup-jna
```
rfk, line 70: I also added a dockerfile that builds all the tools we need, but I forgot to install JNA on my first attempt, so here it is as part of this PR :-)

If you're curious about the whole file, you can view it [here](https://github.com/rfk/uniffi-rs/blob/becd87c30bceca0af989266ca75633cc7059fd6f/docker/Dockerfile-build).

examples/naming-conventions/Cargo.toml
```diff
@@ -1,19 +0,0 @@
-[package]
-name = "uniffi-example-naming-conventions"
```
rfk, line 2: Perhaps controversially, I have deleted the `naming-conventions` example. Reason being that, when I added these tests, I discovered that the `sprites` example was failing in Kotlin due to not being updated for the naming conventions. So I think we can safely check the behaviour of those conventions through the existing examples.

rfk:
@tarikeshaq heads-up, if you do end up working on adding string support, you might like to base it on top of this branch in order to get some basic automated test infra.

rfk:
Alrighty, I'm happy enough with how this works to take it out of "draft". @linacambridge I'm aware you're out for the next few days so no rush, but please take another look over this at your convenience.

uniffi/src/bindings/swift/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+/// declarations, and a `.modulemap` file to tell swift how to use it.
+///
+/// Further complicating matters is that fact that swift doesn't seem to like having multiple
+/// `.modulemap` files referring to headers in the same directory (if you try it, it will complain
+/// about multiple umbrella headers covering the same directory). That's not a problem for
+/// stand-alone crates but it doesn't work well when you have multiple uniffi components defined
+/// in the same workspace.
+///
+/// We work around this by generating the module-mapping instructions in a separate subdirectory
+/// for each module, producing a directory structure that looks like this:
```
rfk, line 38: @linacambridge in particular could you please take a look at the yucky things I had to do here to make this work in a cargo workspace? You may be able to see a better way.

uniffi/src/lib.rs
```diff
@@ -205,11 +206,7 @@ pub fn run_bindgen_helper(component_name: Option<&str>) -> Result<()> {
                 .arg(
                     clap::Arg::with_name("lib_file")
                         .takes_value(true)
-                        .required(if let Some(_) = default_lib_file {
-                            false
-                        } else {
-                            true
-                        })
+                        .required(default_lib_file.is_some())
```
tarikeshaq, line 209:
```suggestion
                        .required(default_lib_file.is_none())
```

Realized while adding the tests for strings that I had to manually set the lib_file when running `generate` 😄

rfk: nice catch, thanks! 😅

uniffi/src/bindings/kotlin/mod.rs
```diff
@@ -57,9 +48,9 @@ pub fn compile_kotlin_bindings(ci: &ComponentInterface, out_dir: &OsStr) -> Resu
         .arg("-d")
         .arg(jar_file)
         .spawn()
-        .map_err(|_| anyhow::anyhow!("failed to spawn `kotlinc`"))?
+        .context("Failed to `kotlinc` to compile the bindings")?
```
linacambridge, line 51:
```suggestion
        .context("Failed to spawn `kotlinc` to compile the bindings")?
```

uniffi/src/bindings/kotlin/mod.rs
```diff
@@ -78,30 +69,30 @@ pub fn run_kotlin_script(out_dir: Option<&OsStr>, script_file: Option<&str>) ->
...
-            .map_err(|_| anyhow!("failed to read directory {:?}", out_dir))?
+            .context("Failed to list target directory when running Kotlin script")?
         {
-            if let Ok(entry) = entry {
-                if let Some(ext) = entry.path().extension() {
-                    if ext == "jar" {
-                        classpath.push(":");
-                        classpath.push(entry.path().to_str().unwrap());
-                    }
+            let entry = entry.context("Directory listing failed while running Koltin script")?;
```
linacambridge, line 74:
```suggestion
            let entry = entry.context("Directory listing failed while running Kotlin script")?;
```

uniffi/src/bindings/swift/mod.rs
linacambridge: I think we can actually get rid of the yucky stuff if we change `umbrella header` to just `header` in the `.modulemap` (that's probably technically more correct, too—"umbrella header" means "includes all the other headers in that directory", which our `-Bridging-Header.h` definitely doesn't do. There's probably other details I'm missing here; I just decided to try it on a whim).

I just tried it locally, and both `cargo test` in the top-level workspace, and `cargo run -- exec tests/bindings/test_geometry.swift` worked! 🎉

uniffi/src/support/tests.rs
```diff
@@ -0,0 +1,141 @@
...
+/// This function takes the top-level directory of a uniffi component crate, and the path to
+/// a foriegn-language test file that exercises that component's bindings. It ensures that the
+/// component is compiled and available for use and then executes the foreign language script,
+/// returning successfully iff the script exits successfully.
+pub fn run_foreign_language_testcase<P: AsRef<Path>>(pkg_dir: &str, test_file: P) -> Result<()> {
+    let test_file: &Path = test_file.as_ref();
+    let cdylib_file = ensure_compiled_cdylib(pkg_dir)?;
+    let out_dir = Path::new(cdylib_file.as_str())
+        .parent()
+        .ok_or_else(|| anyhow::anyhow!("Generated cdylib has not parent directory"))?;
```
linacambridge, line 45:
```suggestion
        .ok_or_else(|| anyhow::anyhow!("Generated cdylib has no parent directory"))?;
```

uniffi/src/support/tests.rs
```diff
@@ -0,0 +1,141 @@
...
+lazy_static! {
+    static ref COMPILED_COMPONENTS: Mutex<HashMap<String, String>> = Mutex::new(HashMap::new());
+    static ref GENERATED_BINDINGS: Mutex<HashSet<(PathBuf, PathBuf, TargetLanguage)>> =
+        Mutex::new(HashSet::new());
+}
+
+/// Execute the given foreign-language script as part of a rust test suite.
+///
+/// This function takes the top-level directory of a uniffi component crate, and the path to
+/// a foriegn-language test file that exercises that component's bindings. It ensures that the
```
linacambridge, line 37:
```suggestion
/// a foreign-language test file that exercises that component's bindings. It ensures that the
```

uniffi/src/support/tests.rs
```diff
@@ -0,0 +1,141 @@
...
+    match cdylibs.len() {
+        0 => bail!("Crate did not produce any cdylibs, it must not be a uniffi component"),
+        1 => (),
+        _ => bail!("Crate produced multiple cdylibs, it must not be a uniffi component"),
+    }
+    let cdylib_files: Vec<_> = cdylibs[0]
+        .filenames
+        .iter()
+        .filter(|nm| match nm.extension().unwrap_or_default().to_str() {
+            Some("dylib") | Some("so") => true,
```
linacambridge, line 99: Would this be `dll` for Windows someday?

rfk: Maybe one day, when @mhammond gets sufficiently ~~annoyed~~ motivated to shave that yak 😁

rfk:
Thanks @linacambridge, that's much nicer than the gross subdirectory hack! I've updated and squashed and will merge this once CI goes green. (Wow I am relieved to have actual CI here!)

PR #20, opened 2020-07-15
fixes missing ffi_support in FfiStr<'_>

tarikeshaq:
Got a couple of compilation errors and thought might as well fix it


rfk:
You might also be interested in the missing string-related pieces described in https://github.com/rfk/uniffi-rs/issues/3

PR #21, opened 2020-07-17
Further refactor backends to use askama macros

### LINKED ISSUE #15: Refactor templates to increase template re-use and make them easier to navigate

jhugman:
Fixes #15.

This is Part 2 of #15.

This introduces askama macros to the backends, notably around declaring and calling syntax for argument lists. 

This should make it _much_ easier to introduce more features, such as properties getters and setters, name mangling etc.


src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -1,31 +1,25 @@
...
+    fun {{ meth.name()|fn_name_kt }}({% call kt::arg_list_decl(meth.arguments()) %}): {{ return_type|type_kt }} {
+        val _retval = {% call kt::to_rs_call_with_prefix("this.handle.get()", meth) %}
+        return {{ "_retval"|lift_kt(return_type) }}
     }
-    {%- endfor %}
+    
+    {%- when None -%}
+    fun {{ meth.name()|fn_name_kt }}({% call kt::arg_list_decl(meth.arguments()) %}) =
+        {% call kt::to_rs_call_with_prefix("this.handle.get()", meth) %}
+    {% endmatch %}
```
jhugman, line 23: This ended up a lot more readable switching to the doing the match around the whole method (like we do in `TopLevelFunctionTemplate.kt`) instead of doing it twice and/or a clunky `return _retVal` when `_retVal` was `Unit`.

src/bindings/kotlin/templates/macros.kt
```diff
@@ -0,0 +1,45 @@
...
+// Arglist as used in kotlin declarations of methods, functions and constructors.
+// Note the var_name_kt and type_kt filters.
+-#}
+
+{% macro arg_list_decl(args) %}
+    {%- for arg in args -%}
+        {{ arg.name()|var_name_kt }}: {{ arg.type_()|type_kt -}}
+        {%- if !loop.last %}, {% endif -%}
+    {%- endfor %}
+{%- endmacro %}
```
jhugman, line 34: This gets called as 

```
call kt::arg_list_decl(meth.arguments())
```

This converts now writes

```kotlin
initialPoint: Point
```

src/bindings/kotlin/templates/macros.kt
```diff
@@ -0,0 +1,45 @@
...
+    {{- prefix }}{% if func.arguments().len() > 0 %}, {% call _arg_list_rs_call(func.arguments()) -%}{% endif -%}
+)
+{%- endmacro -%}
+
+{%- macro _arg_list_rs_call(args) %}
+    {%- for arg in args %}
+        {{- arg.name()|lower_kt(arg.type_()) }}
+        {%- if !loop.last %}, {% endif %}
+    {%- endfor %}
+{%- endmacro -%}
```
jhugman, line 22: This is only called within this file.

This mangles the name in a call to the JNA library in methods/functions/constructors.

```kotlin
lower(initialPoint)
```

src/bindings/kotlin/templates/macros.kt
```diff
@@ -0,0 +1,45 @@
...
+{#-
+// Arglist as used in the _UniFFILib function declations.
+// Note unfiltered name but type_c filters.
+-#}
+{%- macro arg_list_rs_decl(args) %}
+    {%- for arg in args %}
+        {{- arg.name() }}: {{ arg.type_()|type_c -}}
+        {%- if !loop.last %}, {% endif %}
+    {%- endfor %}
+{%- endmacro -%}
```
jhugman, line 45: This is only used in one place, but put it here because the types need to match with arg lists for calls generated by `_arg_list_rs_call`.

src/bindings/kotlin/templates/RustBufferTemplate.kt
```diff
@@ -12,21 +12,16 @@ open class RustBuffer : Structure() {
...
+        internal fun free(buf: RustBuffer.ByValue) =
+            _UniFFILib.INSTANCE.{{ ci.ffi_bytebuffer_free().name() }}(buf)
     }
 
     @Suppress("TooGenericExceptionThrown")
-    fun asByteBuffer(): ByteBuffer? {
-        return this.data?.let {
-            val buf = it.getByteBuffer(0, this.len)
-            buf.order(ByteOrder.BIG_ENDIAN)
-            return buf
```
jhugman, line 29: I started tidying `let` and `return` here; but the whole method got smaller and smaller. Then the other methods sort just kinda lost their `return`s. :D

uniffi/src/bindings/python/templates/ObjectTemplate.py
```diff
@@ -1,29 +1,25 @@
...
-            {{ arg.name()|lower_py(arg.type_()) }}{% if loop.last %}{% else %},{% endif %}
-            {%- endfor %}
-        )
-        return {% match meth.return_type() %}{% when Some with (return_type) %}{{ "_retval"|lift_py(return_type) }}{% else %}None{% endmatch %}
-    {%- endfor %}
\ No newline at end of file
+    {% for meth in obj.methods() -%}
+    {%- match meth.return_type() -%}
+
+    {%- when Some with (return_type) -%}
```
tarikeshaq, line 14: I'm wondering if there is a clean way to merge those 🤔 The difference seems to be the `_retval`. In Rust `_retval` can be `()` and that brings functions with a return_type and those without one closer. Not sure if that helps, or if the same concept exists in the bindings thou 😅  And I'd probably err on the side of consistency.. Anywhoo just dumping my random thoughts on an already great work 😆

uniffi/src/bindings/swift/templates/macros.swift
```diff
@@ -0,0 +1,45 @@
+{# 
+// Template to call into rust. Used in several places.
+// Variable names in `arg_list_decl` should match up with arg lists
+// passed to rust via `_arg_list_rs_call` (we use  `var_name_kt` in `lower_kt`)
```
tarikeshaq, line 4:
```suggestion
// passed to rust via `_arg_list_rs_call` (we use  `var_name_swift` in `lower_swift`)
```

uniffi/src/bindings/swift/templates/macros.swift
```diff
@@ -0,0 +1,45 @@
...
+
+{%- macro _arg_list_rs_call(args) %}
+    {%- for arg in args %}
+        {{- arg.name()|lower_swift(arg.type_()) }}
+        {%- if !loop.last %}, {% endif %}
+    {%- endfor %}
+{%- endmacro -%}
+
+{#-
+// Arglist as used in kotlin declarations of methods, functions and constructors.
```
tarikeshaq, line 25:
```suggestion
// Arglist as used in Swift declarations of methods, functions and constructors.
```

uniffi/src/bindings/swift/templates/macros.swift
```diff
@@ -0,0 +1,45 @@
...
+// Note the var_name_swift and type_kt filters.
```
tarikeshaq, line 26:
```suggestion
// Note the var_name_swift and decl_swift filters.
```

uniffi/src/templates/ObjectTemplate.rs
```diff
@@ -42,21 +35,13 @@ lazy_static::lazy_static! {
 {%- for meth in obj.methods() %}
     #[no_mangle]
     pub extern "C" fn {{ meth.ffi_func().name() }}(
-        {%- for arg in meth.ffi_func().arguments() %}
-        {{ arg.name() }}: {{ arg.type_()|type_c }},
-        {%- endfor %}
-    ) -> {% match meth.ffi_func().return_type() %}{% when Some with (return_type) %}{{ return_type|ret_type_c }}{% else %}(){% endmatch %} {
+        {%- call rs::arg_list_rs_decl(meth.ffi_func().arguments()) %}) -> {% match meth.ffi_func().return_type() %}{% when Some with (return_type) %}{{ return_type|ret_type_c }}{% else %}(){% endmatch %} {
```
tarikeshaq, line 38: I wonder if the return type can also be a macro, although you don't need to add it here, I can add it in my effort in #31

Also, this might be a dumb question, but is there a reason why we do the matching here in the object template, but we do a match before the functions on the TopLevelFunction template? Not that I have an opinion on which is really better, but just wondering

uniffi/src/templates/macros.rs
```diff
@@ -0,0 +1,33 @@
+{# 
+// Template to call into rust. Used in several places.
+// Variable names in `arg_list_decl` should match up with arg lists
+// passed to rust via `_arg_list_rs_call` (we use  `var_name_kt` in `lower_kt`)
```
tarikeshaq, line 4:
```suggestion
// passed to rust via `_arg_list_rs_call` (we use `lift_rs`)
```

uniffi/src/templates/macros.rs
```diff
@@ -0,0 +1,33 @@
+{# 
+// Template to call into rust. Used in several places.
+// Variable names in `arg_list_decl` should match up with arg lists
```
tarikeshaq, line 3: We can probably get rid of those couple of lines (or change them) 😄

uniffi/src/bindings/python/templates/macros.py
```diff
@@ -0,0 +1,56 @@
...
+
+{%- macro _arg_list_rs_call(args) %}
+    {%- for arg in args %}
+        {{- arg.name()|lower_py(arg.type_()) }}
+        {%- if !loop.last %}, {% endif %}
+    {%- endfor %}
+{%- endmacro -%}
+
+{#-
+// Arglist as used in kotlin declarations of methods, functions and constructors.
```
tarikeshaq, line 25:
```suggestion
// Arglist as used in Python declarations of methods, functions and constructors.
```

uniffi/src/bindings/kotlin/templates/TopLevelFunctionTemplate.kt
```diff
@@ -1,31 +1,13 @@
...
-    ): {{ return_type|type_kt }} {
-        val _retval = _UniFFILib.INSTANCE.{{ func.ffi_func().name() }}(
-            {%- for arg in func.arguments() %}
-            {{ arg.name()|lower_kt(arg.type_()) }}{% if loop.last %}{% else %},{% endif %}
-            {%- endfor %}
-        )
-        return {{ "_retval"|lift_kt(return_type) }}
-    }
+fun {{ func.name()|fn_name_kt }}({%- call kt::arg_list_decl(func.arguments()) -%}): {{ return_type|type_kt }} {
+    val _retval = {% call kt::to_rs_call(func) %}
```
tarikeshaq, line 5: This is just beautiful

uniffi/src/templates/ObjectTemplate.rs
jhugman: I don't like the way we have to duplicate the function body, but have sacrificed de-duplication in favour of clarity/readablity.

Then I saw what you were doing in #31 with the `return_type` macro, and thought it a better way.

I wanted to land this work to make it a base for you to work on and improve in #31.

PR #22, opened 2020-07-19
Add support for strings

### LINKED ISSUE #3: Add support for strings in arguments, return values, and record fields

tarikeshaq:
fixes #3 

**This is based on top of #18, with an additional commit. That additional commit is the one relevant for review. I will keep rebasing as #18 changes.** 

This adds support for using strings across our FFIs, a few things to note about my approach (which could be waaaay off, which is why this is a draft and there are still a few TODO comments):

- Since we have an asymmetry between arguments and return values, there is some duplication with the `type_c` values
- Another weird side effect of the asymmetry is two `impl`s for ViaFFi, one for the lifting from `FfiStr` to `&str`, and a second for the lowering from `String` to `* mut c_char` there is probably a better way of doing this, but I'm blanking out 😞 
- Freeing is **not** implemented yet, I imagine we'd be generating a `free_string` type of api that the bindings can use similar to what is done over at a-s
- I am not too familiar with Kotlin/Swift/Python so there is a non-zero chance that I missed something while doing those bindings 😅 
- Haven't thought too much about the `strings as record members` part of #3 so going to test that

TODOs before this is ready:
- [x] Implement (or figure out) freeing
- [ ] Add tests for strings as record members (I'll add a `TodoEntry` to the todolist example)
- [ ] Minimize duplication (Possibly a different strategy of passing strings than what we have at a-s? Can a ByteBuffer be used maybe, with the strategy mentioned in #3, we can lower and lift the length of the string... Hmmmm)
- [ ] Make the PR target the main branch rather than the test-harness branch


rfk:
@tarikeshaq thank you! I took the liberty of editing this PR so that it targets the branch with the test integration rather than `main`, which I think should make it easier to use github's review tools for discussing it. I also added a "TODO" to switch it back before merge. Hope that's OK :-)

examples/todolist/tests/bindings/test_todolist.kts
```diff
@@ -0,0 +1,10 @@
+import uniffi.todolist.*
+
+val todo = TodoList()
+todo.addItem("Write strings support")
+
+assert(todo.getLast() == "Write strings support")
+
+todo.addItem("Write tests for strings support")
+
+assert(todo.getLast() == "Write tests for strings support")
```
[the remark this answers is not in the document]
tarikeshaq: I spent a non-trivial amount of time thinking of something clever. It was either this or `bobo` 😂.

uniffi/src/support/mod.rs
```diff
@@ -158,3 +158,35 @@ unsafe impl<T: Liftable + Lowerable> ViaFfi for Option<T> {
...
+
+    // Hmmmm this should never happen since there is asymmetry with strings
+    // We typically go FfiStr -> &str for argumetns and 
+    // String -> *mut c_char for return values. 
+    // One thing to consider is if we should split this trait into two traits?
+    // But this might not we worth it since most types have symmetry.... aaaaaaaaah.
+    fn into_ffi_value(self) -> Self::Value {
+        unsafe {ffi_support::FfiStr::from_raw(self.as_ptr() as *const i8)}
+    }
+}
```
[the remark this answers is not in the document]
tarikeshaq: I like that idea... Seems to be also inline with the `From` and `Into` traits in the standard library!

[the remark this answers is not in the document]
rfk: Lol, sorry, somehow I missed that you had this same suggestion in the comment here. Anyway, splitting them seems like a reasonable approach to me 👍

uniffi/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -63,6 +64,19 @@ mod filters {
...
+    }
+
+    pub fn ret_type_c(type_: &TypeReference) -> Result<String, askama::Error> {
+        Ok(match type_ {
+            TypeReference::Boolean => "Byte".to_string(),
+            TypeReference::Enum(_) => "Int".to_string(),
+            TypeReference::Record(_) => "RustBuffer.ByValue".to_string(),
+            TypeReference::Optional(_) => "RustBuffer.ByValue".to_string(),
+            TypeReference::Object(_) => "Long".to_string(),
+            TypeReference::String => "Pointer?".to_string(),
```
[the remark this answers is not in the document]
tarikeshaq: Aaah you're right! I did the casual "Copy the type from A-S and worry about it later without thinking" 😅

tarikeshaq:
Thanks for all the reviews @rfk!! I added a second WIP commit that implements the freeing and a most of the string as a record member (The swift bindings are yet to be implemented). It feels a little weird since the lift/lowering process of a string is completely different from the lift/lowering process of a string that is a member of a record.

As for the ViaFfi splitting, since it'll be changing non-string code and this change is starting to grow, I can tackle that in a separate PR after this.

uniffi/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -63,6 +64,20 @@ mod filters {
             TypeReference::Record(_) => "RustBuffer.ByValue".to_string(),
             TypeReference::Optional(_) => "RustBuffer.ByValue".to_string(),
             TypeReference::Object(_) => "Long".to_string(),
+            TypeReference::String => "String".to_string(),
```
linacambridge, line 67: I was wondering, do we need the same workaround as for Python and use `Pointer` here? I remember @thomcc saying something about `String` not being safe for JNA, and if I'm reading our [docs](https://github.com/mozilla/application-services/blob/main/docs/howtos/exposing-rust-components-to-kotlin.md) right, they say to use `Pointer` for Rust strings, but maybe this is different? (And if we could use `Pointer`, would that let us merge `String` and `CString`?)

uniffi/src/support/mod.rs
```diff
@@ -158,3 +158,60 @@ unsafe impl<T: Liftable + Lowerable> ViaFfi for Option<T> {
...
+    // broken down to IntoFfi and TryFromFfi to prevent us from having
+    // to add this
+    fn try_from_ffi_value(_: Self::Value) -> Result<Self> {
+        panic!("Invalid conversion. try_from_ffi_value should not be called on a c_char, an FfiStr<'_> -> &str conversion should be used instead.")
+    }
+}
+
+impl Lowerable for String {
+    fn lower_into<B: BufMut>(&self, buf: &mut B) {
+        buf.put_u32(self.len() as u32); // We limit strings to u32::MAX bytes
```
linacambridge, line 198: Can we `assert!` this, or is that too paranoid? 😉 Who's going to be passing 4 GB strings over the FFI...

uniffi/src/bindings/kotlin/gen_kotlin.rs
thomcc: Summoned!

- For strings passed as arguments from Kotlin to Rust, using `String` on the JVM side (and e.g. `FfiStr<'_>` on Rust) is correct. 
- For strings returned by rust, these must be freed later, and so we need to use a Pointer in order to maintain the same address for when it's later passed to our string freeing function

So, yes, depending on the case, we need to use Pointer for strings otherwise we'll leak them.

uniffi/src/support/mod.rs
thomcc: Please assert, since attacker-controlled strings are possible.

uniffi/src/bindings/kotlin/gen_kotlin.rs
linacambridge: Oh, cool, Tarik's PR is right as-is and I learned something new (it's OK to use `String` for arguments from Kotlin to Rust, just not for returns from Rust to Kotlin).

uniffi/src/support/mod.rs
```diff
@@ -156,3 +158,60 @@ unsafe impl<T: Liftable + Lowerable> ViaFfi for Option<T> {
...
+        buf.put(self.as_bytes());
+    }
+}
+
+// Having this be for &str instead of String would have been nice for consistency but...
+// it seems very dangerous and only possible with some unsafe magic
+// and having a slice referencing the buffer seems like a recipie for
+// disaster... so we have a String here instead.
```
rfk, line 206: I think it's entirely reasonable to insist that lifting can only return owned types, trying to manage lifetimes otherwise seems like pain we don't need to endure anytime soon.

rfk:
@tarikeshaq sorry, I think I closed this by merging https://github.com/rfk/uniffi-rs/pull/18 and now I can't re-open it. Could you please re-open this (or push a fresh one) and target the `main` branch?

PR #23, opened 2020-07-19
WIP experiment with using procedural macros rather than an IDL

rfk:
The conversation over in https://github.com/rfk/uniffi-rs/pull/9 got me interested in messing around with procedural macros, and I finally managed to steal a bit of time to dig in. This is my *extremely* work-in-progress attempt to port the code over to using macros rather than a separate IDL file.

The code here really doesn't work, although I can see a plausible path to making it work. I'm entirely pushing it for early visibility, and the comments below are a bit of brain-dump as I finish up my current timebox.

I like it so far! It's quite fun to work with `syn` and `quote`, and the shape of much of the code here is fairly similar to how it works on top of `weedle` - rather than converting from e.g. a parsed `weedle` enum into our own definition struct, we convert from a parsed `syn` enum, but it works in broadly the same way. It's also **much** easier to produce helpful contextual error messages, as thom suggested in https://github.com/rfk/uniffi-rs/pull/9.

I've opted for a wasm-bindgen style API here, where you decorate otherwise ordinary rust code with macros to export bits of it over the FFI. I like the way this approach plays with the rest of the rust tooling, such as allowing `cargo fmt` to work as expected.

One thing that's a lot harder though, is that each macro only gets to operate on its local information - for example if I'm exporting `pub fun foo(v: Bar)` with a macro, the macro can see that it wants a type named `Bar`, but it doesn't know whether that type has been properly exported or even exists at all. This seems tractable, but annoying. I have vague recollections of wasm-binding needing to do some pretty hacky things to work around this, but I don't really remember the details.

An alternative would be to do something in the style of the [gobject macros mentioned in this recent release note](https://blog.rust-lang.org/2020/07/16/Rust-1.45.0.html#stabilizing-function-like-procedural-macros-in-expressions-patterns-and-statements), where they have a whole custom API definition inside of a `gobject_gen! { ... }` macro invocation. Taking this approach would let us see and operate on the entire interface definition at once, but honestly, I think I'd really miss `cargo fmt`.

### DIFF
--- a/examples/arithmetic/Cargo.toml
+++ b/examples/arithmetic/Cargo.toml
@@ -15,6 +15,8 @@ log = "0.4"
 ffi-support = "0.4"
 uniffi = {path = "../../uniffi"}
 uniffi_macros = {path = "../../uniffi_macros"}
+lazy_static = "1.4"
+static_assertions = "1.1"
 
 [build-dependencies]
 uniffi = {path = "../../uniffi"}
--- a/examples/arithmetic/src/lib.rs
+++ b/examples/arithmetic/src/lib.rs
@@ -2,25 +2,26 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-// Flag to let calling code specify a particular behaviour of integer overflow,
-// since rust has such nice support for these.
-enum Overflow {
+use uniffi_macros::*;
+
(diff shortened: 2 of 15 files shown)


jhugman:
Propose we close this in favour of #385 and #386?

rfk:
> Propose we close this in favour of #385 and #386?

Yeah, sounds good, this branch is too stale to be of further use for driving discussion 👍

PR #31, opened 2020-07-24
Adds error support

### LINKED ISSUE #7: Add support for returning and handling errors

tarikeshaq:
fixes #7 

### So what is in this so far? 
- Simple parsing of attributes
- Generation of `From<Error> for ExternError`
- Adds the `Result` stuff in the templates needed for it to build
- The Kotlin bindings
  - Generates an Error and Exception for each error on the Kotlin side
  - Does the appropriate matching on the error codes in each error to produce the correct Exception subtype
- The Swift bindings
  - Basically the same as the Kotlin bindings

### What is left for future work?
- Support catching errors that are not explicitly defined in the IDL (panics for example) #32 
- Python bindings for error handling

### TODOs before this is ready:
- [x] WebIDL parsing 
- [x] Scaffolding
- [x] Bindings
  - [x] Kotlin
  - [x] ~~Python~~ (Will create an issue for this, but will leave Python out to get this up quicker, any objections are welcome!)
  - [x] Swift
- [x] Address Feedback
  - [x] Use generics to make the `rustCall` nicer - It's a little nicer, could be better
  - [x] ~~Consolidate Errors into one, with multiple exceptions on the bindings side of things~~ So what we have right now is each `Error` defined on the Rust side, will correspond to one `Error` defined on the bindings, then the enum variants will match to exceptions.
  - [x] Try to get rid of the `GenericRustError` nonsense
- [x] Tests!
- [x] Docs! - Kinda done, happy to add any missing docs that need clarification!


uniffi/src/interface/mod.rs
```diff
@@ -699,10 +742,50 @@ impl APIConverter<Method> for weedle::interface::OperationInterfaceMember<'_> {
...
             ffi_func: Default::default(),
+            throws: match &self.attributes {
+                Some(attr) => attr.body.list.find_throws(ci)?,
+                None => None,
+            },
         })
     }
 }
 
+trait FindThrows {
```
tarikeshaq, line 753: This shouldn't be a trait I think, I'll replace it with a function. I don't think they'll be a difference between the `Throws` attributes on methods and functions

uniffi/src/interface/mod.rs
```diff
@@ -699,10 +742,50 @@ impl APIConverter<Method> for weedle::interface::OperationInterfaceMember<'_> {
                 weedle::types::ReturnType::Type(t) => Some(t.resolve_type_definition(ci)?),
             },
             ffi_func: Default::default(),
+            throws: match &self.attributes {
+                Some(attr) => attr.body.list.find_throws(ci)?,
```
tarikeshaq, line 746: This will error out when if sees other attributes (Since we don't support them) I'll add that!

uniffi/src/interface/mod.rs
```diff
@@ -76,6 +76,7 @@ pub struct ComponentInterface {
     records: Vec<Record>,
     functions: Vec<Function>,
     objects: Vec<Object>,
+    error: Option<Enum>,
```
tarikeshaq, line 79: Felt a little iffy about adding a full `error` piece here, but it seemed like a convenient spot if we end up going the "Generate `From<Error> for ExternError` route"

examples/todolist/src/todolist.idl
```diff
@@ -5,10 +5,16 @@ dictionary TodoEntry {
     string text;
 };
 
+enum Error {
+    "NetworkError", "TypeError", "NoTodoError"
+};
+
 interface TodoList {
     constructor();
+    [Throws=Error]
```
tarikeshaq, line 14: A little different from the `[Throws(Error)]` syntax suggested, thoughts?

uniffi/src/templates/ObjectTemplate.rs
```diff
@@ -50,14 +50,14 @@ lazy_static::lazy_static! {
...
+        UNIFFI_HANDLE_MAP_{{ obj.name()|upper }}.{% match meth.error() %}{% when Some with (e) %}call_with_result_mut{% else %}call_with_output_mut{% endmatch %}(&mut err, {{ meth.first_argument().name() }}, |obj| {% match meth.error() %}{% when Some with (e) %}-> Result<{% match meth.ffi_func().return_type() %}{% when Some with (return_type) %}{{ return_type|ret_type_c }}{% else %}(){% endmatch %}>{% else %}{% endmatch %}{
             let _retval = {{ obj.name() }}::{{ meth.name() }}(
                 obj,
                 {%- for arg in meth.arguments() %}
                 {{ arg.name()|lift_rs(arg.type_()) }},
                 {%- endfor %}
-            );
-            {% match meth.return_type() %}{% when Some with (return_type) %}{{ "_retval"|lower_rs(return_type) }}{% else %}{% endmatch %}
+            ){% match meth.error() %}{% when Some with (e) %}?{% else %}{% endmatch %};
+            {% match meth.error() %}{% when Some with (e) %}Ok({% else %}{% endmatch %}{% match meth.return_type() %}{% when Some with (return_type) %}{{ "_retval"|lower_rs(return_type) }}{% else %}_retval{% endmatch %}{% match meth.error() %}{% when Some with (e) %}){% else %}{% endmatch%}
```
tarikeshaq, line 60: Lots of repetitive `match` here, I'll look if there is a way to set a custom reusable block with askama

tarikeshaq: Realized #21 adds exactly the kind of stuff that'll help here 👀

uniffi/src/interface/mod.rs
```diff
@@ -466,6 +487,17 @@ impl APIConverter<Argument> for weedle::argument::SingleArgument<'_> {
     }
 }
 
+impl APIConverter<Error> for weedle::common::Identifier<'_> {
+    fn convert(&self, ci: &ComponentInterface) -> Result<Error> {
+        Ok(Error {
+            name: self.0.to_string(),
+            type_: ci
+                .get_type_definition(self.0)
```
rfk, line 495: Careful, just because the error type isn't known at the time you encounter it in the parse tree, doesn't mean it's invalid. It's OK for the consumer to have done something like:

```
interface Example {
  [Throws=Error]
  void example();
};

enum Error {
  "EXAMPLE"
};
```

There probably needs to be an entirely separate "consistency checking" phase where we walk the parsed `ComponentInterface` and check that all return types and defined, there isn't an enum and a record with the same name, etc. Right now we only catch these errors by virtual of *generating invalid rust scaffolding*, which is not a great developer experience.

But I don't think you should have to tackle that inline in this PR.

uniffi/src/interface/mod.rs
```diff
@@ -582,6 +614,12 @@ impl Constructor {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Error {
+    name: String,
+    type_: TypeReference,
```
rfk, line 620: I don't think you need this `type_` field on the error struct itself, since you can look it up at any time by doing `ci.get_type_definition(err.name)`.

Instead, I think the abstraction you want here is closer to `TypeReference::Record(name)` or `TypeReference::Enum(name)`, where you just have a holder for the name and it's up to the consuming code to look up the corresponding type definition if and when it needs it.

You could consider using `TypeReference::Enum(name)` directly for this, or you could try adding a `TypeReference::Error(name)` variant to give you what you want here in a way that fits with the existing machinery of this module. Either way, I think "a reference to a named error type" is the abstraction you want here, without having to worry about looking up the underlying type definition until some later phase of code generation.

uniffi/src/interface/mod.rs
```diff
@@ -620,6 +659,10 @@ impl Method {
         }
     }
 
+    pub fn error(&self) -> &Option<Error> {
```
rfk, line 662: API nit: it's not clear why `error` is preferred as the name of the function but `throws` as the name of the attribute, and it seems worth being consistent; FWIW my slight preference would be to use `throws` consistently, from IDL definition to struct attribute to public accessor.

uniffi/src/interface/mod.rs
rfk: Thinking about the way the rest of this code is structured, and about a few of my previous attempts at processing the WebIDL here, here's a hand-wavy suggestion for dealing with attributes. Feel free to ignore it since I really haven't tried it out, I'm just kind of squinting at the problem and thinking about how I might approach it, so the below might turn out to be garbage. But, a suggestion:

* Create a new `enum Attribute` that is specifically for parsing function attributes, with one variant per attribute of interest. Right now I guess you'd only have a single variant `Attribute::Throws(name)`.
* Use `impl TryFrom` to destructure weedle's attribute syntax nodes into your much simpler `Attribute` enum, similar to what we do for other things using the `APIConverter` trait.
    * You don't need a reference to the `ComponentInterface` to achieve this, which is why you can `impl TryFrom` rather than using `APIConverter`.
    * This is where you error out if you encounter an unknown attribute or any of the many syntax forms that weedle supports and we don't.
* In the `APIConverter` impl for `Function`, `Method`, etc where you want to handle attributes:
    * map them into a `Vec` of your `Attribute` enum, like:
        * `let attrs: Vec<Attribute> = attrs.body.list.map(|a| a.try_into()).collect::<Result<_>>()?`
    * inspect the list of converted `Attribute`s to figure out what flags you want to set on your object.

That's a lot of machinery! You'd be justified to conclude that you don't really need all that, and you probably don't *need* it for this single attribute. But the reason I suggested it is to set yourself up for future maintainability as we grow additional attributes, because:

* The "bail if we see unexpected attribute syntax" logic can live in a single place, rather than in each location.
* The logic about what attribute sets what flag on each struct can live with the struct itself.

rfk: (Also I am just a really huge fan of `TryFrom`...)

uniffi/src/templates/ObjectTemplate.rs
```diff
@@ -50,14 +50,14 @@ lazy_static::lazy_static! {
         let mut err: ffi_support::ExternError = Default::default(); // XXX TODO: error handling!
         // If the method does not have the same signature as declared in the IDL, then
         // this attempt to call it will fail with a (somewhat) helpful compiler error.
-        UNIFFI_HANDLE_MAP_{{ obj.name()|upper }}.call_with_output_mut(&mut err, {{ meth.first_argument().name() }}, |obj| {
+        UNIFFI_HANDLE_MAP_{{ obj.name()|upper }}.{% match meth.error() %}{% when Some with (e) %}call_with_result_mut{% else %}call_with_output_mut{% endmatch %}(&mut err, {{ meth.first_argument().name() }}, |obj| {% match meth.error() %}{% when Some with (e) %}-> Result<{% match meth.ffi_func().return_type() %}{% when Some with (return_type) %}{{ return_type|ret_type_c }}{% else %}(){% endmatch %}>{% else %}{% endmatch %}{
             let _retval = {{ obj.name() }}::{{ meth.name() }}(
                 obj,
                 {%- for arg in meth.arguments() %}
                 {{ arg.name()|lift_rs(arg.type_()) }},
```
rfk, line 57: Since we're here: note that `lift_rs` comes out as `<TYPE as ViaFfi>::try_from_ffi_value(VALUE).unwrap()`. If we're adding error handling, it would be nice to change that `unwrap` into a `?` or similar and have the `ViaFfi` conversion errors reported up to calling code in a sensible way.

Such errors should of course *never* happen in theory, because we control both sides of the FFI, so why would there ever be a conversion error? But you know what they say about theory and practice.

A strawperson suggestion: consider reserving a special error code (maybe `MAX_INT`?) for reporting these `ViaFfi` conversion errors, so that every function can implicitly throw a "something has gone horribly wrong in the bindings layer" error if something happens to go horribly wrong in the bindings layer.

uniffi/src/interface/mod.rs
rfk: > just because the error type isn't known at the time you encounter it in the parse tree, doesn't mean it's invalid. 

Hrm, sorry, I got a bit confused here - we already have a separate "type finding" step that will handle the use-before-definition problem I mention above, although there are plenty of other consistency checks that it *doesn't* perform. If you decide to keep the current approach then I think you want `resolve_type_definition` rather than `get_type_definition` here. (Which I realize there's not really any way you could know that because this is not well documented so, that's something for me to take a look at adding).

I still feel like you want a `TypeReference` of some sort here rather than a separate struct for errors. But looking that `TypeReference` up by name using `resolve_type_definition` would be the right way to obtain one; see e.g. how we resolve the return type when construction a `Function` instance.

uniffi/src/templates/ObjectTemplate.rs
rfk: Alternatively, I suppose we could handle these conversion errors the same way we handle panics in the underlying rust code, and spin that work off into a separate "formalize and add tests for handling panics" issue.

tarikeshaq:
Thanks for the comments @rfk! I addressed some of the comments in a new commit, I can squash commits later, but wanted to make it easier to see what I changed! I was about to start dealing with the bindings, but I really wanted to get the scaffolding to look better before doing that.

Note: Tests are gonna fail because I moved the `&mut ExternError` to be an argument and haven't delt with the bindings yet

uniffi/src/templates/ObjectTemplate.rs
tarikeshaq: Hmmm yeah, this is an interesting problem, because it exists whether or not the function itself is expected to return a `Result`. I might spin off a different issue to deal with this... Here's what is going through my head right now:

- For functions that already return `Result`s we'd have to figure out a way to map the conversion error to the expected error or create a layer of indirection, and have both the user-defined error and the conversion errors map to that, then use that as the return value of the closure in `call_with_result_mut`
- For functions that don't return `Result`s, If we take the indirection route, then we just define the 'indirection' error anyways and change those calls to be `call_with_result_mut` even though the core rust function itself isn't expected to return a `Result`. 

I'm leaning towards having this as a follow-up bug, but want to make sure that I don't make decisions that make it tough to implement, thoughts?

uniffi/src/interface/mod.rs
tarikeshaq: I `TryFrom`ed it 😄 Added an `Attributes` as a wrapper for `Vec<Attribute>`. And I'm storing that now in methods instead of a single "Error". Can be helpful as we start supporting more attributes I guess, let me know your thoughts 😄

uniffi/src/interface/mod.rs
tarikeshaq: Hmmm ended up storing the full list of attributes in the method here, then later would check if an error exists in there. Seems like the compiler errors out semi-nicely with that if something is missing

uniffi/src/interface/mod.rs
```diff
@@ -485,9 +560,6 @@ impl Enum {
 
 impl APIConverter<Enum> for weedle::EnumDefinition<'_> {
     fn convert(&self, _ci: &ComponentInterface) -> Result<Enum> {
-        if self.attributes.is_some() {
```
tarikeshaq, line 488: We do the attributes parsing beforehand for this. (since the enum might be an error)

uniffi/src/interface/mod.rs
```diff
@@ -699,6 +845,10 @@ impl APIConverter<Method> for weedle::interface::OperationInterfaceMember<'_> {
                 weedle::types::ReturnType::Type(t) => Some(t.resolve_type_definition(ci)?),
             },
             ffi_func: Default::default(),
+            attributes: match &self.attributes {
+                Some(attr) => Attributes::try_from(attr)?,
```
tarikeshaq, line 849: This is missing erroring for valid Attributes that are not valid for methods (it errors out if attributes are completely invalid, but might want to ensure that the attributes are valid for methods, like `Error` for example should only be on `Enum`s

uniffi/src/templates/ObjectTemplate.rs
```diff
@@ -40,25 +40,51 @@ lazy_static::lazy_static! {
...
+        })
+    }
+
+    {% else %}
+#[no_mangle]
+    pub extern "C" fn {{ meth.ffi_func().name() }}(
+        {%- for arg in meth.ffi_func().arguments() %}
+        {{ arg.name() }}: {{ arg.type_()|type_c }},
+        {%- endfor %}
+        err: &mut ffi_support::ExternError,
```
rfk, line 73: Since you're always taking an `err` param here even if the underlying method can't throw, you could consider pushing the `{% match meth.throws() %}` down into the  body of the method, and using to ensure that we always return a `Result`. Something like:

```
        UNIFFI_HANDLE_MAP_foo.call_with_result_mut(err, [...], |obj| -> Result<[...], [... can we use "_" here to infer the error type? ...]> {
            let _retval = {{ obj.name() }}::{{ meth.name() }}(
                obj,
                {%- for arg in meth.arguments() %}
                {{ arg.name()|lift_rs(arg.type_()) }},
                {%- endfor %}
            );

            {% match meth.throws() %}
            {% when Some with (e) %} 
            let _retval = _retval?;
            {% endmatch %}

            Ok([... cast "_retval" into ffi type...])
        })
    }
```

Not sure if that would make it easier or harder to follow...you might be better off just branching on "does it throw?" and "does it have a return type" and doing four different templates for four different combos.

uniffi/src/templates/ObjectTemplate.rs
rfk: I also lean towards a separate issue. Let's continue to panic on conversion errors since they should never happen in practice, and rely on safe panic handling (which we need anyway!) to report them up to callers.  I filed https://github.com/rfk/uniffi-rs/issues/32 for followup.

 I think the main impacts on the current PR will be:

* All FFI functions would need to take an out `ExternError` for handling panics, which might influence how you do codegen for handling other errors.
* You need to avoid conflicting with reserved error codes (IIUC from [the docs](https://docs.rs/ffi-support/0.4.0/ffi_support/struct.ExternError.html) we reserve `0` for success and `-1` for reporting internal errors such as panics.

I can't think of any other corners you might back yourself in to as a result of this PR.

rfk:
Oh! I've also just realized that the error-handling in `ffi-support` relies on the ability of the [`IntoFfi` trait](https://docs.rs/ffi-support/0.4.0/ffi_support/trait.IntoFfi.html) to fill in a default return value in cases where it needs to return an error. You may find you need to add an `ffi_default` function and fiddle around a bit with the traits in order to make this work.

tarikeshaq:
Added some basic kotlin bindings for **Methods** and we can see our first example of Errors across the FFi 👀 (They're able to propage to Kotlin exceptions... Still a decent bit of work to do to clean it up and implement the rest though

uniffi/src/interface/mod.rs
```diff
@@ -59,6 +59,8 @@ pub mod types;
 pub use types::TypeReference;
 use types::TypeResolver;
 
+const GENERIC_RUST_ERROR: &str = "GenericRustError";
```
tarikeshaq, line 62: Aaaah using this to be the name of the error that non-error functions take as there out `err`. It was the lazy thing to do...

uniffi/src/interface/mod.rs
```diff
@@ -96,6 +99,10 @@ impl<'ci> ComponentInterface {
         }
         // Now that the high-level API is settled, we can derive the low-level FFI.
         ci.derive_ffi_funcs()?;
+        ci.errors.push(Error {
+            name: GENERIC_RUST_ERROR.to_string(),
+            values: vec!["UnknownError".to_string()],
```
tarikeshaq, line 104: Note to self to maybe add `Panic` as a variant here and look at the actual conversions that `ffi_support` covers.. I might have this as a part of #32 But seems more relevant to this PR... We'll see

uniffi/src/templates/ErrorTemplate.rs
```diff
@@ -0,0 +1,30 @@
+
+#[derive(Debug, Clone, thiserror::Error)]
+enum GenericRustError {
+    #[error("Unknown error")]
```
tarikeshaq, line 4: Ditto on the `GenericRustError` Wasn't sure this was the smartest way to do this

uniffi/src/templates/ObjectTemplate.rs
```diff
@@ -40,25 +40,51 @@ lazy_static::lazy_static! {
...
-    #[no_mangle]
+
+{% match meth.throws() %}
+    {% when Some with (e) %} 
+#[no_mangle]
     pub extern "C" fn {{ meth.ffi_func().name() }}(
         {%- for arg in meth.ffi_func().arguments() %}
         {{ arg.name() }}: {{ arg.type_()|type_c }},
         {%- endfor %}
+        err: &mut ffi_support::ExternError,
```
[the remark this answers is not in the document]
tarikeshaq: Yup! I was going though the method to see what changes are neccesary before adding the rest 😄

uniffi/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -0,0 +1,96 @@
+@Structure.FieldOrder("code", "message")
```
tarikeshaq, line 1: All this is copied from [`RustError`](https://github.com/mozilla/application-services/blob/main/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/rust/RustError.kt)

Not sure if there are big design decisions that maybe should change?

Seemed okay to me, only big difference is that I'm defining an `Error` and an `Exception` for each error type. An alternative is to have only one `Error` and use the codes to map to different `Exception`s. 

Thoughts?

uniffi/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
```diff
@@ -22,9 +22,9 @@ internal interface _UniFFILib : Library {
     {% for func in ci.iter_ffi_function_definitions() -%}
         fun {{ func.name() }}(
         {%- for arg in func.arguments() %}
-            {{ arg.name() }}: {{ arg.type_()|type_c }}{% if loop.last %}{% else %},{% endif %}
+            {{ arg.name() }}: {{ arg.type_()|type_c }}{% if loop.last && !func.has_error() %}{% else %},{% endif %}
         {%- endfor %}
-        // TODO: When we implement error handling, there will be an out error param here.
+        {% match func.throws() %}{% when Some with (e) %}error: {{e}}.ByReference{% else %}{% endmatch %}
```
tarikeshaq, line 27: The only functions I found that don't take errors are the free/alloc bytebuffers and free strings. And those are defined by macros in `ffi_support` :( If they did take errors, then this logic would be simpler (i.e we'd assume all functions take errors and get rid of the match)

uniffi/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -0,0 +1,96 @@
...
+open class {{e.name()}}Exception(message: String) : Exception(message) {
+    {% for value in e.values() -%}
+    class {{value}}(msg: String) : {{e.name()}}Exception(msg)
+    {% endfor %}
+}
+
+// Helpers for calling Rust with errors:
+// In practice we usually need to be synchronized to call this safely, so it doesn't
+// synchronize itself
+private inline fun <U> nullableRustCallWith{{e.name()}}(callback: ({{e.name()}}.ByReference) -> U?): U? {
```
tarikeshaq, line 79: I'll look at a Generics way of doing those instead of defining multiple `rustCallWith`. Don't know much about Kotlin Generics, but shouldn't be too bad!

uniffi/src/bindings/kotlin/templates/TopLevelFunctionTemplate.kt
```diff
@@ -6,12 +6,15 @@
...
-        )
-        return {{ "_retval"|lift_kt(return_type) }}
+        rustCall { e -> 
+            val _retval = _UniFFILib.INSTANCE.{{ func.ffi_func().name() }}(
+                {%- for arg in func.arguments() %}
+                {{ arg.name()|lower_kt(arg.type_()) }},
+                {%- endfor %}
+                e
+            )
+            return {{ "_retval"|lift_kt(return_type) }}
```
tarikeshaq, line 16: I'll change it once I actually implement top level functions, but this return should be outside the rustCall probably

uniffi/src/templates/macros.rs
```diff
@@ -0,0 +1,20 @@
+{% macro return_type_func(func) %}{% match func.ffi_func().return_type() %}{% when Some with (return_type) %}{{ return_type|ret_type_c }}{%- else -%}(){%- endmatch -%}{%- endmacro -%}
```
tarikeshaq, line 1: Defined some common macros for args/return types/errors/ on the Rust side. I'll keep an eye on the effort on #21 as I tackle the backends 😄

examples/todolist/tests/bindings/test_todolist.kts
```diff
@@ -1,6 +1,20 @@
 import uniffi.todolist.*
 
 val todo = TodoList()
```
tarikeshaq, line 3: Note to self: will need to add a test for a constructor that throws

uniffi/src/bindings/kotlin/templates/macros.kt
```diff
@@ -5,17 +5,21 @@
 #}
 
 {%- macro to_rs_call(func) -%}
-_UniFFILib.INSTANCE.{{ func.ffi_func().name() }}({% call _arg_list_rs_call(func.arguments()) -%})
+ rustCallWith{% match func.ffi_func().throws() %}{% when Some with (e) %}{{e}}{% else %}{% endmatch %} { e ->
```
linacambridge, line 8: How do you feel about moving the `match` up one level, something like:

```text
{% match func.ffi_func().throws() %}
{% when Some with (e) %}
rustCallWith{{e}} { e -> ... }
{% else %}
rustCallWith { e -> ...}
{% endmatch %}
```

(Also, did I read that right? I don't see a `rustCallWith` yet, is that for default errors, or something else? Maybe I should just keep reading 😅) I know it's a couple more duplicated lines, but the inline `match` took me a couple tries to parse on first read!

uniffi/src/bindings/swift/templates/ErrorTemplate.swift
```diff
@@ -0,0 +1,68 @@
...
+    }
+}
+
+internal func nullableRustCallWith{{e.name()}}<T>(_ cb: (UnsafeMutablePointer<Native{{e.name()}}>) throws -> T?) throws -> T? {
+    return try {{e.name()}}.tryUnwrap { err in
+        return try cb(err)
+    }
+}
+
+public enum {{e.name()}}ResultError: Error {
```
linacambridge, line 66: There's an `InternalError` enum in `RustBufferHelper.swift` for these kinds of cases that shouldn't happen, would that be a good fit here?

uniffi/src/bindings/swift/templates/ObjectTemplate.swift
```diff
@@ -2,24 +2,26 @@ public class {{ obj.name() }} {
...
-    public init({% call swift::arg_list_decl(cons.arguments()) -%}) {
+    public init({% call swift::arg_list_decl(cons) -%}) throws {
         self.handle = {% call swift::to_rs_call(cons) %}
     }
     {%- endfor %}
 
     // XXX TODO: destructors or equivalent.
 
+
+    // TODO: Maybe merge the two templates (i.e the one with a return type and the one without)
```
linacambridge, line 13: Mmm, I don't mind that they're spelled out more explicitly, but we can go the other way and merge these, too! At first, I thought duplicating would be more readable than inline `match`es, but it might also mean we accidentally make a copy-pasta error between the two branches.

uniffi/src/interface/mod.rs
linacambridge: Does this mean non-throwing functions still return a `Result<T, GenericRustError>` or?

uniffi/src/bindings/kotlin/templates/macros.kt
tarikeshaq: Ya the `rustCallWith` stuff is an ugly bi-product of me not knowing how to do generics well in Swift and Kotlin and thinking "let me just get it to work". 

I'll change those to be just a `rustCall` and will do some generics magic ✨  to make it work.

But you are right, `rustCallWith` (without any suffix) doesn't exist, but since all functions generated using this take errors, the `else` won't ever happen. It's very ugly I know 😅 Hopefully will be better once I add the generics and simplify the errors to one error.

uniffi/src/interface/mod.rs
tarikeshaq: Ya! And I can get rid of this, but it can be used as a bucket for all 'Unknown' kind of errors (a panic comes to mind, but that'll be a part of #32). Also kinda allows me to go `call_with_result` every time.. 

I'm not a big fan and it feels very hardcoded-y and now I take a second look at it, I'll get rid of it, and have the effort to deal with #32 deal with generic issues that might happen. 😄

uniffi/src/bindings/kotlin/templates/ErrorTemplate.kt
tarikeshaq: So since we have support for multiple errors enums, each with variants on the Rust side, I'm matching each error there to an error here (and swift), and then using the error codes to match variants to exceptions.

Another alternative I'm very happy to change (it won't be much of a change because of the macros ❤️ ) Is to only support one error on the Rust/IDL side, that would put a limitation there, but keep us to one error here (similar to how it's done in A-S)

uniffi/src/templates/TopLevelFunctionTemplate.rs
```diff
@@ -4,23 +4,13 @@
...
-{%- match func.ffi_func().return_type() -%}
-{%- when Some with (return_type) %}
 
 #[no_mangle]
 pub extern "C" fn {{ func.ffi_func().name() }}(
-    {%- call rs::arg_list_rs_decl(func.ffi_func().arguments()) %}) -> {{ return_type|ret_type_c }} {
+    {% call rs::arg_list_rs_decl(func.ffi_func()) %}
+) -> {% call rs::return_type_func(func) %} {
+    // If the provided function does not match the signature specified in the IDL
+    // then this attempt to cal it will not compile, and will give guideance as to why.
```
linacambridge, line 13:
```suggestion
    // then this attempt to call it will not compile, and will give guidance as to why.
```

Side note, I really like how the `to_rs_function_call` macro spells out `<T as uniffi::support::ViaFfi>::try_from_ffi`, so if you're trying to send an unsupported type or pass it at the wrong position, it'll give you a useful error! 👍

uniffi/src/templates/macros.rs
```diff
@@ -24,9 +24,68 @@
...
-{%- macro arg_list_rs_decl(args) %}
-    {%- for arg in args %}
-        {{- arg.name() }}: {{ arg.type_()|type_c -}}
-        {%- if !loop.last %}, {% endif %}
+{%- macro arg_list_rs_decl(func) %}
+    {%- for arg in func.arguments() %}
+        {{- arg.name() }}: {{ arg.type_()|type_c -}}{% if loop.last %}{% else %},{% endif %}
     {%- endfor %}
+    {% if func.has_out_err() %}
+    {% if func.arguments().len() > 0 %},{% endif %} err: &mut ffi_support::ExternError,
```
linacambridge, line 32: Holy cow, so many commas to get right! 😅

uniffi/src/interface/mod.rs
```diff
@@ -363,12 +382,19 @@ impl Function {
...
+
     fn derive_ffi_func(&mut self, ci_prefix: &str) -> Result<()> {
         self.ffi_func.name.push_str(ci_prefix);
         self.ffi_func.name.push_str("_");
         self.ffi_func.name.push_str(&self.name);
         self.ffi_func.arguments = self.arguments.clone();
         self.ffi_func.return_type = self.return_type.clone();
+        // Theoritically this should always be true
+        // but it's this way until we implement handling for panics
+        self.ffi_func.has_out_err = self.throws().is_some();
```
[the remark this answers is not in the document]
tarikeshaq: > Or do you think we'll want all our Swift functions to be marked as throws, and have the caller use try! explicitly?

I was thinking this - which is why I kept all functions as `throws` for now in swift - but now that I think about it, might be better to do the `try!` approach 🤔 Seems to better reflect the fact that the errors are not explicit (i.e if we panic in Rust, maybe it's better to crash swift etc..) either way, I feel like this will surface in #32, for now, all functions have `throws` in swift.

uniffi/src/templates/macros.rs
tarikeshaq: Ya for sure can make it cleaner 👀

uniffi/src/interface/mod.rs
```diff
@@ -591,10 +644,33 @@ impl Constructor {
...
+    values: Vec<String>,
+}
+
+impl Error {
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    pub fn values(&self) -> Vec<&str> {
+        self.values.iter().map(|v| v.as_str()).collect()
```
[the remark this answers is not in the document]
tarikeshaq: I'll have a separate clean-up PR, I think Enums do the same, but this is better!

examples/rondpoint/tests/bindings/test_rondpoint.swift
```diff
@@ -1,10 +1,10 @@
 import rondpoint
 
 let dico = Dictionnaire(un: .deux, deux: false)
-let copyDico = copieDictionnaire(d: dico)
+let copyDico = try! copieDictionnaire(d: dico)
```
rfk, line 4: Oh wow, do we now have to wrap every call to swift functions in a `try!` because of the error handling?

rfk: Actually looks like there was some discussion of this with @linacambridge already below, so don't feel like you need to respond to this directly :-)

tarikeshaq: I'll have this among a couple 'tiny' cleanups in a follow-up PR, also feel free to add any more opinions/reviews on the PR, and I can address them too! Wanted to merge the big chunk of this so that other PRs don't need to wait and we can iterate on smaller PRs improving the error handling.

Issue #33, opened 2020-07-28
Nullable arguments support


eoger:
This currently does not work:
```
namespace bobo {
  double do_something(MyType? arg);
}}
```

Error is
```
thread 'main' has overflowed its stack
fatal runtime error: stack overflow
[1]    78220 abort      cargo run -- generate
```

Issue #41, opened 2020-07-29
SYNC-1549 ⁃ Fix or disable python bindings


eoger:
I am seeing bugs around Optional lift/lower that I'm not sure I know how to fix

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1549)

eoger:
We've decided to allow ourselves to disable failing Python tests

PR #42, opened 2020-07-29
Optional/Nullable args fixes

eoger:
Fixes https://github.com/mozilla/uniffi-rs/issues/33.

This works for Kotlin and Swift, in Python it seems like the work to support nullable types [was never fully completed](https://github.com/mozilla/uniffi-rs/blob/3c36c7b70c68a67c3861e4d14efc6c2857213b22/uniffi/src/bindings/python/gen_python.rs#L91), and I'm not quite sure how to finish it, hence https://github.com/mozilla/uniffi-rs/issues/41.

### DIFF
--- a/examples/sprites/src/sprites.idl
+++ b/examples/sprites/src/sprites.idl
@@ -14,7 +14,8 @@ dictionary Vector {
 };
 
 interface Sprite {
-  constructor(Point initial_position);
+  // Should be an optional, but I had to test nullable args :)
+  constructor(Point? initial_position);
   Point get_position();
   void move_to(Point position);
   void move_by(Vector direction);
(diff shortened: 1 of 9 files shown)


examples/sprites/src/sprites.idl
```diff
@@ -14,8 +14,9 @@ dictionary Vector {
 };
 
 interface Sprite {
-  constructor(Point initial_position);
+  // Should be an optional, but I had to test nullable args :)
+  constructor(Point? initial_position);
```
linacambridge, line 18: Oooh, interesting, I forgot about `optional Point` vs. `Point?`.

examples/sprites/tests/bindings/test_sprites.swift
```diff
@@ -1,5 +1,8 @@
 import sprites
 
+let sempty = Sprite(initialPosition: nil)
+assert( sempty.getPosition() == Point(x: 0, y: 0))
```
linacambridge, line 4: This is making me think more about argument labeling in Swift—labeling all arguments feels pretty wordy, especially when the type and purpose are obvious, but `Sprite(initialPosition: nil)` reads much better than `Sprite(nil)`.

Issue #35, opened 2020-07-28
Boolean field in dictionary support


eoger:
I want to be able to compile the following code:
```
dictionary Point {
  double x;
  double y;
  boolean visited;
};
```

Currently getting errors:
```
error[E0277]: the trait bound `bool: uniffi::support::Lowerable` is not satisfied
  --> /Users/eoger/uniffi-rs/target/debug/build/uniffi-example-geometry-32b57fa3d176b525/out/geometry.uniffi.rs:38:9
   |
38 |         uniffi::support::Lowerable::lower_into(&self.visited, buf);
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `uniffi::support::Lowerable` is not implemented for `bool`
   |
  ::: /Users/eoger/uniffi-rs/uniffi/src/support/mod.rs:17:22
   |
17 |     fn lower_into<B: BufMut>(&self, buf: &mut B);
   |                      ------ required by this bound in `uniffi::support::Lowerable::lower_into`

error[E0277]: the trait bound `u8: uniffi::support::Liftable` is not satisfied
  --> /Users/eoger/uniffi-rs/target/debug/build/uniffi-example-geometry-32b57fa3d176b525/out/geometry.uniffi.rs:47:22
   |
47 |             visited: <u8 as uniffi::support::Liftable>::try_lift_from(buf)?,
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `uniffi::support::Liftable` is not implemented for `u8`
   |
  ::: /Users/eoger/uniffi-rs/uniffi/src/support/mod.rs:68:25
   |
68 |     fn try_lift_from<B: Buf>(buf: &mut B) -> Result<Self>;
   |                         --- required by this bound in `uniffi::support::Liftable::try_lift_from`

error[E0308]: try expression alternatives have incompatible types
  --> /Users/eoger/uniffi-rs/target/debug/build/uniffi-example-geometry-32b57fa3d176b525/out/geometry.uniffi.rs:47:22
   |
47 |             visited: <u8 as uniffi::support::Liftable>::try_lift_from(buf)?,
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `bool`, found `u8`
```

[...]

PR #44, opened 2020-07-29
Fix up support for booleans

### LINKED ISSUE #34: Boolean return support
I want to be able to compile the following
```
boolean return_true();
```

Currently I'm getting the error:
```
87 |     <u8 as uniffi::support::ViaFfi>::into_ffi_value(_retval)
   |                                                     ^^^^^^^ expected `u8`, found `bool`
```

rfk:
This should fix #34 and #35, although I've only added tests for the former.

There's a bunch of `pub`-changes here which I will explain in a comment, and which I'm not wedded to if they don't seem worthwhile. But I discovered a neat thing that I wanted to try out for the rust scaffolding.

### DIFF
--- a/uniffi/src/scaffolding.rs
+++ b/uniffi/src/scaffolding.rs
@@ -24,14 +24,11 @@ mod filters {
 
     pub fn type_rs(type_: &TypeReference) -> Result<String, askama::Error> {
         Ok(match type_ {
-            // These can be passed directly over the FFI without conversion.
             TypeReference::U32 => "u32".to_string(),
             TypeReference::U64 => "u64".to_string(),
             TypeReference::Float => "f32".to_string(),
             TypeReference::Double => "f64".to_string(),
-            TypeReference::Boolean => "u8".to_string(),
-            // These types need conversion, and will require special handling below
-            // when lifting/lowering.
+            TypeReference::Boolean => "bool".to_string(),
             TypeReference::String => "&str".to_string(),
             TypeReference::Enum(name) => name.clone(),
             TypeReference::Record(name) => name.clone(),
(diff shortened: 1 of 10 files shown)


uniffi/src/scaffolding.rs
```diff
@@ -42,41 +39,38 @@ mod filters {
...
-            TypeReference::Enum(_) => "u32".to_string(),
-            TypeReference::Record(_) => "ffi_support::ByteBuffer".to_string(),
-            TypeReference::Optional(_) => "ffi_support::ByteBuffer".to_string(),
+            // Objects don't currently impl `ViaFfi`.
             TypeReference::Object(_) => "u64".to_string(),
-            _ => type_rs(type_)?,
+            _ => format!(
+              "<{} as uniffi::support::ViaFfi>::Value",
+              type_rs(type_)?,
+            ),
```
rfk, line 62: I discovered that you can use the associated types of a trait in the type signature of a function. Expressing it like this means we don't have to repeat the big list of "rust type X maps to FFI type Y" here when generating the scaffolding, we can just look up the mapping on the `ViaFfi` trait.

I like it, but some downsides of this approach:

* The generating rust scaffolding is a bit more visually complex; I would argue that it's actually easier to *understand* because you don't have to account for the mapping between Rust and C types, but it's definitely extra syntax noise.
* This only works if the corresponding rust types are public, since they appear in the type signature of a public function. Hence all the additions of `pub` in the examples. But TBH I think it's reasonable to insist that these types be marked `pub` because they are part of your public API.

If you don't like this, I can revert the change can go back to a big `match` statement like we had here before, but I figured I'd try it out.

eoger: This is the kind of magic that I like, I'll take 1 source of truth over 3, nice cleanup :)

rfk:
> This is the kind of magic that I like

TBH I couldn't quite believe this Just Worked (tm) when I tried it the first time, the richness of rust's type system continues to impress me.

Issue #28, opened 2020-07-22
Add support for sequence types in arguments, return values, and records


rfk:
Sequences (or "lists" or "arrays" or "vectors" or whatever you like) are a pretty important data type that we don't currently have support for. Let's add them!

I think we can use the existing `Optional` type as a good template. Broadly, we'll need to:

* Add a sequence type to one of the examples. I think Tarik's "todolist" example from https://github.com/rfk/uniffi-rs/pull/25 would be a great candidate, since returning a list of todo items is an obvious extension of that functionality.
     * The WebIDL syntax for sequences is pretty straightfoward: [`sequence<T>`](https://heycam.github.io/webidl/#idl-sequence) for any type `T`.
     * There's already some skeleton code for parsing this type in `uniffi/src/interface/types.rs` but it might need some tweaking.
     * Try to `cargo build` the example and confirm that it doesn't work yet.
* Decide on a representation for sequences to cross the FFI layer.
    * Like the existing `Option<T>` type, it seems reasonable to me to always pass them as a bytebuffer.
    * Similar to the strings in https://github.com/rfk/uniffi-rs/pull/25, writing the number of items and then serializing each item in turn to the buffer seems like a reasonable representation in bytes.
* Implement the rust side of that representation in `uniffi/src/support` by implementing the `Liftable` and `Lowerable` traits for `Vec<T>`, and them implementing `ViaFfi` for `Vec<T>` by using them (similar to what's done for `Option<T>`.
* Add support for sequence types in `uniffi/src/scaffolding`, again following the example of option types.
* Add support for sequences in each of the Kotlin, Swift and Python bindings generators.
    * I think we mostly rely on protobufs for this in the current appservices code so you might have to just feel your way around here a bit, again using `Option<T>` as a guide.
* Run `cargo test` to confirm that your updated example code works correctly with sequence types!

data-sync-user:
➤ Edouard Oger commented:

Closed by https://github.com/mozilla/uniffi-rs/commit/aeb4d478845014e1e2ec7d60e5737071b9fa655f

PR #60, opened 2020-07-30
Sequence type support

eoger:
Fixes https://github.com/mozilla/uniffi-rs/issues/28

Note that the python implementation is not complete, so I disabled the test.


examples/todolist/tests/bindings/test_todolist.kts
```diff
@@ -21,3 +21,9 @@ assert(todo.getLast() == "Test Ünicode hàndling without an entry can't believe
 val entry2 = TodoEntry("Test Ünicode hàndling in an entry can't believe I didn't test this at first 🤣")
 todo.addEntry(entry2)
 assert(todo.getLastEntry().text == "Test Ünicode hàndling in an entry can't believe I didn't test this at first 🤣")
+
+assert(todo.getEntries().size == 5)
+
+todo.addEntries(listOf(TodoEntry("foo"), TodoEntry("bar")))
+assert(todo.getEntries().size == 7)
```
tarikeshaq, line 28: Can we have a test verifying the entries from `getEntries`? Like comparing the values of the entries returned themselves

uniffi/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -170,13 +181,26 @@ fun String.Companion.lift(ptr: Pointer): String {
     }
 }
 
+fun<T> lowerSequence(v: List<T>, lowersIntoSize: (T) -> Int, lowerInto: (T, ByteBuffer) -> Unit): RustBuffer.ByValue {
+    return lowerIntoRustBuffer(v, { v -> lowersIntoSizeSequence(v, lowersIntoSize) }, { v, buf -> lowerIntoSequence(v, buf, lowerInto) })
+}
+
+fun<T> lowersIntoSizeSequence(v: List<T>, lowersIntoSize: (T) -> Int): Int {
+    var len = v.size.toLong().lowersIntoSize()
```
tarikeshaq, line 189: I noticed you chose a 64bit value to hold the size, I remember I chose a 32bit one for strings, I wonder if we should be consistent (I have no opinion on whether both should be 4 bytes or 8 bytes, just an observation)

uniffi/src/bindings/python/templates/RustBufferHelper.py
```diff
@@ -46,6 +46,12 @@ def getInt(self):
     def putInt(self, v):
         self._pack_into(4, ">i", v)
 
+    def getLong(self):
+        return self._unpack_from(4, ">Q")
+
+    def putInt(self, v):
```
tarikeshaq, line 52:
```suggestion
    def putLong(self, v):
```

Not a big deal since it's Python 😶

uniffi/src/bindings/python/templates/RustBufferHelper.py
```diff
@@ -46,6 +46,12 @@ def getInt(self):
     def putInt(self, v):
         self._pack_into(4, ">i", v)
 
+    def getLong(self):
+        return self._unpack_from(4, ">Q")
```
tarikeshaq, line 50:
```suggestion
        return self._unpack_from(8, ">Q")
```

Right?

uniffi/src/bindings/kotlin/templates/RustBufferHelpers.kt
tarikeshaq: Oh oh also on a related note, and this is something I did too, we use `Long` Kotlin, but Long is signed. I know it is practically impossible for someone to pass an in-memory value that might cross that limit, but just thought of that. ( I make the same mistake with Strings)

uniffi/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -99,6 +99,9 @@ class Writer {
     }
 
     // Writes an integer in big-endian order.
+    //
+    // Warning: make sure what you are trying to write
```
tarikeshaq, line 103: +1 for this comment 😅

examples/todolist/tests/bindings/test_todolist.kts
tarikeshaq: Also maybe a tests  of lists something that is not a Record, not a big deal thou since I think we'll have to audit our testing anyways eventually 🙃

examples/todolist/src/lib.rs
```diff
@@ -30,6 +30,17 @@ impl TodoList {
         self.items.push(entry.text)
     }
 
+    fn add_entries(&mut self, entries: Vec<TodoEntry>) {
+        self.items.extend(entries.iter().map(|e| e.text.clone()))
```
rfk, line 34: Something to think about for later, but...I reckon we could figure out how to avoid the `clone` here since you own the vec and all its contents.

uniffi/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -78,6 +78,17 @@ fun<T> liftFromOptional(buf: ByteBuffer, liftFrom: (ByteBuffer) -> T): T? {
     return liftFrom(buf)
 }
 
+fun<T> liftSequence(rbuf: RustBuffer.ByValue, liftFrom: (ByteBuffer) -> T): List<T> {
+    return liftFromRustBuffer(rbuf) { buf -> liftFromSequence(buf, liftFrom) }
+}
+
+fun<T> liftFromSequence(buf: ByteBuffer, liftFrom: (ByteBuffer) -> T): List<T> {
+    val len = Math.toIntExact(Long.liftFrom(buf)) // Sorry the JVM can only allocate 2^32 elements.
```
rfk, line 86: In this case, I would lean towards using a `u32` and having the same restriction everywhere.

uniffi/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -279,3 +282,27 @@ extension Optional: Lowerable where Wrapped: Lowerable {
 }
 
 extension Optional: Serializable where Wrapped: Liftable & Lowerable {}
+
+extension Array: Liftable where Element: Liftable {
+    static func lift(from buf: Reader) throws -> Self {
+        // Int is platform dependent, but so is usize (vec size) on the Rust side!
+        let len = Int(truncatingIfNeeded: try UInt64.lift(from: buf))
```
rfk, line 289: If possible, I think we should throw rather than truncate here; it seems better to fail loudly than risk silently doing the wrong thing. Limiting sequence lengths to `u32` might make this problem go away though.

PR #181, opened 2020-07-30
Fixes string sequences

### LINKED ISSUE #173: SYNC-1569 ⁃ Support String sequences as args/return value
It seems like for they are a special case and do not work out of the box (most likely related to `FfiStr`).

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/SYNC-1569)

tarikeshaq:
fixes #173

### DIFF
--- a/uniffi/src/scaffolding.rs
+++ b/uniffi/src/scaffolding.rs
@@ -32,8 +32,15 @@ mod filters {
             TypeReference::String => "&str".to_string(),
             TypeReference::Enum(name) => name.clone(),
             TypeReference::Record(name) => name.clone(),
-            TypeReference::Optional(t) => format!("Option<{}>", type_rs(t)?),
-            TypeReference::Sequence(t) => format!("Vec<{}>", type_rs(t)?),
+
+            // TODO: Clean this up, the `ret_type_rs` is supposed to only be used on return types
+            // and solve the issue of asymmetry with strings.
+            // When a string is a record member or a part of a sequence, we use the `String` type even if the string is an argument
+            // However, when we use strings directly, we use an &str as an argument.
+            // This reeks of future issues, a possible solution is to abandon the asymmetry (and FfiStr) and just use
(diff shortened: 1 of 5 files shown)


uniffi/src/scaffolding.rs
```diff
@@ -32,8 +32,15 @@ mod filters {
...
             TypeReference::Record(name) => name.clone(),
-            TypeReference::Optional(t) => format!("Option<{}>", type_rs(t)?),
-            TypeReference::Sequence(t) => format!("Vec<{}>", type_rs(t)?),
+
+            // TODO: Clean this up, the `ret_type_rs` is supposed to only be used on return types
+            // and solve the issue of asymmetry with strings.
+            // When a string is a record member or a part of a sequence, we use the `String` type even if the string is an argument
+            // However, when we use strings directly, we use an &str as an argument.
+            // This reeks of future issues, a possible solution is to abandon the asymmetry (and FfiStr) and just use
+            // `String` everywhere as a ByteBuffer and give up an extra memory allocation
```
rfk, line 41: This makes sense in the short term. I've been noodling a bit on the problems we've had around the asymmetry with strings here, and e.g. the `RawPointer` hack we had to add for the destructor. I think there's a bit of cleanup we can do in the typing here to simplify it, but I don't fully have my head around it yet. This is a great extra input to consider.

rfk: Could you please file a follow-up issue for this?

rfk:
> I've been noodling a bit on the problems we've had around the asymmetry with strings here,

Broadly, there are two enhancements to the (currently very sparse) typesystem we've built here that might be useful:

* Distinguishing between API-level and FFI-level types as two distinct sets.
* Distinguishing more explicitly between "Owned" and "Borrowed" types.

I'm going to try to experiment a little bit and see if anything nice pops out.

uniffi/src/scaffolding.rs
tarikeshaq: #182 ;)

Issue #36, opened 2020-07-28
Enum field in dictionary support


eoger:
I want to be able to compile this :
```
enum BoboType {
  "TRON",
  "ZORD",
};

dictionary TestMock {
  BoboType name_suffix_variant;
};
```

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1554)

tarikeshaq:
+1 For `BoboType` 🤣 

I think this is a good first issue! 

To make this possible:
- One would need to implement `Lowerable` and `Liftable` for enum types, the traits can be found [here](https://github.com/mozilla/uniffi-rs/blob/main/uniffi/src/support/mod.rs)
- This could fit nicely in [here](https://github.com/mozilla/uniffi-rs/blob/main/uniffi/src/templates/EnumTemplate.rs)
- Probably do the lifting and lowering **very** similar to how `ViaFfi` is implemented on the above file

rfk:
> One would need to implement Lowerable and Liftable for enum types

Right, and in practice, these could probably just call `into_ffi_value`/`try_from_ffi_value` to convert to a `u32` and then read/write it directly in the buffer.

PR #184, opened 2020-07-30
Implement Lowerable/Liftable for enums

eoger:
Fixes https://github.com/mozilla/uniffi-rs/issues/36.

Summary of changes:
- Impl Lowerable/Liftable for enums
- `into_ffi_value` does not consume `self` anymore, uses `&self` instead (had to do this because otherwise we would require enums to implement `Copy` or `Clone`)
- Added the `rondpoint` example, it does round-tripping testing.

### DIFF
--- a/examples/rondpoint/README.md
+++ b/examples/rondpoint/README.md
@@ -1,0 +1,38 @@
+# Example uniffi component: "Rondpoint"
+
+This example is to exercise complex data types through roundtripping.
+
+* [`./src/rondpoint.idl`](./src/rondpoint.idl), the component interface definition which defines the main object and its methods. This is processed by functions in [`./build.rs`](./build.rs)
+  to generate Rust scaffolding for the component.
+* [`./src/lib.rs`](./src/lib.rs), the core implementation of the component in Rust. This basically
+  pulls in the generated Rust scaffolding via `include!()` and fills in function implementations.
+* [`./src/main.rs`](./src/main.rs) generates a helper executable that can be used for working with
+  foreign language bindings, while guaranteeing that those bindings use the same version of `uniffi`
+  as the compiled component.
+* Some small test scripts that double as API examples in each target foreign language:
+  * Kotlin [`./tests/bindings/test_rondpoint.kts`](./tests/bindings/test_rondpoint.kts)
+  * Swift [`./tests/bindings/test_rondpoint.swift`](./tests/bindings/test_rondpoint.swift)
+  * Python [`./tests/bindings/test_rondpoint.py`](./tests/bindings/test_rondpoint.py)
+
+If you want to try it out, you will need:
+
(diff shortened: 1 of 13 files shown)


uniffi/src/scaffolding.rs
```diff
@@ -77,7 +77,7 @@ mod filters {
         // By explicitly naming the type here, we help the rust compiler to type-check the user-provided
         // implementations of the functions that we're wrapping (and also to type-check our generated code).
         Ok(format!(
-            "<{} as uniffi::support::ViaFfi>::into_ffi_value({})",
+            "<{} as uniffi::support::ViaFfi>::into_ffi_value(&{})",
```
tarikeshaq, line 80: I'm wondering whether it's best to have `into_ffi_value` borrow `self`, or make `Lowerable::lower_into` own `self`. Asking because the `Into::into` in the standard library takes ownership, and I'm not sure if that does some stack memory magic 😆. Probably doesn't make a difference, but wondering what you think!

examples/rondpoint/src/lib.rs
```diff
@@ -0,0 +1,29 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
tarikeshaq, line 1: Love this 🤣

examples/rondpoint/README.md
```diff
@@ -0,0 +1,38 @@
...
+  * Python [`./tests/bindings/test_rondpoint.py`](./tests/bindings/test_rondpoint.py)
+
+If you want to try it out, you will need:
+
+* The [Kotlin command-line tools](https://kotlinlang.org/docs/tutorials/command-line.html), particularly `kotlinc`.
+* The [Java Native Access](https://github.com/java-native-access/jna#download) JAR downloaded and its path
+  added to your `$CLASSPATH` environment variable.
+* Python 3
+* The [Swift command-line tools](https://swift.org/download/), particularly `swift`, `swiftc` and
+  the `Foundation` package.
```
rfk, line 24: As a meta-level comment, we can probably figure out how to consolidate these detailed how-to docs from the examples in a single place, rather than copy-pasting them for each one. Not for this PR though.

uniffi/src/scaffolding.rs
rfk, line 80: This seems OK to me, but I do wonder if there are cases where we actually *want* to take an owned reference, e.g. so we can transfer ownership to the foreign language side without a copy. That can be for follow-up work though.

rfk:
Alternately, if we don't want to make these changes to `ViaFfi` ownership, we could implement `Lowerable`/`Liftable` more directly instead of trying to re-use `into_ffi_value`.

eoger:
Let's revisit the `self` ownership `into_ffi_value` another time if we find that it's problematic

PR #186, opened 2020-07-31
Support argument reference in Rust code

### LINKED ISSUE #4: Add support for reference arguments in the rust code

weslenng:
This PR add support to argument reference in Rust code
Closes #4

### DIFF
--- a/uniffi/src/interface/mod.rs
+++ b/uniffi/src/interface/mod.rs
@@ -463,10 +463,15 @@ impl APIConverter<Argument> for weedle::argument::SingleArgument<'_> {
             type_: (&self.type_).resolve_type_definition(ci)?,
             by_ref: match &self.attributes {
                 None => false,
-                Some(def) => def.body.list.iter().find(| attr| match attr {
-                    weedle::attribute::ExtendedAttribute::NoArgs(attr) => (attr.0).0 == "ByRef",
-                    _ => false
-                }).is_some()
+                Some(def) => def
+                    .body
+                    .list
+                    .iter()
+                    .find(|attr| match attr {
+                        weedle::attribute::ExtendedAttribute::NoArgs(attr) => (attr.0).0 == "ByRef",
+                        _ => false,
+                    })
+                    .is_some(),
             },
             optional: self.optional.is_some(),
             default: match self.default {
--- a/uniffi/src/interface/mod.rs
+++ b/uniffi/src/interface/mod.rs
@@ -458,11 +458,6 @@ impl APIConverter<Argument> for weedle::argument::Argument<'_> {
 
 impl APIConverter<Argument> for weedle::argument::SingleArgument<'_> {
     fn convert(&self, ci: &ComponentInterface) -> Result<Argument> {
-        // by_ref: Attributes::try_from(&self.attributes)?.0.iter().iter().any(|attr| match attr {
-        //     Attribute::ByRef => true,
-        //     _ => false,
-        // })
-
         Ok(Argument {
             name: self.identifier.0.to_string(),
             type_: (&self.type_).resolve_type_definition(ci)?,
--- a/uniffi/src/interface/mod.rs
+++ b/uniffi/src/interface/mod.rs
@@ -458,20 +458,22 @@ impl APIConverter<Argument> for weedle::argument::Argument<'_> {
 
 impl APIConverter<Argument> for weedle::argument::SingleArgument<'_> {
     fn convert(&self, ci: &ComponentInterface) -> Result<Argument> {
+        // by_ref: Attributes::try_from(&self.attributes)?.0.iter().iter().any(|attr| match attr {
+        //     Attribute::ByRef => true,
+        //     _ => false,
+        // })
+
         Ok(Argument {
             name: self.identifier.0.to_string(),
             type_: (&self.type_).resolve_type_definition(ci)?,
             by_ref: match &self.attributes {
                 None => false,
-                Some(def) => def
-                    .body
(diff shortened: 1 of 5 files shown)


uniffi/src/interface/mod.rs
```diff
@@ -451,12 +458,16 @@ impl APIConverter<Argument> for weedle::argument::Argument<'_> {
...
         Ok(Argument {
             name: self.identifier.0.to_string(),
             type_: (&self.type_).resolve_type_definition(ci)?,
+            by_ref: match &self.attributes {
+                None => false,
+                Some(def) => def.body.list.iter().find(| attr| match attr {
+                    weedle::attribute::ExtendedAttribute::NoArgs(attr) => (attr.0).0 == "ByRef",
+                    _ => false
+                }).is_some()
+            },
```
rfk, line 470: @tarikeshaq you've been working with attributes a bit over in https://github.com/mozilla/uniffi-rs/pull/31, is there anything from that PR that's worth extracting and landing so it can be re-used here?

rfk:
Thanks for diving in here! I won't have a chance to give this a thorough look right away, but hopefully some other folks can chime in. It looks pretty good from what I can see.

One small request: please run `cargo fmt` to standardize the formatting of the code (which will unblock the one failing CI test on this PR).

weslenng:
> Thanks for diving in here! I won't have a chance to give this a thorough look right away, but hopefully some other folks can chime in. It looks pretty good from what I can see.
> 
> One small request: please run `cargo fmt` to standardize the formatting of the code (which will unblock the one failing CI test on this PR).

Thanks!
I ran `cargo fmt`, Ci should be good now

uniffi/src/interface/mod.rs
tarikeshaq: Yup! Opened #187 for that!

rfk:
Cross-linking for context, @weslenng what do you think about the suggestion in https://github.com/mozilla/uniffi-rs/pull/187#discussion_r463382693 ?

weslenng:
> Cross-linking for context, @weslenng what do you think about the suggestion in [#187 (comment)](https://github.com/mozilla/uniffi-rs/pull/187#discussion_r463382693) ?

It's a good suggestion
I thought there was already something like that in the codebase (but I couldn't find 😅)

Probably I will work on it tomorrow!

rfk:
>  I thought there was already something like that in the codebase (but I couldn't find it 😅)

Heh, honestly, I expect you will find quite a few things missing like that at this stage; it's a pretty new codebase that we're working on moving beyond "quick prototype" stage.

weslenng:
@rfk @tarikeshaq Now is using the `impl` of PR #187, but I'm not sure if is the best approach
Can you guys review it?

PR #187, opened 2020-07-31
Adds attribute parsing

tarikeshaq:
Separates attribute parsing from #31

### DIFF
--- a/uniffi/src/interface/mod.rs
+++ b/uniffi/src/interface/mod.rs
@@ -49,7 +49,7 @@
 //!   * Error messages leave a lot to be desired (and we currently panc on errors parsing the WebIDL).
 //!     If this were to be a real thing we'd need to invest in more empathetic error messages.
 
-use std::{collections::hash_map::Entry, collections::HashMap, str::FromStr};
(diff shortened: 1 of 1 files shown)


rfk:
The relevant discussion from that PR, for context: https://github.com/mozilla/uniffi-rs/pull/31#discussion_r460381577

uniffi/src/interface/mod.rs
```diff
@@ -788,3 +788,41 @@ impl APIConverter<Literal> for weedle::literal::DefaultValue<'_> {
         })
     }
 }
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum Attribute {
+    // Add valid attributes here
```
rfk, line 794: @weslenng I'm curious what you think about using this over in #186. The idea would be to add a `ByRef` variant to the `enum Attribute` here, and a matching case in the `impl TryFrom` below it.

In your PR #186, you would then use it when determining the `by_ref` field, something like this:

```
            by_ref: Attributes::try_from(&self.attributes)?.0.iter().iter().any(|attr| match attr {
                        Attribute::ByRef => true,
                        _ => false,
                    })
```

This would help us avoid a bit of duplicate work between #186 and #31; what do you think?

PR #191, opened 2020-08-01
Error on duplicate attributes

weslenng:
I'm not sure if it's the most efficient way to do this
Suggestions are welcome!

### DIFF
--- a/examples/sprites/src/sprites.idl
+++ b/examples/sprites/src/sprites.idl
@@ -1,6 +1,6 @@
 
 namespace sprites {
-  Point translate([ByRef] Point position, Vector direction);
+  Point translate([ByRef, ByRef] Point position, Vector direction);
 };
 
 dictionary Point {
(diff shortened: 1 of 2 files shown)


rfk:
(Ordinarily, I would ask you to please add some tests for this, but given that the test coverage of this file is currently *extremely* light on I don't think that would be fair to you right now. Adding test coverage here is on our todo list though.)

rfk:
If we wanted to be slightly more efficient we could do this as a `for` loop, checking in the `HashSet` and then appending to the list of attributes. But I don't think it's worth it in this case, the readability here will be more important than the performance.

weslenng:
> We could consider making Attributes directly contain a `HashSet` rather than a `Vec`

I hadn't thought about that, but it's a interesting idea

> (Ordinarily, I would ask you to please add some tests for this, but given that the test coverage of this file is currently _extremely_ light on I don't think that would be fair to you right now. Adding test coverage here is on our todo list though.)

I think I can do that, it's being fun to learn Rust

Issue #32, opened 2020-07-28
Use panic-safety helper functions from ffi-support


rfk:
Panics in the rust code are not currently handled very well, and I suspect would actually trigger undefined behaviour by propagating across the FFI boundary.  We should use the support functions offered by [ffi-support](https://docs.rs/ffi-support/0.4.0/ffi_support/index.html) to ensure safe panic handling in all cases.

The most basic thing that could work:

* All ffi functions should have an `ExternError` out parameter
* The rust scaffolding for all standalone functions and object methods should wrap the call in [`call_with_output`](https://docs.rs/ffi-support/0.4.0/ffi_support/fn.call_with_output.html).

Object methods could instead rely on the `call` helpers of the `HandleMap` for the same effect, but I think it would be worth our while to catch and handle panics in argument conversion or other operations that might occur before we call into the `HandleMap` at all.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1556)

rfk:
For context, note that we currently panic if we receive unexpected data over the FFI, so we need to make that safe as well; ref https://github.com/rfk/uniffi-rs/pull/31#discussion_r460382079

PR #192, opened 2020-08-01
Clarify type conversions by differentiating between API-level and FFI-level types.

### LINKED ISSUE #182: Using ByteBuffer for Strings
One issue we had to deal with when implementing strings in #25 is what to use to represent them. We chose the way it is done in Application Services, which is to use `ffi_support::FfiStr`. This abstraction is great because it saves us some memory allocations and abstracts the null-terminated string that is passed in.

The drawback is it creates asymmetry with strings, we end up lifting a `&str` but lowering a `String`. This is alright and we can live with it, BUT, when we have strings as a part of a `Record`, since a record is serialized into a `ByteBuffer`, the string has to as well.

Now, this creates an extra layer of asymmetry with strings, to sum it up:

- When a string is taken as an argument, we map an `ffi_support::FfiStr` to a `&str`
- When a string is returned, we map a `String` to a `*mut c_char` and provide a free function the caller can use, so that they don't need to do any copying
- When a string is a part of a record/sequence we serialize the string into the ByteBuffer that the record is serialized into

This creates a lot of confusion but saves us performance in the case that strings are being passed in or returned directly.

If we are to use ByteBuffers for Strings everywhere, we'd have exactly one way to deal with strings, BUT, it would mean multiple copying (As we already do with Records and Sequences anyways)

Thoughts?

Hopefully whatever we can come up with can help solve #181 in a better way

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1583)

rfk:
Fixes #182, kind of. This is a big PR but a lot of it is mechanical.

I spent a bit of time noodling on the difficulty we had in handling strings, with an asymmetry between how they appear in arguments, return values, and records. I think the root of the problem was the way that we were smooshing together different types of types.

Previously, we had a single set of types (the `TypeReference` enum) which were used both in API function signatures, and in the signatures of internal FFI functions. This was nice and simple to get started but involved a lot of duplication and mental tracking of the implicit conversions that happen when sending data across the FFI.

This commit splits the types into two different...um...types. We have the `Type` enum to represent the rich high-level types that appear in the component API, and a separate `FFIType` enum to represent the much more constrained set of types that
are used to pass data back-and-forth across the FFI.

To prove out the approach, this commit also removes the asymmetry in how strings are handled in argument position vs return position. The API-level `Type::String` always corresponds to an owned Rust string, and is passed across the FFI as a pointer
represented by the low-level `FFIType::RustString`. Foreign language code can create a `FFIType::RustString` using a new helper function that takes a `FFI::ForeignStringRef` and copies it into the rust world.

The result is a smaller number of moving part in the foreign language binding generation, which I think is a nice little win.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
@@ -60,7 +60,7 @@ class Reader {
             throw InternalError.bufferOverflow
         }
         var value = [UInt8](repeating: 0, count: count)
-        value.withUnsafeMutableBufferPointer({ buffer in 
+        value.withUnsafeMutableBufferPointer({ buffer in
             data.copyBytes(to: buffer, from: range)
         })
         offset = range.upperBound
@@ -140,8 +140,13 @@ extension String: Liftable, Lowerable {
         }
         return String(cString: v)
     }
-    func toFFIValue() -> Self {
-        self
+    func toFFIValue() -> UnsafeMutablePointer<CChar> {
+        var rustErr = NativeRustError(code: 0, message: nil)
+        let rustStr = {{ ci.ffi_string_alloc_from().name() }}(self, &rustErr)
+        if rustErr.code != 0 {
+            fatalError("caught a panic while passing a string across the ffi")
+        }
+        return rustStr
     }
 
     static func lift(from buf: Reader) throws -> Self {
--- a/uniffi_bindgen/src/templates/RustString.rs
+++ b/uniffi_bindgen/src/templates/RustString.rs
@@ -1,0 +1,41 @@
+// Everybody gets basic string support, since they're such a common data type.
+
+/// This helper receives a borrowed foreign language string and
+/// copies it into an owned rust string. It is naturally tremendously unsafe,
+/// because pointers. The main things to remember as a consumer are:
+///
+///   * You must keep the argument buffer alive for the duration of the call,
+///     and avoid mutating it e.g. from other threads.
+///   * You must eventually free the return value by passing it to the string
+///     destructor defined below. Not just "a string destructor like the one
+///     defined below", the very one defined below for this component!
+///
+/// This function will report a panic via its `ExternError` out arg if given
+/// a null or invalid pointer, or a buffer containing non-utf8 bytes.
+#[no_mangle]
+pub unsafe extern "C" fn {{ ci.ffi_string_alloc_from().name() }}(cstr: *const std::os::raw::c_char, err: &mut uniffi::deps::ffi_support::ExternError) -> *mut std::os::raw::c_char {
+    uniffi::deps::ffi_support::call_with_output(err, || {
+        // This logic was copied from ffi_support::FfiStr, changed to panic on invalid data.
+        // We should figure out whether/how to move this back into ffi_support.
+        if cstr.is_null() {
+            panic!("null pointer provided to string_alloc_from");
+        }
+        // This copies the data into a rust-allocated string.
+        let cstr = std::ffi::CStr::from_ptr(cstr);
+        let s = cstr.to_str().expect("Invalid utf8 in foreign string buffer");
+        // And this lowers it back into a rust-owned pointer to return over the FFI.
+        <String as uniffi::ViaFfi>::into_ffi_value(s.to_string())
+    })
+}
+
+/// Free a String that had previously been passed to the foreign language code.
+/// The argument *must* be a uniquely-owned pointer previously obtained from a call
+/// into the rust code that returned a string.
+///
+#[no_mangle]
+pub unsafe extern "C" fn {{ ci.ffi_string_free().name() }}(cstr: *mut std::os::raw::c_char) {
+    // We deliberately don't check the `Result` here, so that callers don't need to pass an out `err`.
+    // There was nothing for us to free in the failure case anyway, so no point in propagating the error.
+    let s = <String as uniffi::ViaFfi>::try_from_ffi_value(cstr);
+    drop(s)
+}
--- a/uniffi/src/lib.rs
+++ b/uniffi/src/lib.rs
@@ -6,6 +6,7 @@ use anyhow::{bail, Result};
 use bytes::buf::{Buf, BufMut};
 use ffi_support::ByteBuffer;
 use std::convert::TryFrom;
+use std::ffi::CString;
 
 // It would be nice if this module was behind a cfg(test) guard, but it
 // doesn't work between crates so let's hope LLVM tree-shaking works well.
@@ -152,7 +153,7 @@ impl<T: Liftable> Liftable for Vec<T> {
 
 pub unsafe trait ViaFfi: Sized {
     type Value;
-    fn into_ffi_value(&self) -> Self::Value;
+    fn into_ffi_value(self) -> Self::Value;
     fn try_from_ffi_value(v: Self::Value) -> Result<Self>;
 }
 
@@ -160,7 +161,7 @@ macro_rules! impl_via_ffi_for_primitive {
   ($($T:ty),+) => {$(
     unsafe impl ViaFfi for $T {
       type Value = Self;
-      #[inline] fn into_ffi_value(&self) -> Self::Value { *self }
+      #[inline] fn into_ffi_value(self) -> Self::Value { self }
       #[inline] fn try_from_ffi_value(v: Self::Value) -> Result<Self> { Ok(v) }
     }
   )+}
@@ -173,7 +174,7 @@ pub trait ViaFfiUsingByteBuffer: Liftable + Lowerable {}
 unsafe impl<T: ViaFfiUsingByteBuffer> ViaFfi for T {
     type Value = ffi_support::ByteBuffer;
     #[inline]
-    fn into_ffi_value(&self) -> Self::Value {
+    fn into_ffi_value(self) -> Self::Value {
         lower(&self)
     }
     #[inline]
@@ -185,7 +186,7 @@ unsafe impl<T: ViaFfiUsingByteBuffer> ViaFfi for T {
 unsafe impl<T: Liftable + Lowerable> ViaFfi for Option<T> {
     type Value = ffi_support::ByteBuffer;
     #[inline]
-    fn into_ffi_value(&self) -> Self::Value {
+    fn into_ffi_value(self) -> Self::Value {
         lower(&self)
     }
     #[inline]
@@ -197,7 +198,7 @@ unsafe impl<T: Liftable + Lowerable> ViaFfi for Option<T> {
 unsafe impl<T: Liftable + Lowerable> ViaFfi for Vec<T> {
     type Value = ffi_support::ByteBuffer;
     #[inline]
-    fn into_ffi_value(&self) -> Self::Value {
+    fn into_ffi_value(self) -> Self::Value {
         lower(&self)
     }
     #[inline]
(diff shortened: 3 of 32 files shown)


uniffi/src/bindings/swift/gen_swift.rs
```diff
@@ -76,96 +76,72 @@ mod filters {
     use super::*;
     use std::fmt;
 
-    /// Declares a C type in the bridging header.
-    pub fn decl_c(type_: &TypeReference) -> Result<String, askama::Error> {
+    /// Declares a Swift type in the public interface for the library.
+    pub fn type_swift(type_: &Type) -> Result<String, askama::Error> {
```
rfk, line 80: These were renamed and re-ordered for consistency with the other language backends.

uniffi/src/interface/mod.rs
```diff
@@ -8,36 +8,36 @@
 
 //! # Component Interface Definition and Types
```
rfk, line 9: The changes in this file are broadly:

* Modifying the ffi-level functions to use ffi-level types.
* Adding the new string-alloc helper function.
* Adding some docs and general code cleanup while I was in here.

uniffi/src/interface/mod.rs
```diff
@@ -270,14 +293,15 @@ impl FromStr for ComponentInterface {
...
+/// uses them to build up the records, enums, objects etc in the provided
+/// `ComponentInterface`.
 trait APIBuilder {
     fn process(&self, ci: &mut ComponentInterface) -> Result<()>;
 }
 
-trait APIConverter<T> {
-    fn convert(&self, ci: &ComponentInterface) -> Result<T>;
-}
-
```
rfk, line 280: I moved some trait implementations around to put them closer to the structs they operate on, in the hope of making it easier to digest the code one piece at a time.

uniffi/src/scaffolding.rs
```diff
@@ -22,81 +22,64 @@ mod filters {
...
+            Type::Sequence(t) => format!("Vec<{}>", type_rs(t)?),
         })
     }
 
-    pub fn type_c(type_: &TypeReference) -> Result<String, askama::Error> {
+    pub fn type_ffi(type_: &FFIType) -> Result<String, askama::Error> {
         Ok(match type_ {
-            // Objects don't currently impl `ViaFfi`.
-            TypeReference::Object(_) => "u64".to_string(),
-            _ => format!("<{} as uniffi::support::ViaFfi>::Value", type_rs(type_)?,),
```
rfk, line 68: Kinda sad to lose this neat trick, but ah well...

uniffi/src/support/mod.rs
```diff
@@ -194,37 +195,33 @@ unsafe impl<T: Liftable + Lowerable> ViaFfi for Vec<T> {
...
-}
-
+/// Support for passing Strings back and forth across the FFI.
+///
+/// Unlike many other implementations of `ViaFfi`, this passes a pointer rather
+/// than copying the data from one side to the other. This is a safety hazard,
+/// but turns out to be pretty nice for useability.
+///
+/// (In practice, we do end up copying the data, the copying just happens on
+/// the foreign language side rather than here in the rust code.)
```
rfk, line 205: If we don't like the unsafety here, we could implement this logic on top of `RustBuffer` and do a little extra copying. But the `CString` stuff seems sufficiently well-developed that we should be able to manage the safety, similar to how ffi_support manages it internally.

uniffi/src/support/mod.rs
```diff
@@ -194,37 +195,33 @@ unsafe impl<T: Liftable + Lowerable> ViaFfi for Vec<T> {
...
 unsafe impl ViaFfi for String {
     type Value = *mut std::os::raw::c_char;
-    fn into_ffi_value(&self) -> Self::Value {
+
+    // This returns a raw pointer to the underlying bytes, so it's very important
+    // that it consume ownership of the String, which is relinquished to the foreign
+    // language code (and can be returned by it passing the pointer back).
```
rfk, line 211: Also this meant that I had to revert `into_ffi_value` to consuming `self` by value.

uniffi/src/scaffolding.rs
```diff
@@ -22,81 +22,62 @@ mod filters {
...
-            TypeReference::U32 => "u32".to_string(),
-            TypeReference::U64 => "u64".to_string(),
-            TypeReference::Float => "f32".to_string(),
-            TypeReference::Double => "f64".to_string(),
-            TypeReference::Boolean => "bool".to_string(),
-            TypeReference::String => "&str".to_string(),
-            TypeReference::Enum(name) => name.clone(),
-            TypeReference::Record(name) => name.clone(),
-
-            // TODO: Clean this up, the `ret_type_rs` is supposed to only be used on return types
```
rfk, line 36: Oh yeah, this got cleaned up as well.

uniffi/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -164,8 +164,8 @@ fun Double.lowerInto(buf: ByteBuffer) {
     buf.putDouble(this)
 }
 
-fun String.lower(): String {
-    return this
+fun String.lower(): Pointer {
+    return _UniFFILib.INSTANCE.{{ ci.ffi_string_alloc_from().name() }}(this)
```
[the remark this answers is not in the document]
tarikeshaq: I like it!
One clarification (in case I'm missing something 😆 ):

Here's what the process looks like in my head:
- Kotlin wants to call todo.addItem("Make a review comment!")
- Kotlin then will call `String.lower()`, which in turn calls `string_alloc` to create a Rust allocation for the String
- Kotlin will pass that Rust owned string to the `addItem` function across the FFI
- The pointer is back in rust land, and Rust re-takes ownership of that string and appropriately drops it when it goes out of scope
- The function is done, we are happy and all memory used is freed.

Sound about right? Asking because I got confused there for a second thinking that strings created here would need an explicit `free` once the function is done, but that is covered by Rust re-taking the ownership right?

_Wonders if there is a scenario where things can go wrong in between the allocation and the function call causing a leak_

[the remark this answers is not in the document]
rfk: > Sound about right? 

Yes, exactly. This is similar to how lowering into a buffer works, where e.g. `Record.lower()` allocates and fills a buffer, passes the buffer to the rust code, and relies on the rust code to free it.

Which...now that you point it out, we don't actually do, because [`ByteBuffer` doesn't implement `Drop`](https://docs.rs/ffi-support/0.4.0/ffi_support/struct.ByteBuffer.html#caveats) 🤦.

I'll file a separate bug to dig into that.

uniffi/src/scaffolding.rs
```diff
@@ -22,81 +22,62 @@ mod filters {
...
+            FFIType::UInt16 => "u16".to_string(),
+            FFIType::Int32 => "i32".to_string(),
+            FFIType::UInt32 => "u32".to_string(),
+            FFIType::Int64 => "i64".to_string(),
+            FFIType::UInt64 => "u64".to_string(),
+            FFIType::Float32 => "f32".to_string(),
+            FFIType::Float64 => "f64".to_string(),
+            FFIType::RustBuffer => "ffi_support::ByteBuffer".into(),
+            FFIType::RustString => "*mut std::os::raw::c_char".into(),
+            FFIType::ForeignStringRef => "ffi_support::FfiStr<'_>".into(),
```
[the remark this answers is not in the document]
rfk: Yes, good catch, this is left over from a previous factoring (and actually it's totally unused, because the only function that takes one of these is `string_free`, and that specifics its signature by hand.

uniffi/src/bindings/kotlin/templates/RustBufferHelpers.kt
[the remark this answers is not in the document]
rfk: I filed https://github.com/mozilla/uniffi-rs/issues/194 about freeing bytebuffers, but I think we may actually be OK due to the way we access the underlying data.

I also filed https://github.com/mozilla/uniffi-rs/issues/195 about the possibility of leaking memory if we get interrupted part-way through lowering arguments to pass over the FFI.

uniffi/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
```diff
@@ -21,9 +21,9 @@ internal interface _UniFFILib : Library {
 
     {% for func in ci.iter_ffi_function_definitions() -%}
     fun {{ func.name() }}(
-        {%- call kt::arg_list_rs_decl(func.arguments()) %}
+        {%- call kt::ffi_arg_list_decl(func.arguments()) %}
```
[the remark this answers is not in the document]
jhugman: :clap:

uniffi/src/bindings/swift/gen_swift.rs
jhugman: I've been meaning to do this. Thank you.

uniffi/src/support/mod.rs
```diff
@@ -194,37 +195,33 @@ unsafe impl<T: Liftable + Lowerable> ViaFfi for Vec<T> {
...
+/// than copying the data from one side to the other. This is a safety hazard,
+/// but turns out to be pretty nice for useability.
+///
+/// (In practice, we do end up copying the data, the copying just happens on
+/// the foreign language side rather than here in the rust code.)
 unsafe impl ViaFfi for String {
     type Value = *mut std::os::raw::c_char;
-    fn into_ffi_value(&self) -> Self::Value {
+
+    // This returns a raw pointer to the underlying bytes, so it's very important
```
rfk, line 209: @thomcc could you please take a peek at what I'm doing with strings here and see if there's any unsafety that I've missed? It's based on the string-handling code in ffi-support, but structured a little differently to fit in with the way the lifting/lowering logic here works.

uniffi/src/templates/RustString.rs
```diff
@@ -0,0 +1,32 @@
...
+///
+/// Panics if given an invalid or nul pointer, or if the buffer contains
+/// invalid UTF-8.
+#[no_mangle]
+pub unsafe extern "C" fn {{ ci.ffi_string_alloc_from().name() }}(cstr: *const std::os::raw::c_char) -> *mut std::os::raw::c_char {
+    // This logic copied from ffi_support::FfiStr, changed to panic on invalid utf-8.
+    // XXX TODO: confirm whether this panics on null.
+    let cstr = std::ffi::CStr::from_ptr(cstr);
+    let s = cstr.to_str().expect("Invalid utf8 in foreign sting buffer");
+    <String as uniffi::support::ViaFfi>::into_ffi_value(s.to_string())
```
rfk, line 21: @thomcc also this bit, which is basically `FfiStr but panics on invalid utf8`.

rfk:
This has now been rebased on top of https://github.com/mozilla/uniffi-rs/pull/205 and on top of the error-support PR, which was a bit of a best of a merge! But hopefully everything here is still intact...

rfk:
> You need to avoid panicing over the FFI. If this is a pervasive problem in uniffi feel free to ignore for now and address later

Thanks, good call - I was mentally deferring this to https://github.com/mozilla/uniffi-rs/issues/32 but that only covers the *generated* functions, not these hand-written helper ones. I should add the panic-handling stuff to those while I'm in here.

uniffi/src/templates/RustString.rs
```diff
@@ -0,0 +1,32 @@
...
+///   * You must eventually free the return value by passing it to the string
+///     destructor defined below. Not just "a string destructor like the one
+///     defined below", the very one defined below for this component!
+///
+/// Panics if given an invalid or nul pointer, or if the buffer contains
+/// invalid UTF-8.
+#[no_mangle]
+pub unsafe extern "C" fn {{ ci.ffi_string_alloc_from().name() }}(cstr: *const std::os::raw::c_char) -> *mut std::os::raw::c_char {
+    // This logic copied from ffi_support::FfiStr, changed to panic on invalid utf-8.
+    // XXX TODO: confirm whether this panics on null.
```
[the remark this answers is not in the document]
rfk: (Just to note for completeness, I have no reason to think that `FfiStr` mis-handles null in this case, the TODO here is entirely from my own mental unwinding of all the things that `FfiStr` does)

uniffi/src/templates/RustString.rs
```diff
@@ -0,0 +1,32 @@
+// Everybody gets basic string support, since they're such a common data type.
+
+/// This helper receives a borrowed foreign language string and
+/// copies it into an owned rust string. It is naturally tremendously unsafe,
+/// because pointers. The main things to remember as a consumer are:
+///
+///   * You must keep the argument buffer alive and static for the duration
```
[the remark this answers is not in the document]
rfk: I mean "unchanging", and will re-word appropriately 👍

uniffi_bindgen/src/templates/RustString.rs
```diff
@@ -0,0 +1,41 @@
...
+///     and avoid mutating it e.g. from other threads.
+///   * You must eventually free the return value by passing it to the string
+///     destructor defined below. Not just "a string destructor like the one
+///     defined below", the very one defined below for this component!
+///
+/// This function will report a panic via its `ExternError` out arg if given
+/// a null or invalid pointer, or a buffer containing non-utf8 bytes.
+#[no_mangle]
+pub unsafe extern "C" fn {{ ci.ffi_string_alloc_from().name() }}(cstr: *const std::os::raw::c_char, err: &mut uniffi::deps::ffi_support::ExternError) -> *mut std::os::raw::c_char {
+    uniffi::deps::ffi_support::call_with_output(err, || {
```
rfk, line 17: @linacambridge a quick heads-up re: https://github.com/mozilla/uniffi-rs/issues/32, just to note that I manually added panic-handling to this new helper function. We'll probably want to nuance it a bit with whatever more sophisticated thing lands in #32.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -140,8 +140,13 @@ extension String: Liftable, Lowerable {
         }
         return String(cString: v)
     }
-    func toFFIValue() -> Self {
-        self
+    func toFFIValue() -> UnsafeMutablePointer<CChar> {
+        var rustErr = NativeRustError(code: 0, message: nil)
+        let rustStr = {{ ci.ffi_string_alloc_from().name() }}(self, &rustErr)
+        if rustErr.code != 0 {
+            fatalError("caught a panic while passing a string across the ffi")
```
rfk, line 147: I manually tested this by inserting a deliberate `panic!` on the rust side, then confirming that the swift (and kotlin) side tests correctly surfaced the panic.

rfk:
I've added very basic panic handling for `string_alloc_from`, and broader panic handling will come in https://github.com/mozilla/uniffi-rs/issues/32 (which is indeed marked as blocking production use). Thanks @thomcc!

Issue #10, opened 2020-06-26
Handle name collisions with internal helpers


linabutler:
All bindings define some internal helper methods on the records and classes they generate—things like `lift`, `lower`, `liftFrom`, `lowerInto`, `lowersToSize`, `toFFIValue`, and `fromFFIValue`—as well as internal types like `RustBufferStream`. What happens if a type declares a method in its IDL with the same name as one of those internal helpers?

This is broadly related to #2, although more specialized.

**Approach one**

* Blanket ban on defining methods with those names. This is probably the easiest and also most correct approach—we make the programmer think of a better name instead of trying to make it work. If we keep the set of reserved method names consistent across all our language bindings (even if they’re not necessarily consistent with the naming conventions of the target language—this is internal generated code, after all), we can keep the size of the blocklist down. Or, we can introduce language-specific blocklists. So `RustBufferStream` would be blocked only in Python, `liftFrom`, `lowerInto`, and `lowersIntoSize` only in Kotlin, and `toFFIValue` and `fromFFIValue` only in Swift.

A benefit of language-specific lists would also help us block accidental use of reserved words. For example, `def` is reserved in Python, but none of the others; `trait` can’t be an argument name in Rust, but can be used freely in others; `func` is reserved only in Swift, `fun` only in Kotlin, and so on. Of course, we could also block these globally—so you couldn’t use `def`, `func`, or `trait` as an argument name in any language.

Drawbacks: authors have to think of new names themselves, and it doesn’t handle extensions and subclasses (do we care?).

**Approach two**. Try to uniquify internal helper names. This would be more helpful for authors, but feels too magical, and has lots of corner cases and hacks. The idea is, we’d have a list of all user-defined method names, and generate a unique one for the internal helper only if there’s a collision. It keeps the generated code clean in most cases, since we’d only mangle names if we see a collision.

There are drawbacks to this, though. First, for methods, we have to rename the whole method globally, even if there’s only one collision. As soon as a type defines, say, a `lower` method, we have to call our internal one `lower0` everywhere. It means we have to keep track of more state, making the parser more complicated. Also, each language has its own notion of what’s a collision and what’s not—in Swift, labels and return types are part of the name, so it’s perfectly OK to have `lower(into writer: Writer)`, `lower(b: Int)`, and `lower() -> Foo`; in other languages, it’s not. And it also doesn’t handle extensions and subclasses—what happens if you subclass or extend a generated type, and define a reserved name on it?

[...]

PR #196, opened 2020-08-03
Add destructors for objects

### LINKED ISSUE #8: Add destructors to Kotlin and Swift
The Python bindings already define `__del__` destructors for classes, but Kotlin and Swift don’t yet. We’ll want to make sure we wire these up and call them correctly, to avoid leaks and double-frees.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1565)

jhugman:
Fixes #8.

This PR adds `deinit` and `__del__` calls into a reserved `object_free` method. (aside: this should be one more example for the need for #10.)

For kotlin, the JVM has a long and storied history of destructor not being the right thing.

Our _best_ hope is for `java.lang.ref.Cleaner` to be implemented in Java 9 and then adopted by Android.

Our _second best_ hope is for explicit `destroy` method. Provided here is an abstract class which checks if the object has been destroyed before each call.

### DIFF
--- a/examples/sprites/tests/bindings/test_sprites.kts
+++ b/examples/sprites/tests/bindings/test_sprites.kts
@@ -11,3 +11,11 @@ assert( s.getPosition() == Point(1.0, 2.0) )
 
 s.moveBy(Vector(-4.0, 2.0))
 assert( s.getPosition() == Point(-3.0, 4.0) )
+
+s.destroy()
+try { 
+    s.moveBy(Vector(0.0, 0.0))
+    assert(false) { "Should not be able to call anything after `destroy`" }
+} catch(e: IllegalStateException) {
+    assert(true)
+}
--- a/uniffi/src/bindings/kotlin/templates/Helpers.kt
+++ b/uniffi/src/bindings/kotlin/templates/Helpers.kt
@@ -1,0 +1,30 @@
+// A handful of classes and functions to support the generated data structures.
+// This would be a good candidate for isolating in its own ffi-support lib.
+
+abstract class FFIObject(
+    private val handle: AtomicLong
+) {
+    open fun destroy() {
+        this.handle.set(0L)
+    }
+
+    internal inline fun <R> callWithHandle(block: (handle: Long) -> R) =
+        this.handle.get().let { handle -> 
+            if (handle != 0L) {
+                block(handle)
+            } else {
+                throw IllegalStateException("${this.javaClass.simpleName} object has already been destroyed")
+            }
+        }
+}
+
+inline fun <T : FFIObject, R> T.use(block: (T) -> R) =
+    try {
+        block(this)
+    } finally {
+        try {
+            this.destroy()
+        } catch (e: Throwable) {
+            // swallow
+        }
+    }
--- a/uniffi/src/bindings/kotlin/templates/ObjectTemplate.kt
+++ b/uniffi/src/bindings/kotlin/templates/ObjectTemplate.kt
@@ -30,19 +30,17 @@ class {{ obj.name()|class_name_kt }}(
     {%- match meth.return_type() -%}
 
     {%- when Some with (return_type) -%}
-    fun {{ meth.name()|fn_name_kt }}({% call kt::arg_list_decl(meth) %}): {{ return_type|type_kt }} {
-        val _retval = callWithHandle {
-            {# // Use it here, kotlin's anonymous argument for lambas/closures.
-            -#}
+    fun {{ meth.name()|fn_name_kt }}({% call kt::arg_list_decl(meth) %}): {{ return_type|type_kt }} =
+        callWithHandle {
             {% call kt::to_rs_call_with_prefix("it", meth) %} 
+        }.let {
+            {{ "it"|lift_kt(return_type) }}
         }
-        return {{ "_retval"|lift_kt(return_type) }}
-    }
     
     {%- when None -%}
     fun {{ meth.name()|fn_name_kt }}({% call kt::arg_list_decl(meth) %}) =
         callWithHandle {
-            {% call kt::to_rs_call_with_prefix("it", meth.ffi_func()) %} 
+            {% call kt::to_rs_call_with_prefix("it", meth) %} 
         }
     {% endmatch %}
     {% endfor %}
--- a/uniffi/src/bindings/kotlin/templates/ObjectTemplate.kt
+++ b/uniffi/src/bindings/kotlin/templates/ObjectTemplate.kt
@@ -1,25 +1,49 @@
+class {{ obj.name()|class_name_kt }}(
+    handle: Long
+) : FFIObject(AtomicLong(handle)) {
 
-class {{ obj.name()|class_name_kt }}(handle: Long) {
-    private var handle: AtomicLong = AtomicLong(handle)
     {%- for cons in obj.constructors() %}
     constructor({% call kt::arg_list_decl(cons) -%}) :
         this({% call kt::to_rs_call(cons) %})
     {%- endfor %}
 
-    // XXX TODO: destructors or equivalent.
+    /**
+     * Disconnect the object from the underlying Rust object.
+     * 
+     * It can be called more than once, but once called, interacting with the object 
+     * causes an `IllegalStateException`.
+     * 
+     * Clients **must** call this method once done with the object, or cause a memory leak.
+     */
+    override fun destroy() {
+        try {
+            callWithHandle {
+                super.destroy() // poison the handle so no-one else can use it before we tell rust.
+                _UniFFILib.INSTANCE.{{ obj.ffi_object_free().name() }}(it)
+            }
+        } catch (e: IllegalStateException) {
+            // The user called this more than once. Better than less than once.
+        }
+    }
 
     {% for meth in obj.methods() -%}
     {%- match meth.return_type() -%}
 
     {%- when Some with (return_type) -%}
     fun {{ meth.name()|fn_name_kt }}({% call kt::arg_list_decl(meth) %}): {{ return_type|type_kt }} {
-        val _retval = {% call kt::to_rs_call_with_prefix("this.handle.get()", meth) %}
+        val _retval = callWithHandle {
+            {# // Use it here, kotlin's anonymous argument for lambas/closures.
+            -#}
+            {% call kt::to_rs_call_with_prefix("it", meth) %} 
+        }
         return {{ "_retval"|lift_kt(return_type) }}
     }
     
     {%- when None -%}
     fun {{ meth.name()|fn_name_kt }}({% call kt::arg_list_decl(meth) %}) =
-        {% call kt::to_rs_call_with_prefix("this.handle.get()", meth) %}
+        callWithHandle {
+            {% call kt::to_rs_call_with_prefix("it", meth.ffi_func()) %} 
+        }
     {% endmatch %}
     {% endfor %}
 }
--- a/uniffi/src/bindings/python/templates/ObjectTemplate.py
+++ b/uniffi/src/bindings/python/templates/ObjectTemplate.py
@@ -3,10 +3,11 @@ class {{ obj.name()|class_name_py }}(object):
     {%- for cons in obj.constructors() %}
     def __init__(self, {% call py::arg_list_decl(cons.arguments()) -%}):
         {%- call py::coerce_args_extra_indent(cons.arguments()) %}
-        self._handle = {% call py::to_rs_call(cons) %}
+        self._handle = {% call py::to_rs_call(cons.ffi_func()) %}
     {%- endfor %}
 
-    # XXX TODO: destructors or equivalent.
+    def __del__(self):
+        _UniFFILib.{{ obj.ffi_object_free().name() }}(handle)
 
     {% for meth in obj.methods() -%}
     {%- match meth.return_type() -%}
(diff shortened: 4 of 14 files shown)


uniffi/src/interface/mod.rs
```diff
@@ -548,13 +564,21 @@ impl Object {
...
         for cons in self.constructors.iter_mut() {
             cons.derive_ffi_func(ci_prefix, &self.name)?
         }
         for meth in self.methods.iter_mut() {
             meth.derive_ffi_func(ci_prefix, &self.name)?
         }
+        let reserved_prefix = format!("_{}", ci_prefix);
+        self.object_free
+            .derive_ffi_func(&reserved_prefix, &self.name)?;
+
```
jhugman, line 581: Adding a `_` prefix makes it much harder to generate colliding method names and top level function names.

uniffi/src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -1,25 +1,34 @@
 
-class {{ obj.name()|class_name_kt }}(handle: Long) {
-    private var handle: AtomicLong = AtomicLong(handle)
-    {%- for cons in obj.constructors() %}
+class {{ obj.name()|class_name_kt }}(
+    internal val handle: Long
```
jhugman, line 3: `handle` is never going to change, so removed `AtomicLong`.

uniffi/src/templates/ObjectTemplate.rs
```diff
@@ -6,41 +6,47 @@
...
-// define_handle_map_deleter!(UNIFFI_HANDLE_MAP_{{ obj.name() }}, {{ obj.name() }}_free);
+    // define_handle_map_deleter!(UNIFFI_HANDLE_MAP_{{ obj.name() }}, {{ obj.name() }}_free);
+    {% let ffi_free = obj.reserved_object_free().ffi_func() -%}
+    #[no_mangle]
+    pub extern "C" fn {{ ffi_free.name() }}(
+        {%- call rs::arg_list_rs_decl(ffi_free.arguments()) %}) {
+        log::debug!("{{ ffi_free.name() }}");
+        let _retval = {{ handle_map }}.delete_u64(handle);
+    }
+
```
jhugman, line 22: This should be replaced with 

```rust
define_handle_map_deleter!({{ handle_map }}, {{ ffi_free.name() }})
```

once #31 lands and I can cargo cult the how that works better.

uniffi/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -0,0 +1,28 @@
+// A handful of classes and functions to support the generated data structures.
+// This would be a good candidate for isolating in its own ffi-support lib.
+
+abstract class FFIObject {
+    internal val isDestroyed = AtomicBoolean(false)
```
rfk, line 5: Something we've done in the hand-written bindings, was use `this.handle == 0` to indicate that the object is dead. Would it make sense to move `handle` up to a property of this parent class in order to do something like that?

uniffi/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    open fun destroy() {
+        isDestroyed.set(true)
+    }
+
+    internal inline fun <R> safelyCall(block: () -> R) =
```
rfk, line 11: Naming nit: things with "safe" in their name are a personal bugbear of mine because it's often not obvious what "safe" means and they regularly turn out to be less safe than advertised. It's probably not a big deal for an internal helper function like this, but still, is there a more descriptive name?

One possible suggestion: if we moved `handle` up to be a property of this parent class, you could make this something like `callWithHandle`. Not sure if that would make the call-side code in each method declaration messier though.

uniffi/src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -1,25 +1,34 @@
...
+
+    {% for cons in obj.constructors() %}
     constructor({% call kt::arg_list_decl(cons.arguments()) -%}) :
         this({% call kt::to_rs_call(cons) %})
     {%- endfor %}
 
-    // XXX TODO: destructors or equivalent.
+    override fun destroy() {
+        super.destroy()
+        {% call kt::to_rs_call_with_prefix("this.handle", obj.reserved_object_free()) %}
```
rfk, line 13: Are we are deliberately calling `super.destroy()` first so that we "poison" the Kotlin side of the object, as a guard against potential concurrency issues. I expect that the underlying handlemap will prevent concurrency issues in practice, but still, if the ordering here is important it may be worth a comment to that effect.

uniffi/src/bindings/python/templates/ObjectTemplate.py
```diff
@@ -6,7 +6,8 @@ def __init__(self, {% call py::arg_list_decl(cons.arguments()) -%}):
         self._handle = {% call py::to_rs_call(cons) %}
     {%- endfor %}
 
-    # XXX TODO: destructors or equivalent.
+    def __del__(self):
```
rfk, line 9: Should this similarly poison the object on the python side, like we do in Kotlin? IIRC it is possible that someone might subclass one of these and extend `__del__` in a way that resurrects the the python-side object after the rust-side destructor has been called. (They really *shouldn't* do that, but it's *possible*).

uniffi/src/interface/mod.rs
```diff
@@ -166,7 +166,7 @@ impl<'ci> ComponentInterface {
 
     pub fn ffi_string_free(&self) -> FFIFunction {
         FFIFunction {
-            name: format!("{}_string_free", self.namespace()),
+            name: format!("_{}_string_free", self.namespace()),
```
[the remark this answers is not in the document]
rfk: I'm told that names with a leading underscore are reserved for use by the compiler in C code. I don't have any particular case in mind where this would cause problems for us, but one alternative might be to put the leading underscore after the namespace, like `{}__bytebuffer_free`.

uniffi/src/interface/mod.rs
```diff
@@ -533,9 +534,24 @@ pub struct Object {
     name: String,
     constructors: Vec<Constructor>,
     methods: Vec<Method>,
+    object_free: Method,
```
rfk, line 537: Does this need to be an API-level `Method`? Naively, it feels like it would suffice to generate just the `FFIFunction` and rely on the automatically-generated foreign-language code for the API-level definition.

In practice that would look very similar to how `bytebuffer_free` and friends work on the `ComponentInterface`:

* The `Object::reserved_object_free` method creates a new `FFIFunction` every time it's called.
* The `Object::iter_ffi_function_definitions` method calls `Object::reserved_object_free` and appends the result to the vec of `FFIFunction` definitions.
* For foreign-language templates directly emit an FFI-level call to the function returned by `Object::reserved_object_free`.

I *think* that would be all you need, and would simplify some of the changes here.

uniffi/src/interface/mod.rs
```diff
@@ -533,9 +534,24 @@ pub struct Object {
...
+            constructors: Default::default(),
+            methods: Default::default(),
+            object_free: Method {
+                name: "object_free".to_string(),
+                arguments: Default::default(),
+                return_type: None,
+                ffi_func: Default::default(),
+            },
+        }
+    }
```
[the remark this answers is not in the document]
rfk: I think this should be `new`, and it's fine for `new` to have arguments if they're obviously needed for making an instance.

However, I don't think you'll need this at all if you can avoid creating an API-level `Method` for this, and use only an `FFIFunction` in the style of the bytebuffer helpers.

uniffi/src/bindings/kotlin/templates/ObjectTemplate.kt
tarikeshaq: This makes sense to me, but I do wonder why we tend to use AtomicLong for handles in Application Services 🤔

uniffi/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -0,0 +1,28 @@
+// A handful of classes and functions to support the generated data structures.
+// This would be a good candidate for isolating in its own ffi-support lib.
+
+abstract class FFIObject {
+    internal val isDestroyed = AtomicBoolean(false)
+
+    open fun destroy() {
+        isDestroyed.set(true)
```
tarikeshaq, line 8: I'm wondering if we should throw an error here if the object has already been destroyed. Can't think of a concrete reason why, but one thought I'm having is if someone calls the destroy multiple times, I'm guessing we would give them an error, but is it better to error here? And what kind of error would they get if the call goes through the FFI?

I guess we could do what the other bindings do, would Swift/Python complain if you call the destructor twice? Or would they just pass it into the FFI regardless?

examples/sprites/tests/bindings/test_sprites.kts
```diff
@@ -11,3 +11,11 @@ assert( s.getPosition() == Point(1.0, 2.0) )
 
 s.moveBy(Vector(-4.0, 2.0))
 assert( s.getPosition() == Point(-3.0, 4.0) )
+
+s.destroy()
```
tarikeshaq, line 15: I'm wondering where we'd put documentation for this, so consumers know they need to call the `destroy` function to do the deallocation. Thoughts?

uniffi/src/templates/ObjectTemplate.rs
```diff
@@ -6,41 +6,47 @@
 // If the caller's implementation of the struct does not match with the methods or types specified
 // in the IDL, then the rust compiler will complain with a (hopefully at least somewhat helpful!)
 // error message when processing this generated code.
-
+{% let handle_map = format!("UNIFFI_HANDLE_MAP_{}", obj.name().to_uppercase()) %}
```
tarikeshaq, line 9: ❤️

uniffi/src/interface/mod.rs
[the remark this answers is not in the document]
jhugman: I added `new`. 

However, avoiding adding an `reserved` `Method`  meant that I was passing in the namespace into the `Object` now.

uniffi/src/interface/mod.rs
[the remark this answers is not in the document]
jhugman: The extra character needs to be before the namespace, otherwise collisions are possible.

Assuming adding a namespace-trailing underscore, this IDL has two collisions.

```idl
namespace {
   void _string_free()
   void Sprite__object_free()
}

class Sprite {

}
```

I'll change this to something more verbose/less exotic.

uniffi/src/bindings/kotlin/templates/Helpers.kt
jhugman: I'm not overly happy with this object being a class— having it as a `AtomicBoolean` would allow this to become an interface again.

uniffi/src/bindings/kotlin/templates/ObjectTemplate.kt
jhugman: Now changed back to `AtomicLong`, so we can zero out the handle when we're destroying it.

uniffi/src/interface/mod.rs
jhugman: I've moved this as directed here; it did mean touching the foreign-language macro templates for all functions and methods, and adding a `method_arguments()` method to `FFIFunction` which elides the first argument.

uniffi/src/bindings/kotlin/templates/Helpers.kt
jhugman: As @rfk says, the `ConcurrentHandleMap` should handle that for us. 

I'm torn here: should we be adding `isDestroyed` so we can encourage/enable defensive programming? Or should we make it idempotent? Or should we just go with what we've got, and review ergonomics once we've started using it?

examples/sprites/tests/bindings/test_sprites.kts
jhugman: Oh that's a really interesting question. 

For now, I've documented in the generated section, however, we do need a better solution for this.

uniffi/src/bindings/python/templates/ObjectTemplate.py
jhugman: Ug. I think I'm going to file a python-only issue to investigate stopping resurrecting python side objects.

uniffi/src/interface/mod.rs
```diff
@@ -416,6 +416,13 @@ impl FFIFunction {
     pub fn return_type(&self) -> Option<&TypeReference> {
         self.return_type.as_ref()
     }
+    pub fn method_arguments(&self) -> Vec<&Argument> {
+        let args = self.arguments.get(1..self.arguments.len());
+        match args {
+            None => Vec::new(),
+            Some(args) => args.iter().collect(),
+        }
+    }
```
linacambridge, line 425:
```suggestion
    pub fn method_arguments(&self) -> &[Argument] {
        self.arguments.get(1..self.arguments.len()).unwrap_or_default()
    }
```

Super small nit, but maybe we can avoid the copy and `match` here?

uniffi/src/templates/ObjectTemplate.rs
```diff
@@ -6,41 +6,47 @@
...
-// These will need to be defined as another FFI function on the Object struct, and included automatically
-// in the set of all FFI functions for use by the bindings.
-// define_handle_map_deleter!(UNIFFI_HANDLE_MAP_{{ obj.name() }}, {{ obj.name() }}_free);
+    {% let ffi_free = obj.ffi_object_free() -%}
+    // define_handle_map_deleter!({{ handle_map }}, {{ ffi_free.name() }});
+    #[no_mangle]
+    pub extern "C" fn {{ ffi_free.name() }}(
+        {%- call rs::arg_list_rs_decl(ffi_free.arguments()) %}) {
+        log::debug!("{{ ffi_free.name() }}");
+        let _retval = {{ handle_map }}.delete_u64(handle);
```
linacambridge, line 20: Nit: `let _ = ...` or `drop(...)`?

uniffi/src/bindings/kotlin/templates/Helpers.kt
linacambridge: I like how you've done it in the PR, making `destroy` idempotent, and throwing if you try to call any methods after destroying the class. It feels like that'll be easier for a-c or whoever to wire it up to `Cleaner` if they want, without having to do additional state tracking.

uniffi/src/interface/mod.rs
```diff
@@ -548,6 +572,20 @@ impl Object {
         self.methods.iter().collect()
     }
 
+    pub fn ffi_object_free(&self) -> FFIFunction {
+        FFIFunction {
+            name: format!("ffi_{}_{}_object_free", self.namespace, self.name),
```
rfk, line 577: Ah yes, I see the trick here with needing more of the global context for the namespace. This seems OK I think, thanks for digging in.

uniffi/src/interface/mod.rs
rfk: Is it now safe to remove this `object_free` property on the struct?

uniffi/src/interface/mod.rs
jhugman: Oh that's soo much nicer! Thank you!

jhugman:
#203 and #31 takes the templates in a different direction. :(

Issue #27, opened 2020-07-22
Allow manual generation of scaffolding and bindings from IDL file


rfk:
The way uniffi is currently set up, optimizes for a very specific use-case: you're building a crate and you're going to publish it as a rust component, so you want uniffi integrated at the top level of your build process (`cargo build` produces the `.so`, `cargo run -- generate` produces the foreign-language bindings, etc).

From some thoughtful feedback from @tarikeshaq over in https://github.com/rfk/uniffi-rs/issues/24#issuecomment-661608571, it's clear that the first set of potential real-life consumers won't be shaped like that. They're more likely to want to tentatively integrate uniffi into the build process of an existing project, and it's not obvious how to do that based on what we have right now.

A strawfox proposal, broadly inspired by how we use protobufs in app-services:

* Invest in making a `uniffi-bindgen` tool that can be used in a standalone fashion, for integration with other build setups. We actually already have the beginnings of this, but it's hidden away in implementation details.
* Have a command like `uniffi-bindgen scaffolding path/to/api.idl` that can read an IDL file and spit out the rust code that flattens it into a C-style FFI. Rather than writing a bunch of `pub extern "C"` functions by hand, consumers could use this tool to generate them automatically from the IDL file and then `include!()` them into their rust code (or perhaps import them as a sub-module?).
* Have a command like `uniffi-bindgen generate --language=<language> path/to/api.idl` that can read an IDL file and spit out the appropriate foreign-language bindings to consume the C-style FFI from above. Rather than writing Kotlin, Swift etc code by hand, consumers could use this tool to generate it automatically from the IDL file, and then incorporate the generated files into their package build.

That's going to feel a bit clunky as a dev experience, in much the same way that working with protobuf files feels a bit clunky for us in app-services (like the CI job whose only purpose is to check that you didn't forget to generate the bindings). But from recent conversations, I think it could set us up to deliver early value quicker, without having to solve a bunch of more general tooling challenges up-front.

Most of the codegen pieces for this already exist, but need to be exposed for easy use in this new way from the command-line.

Thoughts? /cc @eoger @jhugman @linacambridge @tarikeshaq

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1559)

tarikeshaq:
I think this would be helpful! I also imagine we'd have a way of setting/detecting the native library that the generated code would load as a part of this effort.

[...]

PR #205, opened 2020-08-05
Refactor uniffi structure

eoger:
Fixes https://github.com/mozilla/uniffi-rs/issues/27.
So this is something I wanted to do for a while, I believe it reduces the mental load trying to understand how the pieces fit together (especially the testing parts!).

I've split up the project in a 3 crates:
- `uniffi` is the Rust runtime dependency (think `ViaFFI` traits and friends`). Target crates depend only on this.
- `uniffi_bindgen` contains the Rust scaffolding and Kotlin/Swift bindings generators. It is not meant to be depended upon, but installed through `uniffi-bindgen` and run at build time.
- `uniffi_build` is a convenience crate that will call `uniffi-bindgen scaffolding` for you.

I'll go and add review comments to make it easier to review this giant patch, initially I made multiple commits but before I knew it it proved too hard to amend the previous commits when making touchups.


examples/arithmetic/Cargo.toml
```diff
@@ -10,11 +10,8 @@ crate-type = ["cdylib"]
...
-anyhow = "1.0"
-log = "0.4"
-ffi-support = "0.4"
-uniffi = {path = "../../uniffi"}
 uniffi_macros = {path = "../../uniffi_macros"}
+uniffi = {path = "../../uniffi"}
 
 [build-dependencies]
-uniffi = {path = "../../uniffi"}
+uniffi_build = {path = "../../uniffi_build"}
```
eoger, line 17: See how we depend on 1 crate for the bytebuffer traits/methods + tests helpers, and 1 for build time.

examples/arithmetic/src/main.rs
```diff
@@ -1,7 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-fn main() {
-    uniffi::run_bindgen_for_component("arithmetic").unwrap();
```
eoger, line 6: The test helpers already generate the bindings for you.

examples/arithmetic/tests/test_generated_bindings.rs
```diff
@@ -1,4 +1,5 @@
 uniffi_macros::build_foreign_language_testcases!(
+    "src/arithmetic.idl",
```
eoger, line 2: `build_foreign_language_testcases` now takes an `.idl` file as first argument.

uniffi/Cargo.toml
```diff
@@ -5,20 +5,13 @@ authors = ["Firefox Sync Team <sync-team@mozilla.com>"]
...
-anyhow = "1.0"
-askama = "0.9"
-heck = "0.3"
-clap = "2.33"
-object = "0.20"
-serde = "1.0"
-bincode = "1.3"
-cargo_metadata = "0.10.0"
 lazy_static = "1.4"
+log = "0.4"
```
eoger, line 15: These dependencies are re-exported for convenience.

uniffi/src/testing.rs
```diff
@@ -117,25 +122,82 @@ pub fn ensure_compiled_cdylib(pkg_dir: &str) -> Result<String> {
...
+    if generated_bindings.contains(&(idl_file.clone(), out_dir.clone(), language_ext.to_owned())) {
         return Ok(());
     }
     // Nope, looks like we'll have to compile it afresh.
-    let ci = bindings::get_component_interface_from_cdylib(&cdylib_file)?;
-    bindings::write_bindings(&ci, &out_dir, language)?;
-    bindings::compile_bindings(&ci, &out_dir, language)?;
-    generated_bindings.insert((cdylib_file, out_dir, language));
+
+    let status = Command::new("uniffi-bindgen")
```
eoger, line 144: We don't depend on `uniffi-bindgen`, but that's OK we call it on the command line :)

uniffi/src/testing/kotlin.rs
```diff
@@ -0,0 +1,71 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
eoger, line 1: This + the Swift/Python ones are copy-pasted from `uniffi` old code.

uniffi_bindgen/src/bindings/kotlin/mod.rs
```diff
@@ -0,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
eoger, line 1: Nothing new, `git` is confused.

uniffi_bindgen/src/bindings/mod.rs
```diff
@@ -8,7 +8,6 @@
 //! along with some helpers for executing foreign language scripts or tests.
```
eoger, line 8: Deleted methods are in `uniffi` test support code now.

uniffi_bindgen/src/bindings/swift/mod.rs
```diff
@@ -0,0 +1,73 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
eoger, line 1: Nothing new, `git` is confused.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -100,23 +100,6 @@ impl<'ci> ComponentInterface {
         Ok(ci)
     }
 
-    pub fn from_bincode(data: &[u8]) -> Result<Self> {
```
eoger, line 103: The `.idl` file is the source of truth now.

uniffi_bindgen/src/lib.rs
```diff
@@ -0,0 +1,173 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
eoger, line 1: I'd encourage the reviewer to read the updated module doc :)

uniffi_bindgen/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+    target_languages: Vec<&str>,
+    out_dir_override: Option<&OsStr>,
+) -> Result<()> {
+    let component = parse_idl(idl_file)?;
+    let out_dir = get_out_dir(idl_file, out_dir_override)?;
+    for language in target_languages {
+        bindings::write_bindings(&component, &out_dir, language.try_into()?)?;
+    }
+    Ok(())
+}
```
eoger, line 148: The 2 important functions we expose through the CLI (see `main.rs`).

uniffi_build/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use anyhow::{bail, Result};
+use std::{env, process::Command};
+
+pub fn generate_scaffolding(idl_file: &str) -> Result<()> {
```
eoger, line 8: As I said this is just a helper crate.

uniffi_build/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+    println!("cargo:rerun-if-changed={}", idl_file);
+    // Why don't we just depend on uniffi-bindgen and call the public functions?
+    // Calling the command line helps making sure that the generated swift/Kotlin/whatever
+    // bindings were generated with the same version of uniffi as the Rust scaffolding code.
+    let out_dir = env::var("OUT_DIR").map_err(|_| anyhow::anyhow!("$OUT_DIR missing?!"))?;
+    let status = Command::new("uniffi-bindgen")
```
tarikeshaq, line 14: I'm a little conflicted here, I like `uniffi-bindgen` being a separate CLI crate, but the way its setup requires us to run the scaffolding through a command when in reality this mostly will be run on build scripts (Unless we think the consumer should to be able to generate their scaffolding at different times, then I concede)

If we think that it's okay for the consumer to **always** use `uniffi_build` to generate scaffolding (and never `uniffi_bindgen`. Then maybe all the scaffolding logic can be in this crate?

What do you think? I'm genuinely unsure, so okay with whatever path you choose!

If we choose to stay with the way that it is, it would be nice if we can print a nice error message or something if the user doesn't have `uniffi-bindgen` installed

rfk:
> Note that the CI failures is a bootstrapping problem: uniffi-bindgen is not available on crates.io, so it can't execute the tests

Are we feeling ready to publish this thing to crate.io..? :-)

examples/arithmetic/tests/test_generated_bindings.rs
rfk: I agree, it would be nicer to separate this argument more directly. Can be for a followup if it's a PITA though.

uniffi/Cargo.toml
```diff
@@ -5,20 +5,13 @@ authors = ["Firefox Sync Team <sync-team@mozilla.com>"]
 license = "MPL-2.0"
 edition = "2018"
 
-[[bin]]
-name = "uniffi-bindgen"
-path = "src/main.rs"
-
+# The dependency versions here have to agree with the corresponding ones in
+# uniffi_bindgen.
```
rfk, line 9: What will happen if they don't?

uniffi/Cargo.toml
rfk: Yeah, +1 I like this a lot.

uniffi/src/lib.rs
```diff
@@ -2,352 +2,293 @@
...
-    let mut contents = String::new();
-    let mut f = File::open(file_name)?;
-    f.read_to_string(&mut contents)?;
-    Ok(contents)
+use anyhow::{bail, Result};
+use bytes::buf::{Buf, BufMut};
+use ffi_support::ByteBuffer;
+use std::convert::TryFrom;
+
+// I'd be nice this module was behind a cfg(test) guard, but it doesn't
```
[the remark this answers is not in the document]
rfk: Lol I like the subtle blackmailing going on in Ed's version though, "I'd be nice if only you would..." 😆

uniffi/src/lib.rs
```diff
@@ -2,352 +2,293 @@
...
+// work between crates so let's hope LLVM tree-shaking works well.
+pub mod testing;
```
rfk, line 12: I guess we could consider moving this into a fourth "dev-dependencies" crate, but I think this will be fine as is, and I have confidence in the tree-shaking stuff 🤞

uniffi/src/lib.rs
```diff
@@ -2,352 +2,293 @@
...
+
+// Re-export the libs that we use in the generated code,
+// so the consumer doesn't have to depend on them directly.
+pub use anyhow;
+pub use bytes;
+pub use ffi_support;
+pub use lazy_static;
+pub use log;
```
rfk, line 20: This is very good, thank you for figuring this out!

Cargo.toml
```diff
@@ -1,7 +1,9 @@
 [workspace]
 members = [
-  "uniffi",
+  "uniffi_bindgen",
+  "uniffi_build",
   "uniffi_macros",
+  "uniffi",
```
rfk, line 6: At the risk of veering into bikeshed territory, I would be tempted to name this "uniffi_runtime" or similar to give a more direct indication that it contains the run-time support infra. We don't need to change it though, just sharing my reaction to the names here.

uniffi/src/testing.rs
rfk: I'm a little hesitant about this, since we don't have any guarantees that the `uniffi-bindgen` found on the system is the same version as used by the crate under test here. It seems like a footgun and potentially confusing experience for developers.

We have the `pkg_dir` here in order to compile the crate under test, I wonder if we should try to do something like:

* Inspect the cargo metadata to find the version of `uniffi` on which the crate depends.
* Shell out to cargo to build the matching version of `uniffi_bindgen` into our target dir, caching it similar to how we cache the compilation of the crate itself.
* Use that built version of the `uniffi-bindgen` command the run the tests.

I think this would resolve the pending CI failures, and also produce a more consistent/reliable developer experience.

(A simpler version of this might be to just build and use the `uniffi-bindgen` in the current workspace, which would suffice to get our CI passing).

uniffi/src/testing.rs
```diff
@@ -117,25 +122,82 @@ pub fn ensure_compiled_cdylib(pkg_dir: &str) -> Result<String> {
...
+        bail!("Error while generating bindings: {}",);
+    }
+
+    // XXX: technically, namespace != idl file name, but we don't
+    // want to take a dependency on `ComponentInterface` here.
+    let namespace = idl_file
+        .file_stem()
+        .ok_or_else(|| anyhow!("not a file"))?
+        .to_str()
+        .ok_or_else(|| anyhow!("not a string"))?;
```
rfk, line 164: Two alternatives to relying on file-naming convention here:

* Do a half-hearted manual parse of the IDL file, just looking for a regex match on `namespace (\w+)`.
* Add another command to `uniffi-bindgen` to spit out component metadata from the IDL file.

rfk: Alternately, maybe it would make sense to push some or all of the "compile and run foreign-language script against these bindings" logic out into a `uniffi-bindgen` subcommand as well? We already need `uniffi-bindgen` to exist to run the tests.

uniffi/src/testing/python.rs
```diff
@@ -0,0 +1,27 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use anyhow::{bail, Context, Result};
+use std::{env, ffi::OsString, path::Path, process::Command};
+
+/// Execute the specifed python script, with environment based on the generated
+/// artifacts in the given output directory.
+pub fn run_script(out_dir: &Path, script_file: &Path) -> Result<()> {
+    let mut pythonpath = env::var_os("PYTHONPATH").unwrap_or_else(|| OsString::from(""));
+    // This lets java find the compiled library for the rust component.
```
rfk, line 12: This is probably my own typo, but, nit: "java" -> "python"

uniffi_bindgen/src/interface/mod.rs
rfk: 😢 I really liked this neat hack, but yeah, using the `.idl` as the source of truth makes more sense for our setup, at least in the short term.

uniffi_bindgen/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+//! The `uniffi-bindgen` utily provides a command-line tool that can produce code to
+//! consume the Rust library in any of several supported languages.
+//! It is done by calling (in python for example):
+//!
+//! ```text
+//! uniffi-bindgen --language python ./src/example.idl
+//! ```
+//!
+//! This will produce a file `example.py` in the same directory as the .idl file, containing python bindings
+//! to load and use the compiled rust code via its C-compatible FFI.
```
rfk, line 89: If the python bindings are tier-two support level, we should probably not use python as the example language here.

uniffi_bindgen/src/main.rs
```diff
@@ -0,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use anyhow::{bail, Result};
+
+const POSSIBLE_LANGUAGES: &[&str] = &["kotlin", "python", "swift"];
+
+fn main() -> Result<()> {
+    let matches = clap::App::new("uniffi")
```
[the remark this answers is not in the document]
rfk: Yes, please let's.

uniffi_bindgen/src/templates/RustBuffer.rs
```diff
@@ -0,0 +1,8 @@
+// Everybody gets basic bytebuffer allocation and freeing, since it's needed
+// for passing complex types over the FFI.
+#[no_mangle]
+pub extern "C" fn {{ ci.ffi_bytebuffer_alloc().name() }}(size: u32) -> uniffi::ffi_support::ByteBuffer {
```
rfk, line 4: A minor suggestion seeing the generated code, I wonder if it would be worth namespacing this re-exported modules under something like `uniffi::deps::ffi_support` or similar, to make it a little more obvious what's going on when you see this in the generated code.

uniffi_build/src/lib.rs
rfk: > (Unless we think the consumer should to be able to generate their scaffolding at different times, then I concede)

I think this is at least partly the intention, for consumers who want to checkin the generated scaffolding rather than generating it afresh each time (like we currently do for protobuf generated rust code in appservices).

uniffi_macros/src/lib.rs
```diff
@@ -22,12 +22,15 @@ use syn::{punctuated::Punctuated, LitStr, Token};
...
-/// To use it, invoke the macro with one or more file paths relative to the crate root
-/// directory. It will produce one `#[test]` function per file, in a manner designed to
+/// To use it, invoke the macro with the idl file as the first argument, then
+/// one or more file paths relative to the crate root directory.
+/// It will produce one `#[test]` function per file, in a manner designed to
 /// play nicely with `cargo test` and its test filtering options.
 #[proc_macro]
 pub fn build_foreign_language_testcases(paths: proc_macro::TokenStream) -> proc_macro::TokenStream {
-    let paths = syn::parse_macro_input!(paths as FilePaths).0;
+    let mut paths = syn::parse_macro_input!(paths as FilePaths).0;
```
rfk, line 31: I think we should refactor this to be called like `build_foreign_language_testcases("file.idl", ["array", "of", "test", "files"])`, and I think it would be a great opportunity for someone who hasn't done it before, to try messing around with `syn`/`quote` and the infra of doing proc macros. It can be a follow-up though.

rfk:
Also, the backtrace from failing CI is a good indicator that we should take the time to provide a clearer error message when `uniffi-bindgen` cannot be found, it's currently pretty opaque:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: No such file or directory (os error 2)', examples/rondpoint/build.rs:6:5
```

uniffi/src/lib.rs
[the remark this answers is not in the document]
tarikeshaq: I laughed way too loud at this 🤣

uniffi_build/src/lib.rs
eoger: Yes that's the intent, my concern is that just like `prost`, compiling this crate at build time will make the build take way longer. This way you have the flexibility of doing that, or like @rfk said you can generate the scaffolding out-of-build.

.circleci/config.yml
```diff
@@ -2,6 +2,12 @@
 
 version: 2.1
 
+commands:
```
eoger, line 5: This should allow CI to run properly with the current bindgen version.

uniffi/src/testing.rs
```diff
@@ -37,15 +35,22 @@ lazy_static! {
...
         .parent()
-        .ok_or_else(|| anyhow::anyhow!("Generated cdylib has no parent directory"))?;
-    let lang = test_file.extension().unwrap_or_default();
-    ensure_generated_bindings(cdylib_file.as_str(), out_dir, lang)?;
-    bindings::run_script(Some(out_dir), Some(test_file), lang.try_into()?)
+        .ok_or_else(|| anyhow::anyhow!("Generated cdylib has no parent directory"))?
+        .to_str()
+        .unwrap();
+
+    let _lock = UNIFFI_BINDGEN.lock();
```
eoger, line 46: I put all the `compile` and `run_script` machinery back in `uniffi_ bindgen`, leaving a single call to `uniffi-bindgen`. As a bonus we can now run tests outside of Rust.

uniffi_bindgen/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+    if format_code {
+        Command::new("rustfmt").arg(&out_dir).status()?;
+    }
+    Ok(())
+}
+
+// If the crate for which we are generating bindings for depends on
+// a `uniffi` runtime version that doesn't agree with our own version,
+// the developer of that said crate will be in a world of pain.
+fn ensure_versions_compatibility(
```
eoger, line 146: Added this to ensure the crates compatibility between the bindgen and the runtime crates. This basically compares the major versions and fails if they are different.

uniffi_bindgen/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+    let out_dir = get_out_dir(&idl_file, out_dir_override)?;
+    for language in target_languages {
+        bindings::write_bindings(&component, &out_dir, language.try_into()?)?;
+    }
+    Ok(())
+}
+
+// Run tests against the foreign language bindings (generated and compiled at the same time).
+// Note that the cdylib we're testing against must be built already.
+pub fn run_tests(cdylib_dir: &OsStr, idl_file: &OsStr, test_scripts: Vec<&str>) -> Result<()> {
```
eoger, line 214: The new test machinery living in `uniffi_bindgen` is here!

uniffi_bindgen/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+                    .help("Path to crate's Cargo.toml. If not provided the IDL file is assumed to be under src/")
+                )
+                .arg(
+                    clap::Arg::with_name("no_format")
+                        .long("--no-format")
+                        .help("Do not format the generated code with rustfmt (useful for maintainers)"),
+                )
+                .arg(clap::Arg::with_name("idl_file").required(true)),
+        )
+        .subcommand(
```
eoger, line 58: And here is the subcommand to use it.

uniffi_macros/src/lib.rs
```diff
@@ -61,15 +63,24 @@ pub fn build_foreign_language_testcases(paths: proc_macro::TokenStream) -> proc_
...
+        let array_contents;
+        bracketed!(array_contents in input);
+        let test_scripts = Punctuated::<LitStr, Token![,]>::parse_terminated(&array_contents)?
+            .iter()
+            .map(|s| s.value())
+            .collect();
+        Ok(FilePaths {
+            idl_file: idl_file.value(),
+            test_scripts,
+        })
```
eoger, line 84: There might be simpler way of doing this 😅

Cargo.toml
eoger: I actually named it `uniffi_runtime` first, then I figured it would be sad if we didn't have any `uniffi` crate at all 😝 @tarikeshaq any opinion here?

rfk: Haha, yeah, needing `uniffi` crate of *some* sort makes sense! I guess if/when we eventually move to a macro-based approach we'll also want that in a top-level `uniffi` crate if possible, so I'm happy to keep it this like.

uniffi_bindgen/src/bindings/mod.rs
```diff
@@ -97,42 +96,17 @@ where
...
-) -> Result<()>
+pub fn run_script<P1, P2>(out_dir: P1, script_file: P2, language: TargetLanguage) -> Result<()>
 where
     P1: AsRef<Path>,
     P2: AsRef<Path>,
 {
-    let out_dir: Option<&Path> = out_dir.as_ref().map(|v| v.as_ref());
-    let script_file: Option<&Path> = script_file.as_ref().map(|v| v.as_ref());
+    let out_dir = out_dir.as_ref();
+    let script_file = script_file.as_ref();
```
rfk, line 105: These are a lot simpler without all the `Option` stuff, I like it. (I'll probably find myself missing the ability to launch into a shell for a bit of ad-hoc debugging, but we can bring that back later if we need to).

uniffi_bindgen/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+        .to_os_string();
+    filename.push(".uniffi.rs");
+    let mut out_dir = get_out_dir(&idl_file, out_dir_override)?;
+    out_dir.push(filename);
+    let mut f =
+        File::create(&out_dir).map_err(|e| anyhow!("Failed to create output file: {:?}", e))?;
+    write!(f, "{}", RustScaffolding::new(&component))
+        .map_err(|e| anyhow!("Failed to write output file: {:?}", e))?;
+    if format_code {
+        Command::new("rustfmt").arg(&out_dir).status()?;
```
rfk, line 138: nice :-)

uniffi_bindgen/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+        .manifest_path(manifest_path)
+        .exec()
+        .map_err(|e| anyhow!("Failed to run cargo metadata: {:?}", e))?;
+    let uniffi_runtime_deps: Vec<cargo_metadata::Package> = metadata
+        .packages
+        .into_iter()
+        .filter(|p| p.name == "uniffi")
+        .collect();
+    if uniffi_runtime_deps.is_empty() {
+        bail!("It looks like the crate doesn't depend the `uniffi` runtime. Please add `uniffi` as a dependency.");
```
rfk, line 181: nit: "depend the" -> "depend on the"

uniffi_bindgen/src/lib.rs
rfk: Great, thank you!

One nit: I think things are changing fast enough (and we're pre-1.0-enough) that we should just do a full version number check here rather than just the major version, at least until we start being deliberate about releasing versions with semver. Put another way: the commit that prepares our 1.0 release should be the one that makes this do a major-version-only comparison.

uniffi_bindgen/src/lib.rs
rfk: From an API-consumer point of view, I'm lightly surprised to see this take the directory of the cdylib rather than the path to the cdylib itself. Not a big deal, just noting my surprise.

uniffi_bindgen/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+                bail!("Could not determine the crate path from the .idl file");
+            }
+            let cargo_toml = src_dir
+                .parent()
+                .ok_or_else(|| anyhow!("no parent!"))?
+                .join("Cargo.toml");
+            if !cargo_toml.exists() {
+                bail!("Could not find Cargo.toml file")
+            }
+            cargo_toml
```
rfk, line 168: If you didn't want to assume a particular location of the IDL file, I *think* you could `cd` into its parent directory and run `cargo metadata` from there. From local testing it would give you the metadata for the containing workspace as long as you are somewhere in a subdirectory of it, which IIUC is all you need here.

uniffi_macros/src/lib.rs
rfk: We might be able to come up with something more declarative as we learn more, but this LGTM!

Issue #208, opened 2020-08-07
Make `cargo test` in this repo Just Work (tm)


rfk:
As a side-effect of https://github.com/mozilla/uniffi-rs/pull/205, checking out this repo and running `cargo test` no longer works. It instead fails with:

```
It looks like uniffi-bindgen is not installed. You can do so by running `cargo install uniffi-bindgen`
```

I understand how to do so, this is probably the right thing for external crates depending on uniffi, but it's a pretty harsh developer experience for folks contributing to this repo itself.

Ideally, running `cargo test` in this repo would automatically use the `uniffi-bindgen` built from the checkout it is being run in. Not really sure how to make that work just yet, though...

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1632)

rfk:
What I've done so far is to symlink `./target/debug/` into my `$PATH` and then `cargo build -p uniffi_bindgen`. However, I find that each time I make a change to e.g. support some new bit of syntax in the IDL, I need to remember to run `cargo build` again before running `cargo test`. I fear that friction is going to add up pretty quickly over time, especially while we're likely to be adding many new things to uniffi in quick succession.

PR #209, opened 2020-08-07
Use our own bindgen when building our crates, rather than system bindgen

rfk:
When working on uniffi itself, it's very inconvenient to have to rely
on an externally-installed `uniffi-bindgen` command in order to run the
tests for the in-tree example crates. You basically need to remember to
manually `cargo build -p uniffi_bindgen` anytime you make a change to the
bindings genertor, or the tests won't test what you think they should.

This commit adds a feature-flag to the `uniffi` and `uniffi_build` crates
so that they will use the builtin `unifi_bindgen` module from this workspace
rather than whatever one they find on the system. I think this gives a much
better development experience when working on this repo itself, in particular
by allow `cargo test` to Just Work (tm) and always reflect your latest changes.

On the other hand, it means the examples in this repo are doing a thing that
we might not expect actual consuming crates to do.

I'm not thrilled about this solution, but I think it's a decent tradeoff on balance.

(I spent a few hours working on something more elaborate, that would actually
build the `uniffi-bindgen` command in this workspace when needed, but I got
caught up by cargo's build locking which prevented me from doing the required
`cargo build -p uniffi_bindgen` on-demand).

Fixes https://github.com/mozilla/uniffi-rs/issues/208.

### DIFF
--- a/examples/arithmetic/Cargo.toml
+++ b/examples/arithmetic/Cargo.toml
@@ -11,7 +11,7 @@ name = "uniffi_arithmetic"
 
 [dependencies]
 uniffi_macros = {path = "../../uniffi_macros"}
-uniffi = {path = "../../uniffi"}
+uniffi = {path = "../../uniffi", features=["builtin-bindgen"]}
 
 [build-dependencies]
-uniffi_build = {path = "../../uniffi_build"}
+uniffi_build = {path = "../../uniffi_build", features=["builtin-bindgen"]}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,12 +2,6 @@
 
 version: 2.1
 
-commands:
-  make_install_uniffi_bindgen:
-    steps:
-      - run:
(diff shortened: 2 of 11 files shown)


examples/arithmetic/Cargo.toml
```diff
@@ -11,7 +11,7 @@ name = "uniffi_arithmetic"
 
 [dependencies]
 uniffi_macros = {path = "../../uniffi_macros"}
-uniffi = {path = "../../uniffi"}
+uniffi = {path = "../../uniffi", features=["builtin-bindgen"]}
```
eoger, line 14: I wonder if we should add a comment there (or at the feature declaration in `uniffi` and `uniffi_build`) to say that you're not supposed to use this?

rfk: Yes, good idea. I tried to explain it a bit in the code where it's used, but consumers are not likely to look there.

Issue #194, opened 2020-08-03
Passing via ByteBuffer leaks memory


rfk:
As noted in @tarikeshaq's review of https://github.com/mozilla/uniffi-rs/pull/192.

When passing rich structures such as `Record` or `Sequence` *in* to Rust from the foreign language bindings, AFAICT the procedure goes like this:

* Foreign language allocates a `RustBuffer` into which to serialize the structure.
* Foreign language copies data into the `RustBuffer`.
* Foreign language passes the `RustBuffer` to Rust as an argument over the FFI.
* Rust copies data out of the `RustBuffer` into live Rust objects
* Rust discards the `RustBuffer` without freeing the underlying memory, because [`ffi_support::ByteBuffer` does not impl `Drop`](https://docs.rs/ffi-support/0.4.0/ffi_support/struct.ByteBuffer.html#caveats).

That seems suboptimal.

We need to dig into the precise details of `ByteBuffer` on the rust side and see what we need to do in order to free the underlying bytes.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1593)

rfk:
Poking in the code for `ByteBuffer` on the rust side, I'm actually not sure if this is true.

The `destroy` method for `ByteBuffer`, which is what we'd call if we wanted to manually drop the buffer, is implemented by [calling `self.into_vec()` and dropping the result](https://github.com/mozilla/application-services/blob/46e92e24abd29642c26c60020a9f04566680e3eb/components/support/ffi/src/lib.rs#L465).

When we lift values out of the `RustBuffer`, we do so by [explicitly calling `buf.into_vec()`](https://github.com/mozilla/uniffi-rs/blob/8daccdb683768a93ccaa807725f0031e648df10b/uniffi/src/support/mod.rs#L82) and then allowing the resulting `Vec` to go out of scope, which should cause it to be dropped.

So...I think maybe we're OK here, and the docs for `ByteBuffer` should be updated to reflect that converting to a `Vec` will handle the memory cleanup via drop.

@thomcc could you please lend your wisdom here?

thomcc:
[...]

PR #210, opened 2020-08-08
Use destroy_into_vec

weslenng:
`into_vec` throws a warning at compile time

```sh
warning: use of deprecated item 'ffi_support::ByteBuffer::into_vec': Name is confusing, please use `destroy_into_vec` instead
```

### DIFF
--- a/uniffi/src/lib.rs
+++ b/uniffi/src/lib.rs
@@ -91,7 +91,7 @@ pub trait Liftable: Sized {
 }
 
 pub fn try_lift<T: Liftable>(buf: ByteBuffer) -> Result<T> {
-    let vec = buf.into_vec();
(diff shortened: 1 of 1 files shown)


rfk:
Thank you! This addresses https://github.com/mozilla/uniffi-rs/issues/194#issuecomment-669670893.

rfk:
Whoops, I should have noticed that we also need to bump our version number on `ffi-support` in `Cargo.toml`. We currently specify `ffi-support = "0.4"`, which will pull in the latest `0.4.2` version on *new* builds, but won't cause existing checkouts to update to the latest version. I filed https://github.com/mozilla/uniffi-rs/pull/211 to fix.

PR #211, opened 2020-08-08
Update version requirement for ffi-support.

rfk:
As of https://github.com/mozilla/uniffi-rs/pull/210, we depend on `ByteBuffer.destroy_into_vec` which is first available in the 0.4.2 release of `ffi_support`. This uses the tilde syntax to say "any semvar-compatible version that is at least 0.4.2" (which frankly, nothing about `~` makes me think "at least", but ok...)


rfk:
(I am going to abuse my administrator privileges to merge this without review, so that folks with existing checkouts don't get a build error when they next `git pull`).

PR #214, opened 2020-08-09
Clean up python bindings generator, and add support for sequences.

### LINKED ISSUE #39: Audit Python helper size conversions
https://github.com/mozilla/uniffi-rs/blob/3c36c7b70c68a67c3861e4d14efc6c2857213b22/uniffi/src/bindings/python/templates/RustBufferHelper.py#L43-L44.

This looks like this corresponds to a `put_u32` on the Rust size, looking at the [Python docs](https://docs.python.org/3/library/struct.html#format-characters) it looks like we need to use `I` instead.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1551)

rfk:
(This isn't on the critical path for nimbus, so no rush on reviewing this; I was just messing around with an idea on a lazy weekend).

This is a significant refactor of the python bindings generator, with
the aim of making the generated code a bit cleaner and of better hiding
many of our implementation details from the public API. To help prove
out the approach, it adds previously-missing support for sequence types.

The key idea here is that, for any given `ComponentInterface`, we can
finitely enumerate all the possible types used in that interface
(including recursive types like sequences) and can give each such type
a unique name. This lets us define per-type helper methods on the internal
helper objects like `RustBuffer`, rather than putting these as "hidden"
methods on the public classes.

For example, for each type that lowers into a `RustBuffer`, there is
a corresponding `RustBuffer.allocFrom{{ type_name }}` classmethod for
lowering it and a `RustBuffer.consumeInto{{ type_name }}` for lifting it.

If you squint, this is a little bit like defining private traits and
implementing them for each type, except within the constraints of python's
much looser type system.

Fixes #39, among other things.


uniffi_bindgen/src/bindings/python/templates/RustBufferHelper.py
```diff
@@ -1,17 +1,13 @@
-# Helpers for lifting/lowering primitive data types from/to a bytebuffer.
+# Helpers for reading/writing types from/to a bytebuffer.
```
rfk, line 1: On a side note, I tried our calling these "read/write" rather than "lift_from/lower_into", and TBH I think it makes things a bit simpler to understand.

uniffi_bindgen/src/bindings/python/gen_python.rs
```diff
@@ -87,94 +87,82 @@ mod filters {
...
-            Type::Enum(type_name) => format!("{}({})", type_name, nm),
-            Type::String => format!("liftString({})", nm),
-            Type::Record(type_name) => format!("{}._lift({})", type_name, nm),
-            Type::Optional(type_) => format!(
-                "liftOptional({}, lambda buf: {})",
-                nm,
-                lift_from_py(&"buf", type_)?
-            ),
-            Type::Sequence(type_) => format!(
-                "liftSequence({}, lambda buf: {})",
```
rfk, line 163: Python lambdas are a pain, much of this is motivated by not wanting to build an elaborate series of nested lambdas for sequences-of-optional-types and things like that.

uniffi_bindgen/src/bindings/python/templates/RustBufferHelper.py
```diff
@@ -28,59 +24,230 @@ def _pack_into(self, size, format, value):
...
+
+    def read{{ type_name|class_name_py }}(self):
+        count = self._unpack_from(4, ">I")
+        items = []
+        while count > 0:
+            items.append(self.read{{ inner_type.unique_name()|class_name_py }}())
+            count -= 1
+        return items
+
+    def write{{ type_name|class_name_py }}(self, items):
```
rfk, line 246: As a concrete example, in `rondpoint` this produces a `writeSequenceEnumEnumeration` method specialized for writing out sequences of `Enumeration` objects. We also get a `writeRecDictionnaire` method specialized for writing out `Dictionnaire` objects, etc.

rfk:
Just a note that I think we should land https://github.com/mozilla/uniffi-rs/pull/215 before landing this one, given that it's the smaller of two concurrent python-related changes.

uniffi_bindgen/src/bindings/python/templates/RustBufferHelper.py
jhugman: I'd like to do this with all the backends.

rfk: If we like this, I'd be happy to go through the other backends and make them all consistent. @linacambridge @tarikeshaq @eoger feel free to thumbs-up or thumbs-down this comment for feedback :-)

linacambridge: That sounds really good. It did take me a while to wrap my head around the difference between `lift`, `liftFrom`, `lower`, and `lowerInto` at first, and especially since the thing that does the lifting and lowering is called a "reader" and a "writer", changing the names to read and write seems natural.

rfk: Aye, I think we can call it a bit of an experiment in terminology that didn't work out (which, it's good to notice these things and change them before they get too entrenched!)

rfk:
This has been rebased and updated, adding support for maps along with additional support for strings.

@jhugman and @eoger , A++ work on the round-tripping tests, they helped me find a few places here where the python code was subtly doing the wrong thing.

uniffi_bindgen/src/bindings/python/templates/RustBufferTemplate.py
```diff
@@ -16,4 +16,135 @@ def free(self):
         return _UniFFILib.{{ ci.ffi_bytebuffer_free().name() }}(self)
 
     def __str__(self):
-        return "RustBuffer(len={}, data={})".format(self.len, self.data[0:self.len])
\ No newline at end of file
+        return "RustBuffer(len={}, data={})".format(self.len, self.data[0:self.len])
```
rfk, line 19: Note to self: update this to use an explicitly-sized type rather than `ctypes.c_long`.

rfk:
I plan to revisit this once https://github.com/mozilla/uniffi-rs/pull/261 and https://github.com/mozilla/uniffi-rs/pull/262 have landed, as I think those wrap up all of the cross-cutting bindings changes that I wanted to make in this short term.

jhugman:
Revisiting this after reviewing #279; this is where a lot of those ideas came from.

The second language is easier than the first, the third is easier than the second. Let's add that PR to the long list of PRs that should land before this work re-commences on this one?

rfk:
> Let's add that PR to the long list of PRs that should land before this work re-commences on this one?

Yep! By the time we're finished, there's not going to be much left to this PR other than the python bits...which actually feels pretty right 👍

examples/rondpoint/tests/bindings/test_rondpoint.kts
```diff
@@ -88,11 +88,6 @@ fun <T> List<T>.affirmEnchaine(
 val wellKnown = st.wellKnownString("kotlin")
 assert("uniffi 💚 kotlin!" == wellKnown) { "wellKnownString 'uniffi 💚 kotlin!' == '$wellKnown'" }
 
-// NB. Numbers are all signed in kotlin. This makes roundtripping of unsigned numbers tricky to show. 
-// Uniffi does not generate unsigned types for kotlin, but the work tracked is 
-// in https://github.com/mozilla/uniffi-rs/issues/249. Tests using unsigned types are 
-// commented out for now.
```
rfk, line 94: I happened to notice this outdated comment while I was in here; unsigned types are included in the tests these days.

uniffi_bindgen/src/bindings/python/templates/RecordTemplate.py
```diff
@@ -1,49 +1,15 @@
...
-            return buf.finalize()
-        except Exception:
-            buf.discard()
-            raise
-
-    @classmethod
-    def _lowerInto(cls, v, buf):
-        {%- for field in rec.fields() %}
-        {{ "(v.{})"|format(field.name())|lower_into_py("buf", field.type_()) }}
-        {%- endfor %}
```
rfk, line 49: These are now helper methods on the `RustBuffer` class and its friends, in order to hide them from consumers.

uniffi_bindgen/src/bindings/python/templates/macros.py
```diff
@@ -5,15 +5,14 @@
...
 rust_call_with_error(
     {%- match func.throws() -%}
     {%- when Some with (e) -%}
     {{ e|class_name_py }}
     {%- else -%}
     InternalError
-    {%- endmatch -%},
-    _UniFFILib.{{ func.ffi_func().name() }}{% if func.arguments().len() > 0 %},{% endif %}{% call _arg_list_ffi_call(func) -%}
+    {%- endmatch -%}
+    ,_UniFFILib.{{ func.ffi_func().name() }}{% if func.arguments().len() > 0 %},{% endif %}{% call _arg_list_ffi_call(func) -%}
```
rfk, line 15: This little comma tweak makes the call emit entirely on one line, which seemed to help with some formatting issues in the generated code.

uniffi_bindgen/src/bindings/python/templates/wrapper.py
```diff
@@ -20,8 +20,8 @@
 import contextlib
 
 {% include "RustBufferTemplate.py" %}
-
-{% include "RustBufferHelper.py" %}
```
rfk, line 24: I split this up into two files, one for the reading and one for the writing. It would have been nice to group the operations by datatype rather than read/write, but didn't really work out with the structure of the abstractions here.

rfk (commented):
Alrighty, I think this is ready for a proper review, and I'm pleased that after a rebase it's almost entirely changes in the python-related code.

uniffi_bindgen/src/bindings/python/gen_python.rs
```diff
@@ -53,7 +53,7 @@ mod filters {
             FFIType::Float64 => "ctypes.c_double".to_string(),
             FFIType::RustCString => "ctypes.c_voidp".to_string(),
             FFIType::RustBuffer => "RustBuffer".to_string(),
-            FFIType::RustError => "RustError".to_string(),
+            FFIType::RustError => "POINTER(RustError)".to_string(),
```
jhugman, line 56: To my untrained eye, this doesn't look like Python.

uniffi_bindgen/src/bindings/python/templates/RustBufferStream.py
```diff
@@ -0,0 +1,179 @@
...
+        return data
+
+    # For every type used in the interface, we provide helper methods for conveniently
+    # reading that type in a buffer. Putting them on this internal helper object (rather
+    # than, say, as methods on the public classes) makes it easier for us to hide these
+    # implementation details from consumers, in the face of python's free-for-all type
+    # system.
+
+    {%- for typ in ci.iter_types() -%}
+    {%- let canonical_type_name = typ.canonical_name()|class_name_py -%}
```
jhugman, line 33: 🙇‍♂️ 🏅

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -100,30 +100,69 @@ impl<'ci> ComponentInterface {
...
+    /// List the definitions for every Record type in the interface.
     pub fn iter_record_definitions(&self) -> Vec<Record> {
         self.records.to_vec()
     }
 
+    /// Get a Record definition by name, or None if no such Record is defined.
+    pub fn get_record_definition(&self, name: &str) -> Option<&Record> {
+        // TODO: probably we could store these internally in a HashMap to make this easier?
+        self.records.iter().find(|r| r.name == name)
+    }
```
jhugman, line 131: IIUC, we only need this method, but not the other `get_*_definition` methods. I don't understand how this wasn't needed in the other bindings.

jhugman (approved):
I've left some comments and questions, but I'd be happy to see this land.

uniffi_bindgen/src/bindings/python/templates/macros.py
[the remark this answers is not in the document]
rfk: Aye, this got pretty hairy. Some of the changes here were due to problems with indentation, but I'll take another look and see if I can clean it up.

uniffi_bindgen/src/bindings/python/gen_python.rs
rfk: Nice catch, this at least needs `ctypes.POINTER(RustError)` otherwise the imports won't work out correctly.

uniffi_bindgen/src/interface/mod.rs
rfk: Right, we only need `get_record_definition` but I added the others for consistency.

The reason they'd needed here is because the python bindings generate some code for `type::Record` instances as part of doing `ci.iter_types`, while e.g. the Kotlin bindings generate all the code for `type::Record` while doing `ci.iter_record_definitions`. For python, we need to be able to go from the `type::Record(name)` that we get when iterating the types, to the actual definition of the record with the given name.

rfk:
Thanks @jhugman!

PR #215, opened 2020-08-09
Python error handling

### LINKED ISSUE #202: Add error handling for python
Not a priority, but with #31 merged, we added error handling and the binding for it in Kotlin and Swift.

We are missing the bindings for Python though.

What you'd need to do:
 - Check out how the error handling was implemented #31 for Kotlin and Swift
 - Modify the todolist python example to have try catches and assert a certain error. Something like:
```python
from todolist import *

try:
  entry = createEntryFrom("")
  assert(false)
except EmptyStringError:
  # We are okay, the string was empty!
```
 - Generate `Error`s/`Exception`s in Python that correspond to the Errors definined in the `ComponentInterface` over [here](https://github.com/mozilla/uniffi-rs/blob/89b1312a01ecab2922cc8b0f9ef251bb651e7689/uniffi/src/interface/mod.rs#L79)
  - You can see how that is done in the other  bindings, probably something like what is done [here](https://github.com/mozilla/uniffi-rs/blob/89b1312a01ecab2922cc8b0f9ef251bb651e7689/uniffi/src/bindings/kotlin/templates/wrapper.kt#L41) and [here](https://github.com/mozilla/uniffi-rs/blob/main/uniffi/src/bindings/kotlin/templates/ErrorTemplate.kt)
 - Do some python magic until the tests pass

Note: the todolist example also uses sequences... which are also not yet supported in Python, so you'll probably have to play around with the example so that you are not using any other features that are not yet implemented in Python (including uncommenting out this [line](https://github.com/mozilla/uniffi-rs/blob/89b1312a01ecab2922cc8b0f9ef251bb651e7689/examples/todolist/tests/test_generated_bindings.rs#L4) while you are doing your local testing.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1605)

### LINKED ISSUE #206: Code generation failing in Python
I'm trying to do #202 but `cargo run -- generate` fail with:

```sh
thread 'main' panicked at '[TODO: coerce_py(Sequence(Record("TodoEntry")))]', uniffi/src/bindings/python/gen_python.rs:97:18
```
- Mac OS Mojave 10.14.6
- Python 3.8.3

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1626)

weslenng:
This PR add support for Python error handling

Closes #202
Closes #206


uniffi_bindgen/src/bindings/python/templates/wrapper.py
```diff
@@ -23,6 +23,8 @@
 
 {% include "RustBufferHelper.py" %}
 
+{% include "RustErrorTemplate.py" %}
```
weslenng, line 26: @tarikeshaq @rfk I think Python doesn't allow the usage of a generic class-type `ctypes.POINTER(ctypes.Structure)` on args list. I can't make something like `argtypes = (ctypes.POINTER(ctypes.Structure),)` (and send a class which extends `ctypes.Structure`)

I made this file to avoid this. A `RustError` class is declared on top of file to get the Rust result across FFI.
You guys know a better way to solve this? 👀

examples/arithmetic/tests/test_generated_bindings.rs
```diff
@@ -2,7 +2,7 @@ uniffi_macros::build_foreign_language_testcases!(
     "src/arithmetic.idl",
     [
         "tests/bindings/test_arithmetic.py",
-        "tests/bindings/test_arithmetic.kts",
-        "tests/bindings/test_arithmetic.swift",
+        // "tests/bindings/test_arithmetic.kts", # REVIEW_BEFORE_PR
```
weslenng, line 5: For some reason Kotlin and Swift tests is just failing here

uniffi_bindgen/src/bindings/python/templates/wrapper.py
```diff
@@ -31,6 +33,9 @@
 {% include "EnumTemplate.py" %}
 {%- endfor -%}
 
+# Error definitions
+{% include "ErrorTemplate.py" %}
```
weslenng, line 37: @tarikeshaq @rfk The `ErrorTemplate` should generate custom exceptions (based on my Enum on the Rust side) and should have a function that is responsible for:

- Check an error result
- Select the correct custom exception based on the `RustError` code and,
- Raise this exception on the Python side

I didn't quite understand if this is really it 🤔

uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py
```diff
@@ -1,14 +1,15 @@
 {%- match func.return_type() -%}
 {%- when Some with (return_type) %}
 
-def {{ func.name()|fn_name_py }}({%- call py::arg_list_decl(func.arguments()) -%}):
-    {%- call py::coerce_args(func.arguments()) %}
+def {{ func.name()|fn_name_py }}({%- call py::arg_list_decl(func) -%}):
+    {%- call py::coerce_args(func) %}
+    t = RustError() # REVIEW_BEFORE_PR
```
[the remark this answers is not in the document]
weslenng: Oh, this is just a way which I find to test FFI, probably I will remove soon 😅

uniffi_bindgen/src/bindings/python/templates/wrapper.py
weslenng: > Not too familiar with what Python allows here, but this feels okay to me, and we can iterate on it if need be 😄 Also feel free to have this as a part of `ErrorTemplate.py` if you don't wanna add a new file.

It's a good idea!

examples/arithmetic/src/lib.rs
```diff
@@ -2,29 +2,30 @@
...
+    #[error("Integer overflow!")]
+    IntegerOverflow,
 }
 
-pub fn add(a: u64, b: u64, overflow: Overflow) -> u64 {
-    match overflow {
-        Overflow::WRAPPING => a.overflowing_add(b).0,
-        Overflow::SATURATING => a.saturating_add(b),
+fn add(a: u64, b: u64) -> Result<u64> {
+    match a.checked_add(b) {
```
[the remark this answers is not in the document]
weslenng: Nice one!

examples/arithmetic/tests/bindings/test_arithmetic.kts
```diff
@@ -1,6 +1,27 @@
...
+    throw RuntimeException("Should have thrown a IntegerOverflow exception!")
+} catch (e: ArithmeticErrorException) {
+    // It's okay!
+}
 
+assert(add(2, 4) == 6L)
+assert(add(4, 8) == 12L)
+
+try {
+    sub(0, 1)
```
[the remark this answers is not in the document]
tarikeshaq: Never mind this comment 😆 Got confused, it **should** throw an exception (since the subtraction happens on the rust side)

examples/arithmetic/tests/bindings/test_arithmetic.swift
```diff
@@ -1,6 +1,27 @@
 import arithmetic
 
-assert(add(a: 2, b: 3, overflow: .saturating) == 5, "addition works")
+do {
+    let _ = try add(18446744073709551615, 1)
```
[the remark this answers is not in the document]
weslenng: The Swift tests returns a lot of

```sh
error: 'Self' is only available in a protocol or as the result of a method in a class; did you mean 'String'
```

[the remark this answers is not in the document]
weslenng: > The Swift tests returns a lot of
> 
> ```shell
> error: 'Self' is only available in a protocol or as the result of a method in a class; did you mean 'String'
> ```

Probably my environment is missing something, but I fixed the test case using the CI feedback!

examples/arithmetic/tests/bindings/test_arithmetic.py
```diff
@@ -1,6 +1,28 @@
 from arithmetic import *
+from arithmetic import ArithmeticError
```
weslenng, line 2: I don't know why the import star is not working 🤔

uniffi_bindgen/src/bindings/python/templates/macros.py
```diff
@@ -5,19 +5,31 @@
...
+_RustErrorHelper.try_raise(_RustErrorHelperPartial(_UniFFILib.{{ func.ffi_func().name() }},{% call _arg_list_ffi_call(func) -%}))
+{%- else -%}
+_UniFFILib.{{ func.ffi_func().name() }}({% call _arg_list_ffi_call(func) -%})
+{%- endmatch -%}
 {%- endmacro -%}
 
 {%- macro to_ffi_call_with_prefix(prefix, func) -%}
+{%- match func.throws() -%}
+{%- when Some with (e) -%}
+_RustErrorHelper.try_raise(_RustErrorHelperPartial(
```
weslenng, line 19: How can I test this issued code?
It's probably working, but I haven't tested it

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -0,0 +1,51 @@
...
+class RustErrorHelper:
+    def __init__(self):
+        self.err = RustError()
+
+    def __apply__(self, fn):
+        return fn(ctypes.byref(self.err))
+
+    def __raise__(self):
+    {%- for e in ci.iter_error_definitions() %}
+        if self.err.code == {{loop.index}}:
```
tarikeshaq, line 35: Hmmm I'm not sure this does what you expect it to do.

Lets imagine this scenario:
```
[Error]
enum ArithmeticError {
"IntegerOverflow", "DivideByZero"
};

[Error]
enum StringError {
"EmptyStringError", "NullPointer"
};
```
If we receive an `ArithmeticError::DivideByZero` We would raise a `StringError::NullPointer`.

An alternative is to do what Swift does and accept an "empty" error of the correct type in `try_raise` (swift ref [here](https://github.com/mozilla/uniffi-rs/blob/bad6448e1f8e7ac80284f33197272ac833bbf39f/uniffi_bindgen/src/bindings/swift/templates/macros.swift#L10) notice how we pass an `{{e}}.NoError` to the `rustCall` helper).

It's not the best looking approach so feel free to try something else if you have another idea in mind 😄

uniffi_bindgen/src/bindings/python/templates/macros.py
tarikeshaq: Hmmm Ya this would run when we have an `interface`. You can take a look at how the `sprites` example defines an interface if you wanna test this!

examples/arithmetic/tests/bindings/test_arithmetic.py
rfk: In `uniffi_bindgen/src/bindings/python/templates/wrapper.py` at the bottom of the file, there's a blog that defines the special `__all__` attribute on the module. You'll need to add the error classes to that block in order to get them to appear as part of `import *`.

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __str__(self):
+        return "RustError(code={}, message={})".format(self.code, str(self.message, "utf-8")) 
+
+{% for e in ci.iter_error_definitions() %}
+class {{e.name()}}:
+    {%- for value in e.values() %}
+    class {{value}}(Exception):
+        pass
+    {%- endfor %}
```
rfk, line 15: I think this is a good place to start, but having nested class for each error like this is pretty un-idiomatic in python. I'm not sure how to expose our errors in a better way though, so I'm happy to land it like this for now, and we can explore other options later.

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -0,0 +1,51 @@
...
+    _fields_ = [
+        ("code", ctypes.c_int32),
+        ("message", ctypes.c_char_p),
+    ]
+
+    def __str__(self):
+        return "RustError(code={}, message={})".format(self.code, str(self.message, "utf-8")) 
+
+{% for e in ci.iter_error_definitions() %}
+class {{e.name()}}:
```
rfk, line 11: minor point, but: please use `{{ e.name()|class_name_py }}` and similar for `e.value`, to make sure these have the right case convention for python names. (Some of the other python classes don't do this either, but they should; I've tried to fix some of them over in #214

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -0,0 +1,51 @@
...
+    {%- for value in e.values() %}
+    class {{value}}(Exception):
+        pass
+    {%- endfor %}
+
+    @staticmethod
+    def raise_err(err):
```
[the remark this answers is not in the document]
rfk: If you don't want to use a finally block here, you could also try defining a `__del__` method on the `RustError` class, that calls the `string_free` FFI function on `self.message`. Either way, we do need to avoid leaking memory here.

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -0,0 +1,51 @@
...
+        return
+
+    def __reset__(self):
+        del self.err
+        self.err = RustError()
+
+    def try_raise(self, fn):
+        self.__reset__()
+        result = self.__apply__(fn)
+        self.__raise__()
```
rfk, line 47: (I'm not sure how familiar you are with python, so I apologize is some of this sounds too basic)

In general, we shouldn't define our own "__dunder__" style methods, which are typically reserved for special behavior defined by the language interpreter. If we intend these to be private helper methods they should be named with a single leading underscore, like `_reset()` and `_raise()`.

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -0,0 +1,51 @@
...
+        
+        return result
+
+_RustErrorHelper = RustErrorHelper()
```
rfk, line 51: Using a stateful object for this is going to be a threading hazard - imagine two threads calling into two different functions at the same time, but both using `self.err` as their out error parameter. We might accidentally fail to report an error, or report the wrong type of error, if they concurrently modified the shared `self.err` struct.

Fortunately, I don't think you need to use a stateful objectfor this. I suggest using a stateless helper function instead, using python's star-args syntax to make it generic in the number of arguments. Something like:

```
def rust_call_with_err(errorClass, fn, *args):
    err = RustError()
    args = args + (ctypes.byref(self.err),)  # `args` is a tuple, this puts `err` on the end of it
    result = fn(*args)
    if err.code != 0:
        errorClass.raise_err(err)
    return result
```

And then when you come to call an error-throwing FFI function, generate code that looks like this:

```
rust_call_with_err(ArithmeticError, _UniFFILib.arithmetic_add, a, b)
```

This avoids shared state, and by passing in the type of error as part of the call, you avoid the issue that @tarikeshaq mentions in https://github.com/mozilla/uniffi-rs/pull/215/files#r467653835 above.

uniffi_bindgen/src/bindings/python/templates/macros.py
```diff
@@ -5,19 +5,31 @@
...
+    _UniFFILib.{{ func.ffi_func().name() }},{{- prefix }}{% if func.arguments().len() > 0 %},{% endif %}{% call _arg_list_ffi_call(func) %})
+))
```
rfk, line 21: Just a heads-up that you might run into indentation issues if you put newlines in your macro output like this, since you don't have any guarantee about what indentation level the invocation of the macro is at, and python is very strict about indentation.

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
rfk: (Just to be clear: I'm not asking you to change this, I'm just thinking out load about whether there's a better way to do it that we could explore in future)

examples/arithmetic/tests/bindings/test_arithmetic.py
weslenng: Now it makes a lot more sense, thank you!

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
weslenng: > (I'm not sure how familiar you are with python, so I apologize is some of this sounds too basic)

I'm not familiar with Python, I was just trying not conflict with reserved keywords (e.g. `raise`). I'll change asap!

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
weslenng: Nice one!
I hadn't thought about parallelism when I did a stateful helper, it's really a problem

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
weslenng: I've created https://github.com/mozilla/uniffi-rs/issues/223 to keeping track on this!

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
weslenng: @rfk What `|class_name_py` does?

rfk: It basically just makes sure the names are `CamelCase`, which is the preferred naming scheme for classes in python.

weslenng:
Thanks for the feedback, it's probably okay now!

examples/arithmetic/tests/bindings/test_arithmetic.kts
```diff
@@ -1,6 +1,20 @@
 import uniffi.arithmetic.*;
 
-assert(add(2, 3, Overflow.SATURATING) == 5L)
+assert(add(2, 4) == 6L)
+assert(add(4, 8) == 12L)
 
+try {
+    sub(0, 2)
+    throw RuntimeException("Should have thrown a IntegerOverflow exception!")
+} catch (e: ArithmeticErrorException) {
```
rfk, line 9: In swift and python we explicitly test that this was the `IntegerOverflow` variant, but here we test only that it was an `ArithmeticError`. Does it work if we catch `ArithmeticErrorException.IntegerOverflow` here like the other languages do?

(I note that the todolist example has a similar problem, but we don't need to fix that here)

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -0,0 +1,43 @@
...
+        raise Exception("Unknown error code")
+{% endfor %}
+
+def rust_call_with_error(error_class, fn, *args):
+    error = RustError()
+    args_with_error = args + (ctypes.byref(error),)
+
+    result = fn(*args_with_error)
+    if error.code != 0:
+        message = error.__str__()
```
rfk, line 38:
```suggestion
        message = str(error)
```

The `__str__` method is designed to work with the `str()` function via python interpreter magic.

uniffi_bindgen/src/bindings/python/templates/macros.py
```diff
@@ -5,19 +5,27 @@
 #}
 
 {%- macro to_ffi_call(func) -%}
-_UniFFILib.{{ func.ffi_func().name() }}({% call _arg_list_ffi_call(func.arguments()) -%})
+{%- match func.throws() -%}
+{%- when Some with (e) -%}
+rust_call_with_error({{ e|class_name_py }},_UniFFILib.{{ func.ffi_func().name() }},{% call _arg_list_ffi_call(func) -%})
```
rfk: Nice, I really like how this worked out in practice 👍

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -0,0 +1,43 @@
...
+    def raise_err(code, message):
+        {%- for value in e.values() %}
+        if code == {{ loop.index }}:
+            raise {{ e.name()|class_name_py }}.{{ value|class_name_py }}(message)
+        {% endfor %}
+        raise Exception("Unknown error code")
+{% endfor %}
+
+def rust_call_with_error(error_class, fn, *args):
+    error = RustError()
```
rfk, line 33: Please explicitly set `error.code = 0` here, just to make sure we're not passing uninitialized data into Rust. (Probably something in the stack here would do that for us already, but it won't hurt to be extra cautious here).

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
weslenng: Oh, nice!

PR #217, opened 2020-08-10
Roundtrip and stringify testing for primitives

### LINKED ISSUE #175: Equivalence and roundtrip tests for basic types at either end of the ffi.
This issue is to make more systematic the testing of matching the lower/lift implementations between platform languages and rust for the basic types.

This is likely simplest achieved by sending two representations of the same thing from one language, then testing the two at the other end. e.g. 

```rust
fn u32_translates(left: u32, right: String) -> Bool
```

Fuzz testing, or property testing should be sufficient to give us a high degree of confidence that the lower/lift pairs match from platform to rust work as intended. 

This should be a jumping off point to roundtripping through rust.

```rust
fn u32_identity(in_value: u32) -> u32
```

Ideally it should be obvious when a platform is missing testing, so we can avoid getting backends out of sync.

Note: we need to test literals for `bool_identity` in order to have ensure `u32_translates` works.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1576)

jhugman:
Fixes #175 

This PR adds a good deal of certainty around the lowering and lifting code for primitives. 

The idea is that roundtripping gives us some confidence that two rounds of lowering/lifting works, but not that the value that rust gets is the value sent by the foreign language.

e.g. 
1. calling round: lower from swift to C then lift from C to rust.
2. returning round: lower from rust to C, then lift from C to swift.

i.e. if we only test by roundtripping, we only verify that the two lowering/lifting rounds are symmetrical, not that they are both correct.

If we can convince ourselves that one round is correct, and that both rounds are symmetrical, then we can say that both are correct.

To do this, we ask rust to transform the received value into a different, but well known representation, and send _that_ back to be compared in swift.

In this way, we can exercise and test the FFI, without having to get oracle data into rust in another way (e.g. hard coding into rust, or loading a file).

### DIFF
--- a/examples/rondpoint/src/lib.rs
+++ b/examples/rondpoint/src/lib.rs
@@ -50,4 +50,97 @@ fn switcheroo(b: bool) -> bool {
     !b
 }
 
+// Test that values can traverse both ways across the FFI.
+// Even if roundtripping works, it's possible we have
+// symmetrical errors that cancel each other out.
+#[derive(Debug, Clone)]
+struct Retourneur;
+impl Retourneur {
+    fn new() -> Self {
+        Retourneur
+    }
+    fn identique_i8(&self, value: i8) -> i8 {
+        value
+    }
+    fn identique_u8(&self, value: u8) -> u8 {
+        value
+    }
+    fn identique_i16(&self, value: i16) -> i16 {
+        value
+    }
+    fn identique_u16(&self, value: u16) -> u16 {
+        value
+    }
+    fn identique_i32(&self, value: i32) -> i32 {
+        value
+    }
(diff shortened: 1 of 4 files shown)


examples/rondpoint/src/lib.rs
```diff
@@ -50,4 +50,97 @@ fn switcheroo(b: bool) -> bool {
     !b
 }
 
+// Test that values can traverse both ways across the FFI.
+// Even if roundtripping works, it's possible we have
+// symmetrical errors that cancel each other out.
+#[derive(Debug, Clone)]
+struct Retourneur;
```
eoger, line 57: :100: great name

Issue #188, opened 2020-07-31
Figure out optional dictionary fields


tarikeshaq:
Based on the discussion in https://github.com/mozilla/application-services/issues/3440

We need a clear story on how we represent optional fields in dictionaries.

A few things from that discussion:

- We may need to set default values
- Should they be `nullable`?

I Haven't looked at the details here, but opening this for further discussion in this repo

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1588)

jhugman:
Thinking by documenting here:

 * Values with `nullable` types are allowed to be `null`. They most  closely correspond to `T?` in Kotlin and Swift, or `Option<T>` in Rust. 
 * `optional` is allowing the fields or argument to be _omitted_, and the behavior when it is. For JS, the key is simply not there in the object. For Swift and Kotlin, the keys are there, but with default values. Rust has no such concept. 

I think what I'm proposing here is to use `optional` to use foreign language features to fully construct the struct before crossing the FFI.

E.g. in Swift, a `struct` generates a _memberwise initializer_ that accounts for default values.

```swift
struct TVCharacter {
    let name: String
    let yearsActive: Int = 0
}

let roslin = TVCharacter(name: "Laura Roslin")
let adama = TVCharacter(name: "William Adama", yearsActive: 45)
```

Above, `yearsActive` is optional, but not `nullable`. Below, `yearsActive` is nullable, but not optional.

```swift
struct TVCharacter {
    let name: String
    let yearsActive: Int?
}

let starbuck = TVCharacter(name: "Kara Thrace", yearsActive: null)
```

Once the struct is fully constructed with all the values set in Swift, only then is it sent over the FFI.

The only confusion in my mind is this: what happens when we specify a field that is `optional`, but don't provide a default value:

 * fail fast! we can't omit the key, so we need a default value.
 * guess! Essentially make the `Default::default()`.
 * silently make them nullable/`Option<T>` types, and set them to `null`/`None`.

I don't know which of these are the least bad.

@linacambridge @rfk @tarikeshaq ? thoughts?

rfk:
This makes sense to me, thanks for writing it out so clearly. I favour this option in the short term:

> fail fast! we can't omit the key, so we need a default value.

If only because it's easy to change in a backwards-compatible way later.

[...]

PR #218, opened 2020-08-10
Adds first pass of nimbus api

### LINKED ISSUE #178: [demo] Mock Nimbus API
Work to do is define an "expected" .idl API for Nimbus, build a "mock" Rust crate for it and create an example out of it.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1580)

tarikeshaq:
fixes #178, relates to https://github.com/mozilla/application-services/issues/3440

Adds the proposed nimbus API as an example here. 

compiles ✅ 
bindings can be generated ✅ 
tested ❌ 

I intentionally left out the README we usually add in examples, created #219 to consolidate the README's (since they all say the same thing 😆)

### DIFF
--- a/examples/nimbus/src/lib.rs
+++ b/examples/nimbus/src/lib.rs
@@ -1,0 +1,70 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#[derive(Debug, Clone)]
+pub struct ApplicationContext {
+    app_id: Option<String>,
+    app_version: Option<String>,
+    locale: Option<String>,
+    device_manufacturer: Option<String>,
+    device_model: Option<String>,
+    region: Option<String>,
+    debug_tag: Option<String>,
+}
+
+pub struct Experiment {
+    slug: String,
+    user_facing_name: String,
+    user_facing_description: String,
+}
+
+pub struct Config {
+    server_url: Option<String>,
+    uuid: Option<String>,
+    collection_name: Option<String>,
+    bucket_name: Option<String>,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ExperimentError {
+    #[error("Error from the network")]
+    NetworkError,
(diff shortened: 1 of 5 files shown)


examples/nimbus/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+
+type Result<T, E = ExperimentError> = std::result::Result<T, E>;
+pub struct Experiments {}
+
+impl Experiments {
+    pub fn new<P: AsRef<Path>>(_: ApplicationContext, _: P, _: Option<Config>) -> Result<Self> {
+        unimplemented!()
+    }
+
+    pub fn get_experiment_branch<S: Into<String>>(&self, _: S) -> Option<String> {
```
jhugman, line 46: This is surprising. Why not `String` or `&str`?

```rust
<S: Into<String>>
```

tarikeshaq: Aaah this was muscle memory from when we had asymmetry, and I wanted my `todolist` to be able to take in either `&str` or `String`. Having as `String` is better though, thanks!!

examples/nimbus/src/nimbus.idl
```diff
@@ -0,0 +1,45 @@
+namespace nimbus {};
+dictionary ApplicationContext {
+    string? app_id;
+    string? app_version;
+    string? locale;
+    string? device_manufacturer;
+    string? device_model;
+    string? region;
+    string? debug_tag;
```
jhugman, line 9: Nice. Does this mean we can deprioritize #188 now?

tarikeshaq: Yes! it makes more sense to me to be nullable!

rfk:
@tarikeshaq do we still want this, or is it superseded by the actual nimbus crate uniffi integration over in that repo?

tarikeshaq:
Yes it is! Thanks @rfk!

I'll close this. Since the uniffi is already a part of the build process there, we don't need a nimbus specific test here, and it'll be easy to detect if something went wrong there with uniffi. (For future reference if someone comes, along "there" means https://github.com/mozilla/nimbus-sdk)

PR #220, opened 2020-08-10
Fixup destructor crashes

jhugman:
The PR fixes a bug which crept in during simultaneous landing of errors and destructors.

Cursory testing of #217 prompted this discovery.

### DIFF
--- a/examples/todolist/tests/bindings/test_todolist.swift
+++ b/examples/todolist/tests/bindings/test_todolist.swift
@@ -43,6 +43,7 @@ todo.addItems(items: ["bobo", "fofo"])
 assert(todo.getItems().count == 9)
 assert(todo.getItems()[7] == "bobo")
 
+// Ensure deinit doesn't crash.
 for _ in 0..<10 {
     let list = TodoList()
     try! list.addItem(todo: "todo")
--- a/examples/todolist/tests/bindings/test_todolist.swift
+++ b/examples/todolist/tests/bindings/test_todolist.swift
@@ -42,3 +42,8 @@ assert(try! todo.getLast() == "bar")
 todo.addItems(items: ["bobo", "fofo"])
 assert(todo.getItems().count == 9)
 assert(todo.getItems()[7] == "bobo")
+
+for _ in 0..<10 {
+    let list = TodoList()
+    try! list.addItem(todo: "todo")
+}
--- a/uniffi_bindgen/src/templates/ObjectTemplate.rs
+++ b/uniffi_bindgen/src/templates/ObjectTemplate.rs
@@ -12,7 +12,10 @@ uniffi::deps::lazy_static::lazy_static! {
 }
 
     {% let ffi_free = obj.ffi_object_free() -%}
-    uniffi::deps::ffi_support::define_handle_map_deleter!({{ handle_map }}, {{ ffi_free.name() }});
+    #[no_mangle]
+    pub extern "C" fn {{ ffi_free.name() }}(handle: u64) {
+        let _ = {{ handle_map }}.delete_u64(handle);
+    }
 
 {%- for cons in obj.constructors() %}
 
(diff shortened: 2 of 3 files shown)


examples/todolist/tests/bindings/test_todolist.swift
```diff
@@ -42,3 +42,8 @@ assert(try! todo.getLast() == "bar")
 todo.addItems(items: ["bobo", "fofo"])
 assert(todo.getItems().count == 9)
 assert(todo.getItems()[7] == "bobo")
+
```
tarikeshaq, line 45: nano nit: Can we add a comment here that this loop tests destructors?

uniffi_bindgen/src/templates/ObjectTemplate.rs
```diff
@@ -12,7 +12,10 @@ uniffi::deps::lazy_static::lazy_static! {
 }
 
     {% let ffi_free = obj.ffi_object_free() -%}
-    uniffi::deps::ffi_support::define_handle_map_deleter!({{ handle_map }}, {{ ffi_free.name() }});
+    #[no_mangle]
+    pub extern "C" fn {{ ffi_free.name() }}(handle: u64) {
+        let _ = {{ handle_map }}.delete_u64(handle);
```
tarikeshaq, line 17: We can probably open a separate issue for handling the error that comes back from `delete_u64`

uniffi_bindgen/src/templates/ObjectTemplate.rs
```diff
@@ -12,7 +12,10 @@ uniffi::deps::lazy_static::lazy_static! {
 }
 
     {% let ffi_free = obj.ffi_object_free() -%}
-    uniffi::deps::ffi_support::define_handle_map_deleter!({{ handle_map }}, {{ ffi_free.name() }});
+    #[no_mangle]
+    pub extern "C" fn {{ ffi_free.name() }}(handle: u64) {
```
eoger, line 16: @thomcc does this makes sense to define the destructor differently?

linacambridge:
Hmmm, is this because the handle is invalid after the first time it's deleted, so the destructor returns an error the second time it's caught? Didn't we also clear out the handle on the Kotlin side the first time `destroy` was called, to avoid this? Or are the crashes happening in Swift—in which case, how is `deinit` getting called multiple times?

jhugman:
@linacambridge sorry, the PR didn't have a lot of context.

The crash was occurring because the hand-crafted method call on the foreign language side was calling with only the handle, and not the error too. 

i.e. 

```swift
{{ obj.ffi_object_free().name() }}(self.handle)
```

And the proc macro was generating: 

```rust
    #[no_mangle]
    pub extern "C" fn {{ ffi_free.name() }}(handle: u64, err: &mut ExternError) {}
```

Two solutions: 

 1. generate similar looking call sites on the foreign side as other calls that use errors.
 2. use a simpler `object_free` implementation, with simpler call sites.

Generating the errors seems a lot of effort 
```kt
rustCall({{err}}.ByReference()) { err ->
        {{ obj.ffi_object_free().name() }}(handle, err)
}
```

But: what `err` to use? We're going to be throwing away the error anyway aren't we?

Additional considerations: 

 * the destructor implementation is going to get [panic safety][#221], so getting the implementation out of the proc_macro might be a good idea sooner or later.
 * adding these internal methods is difficult with the way we have our askama macros at the moment.

Please advise! @linacambridge @tarikeshaq

uniffi_bindgen/src/templates/ObjectTemplate.rs
jhugman: :+1:

examples/todolist/tests/bindings/test_todolist.swift
jhugman: :+1:

uniffi_bindgen/src/templates/ObjectTemplate.rs
jhugman: Filed #236 so I can land this.

PR #221, opened 2020-08-10
Ensure panic safety for all FFI methods

### LINKED ISSUE #32: Use panic-safety helper functions from ffi-support

linacambridge:
This commit changes all FFI functions to take an `out_err` param, so
that they're called using `call_with_result`. Any functions that don't
have a `[Throws]` annotation in the IDL will throw a generic
`InternalError` instead.

Closes #32.

### DIFF
--- a/examples/arithmetic/tests/bindings/test_arithmetic.kts
+++ b/examples/arithmetic/tests/bindings/test_arithmetic.kts
@@ -13,6 +13,15 @@ try {
 assert(sub(4u, 2u) == 2uL)
 assert(sub(8u, 4u) == 4uL)
 
+assert(div(8u, 4u) == 2uL)
+
+try {
+    div(8u, 0u)
+    throw RuntimeException("Should have panicked when dividing by zero")
+} catch (e: InternalException) {
+    // It's okay!
+}
+
 assert(equal(2u, 2uL))
 assert(equal(4u, 4uL))
 
--- a/examples/arithmetic/tests/bindings/test_arithmetic.py
+++ b/examples/arithmetic/tests/bindings/test_arithmetic.py
@@ -20,6 +20,16 @@ except ArithmeticError.IntegerOverflow:
 assert sub(4, 2) == 2
 assert sub(8, 4) == 4
 
+assert div(8, 4) == 2
+
+try:
+    div(8, 0)
+except InternalError:
+    # It's okay!
+    pass
+else:
+    assert(not("Should have panicked when dividing by zero"))
+
 assert equal(2, 2)
 assert equal(4, 4)
 
--- a/uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
@@ -78,6 +78,23 @@ internal open class RustError : Structure() {
     }
 }
 
+internal open class InternalError : RustError() {
+    class ByReference: InternalError(), RustErrorReference
+
+    @Suppress("ReturnCount", "TooGenericExceptionThrown", "UNCHECKED_CAST")
+    override fun<E: Exception> intoException(): E {
+        if (!isFailure()) {
+            // It's probably a bad idea to throw here! We're probably leaking something if this is
+            // ever hit! (But we shouldn't ever hit it?)
+            throw RuntimeException("[Bug] intoException called on non-failure!")
+        }
+        val message = this.consumeErrorMessage()
+        return InternalException(message) as E
+    }
+}
+
+class InternalException(message: String) : Exception(message)
+
 {%- for e in ci.iter_error_definitions() %}
 internal open class {{e.name()}} : RustError() {
     class ByReference: {{e.name()}}(), RustErrorReference
--- a/uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
@@ -13,6 +13,11 @@ class RustError(ctypes.Structure):
             str(ctypes.cast(self.message, ctypes.c_char_p).value, "utf-8"),
         )
 
+class InternalError(Exception):
+    @staticmethod
+    def raise_err(code, message):
+        raise InternalError(message)
+
 {% for e in ci.iter_error_definitions() %}
 class {{ e.name()|class_name_py }}:
     {%- for value in e.values() %}
--- a/uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
@@ -9,6 +9,36 @@ protocol RustError: LocalizedError {
     static func fromConsuming(_ rustError: NativeRustError) throws -> Self?
 }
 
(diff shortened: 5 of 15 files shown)


linacambridge:
I think Python is broken now because methods without `[Throws]` are now expecting a `RustError` argument as well, but that'll be fixed with #215.

rfk:
Will there be any conflicts or problems if https://github.com/mozilla/uniffi-rs/pull/215 lands ahead of this PR, or should we hold on of that one until this one is landed?

uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
```diff
@@ -9,6 +9,47 @@ protocol RustError: LocalizedError {
...
+// the library level.
+enum InternalError: RustError {
+    case bufferOverflow
+    case incompleteData
+    case unexpectedOptionalTag
+    case unexpectedEnumCase
+    case emptyResult
+    case panic(message: String)
+    case invalidHandle(message: String)
+    case unknownError(code: Int32)
```
rfk, line 22: Possibly contrary opinion: I'm not convinced it's valuable to define an elaborate series of cases for each individual type of internal error that might be thrown. If anyone ever finds themselves matching on these different cases and doing anything other than freaking out and filing a bug against uniffi, something has gone terribly wrong 😅.

In that spirit, the most important thing is that these errors surface with a useful human-readable message that helps us find and fix the bug. If enumerating all the cases helps us achieve that, awesome. But if they're just adding layers of indirection to a fixed string value, then I worry they might set incorrect expectations about how these internal errors should be handled.

I'm not going to r- for that though, just my 2c.

tarikeshaq:
@rfk I think we can land #215. And after this merges, we can spin off another issue to tackle panics in Python based on what we come with here! Thoughts?

> Naively, as a consumer I think I would expect a panicking [Throws(E)] function to throw an InternalError.Panic, rather than throwing e.g. some sort of E.InternalError.Panic variant. Does that mesh with what's being proposed here?

Hmm ya this makes more sense... Although I can't instantly think of a clean way to do this (since we pass the error type as a part of the `rustCall`).

This makes me wonder if we should just `panic` on the side of the binding (`fatalError` in swift and maybe a `RuntimeException` in Kotlin) if we receive an internal error (as opposed to raising a special exception)

I defer to @linacambridge's judgement though 😅

uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
linacambridge: Hmm, thinking some more about this, I'm coming around to what you and Tarik suggested—assert for any kind of internal error instead of bubbling it up as an exception. In real-world code, `[Throws]` is for defining error cases that are meaningfully recoverable—a network error or duplicate experiment name, for example. A Rust-side panic _might_ be, but...how do you fix that? If the app were fully in Rust, it would crash. The only reason we catch panics is to avoid stack unwinding into the foreign language, not as a graceful error recovery mechanism.

rfk:
>  And after this merges, we can spin off another issue to tackle panics in Python based on what we come with here! Thoughts?

Yep. I'll probably try to tackle that as part of my big make-python-work refactor over in #214 (and I'd hate to derail any critical-path nimbus stuff in the service of the python bindings).

linacambridge:
Yes, let's land #215 first, and then this. I'll go ahead and make the changes to fatally assert on an internal error, instead of throwing.

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -78,6 +78,23 @@ internal open class RustError : Structure() {
...
+            // It's probably a bad idea to throw here! We're probably leaking something if this is
+            // ever hit! (But we shouldn't ever hit it?)
+            throw RuntimeException("[Bug] intoException called on non-failure!")
+        }
+        val message = this.consumeErrorMessage()
+        return InternalException(message) as E
+    }
+}
+
+class InternalException(message: String) : Exception(message)
```
linacambridge, line 96: OK, I simplified the error hierarchy here so that we have a catch-all "internal error" instead of getting fancy with reporting panic and invalid handle codes. We might want to report these specially to Sentry, for example, but I totally agree consumers shouldn't be matching on them. WDYT?

linacambridge:
I went ahead and did a cleanup pass and added a quick test—would you mind taking another look please, @tarikeshaq? ☺️

examples/arithmetic/tests/bindings/test_arithmetic.kts
```diff
@@ -13,6 +13,15 @@ try {
 assert(sub(4u, 2u) == 2uL)
 assert(sub(8u, 4u) == 4uL)
 
+assert(div(8u, 4u) == 2uL)
+
+try {
+    div(8u, 0u)
+    throw RuntimeException("Should have panicked when dividing by zero")
+} catch (e: Exception) {
+    // It's okay!
```
rfk, line 22: IIUC this will catch the `RuntimeException` thrown above, meaning the test will always pass even if the rust code didn't panic at all. Should we catch a more specific type of error here?

examples/arithmetic/tests/bindings/test_arithmetic.py
```diff
@@ -20,6 +20,15 @@
 assert sub(4, 2) == 2
 assert sub(8, 4) == 4
 
+assert div(8, 4) == 2
+
+try:
+    div(8, 0)
+    assert(not("Should have panicked when dividing by zero"))
+except:
```
rfk, line 28: This will catch and ignore the `AssertionError` from "should have panicked" above. In this case you can use the `try-else` syntax like this:

```
try:
    div(8, 0)
except:
    pass
else:
    assert False, "Should have panicked"

File: uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py [hunk 5]
```diff
@@ -13,6 +13,11 @@ def __str__(self):
             str(ctypes.cast(self.message, ctypes.c_char_p).value, "utf-8"),
         )
 
+class InternalError:
+    @staticmethod
+    def raise_err(code, message):
+        raise Exception("Unexpected error {}: {}".format(code, message))
```
rfk, line 19: nit: please throw a `RuntimeException` here rather than the generic `Exception` base class.

rfk (approved):
Looks good to me :-)

Issue #234, opened 2020-08-12
Auto-format generated bindings


eoger:
Just like what we do with the Rust scaffolding files, we should do the same with Python/Kotlin/Swift files

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1653)

PR #235, opened 2020-08-12
Format generated bindings

eoger:
Fixes https://github.com/mozilla/uniffi-rs/issues/234.

We auto-format code after generation using:
- `swiftformat` for Swift
- `ktlint` for Kotlin
- `yapf` for Python

This feature is enabled by default, but does not generate a hard error if these tools are not available.


tarikeshaq:
> but I am sad that we're going to stop caring about the formatting of the generated code.

Ya, but maybe this makes us focus more on how the templates look, and assuming "clean templates" `!==` "clean generated code" It could be a good thing 🤷

rfk:
Yep to all the above.

On balance, I think that if this lets us optimize for maintainability of the templates rather than readability of the final code then it's probably a win.

PR #239, opened 2020-08-12
Properly snake-case Swift bindings

### LINKED ISSUE #238: Swift generated bindings case inconsistency
A dictionary
```
dictionary BoboZORD {
}
```

ends up with the following bindings (see how `ZORD` is not camel-cased in the body of the struct):
```
public struct BoboZord: Lowerable, Liftable, Serializable, Equatable {
    public static func == (lhs: BoboZORD, rhs: BoboZORD) -> Bool {
        return true
    }

    static func lift(from buf: Reader) throws -> BoboZORD {
        return try BoboZORD(
        )
    }
}
```

which results in a compilation error

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1656)

eoger:
Fixes #238

### DIFF
--- a/examples/rondpoint/src/lib.rs
+++ b/examples/rondpoint/src/lib.rs
@@ -17,6 +17,17 @@ pub enum Enumeration {
     Trois,
 }
 
+#[allow(non_camel_case_types)]
+#[allow(non_snake_case)]
+pub struct minusculeMAJUSCULEDict {
+    minusculeMAJUSCULEField: bool,
+}
+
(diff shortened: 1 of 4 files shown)


rfk:
> Can't think of a clean way, but it would be so nice if our .name() calls were smart enough to figure out the format on their own.

Possibly simpler, I wonder if we could somehow hide the `name()` method and instead populate the template with `rec | class_name_kt` (so you can't get the name directly, you can only get it through one of the formatting filters).

PR #240, opened 2020-08-12
Output Kotlin bindings into package path

eoger:
Java loves folders, and so when running `uniffi-bindgen generate MYPROJECT --language kotlin --out-dir MY_OUT_DIR`, you will have to create `MY_OUT_DIR/uniffi/MYPROJECT` beforehand for it to make sense and point `--out-dir` to `MY_OUT_DIR/uniffi/MYPROJECT`

Note that I'd be cool to be able to specify the package name with an idl directive (kinda like protobufs)

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/mod.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/mod.rs
@@ -22,10 +22,12 @@ pub fn write_bindings(
     out_dir: &Path,
     try_format_code: bool,
 ) -> Result<()> {
-    let mut kt_file = PathBuf::from(out_dir);
+    let config = Config::from(&ci);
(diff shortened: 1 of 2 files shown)


rfk:
> Note that I'd be cool to be able to specify the package name with an idl directive (kinda like protobufs)

We could also consider a separate per-language config for this sort of stuff (Maybe `uniffi.ini`? A custom section in `Cargo.tom`?).

rfk:
> you will have to create MY_OUT_DIR/uniffi/MYPROJECT beforehand for it to make sense and point --out-dir to MY_OUT_DIR/uniffi/MYPROJECT

To clarify, are you saying that this is what you would *previously* need to do by hand, and this PR makes it so that now happens automatically?

eoger:
Yup you got that right, sorry it wasn't clear!

PR #241, opened 2020-08-13
Clarify terminology around passing values across the FFI.

### LINKED ISSUE #29: Rename `Serializable` to `ViaFFIUsingByteBuffer` in the Swift bindings
..And `Primitive` to `ViaFFI`.

Even though they're not quite the same, the comment already says they're pretty analogous, and it would be nice to have that bit of symmetry—especially with #26 splitting up `ViaFFI` into `IntoFFI` and `TryFromFFI`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1557)

rfk:
Inspired by the conversation in https://github.com/mozilla/uniffi-rs/pull/214#discussion_r467578565,
here's a PR to try to clarify and simplify some terminology.

Previously, we used a loosely-defined mix of terms like "lower" and "lower into"
to describe passing data across the FFI in two very different ways (the former
by converting to a primitive C type, the later by serializing to a bytebuffer).
Sometimes when we said "lower" we meant the former, and sometimes we meant the
later, and this was not consistent between the different codegen backends.

This commit tries to clean up the terminology and bring the Rust, Kotlin and
Swift code generators into agreement on which activities are called what.
Broadly:

* "Lowering" is now exclusively about converting from a high-level language
  type into a low-level C-compatible type, and "lifting" is the reverse.
* Serializing into a bytebuffer is now called "writing" and deserializing
  from a bytebuffer is now called "reading".

I've also consolidated all these into a single trait/protocol/what-have-you
where possible, since we've had confusion in the past where some types implemented
e.g. lowering but not writing.

I haven't touched the python bindings, because I'm working on a bigger
refactor in a separte commit.

Fixes #29

### DIFF
--- a/uniffi/src/lib.rs
+++ b/uniffi/src/lib.rs
@@ -2,6 +2,28 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+//! # Runtime support code for uniffi
+//!
+//! This crate provides the small amount of runtime code that is required by the generated uniffi
+//! component scaffolding in order to transfer data back and forth across the C-style FFI layer,
+//! as well as some utilities for testing the generated bindings.
+//!
+//! The key concept here is the [`ViaFfi`] trait, which must be implemented for any type that can
+//! be passed across the FFI, and which determines:
+//!
+//!  * How to [represent](ViaFfi::Value) values of that type in the low-level C-style type
+//!    system of the FFI layer.
+//!  * How to ["lower"](ViaFfi::lower) rust values of that type into an appropriate low-level
+//!    FFI value.
+//!  * How to ["lift"](ViaFfi::lift) low-level FFI values back into rust values of that type.
+//!  * How to [write](ViaFfi::write) rust values of that type into a bytebuffer, for cases
+//!    where they are part of a compount data structure that is serialized for transfer.
+//!  * How to [read](ViaFfi::read) rust values of that type from bytebuffer, for cases
+//!    where they are received as part of a compound data structure that was serialized for transfer.
+//!
+//! This logic encapsulates the rust-side handling of data transfer. Each foreign-language binding
+//! must also implement a matching set of data-handling rules for each data type.
+
 use anyhow::{bail, Result};
 use bytes::buf::{Buf, BufMut};
 use ffi_support::ByteBuffer;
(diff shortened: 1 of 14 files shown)


uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
```diff
@@ -16,11 +16,11 @@ public enum {{ e.name()|class_name_swift }}: Lowerable, Liftable {
         }
     }
 
-    func lower(into buf: Writer) {
-        self.toFFIValue().lower(into: buf)
+    func write(into buf: Writer) {
+        self.lower().write(into: buf)
```
rfk, line 20: I'm not super familiar with visibility rules in Swift, will these `read` and `write` methods be visible to consumers by default? I'm wondering if we need to e.g. put them in an extension block in order to take them out of the public API.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -120,196 +120,199 @@ class Writer {
...
-    }
-
-    func lower(into buf: Writer) {
-        let len = UInt32(self.utf8.count)
-        buf.writeInt(len)
-        buf.writeBytes(self.utf8)
-    }
+// Types confirming to `ViaFfi` can be transferred back-and-for over the FFI.
+// This is analogous to the Rust trait of the same name.
+protocol ViaFfi: Serializable {
```
rfk, line 132: I've tried to make this map more directly to the matching traits on the rust side, @linacambridge could you please check that I haven't totally destroyed your beautiful creation here?

uniffi_bindgen/src/scaffolding.rs
```diff
@@ -78,7 +78,7 @@ mod filters {
         // implementations of the functions that we're wrapping (and also to type-check our generated code).
         // This will panic if the bindings provide an invalid value over the FFI.
         Ok(format!(
-            "<{} as uniffi::ViaFfi>::try_from_ffi_value({}).unwrap()",
```
rfk, line 81: If y'all hate "lift" and "lower" in practice, we could also change this to use the more verbose "into_ffi_value" and "from_ffi_value" names everywhere, rather than getting rid of them on the rust side.

uniffi/src/lib.rs
```diff
@@ -23,212 +45,291 @@ pub mod deps {
...
-    Ok(value)
-}
-
-macro_rules! impl_lowerable_liftable_for_num_primitive {
+/// Blanket implementation of ViaFfi for numeric primitives.
+///
+/// Numeric primitives have a straightforward mapping into C-compatible numeric types,
+/// sice they are themselves a C-compatible numeric type!
+macro_rules! impl_via_ffi_for_num_primitive {
     ($($T:ty,)+) => { impl_lowerable_liftable_for_num_primitive!($($T),+); };
```
eoger, line 161:
```suggestion
    ($($T:ty,)+) => { impl_via_ffi_for_num_primitive!($($T),+); };
```

uniffi/src/lib.rs
```diff
@@ -2,6 +2,28 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+//! # Runtime support code for uniffi
```
eoger, line 5: I love how well you documented this file, :100:

uniffi/src/lib.rs
```diff
@@ -23,212 +45,291 @@ pub mod deps {
...
+///
+/// This is an unsafe trait (implementing it requires `unsafe impl`) because we can't guarantee
+/// that it's safe to pass your type out to foreign-language code and back again. Buggy
+/// implementations of this trait might violate some assumptions made by the generated code,
+/// or might not match with the corresponding code in the generated foreign-language bindings.
+///
+/// In general, you should not need to implement this trait by hand, and should instead rely on
+/// implementations generated from your component IDL via the `uniffi-bindgen scaffolding` command.
+
+pub unsafe trait ViaFfi: Sized {
```
eoger, line 68: I like that this trait will ensure we don't miss Lowerable/Liftable impls anymore since everything is in a single trait now :100:

uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
eoger: I think by default they won't

jhugman: By default they're `internal`, i.e. won't be accessible outside of the module.

[source][1]

[1]: https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html#//apple_ref/doc/uid/TP40014097-CH41-ID5

eoger:
(Rebased against `main`)

uniffi/src/lib.rs
rfk: I am lightly alarmed that this did not cause any test failures...

rfk:
@eoger thanks, I've merged and squashed your feedback; if this turns out to be blocking you from landing https://github.com/mozilla/uniffi-rs/pull/243 and similar then please feel free to r+ and merge, and we can continue to iterate if other folks have more feedback over time.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
linacambridge: This is exactly what I had in mind with #29, thanks! 💯

Issue #226, opened 2020-08-10
Add map type support


eoger:
Just like sequences, but for associative arrays.

EDIT: renamed the issue to avoid confusion with our current dictionaries support.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1644)

eoger:
For what it's worth I don't see any support for this in the WebIDL spec

rfk:
> For what it's worth I don't see any support for this in the WebIDL spec

Aye. I think we are starting to bump up against the limits of WebIDL (which was always going to happen at *some* point, but TBH I'm a little surprised it's happened so quickly). I'm not sure there's a good way to parse something like `Map<Key, Value>` using weedle.

rfk:
Oh! It does have this `record` type for representing ordered maps:

* https://heycam.github.io/webidl/#idl-record

Which...I think it would be extremely confusing because uniffi turns WebIDL `dictionary` types into something called a `Record`, and we'd be turning the WebIDL `record` type into what many languages refer to as a `dictionary` 🤦

linacambridge:
There's also `maplike` which corresponds to a JS `Map`.

eoger:
That's pretty cool, it looks harder to implement though since it's an interface?

PR #243, opened 2020-08-13
Implement record type

eoger:
Fixes https://github.com/mozilla/uniffi-rs/issues/226.

There's some awkwardness in here, but it's not a shower-stopper imo.

### DIFF
--- a/examples/rondpoint/src/rondpoint.idl
+++ b/examples/rondpoint/src/rondpoint.idl
@@ -2,6 +2,7 @@ namespace rondpoint {
   Dictionnaire copie_dictionnaire(Dictionnaire d);
   Enumeration copie_enumeration(Enumeration e);
   sequence<Enumeration> copie_enumerations(sequence<Enumeration> e);
+  record<DOMString, Enumeration> copie_carte(record<DOMString, Enumeration> c);
   boolean switcheroo(boolean b);
 };
 
--- a/uniffi/src/lib.rs
+++ b/uniffi/src/lib.rs
@@ -28,8 +28,7 @@ use anyhow::{bail, Result};
 use bytes::buf::{Buf, BufMut};
 use ffi_support::ByteBuffer;
 use paste::paste;
-use std::convert::TryFrom;
-use std::ffi::CString;
+use std::{collections::HashMap, convert::TryFrom, ffi::CString};
 
 // It would be nice if this module was behind a cfg(test) guard, but it
 // doesn't work between crates so let's hope LLVM tree-shaking works well.
@@ -75,9 +74,9 @@ pub unsafe trait ViaFfi: Sized {
     /// serializing the data for transfer. In theory it could be possible to build a matching
     /// `#[repr(C)]` struct for a complex data type and pass that instead, but explicit
     /// serialization is simpler and safer as a starting point.
-    type Value;
+    type FfiType;
 
-    /// Lower a rust value of the target type, into an FFI value of type Self::Value.
+    /// Lower a rust value of the target type, into an FFI value of type Self::FfiType.
     ///
     /// This trait method is used for sending data from rust to the foreign language code,
     /// by (hopefully cheaply!) converting it into someting that can be passed over the FFI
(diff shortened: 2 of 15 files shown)


examples/rondpoint/src/rondpoint.idl
```diff
@@ -2,6 +2,7 @@ namespace rondpoint {
   Dictionnaire copie_dictionnaire(Dictionnaire d);
   Enumeration copie_enumeration(Enumeration e);
   sequence<Enumeration> copie_enumerations(sequence<Enumeration> e);
+  record<DOMString, Enumeration> copie_carte(record<DOMString, Enumeration> c);
```
eoger, line 5: - IDL record type is limited: the key *has* to be a string type, however the value can be anything.
- weedle will not parse `string`, so you have to declare a `DOMString` instead (it still maps properly to a rust `String`, we actually ignore the Key type).

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -124,6 +125,14 @@ mod filters {
...
                 write_kt(&"v", &"buf", t)?
             ),
+            Type::Map(t) => format!(
+                "lowerMap({}, {{ k, v -> {} + {} }}, {{ k, v, buf -> {}; {} }})",
+                nm,
+                calculate_write_size(&"k", &Type::String)?,
+                calculate_write_size(&"v", t)?,
+                write_kt(&"k", &"buf", &Type::String)?,
+                write_kt(&"v", &"buf", t)?
+            ),
```
eoger, line 135: 🤮

uniffi_bindgen/src/bindings/python/gen_python.rs
```diff
@@ -97,7 +97,13 @@ mod filters {
             Type::Enum(type_name) => format!("{} = {}({})", nm, type_name, nm),
             Type::Record(type_name) => format!("{} = {}._coerce({})", nm, type_name, nm),
             Type::Optional(t) => format!("(None if {} is None else {})", nm, coerce_py(nm, t)?),
-            Type::Sequence(t) => format!("({} for x in {})", coerce_py(&"x", t)?, nm), // TODO: name hygiene
+            Type::Sequence(t) => format!("({} for x in {})", coerce_py(&"x", t)?, nm), // TODO: name hygiene,
+            Type::Map(t) => format!(
```
eoger, line 101: I couldn't finish the python implementation, I think the other PR needs to land first.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -130,16 +130,16 @@ protocol Serializable {
 // Types confirming to `ViaFfi` can be transferred back-and-for over the FFI.
 // This is analogous to the Rust trait of the same name.
 protocol ViaFfi: Serializable {
-    associatedtype Value
-    static func lift(_ v: Value) throws -> Self
-    func lower() -> Value
+    associatedtype FfiType
```
eoger, line 133: It's unfortunate, but in the `extension Dictionary` below the Swift compiler couldn't disambiguate between `ViaFfi.Value` and `Dictionary.Value` in the `where` declaration

examples/rondpoint/src/rondpoint.idl
tarikeshaq: I think this is an okay start, and we can figure out other key types as we go. Kinda surprised `weedle` wouldn't parse `string`, but oh well.

uniffi/src/lib.rs
```diff
@@ -303,6 +302,46 @@ unsafe impl<T: ViaFfi> ViaFfi for Vec<T> {
...
+    }
+
+    fn try_lift(v: Self::Value) -> Result<Self> {
+        try_lift_from_bytebuffer(v)
+    }
+
+    fn write<B: BufMut>(&self, buf: &mut B) {
+        // TODO: would be nice not to panic here :-/
+        let len = u32::try_from(self.len()).unwrap();
+        buf.put_u32(len); // We limit HashMaps to u32::MAX bytes
```
tarikeshaq, line 325:
```suggestion
        buf.put_u32(len); // We limit HashMaps to u32::MAX items
```

rfk:
> Base automatically changed from clean-up-trait-method-names to main now

Oh, that is very slick, thankyou github!

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
rfk: lol, but also, I wonder if something similar to what I tried for python over in https://github.com/mozilla/uniffi-rs/pull/214 could be slightly less vomit-inducing here; thoughts? (for a future PR of course, not this one)

eoger: What I'm concerned about is that we would generate a lot of extra code for each new record/sequence/etc?

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -96,6 +96,21 @@ fun<T> readSequence(buf: ByteBuffer, readItem: (ByteBuffer) -> T): List<T> {
     }
 }
 
+fun<V> liftMap(rbuf: RustBuffer.ByValue, readItem: (ByteBuffer) -> Pair<String, V>): Map<String, V> {
+    return liftFromRustBuffer(rbuf) { buf -> readMap(buf, readItem) }
+}
+
+fun<V> readMap(buf: ByteBuffer, readItem: (ByteBuffer) -> Pair<String, V>): Map<String, V> {
+    val len = Int.read(buf)
+    @OptIn(ExperimentalStdlibApi::class)
```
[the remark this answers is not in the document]
eoger: It should result in a warning during compilation, if we find this unacceptable we can always switch to a more stable construct.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
rfk: Yeah, I guess it's similar to the use of monomorphization in rust, where you rust generating lots of almost-duplicate code. (I don't think it would balloon too badly in practice, because you'd only get one method per type-you-actually-use-in-a-sequence, not one for every single type. But, worth watching out for).

PR #247, opened 2020-08-14
Allows dynamic error messages

### LINKED ISSUE #245: Richer error types
Right now we only allow for error enums that have a flat structure. This is okay for a simple project, but we often want to report some dynamic information (what was the network error? what was the db error? etc).

 Currently, we have no good way of doing that, except requiring the consumer to define more errors.

This is also related to enums overall since they have the same limitation, but the specific error use case seems more important.

Thoughts? \cc @eoger @rfk @linacambridge

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1666)

tarikeshaq:
fixes #245 

Forces errors variants to have an associate type

i.e now all error variants must have some internal thing they are wrapping, example:

```rust
enum Error {
    NetworkError, // This is an invalid variant, will not compile
    ExperimentError(String), // Valid
    DbError(rusqlite::Error), // This is also valid
    OtherError {
       message: String
     } // Invalid will not compile
}
```

I don't like this too much. 

I would prefer it if we were able to let the consumer decide what each variant looks like.

In order to do that we need a way to match against an enum variant's name, regardless if it has an associated type, is a struct, or is just a variant with no other data. Seems like a reasonable thing, except I couldn't for the life of me find a way to do that. What I mean is something like this for the above struct:
```rust
match err {
  Error::OtherError => ..... // Gives a compiler error, since `OtherError` is a struct variant, and we need to do `Error::OtherError{message:_}`
}
```

We could force the consumer to annotate each variant in the `idl` but not sure if that's too much 😅 

Happy to ditch this if we can think of a nice way to do this.

### DIFF
--- a/uniffi_bindgen/src/templates/ErrorTemplate.rs
+++ b/uniffi_bindgen/src/templates/ErrorTemplate.rs
@@ -12,7 +12,7 @@ impl From<{{e.name()}}> for uniffi::deps::ffi_support::ExternError {
         // we might find that we need to reserve some codes.
         match err {
             {%- for value in e.values() %}
-            {{ e.name()}}::{{value}} => uniffi::deps::ffi_support::ExternError::new_error(uniffi::deps::ffi_support::ErrorCode::new({{ loop.index }}), err.to_string()),
+            {{ e.name()}}::{{value}}{..} => uniffi::deps::ffi_support::ExternError::new_error(uniffi::deps::ffi_support::ErrorCode::new({{ loop.index }}), err.to_string()),
             {%- endfor %}
         }
     }
--- a/examples/arithmetic/src/lib.rs
+++ b/examples/arithmetic/src/lib.rs
@@ -4,16 +4,18 @@
 
 #[derive(Debug, thiserror::Error)]
 enum ArithmeticError {
-    #[error("Integer overflow!")]
-    IntegerOverflow,
+    #[error("Integer overflow on an operation with {a} and {b}")]
+    IntegerOverflow { a: u64, b: u64 },
 }
 
 fn add(a: u64, b: u64) -> Result<u64> {
-    a.checked_add(b).ok_or(ArithmeticError::IntegerOverflow)
+    a.checked_add(b)
+        .ok_or(ArithmeticError::IntegerOverflow { a, b })
 }
 
 fn sub(a: u64, b: u64) -> Result<u64> {
-    a.checked_sub(b).ok_or(ArithmeticError::IntegerOverflow)
+    a.checked_sub(b)
+        .ok_or(ArithmeticError::IntegerOverflow { a, b })
 }
 
 fn equal(a: u64, b: u64) -> bool {
--- a/examples/todolist/src/lib.rs
+++ b/examples/todolist/src/lib.rs
@@ -15,14 +15,18 @@ enum TodoError {
     EmptyTodoList,
     #[error("That todo already exists!")]
     DuplicateTodo,
-    #[error("Empty String error!")]
-    EmptyString,
+    #[error("Empty String error!: {0}")]
+    EmptyString(String),
+    #[error("I am a delegated Error: {0}")]
+    DeligatedError(#[from] std::io::Error),
 }
 
 fn create_entry_with<S: Into<String>>(item: S) -> Result<TodoEntry> {
     let text = item.into();
     if text == "" {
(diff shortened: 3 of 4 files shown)


uniffi_bindgen/src/templates/ErrorTemplate.rs
```diff
@@ -12,7 +12,7 @@ impl From<{{e.name()}}> for uniffi::deps::ffi_support::ExternError {
         // we might find that we need to reserve some codes.
         match err {
             {%- for value in e.values() %}
-            {{ e.name()}}::{{value}} => uniffi::deps::ffi_support::ExternError::new_error(uniffi::deps::ffi_support::ErrorCode::new({{ loop.index }}), err.to_string()),
+            {{ e.name()}}::{{value}}(_) => uniffi::deps::ffi_support::ExternError::new_error(uniffi::deps::ffi_support::ErrorCode::new({{ loop.index }}), err.to_string()),
```
tarikeshaq, line 15: This is the only real change, the rest is just test stuff.

rfk:
Thanks for pushing on this, @tarikeshaq! I have...well, a lot of half-formed thoughts on what's going on here, and a vague sense that we're missing something.

My understanding from your description is:
* We want to allow the rust code to customize `err.to_string` in order to surface a customized error message to the consuming code
* In order to do this, we need to allow the error variants to contain data, instead of just being labels.
* But because of the way we generate code, is means that *every single enum variant* has to have *exactly one* piece of member data.

Is that accurate?

One option here could be to use rusts "rest" pattern `..` to match on an arbitrary number of fields. This seems to work when testing it out on the rust playground:

```
enum E {
    Zero(),
    One(u32),
    Two(u64, u64),
}

impl E {
    fn printme(&self) {
        match self {
            E::Zero(..) => println!("zero"),
            E::One(..) => println!("one"),
            E::Two(..) => println!("two"),
        };
    }
}

fn main() {
    E::Zero().printme();
    E::One(1).printme();
    E::Two(2, 2).printme();
}
```

The only slight annoyance here is the zero-valued variant needs to be constructed using `E::Zero()` rather than `E::Zero`, and gives a pretty unhelpful error message if you try to use the latter.

> We could force the consumer to annotate each variant in the idl but not sure if that's too much 😅

Yeah, I feel like this is one of the places where we're quickly butting up against the mis-match between our needs and the design-space of WebIDL (which I guess shouldn't come as too much of a surprise).

Would this be easier if we were using macros, or a custom IDL, rather than confining ourselves to the limited enum types offered by WebIDL? Do we need it urgently, or could it wait until post-MVP when we might have a chance to revisit that?

tarikeshaq:
Thanks, @rfk!!
> Is that accurate?

Yes, spot on! Before this PR, all our variants were only labels, with no data.

I'm happy to use the `rest` pattern, for now, I think it provides us with all the context we need to propagate good error messages. We can tackle any large changes (macros, custom webidl, etc) in the future post-MVP as we consider those alternatives.

uniffi_bindgen/src/templates/ErrorTemplate.rs
```diff
@@ -12,7 +12,7 @@ impl From<{{e.name()}}> for uniffi::deps::ffi_support::ExternError {
         // we might find that we need to reserve some codes.
         match err {
             {%- for value in e.values() %}
-            {{ e.name()}}::{{value}} => uniffi::deps::ffi_support::ExternError::new_error(uniffi::deps::ffi_support::ErrorCode::new({{ loop.index }}), err.to_string()),
+            {{ e.name()}}::{{value}}(..) => uniffi::deps::ffi_support::ExternError::new_error(uniffi::deps::ffi_support::ErrorCode::new({{ loop.index }}), err.to_string()),
```
linacambridge, line 15: If you change the `(..)` to `{ .. }`, I think you can support both plain and wrapped variants! 😄🎉

tarikeshaq: I swear my jaw just dropped. This is exactly what I was looking for 🎉 😄 🚀 🕺

rfk: Boom! Nice one @linacambridge 😁 

(I don't have a good mental model of *why* that syntax works, because in my head `{}` is for struct literals so I didn't expect it to work with the tuple-like variant..but I love it!)

PR #255, opened 2020-08-27
Generate Firefox Desktop front-end bindings with WebIDL

linacambridge:
This PR is a first cut at a UniFFI backend for Firefox! It's a work in progress, but can generate working bindings now, so I wanted to get it up for review. I'd love your comments! 💜

The backend works a little differently than the others. Firefox expects one header file per interface (and a separate one for the namespace that holds top-level functions), so UniFFI emits multiple `.h` and `.cpp` files. There's also complexity around handling different types. Some are reflected differently depending on whether they're arguments, type parameters, or local variables; some are returned directly and others are passed as out parameters; and the WebIDL implementation signatures have extra leading and trailing arguments. All this makes `macros.cpp` pretty gnarly 🙀

`SharedHeaderTemplate.h` contains the `RustBuffer` helpers and serialization logic. The ideas are the same as the Swift, Kotlin, and Python backends, but the implementation is a pile of C++ code. I'll add some more comments throughout it in the coming days, so that it's a bit easier to digest. This part can also be reused wholesale for our pure-C++ Nimbus interface, there's not much that's WebIDL specific here.

# Getting this building

The setup for this is a little janky for now, but here's how to get this running! You'll need a full (not artifact) build of Firefox, since this touches lots of C++ and Rust code.

1. `git checkout gecko` to switch to the Gecko branch.
2. Make a new folder for the bindings: `mkdir rondpoint-gecko`
3. Generate the Gecko bindings for Rondpoint: `cd uniffi_bindgen; and cargo run -- generate ../examples/rondpoint/src/rondpoint.idl -l gecko -o ../rondpoint-gecko`
4. Symlink the generated WebIDL file into `dom/chrome-webidl` in m-c: `ln -s /uniffi-rs/rondpoint-gecko/Rondpoint.webidl /gecko/dom/chrome-webidl/Rondpoint.webidl`
5. Add `Rondpoint.webidl` to `dom/chrome-webidl/moz.build`, in the `WEBIDL_FILES` section.
6. Symlink the `rondpoint-gecko` folder from step (2) into an existing (or new!) directory in m-c that has a `moz.build` file. Any one will do; I picked `toolkit/components/extensions/storage` for no other reason than it's a pretty small directory and makes it easy to do directory-specific builds.
7. Symlink `uniffi-rs` into m-c.
8. Add `uniffi-example-rondpoint` as a dependency to `toolkit/library/rust/shared/Cargo.toml`, as a `path` dependency inside the symlink from step (7).
9. Add `extern crate uniffi_rondpoint` to `toolkit/library/rust/shared/lib.rs`.
10. Run `mach vendor rust`.
11. Add the generated sources and xpcshell test manifest to the `moz.build` in the folder from step (6), like this:

```py
EXPORTS.mozilla.dom += [
    'rondpoint-gecko/Retourneur.h',
    'rondpoint-gecko/Rondpoint.h',
    'rondpoint-gecko/RondpointShared.h',
    'rondpoint-gecko/Stringifier.h',
]

UNIFIED_SOURCES += [
    'rondpoint-gecko/Retourneur.cpp',
    'rondpoint-gecko/Rondpoint.cpp',
    'rondpoint-gecko/Stringifier.cpp',
]

XPCSHELL_TESTS_MANIFESTS += [
    'uniffi-rs/examples/rondpoint/tests/gecko/xpcshell.ini'
]
```

And then run `mach build`, and you should be good to go! You can run `mach xpcshell-test toolkit/components/extensions/storage/uniffi-rs/examples/rondpoint/tests/gecko/xpcshell.ini` after the build is done! 😊

# Hacks

I had to make two changes to the `Cargo.toml`s in UniFFI to get this building in Firefox, which we'll want to fix before merging:

1. Remove the top-level `Cargo.toml`, so that I could symlink `uniffi-rs` into my m-c tree and add `uniffi-example-rondpoint` as a `path` dependency. I'm not sure this is necessary; if we can avoid symlinking `uniffi-rs` into m-c, we can do without.
2. Remove the `cargo_metadata` dependency. When I tried building Firefox with the version of `cargo_metadata` from Crates.io, I got a panic in one of its `derive` macros. Looks lik that was fixed in oli-obk/cargo_metadata#123 by bumping the Serde version, but that hasn't been published yet, so `mach vendor rust` doesn't know to pull the new version of Serde in. But changing `cargo_metadata` to a Git dependency confused Cargo: `mach vendor rust` didn't vendor _any_ of UniFFI's dependencies, and then `mach build` complained that it couldn't resolve the versions properly. I'm not sure what's going on there, and didn't dig too much more into it.

I also had to change the `crate-type` to `["cdylib", "lib"]`, so that we could link the crate into `gkrust`. If it's just `cdylib`, `extern crate` won't find it.


uniffi_bindgen/src/bindings/gecko/templates/WebIDLTemplate.webidl
```diff
@@ -0,0 +1,67 @@
...
+{% endfor %}
+
+{%- let functions = ci.iter_function_definitions() %}
+{%- if !functions.is_empty() %}
+[ChromeOnly, Exposed=Window]
+namespace {{ ci.namespace()|class_name_webidl }} {
+  {#-
+  // We'll need to figure out how to handle async methods. One option is
+  // to declare them as `async foo()`, or an `[Async]` or `[BackgroundThread]`
+  // attribute in the UniFFI IDL. Kotlin, Swift, and Python can ignore that
```
rfk, line 27: Kotlin might like to automatically dispatch such methods via coroutine or similar, I think this could be worthwhile in more contexts than just Gecko.

linacambridge: Ooooh, yes, great point! And maybe Swift can use a dispatch queue internally, and Python could use `asyncio`! 🚀

For the Nimbus pure-C++ API, we can punt on figuring out the exact details, and leave it up to a wrapper to initialize it and make network requests on a background task queue, just like we do in a-c and Swift now. But it’s definitely worth exploring this, and we’ll certainly want to have a story here before we transition the Nimbus JS SDK to Rust.

uniffi_bindgen/src/bindings/gecko/gen_gecko.rs
```diff
@@ -0,0 +1,355 @@
...
+            Type::Sequence(_) => ReturnPosition::OutParam(type_),
+            _ => ReturnPosition::Return(type_),
+        }
+    }
+}
+
+/// Returns a dummy empty value for the given type. This is used to
+/// implement the `cpp::bail` macro to return early if a WebIDL function or
+/// method throws an exception.
+pub fn default_return_value_cpp(type_: &Type) -> Option<String> {
```
linacambridge, line 55: It’s unfortunate that this has to exist. Basically, if you want to throw an error from inside a WebIDL method with a signature like `int32_t Foo(ErrorResult& aRv)`, you can call `aRv.Throw(...)` in the method, but you still need to return _some_ value—even though it’ll get ignored by the bindings code. Methods that return results via out params don’t need to do this; they can just do `return;` and not set the param.

uniffi_bindgen/src/bindings/gecko/gen_gecko.rs
```diff
@@ -0,0 +1,355 @@
...
+
+impl<'config, 'ci> WebIdl<'config, 'ci> {
+    pub fn new(config: &'config Config, ci: &'ci ComponentInterface) -> Self {
+        Self { config: config, ci }
+    }
+}
+
+/// A shared header file that's included by all our bindings. This defines
+/// common serialization logic and `extern` declarations for the FFI. Note that
+/// the bindings always include this header file, never the other way around.
```
linacambridge, line 94: ...Because otherwise, we’d get a circular dependency when the bindings layer includes one of our interface headers in its header or implementation.

uniffi_bindgen/src/bindings/gecko/gen_gecko.rs
```diff
@@ -0,0 +1,355 @@
...
+        Self { config: config, ci }
+    }
+}
+
+/// A header file generated for a namespace with top-level functions.
+#[derive(Template)]
+#[template(syntax = "c", escape = "none", path = "NamespaceHeaderTemplate.h")]
+pub struct NamespaceHeader<'config, 'ci, 'functions> {
+    config: &'config Config,
+    ci: &'ci ComponentInterface,
```
linacambridge, line 113: It may be worth having these structs hold just the pieces they need from the `ComponentInterface`, like the namespace name, instead of the full interface _and_ the specific thing they’re rendering.

uniffi_bindgen/src/bindings/gecko/gen_gecko.rs
```diff
@@ -0,0 +1,355 @@
...
+            // Note: Not `unrestricted double`; we don't want to allow NaNs
+            // and infinity.
+            Type::Float64 => "double".into(),
+            Type::Boolean => "boolean".into(),
+            Type::String => "DOMString".into(),
+            Type::Enum(name) | Type::Record(name) | Type::Object(name) => class_name_webidl(name)?,
+            Type::Error(name) => panic!("[TODO: type_webidl({:?})]", type_),
+            Type::Optional(type_) => format!("{}?", type_webidl(type_)?),
+            Type::Sequence(type_) => format!("sequence<{}>", type_webidl(type_)?),
+            Type::Map(type_) => format!("record<DOMString, {}>", type_webidl(type_)?),
```
linacambridge, line 219: Terminology clarification that probably ought to go in a comment: what UniFFI calls a “record”, Firefox calls a “dictionary”. What UniFFI calls a “map”, Firefox calls a “record”. What UniFFI calls an “optional”, Firefox calls “nullable” (“optional” means something different in WebIDL/Firefox).

uniffi_bindgen/src/bindings/gecko/gen_gecko.rs
```diff
@@ -0,0 +1,355 @@
...
+            FFIType::UInt16 => "uint16_t".into(),
+            FFIType::Int32 => "int32_t".into(),
+            FFIType::UInt32 => "uint32_t".into(),
+            FFIType::Int64 => "int64_t".into(),
+            FFIType::UInt64 => "uint64_t".into(),
+            FFIType::Float32 => "float".into(),
+            FFIType::Float64 => "double".into(),
+            FFIType::RustBuffer => "RustBuffer".into(),
+            FFIType::RustString => "char*".into(),
+            FFIType::RustError => "NativeRustError".into(),
```
linacambridge, line 238:
```suggestion
            FFIType::RustError => "RustError".into(),
```

uniffi_bindgen/src/bindings/gecko/gen_gecko.rs
```diff
@@ -0,0 +1,355 @@
...
+            FFIType::ForeignStringRef => "const char*".into(),
+        })
+    }
+
+    pub fn ret_type_ffi(type_: &Type) -> Result<String, askama::Error> {
```
linacambridge, line 243: Needs a comment—this is used in `macros.cpp`, to generate the specialization declaration for lifting return values from the FFI call, and for declaring the return type of an FFI call.

uniffi_bindgen/src/bindings/gecko/gen_gecko.rs
```diff
@@ -0,0 +1,355 @@
...
+            Type::Int16 => "int16_t".into(),
+            Type::UInt16 => "uint16_t".into(),
+            Type::Int32 => "int32_t".into(),
+            Type::UInt32 => "uint32_t".into(),
+            Type::Int64 => "int64_t".into(),
+            Type::UInt64 => "uint64_t".into(),
+            Type::Float32 => "float".into(),
+            Type::Float64 => "double".into(),
+            Type::Boolean => "bool".into(),
+            Type::String => "nsAString".into(),
```
linacambridge, line 286: This isn’t conceptually right (it should be `nsString`), but it _is_ the type we need to specialize `ViaFfi::Lift` and `Lower`. That might need to go into its own method...

And also, `type_cpp` needs a better name—something that captures this is the type expressed in a type parameter or when declaring a value.

uniffi_bindgen/src/bindings/gecko/gen_gecko.rs
```diff
@@ -0,0 +1,355 @@
...
+        Ok(nm.to_string().to_camel_case())
+    }
+
+    pub fn field_name_cpp(nm: &str) -> Result<String, askama::Error> {
+        Ok(format!("m{}", nm.to_camel_case()))
+    }
+
+    pub fn enum_variant_cpp(nm: &dyn fmt::Display) -> Result<String, askama::Error> {
+        // TODO: Make sure this does the right thing for hyphenated variants.
+        // Example: "bookmark-added" becomes `Bookmark_added`.
```
linacambridge, line 352: Let’s add a test case for hyphenated enum variants to UniFFI, and verify it gets reflected correctly in the binding layer.

uniffi_bindgen/src/bindings/gecko/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+pub struct Bindings {
+    webidl: String,
+    shared_header: String,
+    sources: Vec<Source>,
+}
+
+/// Generate uniffi component bindings for Gecko.
+///
+/// Bindings to a Rust interface for Gecko involves more than just generating a
+/// `.cpp` file.
```
linacambridge, line 36: Let’s move the comment about `Codegen.py` out of `generate_bindings` and up here.

uniffi_bindgen/src/bindings/gecko/templates/InterfaceHeaderTemplate.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include "mozilla/dom/{{ ci.namespace()|class_name_webidl }}Shared.h"
+
+namespace mozilla {
+namespace dom {
+
+class {{ obj.name()|class_name_cpp }} final : public nsISupports, public nsWrapperCache {
+ public:
+  NS_DECL_CYCLE_COLLECTING_ISUPPORTS
+  NS_DECL_CYCLE_COLLECTION_SCRIPT_HOLDER_CLASS({{ obj.name()|class_name_cpp }})
```
linacambridge, line 24: Technically, since we only ever return new objects, we might not need cycle collection...but we do need to cycle-collect `mGlobal` if we hold onto it, and we may need it once we start returning references, _and_ also all this might go away and get replaced with XPIDL—so for now, copypasta it is! 🍝😁

uniffi_bindgen/src/bindings/gecko/templates/InterfaceTemplate.cpp
```diff
@@ -0,0 +1,97 @@
...
+  auto handle = {{ cons.ffi_func().name() }}(
+    {%- let args = cons.arguments() %}
+    {% call cpp::to_ffi_args(args) -%}
+    {%- if cons.throws().is_some() %}
+    {% if !args.is_empty() %},{% endif %}&err
+    {% endif %}
+  );
+  {%- if cons.throws().is_some() %}
+  if (err.mCode) {
+    aRv.ThrowOperationError(err.mMessage);
```
linacambridge, line 59: Error types are fun. Firefox has a hard-coded list of DOM exceptions (there aren’t that many! And I think the guidance for WebIDL APIs is to use a limited set of built-in exceptions, like `AbortError`, `InvalidStateError`, etc.) that can be thrown on an `ErrorResult&`...which is kind of at odds with how we expose errors. We might need to use `jsapi` here to make a custom error object, and then call `aRv.ThrowJSException`.

Propagating errors with all the right type information is still a big `TODO` for this PR.

uniffi_bindgen/src/bindings/gecko/templates/InterfaceTemplate.cpp
```diff
@@ -0,0 +1,97 @@
...
+  return result.forget();
+}
+{%- endfor %}
+
+{% for meth in obj.methods() %}
+{% call cpp::decl_return_type(meth) %} {{ obj.name()|class_name_cpp }}::{{ meth.name()|fn_name_cpp }}(
+  {% call cpp::decl_method_args(meth) %}
+) {
+  if (mHandle == -1) {
+    aRv.ThrowOperationError("Can't use destroyed handle");
```
linacambridge, line 74: Every method has a `[Throws]` annotation because it can throw one of these internal errors. Let’s drop this check (and setting the handle to `-1`) entirely. I think the `ConcurrentHandleMap` will fail any calls that try to use a destroyed handle, and this situation should never happen, anyway—when the destructor for a WebIDL object runs, it’s guaranteed that JS has GC’ed it.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,795 @@
...
+    {%- endif %}
+);
+
+{% endfor -%}
+
+}  // extern "C"
+
+namespace mozilla {
+namespace dom {
+// namespace {{ ci.namespace() }} {
```
linacambridge, line 55: It would be nice if we could namespace the UniFFI bindings, but Firefox’s binding layer specifically expects all interface implementations to live in `mozilla::dom`.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,795 @@
...
+  /// closure takes a `Span` pointing to the raw bytes, which it can use to
+  /// copy the bytes into an `nsCString` or `nsString`.
+  ///
+  /// Safety: The closure must copy the span's contents into a new owned string.
+  /// It must not hold on to the span, as its contents will be invalidated when
+  /// the backing Rust byte buffer is freed. It must not call any other methods
+  /// on the reader.
+  template <typename T>
+  void ReadRawString(
+      T& aString,
```
linacambridge, line 125: Using an out parameter here because `AString` types (`nsACString` and `nsAString`) can’t be returned by value.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,795 @@
...
+                       aNewOffset.value() <= mBuffer.Capacity());
+  }
+
+  template <typename T>
+  void WriteAt(const T& aValue,
+               const std::function<void(size_t, const T&)>& aClosure) {
+    CheckedInt<size_t> newOffset = mOffset;
+    newOffset += sizeof(T);
+    AssertInBounds(newOffset);
+    mBuffer.SetLength(newOffset.value());
```
linacambridge, line 284: Add a comment: we’ve reserved the capacity in `mBuffer` to hold the entire object, so this won’t allocate, but we still need to update the length or we’ll get an out-of-bounds exception indexing into the array.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,795 @@
...
+
+/// A "trait" with specializations for types that can be transferred back and
+/// forth over the FFI. This is analogous to the Rust trait of the same name.
+/// As above, this gives us compile-time type checking for type pairs. If
+/// `ViaFfi<T, U>::Lift(U, T)` compiles, we know that a value of type `U` from
+/// the FFI can be lifted into a value of type `T`.
+template <typename T, typename FfiType>
+struct ViaFfi {
+  static bool Lift(const FfiType& aLowered, T& aLifted) = delete;
+  static FfiType Lower(const T& aLifted) = delete;
```
linacambridge, line 317: This returns `FfiType` instead of using an out param because we call `ViaFfi<T, U>::Lower(T)` for each argument in the argument list when calling an FFI function. An out param would mean we need to allocate some temporary stack variables to hold them, and I’ve nooo idea how to express all that in Askama 🤯 Fortunately, FFI types are all primitives, so there’s no fancy copy constructor or move shenanigans going on here, as would be the case with `Lift`.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,795 @@
...
+struct Serializable<nsACString> {
+  static size_t Size(const nsACString& aValue) {
+    CheckedInt<size_t> size(aValue.Length());
+    size += sizeof(uint32_t);  // For the length prefix.
+    MOZ_RELEASE_ASSERT(size.isValid());
+    return size.value();
+  }
+
+  static bool ReadFrom(Reader& aReader, nsACString& aValue) {
+    aValue.Truncate();
```
linacambridge, line 402: Just in case the out param has garbage in it.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,795 @@
...
+    aLifted.Truncate();
+    aLifted.Append(MakeStringSpan(aLowered));
+    return true;
+  }
+
+  static char* Lower(const nsACString& aLifted) {
+    RustError err = {0, nullptr};
+    char* lowered = {{ ci.ffi_string_alloc_from().name() }}(aLifted.BeginReading(), &err);
+    if (err.mCode) {
+      MOZ_ASSERT(false, "Failed to copy narrow string to Rust string");
```
linacambridge, line 430: `MOZ_ASSERT` doesn’t fire in release (`MOZ_RELEASE_ASSERT` does, though), which seems fine for things that aren’t security-sensitive—Rust will take care of crashing the app if it gets a `nullptr` for a string, or a weird serialization.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,795 @@
...
+    return StringTraits<nsAString>::ReadFrom(aReader, aValue);
+  }
+
+  static void WriteInto(Writer& aWriter, const nsAString& aValue) {
+    StringTraits<nsAString>::WriteInto(aWriter, aValue);
+  }
+};
+
+template <>
+struct ViaFfi<nsAString, char*> {
```
linacambridge, line 491: Technically, we can specialize `ViaFfi` for `nsString`, too—I didn’t because we only lift `char *`s into `nsAString&` out params, not `nsString`s on the heap or `nsAutoString`s on the stack.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,795 @@
...
+
+  static char* Lower(const nsAString& aLifted) {
+    // Encode the string to UTF-8, then make a Rust string from the contents.
+    // This copies the string twice, but is safe.
+    nsAutoCString utf8;
+    CopyUTF16toUTF8(aLifted, utf8);
+    RustError err = {0, nullptr};
+    char* lowered = {{ ci.ffi_string_alloc_from().name() }}(utf8.BeginReading(), &err);
+    if (err.mCode) {
+      MOZ_ASSERT(false, "Failed to copy wide string to Rust string");
```
linacambridge, line 506:
```suggestion
      MOZ_ASSERT(false, "Failed to make Rust string from nsAString");
```

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,795 @@
...
+    CheckedInt<size_t> size(1);
+    size += Serializable<T>::Size(aValue.Value());
+    MOZ_RELEASE_ASSERT(size.isValid());
+    return size.value();
+  }
+
+  static bool ReadFrom(Reader& aReader, dom::Nullable<T>& aValue) {
+    uint8_t hasValue = aReader.ReadUInt8();
+    if (hasValue != 0 && hasValue != 1) {
+      return false;
```
linacambridge, line 554: Assert about the unexpected tag here...

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,795 @@
...
+  static bool Lift(const RustBuffer& aLowered, T& aLifted) {
+    auto reader = Reader(aLowered);
+    if (!Serializable<T>::ReadFrom(reader, aLifted)) {
+      return false;
+    }
+    if (reader.HasRemaining()) {
+      MOZ_ASSERT(false);
+      return false;
+    }
+    {{ ci.ffi_bytebuffer_free().name() }}(aLowered);
```
linacambridge, line 701: Freeing a `const RustBuffer&` seems sketchy, even though it’s not actually `const`. Since we’re not inheriting from `Serializable`, just specializing, we can probably change this signature to `Lift(RustBuffer&)`, and the compiler will pick this overload instead of the deleted function in the base template.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,795 @@
...
+};
+
+/// A "trait" struct with specializations for types that can be read and
+/// written into a byte buffer. This struct is specialized for all serializable
+/// types.
+template <typename T>
+struct Serializable {
+  /// Returns the size of the serialized value, in bytes. This is used to
+  /// calculate the allocation size for the Rust byte buffer.
+  static size_t Size(const T& aValue) = delete;
```
linacambridge, line 300: I wonder if we should have this return a `CheckedInt<size_t>`, and only do the unwrapping and `MOZ_RELEASE_ASSERT(size.isValid())` once, when we actually allocate the buffer.

uniffi_bindgen/src/bindings/gecko/templates/WebIDLTemplate.webidl
```diff
@@ -0,0 +1,67 @@
...
+  {% endif %}
+  constructor(
+      {%- for arg in cons.arguments() %}
+      {{ arg.type_()|type_webidl }} {{ arg.name() }}{%- if !loop.last %}, {% endif %}
+      {%- endfor %}
+  );
+  {%- endfor %}
+
+  {% for meth in obj.methods() -%}
+  [Throws]
```
linacambridge, line 59: See note above; get rid of the `mHandle != -1` check and then make this conditional on `method.throws()`, just like the others.

uniffi_bindgen/src/bindings/gecko/templates/macros.cpp
```diff
@@ -0,0 +1,120 @@
...
+    {%- when Some with (type_) -%}
+      {%- match ReturnPosition::for_type(type_) -%}
+        {%- when ReturnPosition::Return with (type_) -%} {{ type_|ret_type_cpp }}
+        {%- when ReturnPosition::OutParam with (_) -%} void
+      {%- endmatch -%}
+    {%- else -%} void
+  {%- endmatch -%}
+{%- endmacro -%}
+
+{# /* Declares a list of arguments for a WebIDL constructor. A constructor takes
```
linacambridge, line 26: These lists look quite hideous in the generated code, but I haven’t figured out the right mix of `%}` and `-%}` to make Askama strip the right amount of whitespace so that argument lists look pretty. The fixed arguments, like `GlobalObject`, and inserting the right number of commas make this tricky. They are readable as they are, so maybe that’s good enough?

uniffi_bindgen/src/bindings/gecko/templates/macros.cpp
```diff
@@ -0,0 +1,120 @@
...
+  {%- for arg in args -%}
+  {{ arg.type_()|arg_type_cpp }} {{ arg.name() -}}{%- if !loop.last -%},{%- endif -%}
+  {%- endfor -%}
+  {%- call _decl_out_param(func) -%}
+  {%- match func.return_type() -%}
+    {%- when Some with (type_) -%}
+      {%- match ReturnPosition::for_type(type_) -%}
+        {%- when ReturnPosition::OutParam with (type_) -%},
+        {%- else -%}
+          {%- if !args.is_empty() %}, {% endif -%}
```
linacambridge, line 72: All this just to figure out if we need a trailing comma or not... 😭

uniffi_bindgen/src/bindings/gecko/templates/macros.cpp
```diff
@@ -0,0 +1,120 @@
...
+{# /* Returns a result from a function or method. This lifts the result from the
+      given `var`, and returns it by value or via an "out param" depending on
+      the function's return type. */ #}
+{%- macro return(func, var) -%}
+{% match func.return_type() -%}
+{%- when Some with (type_) -%}
+  {% match ReturnPosition::for_type(type_) -%}
+  {%- when ReturnPosition::OutParam with (type_) -%}
+  DebugOnly<bool> ok_ = detail::ViaFfi<{{ type_|type_cpp }}, {{ type_|ret_type_ffi }}>::Lift({{ var }}, aRetVal_);
+  MOZ_ASSERT(ok_);
```
linacambridge, line 89: We should have already asserted before returning `false` in the serializers—maybe this deserves a `MOZ_RELEASE_ASSERT`?

examples/rondpoint/tests/gecko/test_rondpoint.js
```diff
@@ -0,0 +1,141 @@
+add_task(async function test_rondpoint() {
```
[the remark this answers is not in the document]
nnethercote: What is `add_task`? I can see it used in this file, but I can't see it anywhere else.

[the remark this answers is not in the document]
mhammond: I imagine the idea is to run these tests using `mach xpcshell-test` somehow - no idea what process lina has in mind for having mach understand them though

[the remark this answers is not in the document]
linacambridge: Yep, this file is an xpcshell test!

> no idea what process lina has in mind for having mach find them though

Haha, my workflow so far has been “symlink `rondpoint` into m-c and add it to `moz.build` with all the other stuff”, but that’s not going to fly for anything more than a demo. Still, if you want to take it for a spin locally, there are only 11 easy steps! 😆

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,794 @@
...
+#include "mozilla/Utf8.h"
+
+#include "mozilla/dom/BindingDeclarations.h"
+#include "mozilla/dom/Record.h"
+#include "mozilla/dom/{{ ci.namespace()|class_name_webidl }}Binding.h"
+
+extern "C" {
+
+struct RustBuffer {
+  int64_t mLen;
```
nnethercote, line 24: @rfkelly was talking about changing buffer lengths to `int32_t`.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,794 @@
...
+ public:
+  explicit Reader(const RustBuffer& aBuffer) : mBuffer(aBuffer), mOffset(0) {}
+
+  /// Indicates if the offset has reached the end of the buffer.
+  bool HasRemaining() {
+    return static_cast<int64_t>(mOffset.value()) < mBuffer.mLen;
+  }
+
+  /// Helpers to read fixed-width primitive types at the current offset.
+  /// Fixed-width integers are read in big endian order.
```
nnethercote, line 71: I always find endianness tricky to think about, and I like to document things carefully. With that in mind, for me this comment raises as many questions as it answers. How/why are they read in big-endian order? Is big-endian used for all serialization in `uniffi-rs`? Some more explanation would be helpful.

(It might not hurt to mention that "network order" == "big-endian"... I had to google that to remind myself.)

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
nnethercote: This seems like a good comment to put in a comment :)

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,794 @@
...
+  }
+
+  void WriteInt8(const int8_t& aValue) {
+    WriteUInt8(BitwiseCast<uint8_t>(aValue));
+  }
+
+  // This code uses `memcpy` and other eye-twitchy patterns because it
+  // originally wrote values directly into a `RustBuffer`, instead of
+  // an intermediate `nsTArray`. Once #251 is fixed, we can return to
+  // doing that, and remove `ToRustBuffer`.
```
nnethercote, line 174: The `Reader` code also uses `memcpy` but it doesn't have a comment like this. Should it?

Also, there should be a mention of big-endian encoding to match the one in `Reader`.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,794 @@
...
+  /// closure takes a `Span` pointing to the byte buffer, which it should fill
+  /// with bytes and return the actual number of bytes written.
+  ///
+  /// This function is (more than a little) convoluted. It's written this way
+  /// because we want to support UTF-8 and UTF-16 strings. The "size hint" is
+  /// the maximum number of bytes that the closure can write. For UTF-8 strings,
+  /// this is just the length. For UTF-16 strings, which must be converted to
+  /// UTF-8, this can be up to three times the length. Once the closure tells us
+  /// how many bytes it's actually written, we can write the length prefix, and
+  /// advance the current offset.
```
nnethercote, line 227: Why does `ReadRawString` only support UTF-8 when this supports UTF-8 and UTF-16?

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
nnethercote: Another good thing to put in a comment.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,794 @@
...
+};
+
+/// A "trait" struct with specializations for types that can be read and
+/// written into a byte buffer. This struct is specialized for all serializable
+/// types.
+template <typename T>
+struct Serializable {
+  /// Returns the size of the serialized value, in bytes. This is used to
+  /// calculate the allocation size for the Rust byte buffer.
+  static CheckedInt<size_t> Size(const T& aValue) = delete;
```
nnethercote, line 299: Why is the return type `CheckedInt`?

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,794 @@
...
+UNIFFI_SPECIALIZE_SERIALIZABLE_PRIMITIVE(uint16_t, ReadUInt16, WriteUInt16);
+UNIFFI_SPECIALIZE_SERIALIZABLE_PRIMITIVE(int16_t, ReadInt16, WriteInt16);
+UNIFFI_SPECIALIZE_SERIALIZABLE_PRIMITIVE(uint32_t, ReadUInt32, WriteUInt32);
+UNIFFI_SPECIALIZE_SERIALIZABLE_PRIMITIVE(int32_t, ReadInt32, WriteInt32);
+UNIFFI_SPECIALIZE_SERIALIZABLE_PRIMITIVE(uint64_t, ReadUInt64, WriteUInt64);
+UNIFFI_SPECIALIZE_SERIALIZABLE_PRIMITIVE(int64_t, ReadInt64, WriteInt64);
+UNIFFI_SPECIALIZE_SERIALIZABLE_PRIMITIVE(float, ReadFloat, WriteFloat);
+UNIFFI_SPECIALIZE_SERIALIZABLE_PRIMITIVE(double, ReadDouble, WriteDouble);
+
+/// Booleans are passed as unsigned integers over the FFI, because JNA doesn't
```
nnethercote, line 355: Change "unsigned integers" to "Uint8"s... I was briefly confused because I erroneously assumed a UInt32.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
nnethercote: Again: put this in a comment :)

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
```diff
@@ -0,0 +1,794 @@
...
+    aWriter.WriteRawString(aValue.Length(), [&](Span<char> aRawString) {
+      memcpy(aRawString.Elements(), aValue.BeginReading(), aRawString.Length());
+      return aRawString.Length();
+    });
+  }
+};
+
+template <>
+struct ViaFfi<nsACString, char*> {
+  static MOZ_MUST_USE bool Lift(const char*& aLowered, nsACString& aLifted) {
```
nnethercote, line 422: `[[nodiscard]]` is now preferred to `MOZ_MUST_USE`. See [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1571631).

uniffi_bindgen/src/bindings/gecko/templates/WebIDLTemplate.webidl
```diff
@@ -0,0 +1,69 @@
...
+
+{%- let functions = ci.iter_function_definitions() %}
+{%- if !functions.is_empty() %}
+[ChromeOnly, Exposed=Window]
+namespace {{ ci.namespace()|class_name_webidl }} {
+  {#-
+  // We'll need to figure out how to handle async methods. One option is
+  // to declare them as `async foo()`, or an `[Async]` or `[BackgroundThread]`
+  // attribute in the UniFFI IDL. Kotlin, Swift, and Python can ignore that
+  // anno; Gecko will generate a method that returns a `Promise` instead, and
```
nnethercote, line 28: Is "anno" short for "annotation"? If so, I've not seen that abbreviation before. Probably best to write the full word.

nnethercote (commented):
This looks good to me. A lot of boilerplate-y stuff, but that's unavoidable, and it's easy to read and well-commented. I have a few nits, mostly about improving comments.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
rfk: In fact this chance has now landed on the `main` branch.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
rfk: > Is big-endian used for all serialization in uniffi-rs?

It is, and it's probably worth a separate doc for the serialization format as a whole, since it's rather ad-hoc.

Some light background that might only be known to myself: I arbitrarily decided that everything should be big-endian when writing the first version of the serialization code, and I've no particular attachment to that decision. I wonder if it would be better to explicitly commit to using the native machine endianness, since we only ever expect to be serializing between two codebases running on the same machine (in the same process no less).

uniffi_bindgen/src/bindings/gecko/templates/InterfaceTemplate.cpp
linacambridge: I filed https://github.com/mozilla/uniffi-rs/issues/295.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
linacambridge: I was using `CheckedInt` throughout to avoid accidental overflows when serializing, but with the `RustBuffer` changes on `main` now, `Size` isn't necessary anymore. I think we can remove `CheckedInt` from the other uses, too, since `RustBuffer`s can only hold an `int32_t` worth of bytes.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
linacambridge: It supports reading into a UTF-8 and a UTF-16 string, too—but the bytes in the `RustBuffer` are always UTF-8 encoded, to match Rust strings. But I realized the signatures for `ReadRawString` and `WriteRawString` were super confusing, especially with the lambda doing strange things, so I flattened them into `ReadCString` and `ReadString` methods. It's a little more code, but a lot more readable, so seems like worth the trade-off.

linacambridge:
OK, I think I've addressed all of @nnethercote's feedback, and rebased all the latest changes!

@rfk, would you mind giving this another quick glance, too? Thanks!

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
linacambridge: Yep, the length is now `int32_t` and there's also a new capacity field.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
linacambridge: I clarified in the functions themselves that this is because the serialization format uses big endian (I don't think the caller needs to know that one way or the other, any more than it needs to know we serialize floats and doubles by their bit pattern), and I also agree about documenting the format, maybe as part of the [ADR that introduced it](https://github.com/mozilla/uniffi-rs/blob/757a028278f8c8d81754e6a32ea3844012669deb/docs/adr/0002-serialize-complex-datatypes.md), or as an appendix in the tutorial! 👍🏻

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
linacambridge: I actually removed this comment because `Writer` writes directly into a `RustBuffer`, since it's now growable and we can avoid copying everything into the `nsTArray`, then copying everything again into a `RustBuffer`.

uniffi_bindgen/src/bindings/gecko/templates/SharedHeaderTemplate.h
linacambridge: Ahh, oops, good call! Done! I also called out that it's an implementation detail of UniFFI, instead of mentioning JNA specifically.

rfk (approved):
This looks really good.

I mean, I kinda glazed over a bit at the intricacies of the C++ code because I trust that more experienced folks are looked at it, but I'm confident I could use the guideposts here to dig in more to the details if needed!

I note that the generated bindings occupy a kind of global namespace, e.g. we get a class `mozilla::dom::Optionnuer` in a header file conditioned on `#ifndef mozilla_dom_Optionneur`. Are we able to add the uniffi package name as a level here, to give e.g. `mozilla::dom::Rondpointer::Optionnuer` and similar? I'm concerned about accidental name conflicts in future, but I'm also cautious about potential limitations of the way Firefox generates the bindings.

Similarly, if we had multiple uniffi libraries being build together in Firefox, would they run into trouble by both declaring `RustBuffer` structs etc? I think maybe it's OK because the `RondpointShared.h` file is only included by its `.cpp` files, not by any other header files. But I'm no expert, so kind of just looking for explicit re-assurance here :-)

A couple of other broad nits I noticed:

* When running `cargo test` in this branch, I get quite a few warnings about unused symbols which it's probably worth cleaning up.
* Running `uniffi-bindgen` via the `cargo run --` method you describe in the PR works, but if I try to build and run the executable via `./target/debug/uniffi-bindgen` using the same arguments, it produces an error `'Missing $CARGO_MANIFEST_DIR, cannot load config file: NotPresent'`. We should expect that some consumers will use an external `uniffi-bindgen` rather than building and running it from within this repo.

uniffi/src/testing.rs
```diff
@@ -56,6 +56,8 @@ pub fn run_foreign_language_testcase(pkg_dir: &str, idl_file: &str, test_file: &
 /// Internally, this function does a bit of caching and concurrency management to avoid rebuilding
 /// the component for multiple testcases.
 pub fn ensure_compiled_cdylib(pkg_dir: &str) -> Result<String> {
+    unimplemented!()
```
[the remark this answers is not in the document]
rfk: I think this is causing CI test failures.

linacambridge:
> I mean, I kinda glazed over a bit at the intricacies of the C++ code because I trust that more experienced folks are looked at it, but I'm confident I could use the guideposts here to dig in more to the details if needed!

Thank you so much for digging into this, Ryan, that's mainly what I wanted from your review! 🚀

> I note that the generated bindings occupy a kind of global namespace, e.g. we get a class `mozilla::dom::Optionnuer` in a header file conditioned on `#ifndef mozilla_dom_Optionneur`. Are we able to add the uniffi package name as a level here, to give e.g. `mozilla::dom::Rondpointer::Optionnuer` and similar?

I don't think so 😞 I covered this in the Theory of Operation—which would be nice to convert into Markdown and check in to the repo, adding that to my list for this week ☑️ —but the binding declarations generated by Firefox's `Codegen.py` expect headers to be in `mozilla/dom/InterfaceName.h`, anda emit forward declarations that expect to find the interfaces and namespaces directly under `mozilla::dom`. (I think that also means type names will have to be globally unique, across UniFFI interfaces _and_ all the Web APIs currently in Gecko).

> Similarly, if we had multiple uniffi libraries being build together in Firefox, would they run into trouble by both declaring `RustBuffer` structs etc?

Yes, they would, because they're declared in the `extern "C"` block—thanks for noting this! I'll get that fixed up before landing, by prefixing `RustBuffer`, `ForeignBytes`, and `RustError` with the `ffi_namespace()`, like you do for the functions. It does mean we'll end up with duplication if we vendor multiple UniFFI components, but that's already the case today, and I suspect making cross-component helper deduping work is going to be more involved.

examples/arithmetic/Cargo.toml
```diff
@@ -7,7 +7,7 @@ license = "MPL-2.0"
 publish = false
 
 [lib]
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "lib"]
```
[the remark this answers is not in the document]
linacambridge: I'll go ahead and remove this since `arithmetic` doesn't have an xpcshell test, but we do need to do this in every crate we plan to vendor in m-c—since we're importing it as a Rust library via `extern crate`, not a `.dylib`.

uniffi/src/testing.rs
```diff
@@ -10,7 +10,7 @@
 //! the `uniffi_macros` crate.
 
 use anyhow::{bail, Result};
-use cargo_metadata::Message;
+// use cargo_metadata::Message;
```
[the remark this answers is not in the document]
linacambridge: Oops, I meant to revert these, thanks!

uniffi_bindgen/src/bindings/gecko_js/gen_gecko_js.rs
```diff
@@ -0,0 +1,463 @@
...
+    BindingArgument, BindingFunction, ReturnBy, ReturningBindingFunction, ThrowBy,
+};
+
+/// Config options for the generated Firefox front-end bindings. Note that this
+/// can only be used to control details *that do not affect the underlying
+/// component*, since the details of the underlying component are entirely
+/// determined by the `ComponentInterface`.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct Config {
+    // ...
```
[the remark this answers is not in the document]
linacambridge: Hmm, good question! There's limited customization that we can do, just because `Codegen.py` expects files and C++ namespaces to be laid out a certain way. I could imagine letting consumers specify a prefix, to make the WebIDL interface, namespace, dictionary, and type names globally unique, though—can you think of others we might want now?

uniffi_bindgen/src/bindings/gecko_js/gen_gecko_js.rs
```diff
@@ -0,0 +1,463 @@
...
+    }
+
+    pub fn field_name_cpp(nm: &str) -> Result<String, askama::Error> {
+        Ok(format!("m{}", nm.to_camel_case()))
+    }
+
+    pub fn enum_variant_cpp(nm: &dyn fmt::Display) -> Result<String, askama::Error> {
+        // TODO: Make sure this does the right thing for hyphenated variants.
+        // Example: "bookmark-added" should become `Bookmark_added`, because
+        // that's what Firefox's `Codegen.py` spits out.
```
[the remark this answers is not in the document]
linacambridge: I filed https://github.com/mozilla/uniffi-rs/issues/294 for this, but the consequences is that hyphenated names will produce a page-long compilation error because the enum variant that UniFFI generates doesn't exist in the variants that `Codegen.py` spits out 😬

uniffi_bindgen/src/bindings/gecko_js/templates/RustBufferHelper.h
```diff
@@ -0,0 +1,630 @@
...
+};
+
+/// Nullable values are prefixed by a tag: 0 if none; 1 followed by the
+/// serialized value if some. These are turned into Rust `Option<T>`s.
+///
+/// Fun fact: WebIDL also has a `dom::Optional<T>` type. They both use
+/// `mozilla::Maybe<T>` under the hood, but their semantics are different.
+/// `Nullable<T>` means JS must pass some value for the argument or dictionary
+/// field: either `T` or `null`. `Optional<T>` means JS can omit the argument
+/// or member entirely.
```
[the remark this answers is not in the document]
linacambridge: It's a fun fact that's specific to Gecko, I'll either clarify or take this out entirely. In WebIDL, `Nullable<T>` means "a value of type T or `null`", and `Optional<T>` means "an argument or a dictionary member that can be omitted entirely". Something can be [both](https://searchfox.org/mozilla-central/rev/f27594d62e7f1d57626889255ce6a3071d67209f/dom/webidl/Animation.webidl#20-21) `Nullable` and `Optional`, in which case it gets reflected [like this](https://searchfox.org/mozilla-central/rev/f27594d62e7f1d57626889255ce6a3071d67209f/dom/animation/Animation.cpp#124-126).

uniffi_bindgen/src/bindings/gecko_js/templates/WebIDLTemplate.webidl
```diff
@@ -0,0 +1,84 @@
...
+{%- let functions = ci.iter_function_definitions() %}
+{%- if !functions.is_empty() %}
+[ChromeOnly, Exposed=Window]
+namespace {{ ci.namespace()|class_name_webidl }} {
+  {#-
+  // We'll need to figure out how to handle async methods. One option is
+  // to declare them as `async foo()`, or an `[Async]` or `[BackgroundThread]`
+  // attribute in the UniFFI IDL. Kotlin, Swift, and Python can ignore that
+  // anno; Gecko will generate a method that returns a `Promise` instead, and
+  // dispatches the task to the background thread.
```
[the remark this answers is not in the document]
linacambridge: https://github.com/mozilla/uniffi-rs/issues/298, for posterity 😄

uniffi_bindgen/src/bindings/gecko_js/templates/macros.cpp
```diff
@@ -0,0 +1,59 @@
+{# /* Calls an FFI function. */ #}
```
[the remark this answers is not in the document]
linacambridge: Seriously! It's a whole feast of curly braces, percents, angle brackets, and colons—lovely C++ 😭

rfk:
> prefixing RustBuffer, ForeignBytes, and RustError with the ffi_namespace(), like you do for the functions.
> It does mean we'll end up with duplication

I have perhaps too much faith in Sufficiently Smart Compilers, but hopefully these duplicated struct definitions would not result in duplicate code in the compiled artifacts, would they?

rfk:
> (I think that also means type names will have to be globally unique, across UniFFI interfaces and all the Web APIs currently in Gecko).

Hrm. I can see this being a problem (and worse, one that catches us by surprise). If you think there's scope for adding some sort of unique-ifying prefix to these names, it's probably worth putting the infra for that in place sooner rather than later.

linacambridge:
> I have perhaps too much faith in Sufficiently Smart Compilers, but hopefully these duplicated struct definitions would not result in duplicate code in the compiled artifacts, would they?

Good question—my knowledge of how C++ compilers inline templates, functions, and types is nonexistent 😭 Just like you suggested profiling the size impact of vendoring Nimbus and the bindings, it would be worth measuring the impact of vendoring multiple UniFFI components.

The one thing I'm worried about that might make inlining difficult is that the serialization helpers call into component-specific allocation functions. In `libxul` or an App Services-style megazord, `ffi_rondpoint_abc123_rustbuffer_alloc` does the exact same thing as `ffi_arithmetic_789xyz_rustbuffer_alloc`—but they're exported as two different symbols, and we end up generating two copies of the helpers wrapping them—`rondpoint_detail::{Reader, Writer, Serializable, ViaFfi}` and `arithmetic_detail::{Reader, Writer, Serializable, ViaFfi}`, plus all the specializations. Is Clang smart enough to recognize that, and optimize some (or all) those calls out? I don't know 😞

linacambridge:
> Hrm. I can see this being a problem (and worse, one that catches us by surprise). If you think there's scope for adding some sort of unique-ifying prefix to these names, it's probably worth putting the infra for that in place sooner rather than later.

Good idea. Let's do that now, so that this doesn't sneak up on us later.

examples/rondpoint/tests/gecko/test_rondpoint.js
```diff
@@ -0,0 +1,141 @@
...
+  // Floats.
+  [0, 1, 0.25].forEach(v => equal(rt.identiqueFloat(v), v));
+
+  // Doubles.
+  [0, 1, 0.25].forEach(v => equal(rt.identiqueDouble(v), v));
+
+  // Strings.
+  [
+    "",
+    "abc",
```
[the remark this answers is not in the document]
linacambridge: Done, and I went ahead and ported over the `Optionneur` tests while we were at it.

uniffi/src/testing.rs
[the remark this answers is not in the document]
linacambridge: @dmose took care of these, and the tests pass now! 🎉

nnethercote (approved):
LGTM!

linacambridge:
Woohoo, thanks so much everyone! 😅 The latest push has the `definition_prefix` config option, prefixed names for the C structs (so we should be able to compile multiple UniFFIed components in Firefox), and a last-minute fix for nullable strings uncovered by the Rondpoint tests 👍🏻 (that example is paying for itself over and over!)

I'll go ahead and mash the "merge" button now :shipit: and we can file follow-ups for anything we forgot!

Issue #213, opened 2020-08-09
Document uniffi-bindgen


tarikeshaq:
Our README.md doesn't include any information about the CLI, we should probably add it. In #212 We add a couple of mentions of it, but we might benefit from having a section there explaining it.

- This should explain how one can download the tool (currently using `cargo install --path`, once it's on crates.io we'll add that too)
- Maybe an explanation of how the tool fits in the grand scheme of things.
- A mention of how to find more information (telling users to run the `--help` may be enough here)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1638)

PR #260, opened 2020-09-02
Uniffi user manual

eoger:
Fixes https://github.com/mozilla/uniffi-rs/issues/213.

[Rendered](https://mozilla.github.io/uniffi-rs/)

The intention is to give a hands-on guide to the user on how to integrate uniffi with their own crates.
I think we might be missing some sections but it's a start, so I wouldn't mind some feedback on this :)


docs/manual/src/Overview.md
```diff
@@ -0,0 +1,18 @@
+# uniffi
+
+Uniffi is a tool that automatically generates foreign-language bindings targeting Rust libraries.  
+It fits in the practice of consolidating business logic in a single Rust library while targeting multiple platforms, making it simpler to develop and maintain a cross-platform codebase.  
+Note that this tool will not help you ship a Rust library to these platforms, but simply not have to write bindings code by hand [[0]](https://i.kym-cdn.com/photos/images/newsfeed/000/572/078/d6d.jpg).
+
+## Design
+
+uniffi requires to write an Interface Definition Language ([IDL](https://heycam.github.io/webidl/)) file describing the methods and data structures available to the targeted languages.  
+This .idl file, whose definitions must match with the exposed Rust code, is then used to generate Rust *scaffolding* code and foreign-languages *bindings*. This process can take place either during the build process or be manually initiated by the developer.
```
rfk, line 10: This page would really benefit from a high-level diagram, along the lines of the one we have in the README for application-services (but with a few extra bits for where uniffi fits in).

docs/manual/src/idl/interfaces.md
```diff
@@ -0,0 +1,40 @@
+# Interfaces/Objects
+
+Interfaces refer to the Rust world as a struct with an `impl` block containing methods. In the Kotlin or Swift world, it's a class.
```
rfk, line 3: suggestion: "refer to" -> "are represented in" or similar.

docs/manual/src/idl/interfaces.md
```diff
@@ -0,0 +1,40 @@
+# Interfaces/Objects
+
+Interfaces refer to the Rust world as a struct with an `impl` block containing methods. In the Kotlin or Swift world, it's a class.  
+Because of Reasons, in the uniffi world it is impossible to be both an Object and a [Dictionary](./structs.md).
```
rfk, line 4: If we wanted to be slightly less mysterious about The Reasons, we could describe objects are being passed around by reference while dictionaries are passed around by value (or by copy or similar).

docs/manual/src/tutorial/Prerequisites.md
```diff
@@ -0,0 +1,18 @@
...
+`cargo install --git https://github.com/mozilla/uniffi-rs --branch main uniffi_bindgen`
+
+You can see what it can do with `uniffi-bindgen --help`, but let's leave it aside for now.
+
+## Build your crate as a cdylib
+
+Ensure your crate builds as a `cdylib` by adding
+```toml
+crate-type = ["cdylib"]
+```
```
rfk, line 17: I believe you also need to set a specific name for the resulting `.so` file.

docs/manual/src/tutorial/Rust_scaffolding.md
```diff
@@ -0,0 +1,39 @@
...
+
+First, add `uniffi` to your crate dependencies: it is the runtime support code that powers uniffi's serialization of data types across languages:
+
+<!-- TODO: Use a published version -->
+```toml
+[dependencies]
+uniffi = { git = "https://github.com/mozilla/uniffi-rs", branch = "main" }
+```
+
+Then let's add `uniffi_build` to your build dependencies: it generates the Rust scaffolding code that exposes our Rust functions as a C abi.
```
rfk, line 15: I recall thom having very specific opinions on how to describe this C layer, and I'm not sure "C abi" is technically accurate. Something like "C-compatible FFI layer" or similar might work.

rfk (approved):
TBH I'm surprised how much I like the book format for this, nice work! I left a few comments below, but I think we should err on the side of landing this and iterating rather than trying to polish it over and over in a single PR. A great starting point!

eoger:
Thanks for the review, I've added a diagram.

PR #261, opened 2020-09-07
Make RustBuffer growable on demand, grow it rather than pre-calculating final size.

### LINKED ISSUE #251: Replace `calculateWriteSize` with a growable buffer
Each of our foreign-language bindings generators currently implements a `calculateWriteSize` method of some sort, which is used when serializing complex datatypes to send over the FFI like so:

1. Call `calculateWriteSize(v)` to determine the precise number bytes into which `v` will serialize.
2. Allocate a `RustBuffer` of exactly that size.
3. Serialize `v` into the allocated buffer.

In retrospect, I think this was a premature optimization, and it's not serving us well in practice. Two downsides come to mind:

* It adds nontrivial cognitive overhead to working on the bindings generator, multiplied by the number of bindings backends.
* It adds overhead for serializing strings in languages where utf8 is not the native encoding, since you end up converting the string to utf8 twice - once to find out how many bytes it occupies, then a second time to actually serialize it.

Notably, the rust scaffolding does *not* pre-calculate sizes in this way when returning complex datatypes; it just uses a `Vec` and allows the `Vec` to automagically grow as required.

I think we should do the same thing on the foreign-language side, as follows:

* Add an FFI function `ffi_bytebuffer_reserve(buf, n)` which calls through to [Vec::reserve](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reserve) in the rust code.
* In each of the foreign-language bindings, at the point where we check that we're not writing past the end of the buffer, have it call `ffi_bytebuffer_reserve` rather than throwing an error.
* Delete all the `calcaulteWriteSize` stuff from the foreign-language bindings.

Rust's `std::vec::Vec` does not guarantee any particular growth strategy when calling `reserve` multiple times, but it *does* claim that calling `push()` multiple times will have `O(1)` amortized cost per element That's good enough a guarantee for our purposes here, and we should expect that serializing any particular value in this way will involve a sub-linear number of calls to `ffi_bytebuffer_reserve`. (In practice right now `Vec::reserve` always at least doubles the size of the underlying buffer, which is what we'd do if we had to manage a growable buffer for ourselves).

Any objections?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1678)

rfk:
Fixes #251, with some additional cleanups that kind of fell out along the way.

The Kotlin and Python bindings currently do an awkward pre-calculation of the serialized
size of a record, so that they can pre-allocate a `RustBuffer` into which to write it.
This commit changes things to use a grown-on-demand `RustBuffer`, which brings with it
some nice improvements.

The broad strokes of this PR:

* The `RustBuffer` struct now exposes the `capacity` field from Rust, and a new
  `reserve` method mirroring the behaviour of `Vec<u8>::reserve`. This allows
  the foreign-language code to grow rust buffers on demand.

* Since `RustBuffer` is now specific to how `Vec`s are implemented in rust, there
  is a new `ForeignBytes` struct that can be used for passing a buffer of bytes
  *in* to rust from the foreign language code

* When serializing complex data types, the Kotlin and Python bindings now grow
  the target buffer on demand rather than pre-calculating its final size.

* The swift bindings still serialize into a local byte array and then copy it
  over into a `RustBuffer` at the end, but this process is slightly simpler
  thanks to the new `ForeignBytes` struct.

* Since we can no longer use the `ffi_support::ByteBuffer` struct for buffers,
  we are no longer bound to its pre-canned destructor functions. This means we
  can make all FFI functions take an `&err` out param, simplifying some of the
  code gen logic.


uniffi/src/lib.rs
```diff
@@ -16,19 +16,25 @@
 //!  * How to ["lower"](ViaFfi::lower) rust values of that type into an appropriate low-level
 //!    FFI value.
 //!  * How to ["lift"](ViaFfi::lift) low-level FFI values back into rust values of that type.
-//!  * How to [write](ViaFfi::write) rust values of that type into a bytebuffer, for cases
+//!  * How to [write](ViaFfi::write) rust values of that type into a buffer, for cases
 //!    where they are part of a compount data structure that is serialized for transfer.
-//!  * How to [read](ViaFfi::read) rust values of that type from bytebuffer, for cases
+//!  * How to [read](ViaFfi::read) rust values of that type from buffer, for cases
```
rfk, line 21: FYI, a lot of minor changes here are around trying to use consistent terminology across the different backends, saying "rust buffer" when we mean a `Vec<u8>` owned by rust.

uniffi/src/lib.rs
```diff
@@ -223,24 +229,188 @@ unsafe impl ViaFfi for bool {
...
+/// are safe) or it came from foreign-language code (which should be encforcing
+/// safety invariants).
+///
+/// This struct is based on `ByteBuffer` from the `ffi-support` crate, but modified
+/// to retain unallocated capacity rather than truncating to the occupied length.
+#[repr(C)]
+pub struct RustBuffer {
+    /// The allocated capacity of the underlying `Vec<u8>`.
+    /// In rust this is a `usize`, but we use an `i32` for compatibility with JNA.
+    capacity: i32,
```
rfk, line 262: Note that this is different from `ffi_support`, which uses an `i64` for the `len` field. I think we're mostly sticking to `i32` for length-related things due to JVM limitations so I figured it was worth being consistent here.

uniffi/src/lib.rs
```diff
@@ -223,24 +229,188 @@ unsafe impl ViaFfi for bool {
...
+                .expect("buffer length negative or overflowed");
+            unsafe { Vec::from_raw_parts(self.data, len, capacity) }
+        }
+    }
+
+    /// Reclaim memory stored in this `RustBuffer`.
+    pub fn destroy(self) {
+        drop(self.destroy_into_vec());
+    }
+}
```
rfk, line 332: TODO: it's probably worth adding some tests for this (or at least, copying across existing tests from ffi-support)

rfk:
I'll add for completeness here, I think it would be *even better* if we could just pass serialized data as a `ForeignBytes` struct rather than copying it into rust and then reading it out of rust-owned buffer. To enable that we would need figure out how to update the `ViaFfi` trait to deal with asymmetric types, so that we can receive arguments serialized as a `ForeignBytes` but return results serialized as a `RustBuffer`. That's for follow-up work.

badboy (commented):
I gave the Rust code a look (and briefly peaked at how it's used in the Kotlin part).
I think this takes care of checking all invariants correctly.
I left a couple of comments inline, I especially think that commenting on why some parts are the way they are will help in the long run.

uniffi/src/lib.rs
```diff
@@ -400,3 +576,159 @@ unsafe impl ViaFfi for String {
...
+        assert_eq!(rbuf.destroy_into_vec().as_slice(), &[0u8, 0, 0, 0, 0]);
+
+        let rbuf = RustBuffer::new_with_size(0);
+        assert!(!rbuf.data.is_null());
+        assert_eq!(rbuf.destroy_into_vec().as_slice(), &[0u8; 0]);
+    }
+
+    #[test]
+    fn test_rustbuffer_null_means_empty() {
+        let rbuf = RustBuffer {
```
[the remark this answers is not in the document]
rfk: That's exactly right, yeah, the idea here is to have some guards against broken foreign-language code passing an invalid `RustBuffer` struct over the FFI. I'll update the comments to make this clearer, thanks!

uniffi/src/lib.rs
```diff
@@ -223,24 +229,194 @@ unsafe impl ViaFfi for bool {
...
+    }
+
+    /// Creates a `RustBuffer` zero-filed to the requested size.
+    ///
+    /// The resulting vector will not be automatically dropped; you must
+    /// arrange to call `destroy` or `destroy_into_vec` when finished with it.
+    pub fn new_with_size(size: usize) -> Self {
+        // Note: `Vec` requires this internally on 64 bit platforms (and has a
+        // stricter requirement on 32 bit ones), so this is just to be explicit.
+        assert!(size < i64::MAX as usize);
```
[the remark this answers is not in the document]
rfk: nice catch 👍

uniffi/src/lib.rs
```diff
@@ -223,24 +229,194 @@ unsafe impl ViaFfi for bool {
...
+    }
+
+    /// Consumes a `Vec<u8>` and returns its raw parts as a `RustBuffer`.
+    ///
+    /// The resulting vector will not be automatically dropped; you must
+    /// arrange to call `destroy` or `destroy_into_vec` when finished with it.
+    pub fn from_vec(v: Vec<u8>) -> Self {
+        let mut v = std::mem::ManuallyDrop::new(v);
+        Self {
+            capacity: i32::try_from(v.capacity()).expect("buffer capacity cannot fit into a i32."),
```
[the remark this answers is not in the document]
rfk: nice catch, I'm going to apply to suggested fix 👍

uniffi/src/lib.rs
```diff
@@ -400,3 +576,159 @@ unsafe impl ViaFfi for String {
...
+            data: v.as_mut_ptr(),
+        };
+        rbuf.destroy_into_vec();
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_rustbuffer_provided_len_must_be_non_negative() {
+        let mut v = vec![0u8, 1, 2];
+        v.shrink_to_fit();
```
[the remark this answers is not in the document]
rfk: I don't think this is required, I'll remove it to avoid needed confusion.

uniffi/src/lib.rs
```diff
@@ -223,24 +229,194 @@ unsafe impl ViaFfi for bool {
...
+///     ownership system. To avoid memory leaks they *must* passed back into
+///     rust and either explicitly destroyed using `RustBuffer::destroy`, or
+///     converted back to a `Vec<u8>` using `RustBuffer::destroy_into_vec`
+///     (which will then be dropped via rust's usual ownership-tracking system).
+///
+/// Implementation note: all the fields of this struct are private, so you can't
+/// manually construct instances that don't come from a `Vec<u8>`. If you've got
+/// a `RustBuffer` then it either came from a public constructor (all of which
+/// are safe) or it came from foreign-language code (which should be encforcing
+/// safety invariants).
```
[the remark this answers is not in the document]
rfk: Added, with the additional caveat that foreign-language code can safely construct an empty buffer using a null pointer.

rfk:
@linacambridge, pinging you for an r? on this one because I think it'll help reduce the complexity of your desktop prototypes.

uniffi/src/ffi/foreignbytes.rs
```diff
@@ -0,0 +1,118 @@
...
+///
+/// Implementation note: all the fields of this struct are private and it has no
+/// constructors, so consuming crates cant create instances of it. If you've
+/// got a `ForeignBytes`, then you received it over the FFI and are assuming that
+/// the foreign language code is upholding the above invariants.
+///
+/// This struct is based on `ByteBuffer` from the `ffi-support` crate, but modified
+/// to give a read-only view of externally-provided bytes.
+#[repr(C)]
+pub struct ForeignBytes {
```
linacambridge, line 28: Oooh, I like this conceptual split between buffers allocated by the host language, and buffers allocated in Rust!

uniffi/src/ffi/rustbuffer.rs
```diff
@@ -0,0 +1,300 @@
...
+///
+/// Foreign-language code should not construct `RustBuffer` structs other than
+/// by receiving them from a call into the Rust code, and should not modify them
+/// apart from the following safe operations:
+///
+///   * Writing bytes into the buffer pointed to by `data`, without writing
+///     beyond the indicated `capacity`.
+///
+///   * Adjusting the `len` property to indicate the amount of data written,
+///     while ensuring that 0 <= `len` <= `capacity`.
```
linacambridge, line 33: Nice! This will let the Desktop bindings overallocate and then write directly into the buffer without using an intermediate `nsTArray`, and we can also change the Swift bindings to write into a buffer directly without first going through `Data`, like you suggested doing for Kotlin.

uniffi/src/ffi/rustbuffer.rs
```diff
@@ -0,0 +1,300 @@
...
+    /// Panics if the requested size is too large to fit in an `i32`, and
+    /// hence would risk incompatibility with some foreign-language code.
+    pub fn new_with_size(size: usize) -> Self {
+        assert!(
+            size < i32::MAX as usize,
+            "RustBuffer requested size too large"
+        );
+        let mut buf = vec![];
+        buf.resize(size, 0);
+        Self::from_vec(buf)
```
linacambridge, line 103: Would this work?

```suggestion
        Self::from_vec(vec![0; size])
```

uniffi_bindgen/src/bindings/kotlin/templates/wrapper.kt
```diff
@@ -12,9 +12,9 @@ package {{ config.package_name }};
 //
 // However, it's important that the detils of how this helper code works (e.g. the
 // way that different builtin types are passed across the FFI) exactly match what's
-// expected by the rust code on the other side of the interface. In practice right
+// expected by the Rust code on the other side of the interface. In practice right
 // now that means coming from the exact some version of `uniffi` that was used to
-// compile the rust component. The easiest way to ensure this is to bundle the Kotlin
+// compile the Eust component. The easiest way to ensure this is to bundle the Kotlin
```
linacambridge, line 17:
```suggestion
// compile the Rust component. The easiest way to ensure this is to bundle the Kotlin
```

linacambridge (approved):
Wonderful!

uniffi_bindgen/src/bindings/kotlin/templates/wrapper.kt
rfk: 😆

PR #264, opened 2020-09-08
Consolidate examples READMEs

### LINKED ISSUE #219: Consolidate general information in example README's into one README
All our examples have README's that basically say the same thing (for example [here](https://github.com/mozilla/uniffi-rs/blob/main/examples/arithmetic/README.md)'s the `arithmetic` README, and [here](https://github.com/mozilla/uniffi-rs/blob/main/examples/todolist/README.md)'s the `todolist` README) We should put that general information (the required libraries etc) in the [README](https://github.com/mozilla/uniffi-rs/blob/main/examples/README.md) in the root examples directory

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1640)

eoger:
Fixes #219


rfk (approved):
👍

PR #273, opened 2020-09-10
Add toml file to configure package_name for kotlin binding

### LINKED ISSUE #258: Allow setting the package name in the generated kotlin file
Right now the generated kotlin file has the package name `uniffi.{name_of_lib}`. It is useful for Nimbus though if it was possible to override that.

This is useful because at https://github.com/mozilla/nimbus-sdk/pull/18 we publish the generated code directly.

The generated package name is set here https://github.com/mozilla/uniffi-rs/blob/a23fd65e1777c8274c80feab139b906442cdb02e/uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs#L21

There are a couple of options, one being having the package name in the idl, the other being having it as a parameter to `uniffi-bindgen`. Since the package name may not be relevant to all the languages, we may prefer the latter option

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1722)

jhugman:
Fixes #258. 

This PR introduces a `uniffi.toml` config file that sits in `env::var("CARGO_MANIFEST_DIR")`.

We use `serde:toml:de` to deserialize into a config object, and generate a config object from the `ComponentInterface` then merge them.

This is certainly a bit over engineered for just the package name, but this is preferable to the proliferation of command line args.


rfk:
> I'd prefer it to be in env::var("CARGO_MANIFEST_DIR")

Without thinking super deeply about it, I think this is my first preference as well. Thinking of *the crate itself as a whole* as the thing that's being processed by uniffi seems like a good mental model to me, especially if we do end up moving to a macro-based approach.

> I think a better way of doing all this would be to serde::toml::Deserialize into a config struct,

If `serde` isn't working out, you could also consider making a `Config` struct and using your own custom `TryInto` implementation to convert into it. It might at least capture all the Postel-related code into a single location.

examples/geometry/uniffi.toml
```diff
@@ -0,0 +1 @@
+[bindings.swift]
```
jhugman, line 1: There's nothing to do with the swift bindings, but look, the kotlin ones don't need to be there and everything still works.

uniffi_bindgen/src/bindings/kotlin/mod.rs
```diff
@@ -18,11 +18,11 @@ pub use gen_kotlin::{Config, KotlinWrapper};
 use super::super::interface::ComponentInterface;
 
 pub fn write_bindings(
+    config: &Config,
```
jhugman, line 21: This file is pretty much all plumbing, feeding the config from the callers down to `gen_kotlin`.

uniffi_bindgen/src/bindings/mod.rs
```diff
@@ -59,8 +61,39 @@ impl TryFrom<String> for TargetLanguage {
...
+impl MergeWith for Config {
+    fn merge_with(&self, other: &Self) -> Self {
+        Config {
+            kotlin: self.kotlin.merge_with(&other.kotlin),
+            swift: self.swift.merge_with(&other.swift),
+            python: self.python.merge_with(&other.python),
+        }
+    }
+}
+
```
jhugman, line 93: This is pretty much the pattern: we define a struct that is `Deserializable`, give an implementation from `ComponentInterface`, and how we're going to merge the two together.

uniffi_bindgen/src/lib.rs
```diff
@@ -267,3 +303,39 @@ fn slurp_file(file_name: &Path) -> Result<String> {
...
+
+impl<T: Clone> MergeWith for Option<T> {
+    fn merge_with(&self, other: &Self) -> Self {
+        match (self, other) {
+            (Some(_), _) => self.clone(),
+            (None, Some(_)) => other.clone(),
+            (None, None) => None,
+        }
+    }
+}
```
jhugman, line 341: This is the cleverest bit in the PR, and it's not very clever.

uniffi_bindgen/src/lib.rs
```diff
@@ -230,15 +243,38 @@ pub fn run_tests<P: AsRef<Path>>(
...
+        .expect("Missing $CARGO_MANIFEST_DIR, cannot load config file")
+        .into();
+
+    let config_file = pkg_path.join("uniffi.toml").canonicalize();
+
+    if config_file.is_err() {
+        return Ok(default_config);
+    }
+
+    let config_file = config_file.unwrap();
```
linacambridge, line 268: Minor nit: it's a bit more idiomatic Rust to write it like this...

```suggestion
    let config_file = match config_file {
        Ok(f) => f,
        Err(_) => return Ok(default_config),
    };
```

...Even though they're conceptually the same.

uniffi_bindgen/src/lib.rs
```diff
@@ -230,15 +243,38 @@ pub fn run_tests<P: AsRef<Path>>(
...
+    let contents = slurp_file(&config_file)
+        .map_err(|_| anyhow!("Failed to read config file from {:?}", &config_file))?;
```
linacambridge, line 270: Another small nit: I like using `.context()` and `.with_context()` (you might need to `use anyhow::Context` at the top) instead of `map_err`, like this...

```suggestion
    let contents = slurp_file(&config_file)
        .with_context(|| format!("Failed to read config file from {:?}", &config_file))?;
```

Because that preserves the original error, so you'll get, say, the syntax error in addition to the context of where it came from.

linacambridge (approved):
This looks perfect, thanks @jhugman! I really like your `MergeWith` trait, it makes it easy to see where the defaults come from, and how they bubble up.

There's probably a "clever" way to do this with Serde, like a custom `Default` implementation on a struct with a reference to the `ComponentInterface`, and sprinkling some `serde(default)`s and `serde(flatten)`s throughout, or a custom deserializer at worst...but I think that would be much more convoluted than what you've done here.

PR #277, opened 2020-09-11
Mention how to install uniffi-bindgen if we fail to execute it.

mhammond:
Also updates other instructions to reflect the name is `uniffi_bindgen`
rather than `uniffi-bindgen`.

### DIFF
--- a/uniffi_bindgen/src/lib.rs
+++ b/uniffi_bindgen/src/lib.rs
@@ -58,7 +58,7 @@
 //!
 //! ### 3) Generate and include component scaffolding from the IDL file
 //!
-//! First you will need to install `uniffi-bindgen` on your system using `cargo install uniffi-bindgen`.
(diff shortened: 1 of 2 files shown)


rfk (approved):
r+, but also, I wonder if we should change the crate name so that it is in fact the more idiomatic `uniffi-bindgen`...

mhammond:
>  I wonder if we should change the crate name so that it is in fact the more idiomatic `uniffi-bindgen`...

I assumed that situation was forced on us - +1 to unifiiing that if possible 🥴

rfk:
Oh no, it looks like hyphens-versus-underscores is a [whole thing](https://github.com/rust-lang/cargo/issues/2775) and now that I've published with the underscore, we might not be able to use the hyphen.

For example, linking to the crate as [uniffi-bindgen](https://crates.io/crates/uniffi-bindgen) works as far as crates.io is concerned, but attempting to `cargo install uniffi-bindgen` fails as you note above. Bleh.

PR #279, opened 2020-09-12
Make experimental Kotlin features optional for consumers.

### LINKED ISSUE #259: Generated Kotlin uses experimental unsigned types
I just noticed this while building the nimbus-sdk kotlin bindings added by Tarik. We get a lot of warnings about using [experimental unsigned types](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-experimental-unsigned-types/) without using the annotation saying "we know we are".

The warning:
`This declaration is experimental and its usage should be marked with '@kotlin.ExperimentalUnsignedTypes' or '@OptIn(kotlin.ExperimentalUnsignedTypes::class)'`

This may come up as an issue if used in a project that has warnings as errors enabled.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1730)

rfk:
(This builds on #262, but only because I didn't want to deal with the merge conflicts in `RustBufferHelpers.kt`).

Previously, we were using two experimental features of Kotlin, in different ways:

* Unsigned types, both as API-level types and for some internal helpers.
* The `buildMap` function, for easily building a Map.

Both of these were emitted as part of the generated Kotlin code, regardless of whether the component actually needed them.

In the first case, consumers would get a warning that they need to opt-in to using experimental unsigned types, even if the API they were consuming did not use any unsigned types.

In the second case, we were silently opting in to an experimental API, meaning that consumers might find the generated code broken by a future Kotlin release without any warning.

This commit removes default-on experimental Kotlin APIs. Now, the only time Kotlin consumers will hear about experimental APIs is if they're using a component with unsigned integers in its public API, in which case they will need to explicitly opt in (and hence will be aware of the potential for bustage if the feature gets removed or changed in future Kotlin releases).

The key here was to refactor the ways the `ComponentInterface` deals with types, so that we can inspect the set of types that
are actually used by the interface and emit only the code required for those types. This is encapsulated in a new `TypeUniverse`
struct whose job is to maintain that whole-interface view of the set of types in use.

Fixes #259, in the sense that I tried compiling nimbus-sdk with the branch and it didn't emit any warnings.


uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -114,199 +261,133 @@ fun String.Companion.lift(rbuf: RustBuffer.ByValue): String {
...
-    return liftFromRustBuffer(rbuf) { buf -> readSequence(buf, readItem) }
-}
+{% when Type::Optional with (inner_type) -%}
+{% let inner_type_name = inner_type|type_kt %}
 
-fun<T> readSequence(buf: ByteBuffer, readItem: (ByteBuffer) -> T): List<T> {
-    val len = Int.read(buf)
-    return List<T>(len) {
-        readItem(buf)
+fun lift{{ type_name }}(rbuf: RustBuffer.ByValue): {{ inner_type_name }}? {
```
rfk, line 288: So this may be slightly controversial, and I welcome pushback here if you don't like it.

I've opted to "monomorphize" the helper functions for `Optional`, `Sequence` and `Map` types, emitting a separate helper for each concrete instance of such a type rather than a single generic helper that takes a lambda. I find this form more readable as I don't have to mentally untangle to lambda, and it happens to fit nicely with the "iterate over all the types" approach.

The downside may be slightly increased code size, although I have a hard time imagining it'll make a big difference in practice. I also think reasonable people could disagree with me about which is the more readable approach :-)

Anyway, if y'all hate this, an alternative would be to:

* Add methods on `TypeUniverse` like `has_map_types()`, `has_sequence_types()` etc.
* Change the template here to conditionally emit the Map/Sequence/Optional helpers only when those types are actually used in the interface.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -44,27 +44,26 @@
...
 
 pub mod types;
-use types::TypeResolver;
+use types::TypeUniverse;
 pub use types::{FFIType, Type};
 
 /// The main public interface for this module, representing the complete details of an interface exposed
 /// by a rust component and the details of consuming it via an extern-C FFI layer.
 ///
-#[derive(Debug, Default, Serialize, Deserialize)]
```
rfk, line 60: We no longer attempt to serialize/deserialize `ComponentInterface` structs, so I've removed the derivations here rather than adding them to my new structs.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -336,13 +323,21 @@ impl APIBuilder for weedle::Definition<'_> {
                     false
                 };
                 if is_error {
-                    ci.add_error_definition(d.convert(ci)?)
+                    let err = d.convert(ci)?;
+                    ci.add_error_definition(err)
```
rfk, line 327: This and similar changes, is because `convert()` and friends now take an `&mut ci`, and Rust's order of evaluation tries to take the mutable reference for `ci.add_error_definition` before evaluating any of its arguments. So it thinks I'm trying to do a double-mutable-borrow unless I pull the conversion into a separate line.

uniffi_bindgen/src/interface/types.rs
```diff
@@ -128,243 +132,349 @@ impl From<&Type> for FFIType {
...
-            bail!("please don't shadow builtin types (enum {})", name);
-        }
-        ci.add_type_definition(self.identifier.0, Type::Enum(name))
+        types.add_type_definition(self.identifier.0, Type::Enum(name))
     }
 }
 
 impl TypeFinder for weedle::TypedefDefinition<'_> {
-    fn find_type_definitions(&self, ci: &mut ComponentInterface) -> Result<()> {
+    fn add_type_definitions_to(&self, types: &mut TypeUniverse) -> Result<()> {
```
rfk, line 317: Most of the changes in the `TypeFinder` and `TypeResolve` trait impls are simple renamings, I don't think I've made any substantial changes to the logic. (Apart from moving the "dont override builtin names" check into `TypeUniverse.add_type_definition` so that it's done in a single consistent spot).

uniffi_bindgen/src/interface/types.rs
```diff
@@ -128,243 +132,349 @@ impl From<&Type> for FFIType {
...
-/// resolve types by name.
+/// a parsed IDL file, turning them into a corresponding [Type] struct. It uses the
+/// known type definitions in a [TypeUniverse] to resolve names, and so it assumes
+/// that we've already done a [TypeFinder] pass.
 ///
 /// (And to be honest, a big part of its job is to error out if we encounter any of the
 /// many, many WebIDL type definitions that are not supported by uniffi.)
 pub(crate) trait TypeResolver {
-    fn resolve_type_definition(&self, ci: &ComponentInterface) -> Result<Type>;
+    fn resolve_type_expression(&self, types: &mut TypeUniverse) -> Result<Type>;
```
rfk, line 339: The thing this method operates on is not a "definition" in any meaningful sense - it's an expression for a type, built up from the named types we discovered in the type-finding pass. So, I've renamed the method.

rfk:
Also, cross-linking for completeness, the `TypeUniverse` approach used here subsumes all the `iter_types` machinery I was messing around with over in https://github.com/mozilla/uniffi-rs/pull/214; I think this version is easier to understand and also likely to scale better.

uniffi_bindgen/src/interface/types.rs
```diff
@@ -128,243 +132,349 @@ impl From<&Type> for FFIType {
...
 /// we know how to resolve names to types when building up the full interface
 /// definition.
 pub(crate) trait TypeFinder {
-    fn find_type_definitions(&self, ci: &mut ComponentInterface) -> Result<()>;
+    fn add_type_definitions_to(&self, types: &mut TypeUniverse) -> Result<()>;
 }
 
-impl<T: TypeFinder> TypeFinder for Vec<T> {
-    fn find_type_definitions(&self, ci: &mut ComponentInterface) -> Result<()> {
+impl<T: TypeFinder> TypeFinder for &Vec<T> {
```
linacambridge, line 267: Can we make this `&[T]` so you can resolve types for any kind of slice?

linacambridge (approved):
On the one hand, this does feel a little hacky...one of the points you raised a while back is that our bindings will have to target the lowest common denominator, by design. It is a little unfortunate that Kotlin’s lack of support for unsigned integers mean we have to make string, sequence, and map sizes signed, even though our other typed languages—Swift and C++—can support them without a hitch. We also have to do a bit more work to make sure they’re in range, and, for languages that _do_ support unsigned types, we need to first cast them to signed.

But I think that’s a philosophical objection, and it would be a shame if our Kotlin consumers had a worse experience with spurious warnings about unsigned types just because all the other languages support them.

Practically, I think this is the right way to go. It does restrict data structure sizes even more, but if you’re trying to pass a 2+ GB map over the FFI, you’re going to have a bad time for many reasons other than integer width.

Overall, this looks wonderful!

uniffi_bindgen/src/interface/types.rs
```diff
@@ -128,243 +132,349 @@ impl From<&Type> for FFIType {
...
+///
+/// You could imagine this struct doing some clever interning of names and so-on in future,
+/// to reduce the overhead of passing around [Type] instances. For now we just do a whole
+/// lot of cloning.
+#[derive(Debug, Default)]
+pub(crate) struct TypeUniverse {
+    // Named type definitions (including aliases).
+    type_definitions: HashMap<String, Type>,
+    // All the types in the universe, by canonical type name.
+    // We should use a `HashSet`
```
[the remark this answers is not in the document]
rfk: Heh, I was originally going to use a `HashMap` and perform some extra cross-checks, and this comment was going to be to explain why that was...but in the end I didn't go that route.

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -13,6 +13,28 @@ fun<T> liftFromRustBuffer(rbuf: RustBuffer.ByValue, readItem: (ByteBuffer) -> T)
...
+        throw e
+    }
+}
+
+// For every type used in the interface, we provide helper methods for conveniently
+// lifting and lowering that type from C-compatible data, and for reading and writing
+// values of that type in a buffer.
+
+{% for typ in ci.iter_types() %}
+{% let type_name = typ.canonical_name()|class_name_kt %}
```
jhugman, line 33: Nit: naming. I've scrolled up to here to find out what this was. 

Would prefer  something more descriptive: `canonical_type_name` or `type_identifier`.

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -53,39 +115,104 @@ fun Long.Companion.read(buf: ByteBuffer): Long {
...
     return v.toUShort()
 }
 
+@ExperimentalUnsignedTypes
 fun UShort.Companion.read(buf: ByteBuffer): UShort {
     return UShort.lift(buf.getShort())
 }
 
+@ExperimentalUnsignedTypes
+fun UShort.lower(): Short {
```
jhugman, line 161: This might be an ideal time to change the visibility of these methods to `private` or `internal`.

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -114,199 +261,133 @@ fun String.Companion.lift(rbuf: RustBuffer.ByValue): String {
...
+fun String.write(buf: RustBufferBuilder) {
+    val byteArr = this.toByteArray(Charsets.UTF_8)
+    buf.putInt(byteArr.size)
+    buf.put(byteArr)
 }
 
-fun<T> liftSequence(rbuf: RustBuffer.ByValue, readItem: (ByteBuffer) -> T): List<T> {
-    return liftFromRustBuffer(rbuf) { buf -> readSequence(buf, readItem) }
-}
+{% when Type::Optional with (inner_type) -%}
```
jhugman, line 285: Nit: we're getting to library specific code now.

Suggest adding a comment in the generate code to make this easier to find?

uniffi_bindgen/src/interface/types.rs
```diff
@@ -128,243 +132,348 @@ impl From<&Type> for FFIType {
...
+            Type::Enum(nm) => format!("Enum{}", nm),
+            Type::Record(nm) => format!("Record{}", nm),
+            // Recursive types.
+            // These add a prefix to the name of the underlying type.
+            // The component API definition cannot give names to recursive types, so as long as the
+            // prefixes we add here are all unique amongst themselves, then we have no chance of
+            // acccidentally generating name collisions.
+            Type::Optional(t) => format!("Optional{}", t.canonical_name()),
+            Type::Sequence(t) => format!("Sequence{}", t.canonical_name()),
+            Type::Map(t) => format!("Map{}", t.canonical_name()),
```
jhugman, line 172: There's lots of places in the code where you're immediately calling `name.to_camel_case()`.

```
class_name_kt(&type_.canonical_name())?
```

Does heck respect the existing case? 

Just found: `lowerSequenceEnumEnumeration`, so yes it appears it does.

uniffi_bindgen/src/interface/types.rs
jhugman: I'm still not wild about the name, (either the `resolve` or the `expression` bits).

This might be that the difference in naming between the `TypeResolver` and `TypeFinder` is so small.

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -53,39 +115,104 @@ fun Long.Companion.read(buf: ByteBuffer): Long {
...
+fun UShort.lower(): Short {
+    return this.toShort()
+}
+
+@ExperimentalUnsignedTypes
+fun UShort.write(buf: RustBufferBuilder) {
+    buf.putShort(this.toShort())
+}
+
+{% when Type::UInt32 -%}
```
jhugman, line 170: FWIW I love this grouping by type rather than reader and writers.

uniffi_bindgen/src/interface/types.rs
```diff
@@ -128,243 +132,348 @@ impl From<&Type> for FFIType {
...
+        }
+    }
+}
+
+/// The set of all possible types used in a particular component interface.
```
jhugman, line 177: Might this be a good time to split up `types.rs` in to a few smaller files?

jhugman (approved):
Very much like this PR. I've left a few comments, and some nits, but I like the type universe/oracle, divvying up by types and the monomorphising of types. 🚚 🚢 :shipit: 🚀

uniffi_bindgen/src/interface/types.rs
rfk: That's fair; I think the new name is *better*, but that's not the same thing as *good*.

"Resolve" is probably not what we're really mean here. Not for this PR, but I'd be happy to try to brainstorm something better. Maybe something using "builder" could be helpful, since this builds more complex structured types out of the named primitives we found in the previous pass?

Good naming is hard :-/

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt
rfk: Oof, yes, it had somehow slipped my mind that Kotlin defaults to public visibility. I'll update these to `internal`.

uniffi_bindgen/src/interface/types.rs
rfk: Yeah, I think it is time, but I don't want to do it as part of this PR; filed https://github.com/mozilla/uniffi-rs/issues/286 for followup

PR #280, opened 2020-09-12
Include a short checksum in generated FFI function names, for extra link-time safety.

rfk:
Here's a little idea I've been toying with, interested to hear what y'all think...

Since the foreign-language bindings are generated separately from the compiled Rust scaffolding,
user error might lead to a situation where the bindings try to link against an incompatible
version of the compiled code. With luck this would produce a link-time error, but it might
also lead to arbitrary memory unsafety, depending on what's changed between the two sides
of the FFI boundary.

This commit attempts to add an additional layer of safety to prevent the foreign-language
bindings from loading incompatible Rust code. We insert a short checksum into the name of
every single FFI function, which will change if the WebIDL interface changes or if the version
of uniffi changes. Theory being, the bindings can't load and call incompatible FFI functions
if they don't know what they're called!

With this in place, attempts to use foreign-language bindings generated from a different
IDL than the Rust code, or with a different version of uniffi, should fail with an inscrutable
link-time error. But at least they'll fail early and safely.

This may also cause some failures in cases that would have been safe, such as if the two
sounds of the boundary were generated using two different versions of Rust. But if that does
somehow happen, it probably indicates a fragile build process that's at risk of bigger
incompatible probablems, so catching it early seems worthwhile.


uniffi_bindgen/src/interface/mod.rs
```diff
@@ -119,12 +125,62 @@ impl<'ci> ComponentInterface {
...
+    /// Note that this is designed to prevent accidents, not attacks, so there is no need for the
+    /// checksum to be cryptographically secure.
+    ///
+    /// TODO: it's not clear to be if the derivation of `Hash` is actually deterministic enough to
+    /// ensure the guarantees above, or if it might be sensitive to e.g. compiler-driven re-ordering
+    /// of struct field. Let's see how it goes...
+    pub fn checksum(&self) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        // Our implementation of `Hash` mixes in all of the public API of the component,
+        // as well as the version string of uniffi.
```
rfk, line 153: You could imagine a version of this that just does `sha256(uniffi_version + webidl_file_contents)` for a similar effect, rather than (ab)using the `DefaultHasher` like this. I've opted for this approach because I think it's less committal about our ongoing use of WebIDL post-MVP.

linacambridge (approved):
I really like this idea! It’s not exactly name mangling, but reminds me of it—and I also like how it’s a cheap but effective guarantee that both sides of the binding match.

rfk (commented):
@eoger we briefly discussed something like this a little while ago, I'd be interested in your take as well.

eoger (approved):
This is great, I also wonder if we should do something similar with the run functions brought by `uniffi` (although we already have a check for that)

PR #282, opened 2020-09-16
Don't canonicalize paths fed to `cargo metadata`

Gankra:
Cargo has a tendency to mishandle \\\\?\\ paths on windows.
See: https://github.com/rust-lang/cargo/issues/8626

I was having trouble running `cargo build` on windows to check the generated
code, and this seems to fix the issue?


rfk (approved):
Thanks 👍 

I don't think we were canonicalizing here for any particular reason

PR #283, opened 2020-09-16
Add default literals for `optional` arguments

### LINKED ISSUE #263: Figure out optional arguments with defaults
This is similar to the [issue about optional properties in records][1], namely that of default values in optional arguments in function definitions.

There are several times fxa_client uses optional/defaulted arguments.

```idl
ProfileInfo get_profile(optional boolean ignore_cache=false);
string authorize_oauth_code(string client_id, sequence<string> scopes, string state, optional string access_type = "online");
AccessTokenInfo get_access_token(string scope, optional u32 ttl);
```

This is a strawfox proposal implementing them in uniffi to use caller language features to implement them.

```kotlin
pub fun getProfile(ignoreCache: Boolean = false): ProfileInfo =
    callWithHandle { UNIFFI_STRAWFOX_FirefoxAccount_get_profile(it, ignoreCache.lower()) }
    let { ProfileInfo.lift(it) }
```

Note the `= false` default in the method declaration.

Reasons to like this: 

 * documented in the caller language.
 * generated code is very simple

Reason to not like this: 

 * the default value is not contained in rust, but in the IDL file itself.

For reference, in the hand written bindings for `ttl`, [`get_access_token`][2] effectively implemented an `Option<u32>`; the defaulting happened in the rust.

This might be desirable, especially when the bindings are handwritten and hard to keep in sync.

Looking for more experienced rust devs for thoughts: @eoger @linacambridge @rfk ?

[1]: 188
[2]: https://github.com/mozilla/application-services/pull/2896

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1743)

jhugman:
Fixes #263.

This PR fleshes out the parsing and output of Literals, and used in default arguments and defaulted values for dictionaries. This corresponds to the WebIDL keyword `optional`.


uniffi_bindgen/src/bindings/swift/gen_swift.rs
```diff
@@ -121,6 +121,52 @@ mod filters {
...
+                    Radix::Hexadecimal => format!("{:#x}", i),
+                },
+            )?,
+            Literal::UInt(i, radix, type_) => typed_number(
+                type_,
+                match radix {
+                    Radix::Octal => format!("0o{:o}", i),
+                    Radix::Decimal => format!("{}", i),
+                    Radix::Hexadecimal => format!("{:#x}", i),
+                },
```
jhugman, line 164: This level of duplication felt acceptable. 

Using the 8 byte `u64`, `i64` and `f64` as internal representations for number literals was the compromise between lots of duplication, and flexibility of representation.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -358,7 +358,7 @@ extension Dictionary: ViaFfiUsingByteBuffer, ViaFfi, Serializable where Key == S
         let len: UInt32 = try buf.readInt()
         var dict = [String: Value]()
         dict.reserveCapacity(Int(len))
-        for _ in 1...len {
+        for _ in 0..<len {
```
jhugman, line 361: Tests found that empty lists crashes the swift bindings.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -911,33 +921,119 @@ impl APIConverter<Field> for weedle::dictionary::DictionaryMember<'_> {
...
+                // 1. weedle parses "0" as an octal literal, but we most likely want to treat this as a decimal.
+                // 2. Explicitly negatively signed hex numbers won't convert via i64 very well if they're not 64 bit.
+                //    For ease of implementation, output will use decimal.
+                Radix::Decimal
+            } else {
+                radix
+            };
+
+            let trimmer = Regex::new("^(-?)(0x)?").unwrap();
+            let string = trimmer.replace_all(string, "$1").to_lowercase();
```
jhugman, line 985: Possible improvement/optimization: putting this in a `lazy_static!` macro.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -911,33 +921,119 @@ impl APIConverter<Field> for weedle::dictionary::DictionaryMember<'_> {
...
 // Represents a literal value.
 // Used for e.g. default argument values.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum Literal {
     Boolean(bool),
     String(String),
-    // TODO: more types of literal
+    UInt(u64, Radix, Type),
+    Int(i64, Radix, Type),
+    Float(f64, Type),
```
jhugman, line 952: I'm looking for feedback on the internal representation here.

We infer the type of the number from the type of the argument.

Initially I wanted to use the corresponding rust type to be represented, but that meant an awful lot of duplication, especially around radixes and `format!("{}", n)`.

The simplest thing to do would be to use to pass the `String` straight to the bindings, and then let the bindings figure out the differences in parsing and rendering numbers between IDL and their own language; e.g. `e` v `E`, octal llteral. This is not perfect, but this might yet be the right solution.

Presented here is a halfway between these two: using 8 byte numbers to represent signed and unsigned integers, and floats. We have to do a little bit of extra work— mostly involving the regex crate and some extra `format!` calls— but not much; it also brings the small but real possibility of accuracy lost converting between `f64` and `f32`, though I'm not sure how to quantify that.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -911,33 +921,119 @@ impl APIConverter<Field> for weedle::dictionary::DictionaryMember<'_> {
...
+            (weedle::literal::DefaultValue::Boolean(b), Type::Boolean) => Literal::Boolean(b.0),
+            (weedle::literal::DefaultValue::String(s), Type::String) => {
+                Literal::String(s.0.to_string())
+            }
+            (weedle::literal::DefaultValue::EmptyArray(_), Type::Sequence(_)) => {
+                Literal::EmptySequence
+            }
+            (weedle::literal::DefaultValue::String(s), Type::Enum(_)) => {
+                Literal::Enum(s.0.to_string(), type_.clone())
+            }
```
jhugman, line 1028: I was very pleased to work out we could default enums.

uniffi_bindgen/src/interface/mod.rs
jhugman: This PR needs rebasing, and I notice `f64` needs to derive `Hash`. The simplest thing to do here is to move `Literal::Float` to use the weedle parsed `String`.

jhugman (commented):
Leaving this here for feedback before a rebase.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1053,33 +1063,119 @@ impl APIConverter<Field> for weedle::dictionary::DictionaryMember<'_> {
...
-            default: match self.default {
-                None => None,
-                Some(v) => Some(v.value.convert(ci)?),
-            },
+            default,
         })
     }
 }
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, Hash)]
```
rfk, line 1079: Note: as of latest `main`, you'll need to remove `Serialize` and `Deserialize` from here.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1053,33 +1063,119 @@ impl APIConverter<Field> for weedle::dictionary::DictionaryMember<'_> {
...
+            let (string, radix) = match literal {
+                weedle::literal::IntegerLit::Dec(v) => (v.0, Radix::Decimal),
+                weedle::literal::IntegerLit::Hex(v) => (v.0, Radix::Hexadecimal),
+                weedle::literal::IntegerLit::Oct(v) => (v.0, Radix::Octal),
+            };
+            // This is the radix of the parsed number, passed to `from_str_radix`.
+            let src_radix = radix as u32;
+            // This radix tells the backends how to represent the number in the output languages.
+            let dest_radix = if string == "0" || string.starts_with('-') {
+                // 1. weedle parses "0" as an octal literal, but we most likely want to treat this as a decimal.
```
rfk, line 1118: wow; octal zero I hope 😬

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1053,33 +1063,119 @@ impl APIConverter<Field> for weedle::dictionary::DictionaryMember<'_> {
...
         })
     }
 }
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, Hash)]
+pub enum Radix {
+    Decimal = 10,
+    Octal = 8,
+    Hexadecimal = 16,
+}
```
rfk, line 1084: Is preserving the radix necessary for correctness in some cases, or is it entirely about readability of the generated code? (I'm not opposed to doing it for readability reasons, just want to make sure I understand).

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1053,33 +1063,119 @@ impl APIConverter<Field> for weedle::dictionary::DictionaryMember<'_> {
...
+    Null,
 }
 
-impl APIConverter<Literal> for weedle::literal::DefaultValue<'_> {
-    fn convert(&self, _ci: &ComponentInterface) -> Result<Literal> {
-        Ok(match self {
-            weedle::literal::DefaultValue::Boolean(b) => Literal::Boolean(b.0),
-            weedle::literal::DefaultValue::String(s) => Literal::String(s.0.to_string()),
-            _ => bail!("no support for {:?} literal yet", self),
+impl TypedAPIConverter<Literal> for weedle::literal::DefaultValue<'_> {
```
rfk, line 1101: Since you only `impl` this for a single type, I wonder if it would be simpler to have `convert_default_value` as a standalone function to begin with, rather than making it a trait. Or, do you anticipate adding more `impl`s for this shortly?

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1053,33 +1063,119 @@ impl APIConverter<Field> for weedle::dictionary::DictionaryMember<'_> {
...
 // Represents a literal value.
 // Used for e.g. default argument values.
 #[derive(Debug, Clone, Serialize, Deserialize, Hash)]
 pub enum Literal {
     Boolean(bool),
     String(String),
-    // TODO: more types of literal
+    UInt(u64, Radix, Type),
+    Int(i64, Radix, Type),
+    Float(String, Type),
```
rfk, line 1094: Naively, I was expecting to see separate `UInt8`, `UInt16` etc variants on this enum. I don't object to this more succinct representation if it makes the code cleaner, but it may be worth a short comment here to explain the choice since it stands out as being a bit different from other type-related enums.

rfk (commented):
I didn't dig too much into the nitty-gritty parsing details, but did an initial high-level pass and left some comments. Overall, seems to be coming together nicely 👍

uniffi_bindgen/src/interface/mod.rs
jhugman: Preserving the radix is only for readabilty/documentation for the generated interfaces.

There are cases when we can't preserve radix (octal numbers, anyone?), or our choice of representation mean that it's not possible (negative hex and octal numbers); in both these cases we use decimal because correctness beats readability.

uniffi_bindgen/src/interface/mod.rs
jhugman: I'm not anticipating any additional implementations. I've made the change to a standalone function as suggested.

uniffi_bindgen/src/interface/mod.rs
jhugman: :D Yeah. Thought that'd be a bit surprising to our API consumers finding `0o0` in the API.

uniffi_bindgen/src/interface/mod.rs
jhugman: Added a comment about radices.

uniffi_bindgen/src/interface/mod.rs
jhugman: Added extra comments around these representations.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1068,17 +1079,105 @@ impl APIConverter<Field> for weedle::dictionary::DictionaryMember<'_> {
...
+        let dest_radix = if string == "0" || string.starts_with('-') {
+            // 1. weedle parses "0" as an octal literal, but we most likely want to treat this as a decimal.
+            // 2. Explicitly negatively signed hex numbers won't convert via i64 very well if they're not 64 bit.
+            //    For ease of implementation, output will use decimal.
+            Radix::Decimal
+        } else {
+            radix
+        };
+
+        let trimmer = Regex::new("^(-?)(0x)?").unwrap();
```
linacambridge, line 1131: I'm not too worried about this, since we're still prototyping and might eventually use more complex regexes elsewhere—but bringing in `regex` can bump up compile times. For prefix-stripping like this, I wonder if we can use plain-ol' string functions, maybe something like this:

```rust
let string = if string.starts_with('-') {
    ("-".to_string() + string[1..].strip_prefix("0x").unwrap_or(&string[1..])).to_lowercase()
} else {
    string.strip_prefix("0x").unwrap_or(&string).to_lowercase()
};
```

WDYT?

uniffi_bindgen/src/interface/mod.rs
linacambridge: Thanks for the extra comments! I also like your approach of using the widest representation possible, and truncating to the actual type in the binding backends, to simplify the code.

linacambridge (approved):
Wow, this is fantastic, @jhugman! Everything looks solid, and the battery of tests is great.

It would be super helpful to have for the JS and C++ backends, too, which already have some machinery for returning a dummy default value for methods that throw a JS exception—all that code can piggyback on your default literals now.

Ship it! :shipit:

uniffi_bindgen/src/bindings/python/gen_python.rs
```diff
@@ -62,6 +62,18 @@ mod filters {
...
+    pub fn literal_py(literal: &Literal) -> Result<String, askama::Error> {
+        Ok(match literal {
+            Literal::Boolean(v) => format!("{}", v),
+            Literal::String(s) => format!("'{}'", s),
+            Literal::Null => "None".into(),
+            Literal::EmptySequence => "[]".into(),
+            Literal::EmptyMap => "{}".into(),
+
+            _ => panic!("Literal unsupported by python"),
+        })
```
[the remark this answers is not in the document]
jhugman: Add this to the list of things to do _after_ #214  has landed.

uniffi_bindgen/src/interface/mod.rs
jhugman: Oh, magic.

Not since Java 1.3 have I had to think about not having regular expressions.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -84,6 +84,49 @@ mod filters {
...
+
+                Type::Float32 => format!("{}f", num_str),
+                Type::Float64 => num_str,
+                _ => panic!("Unexpected literal: {} is not a number", num_str),
+            })
+        }
+
+        Ok(match literal {
+            Literal::Boolean(v) => format!("{}", v),
+            Literal::String(s) => format!("\"{}\"", s),
```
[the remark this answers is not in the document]
jhugman: 😷 

It appears weedle doesn't let us escape quotes, 

```rust
    /// Represents a string value
    ///
    /// Follow `/"[^"]*"/`
    #[derive(Copy)]
    struct StringLit<'a>(
        &'a str = ws!(do_parse!(
            char!('"') >>
            s: take_while!(|c| c != '"') >>
            char!('"') >>
            (s)
        )),
    )
```

I added some comments at the frontend, and changed the quotes in the python backend.

Issue #268, opened 2020-09-10
Write an ADR for Desktop platform (C++) consumers


linacambridge:
We expect to have two different kinds of consumers in Firefox Desktop: front-end and platform.

Front-end consumers are written in JS, and will require a JS-exposed API (like XPIDL or WebIDL) to call into Rust code. We have a spike for this over in #255, and an ongoing discussion in [this ADR](https://docs.google.com/document/d/1XVBvjt52EKgaaAflMDqUGXQy2AK3sufgEFIu1ZjWbqw/edit?usp=sharing). We have a pretty good understanding of the problem, even though all the solutions are a little gnarly.

Platform consumers are a different beast—think of something like graphics, storage, or a DOM API. These are written in C++, so we’ll want to generate an idiomatic C++ binding for the Rust component that uses Gecko platform types (for example, `mozilla::Maybe` for nullable types, `mozilla::Result` for reporting errors, `mozilla::HashMap` for maps, and `nsTArray` for sequences).

I don’t think we have as good of an understanding of platform consumers, though the solutions should be simpler than for front-end: it’s a lot easier to generate a “vanilla” C++ header and source file that calls into Rust, than it is to generate a WebIDL interface backed by some gnarly C++, all so that JS can call into Rust.

This ticket covers writing up an ADR with options for those platform consumers. The ADR should also cover a path for landing this in Desktop.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1759)

linacambridge:
Well, that wasn’t too hard! 😅 Some copy-pasting and wordsmithing later, here’s the start of a doc: https://docs.google.com/document/d/1kaVliEaNzhLthudBmWSJnmbyC70P0AQqNGE_4K0HYmI/edit

PR #285, opened 2020-09-17
Generate Firefox Desktop platform bindings (C++)

### LINKED ISSUE #269: Prototype bindings for Firefox Desktop platform (C++) consumers
Once we have our ADR for platform consumers (#268), we’ll want to make a prototype. I don’t have many details to add here yet, since we don’t know how we’ll approach this—that’s what the ADR is for! 😁

I think we can reuse the core of #255. The `RustBuffer` serialization machinery is going to be useful here as well, except instead of serializing WebIDL-specific types like `mozilla::dom::Nullable`, `dom::Record`, and `dom::Sequence`, we can serialize `mozilla::Maybe`, `mozilla::HashMap`, and `nsTArray`, respectively. We’ll also need to generate some code for exposing functions, records, and objects to C++, too, like we do for Swift, Kotlin, and Python.

Good news: because we’re only generating a C++ API, and not any IDLs, I suspect we can land this in Desktop quicker than whatever we do for JS.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-18)
┆Issue Number: UNIFFI-18

linacambridge:
Implementing Option A in the ADR from #268. The serialization helpers are copy-pasted from #255, with some of @nnethercote's review feedback incorporated. I'll address the rest tomorrow, fix the comments (many of them are still out-of-date) and request formal review then.

Linking this into Firefox locally is a bit easier than the WebIDL. Just symlink `uniffi-rs` into the tree, run `cargo run -- generate`, and add this to a `moz.build`:

```python
EXPORTS.mozilla += [
    'Arithmetic.h',
]

UNIFIED_SOURCES += [
    'Arithmetic.cpp',
]

TEST_DIRS += [
    '../../../uniffi-rs/examples/arithmetic/tests/gecko/gtest',
]
```

Add the example as a dependency to `toolkit/library/rust/shared/Cargo.toml `:

```toml
uniffi-example-arithmetic = { path = "../../../uniffi-rs/examples/arithmetic" }
```

Link it into `toolkit/library/rust/shared/lib.rs`:

```rust
extern crate uniffi_arithmetic;
```

And run the GTest: `mach gtest 'TestArithmetic.*'`.

Closes #269.


rfk:
Clearing out some issues that we're not actively working on; I'm going to consolidate context on the Gecko bindings into a single new issue at #991

PR #288, opened 2020-09-22
No bug - fixup warnings in cargo test

jhugman:
This PR fixes two sets of annoying warnings seen a lot in the `cargo test`:

```
warning: /Users/jhugman/workspaces/mozilla/uniffi-rs/uniffi_bindgen/Cargo.toml: only one of `license` or `license-file` is necessary
warning: /Users/jhugman/workspaces/mozilla/uniffi-rs/uniffi_macros/Cargo.toml: only one of `license` or `license-file` is necessary
warning: /Users/jhugman/workspaces/mozilla/uniffi-rs/uniffi_build/Cargo.toml: only one of `license` or `license-file` is necessary
warning: /Users/jhugman/workspaces/mozilla/uniffi-rs/uniffi/Cargo.toml: only one of `license` or `license-file` is necessary
```

and 

```
/Users/jhugman/workspaces/mozilla/uniffi-rs/target/debug/rondpoint.swift:127:9: warning: using '_' to ignore the result of a Void-returning function is redundant
        _ = withUnsafeBytes(of: &value) { bytes.append(contentsOf: $0) }
```

This should not change any other behaviour.

### DIFF
--- a/uniffi/Cargo.toml
+++ b/uniffi/Cargo.toml
@@ -6,7 +6,6 @@ homepage = "https://mozilla.github.io/uniffi-rs"
 repository = "https://github.com/mozilla/uniffi-rs"
 version = "0.1.0"
 authors = ["Firefox Sync Team <sync-team@mozilla.com>"]
-license = "MPL-2.0"
 license-file = "../LICENSE"
 edition = "2018"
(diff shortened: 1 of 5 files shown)


linacambridge (approved):
Thanks!

Issue #266, opened 2020-09-09
Remove mentions of “bridging header” from the Swift bindings


linacambridge:
The Swift backend generates a header file with C declarations for all the FFI symbols. `Template-Bridging-Header.h` serves the same purpose as `NamespaceLibraryTemplate.py` and `NamespaceLibraryTemplate.kt`. The Swift compiler uses that header file to generate FFI function declarations (it knows where to look for the header thanks to the `.modulemap`), and expose them to Swift.

But the term “bridging header” means something veeeery specific in Cocoa-land, and I’m pretty sure that’s not what this is 😅 So let’s rename `Template-Bridging-Header.h` to just `NamespaceLibraryTemplate.h` to match the other backends, and scrub all mentions of “bridging header” from the comments to avoid confusion.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-17)
┆Issue Number: UNIFFI-17

rfk:
Oh hey, I did this accidentally over in #984!

PR #289, opened 2020-09-22
Add an example iOS app to prove out crates running on iOS.

### LINKED ISSUE #237: Swift bindings only work in tests
Specifically we include the `import uniffi_{{ ci.namespace() }}` statement at the beginning of these bindings, which I think is there because in tests we compile the Swift code into `libuniffi_{namespace}.dylib` to be able to be used into the Swift REPL. 
In regular production contexts the headers are included in the project and this statement causes a hard error.
@linacambridge I saw you worked on this, does what I describe makes sense?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1655)

jhugman:
Fixes #237.

This PR looks larger than it really is.

The bulk of the file changes are in generating an iOS app from a standard SwiftUI template from Xcode.

Added: an `is_testing` argument passed down to the bindings, so the user doesn't have to think about the difference between `cargo test` and invoking `uniffi-bindgen` in different ways.
Modified: it's not clear to me how module maps work. However, the ones we were generating would use an absolute path for the header. This has changed to use a relative path.
Added: a `module_name` config variable, similar to `package_name` for kotlin package_name.


examples/app/ios/IOSApp/ContentView.swift
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+    }
+}
+
+extension TodoEntry: Hashable {
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(text)
+    }
+}
```
jhugman, line 48: We may want consider generating this for all our records.

examples/app/ios/IOSApp/ContentView.swift
```diff
@@ -0,0 +1,54 @@
...
+
+struct ContentView_Previews: PreviewProvider {
+    static var previews: some View {
+        ContentView(todoList: TodoList())
+    }
```
jhugman, line 53: Wanted: `TodoListProtocol` that foreign languages can mock. See #315 .

examples/app/shared/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "uniffi-example-app"
+edition = "2018"
+version = "0.1.0"
+authors = ["Firefox Sync Team <sync-team@mozilla.com>"]
+license = "MPL-2.0"
+publish = false
+
+[lib]
+crate-type = ["staticlib", "cdylib"]
```
jhugman, line 10: We use `staticlib` for iOS and `cdylib` for testing.

uniffi_bindgen/src/bindings/swift/templates/wrapper.swift
```diff
@@ -2,7 +2,9 @@
 // Trust me, you don't want to mess with it!
 
 import Foundation
-import uniffi_{{ ci.namespace() }}
+{% if is_testing -%}
+import {{ config.module_name() }}
+{% endif -%}
```
jhugman, line 7: This is the real meat of making it work.

The rest is convincing ourselves that it does.

examples/app/ios/IOSApp-Bridging-Header.h
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+#ifndef IOSApp_Bridging_Header_h
+#define IOSApp_Bridging_Header_h
+
+#import "todolist-Bridging-Header.h"
```
eoger, line 9: That reminds me we should really do https://github.com/mozilla/uniffi-rs/issues/266.

examples/app/ios/IOSApp.xcodeproj/project.xcworkspace/xcuserdata/jhugman.xcuserdatad/IDEFindNavigatorScopes.plist
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
eoger, line 1: You might want to delete these `userdata` folders from your commit.

examples/app/shared/tests/bindings/test_todolist.kts
```diff
@@ -0,0 +1,55 @@
+import uniffi.todolist.*
```
eoger, line 1: Do we want to copy the tests too?

uniffi_bindgen/src/bindings/swift/mod.rs
```diff
@@ -31,26 +31,32 @@ pub fn write_bindings(
...
     let mut h = File::create(&header_file).context("Failed to create .h file for bindings")?;
     write!(h, "{}", header)?;
 
     let mut m =
         File::create(&module_map_file).context("Failed to create .modulemap file for bindings")?;
-    write!(m, "{}", generate_module_map(&ci, &header_file)?)?;
+    write!(
+        m,
+        "{}",
+        generate_module_map(config, &ci, &Path::new(&header_filename))?
```
eoger, line 58: I wonder if we need the modulemap also? I don't think we do, so maybe we could in a future patch guard it with `is_testing`?

eoger (commented):
Nice! Do you think it might be worth adding a README to the app dir explainings how we tied everything together in the XCode project? ("No" is a valid answer though, the uniffi goal is not to explain the full rust+ios integration steps).

examples/todolist/Cargo.toml
```diff
@@ -7,7 +7,7 @@ license = "MPL-2.0"
 publish = false
 
 [lib]
-crate-type = ["cdylib"]
+crate-type = ["staticlib", "cdylib"]
```
jhugman, line 10: Generate the `.a` file.

jhugman (commented):
Addressed review comments; removed the second copy of todolist completely. This makes everything a lot tighter. 

There is scope for a README.md in `example/app` or `example/app/ios`, which should cover the targets to install, and the steps to getting it to run in Xcode. This might be a bit too bare bones, given that we'll almost certainly be going through a megazord or packaging via carthage.

Currently we can 

```sh
$ cargo clean 
$ cd example/app/ios
$ xcodebuild clean 
```

Then hitting play on Xcode.

eoger (approved):
The README is superb, thank you!

eoger:
Note: we'll probably want to squash before merging.

Issue #290, opened 2020-09-23
Push a uniffi-rs release


eoger:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1789)

PR #291, opened 2020-09-23
Release v0.2.0

eoger:
Fixes https://github.com/mozilla/uniffi-rs/issues/290.

### DIFF
--- a/uniffi_build/Cargo.toml
+++ b/uniffi_build/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "uniffi_build"
-version = "0.1.0"
+version = "0.2.0"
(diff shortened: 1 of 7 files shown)


uniffi_build/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "uniffi_build"
-version = "0.1.0"
+version = "0.2.0"
```
eoger, line 3: Technically it's a `0.1.1` since there's not breaking changes in `uniffi_build`, but I think it is easier to reason about this by aligning versions.

linacambridge (approved):
:shipit:

Issue #301, opened 2020-09-29
gecko branch is failing tests


dmose:
Happily, @linacambridge told me how to fix it, and I've got a PR forthcoming... it bumps the cargo_metadata dependency version and starts using it again.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1807)

PR #303, opened 2020-09-29
Finish the fix for #301

dmose:
Somehow my commit missed a couple of the Cargo.toml files that I had changed.  So this wants to land there too...

CC @linacambridge


linacambridge (approved):
Whoops, thanks for polishing this off!

linacambridge:
(I see assigning reviewers works for you now, too, yay! 🎉)

Issue #300, opened 2020-09-28
Add optional defaults for arguments and properties in Python


jhugman:
Once #214 lands, defaulting arguments becomes testable in Python (see #283 for equivalent in kotlin and swift).

This issue covers:

* [ ] adding bindings for integer and floating point literals.
* [ ] testing the `Optionneur` object in python
* [ ] ensuring optional arguments work for constructors in Record types. (this will follow the relatively simple work of #188 in other languages)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-27)
┆Issue Number: UNIFFI-27

mhammond:
All this works 🎉

PR #305, opened 2020-09-29
Render all types of literal in the python bindings.

rfk:
I should have rebased https://github.com/mozilla/uniffi-rs/pull/214 before merging, as the new python rondpoint tests interacted badly with the concurrent PR to add support for default arguments. This PR does just enough work to unbreak CI, but doesn't add any tests or similar for the actual use of optional args in python. That work is tracked in https://github.com/mozilla/uniffi-rs/issues/300

### DIFF
--- a/uniffi_bindgen/src/bindings/python/gen_python.rs
+++ b/uniffi_bindgen/src/bindings/python/gen_python.rs
@@ -67,14 +67,26 @@ mod filters {
 
     pub fn literal_py(literal: &Literal) -> Result<String, askama::Error> {
         Ok(match literal {
-            Literal::Boolean(v) => format!("{}", v),
+            Literal::Boolean(v) => {
+                if *v {
+                    "True".into()
+                } else {
+                    "False".into()
+                }
+            }
             // use the double-quote form to match with the other languages, and quote escapes.
             Literal::String(s) => format!("\"{}\"", s),
             Literal::Null => "None".into(),
             Literal::EmptySequence => "[]".into(),
             Literal::EmptyMap => "{}".into(),
-
-            _ => panic!("Literal unsupported by python"),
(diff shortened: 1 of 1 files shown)


uniffi_bindgen/src/bindings/python/gen_python.rs
```diff
@@ -67,14 +67,26 @@ mod filters {
...
     pub fn literal_py(literal: &Literal) -> Result<String, askama::Error> {
         Ok(match literal {
-            Literal::Boolean(v) => format!("{}", v),
+            Literal::Boolean(v) => {
+                if *v {
+                    "True".into()
+                } else {
+                    "False".into()
+                }
+            }
```
jhugman, line 76: Zoinks. Good catch.

uniffi_bindgen/src/bindings/python/gen_python.rs
```diff
@@ -67,14 +67,26 @@ mod filters {
...
             Literal::EmptyMap => "{}".into(),
-
-            _ => panic!("Literal unsupported by python"),
+            Literal::Enum(v, type_) => match type_ {
+                Type::Enum(name) => format!("{}.{}", class_name_py(name)?, enum_name_py(v)?),
+                _ => panic!("Unexpected type in enum literal: {:?}", type_),
+            },
+            // TODO: render with the intended radix.
+            Literal::Int(i, _radix, _type_) => format!("{}", i),
+            Literal::UInt(i, _radix, _type_) => format!("{}", i),
```
jhugman, line 88: #300 will be interested in this.

jhugman (approved):
:shipit: I'm really pleased to see `test_rondpoint.py` running now. ✅

PR #309, opened 2020-10-01
Introduce a "shadow type system" for WebIDL

### LINKED ISSUE #308: Dictionaries can't be nullable in Gecko WebIDL
@piatra found this building the Nimbus bindings in Firefox. The [`NimbusClient` constructor](https://github.com/mozilla/nimbus-sdk/blob/main/nimbus/src/nimbus.idl) takes an `ExperimentConfig? config`, but then Firefox fails to build with:

```
WebIDL.WebIDLError: error: An argument cannot be a nullable dictionary or nullable union containing a dictionary, /Users/lina/Code/gecko/toolkit/components/nimbus/Nimbus.webidl line 39:25
```

So we might either need to mark dictionaries as `optional` instead of nullable (which would only work if the dictionary was the last argument—a non-`optional` argument following an `optional` one would also be a syntax error), _or_ make dictionary arguments non-nullable and force the caller to pass an empty dictionary (`{}`).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1815)

linacambridge:
Although UDL and WebIDL share the same syntax, their semantics are
completely different. WebIDL requires certain keywords where UDL
doesn't, dictionary members are optional by default, and some
constructs, like nullable dictionaries, aren't allowed in WebIDL
at all.

Further, translating UDL types directly into WebIDL without
considering their surrounding context makes for some un-idiomatic
JS interfaces. For example, consider `DOMString?`, meaning a nullable
string. It's OK to take an argument of this type, but a dictionary
member should probably just be `DOMString` instead of
`required DOMString?`, because dictionaries with mandatory but
explicitly `null` properties are awkward and don't read well. In
UniFFI, both "nullable" and "optional" map to the `Optional` type,
while in WebIDL, they have syntactic and semantic differences. But
other dictionary members _may_ need to be `required`.

All these WebIDL-level types map to _different_ C++ types in the
binding, meaning we need the extra context to keep them apart.
(Confusingly, for strings, a `String` and an `Optional(String)`
map to the same C++ type). And some C++ method declarations might
need to take extra arguments, or "out parameters" for returning
complex types.

Expressing all these edge cases, especially when combined, is
impossible to do in the Askama templates and filters.

`webidl.rs` already had some extension traits to help with declaring
C++ arguments and methods. This commit extends them into a fuller
"shadow type system" that wraps the UniFFI types into the more
pedantic types that WebIDL expects.

With this commit, it should be possible to compile an idiomatic
Nimbus binding (no compilation errors, and no objects filled with
`null`s) into Firefox Desktop.

Closes #308, and probably some other undiscovered bugs.


linacambridge:
WDYT @rfk? I don't know if this is the best way to go about it, but after spending some time trying to do all this in the Askama filters, I felt like we'd need a full map of UniFFI types to WebIDL types eventually...

rfk:
> WDYT @rfk? 

To a first approximation, I'm just relieved that the only changes here are under `./gecko_js/` 😅 

I'll try to dig into this later today.

uniffi_bindgen/src/bindings/gecko_js/webidl.rs
```diff
@@ -2,171 +2,537 @@
...
+    Interface(String),
+
+    /// Dictionaries correspond to UniFFI records.
+    Dictionary(String),
 
-use crate::interface::{Argument, Constructor, Function, Method, Type};
+    /// `Nullable` and `Optional` both correspond to UniFFI optional types.
+    /// Semantically, "nullable" means "must be passed as an argument or a
+    /// dictionary member, but can be `null`". "Optional" means the argument
+    /// or member can be omitted entirely.
```
rfk, line 56: I don't know where to capture it, but the way I explain this to myself is "most languages don't have both `null` and `undefined`" but the difference is very important to javascript.

uniffi_bindgen/src/bindings/gecko_js/webidl.rs
```diff
@@ -2,171 +2,537 @@
...
+//! These combinations of special cases are impossible to express in Askama, so
+//! we have a "shadow type system" that wraps the UniFFI interface types.
+//! Capturing this logic here lets us keep our templates and filters clean.
+
+use crate::interface::{
+    Argument, Constructor, FFIFunction, FFIType, Field, Function, Literal, Method, Object, Record,
+    Type,
+};
+
+/// WebIDL types correspond to UniFFI interface types, s
```
rfk, line 39: trailing `,s` on the end of here, makes we wonder if it's got something important to say :-)

After our conversation earlier, in my head I was lightly expecting to see this implemented as a *trait* rather than as a wrapper type. Is there any interesting context about why a wrapper type makes more sense that is worth capturing in comments here?

(I'm not asking you to try to change it to use traits or anything like that, and "no, there is no interesting context" is a perfectly acceptable answer).

uniffi_bindgen/src/bindings/gecko_js/templates/WebIDLTemplate.webidl
```diff
@@ -39,16 +42,16 @@ namespace {{ context.namespace()|class_name_webidl(context) }} {
 };
 {% endif -%}
 
-{%- for obj in ci.iter_object_definitions() %}
+{%- for obj in interfaces %}
```
rfk, line 45: nit: consider naming the variable here `webidl_interfaces` or similar, by symmetry with the method names like `webidl_arguments`. I quite like the way they call out explicitly that we're working with the shadow types rather than the usual ones.

uniffi_bindgen/src/bindings/gecko_js/mod.rs
```diff
@@ -75,15 +75,38 @@ pub fn generate_bindings(config: &Config, ci: &ComponentInterface) -> Result<Vec
...
+    let functions: Vec<_> = ci
+        .iter_function_definitions()
+        .into_iter()
+        .map(webidl::WebIDLFunction::new)
+        .collect();
+    let interfaces: Vec<_> = ci
+        .iter_object_definitions()
+        .into_iter()
+        .map(webidl::WebIDLInterface::new)
+        .collect();
```
rfk, line 93: You could also consider grouping these together into some sort of `GeckoComponentInterface` struct, by analogy to our `ComponentInterface` struct. Might make the mental model coalesce a little more completely, and make them easier to pass around. It's fine as-is though, if that doesn't sound worthwhile to you.

rfk (approved):
Overall, I like this a lot, feels like a good way to wrangle the complexity and it kind of "plugs in" to my mental model of the `ComponentInterface` abstractions in a sensible way. I actually wonder if leaning more in to this style could help clean up some of the other language bindings! (not for this PR though...)

uniffi_bindgen/src/bindings/gecko_js/webidl.rs
linacambridge: Oops, half-complete comment, how embarrassing! 😆

I ran with your idea of using extension traits for `Function`, `Field`, `Argument`, etc.—that actually makes things a lot cleaner, since we don't need to wrap `Object` and `Record` 👍🏻😁  I think that was how I had it originally, and then during some intermediate refactoring (before I added `WebIDLType`) switched to wrappers. I'm not sure `WebIDLType` can be a trait, though—it expands "nullable" to "nullable and optional", so we need to hold an extra variant, and since it's recursive, I think we need to redefine those variants.

I did move `Interface` and `Dictionary` into `Wrapped` (and renamed it to `NonRecursive`), since there's no reason to have those separate. And I renamed `Record` back to `Map`, since that's the UniFFI type name for it.

Thanks for encouraging me to think more about modeling this!

uniffi_bindgen/src/bindings/gecko_js/webidl.rs
linacambridge: Oh, that's a great summary, I added that to the comment! (As if that wasn't confusing enough, there's an observable difference between explicitly passing `undefined` and omitting it entirely—`{ foo: undefined }` and `{}`—but thankfully all the specs seem to be moving away from that).

lattice0:
why the gecko_js binding was deleted?

mhammond:
Because we never got it working and decided on a different approach

PR #310, opened 2020-10-01
Consolidate and fix logic for guessing the crate root.

rfk:
This fixes some bustage on CI on current `main`. Also adds a test for completeness.

### DIFF
--- a/uniffi_bindgen/src/lib.rs
+++ b/uniffi_bindgen/src/lib.rs
@@ -161,11 +161,7 @@ fn ensure_versions_compatibility(
             metadata_cmd.manifest_path(p);
         }
         None => {
-            metadata_cmd.current_dir(
-                idl_file
-                    .parent()
-                    .ok_or_else(|| anyhow!("IDL file has no parent folder!"))?,
-            );
+            metadata_cmd.current_dir(guess_crate_root(idl_file)?);
         }
     };
     let metadata = metadata_cmd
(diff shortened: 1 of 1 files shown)


uniffi_bindgen/src/lib.rs
```diff
@@ -267,11 +280,8 @@ fn get_config(
     let config_file: Option<PathBuf> = match config_file_override {
         Some(cfg) => Some(PathBuf::from(cfg)),
         None => {
-            let path_guess = idl_file
-                .parent()
-                .ok_or_else(|| anyhow!("IDL file has no parent folder!"))?
-                .join("uniffi.toml");
```
rfk, line 273: We need to `parent()` twice here, once for the filename and once for the `src` directory.

uniffi_bindgen/src/lib.rs
```diff
@@ -161,11 +161,7 @@ fn ensure_versions_compatibility(
             metadata_cmd.manifest_path(p);
         }
         None => {
-            metadata_cmd.current_dir(
-                idl_file
-                    .parent()
-                    .ok_or_else(|| anyhow!("IDL file has no parent folder!"))?,
```
rfk, line 167: This was incorrectly setting the current directory to `./src` rather than the crate root, but `cargo metadata` was walking the directory tree to find the closest enclosing `Cargo.toml` file, so it didn't cause a bug in practice.

eoger (approved):
Great factorization, thanks!

PR #314, opened 2020-10-03
Special-case "optionals with a default value" and fix compilation errors in Gecko JS

linacambridge:
Optional arguments and members with a default value are a bag of special cases.

On the C++ side, `optional` arguments with default values are passed as `const T&`, and optional dictionary members are type `T` (for optional arguments, they're `const dom::Optional<T>&` and `dom::Optional<T>`, respectively). But the Rust side is expecting them to be serialized as `Option<T>`. It's basically the same problem as strings, where null and non-null strings have the same `const nsAString&`/`nsString` type—the Rust side is expecting to read `1`/`0` followed by the serialization of the type, but the C++ side isn't passing that `1`/`0`.

So the first commit adds a special `WebIDLType::OptionalWithDefaultValue` that's emitted as `T`, except when we call `ViaFfi`, where it passes `Nullable = true`. It also adds a helper for serializing non-null things on the C++ side as if they were `Some<T>` for Rust.

The second commit fixes some errors in the generated Nimbus bindings. `dom::Optional<T>` can't be constructed directly, and `ThrowOperationError` takes a `const ACString&`, not a raw `char*`, so we have to wrap it in an `nsDependentCString`.

@dmose and I paired on this yesterday, and got the Nimbus SDK building with these changes. We haven't written an xpcshell test for it yet—it's crashing because it's trying to use Viaduct to make a network request from the main thread, [which is forbidden](https://searchfox.org/mozilla-central/rev/7ef5cefd0468b8f509efe38e0212de2398f4c8b3/toolkit/components/viaduct/Viaduct.cpp#45)—but it compiles and runs.

### DIFF
--- a/uniffi_bindgen/src/bindings/gecko_js/webidl.rs
+++ b/uniffi_bindgen/src/bindings/gecko_js/webidl.rs
@@ -50,6 +50,13 @@ pub enum WebIDLType {
     Nullable(Box<WebIDLType>),
     Optional(Box<WebIDLType>),
 
+    /// Optionals with a default value are a grab bag of special cases in Gecko.
+    /// In the generated C++ bindings, the type of an optional with a default
+    /// value is `T`, not `Optional<T>`. However, it must be serialized as if
+    /// it's an `Optional<T>`, since that's what the Rust side of the FFI
+    /// expects.
+    OptionalWithDefaultValue(Box<WebIDLType>),
+
     /// Sequences are the same as their UniFFI counterparts.
     Sequence(Box<WebIDLType>),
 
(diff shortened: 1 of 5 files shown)


uniffi_bindgen/src/bindings/gecko_js/webidl.rs
```diff
@@ -50,6 +50,13 @@ pub enum WebIDLType {
     Nullable(Box<WebIDLType>),
     Optional(Box<WebIDLType>),
 
+    /// Optionals with a default value are a grab bag of special cases in Gecko.
+    /// In the generated C++ bindings, the type of an optional with a default
+    /// value is `T`, not `Optional<T>`. However, it must be serialized as if
+    /// it's an `Optional<T>`, since that's what the Rust side of the FFI
+    /// expects.
```
eoger, line 57: What a headache, I wonder if we shouldn't have disallowed optionals without default values in IDL files.

eoger (approved):
Ship it!
Dancing kid

PR #316, opened 2020-10-06
Add defaults for properties in Records

### LINKED ISSUE #188: Figure out optional dictionary fields

jhugman:
Fixes #188.

This small PR adds to #283.

### DIFF
--- a/examples/rondpoint/tests/bindings/test_rondpoint.kts
+++ b/examples/rondpoint/tests/bindings/test_rondpoint.kts
@@ -57,6 +57,7 @@ listOf(-1, 0, 1).map { DictionnaireNombresSignes(it.toByte(), it.toShort(), it.t
 listOf(0, 1).map { DictionnaireNombres(it.toUByte(), it.toUShort(), it.toUInt(), it.toULong()) }
     .affirmAllerRetour(rt::identiqueNombres)
 
+
 rt.destroy()
 
 // Test one way across the FFI.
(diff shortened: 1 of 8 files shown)


uniffi_bindgen/src/bindings/python/templates/RecordTemplate.py
```diff
@@ -1,5 +1,12 @@
 class {{ rec.name()|class_name_py }}(object):
-    def __init__(self,{% for field in rec.fields() %}{{ field.name()|var_name_py }}{% if loop.last %}{% else %}, {% endif %}{% endfor %}):
+    def __init__(self, {% for field in rec.fields() -%}
+        {{ field.name()|var_name_py -}}
+                {%- match field.default_value() %}
+            {%- when Some with(literal) %} = {{ literal|literal_py }}
+            {%- else %}
+        {%- endmatch -%}
```
rfk, line 7: FYI, this will cause problems in python if there are non-default fields following a default field:

```
>>> def test(one, two=2, three):
...   pass
  File "<stdin>", line 1
SyntaxError: non-default argument follows default argument>>>                                                       
```

rfk (approved):
Wasn't sure if this was ready for review, but it seems reasonable to me 👍

I don't mind if we punt on the argument-ordering issue in python, as long as it works sensible in Kotlin and Swift.

uniffi_bindgen/src/bindings/python/templates/RecordTemplate.py
jhugman: Yuck. Ok. This was meant to be a quick PR; I think I favour pulling out defaults until we have the bandwidth for #300.

examples/rondpoint/tests/bindings/test_rondpoint.kts
```diff
@@ -57,6 +57,7 @@ listOf(-1, 0, 1).map { DictionnaireNombresSignes(it.toByte(), it.toShort(), it.t
 listOf(0, 1).map { DictionnaireNombres(it.toUByte(), it.toUShort(), it.toUInt(), it.toULong()) }
     .affirmAllerRetour(rt::identiqueNombres)
 
+
```
jhugman, line 60: Nit: whitespace

jhugman:
> I don't mind if we punt on the argument-ordering issue in python

Removed the python implementation in favour of #300.

PR #317, opened 2020-10-06
Adds public protocols/interfaces for Objects.

### LINKED ISSUE #315: Generate interfaces for mocking and previewing Objects.
Uniffi objects (and namespace functions) manage the calling into rust and getting output from rust.

This makes it difficult to mock for testing, or use as preview objects when using SwiftUI or Jetpack Compose.

This issue is for foreign language developer ergonomics. It isn't *necessary* to help rust code deployed, but helps uniffi users to develop code around the Rust code.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1822)

jhugman:
Fixes #315 

Adds simple bindings for kotlin and swift to add interfaces to allow for mocking, testing and previewing.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift
@@ -1,4 +1,4 @@
-public struct {{ rec.name()|class_name_swift }}:  ViaFfiUsingByteBuffer, ViaFfi, Equatable {
+public struct {{ rec.name()|class_name_swift }}:  ViaFfiUsingByteBuffer, ViaFfi, Equatable, Hashable {
     {%- for field in rec.fields() %}
     let {{ field.name()|var_name_swift }}: {{ field.type_()|type_swift }}
     {%- endfor %}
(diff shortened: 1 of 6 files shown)


uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift
```diff
@@ -1,4 +1,4 @@
-public struct {{ rec.name()|class_name_swift }}:  ViaFfiUsingByteBuffer, ViaFfi, Equatable {
+public struct {{ rec.name()|class_name_swift }}:  ViaFfiUsingByteBuffer, ViaFfi, Equatable, Hashable {
```
rfk, line 1: I don't object to it, but I'm curious why you needed to add `Hashable` here.

rfk (approved):
Thanks! I think this makes sense, and it mirrors what we've done by hand in some of the existing rust components like logins.

rfk:
I also think it would be useful to document expectations here for consumers; maybe there's a place in the [the manual](https://github.com/mozilla/uniffi-rs/tree/main/docs/manual) where it would make sense to add that?

uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift
jhugman: No real need to do it here, this PR came out of some developer ergonomics when building an iOS app.

Issue #293, opened 2020-09-25
Change `.idl` extension to `.udl`


linacambridge:
Firefox uses the `.idl` extension for XPIDL, and `.webidl` for WebIDL. UniFFI IDL uses WebIDL syntax, but not the semantics...and the `.idl` extension. Dan suggested changing the extension to `.uidl` to help clarify that it's our own dialect of WebIDL that we're using, not "standard" WebIDL and definitely not XPIDL.

WDYT?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1795)

linacambridge:
If you pronounce "UIDL" as "oo-ee-dl", it sounds like `weedle`, which is the [Rust WebIDL parser](https://crates.io/crates/weedle) that we use! 😁👍🏻

rfk:
I support it 👍 

(Mentally, I am going to associate it with these [super classy pre-mixed vodka drinks called UDLs](https://www.danmurphys.com.au/search?searchTerm=udl), so I'll have to try really hard to remember the "I" is in there; but I can live with that...)

linacambridge:
Or we could go with `.udl`, we're fancy, too! 🍸🥤

jhugman:
I support this message. 👍 

Does this mean we can rule out supporting `DOMString`?

linacambridge:
> Does this mean we can rule out supporting `DOMString`?

I _think_ so! Making sure I understand, though, by supporting `DOMString`, did you mean switching from `string` to `DOMString`, or something else?

linacambridge:
Oh, for completeness, I wanted to capture some of the differences between UDL and ["standard" WebIDL](https://heycam.github.io/webidl/). Some of these are described in a comment in the Gecko JS backend, but it might be helpful to capture them in prose, too:

* UDL has the following aliases for numeric types: `u8`, `i8`, `u16`, `i16`, `u32`, `i32`, `u64`, `i64`, `f32`, `f64`. WebIDL calls these `octet`, `byte`, `unsigned short`, `short`, `unsigned long`, `long`, `unsigned long long`, `float`, `double`.
* UDL doesn't support [`unrestricted float`](https://heycam.github.io/webidl/#idl-unrestricted-float) or [`unrestricted double`](https://heycam.github.io/webidl/#idl-unrestricted-double).
[...]

PR #319, opened 2020-10-07
Rename idl to udl

eoger:
Fixes https://github.com/mozilla/uniffi-rs/issues/293.

This is a big breaking change, but it's worth doing before uniffi deploys in Mozilla's projects.


docs/manual/src/Overview.md
```diff
@@ -6,8 +6,8 @@ Note that this tool will not help you ship a Rust library to these platforms, bu
 
 ## Design
 
-uniffi requires to write an Interface Definition Language ([IDL](https://heycam.github.io/webidl/)) file describing the methods and data structures available to the targeted languages.  
-This .idl file, whose definitions must match with the exposed Rust code, is then used to generate Rust *scaffolding* code and foreign-languages *bindings*. This process can take place either during the build process or be manually initiated by the developer.
+uniffi requires to write an Interface Definition Language (loosely based on [WebIDL](https://heycam.github.io/webidl/)) file describing the methods and data structures available to the targeted languages.
```
jhugman, line 9: Syntactically the same, but with different standard types and annotations.

README.md
```diff
@@ -46,11 +46,11 @@ continues to evolve.
 Using `uniffi`, you can:
 
 * Implement your software component as a `cdylib` crate in Rust; let's say the code is in `./src/lib.rs`.
-* Specify the desired component API using an *Interface Definition Language* (specifically, a variant of WebIDL) in a separate file like `./src/lib.idl`.
-* Run `uniffi-bindgen scaffolding ./src/lib.idl` to generate a bunch of boilerplate rust code that exposes this API as a C-compatible FFI layer.
+* Specify the desired component API using an *Interface Definition Language* (specifically, a variant of WebIDL) in a separate file like `./src/lib.udl`.
```
jhugman, line 49: We don't say anywhere what `UDL` stands for.

jhugman (approved):
This looks good. 

Call out that we don't define what UDL stands for anywhere.

rfk (approved):
Thanks you for working through all these details!

PR #325, opened 2020-10-12
Make Rust warnings break the build

dmose:


dmose:
Hmm, that didn't do it.   I bet I need to put this inside the `profile.sh` in the Docker image...

rfk (approved):
👍 thanks!

PR #332, opened 2020-10-21
Update dependencies

piatra:
I couldn't find much in our dependencies that could be removed (they also don't have many sub-dependencies), only did some small updates.
I also did a license check
```
Apache-2.0 OR BSL-1.0 (1): ryu
Apache-2.0 OR MIT (31): anyhow, askama, askama_derive, askama_escape, askama_shared, bitflags, cfg-if, ffi-support, glob, heck, itoa, lazy_static, log, paste, pest, proc-macro2, quote, semver, semver-parser, serde, serde_derive, serde_json, syn, thiserror, thiserror-impl, toml, ucd-trie, unicode-segmentation, unicode-width, unicode-xid, version_check
MIT (6): bytes, cargo_metadata, clap, nom, textwrap, weedle
MIT OR Unlicense (1): memchr
MPL-2.0 (9): uniffi, uniffi-example-arithmetic, uniffi-example-geometry, uniffi-example-rondpoint, uniffi-example-sprites, uniffi-example-todolist, uniffi_bindgen, uniffi_build, uniffi_macros
```
Everything is compatible according to [our license compatibility list](https://www.mozilla.org/en-US/MPL/license-policy/#Licenses_Compatible_with_the_MPL). [`memchr`](https://github.com/BurntSushi/rust-memchr) (a dependency of `askama`) has a double MIT and Unlicense license.

I'm not sure if we want to check in the `Cargo.lock` file (currently ignored)?
- [ ] todo maybe

### DIFF
--- a/docs/manual/src/tutorial/Rust_scaffolding.md
+++ b/docs/manual/src/tutorial/Rust_scaffolding.md
@@ -8,7 +8,7 @@ First, add `uniffi` to your crate dependencies: it is the runtime support code t
 
 ```toml
 [dependencies]
-uniffi = "0.3"
+uniffi = "0.4"
 ```
 
 Important note: the `uniffi` version must be the same as the `uniffi-bindgen` command-line tool installed on your system.
@@ -17,7 +17,7 @@ Then let's add `uniffi_build` to your build dependencies: it generates the Rust 
 
 ```toml
 [build-dependencies]
-uniffi_build = "0.3"
+uniffi_build = "0.4"
 ```
 
 Then create a `build.rs` file next to `Cargo.toml` that will use `uniffi_build`:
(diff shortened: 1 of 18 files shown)


rfk:
> I'm not sure if we want to check in the Cargo.lock file (currently ignored)?

Hrm, good question. I deliberately didn't check it in initially, because in my head `uniffi` is a library for use by other crates, and [libraries shouldn't check in their `Cargo.lock`](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries). But in its current usage uniffi is more like a command-line tool, which IIUC typically *would* check in `Cargo.lock` to version control.

On balance I think it' would be better for us to check it in, so that tagged releases of uniffi-bindgen come with a consistent set of dependencies. Any other opinions/objections?

rfk (approved):

rfk:
I don't think we need to block this on the `Cargo.lock` question, going to merge if CI is green.

Issue #334, opened 2020-10-26
Revert struct-padding workaround once upstream bug is fixed in JNA


rfk:
The code generated by uniffi makes heavy use of passing structs by value, and this seems to have uncovered an [argument-passing bug in JNA](https://github.com/java-native-access/jna/issues/1259). We landed a workaround for this bug in https://github.com/mozilla/uniffi-rs/pull/335 but it works by artificially padding our structs, so it seems likely to have some runtime performance costs.

(Briefly: it pads our structs to be more than 16 bytes in size, so that they will always be passed as a pointer, which works around the bug).

Once the underlying issue has been resolved in a JNA release, we should revert the above workaround.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-33)
┆Issue Number: UNIFFI-33

rfk:
The latest release of JNA (5.8) has an updated version of `libffi` and appears to have fixed the issue!

rfk:
I believe this bug is now actionable, and could make a good first issue. Basically, we need to:

* Remove the `padding` and `padding2` fields that were added in #335 (there are other cleanups in that PR, so we shouldn't revert the whole thing). This will need to be done in both the Rust definitions, and the various foreign-language bindings.
* Document somewhere in the manual that we require JNA 5.7 or greater; I don't think we have a good place for this yet, but it could go somewhere in the "Kotlin" section.
* Update `docker/Dockerfile-build` to install the newer version of JNA; this won't have any effect on tests passing because we weren't able to repro that crash in CI, but it'll be useful for completeness.

We'll also have to keep a careful eye out for any increase in crashes on Android once the new version rolls out to consumers.

badboy:
I will take a shot at this.

rfk:
sounds good - especially getting your head around the docker image and tests will be valuable as well

PR #335, opened 2020-10-26
Add workaround for JNA bug when passing small structs by value.

rfk:
The current version of JNA appears to have a bug when passing small structs
as arguments by value, on arm64 platforms, when the function call has too
many arguments to pass the struct in registers. It turns out that our generated
code does this a lot thanks to the way we pass around `RustBuffer` structs.

This commit adds a temporary workaround for the issue, by padding our structs
to be larger than 16 bytes, which causes them to be passed as pointers rather
than by value.

See https://github.com/mozilla/application-services/issues/3646 for investigation
of the JNA issue, and https://github.com/mozilla/uniffi-rs/issues/334 to track
the work of removing this workaround once a fix is released in upstream JNA.

### DIFF
--- a/uniffi/src/ffi/foreignbytes.rs
+++ b/uniffi/src/ffi/foreignbytes.rs
@@ -31,9 +31,30 @@ pub struct ForeignBytes {
     len: i32,
     /// The pointer to the foreign-owned bytes.
     data: *const u8,
+    /// This forces the struct to be larger than 16 bytes, as a temporary workaround for a bug in JNA.
+    /// See https://github.com/mozilla/uniffi-rs/issues/334 for details.
+    padding: i64,
+    padding2: i32,
 }
 
 impl ForeignBytes {
+    /// Creates a `ForeignBytes` from its constituent fields.
+    ///
+    /// This is intended mainly as an internal convenience function and should not
+    /// be used outside of this module.
+    ///
+    /// # Safety
+    ///
+    /// You must ensure that the raw parts uphold the documented invariants of this class.
+    pub unsafe fn from_raw_parts(data: *const u8, len: i32) -> Self {
+        Self {
+            len,
+            data,
+            padding: 0,
+            padding2: 0,
+        }
+    }
+
     /// View the foreign bytes as a `&[u8]`.
     ///
     /// # Panics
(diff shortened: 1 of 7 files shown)


rfk:
@travis79 I made a local build of the glean+nimbus sample app using this patched version of uniffi, and AFAICT it resolves the arm64 bug I was seeing with invalid memory reads. Any chance you could try this out on your device that was also experiencing a crasher, and see if it works for you?  You'd need to:

* Configure your nimbus-sdk rust project to use this version of uniffi, e.g. using a [git commit dependency](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories).
* Install this version as `uniffi-bindgen` in your path (so it can be found by the gradle script to generate Kotlin bindings.

I'm confident enough that this resolves a real issue that I think we should proceed with merging and cutting a new uniffi release that includes it, but it would be worth getting confirmation from your device as well for completeness.

rfk:
Here's the concrete repro of the bug, for context: https://github.com/rfk/arm64-jna-structbug

rfk:
Also for context, the upstream bug report: https://github.com/java-native-access/jna/issues/1259

eoger (approved):
Oh man this is a pretty bad bug, good job investigating it!

travis79:
> @travis79 I made a local build of the glean+nimbus sample app using this patched version of uniffi, and AFAICT it resolves the arm64 bug I was seeing with invalid memory reads. Any chance you could try this out on your device that was also experiencing a crasher, and see if it works for you? You'd need to:
> 
>     * Configure your nimbus-sdk rust project to use this version of uniffi, e.g. using a [git commit dependency](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories).
> 
>     * Install this version as `uniffi-bindgen` in your path (so it can be found by the gradle script to generate Kotlin bindings.
> 
> 
> I'm confident enough that this resolves a real issue that I think we should proceed with merging and cutting a new uniffi release that includes it, but it would be worth getting confirmation from your device as well for completeness.

This is a little non-trivial to test locally because the application I was testing this in was depending on the nimbus in the appservices megazord.  Let me see if I can build and publish the megazord locally.  I trust your judgement that this resolves the issue if we want to go ahead and merge this, it might make it a little easier for me to confirm if it percolated down to the prototype megazord with nimbus in it.

rfk:
@travis79 FYI the 0.4.0 release is out with this fix included.

PR #339, opened 2020-11-04
Fix #338 (JS bindings crash) by padding struct in C++ template

### LINKED ISSUE #338: JNA workaround struct changes didn't get propagated to C++ templates
In #335, a workaround for a JNA bug was added by padding out the RustBuffer and ForeignBytes code.  The workaround was propagated to most of the templates (Kotlin, etc), but the not to the C++ code used in the JS bindings.  Patch forthcoming which fixes that.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1872)

dmose:


uniffi_bindgen/src/bindings/gecko_js/templates/FFIDeclarationsTemplate.h
```diff
@@ -4,11 +4,14 @@ struct {{ context.ffi_rustbuffer_type() }} {
   int32_t mCapacity;
   int32_t mLen;
   uint8_t* mData;
+  uint64_t mPadding;
```
rfk, line 7: nit: I think this should be `int64_t` rather than `uint64_t` (which I don't think should actually matter in practice but hey, who knows...)

rfk (approved):
Thanks for tracking this down!

In the other bindings backends, I left a little breadcrumb comment like:

```
// Ref https://github.com/mozilla/uniffi-rs/issues/334 for the extra "padding" arg.
```

In an attempt to make sure future readers would be able to find the right context. Please consider adding it here as well for consistency.

uniffi_bindgen/src/bindings/gecko_js/templates/FFIDeclarationsTemplate.h
dmose: Good catch; fixed.

PR #340, opened 2020-11-04
Flush struct memory when writing to a RustBuffer.

rfk:
I need to understand better what's happening here, but pushing this for visibility.

@jhugman please let me know if this fixes the issue you were having with long strings in callbacks.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/RustBufferTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/RustBufferTemplate.kt
@@ -64,7 +64,7 @@ class RustBufferBuilder() {
 
     init {
         val rbuf = RustBuffer.alloc(16) // Totally arbitrary initial size
-        rbuf.len = 0
+        rbuf.writeField("len", 0)
         this.setRustBuffer(rbuf)
     }
 
@@ -78,7 +78,10 @@ class RustBufferBuilder() {
 
     fun finalize() : RustBuffer.ByValue {
         val rbuf = this.rbuf
-        rbuf.len = this.bbuf!!.position()
+        // Ensure that the JVM-level field is written through to native memory
+        // before turning the buffer, in case its recipient uses it in a context
+        // JNA doesn't apply its automatic synchronization logic.
+        rbuf.writeField("len", this.bbuf!!.position())
         this.setRustBuffer(RustBuffer.ByValue())
         return rbuf
     }
@@ -95,7 +98,7 @@ class RustBufferBuilder() {
         // here, trying the write and growing if it throws a `BufferOverflowException`.
         // Benchmarking needed.
         if (this.bbuf!!.position() + size > this.rbuf.capacity) {
-            this.rbuf.len = this.bbuf!!.position()
+            rbuf.writeField("len", this.bbuf!!.position())
             this.setRustBuffer(RustBuffer.reserve(this.rbuf, size))
         }
         write(this.bbuf!!)


rfk:
So as far as I can tell (e.g. reading [the overview](https://java-native-access.github.io/jna/4.2.0/overview-summary.html#structures)) JNA implements structures by allocating a chunk of native memory for the structure, and synchronizing field values from the JVM into and out of that flat native memory. It *mostly* does this automagically by default, e.g. writing out JVM field values before passing a struct as an argument to a function call and reading them back after the call.

I think @jhugman may have been hitting a case where this wasn't happening automatically, when returning a structure from a callback. If true, that might plausibly be considered a bug in JNA, or at least a surprising behaviour, since it seems reasonable to me that it could apply auto-sync behaviour that that poing. I'd have to dig in to more details to confirm.

On the other hand, it's quite rare that we want to be able to modify struct fields from the Kotlin side, and there may be some small performance benefit to turning off this automatic syncing and instead doing it by hand in the few places where we need it (such as the places touched by this PR).

jhugman:
> hitting a case where this wasn't happening automatically, when returning a structure from a callback […]
> [or] it's quite rare that we want to be able to modify struct fields from the Kotlin side

So a couple of questions:

 1. Can we tell the difference between these theories? e.g. I am not able reproduce it by sending long strings to Rust as part of complex types (Option or Vec). 
 2. Should we use `writeField` more liberally, e.g. for `capacity`?
 3. Is your performance benefit about using` volatile` for `capacity` or `len`, and only use `writeField` during `finalize()`?

rfk:
> Can we tell the difference between these theories? e.g. I am not able reproduce it by sending long strings to Rust as part of
> complex types (Option or Vec). 

I'm not sure, I haven't dug into the details enough. I think the issue would only surface when we write to a struct field at the Kotlin level, and basically the only place we do that is here where we modify `len`.

> 2. Should we use writeField more liberally, e.g. for capacity?

I don't believe we ever write to `capacity` or `data` from Kotlin, these are managed by the native code.

> 3. Is your performance benefit about using volatile for capacity or len, and only use writeField during finalize()?

Maybe both 😁 

For example, what I *think* is happening with this PR where we have:

```
            rbuf.writeField("len", this.bbuf!!.position())
            this.setRustBuffer(RustBuffer.reserve(this.rbuf, size))
```

Is something like:

1. We write the new length into the `len` object field in the JVM and to the corresponding location native memory.
2. We pass the struct as an argument to a native function:
    2.1 JNA copies the value from `capacity`, `len` and `data` from the object's fields in the JVM into the native memory.
    2.2 JNA calls the native function, passing the native memory.
    2.3 JNA copies the values from native memory back into the `capacity`, `len` and `data` object fields in the JVM.

The copies done in (2.1) are entirely superfluous, since we should never be changing `capacity` or `data` from Kotlin code, and we already wrote `len` in the step above.

Alternatively, the write to native memory done in (1) is not necessary because JNA will do it for us automatically in (2.1). The one place where we can't count on JNA to do it for us is in `finalize`, so we could change this to only use `writeField` there.

Both of these are probably small potatoes in terms of performance though, so I'm tempted to say let's leave this PR as it is in the interests of being obviously correct.

rfk:
> I'm tempted to say let's leave this PR as it is in the interests of being obviously correct.

In fact, I'm going to take the "WIP" off of this given that it fixes the issue you were seeing. r?

jhugman:
This fixes the issue I'm seeing. Thank you. 

Agreed, this isn't an opportunity to optimize, just understand a little better what is going on. It might be worth filing a bug with https://github.com/java-native-access/jna when we can isolate the bug (eventually), but I'm happy to r+ this now.

jhugman (approved):
:ship: :shipit: 🚀 ✅

PR #341, opened 2020-11-05
Update cargo_metadata dependency.

rfk:
This will avoid us pulling in multiple versions of some of our dependencies when used in conjunction with this PR: https://github.com/mozilla/nimbus-sdk/pull/49#discussion_r517688219


mhammond (approved):
:stamp:

Issue #342, opened 2020-11-05
Scope out adding support for callback interfaces


jhugman:
### Problem

1. Currently application-services uses viaduct to perform network calls (so Rust can use a `concept-fetch` android-component).
2. Nimbus is calling `RemoteSettings` an HTTP request directly, but we'd like to be able to use a `RemoteSettingsClient` in nimbus and outside of nimbus (so Gecko and GeckoView) can use it.
3. Glean is needed to record events deep within a Rust implementation.
4. Many forms of async infrastructure (e.g. Promises, Coroutines, Observables, callbacks) that would make writing APIs more expressive and flexible. Callbacks and callback interfaces are routes to these.

WebIDL provides two types:

[Callback interfaces][1]:

```
callback interface Filesystem {
    void write_text(string filepath, string text);
    string read_text(string filepath);
};

namespace my_lib {
    set_filesystem(Filesystem fs);
}
```

[Callback functions][2]:
```
callback TaskFinished = void task_finished(string status);

namespace my_lib {
    do_task(string payload, TaskFinished callback);
}
```

[1]: https://heycam.github.io/webidl/#dfn-callback-interface
[2]: https://heycam.github.io/webidl/#idl-callback-functions

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1883)

jhugman:
Following discussion with @eoger @dmose @jdragojevic : we've scoped this work out to be three separate parts: for kotlin, swift and JS. 

Since this does not have to be on the critical path, we should close this as done, and open an issue for each of the target languages.

Issue #197, opened 2020-08-03
Support returning objects from functions and methods


jhugman:
As part of #37 and complementing #40. Putting this here so it doesn't get lost.

We should be able to return objects from functions and methods: 

e.g. 

```kotlin
val pool = new ThingPool()
val thing = pool.getOrCreate(id = "abc")

// later on 
val sameThing = pool.getOrCreate(id = "abc")

// The objects are equal.
assert(thing == sameThing)

// The objects are the same.
// it might be ok if this isn't true, but we need to discover the assumptions and document them.
assert(thing === sameThing)
```

The `===` case is the hard bit! We could maintain a `handle_map` on the platform side, but then we have some book keeping to do around destruction, including if the object is destructed on the Rust side. If the objects **must** be the same, then we should do it in a separate issue.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1596)

rfk:
Ref https://github.com/mozilla/uniffi-rs/pull/366#discussion_r557002595 for a concrete use-case for this.

rfk:
Since this was on my radar, a bit of a brain-dump of another possible source of complexity here...

When returning an object from a Rust function, how do we know whether this was an object that was already in the handlemap, or whether it needs to be added to the handlemap? If it's an existing instance, how do we discover its existing handle in order to return that value across the FFI? I feel like we may have to expose some of the internal details of how the handlemap works so that the hand-written Rust code for the component can cooperate with the generated Rust scaffolding in order to make this work.

rfk:
This is now possible as of v0.12.0.

PR #344, opened 2020-11-06
Add callback interfaces for Kotlin

### LINKED ISSUE #352: Implement callback interfaces for Kotlin
This was #342 ; but now covers just the work in Kotlin.

[Callback interfaces][1]:

```
callback interface Filesystem {
    void write_text(string filepath, string text);
    string read_text(string filepath);
};

namespace my_lib {
    set_filesystem(Filesystem fs);
}
```

[1]: https://heycam.github.io/webidl/#dfn-callback-interface

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1922)

jhugman:
This PR implements the WebIDL `callback interface`, to setup Rust calling into Kotlin.

A callback interface describes a trait in Rust. Uniffi generates a corresponding interface in Kotlin for the client code to implement. The client code sends the implementation to rust, where Rust treats it as a `Box<dyn Trait>`. Any call to that trait is then sent back to Kotlin.

Related to #342 . (fixes for Kotlin)
Fixes #352 .

## Example Problem

We'd like to have a Rust trait to fetch a document which will drive experiments:

```rust
trait FetchClient {
    fn fetch(filepath: String) -> String
}

struct NimbusClient {
    client: Box<dyn FetchClient>,
}

impl NimbusClient {
    fn set_fetch_client(client: Box<dyn FetchClient>) {
        self.client = client;
    }

   fn update_experiments() {
        let json_string =  self.client.fetch("experiments.json");
        …
    }
}
```

Declared in the IDL:

```idl
callback interface FetchClient {
    string fetch(string filepath);
};

interface NimbusClient {
    set_fetch_client(FetchClient client);
    update_experiments();
};
```

And used in Kotlin:

```kotlin
class ConceptFetchClient(
    val baseURL: String
): FetchClient {
    override fun fetch(filepath: String): String {
       let url = URL(baseURL, filepath)
       …
    }
}

val nimbus = NimbusClient()
nimbus.setFetchClient(ConceptFetchClient(BASE_URL))
nimbus.updateExperiments() // uses ConceptFetchClient
```

## Implementation

This PR presents a strawfox implementation.

We sends JNA Callbacks to send a callback interface to Rust (a.k.a. `ForeignCallback`).

This implementation of the `ForeignCallback` is individual to the callback interface— on the kotlin side; in this case this is called: `CallbackInterfaceFetchClientFFI`.

As the `ConceptFetchClient` is sent from kotlin, it is placed in a kotlin `ConcurrentHandleMap`.

A Rust `impl` of corresponding `FetchClient` trait is generated which proxies all calls through the `ForeignCallback` to kotlin, i.e. `CallbackInterfaceFetchClientProxy` in Rust calls to `CallbackInterfaceFetchClientFFI` in kotlin.

The callback object i.e. the `ConceptFetchClient` is then found, and then methods are called from `CallbackInterfaceFetchClientFFI` to the client code `ConceptFetchClient`.

## Issues

 * No support for errors, or `Result`.
 * I'm a bit worried about threading with a second concurrent handlemap and the callback being able to call back into Rust.
 * I'm a bit worried about deadlocks in the same way as #197 (we're passing callbacks across the FFI boundary)
 * Testing of the Kotlin handlemap is non-existent; I have been unable to test if Drop was working properly with a `WeakReference`, but that should be more easily tested with handlemap testing.
 * I've cargo culted quite a lot of the viaduct code for setting callbacks, though I don't know if this is still the best way of doing things.
 * Other rusty things to do with `ffi_support` and saftey, I'm not super confident on.

I'm not sure about implementing `Send`, but it seemed the only way I could get a callback interface storable in an object.


rfk:
Sorry I haven't had time to look at this in any detail, and I probably won't this week either :-(
Ping me if I don't seem to have gotten around to it by next week?

jhugman (commented):
First pass with comments.

uniffi_bindgen/src/scaffolding.rs
```diff
@@ -61,27 +62,70 @@ mod filters {
...
+    /// Get a Rust expression for writing a value into a byte buffer.
+    pub fn write_rs(
+        nm: &dyn fmt::Display,
+        target: &dyn fmt::Display,
+        type_: &Type,
+    ) -> Result<String, askama::Error> {
+        Ok(match type_ {
+            Type::CallbackInterface(type_name) => format!(
+                "Box::new(<{}Proxy as uniffi::ViaFfi>::write(&{}, {})",
+                type_name, nm, target,
```
rfk, line 106: I don't think it should be possible to write `CallbackInterfaces` into a buffer, and it may be better to bail out here rather than generate code for it at all. (Also I think the closing paren on `Box::new(` may be missing here?)

uniffi_bindgen/src/templates/CallbackInterfaceTemplate.rs
```diff
@@ -0,0 +1,102 @@
+{#
+// For each callback interface in the UDL, we assume the caller has provided a corresponding
+// Rust `trait` with the declared methods. We provide the traits for sending it across the FFI.
+// If the caller's trait does not match the shape and types declared in the UDL then the rust
+// compiler will complain with a type error.
+// 
+// The generated proxy will implement `Drop`, `Send` and `Debug`.
```
rfk, line 7: The comment here doesn't explain what the "generated proxy" is, it may be worth a sentence or two to do so (otherwise I don't know what the significance of it implementing these traits is)

uniffi_bindgen/src/templates/CallbackInterfaceTemplate.rs
```diff
@@ -0,0 +1,102 @@
...
+#[derive(Debug)]
+struct {{ trait_impl }} {
+  handle: u64
+}
+
+// We need Send so we can stash the callbacks in objects we're sharing with foreign languages
+// i.e. in the handle map.
+// Prepared for our target trait to declare:
+//  `trait {{ trait_name }}: Send + std::fmt::Debug`
+unsafe impl Send for {{ trait_impl }} {}
```
rfk, line 31: We should document here the reasons why it's safe to `impl Send` for this struct. Actually I'm surprised that it's not automatically `Send`, but I haven't finished following all the threads here yet.

uniffi_bindgen/src/templates/CallbackInterfaceTemplate.rs
```diff
@@ -0,0 +1,102 @@
...
+    {% else -%}
+    {%- endmatch -%} { 
+
+    {#- Method body #}
+        uniffi::deps::log::debug!("{{ obj.name() }}.{{ meth.name() }}");
+        let callback = uniffi::get_foreign_callback(&{{ foreign_callback_holder }}).unwrap();
+
+    {#- Packing args into a RustBuffer #}
+        {% if meth.arguments().len() == 0 -%}
+        let args_rbuf = uniffi::RustBuffer::new();
```
rfk, line 50: nit: I think it would be cleaner not to special-case a zero-length arguments list here, and just let the `for arg in meth.arguments()` loop execute zero times. The end result will be exactly the same (`RustBuffer::new()` calls `RustBuffer::from_vec` with an empty vec under the hood).

uniffi_bindgen/src/templates/CallbackInterfaceTemplate.rs
```diff
@@ -0,0 +1,102 @@
...
+        {%- endmatch %}
+    }
+    {%- endfor %}
+}
+
+unsafe impl uniffi::ViaFfi for {{ trait_impl }} {
+    type FfiType = u64;
+
+    fn lower(self) -> Self::FfiType {
+        self.handle
```
rfk, line 80: Is it supported to return these back to the foreign-language code? If we don't have a good use-case for it I'd be tempted to make this an error, even if the implementation is obvious.

uniffi_bindgen/src/templates/CallbackInterfaceTemplate.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    fn write<B: uniffi::deps::bytes::BufMut>(&self, buf: &mut B) {
+        buf.put_u64(self.handle);
+    }
+
+    fn try_read<B: uniffi::deps::bytes::Buf>(buf: &mut B) -> uniffi::deps::anyhow::Result<Self> {
+        uniffi::check_remaining(buf, 8)?;
+        <Self as uniffi::ViaFfi>::try_lift(buf.get_u64())
+    }
```
rfk, line 94: As above, do we expect to ever want to read/write these in a buffer in practice? I don't have a specific failure mode in mind, it just makes me nervous...

uniffi_bindgen/src/templates/CallbackInterfaceTemplate.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl Drop for {{ trait_impl }} {
+    fn drop(&mut self) {
+        let callback = uniffi::get_foreign_callback(&{{ foreign_callback_holder }}).unwrap();
+        unsafe { callback(self.handle, 0, Default::default()) };
```
rfk, line 100: using a named const for the `0` here might make it easier to understand what this does at a glance; IIUC it would be something like `const IDX_CALLBACK_FREE = 0` or similar.

rfk: The fact that we implement `Drop` here makes me very suspicious of writing these into a buffer etc, because we could serialize it to a buffer, drop it, then deserialize into a fresh instance with a handle that's been dropped.

uniffi_bindgen/src/interface/types.rs
```diff
@@ -64,6 +64,9 @@ pub enum FFIType {
     /// The string is owned by rust and allocated on the rust heap, and must be freed by
     /// passing it to the appropriate `string_free` FFI function.
     RustError,
+    /// A pointer to a single function in to the foreign language.
+    /// Typically, this isn't the callback that's exposed to foreign languages
```
rfk, line 68: nit: this comment feels incomplete - if it's not typically that, then what typically *is* it?

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1039,6 +1070,84 @@ impl APIConverter<Error> for weedle::EnumDefinition<'_> {
...
+        //    avoids a weird circular depenendency in the calculation.
+        self.name.hash(state);
+        self.methods.hash(state);
+    }
+}
+
+impl APIConverter<CallbackInterface> for weedle::CallbackInterfaceDefinition<'_> {
+    fn convert(&self, ci: &mut ComponentInterface) -> Result<CallbackInterface> {
+        if self.attributes.is_some() {
+            bail!("interface attributes are not supported yet");
```
rfk, line 1128: nit: please say "callback interface" rather than just "interface" in these error strings.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -152,6 +153,17 @@ impl<'ci> ComponentInterface {
         self.objects.iter().find(|o| o.name == name)
     }
 
+    /// List the definitions for every Callback Interface type in the interface.
+    pub fn iter_callback_interface_definitions(&self) -> Vec<CallbackInterface> {
+        self.callback_interfaces.to_vec()
+    }
+
+    /// Get an Callback interface definition by name, or None if no such interface is defined.
```
rfk, line 161: nit: "an" -> "a"

uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceTemplate.kt
```diff
@@ -0,0 +1,75 @@
...
+    @Suppress("TooGenericExceptionCaught")
+    override fun invoke(handle: Long, method: Int, args: RustBuffer.ByValue): RustBuffer.ByValue {
+        return {{ callback_internals }}.handleMap.callWithResult(handle) { cb -> 
+            when (method) {
+                0 -> {{ callback_internals }}.drop(handle)
+                {% for meth in obj.methods() -%}
+                {% let method_name = format!("invoke_{}", meth.name())|fn_name_kt -%}
+                {{ loop.index }} -> this.{{ method_name }}(cb, args)
+                {% endfor %}
+                else -> RustBuffer.ByValue()
```
rfk, line 26: In some future iteration, I guess this is where we'd signal an error back to the Rust code. Any thoughts on how that might work? (Or should we just forbid it for now?)

uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceTemplate.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    {% endfor %}
+}
+
+internal object {{ callback_internals }}: CallbackInternals<{{ type_name }}>(
+    {{ callback_interface_impl }}()
+) {
+    override fun register(lib: _UniFFILib) {
+        rustCall(InternalError.ByReference()) { err ->
+            lib.{{ obj.ffi_init_callback().name() }}(foreignCallback, err)
```
rfk, line 72: Oh, heh, it took me a bit of confusion to understand that `foreignCallback` here is the name of the thing you pass in to the constructor, but to know that you have to look at the implementation of `CallbackInternals`.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -28,3 +28,73 @@ inline fun <T : FFIObject, R> T.use(block: (T) -> R) =
...
+                        leftMap[handle] = obj
+                        rightMap[obj] = handle
+                    }
+            }
+
+    fun <R> callWithResult(handle: Handle, fn: (T) -> R): R =
+        lock.withLock {
+            val obj = leftMap[handle] ?: throw RuntimeException("Panic: handle not in handlemap")
+            fn.invoke(obj)
+        }
```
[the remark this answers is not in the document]
rfk: Right, good call. And things should be safe here once we've got a reference to the object out of the handlemap, because  the GC will keep it alive during the call.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -28,3 +28,73 @@ inline fun <T : FFIObject, R> T.use(block: (T) -> R) =
...
+
+    abstract fun register(lib: _UniFFILib)
+
+    fun drop(handle: Long): RustBuffer.ByValue {
+        return handleMap.remove(handle).let { RustBuffer.ByValue() }
+    }
+
+    fun lift(n: Long) = handleMap.get(n)
+
+    fun read(buf: ByteBuffer) = lift(buf.getLong())
```
rfk, line 91: As with the Rust side of things, I'm kind of dubious about `lift`/`read`/`write` here, since I don't think we can guarantee that the handle is still good.

uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceTemplate.kt
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    {% for meth in obj.methods() -%}
+    {% let method_name = format!("invoke_{}", meth.name())|fn_name_kt %}
+    private fun {{ method_name }}(kotlinCallbackInterface: {{ type_name }}, args: RustBuffer.ByValue): RustBuffer.ByValue =
+        try {
+        {#- Unpacking args from the RustBuffer #}
+            {%- if meth.arguments().len() != 0 -%}
+            {#- Calling the concrete callback object #}
+            val buf = args.asByteBuffer() ?: throw InternalError("No ByteBuffer in RustBuffer; this is a Uniffi bug")
```
rfk, line 38: What will actually happen to this exception if it does get thrown? Or equivalently, what will happen if the user's provided implementation method throws?

The JNA docs say:

> A callback should generally never throw an exception, since it doesn't necessarily have an encompassing
> Java environment to catch it. Any exceptions thrown will be passed to the default callback exception handler.

But it's not at all obvious to me what that actually means, or how the rust code will respond if allowed to continue running after this happens.

Should we try to catch exceptions and signal the error to the Rust code, maybe using an `out ExternError` argument like we do for calls from the foreign language into Rust?

rfk (commented):
This is impressive stuff! I left a bunch of comments as I worked my way through it, but I don't think I came away with any substantial concerns. 

@jhugman are you looking to merge this and iterate, or are you after an initial round of feedback and plan to keep working on some of the outstanding things mentioned in the PR description?

The one thing I do think we need is some narrative docs about the high-level design of the system (like the kind of docs we *wish* we had about the serialization format). What do you think about starting a "design docs" subdirectory under `./docs` to contain such a thing?

examples/callbacks/src/callbacks.udl
```diff
@@ -0,0 +1,37 @@
...
+  string? get_option(string? v, boolean arg2);
+  sequence<i32> get_list(sequence<i32> v, boolean arg2);
+};
+
+interface RoundTripperToRust {
+  boolean get_bool(RoundTripper callback, boolean v, boolean arg2);
+  string get_string(RoundTripper callback, string v, boolean arg2);
+  string? get_option(RoundTripper callback, string? v, boolean arg2);
+  sequence<i32> get_list(RoundTripper callback, sequence<i32> v, boolean arg2);
+};
```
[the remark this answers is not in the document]
jhugman: I started in `rondpoint` but quickly realised:

0. We couldn't land kotlin-only code and keep the `test_rondpoint.swift` and `test_rondpoint.py` going.
1. I wanted to write identifiers in English again.

uniffi/src/ffi/foreigncallbacks.rs
```diff
@@ -0,0 +1,35 @@
...
+    }
+    old_ptr == 0
+}
+
+/// Get the function pointer to the ForeignCallback. Panics if the callback
+/// has not yet been initialized.
+pub fn get_foreign_callback(callback_ptr: &AtomicUsize) -> Option<ForeignCallback> {
+    let ptr_value = callback_ptr.load(Ordering::SeqCst);
+    unsafe { std::mem::transmute::<usize, Option<ForeignCallback>>(ptr_value) }
+}
```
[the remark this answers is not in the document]
jhugman: It certainly tidied things up on the kotlin side with a generic `CallbackInternals` class. This might be the basis of a rust `ForeignCallbackInternals`.

uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceTemplate.kt
jhugman: I should add a `this.foreignCallback`.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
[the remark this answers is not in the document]
jhugman: The `ConcurrentHandleMap` in Rust sounds like it has this same implementation and similar deadlock weakness.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
jhugman: > I don't think we can guarantee that the handle is still good.

I'm not sure I understand why it's not true. I'll have to think about this.

jhugman:
> are you looking to merge this and iterate, or are you after an initial round of feedback and plan to keep working on some of the outstanding things mentioned in the PR description?

AFAICT, the biggest unresolved issue left is trapping, logging and returning errors, which I'm not sure of the best path forward here.

I think a round of feedback would be useful and then spin out the missing pieces in to separate issues. i.e. land and iterate.

However, this isn't yet on the critical path, so it's going to be a while before it's ready to land.

travis79 (commented):
This is awesome! I defer to @rfk's expertise on this, but I'm excited to try it out.

uniffi_bindgen/src/interface/mod.rs
jhugman: Yes. This was a bit of copy pasta.

uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceTemplate.kt
rfk: Broadly agreed, the more symmetry we can have between `interface` and `callback interface` the better things are likely to be.

examples/callbacks/src/callbacks.udl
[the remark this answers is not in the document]
jhugman: I changed the names to not be `RoundTripper` and to try and reflect what language the called method is written in.

jhugman (commented):
@rfk Thanks for feedback! I think this is ready for review now.

rfk:
Great! I'll make sure this is in my review queue but probably won't get a chance to look at it until end of week or early next.

uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceTemplate.kt
rfk: Could you please add a brief `TODO` comment in the code here with a link to the follow-up issue? I think it's important that we remind our future selves that the `else` branch here is a temporary situation.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -28,3 +28,80 @@ inline fun <T : FFIObject, R> T.use(block: (T) -> R) =
             // swallow
         }
     }
+
+internal typealias Handle = Long
+internal class ConcurrentHandleMap<T>(
+    private val leftMap: MutableMap<Handle, T> = mutableMapOf(),
+    private val rightMap: MutableMap<T, Handle> = mutableMapOf()
+) {
+    private val lock = java.util.concurrent.locks.ReentrantLock()
```
rfk, line 37: IIUC, this lock only protects concurrent access to the maps, and we release it before invoking the callback. So, it's possible to have multiple callbacks being dispatched at the same time. Accurate?

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
rfk: Revisiting this, I'm not sure I recall exactly what my concern was. I guess the other implicit link in the chain of argument here is that we don't ever `read` or `write` as part of some long-term serialization that might keep a reference to the handle around after it's been freed, we only do them as part of a transient shuffling of data from one side of the FFI to the other. So...I think I'm OK with this having concrete `read` and `write` implementations 👍

uniffi_bindgen/src/bindings/kotlin/templates/wrapper.kt
```diff
@@ -57,4 +60,9 @@ import java.util.concurrent.atomic.AtomicLong
 {% include "ObjectTemplate.kt" %}
 {% endfor %}
 
+// Callback Interfaces
+{% for obj in ci.iter_callback_interface_definitions() %}
```
rfk, line 64: nano-nit: consider using a different name for the loop index variable here, e.g. `cb` or `callback` or something.

uniffi_bindgen/src/templates/CallbackInterfaceTemplate.rs
rfk: I tried commenting out this `unsafe impl Send` and `cargo test` still ran successfully to completion, which suggests to me that Rust is now automatically deriving a `Send` impl for us. Maybe a refactor has made it automatically `Send`?

What do you think about switching from `unsafe impl Send` to statically asserting that it is implemented? This would help remind us to re-check any assumptions here if we ever change the struct in a way that makes it no longer automatically `Send`.

(I'd also support taking a depending on the [`static_assertions`](https://docs.rs/static_assertions) crate for this, as it may be a bit awkward to express this as an expression suitable for use with `ffi_support::static_assert!`).

rfk (approved):
Sorry for the review delay here, the end of year kind of got away from me a bit.

This looks great! I left a bunch of small comments and nits and questions but nothing that should block landing this PR and iterating further in followups. I look forward to finding some time to work on a callback implementation for Python based on this :-)

One thing to consider is a brief "high-level overview of how callbacks work" document for our future reference. The module-level docstring of `foreigncallbacks.rs` might be a good place for this, absent a dedicated "design docs" area in the repo or manual.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
jhugman: Yes, correct.

PR #345, opened 2020-11-09
Compatibility tweaks for Rust v1.43.0

rfk:
The version of Rust in mozilla-central is currently pinned to v1.43.0,
so we need to make sure we're compatible with it as well.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,11 @@
+This directory contains a Dockerfile for building the
+`uniffi-ci` docker image that we use for running tests
+in CI. To build a new version of this docker image, use:
+
+```
+docker build -t rfkelly/uniffi-ci -f Dockerfile-build .
+docker push rfkelly/uniffi-ci
+```
+
+That only works if you're `rfkelly`; we need to figure out
+a better strategy for maintainership of said docker image.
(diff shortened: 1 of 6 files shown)


docker/README.md
```diff
@@ -0,0 +1,11 @@
...
+`uniffi-ci` docker image that we use for running tests
+in CI. Build a new version of this docker image, use:
+
+```
+docker build -t rfkelly/uniffi-ci -f Dockerfile-build .
+docker push rfkelly/uniffi-ci
+```
+
+That only works if you're `rfkelly`; We need to figure out
+a better strategy for maintainership of said docker image.
```
rfk, line 11: FYI I've already pushed a new version of this Docker image using the changed below (and muddling through that process inspired me to write this short doc).

uniffi/Cargo.toml
```diff
@@ -18,7 +18,7 @@ ffi-support = "~0.4.2"
 lazy_static = "1.4"
 log = "0.4"
 # Regular dependencies
-cargo_metadata = "0.12"
+cargo_metadata = "0.11"
```
rfk, line 21: The latest `cargo_metadata` depends on a version of the `semver` crate that uses features not available in older Rust.

mhammond (approved):

PR #350, opened 2020-11-18
Start generating async gecko_js bindings rather than sync one - WIP

dmose:
Here's a PR that makes the gecko_js bindings more or less entirely async.  Excitingly, it makes the 'hot garbage' that Lena's comments refer to even hotter _and_ garbagier! A bunch of cleanup will be needed here, post-demo...

I've checked in the WebIDL and C++ generated by this branch into the `io-work` branch.

Each time I change one of the uniffi templates, I've been using this shell command from the uniffi_bindgen directory of my own uniffi_rs repo on this branch:

```
cargo run -- generate ../../mozilla-central/third_party/rust/nimbus-sdk/src/nimbus.idl -l gecko_js \
 -o ../../mozilla-central/toolkit/components/nimbus && (cd ../mozilla-central && ./mach build)
```

to generate new C++ code into the right place in my mozilla-central tree and compile it.

### DIFF
--- a/uniffi_bindgen/src/bindings/gecko_js/templates/InterfaceTemplate.cpp
+++ b/uniffi_bindgen/src/bindings/gecko_js/templates/InterfaceTemplate.cpp
@@ -46,11 +46,6 @@ JSObject* {{ obj.name()|class_name_cpp(context) }}::WrapObject(
   return dom::{{ obj.name()|class_name_cpp(context) }}_Binding::Wrap(aCx, this, aGivenProto);
 }
 
-using nimbus_1725_MozPromise = MozPromise < {{context.ffi_rustbuffer_type()}}, {{context.ffi_rusterror_type()}}, false >;
-
-using nimbus_1725_MozPromiseVoid = MozPromise < void, {{context.ffi_rusterror_type()}}, false >;
-
-
 RefPtr<nsISerialEventTarget> {{ obj.name()|class_name_cpp(context) }}::GetBackgroundTarget() {
   if (!mBackgroundET) {
     MOZ_ALWAYS_SUCCEEDS(NS_CreateBackgroundTaskQueue(
--- a/uniffi_bindgen/src/bindings/gecko_js/templates/InterfaceTemplate.cpp
+++ b/uniffi_bindgen/src/bindings/gecko_js/templates/InterfaceTemplate.cpp
@@ -48,6 +48,9 @@ JSObject* {{ obj.name()|class_name_cpp(context) }}::WrapObject(
 
 using nimbus_1725_MozPromise = MozPromise < {{context.ffi_rustbuffer_type()}}, {{context.ffi_rusterror_type()}}, false >;
 
+using nimbus_1725_MozPromiseVoid = MozPromise < void, {{context.ffi_rusterror_type()}}, false >;
+
+
 RefPtr<nsISerialEventTarget> {{ obj.name()|class_name_cpp(context) }}::GetBackgroundTarget() {
   if (!mBackgroundET) {
     MOZ_ALWAYS_SUCCEEDS(NS_CreateBackgroundTaskQueue(
--- a/uniffi_bindgen/src/bindings/gecko_js/templates/InterfaceTemplate.cpp
+++ b/uniffi_bindgen/src/bindings/gecko_js/templates/InterfaceTemplate.cpp
@@ -5,18 +5,25 @@
 
 #include "mozilla/dom/{{ obj.name()|header_name_cpp(context) }}.h"
 #include "mozilla/dom/{{ context.namespace()|header_name_cpp(context) }}Shared.h"
+#include "mozilla/dom/Promise.h"
+#include "mozilla/MozPromise.h"
+#include "nsISerialEventTarget.h"
+#include "MainThreadUtils.h"
+#include "nsAppRunner.h"
 
 namespace mozilla {
-namespace dom {
+  namespace dom {
 
-// Cycle collection boilerplate for our interface implementation. `mGlobal` is
-// the only member that needs to be cycle-collected; if we ever add any JS
-// object members or other interfaces to the class, those should be collected,
-// too.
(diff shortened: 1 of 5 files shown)


uniffi_bindgen/src/bindings/gecko_js/templates/InterfaceTemplate.cpp
```diff
@@ -39,6 +46,16 @@ JSObject* {{ obj.name()|class_name_cpp(context) }}::WrapObject(
   return dom::{{ obj.name()|class_name_cpp(context) }}_Binding::Wrap(aCx, this, aGivenProto);
 }
 
+using nimbus_1725_MozPromise = MozPromise < {{context.ffi_rustbuffer_type()}}, {{context.ffi_rusterror_type()}}, false >;
```
rfk, line 49: There is an `ffi_namespace()` method on the `ComponentInterface` struct that will produce this `nimbus_1725` prefix for you.

dmose:
OK, this branch now generates bindings that actually compile.  I've checked in updated bindings on io-work branch, so it compiles again.

uniffi_bindgen/src/bindings/gecko_js/templates/InterfaceTemplate.cpp
dmose: Good to know; thanks!

uniffi_bindgen/src/bindings/gecko_js/templates/macros.cpp
```diff
@@ -1,21 +1,45 @@
...
+       {{ arg.name() }}={% match arg.type_() %}{% when Type::String %}PromiseFlatString({{ arg.name() }}){% else %}{{ arg.name() }}{% endmatch %}
+
+      {%- if !loop.last %},{% endif -%}
+      {%- endfor %}]() {
+        if (XRE_IsParentProcess() && NS_IsMainThread()) {
+          MOZ_CRASH("lambda called outside of parent process main thread");
+        }
+
+        {{ context.ffi_rusterror_type() }} err = {0, nullptr};
+        const {% match func.ffi_func().return_type() %}{% when Some with(type_) %}{{ type_|type_ffi(context) }} loweredRetVal_ = {% else %}int8_t loweredRetVal_ = 0; // MozPromise doesn't support void, so we use a dummy boolean (sigh){% endmatch %}
```
dmose, line 36: Very helpful comment from @brennie:

> MozPromise doesnt allow void, but there is a special Ok type for when you want void: https://searchfox.org/mozilla-central/source/mfbt/Result.h#22-27

uniffi_bindgen/src/bindings/gecko_js/templates/InterfaceTemplate.cpp
```diff
@@ -39,6 +46,14 @@ JSObject* {{ obj.name()|class_name_cpp(context) }}::WrapObject(
   return dom::{{ obj.name()|class_name_cpp(context) }}_Binding::Wrap(aCx, this, aGivenProto);
 }
 
+RefPtr<nsISerialEventTarget> {{ obj.name()|class_name_cpp(context) }}::GetBackgroundTarget() {
+  if (!mBackgroundET) {
+    MOZ_ALWAYS_SUCCEEDS(NS_CreateBackgroundTaskQueue(
+      "SystemInfoThread", getter_AddRefs(mBackgroundET)));
```
mhammond, line 52: nit: the name of this queue is copy-pasta and should be updated

uniffi_bindgen/src/bindings/gecko_js/templates/macros.cpp
```diff
@@ -35,25 +59,37 @@
...
+      {%- when ReturnBy::Void %}
+      promise->MaybeResolveWithUndefined();
+      {%- endmatch %}
+    },
+    [promise]({{ context.ffi_rusterror_type() }} err) {
+      // XXX put the message into the error
+      // (a la aRv.ThrowOperationError(nsDependentCString(err.mMessage))
+      promise->MaybeReject(NS_ERROR_FAILURE);
+    });
+  return promise.forget(); // XXX i assume .forget is needed, check on this
```
mhammond, line 94: Pretty sure this is correct - although I'm a very very long way from any kind of expert here, IIUC, `Promise::Create()` returns an `already_AddRefed<>` and you don't need any other references, so .`forget()` is the right thing.

uniffi_bindgen/src/bindings/gecko_js/templates/macros.cpp
```diff
@@ -35,25 +59,37 @@
...
+      MOZ_ASSERT(ok_);
+      promise->MaybeResolve(retVal_);
+      {%- when ReturnBy::Void %}
+      promise->MaybeResolveWithUndefined();
+      {%- endmatch %}
+    },
+    [promise]({{ context.ffi_rusterror_type() }} err) {
+      // XXX put the message into the error
+      // (a la aRv.ThrowOperationError(nsDependentCString(err.mMessage))
+      promise->MaybeReject(NS_ERROR_FAILURE);
```
mhammond, line 92: One difference I notice here is that before your patch, functions that weren't declared as throwing an error would `MOZ_ASSERT` on failure, where this is doing a "normal" rejection. The MOZ_ASSERT kinda makes some sense to me (even if you then reject) or at least make alot of noise (eg, IIUC, we'd still fail on, say, a poisoned mutex, which we'd probably want to know about ASAP)

uniffi_bindgen/src/bindings/gecko_js/templates/InterfaceTemplate.cpp
dmose: good catch; thanks

uniffi_bindgen/src/bindings/gecko_js/templates/macros.cpp
dmose: Yeah, that was me cutting corners to make the demo go; thanks for the reminder to clean this up.  :-)

rfk:
Clearing out some issues that we're not actively working on; I'm going to consolidate context on the Gecko bindings into a single new issue at #991

PR #357, opened 2020-12-08
Don't use i32 helpers in generated Enum code.

### LINKED ISSUE #356: Don't use Int type helpers in Enum definition
The nimbus Kotlin bindings currently fail to compile with:

```
[task 2020-12-08T06:07:27.105Z] > Task :nimbus:compileDebugKotlin FAILED
[task 2020-12-08T06:07:27.105Z] e: /builds/worker/checkouts/src/components/nimbus/build/generated/source/uniffi/debug/java/org/mozilla/experiments/nimbus/nimbus.kt: (606, 55): Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
[task 2020-12-08T06:07:27.105Z] internal fun kotlin.Boolean.Companion.read(buf: ByteBuffer): Boolean defined in org.mozilla.experiments.nimbus in file nimbus.kt
[task 2020-12-08T06:07:27.105Z] internal fun kotlin.Byte.Companion.read(buf: ByteBuffer): Byte defined in org.mozilla.experiments.nimbus in file nimbus.kt
[task 2020-12-08T06:07:27.105Z] internal fun kotlin.String.Companion.read(buf: ByteBuffer): String defined in org.mozilla.experiments.nimbus in file nimbus.kt
```

The line causing the error is the `read` definition of an `enum class`, which is attempting to use `Int.read` as a helper function. I believe this is failing because the nimbus IDL doesn't use `i32` anywhere, so we don't generate any of the helpers for `Int`.

The generated code should use `buf.getInt()` directly, rather than the `Int.read` helper. PR incoming.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1926)

rfk:
Previously, the generated Kotlin code for enums would try to use
helpers from the `Int` companion class, but we would only emit the
code for those helpers if the interface was also using the integer
type.

Now, the generated Kotlin code for enums directly uses lower-level
methods for reading and writing integers, so that it doesn't depend
on other generated code.

Fixes #356.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,5 +8,6 @@ members = [
   "examples/geometry",
   "examples/rondpoint",
   "examples/sprites",
-  "examples/todolist"
+  "examples/todolist",
+  "fixtures/regressions/enum-without-i32-helpers",
 ]
--- a/fixtures/regressions/README.md
+++ b/fixtures/regressions/README.md
@@ -1,0 +1,7 @@
+# Regression tests for uniffi components
+
+This directory contains standalone uniffi component crates designed
+as regression tests for specific historical bugs. They're written
+for automated execution and *not* for being easily understandable
+by users; if you're trying to get your head around uniffo then the
+"examples" directory will be a much better bet.
--- a/fixtures/regressions/enum-without-i32-helpers/Cargo.toml
+++ b/fixtures/regressions/enum-without-i32-helpers/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "i356-enum-without-int-helpers"
+edition = "2018"
+version = "0.5.0"
+authors = ["Firefox Sync Team <sync-team@mozilla.com>"]
+license = "MPL-2.0"
+publish = false
+
+[lib]
+crate-type = ["cdylib"]
+name = "uniffi_regression_test_i356"
+
+[dependencies]
+uniffi_macros = {path = "../../../uniffi_macros"}
+uniffi = {path = "../../../uniffi", features=["builtin-bindgen"]}
+
+[build-dependencies]
+uniffi_build = {path = "../../../uniffi_build", features=["builtin-bindgen"]}
--- a/fixtures/regressions/enum-without-i32-helpers/README.md
+++ b/fixtures/regressions/enum-without-i32-helpers/README.md
@@ -1,0 +1,10 @@
+# Regression test for issue #356
+
+In issue [#356](https://github.com/mozilla/uniffi-rs/issues/356)
+we discovered that the generated bindings for Enum classes were
+using helper functions from the i32 data type, but we would only
+generate those functions if the component interface was also
+using that type directly.
+
+This crate is a minimal reproduction of the issue, and its Kotlin
+bindings would fail to compile in the presence of the bug.
--- a/fixtures/regressions/enum-without-i32-helpers/build.rs
+++ b/fixtures/regressions/enum-without-i32-helpers/build.rs
@@ -1,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+fn main() {
+    uniffi_build::generate_scaffolding("./src/test.udl").unwrap();
+}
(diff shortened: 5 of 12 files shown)


rfk:
@jhugman r? This blocks the release of an updated nimbus over in https://github.com/mozilla/application-services/pull/3748.

Cargo.toml
```diff
@@ -8,5 +8,6 @@ members = [
   "examples/geometry",
   "examples/rondpoint",
   "examples/sprites",
-  "examples/todolist"
+  "examples/todolist",
+  "regression_tests/i356-enum-without-int-helpers",
```
rfk, line 12: I didn't want to land this without a test, and I didn't want to contrive a new example just to test this edge-case (especially since the bug depends on the *absence* of something in the UDL, rather than its presence). So, I'm trying out a new suite of explicit regression tests.

The setup here is a little wordy for my liking, and if we generate more of these then we may want to invest in some automation so they can be written more succinctly. But it's a start!

jhugman: I'm so pleased you have a test!

I think we could bike shed this, but happy to land as is.

FWIW, I think my pref would be:

`fixtures/regressions/enums-without-i32-helper` and then put the URL to the issue in the README.md of the fixture.

regression_tests/README.md
```diff
@@ -0,0 +1,7 @@
+# Regression tests for uniffi components
+
+This directory contains standalone uniffi component crates designed
+as regression tests for specific historical bugs. They're written
+for automated execution and *not* for being easily understandable
+by users; if you're trying to get your head around uniff then the
```
jhugman, line 6: Nit: uniffi

regression_tests/i356-enum-without-int-helpers/src/test.udl
```diff
@@ -0,0 +1,12 @@
+// We define an enum, and do not use the i32 datatype.
+// This should compile and run fine, but in https://github.com/mozilla/uniffi-rs/issues/356
+// we were accidentally assuming the presence of i32 datatype helpers in the generated code.
```
jhugman, line 3: Nice.

regression_tests/i356-enum-without-int-helpers/tests/bindings/test.swift
```diff
@@ -0,0 +1,4 @@
+import regression_test_i356
+
+assert(which(arg: true) == Which.yeah)
+assert(which(arg: false) == Which.nah)
```
jhugman, line 4: Nit: exclude `Which`. i.e. 

```swift
assert(which(arg: true) == .yeah)
```

jhugman: Also: I note that Swift is using the `UInt32.read` extension methods, so should we change that now? or file an issue?

regression_tests/i356-enum-without-int-helpers/tests/bindings/test.kts
```diff
@@ -0,0 +1,4 @@
+import uniffi.regression_test_i356.*;
+
+assert(which(true) == Which.YEAH)
+assert(which(false) == Which.NAH)
```
jhugman, line 4: So I guess this wouldn't work before, but now it does? That's what makes it a useful test?

jhugman (approved):
I have a few nits, but nothing that needs a second round of review.

regression_tests/i356-enum-without-int-helpers/tests/bindings/test.kts
rfk: Prior to the fix, we wouldn't even get to running this, it would fail to compile. Which, it's probably worth me adding a comment to that effect in the test here for clarity.

Cargo.toml
rfk: WFM, I will update to the suggested layout. Something about the shorter directory names is very pleasing :-)

regression_tests/i356-enum-without-int-helpers/tests/bindings/test.swift
rfk: Nice catch. It looks like the swift bindings generate these helpers unconditionally, rather than inspecting the types used in the UDL. Let's file an issue for followup to do conditional generation, and we can count on this new test to catch the problem if we reintroduce it there. (I confirmed that the test would catch it by commenting out the `extension UInt32: Primitive, ViaFfi` in the generated swift bindings, and this swift test does indeed fail with an appropriate compile-time error in that case).

rfk: I filed https://github.com/mozilla/uniffi-rs/issues/362 for followup.

rfk:
Updated per review feedback, I'll merge this once CI goes green.

PR #363, opened 2020-12-08
Release v0.6.0

rfk:
I made a `v0.6.0` release using `cargo release`, and even with some rough edges it was still a big improvement over doing it manually! I've added some docs as part of this commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ It's not (yet) quite an ideal fit for our workflow, but it helps! Steps:
     * This is a limitation of using `cargo release` in a workspace,
       ref [sunng87/cargo-release#222](https://github.com/sunng87/cargo-release/issues/222)
 5. Tag the release commit in github.
-    * `git tag v0.6.0`
-    * `git push origin v0.6.0`
+    * `git tag vX.Y.Z`
+    * `git push origin vX.Y.Z`
 6. Push your branch, and make a PR to request it be merged to the main branch.
     * `git push origin`
--- a/README.md
+++ b/README.md
@@ -159,19 +159,20 @@ Please check the project's [code of conduct](./CODE_OF_CONDUCT.md).
 We use [cargo-release](https://crates.io/crates/cargo-release) to simplify the release process.
 It's not (yet) quite an ideal fit for our workflow, but it helps! Steps:
 
-1. Start a new branch for the release:
-    * `git checkout -b release-vX.Y.Z`
-    * `git push -u origin release-vX.Y.Z`
-2. Run `cargo release --dry-run -vv X.Y.Z` and check that the things
+1. Start a new branch for the release. We typically use a single branch for all point releases,
+   so it should be named something like `release-v0.6.x`:
+    * `git checkout -b release-v{MAJOR}.{MINOR}.x`
+    * `git push -u origin release-v{MAJOR}.{MINOR}.x`
+2. Run `cargo release --dry-run -vv {MAJOR}.{MINOR}.{PATCH}` and check that the things
    it is proposing to do seem reasonable.
-3. Run `cargo release X.Y.Z` to bump version numbers and
+3. Run `cargo release {MAJOR}.{MINOR}.{PATCH}` to bump version numbers and
    publish the release to crates.io.
 4. Run `git commit --amend` to fix up the version number in the commit message.
-    * Manually replace `{{version}}` with `vX.Y.Z`.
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -153,3 +153,25 @@ a suitable format for distribution (e.g. a `.jar` or android package for Kotlin)
...
+   publish the release to crates.io.
+4. Run `git commit --amend` to fix up the version number in the commit message.
+    * Manually replace `{{version}}` with `vX.Y.Z`.
+    * This is a limitation of using `cargo release` in a workspace,
+      ref [sunng87/cargo-release#222](https://github.com/sunng87/cargo-release/issues/222)
+5. Tag the release commit in github.
+    * `git tag v0.6.0`
+    * `git push origin v0.6.0`
+6. Push your branch, and make a PR to request it be merged to the main branch.
+    * `git push origin`
```
dmose, line 177: I'm confused; didn't we just push the branch on line 175?

uniffi_build/Cargo.toml
```diff
@@ -12,7 +12,7 @@ keywords = ["ffi", "bindgen"]
 
 [dependencies]
 anyhow = "1"
-uniffi_bindgen = { path = "../uniffi_bindgen", optional = true, version = "0.5.0" }
+uniffi_bindgen = { path = "../uniffi_bindgen", optional = true, version = "^0.6.0"}
```
dmose, line 15: nit: In my idealized imagination, we'd just have "0.6.0", without the caret, since I'm not convinced that we want to complicate our (theoretical) testing matrix in this way.  If that's easy to do with `cargo release`, that'd be nice.  That said, if it's not, (maybe it's part of the fix to sunng87/cargo-release?), it's probably not worth worrying about.

README.md
rfk: Nope; we pushed the tag, not the branch.

uniffi_build/Cargo.toml
rfk: Yeah, this was the default behavior of `cargo release`, I couldn't see a way to change it, and it didn't seem like enough of a problem relative to the value of not having to make all these changes by hand.

rfk:
I made a patch release with a strict `=` requirement, does this match your idealized imagination @dmose?

(Per [the docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio), our previous version was actually an implicit caret and `cargo release` was just making it explicit).

uniffi_build/Cargo.toml
rfk: Aha, looks like using an explicit `=` requirement rather than an implicit `^` will do the right thing, and is more in line with our expectations anyway.

dmose: Sounds good.

examples/arithmetic/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "uniffi-example-arithmetic"
 edition = "2018"
-version = "0.6.0"
+version = "0.6.1"
```
dmose, line 4: So this looks like it's going to push 0.6.1 to the 0.6.0 branch.  If that's really what we want (I suspect not), perhaps the branch should use the string 0.6.x rather than 0.6.0 in its name.

rfk: Yesh, I think you're right, `release-v0.6.x` is closer to what I imagine us using here.

dmose: So what's the right way forward here? Seems like we might be able to (theoretically) change in flight, but given the pre-existing 0.6.{2+}, I'm guessing that might not be practical this time around.  I'm happy to follow your lead here.  I'm going to mark this approved, so that if you're confident you've got something we should go forward with, it won't be blocked from landing.  That said, feel free to ask for another look from me, which I can to Friday Pacific time.

dmose (approved):
r/rs=dmose; feel free to ask for another glance if you think that'll be helpful

examples/arithmetic/Cargo.toml
rfk: Thanks. I don't think the branching setup matters *too* much here this time around, since it's mostly the tags that are important, and they only associate with a commit rather than with a specific branch. What I propose to do here is:

* Update the docs to discuss a `release-v0.6.x`-format branch approach for future reference.
* Merge this to `main` with your r+.
* Push it to a `release-v0.6.x` and see if I can delete this `release-v0.6.0` branch, for completeness.

dmose:
Sounds good!

Issue #244, opened 2020-08-14
Unnecessary sources of overhead in uniffi


thomcc:
(Uh, this was way more than I intended to write about this 😓... I think my thoughts on the locking situation probably were worth writing down, since in general it helps explain what I believe the problem space to be, what lead me to the HandleMap approach, why locking on the other langs is tricky, etc. This was probably important to get down somewhere, so I'm glad I did)

---

I've shared this crate with a few people, but in general they respond negatively when they realize how much overhead uniffi introduces. (Filing this because a friend of mine finally got back to me about uniffi today, but I've heard it 3 or so times)

There are two main causes of overhead:
1. Serialization across the boundary.
2. Using ConcurrentHandleMap, when ConcurrentHandleMap requires a locking a `std::sync::RwLock`, a `std::sync::Mutex`, and performing a number of checks on the handle for validity.

## Serialization

The 1st could be avoided for e.g. `#[repr(C)]` structs and enums with defined `repr` (C, iN, C+iN. note that this well defined for way more cases than you'd expect, thanks to the needs of cbindgen and webrender).

Passing and returning `Vec<T>` for ffi-safe `T` or `String` can be done by converting to a repr-c struct containing (pointer, length, capacity).

Hell, right now we use nul-terminated C-strings (I think), and really there's no real reason to do that over e.g. `#[repr(C)] struct Str(*const c_char, usize);`.

## Locking

Various options.

1. Supporting the use non-`std::sync` locking (probably `parking_lot`) in `ffi_support::ConcurrentHandleMap`. This is essentially trivial.

2. A `ffi_support::ConcurrentHandleMap` which is really a `RwLock<HandleMap<RwLock>>` or `RwLock<HandleMap<AtomicRefcell>>` instead of a `Mutex`. Note that `RwLock` has higher over head than Mutex, so it would *really* only be good if we found a way to opt-out of the locking that happens on the non-rust side of the FFI. OTOH, `AtomicRefcell` is really only viable because we do all the locking on the other side.

3. There exist algorithms for writing a lock-free HandleMap-alikes\*. This is not a good choice for uniffi, as we're using handles because they're more "obviously correct" than pointers, and lock-free data structures are almost never "obviously correct".

4. Finally, and what I think is a better safety vs complexity trade-off: just use pointers. (Note that you need to enforce `T: Send` for any type you do this with, and will be relying on locking happening in the caller. More on that in a sec). Continued below.

### Using Pointers

[...]

Issue #40, opened 2020-07-29
Support passing objects as function/method arguments


rfk:
It is not currently possible to pass objects as arguments to functions or methods, apart from the special-case handling of the implicit first argument in a method call.

In theory this should be easy right? You accept the `u64` handle as an argument, look up the object in the appropriate handle-map, and you're off to the races!

There are significant complexities in practice. I don't think we need this functionality with any urgency but I wanted to put a bug on file for why this is harder than it might seem. Some thoughts below, please comment with others if you have them.

If the object argument needs to be mutable, then we can't just grab the object out of the handle-map, we have to use a closure that manages locking for exclusive access. If we have several object arguments, we would need to codegen a series of nested closures, which could get messy fast.

Mutable object arguments are also a deadlock hazard. Imagine a function call that takes two object arguments, like `modify(obj1, obj2)`, and a concurrent call to `modify(obj2, obj1)`. If we naively try to access each object argument in turn inside the generated FFI function, then the first call will lock `obj1`, and second call will lock `obj2`, and then they'll both deadlock waiting for the lock on their second argument. We might be able to mitigate that by locking object arguments in a consistent order, such as always trying to lock arguments with lower numeric handles first. But that'll make the codegen even messier.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1550)

rfk:
Ref https://github.com/mozilla/uniffi-rs/pull/366#discussion_r557002595 for a concrete use-case for this.

Also of note: the proposal for reducing locking in https://github.com/mozilla/uniffi-rs/pull/368 might also help us with the deadlock hazard described above, or at the very least would help make the deadlock hazard visible in the code instead of being hidden away in the generated bindings.

rfk:
This is now possible as of v0.12.0

PR #366, opened 2020-12-21
Experiment with using raw pointers for objects.

rfk:
Opening for early visibility.

As noted in [SDK-157](https://jira.mozilla.com/browse/SDK-157), the way we use a `ConcurrentHandleMap` for managing objects means that we foist certain threading/locking behaviors on all API conumers. It's currently only possible for a single thread to be operating on a given Rust object at any one time, even if there are operations that could safely be performed without synchronization. This is making it awkward to create the desired API for the Nimbus SDK, where we want to offer a synchronous `getExperimentBranch` method that will return immediately, even if there is a concurrent `updateExperiments` call running on a another thread.

A little while back, @thomcc posted the excellent https://github.com/mozilla/uniffi-rs/issues/244 discussing sources of overhead in uniffi, and one thing he suggested there was that we could consider using raw pointers instead of a `HandleMap`. Many of the features of `HandleMap` are designed to guard against programmer error, so they're more useful in a manually-written FFI layer than in generated bindings.

This PR is me experimenting with using raw pointers instead of a `ConcurrentHandleMap`. I plan to use it to see whether the handlemap is indeed the source of the blocking behaviour seen in [SDK-157](https://jira.mozilla.com/browse/SDK-157).

This is emphatically *not* suitable for merging, and in fact it contains known very-dangerous-completely-unsafe behaviours (such as just assuming we can get an `&mut T` from a pointer interpreted as a `Box<T>` without worrying about a little thing called *locking*). I need to get my head around what would be required to make this safe, but the unsafe version is suitable for my investigations in the meantime.

### DIFF
--- a/docs/manual/src/SUMMARY.md
+++ b/docs/manual/src/SUMMARY.md
@@ -24,3 +24,9 @@ Hello everyone
 # Swift
 
 - [Integrating with XCode](./swift/xcode.md)
+
+# Internals
+
+- [Lifting, Lowering, and Serialization](./internals/lifting_and_lowering.md)
+  - [Serialization format](./internals/serialization_format.md)
+- [Managing object references](./internals/object_references.md)
--- a/docs/manual/src/udl/interfaces.md
+++ b/docs/manual/src/udl/interfaces.md
@@ -156,7 +156,7 @@ impl Counter {
 }
 ```
 
-This version uses the `AtomicU64` struct for interior mutability, which is both `Sync` and
+This version uses an `AtomicU64` for interior mutability, which is both `Sync` and
 `Send` and hence will compile successfully:
 
 ```rust
@@ -178,6 +178,3 @@ impl Counter {
     }
 }
 ```
-
-Uniffi aims to uphold Rust's safety guarantees at all times, without requiring the
-foreign-language code to know or care about them.
--- a/docs/manual/src/udl/interfaces.md
+++ b/docs/manual/src/udl/interfaces.md
@@ -72,3 +72,112 @@ func display(list: TodoListProtocol) {
     }
 }
 ```
+
+# Concurrent Access
+
+Since interfaces represent mutable data, uniffi has to take extra care
+to uphold Rust's safety guarantees around shared and mutable references.
+The foreign-language code may attempt to operate on an interface instance
+from multiple threads, and it's important that this not violate Rust's
+assumption that there is at most a single mutable reference to a struct
+at any point in time.
+
+By default, uniffi enforces this using runtime locking. Each interface instance
+has an associated lock which is transparently acquired at the beginning of each
+call to a method of that instance, and released once the method returns. This
+approach is simple and safe, but it means that all method calls on an instance
+are run in a strictly sequential fashion, limiting concurrency.
+
+You can opt out of this protection by marking the interface as threadsafe:
+
+```idl
+[Threadsafe]
+interface Counter {
+    constructor();
+    void increment();
+    u64 get();
+};
+```
+
+The uniffi-generated code will allow concurrent method calls on threadsafe interfaces
+without any locking.
+
+For this to be safe, the underlying Rust struct must adhere to certain restrictions, and
+uniffi's generated Rust scaffolding will emit compile-time errors if it does not.
+
+The Rust struct must not expose any methods that take `&mut self`. The following implementation
+of the `Counter` interface will fail to compile because it relies on mutable references:
+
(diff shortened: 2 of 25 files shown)


thomcc:
> such as just assuming we can get an &mut T from a pointer interpreted as a Box<T> without worrying about a little thing called locking

FYI just having aliased `Box<T>` is already UB, regardless if lock it or turn it into a `&mut T` or `&T`, see https://github.com/rust-lang/unsafe-code-guidelines/issues/258

docs/manual/src/SUMMARY.md
```diff
@@ -24,3 +24,9 @@ Hello everyone
 # Swift
 
 - [Integrating with XCode](./swift/xcode.md)
+
+# Internals
```
rfk, line 28: Much like our lifting/lowering/serialization story, the way we currently use handlemaps is under-documented. Let's add a new section to the manual that we can use for narrative docs about such things.

docs/manual/src/udl/interfaces.md
```diff
@@ -71,4 +71,112 @@ func display(list: TodoListProtocol) {
         print($0)
     }
 }
-```
\ No newline at end of file
+```
+
+# Concurrent Access
```
rfk, line 76: Based on this experiment and subsequent conversations, here's a concrete proposal for how to expose uniffi interfaces in a way that gives us more control over the concurrency story. Feedback most welcome at this stage! /cc @mhammond @jhugman @dmose.

docs/manual/src/udl/interfaces.md
```diff
@@ -71,4 +71,112 @@ func display(list: TodoListProtocol) {
...
+By default, uniffi enforces this using runtime locking. Each interface instance
+has an associated lock which is transparently acquired at the beginning of each
+call to a method of that instance, and released once the method returns. This
+approach is simple and safe, but it means that all method calls on an instance
+are run in a strictly sequential fashion, limiting concurrency.
+
+You can opt out of this protection by marking the interface as threadsafe:
+
+```idl
+[Threadsafe]
```
rfk, line 94: Bikeshedding also welcome on how to expose this in the UDL...

docs/manual/src/udl/interfaces.md
rfk: (Note that the PR doesn't actually implement proposal yet, although I'm confident that it can be implemented without much fuss)

mhammond: This seems awesome. While it doesn't handle the "multiple bindings" use-case, I wonder if we can avoid uniffi knowing about that kind of setup? Eg, I wonder if we do arrange so that instead of lib.rs having, say:

```
    pub fn get_active_experiments(&self) -> Result<Vec<EnrolledExperiment>> {
        get_enrollments(self.db()?)
    }
```

instead it does something like:
```
    pub fn get_active_experiments(&self) -> Result<Vec<EnrolledExperiment>> {
       handwavey_get_static_instance.get_active_experiments()
    }
```

(ie, that it's OK if there are multiple instances of the bindings, because they all just delegate). I'd need to play a little to see if that might work in practice though. Did you have anything in mind for that use-case?

rfk: (We discussed this a little in meetings and slack, but commenting here for posterity)

I'd be interested in exposing the singleton concept more directly through the IDL. For example, imagine a Nimbus API definition that looks like this:

```
namespace nimbus {
  // Returns the global NimbusClient singleton, or null if it has not been initialized yet.
  NimbusClient? get_global_nimbus_client();

  // Sets the global NimbusClient singleton, or throws if it has already been set.
  [Throws=Error]
  void set_global_nimbus_client(NimbusClient client);
};

interface NimbusClient {
  // all the usual methods here
}
```

Uniffi doesn't support this today. Specifically, we don't let you pass object references as arguments (#40) or return them from functions (#197), so you wouldn't be able to declare the getter/setter for the singleton.

However, I think it's clear that we *want* to support such functionality at some point, and I don't think there's any deep technical reason why we couldn't do so, there's just a bunch of details to carefully work through (as described in the linked issues).

rfk:
This was a useful experiment, but as discussed in https://github.com/mozilla/uniffi-rs/pull/368 I believe we're going to stick with using a handlemap for now. I've migrated to docs proposed here to a standalone PR at https://github.com/mozilla/uniffi-rs/pull/371, included a link to this experiment in the ADR doc for future reference, and will close this PR as having served its purpose.

PR #367, opened 2021-01-13
Add some narrative docs about important internal implementation details.

### LINKED ISSUE #297: Document the UniFFI serialization format
We have a [fantastic ADR](https://github.com/mozilla/uniffi-rs/blob/757a028278f8c8d81754e6a32ea3844012669deb/docs/adr/0002-serialize-complex-datatypes.md) that explains why we used our own serialization format, but no documentation for the format itself. Even though it's an internal implementation detail—if we wanted to, we could switch to a different format, or even avoid serialization altogether, and neither the Rust component nor the consumers would be the wiser—it might be good to document it, for our own use.

The general idea:

* Fixed-width integers (signed and unsigned, 8, 16, 32, and 64 bits) are serialized in big-endian order.
* Booleans are serialized as `int8_t`, either 1 or 0.
* Floats and doubles are converted to their bit patterns (`float` becomes `uint32_t`; `double`, `uint64_t`).
* Strings are length-prefixed (with the number of bytes) and then UTF-8 encoded, _without_ a trailing null character.
* Sequences are length-prefixed (number of elements in the sequence), followed by the serialization of each element.
* Maps are length-prefixed (number of pairs in the map), followed by the serialization of each key and its value, one after the other.
* Optionals are either 1 (`int8_t`) followed by the serialization of the value for `Some<T>`, or `0` for `None`.

Length prefixes are 32-bit _signed_ integers. (In general, we use signed types, because unsigned types are experimental in Kotlin).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1800)

rfk:
Specifically:

* What "lifting" and "lowering" are and how they're done.
* The serialization format we use for lowering complex data types.
* How and why we use handlemaps for managing object references.

(These describe the *current* state of the world, but I want to get them into the repo so that I can propose *diffs* to that state of the world in order to resolve the current threading/concurrency issues).

Fixes #297.

### DIFF
--- a/docs/manual/src/SUMMARY.md
+++ b/docs/manual/src/SUMMARY.md
@@ -24,3 +24,8 @@ Hello everyone
 # Swift
 
 - [Integrating with XCode](./swift/xcode.md)
+
+# Internals
+
+- [Lifting, Lowering, and Serialization](./internals/lifting_and_lowering.md)
+- [Managing object references](./internals/object_references.md)
--- a/docs/manual/src/internals/lifting_and_lowering.md
+++ b/docs/manual/src/internals/lifting_and_lowering.md
@@ -1,0 +1,94 @@
+# Lifting, Lowering and Serialization
+
+Uniffi is able to transfer rich data types back-and-forth between the Rust
+code and the foreign-language code via a process we refer to as "lowering"
+and "lifting".
+
+Recall that uniffi interoperates between different languages by defining
+a C-style FFI layer which operates in terms of primitive data types and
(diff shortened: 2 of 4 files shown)


dmose (approved):
This is great! I already knew a fair bit of this, but having all of it written down in detail including the not-always-obvious invariants and details is super-helpful. Note that I did not compare the docs to the code in order to ensure that they are all correct, as I'm guessing that's not necessary here. r=dmose

PR #368, opened 2021-01-13
Add an ADR for our "threadsafe interfaces" plan.

rfk:
### [Click for Rendered View](https://github.com/mozilla/uniffi-rs/blob/adr-threadsafe-interface/docs/adr/0003-threadsafe-interfaces.md)

Based on conversations over the last few days, here is a concrete proposal for how we can resolve the underlying issue that casued the [recent Fenix startup regression](https://github.com/mozilla/uniffi-rs/blob/adr-threadsafe-interface/docs/adr/0003-threadsafe-interfaces.md). I hope I have captured the current consensus accurately, but this is an ADR, so please do weigh in if you don't agree with something here or if you've got other considerations to include!

Something new for this ADR: I have included an [implementation sketch](https://github.com/mozilla/uniffi-rs/blob/adr-threadsafe-interface/docs/adr/0003-threadsafe-interfaces.md#implementation-sketch) section with a high-level overview of how I'd go about implementing this. I'm hoping this will empower someone who is not me to actually implement this and learn a bit more about this part of the uniffi codebase.

### DIFF
--- a/docs/adr/0003-threadsafe-interfaces.md
+++ b/docs/adr/0003-threadsafe-interfaces.md
@@ -1,0 +1,194 @@
+# Let consumers opt out of HandleMap locking if their interface is threadsafe
+
+* Status: proposed
+* Deciders: rfkelly, jhugman, mhammond, dmosedale
+* Date:  2021-01-13
+
+## Context and Problem Statement
+
+Uniffi currently uses a very coarse locking strategy for managing concurrent access to object instances,
+which has caused us to accidentally ship code in a product that [blocked the main thread on network I/O](https://jira.mozilla.com/browse/SDK-157).
+We need to enable finer-grained concurrency control in order to provide the desired API for a key consumer.
+
+Currently, every interface has a corresponding [ffi_support::ConcurrentHandleMap](https://docs.rs/ffi-support/0.4.0/ffi_support/handle_map/struct.ConcurrentHandleMap.html) that is responsible for owning all instances of
+that interface and for handing out references to them in a mutability-safe and threadsafe manner. This
+ensures that the generated code is safe in the face of concurrent operations, but has a substantial
+runtime cost: only one method call can be executed on an instance at any time. Any attempt to call an
+object method while a concurrent method is already executing, will block until the previous call has completed.
+
+The desired API for Project Nimbus includes methods that will be called synchronously from the main thread,
+and hence must not block on network or disk I/O. Such an API cannot be built with uniffi as currently
+implemented.
+
+## Decision Drivers <!-- optional -->
+
+* Enabling consumers to control the potential blocking behaviour of their generated APIs.
+* Ensure safety of the generated code.
+* Ship a solution in a timely manner to unblock Project Nimbus.
+
+## Considered Options
+
+* Option A: Do nothing, require consumers to assume that all method calls might block.
+* Option B: Let consumers mark interface definitions as threadsafe to opt in to a less-locking handlemap.
+* Option C: Insist that all interfaces be threadsafe and replace the handlemap with raw pointers.
+* Option D: Use a less-locking handlemap and rely on calling code to behave safely.
+
+## Decision Outcome
+
+Chosen option:
+ * **Option B: Let consumers mark interface definitions as threadsafe to opt in to a less-locking handlemap.**
+
+The choice here comes down to safety and simplicity. By making a more-concurrency-friendly handlemap
+we can maintain the current strict enforcement of Rust's mutability-safety and thread-safety guarantees,
+even in the face of errors in the generated bindings. It seems to be a relatively small change, and
+by making it opt-in we avoid creating busywork for other consumers who are not urgently facing this
+problem.
+
+One downside is that consumers need to opt-in to the fix, meaning that the default behavior may still
+be surprising to new consumers. We'll mitigate this with docs and will consider revisiting the default
+behaviour if the majority of consumers adopt the new approach.
+
+This choice does also punt some potential performance improvements to future work, but that seems in keeping
+with where we are in the project's lifecycle.
+
+## Pros and Cons of the Options
+
+### Option A: Do nothing, require consumers to assume that all method calls might block.
+
+Make no changes to uniffi, and instead accept the fact that method calls are executed serially.
+Document this limitation and work with consumers to update their API definitions to account for it.
+
+* Good, because we can ship this quickly.
+* Good, because we don't give up any of the safety guarantees of the current approach.
+* Bad, because it makes it basically impossible to meet the needs of one of our key early consumers,
+  and would force us into awkward compromises around a suboptimal API.
+* Bad, because it makes uniffi less attractive to potential future consumers.
+* Bad, because it keeps all the run-time overhead of the handlemap.
+* Bad, because the default behaviour still has a hidden mutex, which might be a nasty surprise
+  for future consumers in the same way that it was for Nimbus.
+
+Ultimately, we want uniffi to be a tool that helps consumers deliver value, not something that
+foist limitations upon them, which makes this option unattractive.
+
+### Option B: Let consumers mark interface definitions as threadsafe to opt in to a less-locking handlemap.
+
+Implement a variant of `ConcurrentHandleMap` that does not protect its members with a Mutex, which
+is the main source of constraints on concurrent execution in the current setup. Instead, this handlemap
+would only give out immutable references to its members, and would insist that its members are `Send`
+and `Sync` so they can be safely accessed from multiple threads.
+
+Introduce a new annotation to the UDL so that `interface` definitions can be declared as threadsafe.
+In the generated Rust scaffolding, use the new handlemap for threadsafe interfaces but keep using the
+existing `ConcurrentHandleMap` by default.
+
+* Good, because this is a relatively small change from the current behaviour, which should be fairly
+  quick to ship.
+* Good, because this is a non-breaking change for consumers who don't opt in to it.
+* Good, because the `Sync` and `Send` bounds on the underlying struct will help consumers
+  to implement their own fine-grained concurrency control while being supported by Rust's
+  compile-time guarantees.
+* Good, because consumers don't have to care about this until they discover that they need
+  fine-grained locking.
+* Good, because it leaves the door open to further improvements (like "Option C") in the
+  future if we decide that makes sense later.
+* Bad, because it keeps all the run-time overhead of the handlemap.
+* Bad, because now we need to maintain two handlemap variants.
+* Bad, because the default behaviour still has a hidden mutex, which might be a nasty surprise
+  for future consumers in the same way that it was for Nimbus.
+
+This is the solution we have ultimately selected.
+
+### Option C: Insist that all interfaces be threadsafe and replace the handlemap with raw pointers.
+
+Stop using `ConcurrentHandleMap` to intermediate object access. Instead, put each object instance
+in a `Box` and use `Box::into_raw` to transfer ownership of the box to the foreign language code
+as a raw pointer.
+
+Maintain safety by insisting that any struct implementing a UDL `interface` must be `Sync` and `Send`,
+and by refusing to hand out mutable references to the boxed instance.
+
+* Good, because it removes the issue as a potential footgun for all consumers.
+* Good, because the `Sync` and `Send` bounds on the underlying struct will help consumers
+  to implement their own fine-grained concurrency control while being supported by Rust's
+  compile-time guarantees.
+* Good, because it likely reduces runtime overhead and gives a small performance boost.
+* Bad, because it's a breaking change for all consumers, even if they don't care about this issue.
+* Bad, because it increases the amount of unsafe rust that we need to emit in the generated scaffolding.
+* Bad, because we lose some additional runtime checks provided by the handlemap, such as guarding
+  against passing a handle that belongs to a different datatype.
+* Bad, because it's a non-trivial departure from the current approach which will take more time to QA.
+
+This seems like a promising longer-term option, especially if we find that the majority of consumers
+are opting in to the fix proposed in this ADR. But the additional complexity weights heavily against
+trying to ship this as an initial fix for Project Nimbus.
+
+### Option D: Use a less-locking handlemap and rely on calling code to behave safely.
+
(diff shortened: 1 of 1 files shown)


rfk:
Optimistically assuming approval for the plan proposed here, I've also filed https://github.com/mozilla/uniffi-rs/issues/370 to track the work of implementing it.

docs/adr/0003-threadsafe-interfaces.md
```diff
@@ -0,0 +1,196 @@
...
+
+The `NonLockingHandleMap<T>` struct should be capable of handing out `&T` references but should never hand
+out a `&mut T`. It should also require that `T: Send + Sync`.
+
+### Use `NonLockingHandleMap` in the Rust scaffolding for `[Threadsafe]` instances
+
+In [ObjectTemplate.rs](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/templates/ObjectTemplate.rs), check the `threadsafe` property of the object definition. If it's true, use `NonLockingHandleMap` instead of `ConcurrentHandleMap` in the [lazy static that declares the handlemap for that interface](https://github.com/mozilla/uniffi-rs/blob/803bb3d79daa8ea088fb2d8f05c08ada09821986/uniffi_bindgen/src/templates/ObjectTemplate.rs#L11).
+
+By ensuring that `NonLockingHandleMap` and `ConcurrentHandleMap` expose a similar API, this will hopefully
+be a fairly minimal if-then-else that just chooses the correct name of the struct to use.
```
jhugman, line 180: I expect we're going to have to delve into the method bodies themselves, to add or remove `mut` annotations in various places.

jhugman (approved):
This looks like a really nice path forward.

docs/adr/0003-threadsafe-interfaces.md
rfk: Yes, I think you're right; we may even find we want to define a small `HandleMap` trait of our own and adapt `ConcurrentHandleMap` to it, rather than tying ourselves to to the API it already offers.

dmose (commented):
Any chance of word-wrapping this more narrowly?  As is, making it readable in the comment view of a github diff window means dealing with very tiny fonts.

docs/adr/0003-threadsafe-interfaces.md
```diff
@@ -0,0 +1,196 @@
...
+if present. The way that we [handle the `[ByRef]` annotation on method arugments](https://github.com/mozilla/uniffi-rs/blob/803bb3d79daa8ea088fb2d8f05c08ada09821986/uniffi_bindgen/src/interface/mod.rs#L658)
+will likely serve as a good example to follow.
+
+### Implement a less-locking HandleMap variant
+
+The `ffi_support` crate provides a basic non-locking [`HandleMap<T>`](https://docs.rs/ffi-support/0.4.0/ffi_support/handle_map/struct.HandleMap.html) struct,
+and it implements `ConcurrentHandleMap` as a thin wrapper around a `RwLock<HandleMap<Mutex<T>>>`. We can make our own variant that removes the inner mutex, as a thin wrapper around a`RwLock<HandleMap<T>>`. (The outer `RwLock` is still needed in order to guard mutations of the handlemap itself, and we must be careful to only hold it for short periods).
+
+Bikeshed name: `NonLockingHandleMap`, implemented under the [`uniffi::ffi` module](https://github.com/mozilla/uniffi-rs/tree/main/uniffi/src/ffi). It needs to live in the `uniffi` crate so that it can be used by the generated Rust scaffolding at runtime.
+
```
dmose, line 164: How about 'LessLockingHandleMap' or 'MinimalLockingHandleMap', since there _is_ still a lock in there, and projects that don't require a singleton (i.e. not Nimbus) could still potentially block on locks during object construction/destruction?  If this is too edge-casey to scare all code readers about, it'd be nice to at least document the possibility in the reference.

docs/adr/0003-threadsafe-interfaces.md
```diff
@@ -0,0 +1,196 @@
...
+By ensuring that `NonLockingHandleMap` and `ConcurrentHandleMap` expose a similar API, this will hopefully
+be a fairly minimal if-then-else that just chooses the correct name of the struct to use.
+
+### Optional: Add a `[Blocking]` annotation to methods and functions
+
+As a nod toward a possible future where uniffi can provide more help in dealing with blocking vs
+async I/O, we could also support annotating individual methods/functions to indicate whether they
+may block. We don't have concrete plans to *use* this information right now, but it could serve
+as useful documentation to consumers of the interface.
```
dmose, line 187: I would assert that we don't really want this to be optional, in the sense that I think there's high value in (at the very least) encouraging method implementors to document if they're going to make something blocking as part of their contract with the API consumer.

That said, I almost want to suggest different semantics: have a developer affordance to require the developer to explicitly think through and document what their intent is at the time the code is written.  Which is to say, have [Blocking] and [NonBlocking] attributes, and disallow attribute-free methods.

dmose (commented):
Note that I'm still new to Rust (i.e. I just learned about `Sync` and `Send` in order to do this review), so there's a decent bit of stuff here that I don't have tons of context around (eg `Box` proposals).  So definitely don't land this with just my review.  :-)

rfk:
> Any chance of word-wrapping this more narrowly?

I'll give it a shot, but also, that's why there's a link to the rendered view in the issue comment 😁

docs/adr/0003-threadsafe-interfaces.md
rfk: Oh, interesting suggestion. I think I'd be OK with trying that out as a required annotation and see if we like it, as we can always remove the restriction later.

dmose: Indeed; that's the nice thing about only having one real client -- it's still pretty easy to experiment with stuff like this.  :-)

dmose: I guess constructors would need to be declared as [Blocking], since there is gonna be some locking around the handlemap...

rfk: I'm not sure what we intend the semantics of `[Blocking]` to mean, but subjectively, there feels like a significant difference between "blocks on network or disk I/O" and "takes a mutex protecting a purely-in-memory data structure". *Any* method call on an object is going to have to interact with the handlemap mutex, so I worry that considering handlemap access a blocking operation will eliminate any documentary value that the annotations may provide.

rfk: By contrast, I think there *is* value in being able to know whether or not the constructor does disk I/O via some sort of annotation.

docs/adr/0003-threadsafe-interfaces.md
rfk: FWIW I will not block (hah!) this on the selection of any particular name; all of the suggestions so far are acceptable to me and I'll be happy to have a verbose name if it conveys an accurate mental model of how the thing behaves. (Which I feel like `ConcurrentHandleMap` doesn't necessarily do - is it "concurrent" because it's safe to use with concurrent access, or because it enables lots of concurrency?)

docs/adr/0003-threadsafe-interfaces.md
```diff
@@ -0,0 +1,196 @@
...
+Make no changes to uniffi, and instead accept the fact that method calls are executed serially.
+Document this limitation and work with consumers to update their API definitions to account for it.
+
+* Good, because we can ship this quickly.
+* Good, because we don't give up any of the safety guarantees of the current approach.
+* Bad, because it means non-trivial rework for one of our key early consumers.
+* Bad, because it makes uniffi less attractive to potential future consumers.
+* Bad, because it keeps all the run-time overhead of the handlemap.
+* Bad, because the default behaviour still has a hidden mutex, which might be a nasty surprise
+  for future consumers in the same way that it was for Nimbus.
```
mhammond, line 67: Not mentioned here is that it's a *requirement* for nimbus which would have been impossible to meet if we did nothing.

docs/adr/0003-threadsafe-interfaces.md
```diff
@@ -0,0 +1,196 @@
...
+## Decision Drivers <!-- optional -->
+
+* Enabling consumers to control the potential blocking behaviour of their generated APIs.
+* Ensure safety of the generated code.
+* Ship a solution in a timely manner to unblock Project Nimbus.
+
+## Considered Options
+
+* Option A: Do nothing, require consumers to assume that all method calls might block.
+* Option B: Let consumers mark interface definitions as threadsafe to opt in to a non-locking handlemap.
```
mhammond, line 32: I had a mild reaction to "non-locking handlemap" as it might imply no locks are necessary at all, but I expect locks will be necessary to protect the handlemap itself? Not sure it's worth trying to squeeze this in though.

docs/adr/0003-threadsafe-interfaces.md
```diff
@@ -0,0 +1,196 @@
...
+existing `ConcurrentHandleMap` by default.
+
+* Good, because this is a relatively small change from the current behaviour, which should be fairly
+  quick to ship.
+* Good, because this is a non-breaking change for consumers who don't opt in to it.
+* Good, because the `Sync` and `Send` bounds on the underlying struct will help consumers
+  to implement their own fine-grained concurrency control while being supported by Rust's
+  compile-time guarantees.
+* Good, because consumers don't have to care about this until they discover that they need
+  fine-grained locking.
```
mhammond, line 90: "Good, because it leaves the door open to "Option C" in the future if we decide that makes sense later."?

docs/adr/0003-threadsafe-interfaces.md
mhammond: Agreed, the existing name is itself bad - so my 2c would be something like "NonConcurrent..." - ie, if you understand how the term "concurrent" applies to the existing one, you automagically understand what's different about the new one.

docs/adr/0003-threadsafe-interfaces.md
mhammond: I agree with all the above, but in the short term, I believe the main value is documenting expectations, so would like to see *something* here sooner rather than later (eg, we will have `getExperimentBranch()` which doesn't "block", but now there's a `getExperimentBranches()` function which, IIUC, does *not* need these semantics - which will not be at all obvious without some url annotation

mhammond (approved):
Lovely! Most of my comments don't require changes, but I thought them worth noting.

mhammond:
A random thought that will probably go nowhere:

For the sake of the strawman, let's say the annotation on the method could be [Blocking] and [Not-Blocking]. Any function marked as [Blocking] is expected to take a lock. It's not clear how uniffi might be able to help here - taking a mutex on those functions isn't that helpful as the entire object needs to be Send+Sync, and this isn't really the right granularity. But maybe there could be another Rust method called, and *that* is passed in.

To make this clearer - in the nimbus PR, we have something like:

```rust
struct Nimbus {
  db: Mutex<Database>
}

impl Nimbus {
  pub fn blocking(&self, ...) {
    let db = self.some_mutex.lock().unwrap();
    ...
  }
  pub fn not_blocking(&self, ...) {
    ... // doesn't take the mutex
  }
}

```

Whereas I can imagine something like:
```rust

struct ProtectedData {
  db: Database,
}

struct Nimbus {
  data: Mutex<ProtectedData>
}

impl Nimbus {
  // called by uniffi for all functions marked as [Locking] and the result is passed as the 2nd param to the functions
  fn uniffi_lock(&self) -> Handwave {
    self.data.lock.handwave()
  }

  pub fn blocking(&self, data: &ProtectedData, ...) {
  }
  pub fn non_blocking(&self) {
  }
}
```

and uniffi still manages the mutex for us. Or something.

1/2 baked, and we certainly don't want to block (ha!) on this, but there might be a grain of a way nimbus can help enforce semantics and reduce boilerplate. A key limitation is that it removes flexibility, so we might *still* end up needing a way to overwrite the semantics if someone truly has free-threading requirements, or at least invent a way that the app itself is in charge of the semantics.

dmose:
A very interesting idea; would love to hear others' thoughts on it...

docs/adr/0003-threadsafe-interfaces.md
```diff
@@ -0,0 +1,196 @@
...
+would grow a corresponding boolean `threadsafe` field and corresponding public accessor.
+When [building an instance of this struct from the UDL](https://github.com/mozilla/uniffi-rs/blob/803bb3d79daa8ea088fb2d8f05c08ada09821986/uniffi_bindgen/src/interface/mod.rs#L786),
+we would inspect the list of attributes to look for one named "Threadsafe" and set the field
+if present. The way that we [handle the `[ByRef]` annotation on method arugments](https://github.com/mozilla/uniffi-rs/blob/803bb3d79daa8ea088fb2d8f05c08ada09821986/uniffi_bindgen/src/interface/mod.rs#L658)
+will likely serve as a good example to follow.
+
+### Implement a less-locking HandleMap variant
+
+The `ffi_support` crate provides a basic non-locking [`HandleMap<T>`](https://docs.rs/ffi-support/0.4.0/ffi_support/handle_map/struct.HandleMap.html) struct,
+and it implements `ConcurrentHandleMap` as a thin wrapper around a `RwLock<HandleMap<Mutex<T>>>`. We can make our own variant that removes the inner mutex, as a thin wrapper around a`RwLock<HandleMap<T>>`. (The outer `RwLock` is still needed in order to guard mutations of the handlemap itself, and we must be careful to only hold it for short periods).
```
rfk, line 161: Ref https://github.com/mozilla/uniffi-rs/pull/372#discussion_r558620397, this type is not sufficient because I don't think Rust will let us release the `RwLock` while there is an outstanding reference to an item from the map. As well it shouldn't: if we let one thread operate on an `&T` while a concurrent thread deletes the underlying `T` from the handlemap, we'd risk running `T::drop` while the `&T` was still alive.

I expect this will need to be `RwLock<HandleMap<Arc<T>>>` so that we can reference-count our way around this issue. Leaving a note for myself to update the final ADR before merge.

dmose:
I've filed a separate issue (#378) on the [Blocking]/[Non-Blocking] implementation itself.  Maybe the specific proposal wants to be nailed down before this PR is landed, or maybe not.

rfk:
I've updated the doc to fold in the comments above, and added links to the implementation PR and follow-up issue from `[Blocking]`. I'll go ahead and merge this once CI gives me the all-clear. Thanks folks!

Issue #370, opened 2021-01-14
Implement the concurrency-control solution from ADR-0003


rfk:
In #368 we are proposing some changes to the way that uniffi handles concurrent access to object instances. That bug is specifically for agreeing on and documenting a plan, so I'm filing here a separate bug to track the work of implementing said plan.

The ADR contains an [implementation sketch](https://github.com/mozilla/uniffi-rs/blob/adr-threadsafe-interface/docs/adr/0003-threadsafe-interfaces.md#implementation-sketch) section with an outline of the work to be done here.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2005)

rfk:
@jhugman volunteered to take a look at the implementation work here, so I'm optimistically assigning him this issue.

data-sync-user:
➤ Janet Dragojevic commented:

closing in favor of SDK-170

data-sync-user:
➤ Ryan Kelly commented:

Closing as a dupe of https://jira.mozilla.com/browse/SDK-170

PR #371, opened 2021-01-14
Add narrative docs for threadsafe instances.

rfk:
I wrote these docs are part of working through the proposal in https://github.com/mozilla/uniffi-rs/pull/368 and figured I may as well push them for visibility. They may or may not turn out to be a useful starting point for the implementation work connected to #370. /cc @jhugman

### DIFF
--- a/docs/manual/src/internals/lifting_and_lowering.md
+++ b/docs/manual/src/internals/lifting_and_lowering.md
@@ -29,7 +29,7 @@ namespace example {
 
 Calling this function from foreign language code involves the following steps:
 
-1. The user-provided calling code invokes the `add_to_list` function that is expoed by the
+1. The user-provided calling code invokes the `add_to_list` function that is exposed by the
    uniffi-generated foreign language bindings, passing `item` as an appropriate language-native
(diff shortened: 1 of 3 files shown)


rfk:
(N.B. this PR currently targets #367 rather than `main`, since it builds on top of the docs added therein)

docs/manual/src/internals/lifting_and_lowering.md
```diff
@@ -0,0 +1,93 @@
...
+
+```idl
+namespace example {
+  sequence<i32> add_to_list(i32 item);
+}
+```
+
+Calling this function from foreign language code involves the following steps:
+
+1. The user-provided calling code invokes the `add_to_list` function that is expoed by the
```
mhammond, line 31: nit: exposed

mhammond (approved):
I didn't get to the end up this, but found a nit I thought I'd share before I take off for the weekend :)

rfk:
This was lightly updated to reflect that actually merged, but I'm going to land it on the strength of Mark's r+.

Issue #286, opened 2020-09-18
Split up `uniffi_bindgen/src/interface` into smaller files


rfk:
GitHub PRs have started low-key calling me out by hiding changes to `uniffi_bindgen/src/interface/mod.rs` and `uniffi_bindgen/src/interface/types.rs` by default, which is probably a sign that these gigantic files need to be split up into more manageable chunks. Off the top of my head, I'm thinking:

* `interface/mod.rs`    <-- just the `ComponentInterface` struct, maybe handling of namespace
* `interface/record.rs`  <-- separate files for the major chunks of interface functionality
* `interface/object.rs`
* `interface/function.rs`
* `interface/enum.rs`
* `interface/error.rs`
* `interface/types/mod.rs`     <-- just `Type` and `TypeUniverse`, no ffi stuff or webidl stuff
* `interface/types/finder.rs`   <- `TypeFinder` and its impls
* `interface/types/resolver.rs`  <- the suboptimally-named `TypeResolver` and its impls
* `interface/ffi/mod.rs`   <-- `FFIFunction` and friends
* `interface/ffi/types.rs`  <-- `FFIType`

But, probably we should just try it out and feel out what the right split is as we go.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1780)

PR #372, opened 2021-01-15
SDK-170 Add a Threadsafe annotation for interfaces

jhugman:
Fixes [SDK-170][1].

This follows the implementation plan in #368.

1. We add a `Threadsafe` annotation for UDL; this is a marker to show that the client Rust library is declaring itself as able to do its own locking.
2. We add a second implementation of `ffi_support::HandleMap`, called `NonLockingHandleMap`. `NonLockingHandleMap` gives out `&T`; `ConcurrentHandleMap` gives out `&mut T`. 
3. We use the presence/absence of the `Threadsafe` annotation to choose which handle map to use for a given object.
4. We show with a kotlin example that we can exercise the Rust code from two threads _at the same time_.

[1]: https://jira.mozilla.com/browse/SDK-170


uniffi/src/ffi/handle_map.rs
```diff
@@ -0,0 +1,243 @@
+use ffi_support::ExternError;
+use ffi_support::Handle;
+use ffi_support::HandleError;
+use ffi_support::HandleMap;
+use ffi_support::IntoFfi;
+use std::sync::RwLock;
```
jhugman, line 6: This file is copied from the `ffi_support::handle_map::ConcurrentHandleMap`, and edited.

TODO:

 * test test test. I haven't yet copied over the tests, and I don't know what will still make sense.
 * re-write/edit the documentation,

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1303,6 +1318,22 @@ impl Attributes {
...
+    fn validate_for_object(&self) -> Result<()> {
+        for attr in self.0.iter() {
+            match attr {
+                Attribute::Threadsafe => continue,
+                _ => bail!(format!("{:?} not supported for Objects", attr)),
+            }
+        }
+        Ok(())
+    }
+}
```
jhugman, line 1331: Currently we're not validating which attributes are used for each syntactical construct. I propose we should.

uniffi_bindgen/src/scaffolding.rs
```diff
@@ -22,6 +22,18 @@ mod filters {
...
+        expr_value: &bool,
+        then_value: &dyn fmt::Display,
+        else_value: &dyn fmt::Display,
+    ) -> Result<String, askama::Error> {
+        Ok(if *expr_value {
+            then_value.to_string()
+        } else {
+            else_value.to_string()
+        })
+    }
```
jhugman, line 35: Note to self: raise an issue with @djc to add a default filter to askama.

jhugman (commented):
The lack of tests and documentation in `NonLockingHandleMap` mean this isn't ready to land, but please feel free to review.

uniffi_bindgen/src/scaffolding.rs
djc: For simple cases, you might just be able to use `then_value.unwrap_or(else_value.to_string())` in the template, but I think it could some facilities to Askama to facilitate patterns like this.

uniffi/src/ffi/handle_map.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// (It may also panic if the handle map detects internal state corruption,
+    /// however this should not happen except for bugs in the handle map code).
+    pub fn get<F, E, R>(&self, h: Handle, callback: F) -> Result<R, E>
+    where
+        F: FnOnce(&T) -> Result<R, E>,
+        E: From<HandleError>,
+    {
+        let map = self.map.read().unwrap();
+        let hm = map.get(h)?;
+        callback(&*hm)
```
rfk, line 136: Hrm, so, this is where we get into the tricky bits. IIUC what we want to do here is:

* Take the read lock
* Read the instance out of the handlemap
* Release the read lock
* Operate on the retrieved instance.

In order to do that, you'll need the lock guard `map` to go out of scope, similar to what's done in the implementation of `delete`. Something like:

```
let obj = {
    let map = self.map.read().unwrap();
    map.get(h)?;
}
callback(&*obj)
```

However! Rust probably won't let that code compile as written above, because it won't let you keep a reference to `&obj` alive after releasing the read lock that was guarding it. If it did, a concurrent thread might remove the object from the handlemap, running its destructor while there's still a live reference held by another thread.

To make the Rust compiler happy, we'll probably need to take a *copy* of the object out of the handlemap, so that it knows other threads can safely drop the instance contained therein:

```
let obj = {
    let map = self.map.read().unwrap();
    map.get(h)?.clone(); // <-- we're going to operate on a copy of the object
}
callback(&*obj)
```

But of course we don't want to actually copy our instances! The semantics we want here are reference counting: don't run the destructor until the instance is removed from the handle *and* there are zero concurrent threads holding a reference to the instance. That's what [Arc](https://doc.rust-lang.org/std/sync/struct.Arc.html) is for.

So, I think we'll have to actually make the type here be:

```
pub struct NonLockingHandleMap<T>
where
    T: Sync + Send,
{
    // Put each instance in an Arc, for refernece-counting.
    pub map: RwLock<HandleMap<Arc<T>>>,
}
```

If we do so, the `.clone()` version of the code above will increment a reference count instead of doing an actual copy, giving us the semantics we want.

rfk: (This sounds terrifically complicated, but I'm glad of the way Rust is forcing us to work through this explicitly!)

An alternative could be to reach for unsafe code here, since our generated bindings will ensure that we never actually try to drop from the handlemap until all active method calls have completed. But I think I like the additional level of protection against bugs in the generated code, at least for now.

uniffi/src/ffi/handle_map.rs
```diff
@@ -0,0 +1,243 @@
...
+        F: std::panic::UnwindSafe + FnOnce(&T) -> Result<R, E>,
+        ExternError: From<E>,
+        R: IntoFfi,
+    {
+        use ffi_support::call_with_result;
+        call_with_result(out_error, || -> Result<_, ExternError> {
+            // We can't reuse get_mut here because it would require E:
+            // From<HandleError>, which is inconvenient...
+            let h = Handle::from_u64(h)?;
+            let map = self.map.read().unwrap();
```
rfk, line 181: A similar comment applies here re: letting `map` go out of scope as quickly as possible, to avoid holding the lock during the method call.

uniffi_bindgen/src/interface/mod.rs
rfk: +1, yes, we should validate that e.g. you're not setting `[Throws`] on an `instance` declaration. You could consider using separate structs for this, e.g. `InstanceAttributes` and `FunctionAttributes` rather than a single `Attributes` struct, which is probably going to seem a bit verbose with this all-in-one-file approach, but would make sense if we split out the core types into their own separate files as proposed in https://github.com/mozilla/uniffi-rs/issues/286.

uniffi_bindgen/src/templates/macros.rs
```diff
@@ -50,14 +50,15 @@ UNIFFI_HANDLE_MAP_{{ obj.name()|upper }}.insert_with_output(err, || {
 {% endmacro %}
 
 {% macro to_rs_method_call(obj, meth) %}
+{% let mut_suffix = obj.threadsafe()|choose("", "_mut") -%}
```
rfk, line 53: Hrm, this `mut_suffix` business seems unfortunate. I feel like we might be able to clean this up by using a custom trait, along the lines of this but with better names:

```
trait UniffiMethodCall {
  pub fn call_method_with_result(&self, err, callback) -> Result;
}

impl UniffiMethodCall for ConcurrentHandleMap {
  pub fn call_method_with_result(&self, err, callback) -> Result {
      // do the mut thing internally
      self.call_with_result_mut(...)
  }
}

impl UniffiMethodCall for NonLockingHandleMap {
  pub fn call_method_with_result(&self, err, callback) -> Result {
      // no mut here
      self.call_with_result(...)
  }
}
```

But that may be more fiddling around is really necessary, and I'd be happy to land the simpler `mut_suffix` version to get this out the door.

rfk (commented):
This is shaping up nicely, thanks @jhugman! (Just thought I'd pop in for a quick look before heading out for PTO).

This has also highlighted a deficiency in the implementation proposal from the ADR, inasmuch as the Rust compiler will prevent us from release the locks as early as we'd like if we try to do it naively. I've added some comments below about using `Arc` to make things safe, and will cross-link with the ADR so that it can be updated as well.

uniffi/src/ffi/handle_map.rs
rfk: I'm also curious if the test as written would have uncovered the fact that this method is unintentionally holding the `RwLock` for the duration of the method call...

uniffi/src/ffi/handle_map.rs
```diff
@@ -0,0 +1,243 @@
+use ffi_support::ExternError;
+use ffi_support::Handle;
+use ffi_support::HandleError;
+use ffi_support::HandleMap;
+use ffi_support::IntoFfi;
+use std::sync::RwLock;
+
+pub struct NonLockingHandleMap<T>
```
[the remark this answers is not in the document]
rfk: The discussion about `Arc` below makes me wonder if a better pair of names would be `MutexHandleMap` and `ArcHandleMap` or similar. Dunno, naming is hard :-/

uniffi/src/ffi/handle_map.rs
badboy: +1 from me on using an Arc. Even if we control the generated bindings around it right now, we shouldn't pre-optimize that using unsafe code if Arc does the job for us.
IMO it's also good to explicitly call `Arc::clone(&obj)` to make it obvious to the reader that the clone is _cheap_

examples/threadsafe/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+struct ThreadsafeCounter {
+    is_busy: AtomicBool,
+    count: AtomicI32,
+}
+
+unsafe impl Send for ThreadsafeCounter {}
+unsafe impl Sync for ThreadsafeCounter {}
```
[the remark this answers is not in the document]
jhugman: This is a good point, and well made.

I've added a `static_assertions::assert_implements_all!` for good measure.

uniffi/src/ffi/handle_map.rs
jhugman: > I'm also curious if the test as written would have uncovered the fact that this method is unintentionally holding the RwLock for the duration of the method call...

I'm not rightly sure.

The test above is calling a long running method on one thread, then calling many times a short running method on another.

Your analysis would suggest the `read` lock around the handle map is preventing concurrent access? So I wonder if the following would happen:

 * the chance of inadvertent locking is low, so if we run the test multiple times, perhaps we'd see a failure?
 * the number of `incrementIfBusy()` calls that increment the count changes between implementations?
 * the test should fail everytime? This is not happening.

jhugman: FWIW: the number of `incrementIfBusy()` calls hasn't noticeably changed between the two implementations.

uniffi_bindgen/src/templates/macros.rs
jhugman: This definitely looks better than changing to code that gets generated.

It's more complicated than I thought, because the callback type for the two implementations varies as `FnOnce(&T)` and `FnOnce(&mut T)`.

uniffi_bindgen/src/scaffolding.rs
jhugman: > `then_value.unwrap_or(else_value.to_string())`

I don't _think_ this is what I'm asking for. Your suggestion in trinary form looks like: 

`then_value != null ? then_value : else_value`

I'm looking for 

`expr_value ? then_value : else_value`

You _can_ express what my implementation does in askama:

```
{% let string %}
{% if expr_value %}
{% let string = "then value" %}
{% else %}
{% let string = "else value" %}
```

But it's quite verbose.

uniffi/src/ffi/handle_map.rs
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    /// Convenient wrapper for `get` which takes a `u64` that it will convert to
+    /// a handle.
+    ///
+    /// The other benefit (besides convenience) of this over the version
+    /// that takes a [`Handle`] is that it allows handling handle-related errors
+    /// in one place.
+    ///
+    /// This takes the map's `read` lock.
```
mhammond, line 143: nit: this comment about the readlock should probably be the same as, say, `call_with_result`, otherwise the casual reader might think this method take the lock for the entire time?

uniffi/src/ffi/handle_map.rs
```diff
@@ -0,0 +1,414 @@
...
+        self.call_with_result(out_error, h, |r| -> Result<_, HandleError> {
+            Ok(callback(r))
+        })
+    }
+
+    /// Use `constructor` to create and insert a `T`, while inside a
+    /// [`call_with_result`] call (to handle panics and map errors onto an
+    /// `ExternError`).
+    ///
+    /// This takes the map's `write` lock.
```
mhammond, line 201: nit: same comment here re locks - the write lock is not held while the ctor runs, which is probably worth explicitly noting.

(As I mentioned elsewhere, it might make sense to document the lock semantics on the struct itself and avoid duplicating it for every method call - best I can tell, the semantics are identical in all cases (other than read vs write locks)

uniffi/src/ffi/handle_map.rs
```diff
@@ -0,0 +1,414 @@
...
+    ) -> R::Value
+    where
+        F: std::panic::UnwindSafe + FnOnce(&T) -> R,
+        R: IntoFfi,
+    {
+        self.call_with_output(out_error, h, callback)
+    }
+}
+
+/// Tests that check our behavior when panicing.
```
mhammond, line 335: nit: panicking

uniffi/src/ffi/handle_map.rs
```diff
@@ -0,0 +1,414 @@
+use ffi_support::{ExternError, Handle, HandleError, HandleMap, IntoFfi};
```
mhammond, line 1: nit: `handle_map.rs` seems a little unfortunate since it doesn't actually expose an object called HandleMap, even though such a name is used heavily. `arc_handle_map.rs` seems a bit less surprising, but is long. I don't really care about this, but thought it worth mentioning.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -808,6 +816,11 @@ impl APIConverter<Object> for weedle::InterfaceDefinition<'_> {
         if object.constructors.is_empty() {
             object.constructors.push(Default::default());
         }
+
+        object.threadsafe = attributes
```
mhammond, line 820: (jeez, sometimes I hate rustfmt :)

mhammond (commented):
A drive-by - this is looking great! Just a few nits around comments/docs that you should feel free to do what you like with.

uniffi_bindgen/src/interface/mod.rs
jhugman: Raised #373. This is worth doing, just not in this pull request. I've got a PR ready to go, but I don't want to slow down reviews.

examples/threadsafe/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use static_assertions;
+use std::sync::atomic::{AtomicBool, AtomicI32, Ordering};
+
+/// Simulation of a task doing something to keep a thread busy.
+/// Up until now, everything has been synchronous and blocking, so
+/// we can use this to simulate a CPU bound or I/O task.
+fn busy_wait(cycles: i32) -> i32 {
```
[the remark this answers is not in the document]
jhugman: I didn't think of a sleep! Added now.

uniffi/src/ffi/handle_map.rs
jhugman: Renamed to `handle_maps`, because it defines `ArcHandleMap`, `MutexHandleMap` and `UniffiMethodCall`.

examples/threadsafe/src/lib.rs
```diff
@@ -2,22 +2,22 @@
...
+    use std::{thread, time};
+    let ms = ms as u64;
+    let millis = time::Duration::from_millis(ms);
+    thread::sleep(millis);
 }
 
+// This counter is a naive implementation of a counter that _should_ allow the
+// waiting on one thread, and counting on another thread.
+// It relies on uniffi's default locking mechanisms, provided by the
+// `MutexHandleMap`, so doesn't actually work.
```
mhammond, line 20: nit: maybe clarify what "doesn't actually work" means?

examples/threadsafe/tests/bindings/test_threadsafe.kts
```diff
@@ -42,7 +42,7 @@ fun countWhileBusy(busyWait: () -> Int, incrementIfBusy: () -> Int): Int {
 // long enough for us to incrementIfBusy.
 // If the Rust compiler gets too smart, and optimizes the busy wait loop away,
```
mhammond, line 43: seems extremely unlikely the compiler will optimize the sleeping version away? ISTM the main risk here now is the speed of the hardware?

mhammond (approved):
Lovely work James 👍

badboy:
Retro-active 👍

uniffi/src/ffi/handle_map.rs
rfk: > I'm also curious if the test as written would have uncovered the fact that this method is unintentionally holding the
> RwLock for the duration of the method call

I belatedly realized why the test didn't catch the issue I was worried about here, and wanted to record it for future reference.

This method call was holding a *read* lock for the duration of the call, and since multiple threads can concurrently take a read lock, it would not block running ordinary method calls on other threads. What it would block is any attempt to take a *write* lock on the handlemap, meaning that a long-running method call could have blocked constructors and destructors from running concurrently.

That probably would have been fine for Nimbus, since we only ever construct a single immortal instance. But it still seems worthwhile to use an `Arc` and remove this as a future footgun.

rfk (commented):
Just dropping in after the fact to comment that this is 💯, thanks @jhugman! 🎉 🎉

Issue #37, opened 2020-07-28
Static and/or multiple constructors support


eoger:
In the fxa-client crate the `FirefoxAccount` can be instanced in two ways: either by its regular constructor or by supplying a JSON string representing a previous state. IIUC we don't have support for an alternative interface constructor right now..

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1553)

PR #375, opened 2021-01-21
Refactor attributes to be validate IDL

### LINKED ISSUE #373: Split out Attributes structs into structure specific structs.
As part of #372 @rfk [suggested][1] splitting out conversion and validation of `Attributes` into separate specific structs.

[1]: https://github.com/mozilla/uniffi-rs/pull/372#discussion_r558621792

e.g. `EnumAttributes`, `InterfaceAttributes`, `ConstructorAttributes`.

The use of each of attributes can be put directly in these structs (e.g. `threadsafe()` or `has_error()`).

This issue covers this request, because it's a significant enough piece of work to slow down the review process of #372.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2012)

jhugman:
Fixes #373.

Introduces more structure around the `Attributes` struct.

### DIFF
--- a/uniffi_bindgen/src/interface/mod.rs
+++ b/uniffi_bindgen/src/interface/mod.rs
@@ -434,8 +434,7 @@ impl APIBuilder for weedle::Definition<'_> {
             weedle::Definition::Enum(d) => {
                 // We check if the enum represents an error...
                 let is_error = if let Some(attrs) = &d.attributes {
-                    let attributes = Attributes::try_from(attrs)?;
-                    attributes.contains_error_attr()
+                    EnumAttributes::try_from(attrs)?.contains_error_attr()
                 } else {
                     false
                 };
@@ -526,7 +525,7 @@ pub struct Function {
     arguments: Vec<Argument>,
     return_type: Option<Type>,
     ffi_func: FFIFunction,
-    attributes: Attributes,
+    attributes: FunctionAttributes,
 }
 
 impl Function {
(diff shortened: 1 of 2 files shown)


uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1318,44 +1351,100 @@ impl Attributes {
...
+pub struct ConstructorAttributes(Vec<Attribute>);
+
+impl ConstructorAttributes {
+    fn get_throws_err(&self) -> Option<&str> {
+        self.0.iter().find_map(|attr| match attr {
+            // This will hopefully return a helpful compilation error
+            // if the error is not defined.
+            Attribute::Throws(inner) => Some(inner.as_ref()),
+            _ => None,
+        })
```
rfk, line 1383: It's not obvious to me from reading, why does this need to be its own impl rather than re-using `MethodAttributes`, similar to what we do for `FunctionAttributes` above? In fact I wonder whether we could have *just* the `FunctionAttributes` struct for now and make `Method` and `Constructor` use it directly, until we reach a point where their needs diverge.

rfk (approved):
Looks like a good cleanup, thank you!

uniffi_bindgen/src/interface/mod.rs
jhugman: Oh, good catch. I had done this quite mechanically.

I expect this will only change when we implement #37 .

PR #381, opened 2021-02-08
Add support for enums with associated data (aka "tagged unions", aka "sum types", aka ...)

rfk:
Our fxa-client component makes use of a Rust enum with associated data to represent the different kinds of event that can happen to your account. The [Rust code](https://github.com/mozilla/application-services/blob/9ffdf88cdb269bf6f6769be67802f85983ae8d97/components/fxa-client/src/lib.rs#L276) looks like this:

```
pub enum AccountEvent {
    IncomingDeviceCommand(Box<IncomingDeviceCommand>),
    ProfileUpdated,
    DeviceConnected {
        device_name: String,
    },
    [...other cases omitted...]
}
```

In the hand-written bindings, we [reflect this to Kotlin as a sealed class](https://github.com/mozilla/application-services/blob/9ffdf88cdb269bf6f6769be67802f85983ae8d97/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/AccountEvent.kt#L11) and [reflect it to Swift as an enum with associated data](https://github.com/mozilla/application-services/blob/9ffdf88cdb269bf6f6769be67802f85983ae8d97/components/fxa-client/ios/FxAClient/FxAccountDevice.swift#L127).

UniFFI doesn't yet support this kind of data structure, which makes it awkward to use for the fxa-client component. So let's see if we can add it!

WebIDL doesn't seem to have a notion of a datatype like this or offer any ready syntax for it (It has union types, but they're like C-style unions and do not have an explicit discriminant). I messed around a bit and came up with a syntax that doesn't seem too bad, and is similar to how we're implemented errors. This PR contains just the docs and test changes that would happen *if* we chose to implement tagged unions in this manner. I'm looking for early feedback on whether this seems like a good idea.

/cc @mhammond @lougeniaC64 @jhugman @dmose

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ Things that are implemented so far:
 
 * Primitive numeric types, equivalents to those offered by Rust (`u32`, `f64`, etc).
 * Strings (which are always UTF-8, like Rust's `String`).
-* C-style enums (just the discriminant, no associated data).
+* Enums, including enums with associated data (aka "tagged unions" or "sum types").
 * C-style structs containing named fields (we call these *records*).
 * Sequences of all of the above (like Rust's `Vec<T>`).
 * Optional instances of all of the above (like Rust's `Option<T>`).
@@ -98,7 +98,6 @@ Things that are implemented so far:
 
 Things that are not implemented yet:
 
-* Enums with associated data.
 * Union types.
 * Efficient access to binary data (like Rust's `Vec<u8>`).
 * Passing object references to functions or methods.
--- a/docs/manual/src/internals/lifting_and_lowering.md
+++ b/docs/manual/src/internals/lifting_and_lowering.md
@@ -62,7 +62,7 @@ Calling this function from foreign language code involves the following steps:
 | `T?` | `RustBuffer` struct pointing to serialized bytes |
 | `sequence<T>` | `RustBuffer` struct pointing to serialized bytes |
 | `record<DOMString, T>` | `RustBuffer` struct pointing to serialized bytes |
-| `enum` | `uint32_t` indicating variant, numbered in declaration order starting from 1  |
+| `enum` and `[Enum] interface` | `RustBuffer` struct pointing to serialized bytes |
 | `dictionary` | `RustBuffer` struct pointing to serialized bytes |
 | `interface` | `uint64_t` opaque integer handle |
 
@@ -84,7 +84,7 @@ The details of this format are internal only and may change between versions of 
 | `T?` | If null, serialized `boolean` false; if non-null, serialized `boolean` true followed by serialized `T` |
 | `sequence<T>` | Serialized `i32` item count followed by serialized items; each item is a serialized `T` |
 | `record<DOMString, T>` | Serialized `i32` item count followed by serialized items; each item is a serialized `string` followed by a serialized `T` |
-| `enum` | Serialized `u32` indicating variant, numbered in declaration order starting from 1 |
+| `enum` and `[Enum] interface` | Serialized `i32` indicating variant, numbered in declaration order starting from 1, followed by the serialized values of the variant's fields in declaration order |
 | `dictionary` | The serialized value of each field, in declaration order |
 | `interface` | *Cannot currently be serialized* |
 
--- a/docs/manual/src/udl/enumerations.md
+++ b/docs/manual/src/udl/enumerations.md
@@ -1,6 +1,7 @@
 # Enumerations
 
 An enumeration defined in Rust code as
+
 ```rust
 enum Animal {
     Dog,
@@ -17,4 +18,25 @@ enum Animal {
 };
 ```
 
-Note that enumerations with associated data are not yet supported.
+Enumerations with associated data require a different syntax,
+due to the limitations of using WebIDL as the basis for UniFFI's interface language.
+An enum like this in Rust:
+
+```rust
+enum IpAddr {
+  V4 {q1: u8, q2: u8, q3: u8, q4: u8},
+  V6 {addr: string},
+}
+```
+
+Can be exposed in the UDL file with:
+
+```idl
+[Enum]
+interface IpAddr {
+  V4(u8 q1, u8 q2, u8 q3, u8 q4);
+  V6(string addr);
+};
+```
+
+Only enums with named fields are supported by this syntax.
--- a/examples/rondpoint/src/rondpoint.udl
+++ b/examples/rondpoint/src/rondpoint.udl
@@ -2,7 +2,7 @@ namespace rondpoint {
   Dictionnaire copie_dictionnaire(Dictionnaire d);
   Enumeration copie_enumeration(Enumeration e);
   sequence<Enumeration> copie_enumerations(sequence<Enumeration> e);
-  record<DOMString, Enumeration> copie_carte(record<DOMString, Enumeration> c);
+  record<DOMString, EnumerationAvecDonnees> copie_carte(record<DOMString, EnumerationAvecDonnees> c);
   boolean switcheroo(boolean b);
 };
 
@@ -18,6 +18,13 @@ enum Enumeration {
     "Un",
     "Deux",
     "Trois",
+};
+
+[Enum]
+interface EnumerationAvecDonnees {
+  Zero();
+  Un(u32 premier);
+  Deux(u32 premier, string second);
 };
 
 dictionary Dictionnaire {
--- a/examples/rondpoint/tests/bindings/test_rondpoint.kts
+++ b/examples/rondpoint/tests/bindings/test_rondpoint.kts
@@ -6,7 +6,24 @@ assert(dico == copyDico)
 
 assert(copieEnumeration(Enumeration.DEUX) == Enumeration.DEUX)
 assert(copieEnumerations(listOf(Enumeration.UN, Enumeration.DEUX)) == listOf(Enumeration.UN, Enumeration.DEUX))
-assert(copieCarte(mapOf("1" to Enumeration.UN, "2" to Enumeration.DEUX)) == mapOf("1" to Enumeration.UN, "2" to Enumeration.DEUX))
+assert(copieCarte(mapOf(
+    "0" to EnumerationAvecDonnees.Zero,
+    "1" to EnumerationAvecDonnees.Un(1u),
+    "2" to EnumerationAvecDonnees.Deux(2u, "deux")
+)) == mapOf(
+    "0" to EnumerationAvecDonnees.Zero,
+    "1" to EnumerationAvecDonnees.Un(1u),
+    "2" to EnumerationAvecDonnees.Deux(2u, "deux")
+))
+
+val var1: EnumerationAvecDonnees = EnumerationAvecDonnees.Zero
+val var2: EnumerationAvecDonnees = EnumerationAvecDonnees.Un(1u)
+val var3: EnumerationAvecDonnees = EnumerationAvecDonnees.Un(2u)
+assert(var1 != var2)
+assert(var2 != var3)
+assert(var1 == EnumerationAvecDonnees.Zero)
+assert(var1 != EnumerationAvecDonnees.Un(1u))
+assert(var2 == EnumerationAvecDonnees.Un(1u))
 
 assert(switcheroo(false))
 
--- a/uniffi_bindgen/src/bindings/gecko_js/templates/SharedHeaderTemplate.h
+++ b/uniffi_bindgen/src/bindings/gecko_js/templates/SharedHeaderTemplate.h
@@ -29,13 +29,17 @@ namespace dom {
 namespace {{ context.detail_name() }} {
 
 {% for e in ci.iter_enum_definitions() %}
+{% if !e.is_flat() %}
+MOZ_STATIC_ASSERT(false, "Sorry the gecko-js backend does not yet support enums with associated data: {{ e.name() }}");
+{% else %}
 template <>
-struct ViaFfi<{{ e.name()|class_name_cpp(context) }}, uint32_t> {
-  [[nodiscard]] static bool Lift(const uint32_t& aLowered, {{ e.name()|class_name_cpp(context) }}& aLifted) {
-    switch (aLowered) {
+struct Serializable<{{ e.name()|class_name_cpp(context) }}> {
+  [[nodiscard]] static bool ReadFrom(Reader& aReader, {{ e.name()|class_name_cpp(context) }}& aValue) {
+    auto variant = aReader.ReadInt32();
+    switch (variant) {
       {% for variant in e.variants() -%}
       case {{ loop.index }}:
-        aLifted = {{ e.name()|class_name_cpp(context) }}::{{ variant|enum_variant_cpp }};
+        aValue = {{ e.name()|class_name_cpp(context) }}::{{ variant.name()|enum_variant_cpp }};
         break;
       {% endfor -%}
       default:
@@ -45,30 +49,18 @@ struct ViaFfi<{{ e.name()|class_name_cpp(context) }}, uint32_t> {
     return true;
   }
 
-  [[nodiscard]] static uint32_t Lower(const {{ e.name()|class_name_cpp(context) }}& aLifted) {
-    switch (aLifted) {
+  static void WriteInto(Writer& aWriter, const {{ e.name()|class_name_cpp(context) }}& aValue) {
+    switch (aValue) {
       {% for variant in e.variants() -%}
-      case {{ e.name()|class_name_cpp(context) }}::{{ variant|enum_variant_cpp }}:
-        return {{ loop.index }};
+      case {{ e.name()|class_name_cpp(context) }}::{{ variant.name()|enum_variant_cpp }}:
+        aWriter.WriteInt32({{ loop.index }});
       {% endfor -%}
       default:
         MOZ_ASSERT(false, "Unknown raw enum value");
     }
-    return 0;
   }
 };
-
-template <>
-struct Serializable<{{ e.name()|class_name_cpp(context) }}> {
-  [[nodiscard]] static bool ReadFrom(Reader& aReader, {{ e.name()|class_name_cpp(context) }}& aValue) {
-    auto rawValue = aReader.ReadUInt32();
-    return ViaFfi<{{ e.name()|class_name_cpp(context) }}, uint32_t>::Lift(rawValue, aValue);
-  }
-
-  static void WriteInto(Writer& aWriter, const {{ e.name()|class_name_cpp(context) }}& aValue) {
-    aWriter.WriteUInt32(ViaFfi<{{ e.name()|class_name_cpp(context) }}, uint32_t>::Lower(aValue));
-  }
-};
+{% endif %}
 {% endfor %}
 
 {% for rec in ci.iter_record_definitions() -%}
--- a/uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
@@ -1,19 +1,88 @@
+{#
+// Kotlin's `enum class` constuct doesn't support variants with associated data,
+// but is a little nicer for consumers than its `sealed class` enum pattern.
+// So, we switch here, using `enum class` for enums with no associated data
+// and `sealed class` for the general case.
+#}
+
+{% if e.is_flat() %}
+
 enum class {{ e.name()|class_name_kt }} {
-    {% for variant in e.variants() %}
-    {{ variant|enum_variant_kt }}{% if loop.last %};{% else %},{% endif %}
-    {% endfor %}
+    {% for variant in e.variants() -%}
+    {{ variant.name()|enum_variant_kt }}{% if loop.last %};{% else %},{% endif %}
+    {%- endfor %}
 
     companion object {
-        internal fun lift(n: Int) =
-            try { values()[n - 1] }
+        internal fun lift(rbuf: RustBuffer.ByValue): {{ e.name()|class_name_kt }} {
+            return liftFromRustBuffer(rbuf) { buf -> {{ e.name()|class_name_kt }}.read(buf) }
+        }
+
+        internal fun read(buf: ByteBuffer) =
+            try { values()[buf.getInt() - 1] }
             catch (e: IndexOutOfBoundsException) {
                 throw RuntimeException("invalid enum value, something is very wrong!!", e)
             }
-
-        internal fun read(buf: ByteBuffer) = lift(buf.getInt())
     }
 
-    internal fun lower() = this.ordinal + 1
+    internal fun lower(): RustBuffer.ByValue {
+        return lowerIntoRustBuffer(this, {v, buf -> v.write(buf)})
+    }
 
-    internal fun write(buf: RustBufferBuilder) = buf.putInt(this.lower())
+    internal fun write(buf: RustBufferBuilder) {
+        buf.putInt(this.ordinal + 1)
+    }
 }
+
+{% else %}
+
+sealed class {{ e.name()|class_name_kt }} {
+    {% for variant in e.variants() -%}
+    {% if !variant.has_fields() -%}
+    object {{ variant.name()|class_name_kt }} : {{ e.name()|class_name_kt }}()
+    {% else -%}
+    data class {{ variant.name()|class_name_kt }}(
+        {% for field in variant.fields() -%}
+        val {{ field.name()|var_name_kt }}: {{ field.type_()|type_kt}}{% if loop.last %}{% else %}, {% endif %}
+        {% endfor -%}
+    ) : {{ e.name()|class_name_kt }}()
+    {%- endif %}
+    {% endfor %}
+
+    companion object {
+        internal fun lift(rbuf: RustBuffer.ByValue): {{ e.name()|class_name_kt }} {
+            return liftFromRustBuffer(rbuf) { buf -> {{ e.name()|class_name_kt }}.read(buf) }
+        }
+
+        internal fun read(buf: ByteBuffer): {{ e.name()|class_name_kt }} {
+            return when(buf.getInt()) {
+                {%- for variant in e.variants() %}
+                {{ loop.index }} -> {{ e.name()|class_name_kt }}.{{ variant.name()|class_name_kt }}{% if variant.has_fields() %}(
+                    {% for field in variant.fields() -%}
+                    {{ "buf"|read_kt(field.type_()) }}{% if loop.last %}{% else %},{% endif %}
+                    {% endfor -%}
+                ){%- endif -%}
+                {%- endfor %}
+                else -> throw RuntimeException("invalid enum value, something is very wrong!!")
+            }
+        }
+    }
+
+    internal fun lower(): RustBuffer.ByValue {
+        return lowerIntoRustBuffer(this, {v, buf -> v.write(buf)})
+    }
+
+    internal fun write(buf: RustBufferBuilder) {
+        when(this) {
+            {%- for variant in e.variants() %}
+            is {{ e.name()|class_name_kt }}.{{ variant.name()|class_name_kt }} -> {
+                buf.putInt({{ loop.index }})
+                {% for field in variant.fields() -%}
+                {{ "(this.{})"|format(field.name())|write_kt("buf", field.type_()) }}
+                {% endfor -%}
+            }
+            {%- endfor %}
+        }.let { /* this makes the `when` an expression, which ensures it is exhaustive */ }
+    }
+}
+
+{% endif %}
--- a/uniffi_bindgen/src/bindings/python/templates/EnumTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/EnumTemplate.py
@@ -1,4 +1,59 @@
+{#
+# Python has a built-in `enum` module which is nice to use, but doesn't support
+# variants with associated data. So, we switch here, and generate a stdlib `enum`
+# when none of the variants have associated data, or a generic nested-class
+# construct when they do.
+#}
+{% if e.is_flat() %}
+
 class {{ e.name()|class_name_py }}(enum.Enum):
     {% for variant in e.variants() -%}
-    {{ variant|enum_name_py }} = {{ loop.index }}
+    {{ variant.name()|enum_name_py }} = {{ loop.index }}
     {% endfor %}
+
+{% else %}
+
+class {{ e.name()|class_name_py }}(object):
+    def __init__(self):
+        raise RuntimeError("{{ e.name()|class_name_py }} cannot be instantiated directly")
+
+    # Each enum variant is a nested class of the enum itself.
+    {% for variant in e.variants() -%}
+    class {{ variant.name()|enum_name_py }}(object):
+        def __init__(self,{% for field in variant.fields() %}{{ field.name()|var_name_py }}{% if loop.last %}{% else %}, {% endif %}{% endfor %}):
+            {% if variant.has_fields() %}
+            {%- for field in variant.fields() %}
+            self.{{ field.name()|var_name_py }} = {{ field.name()|var_name_py }}
+            {%- endfor %}
+            {% else %}
+            pass
+            {% endif %}
+
+        def __str__(self):
+            return "{{ e.name()|class_name_py }}.{{ variant.name()|enum_name_py }}({% for field in variant.fields() %}{{ field.name() }}={}{% if loop.last %}{% else %}, {% endif %}{% endfor %})".format({% for field in variant.fields() %}self.{{ field.name() }}{% if loop.last %}{% else %}, {% endif %}{% endfor %})
+
+        def __eq__(self, other):
+            if not other.is_{{ variant.name()|var_name_py }}():
+                return False
+            {%- for field in variant.fields() %}
+            if self.{{ field.name()|var_name_py }} != other.{{ field.name()|var_name_py }}:
+                return False
+            {%- endfor %}
+            return True
+    {% endfor %}
+
+    # For each variant, we have an `is_NAME` method for easily checking
+    # whether an instance is that variant.
+    {% for variant in e.variants() -%}
+    def is_{{ variant.name()|var_name_py }}(self):
+        return isinstance(self, {{ e.name()|class_name_py }}.{{ variant.name()|enum_name_py }})
+    {% endfor %}
+
+# Now, a little trick - we make each nested variant class be a subclass of the main
+# enum class, so that method calls and instance checks etc will work intuitively.
+# We might be able to do this a little more neatly with a metaclass, but this'll do.
+{% for variant in e.variants() -%}
+{{ e.name()|class_name_py }}.{{ variant.name()|enum_name_py }} = type("{{ e.name()|class_name_py }}.{{ variant.name()|enum_name_py }}", ({{ e.name()|class_name_py }}.{{variant.name()|enum_name_py}}, {{ e.name()|class_name_py }},), {})
+{% endfor %}
+
+{% endif %}
--- a/uniffi_bindgen/src/bindings/python/templates/RustBufferBuilder.py
+++ b/uniffi_bindgen/src/bindings/python/templates/RustBufferBuilder.py
@@ -97,7 +97,7 @@ class RustBufferBuilder(object):
     {% when Type::Boolean -%}
 
     def writeBool(self, v):
-        self._pack_into(1, ">b", 0 if v else 1)
+        self._pack_into(1, ">b", 1 if v else 0)
 
     {% when Type::String -%}
 
@@ -128,10 +128,21 @@ class RustBufferBuilder(object):
         raise InternalError("RustBufferStream.write() not implemented yet for {{ canonical_type_name }}")
 
     {% when Type::Enum with (enum_name) -%}
+    {%- let e = ci.get_enum_definition(enum_name).unwrap() -%}
     # The Enum type {{ enum_name }}.
 
     def write{{ canonical_type_name }}(self, v):
+        {%- if e.is_flat() %}
         self._pack_into(4, ">i", v.value)
+        {%- else -%}
+        {%- for variant in e.variants() %}
+        if v.is_{{ variant.name()|var_name_py }}():
+            self._pack_into(4, ">i", {{ loop.index }})
+            {%- for field in variant.fields() %}
+            self.write{{ field.type_().canonical_name()|class_name_py }}(v.{{ field.name() }})
+            {%- endfor %}
+        {%- endfor %}
+        {%- endif %}
 
     {% when Type::Record with (record_name) -%}
     {%- let rec = ci.get_record_definition(record_name).unwrap() -%}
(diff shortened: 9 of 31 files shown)


docs/manual/src/udl/enumerations.md
```diff
@@ -17,4 +18,25 @@ enum Animal {
...
+
+Can be exposed in the UDL file with:
+
+```idl
+[TaggedUnion]
+interface IpAddr {
+  V4(u8 q1, u8 q2, u8 q3, u8 q4);
+  V6(string addr);
+};
+```
```
rfk, line 40: This is the proposed UDL syntax. I think it looks nice, but it's a pretty serious abuse of WebIDL syntax.  It parses, but e.g. the `V4` line actually parses as an anonymous special method returning a datatype named `V4` rather than as a field or method named `V4`. I personally have no problem with us re-interpreting that as a tagged union, but it seems gross so YMMV. I haven't been able to come up with anything better though :-/

I had hoped to be able to do something clever with the existing WebIDL `enum` declaration, along the lines of:

```
[TaggedUnion]
enum IpAddr {
  [Fields(u8)] "V4",
  [Fields(u8, string)] "V6",
}
```

But I couldn't come up with something that would parse. In particular, attributes don't seem to be supported on the variants of an enum.

examples/rondpoint/tests/bindings/test_rondpoint.kts
```diff
@@ -6,7 +6,15 @@ assert(dico == copyDico)
...
-assert(copieCarte(mapOf("1" to Enumeration.UN, "2" to Enumeration.DEUX)) == mapOf("1" to Enumeration.UN, "2" to Enumeration.DEUX))
+assert(copieCarte(mapOf(
+    "0" to EnumerationAvecDonnees.ZERO(),
+    "1" to EnumerationAvecDonnees.UN(1),
+    "2" to EnumerationAvecDonnees.DEUX(2, "deux")
+)) == mapOf(
+    "0" to EnumerationAvecDonnees.ZERO(),
+    "1" to EnumerationAvecDonnees.UN(1),
+    "2" to EnumerationAvecDonnees.DEUX(2, "deux")
+))
```
rfk, line 17: (I haven't actually run any of these examples, just trying my best to sketch out what the resulting code would actually look like to use).

rfk:
I haven't even *begun* to think about what this would look like for the Gecko-JS bindings.

docs/manual/src/udl/enumerations.md
```diff
@@ -17,4 +18,25 @@ enum Animal {
...
+enum IpAddr {
+  V4 {q1: u8, q2: u8, q3: u8, q4: u8},
+  V6 {addr: string},
+}
+```
+
+Can be exposed in the UDL file with:
+
+```idl
+[TaggedUnion]
```
rfk, line 35: Please feel welcome to propose your favourite colour for this bikeshed. I chose "tagged union" because that's the canonical name of the Wikipedia page about this concept.

examples/rondpoint/tests/bindings/test_rondpoint.kts
jhugman: In Kotlin, since they're inner classes, they'd be spelled with CamelCase. 

They'd be defined as:

```kotlin
sealed class EnumerationAvecDonnees {
    object Zero: EnumerationAvecDonnees() // object for noargs
    data class Un(val value: Int): EnumerationAvecDonnees()
    data class Deux(val value: Int, val nombre: String): EnumerationAvecDonnees()
}

assert(EnumerationAvecDonnees.Zero != EnumerationAvecDonnees.Deux(2, "two"))
```

docs/manual/src/udl/enumerations.md
jhugman: I love the syntax you've arrived at.

This and #379 reminds me we never quite resolved the question of writing our own parser. Soon is coming the time to reconsider [ADR-0001][1]!

[1]: https://github.com/mozilla/uniffi-rs/blob/600882a0a36191ad4565bda62a22841984d777ae/docs/adr/0001-mvp-webidl.md

examples/rondpoint/tests/bindings/test_rondpoint.kts
rfk: > (in retrospect, this comment feels like "Well Actually". Sorry)

Heh, no worries at all, exactly the sort of early feedback I'm after 👍

docs/manual/src/udl/enumerations.md
rfk: > Soon is coming the time to reconsider ADR-0001!

Yes! I have many half-baked thoughts on this. What do you think is a good way to structure that kind of discussion, should we e.g. open an open-ended "discuss" issue where we can kick some ideas around?

README.md
```diff
@@ -88,7 +88,7 @@ Things that are implemented so far:
 
 * Primitive numeric types, equivalents to those offered by Rust (`u32`, `f64`, etc).
 * Strings (which are always UTF-8, like Rust's `String`).
-* C-style enums (just the discriminant, no associated data).
+* Enums, including enums associated data (aka "Tagged Unions").
```
mhammond, line 91: nit: add "with"

docs/manual/src/internals/lifting_and_lowering.md
```diff
@@ -85,6 +86,7 @@ The details of this format are internal only and may change between versions of
 | `sequence<T>` | Serialized `i32` item count followed by serialized items; each item is a serialized `T` |
 | `record<string, T>` | Serialized `i32` item count followed by serialized items; each item is a serialized `string` followed by a serialized `T` |
 | `enum` | Serialized `u32` indicating variant, numbered in declaration order starting from 1 |
+| `[TaggedUnion] interface` | Serialized `u32` indicating variant as per `enum`, followed by the serialized value of each field, in declaration order |
```
mhammond, line 89: I don't think "union" and "enum" are simple synonyms in rust - is `TaggedEnum` a better name?

docs/manual/src/udl/enumerations.md
mhammond: heh - see above - I mildly prefer "tagged enum" given we are in rust.

docs/manual/src/udl/enumerations.md
mhammond: +1 to what James said, although I'm slightly more skeptical we will find time to rewrite the parser until that's our final option :)

docs/manual/src/udl/enumerations.md
```diff
@@ -17,4 +18,25 @@ enum Animal {
...
+interface IpAddr {
+  V4(u8 q1, u8 q2, u8 q3, u8 q4);
+  V6(string addr);
+};
+```
+
+Only enums with named fields are supported by this syntax.
```
mhammond, line 42: (to reinforce the naming thing - the docs keep calling them "enums" :)

examples/rondpoint/src/rondpoint.udl
```diff
@@ -2,7 +2,7 @@ namespace rondpoint {
   Dictionnaire copie_dictionnaire(Dictionnaire d);
   Enumeration copie_enumeration(Enumeration e);
   sequence<Enumeration> copie_enumerations(sequence<Enumeration> e);
-  record<DOMString, Enumeration> copie_carte(record<DOMString, Enumeration> c);
+  record<DOMString, EnumerationAvecDonnees> copie_carte(record<DOMString, EnumerationAvecDonnees> c);
```
mhammond, line 5: A big +1 for the french :)

(On a completely less serious note, I wonder if the French in rondpoint is a bigger risk than the parser in the shorter term ;)

mhammond (commented):
This is awesome. I skipped over the parser code, but will try and get back to it over the next few days, but thought I'd share my hearty +1 in the meantime.

docs/manual/src/internals/lifting_and_lowering.md
rfk: Yeah, that's a good point re: the likelihood the name here will be interpreted somewhere close to Rust's semantics.

examples/rondpoint/src/rondpoint.udl
rfk: On the other hand, trying to switch this away from French would be like going through the Sync codebase on the removing all the puns...I don't think I'm ready for that just yet 😢 😅

rfk:
A quick update here - there are plenty of rough bits remaining, but the modified Kotlin test seems to be running and working as intended!

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -1,19 +1,94 @@
-enum class {{ e.name()|class_name_kt }} {
-    {% for variant in e.variants() %}
-    {{ variant|enum_variant_kt }}{% if loop.last %};{% else %},{% endif %}
+{#
+// Kotlin's `enum class` constuct doesn't support variants with associated data,
+// but is a little nicer for consumers than its `sealed class` enum pattern.
+// So, we switch here, using `enum class` for enums with no associated data
+// and `sealed class` for the general case.
```
rfk, line 5: @jhugman here's a design question I'd appreciate your insight on. Kotlin's `enum class` declaration doesn't support associated data in the way we want to use it, so in the general case we have to use the `sealed class` approach. However, when there isn't any associated data, Kotlin's `enum` seems to be a little nicer to work with. Here I've opted to switch between them - enums without data will be an `enum class` just as they were under the previous implementation, while newly-supported enums with data will be a `sealed class`.

This has the nice property of being backwards-compatible for existing consumers of existing UDL files. But I'm a little concerned about potential developer confusion in future.

For example, suppose someone currently has a plain-old-enum with no data in their UDL file, and their Kotlin consumers are consuming it through an `enum class`. The component API evolves and adds a new case to the enum, this time with some associated data. For Kotlin consumers this addition will *change the entire type* of the thing they're dealing with, including changing the capitalization of existing items as it switches from being an `enum class` to a `sealed class`.

Maybe this is OK, since consumers will need to update their code to handle the new variant anyway. Or maybe it's leaking implementation details in a way that we shouldn't need to do.

Thoughts?

jhugman: It's a great question. 

From our (nimbus + uniffi) point of view, I don't think it matters: we have a tiny amount of code using UDL enums, so changing them will have some coordination cost when changing them to a new format.

For android devs: adding another `enum` variant is going to be a breaking change.

There is some JVM cost to using sealed classes over Java style (around class initialization, maximum method counts etc). I don't think we're at the stage we need to worry about that.

While thinking about this, it did strike me that _Kotlin_ has had to make a similar choice: Java enums existed after  Java5. Kotlin came up with a better way, yet still supported old school Java enums. I should imagine their backwards compatibility problem was more persuasive/pressing than ours.

jhugman: Other things this question made me think about:

 * [Typescript can do Tagged Unions][1], and whatever Javascript can support natively.
 * [Javascript has barely has the simple enums][0].

Until we have a idiomatically _consumable_ version of tagged unions in JS, we're going to be discouraging use of this feature.



[0]: https://www.andrewnater.com/posts/enums-in-javascript/
[1]: https://gregoryppabian.medium.com/creating-tagged-unions-in-typescript-ce727a306878

docs/manual/src/udl/enumerations.md
jhugman: I agree with Mark here: I expect we'll only seriously consider doing the work once we've exhausted.

That'd be the wrong time for a discussion, so yes, let's start a doc/issue and iterate.

rfk: > That'd be the wrong time for a discussion, so yes, let's start a doc/issue and iterate.

Sounds good 👍 

(FWIW, I was pretty close to declaring "WebIDL has no good syntax for this" before I discovered the anonymous-special-method-returning-a-named-datatype hackery you see here, so I won't be surprised if we run out of runway with WebIDL sooner than we expect)

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
rfk: > Until we have a idiomatically consumable version of tagged unions in JS, we're going to be discouraging use of this feature.

Yeah. Python has a similar problem, which is one of the reasons I like having the python backend here, to think about how things will look in a language with basically the opposite of a Rust-style strong algebraic type system.

docs/manual/src/internals/lifting_and_lowering.md
jhugman: I spent a good deal of time thinking about this yesterday, and still couldn't remember `TaggedUnion`. Having said that: what a lovely bikeshed!

docs/manual/src/udl/enumerations.md
dmose: @rfk So will any WebIDL we generate (not just the internal UDL) for Firefox have a way of expressing this stuff?

examples/rondpoint/src/rondpoint.udl
dmose: How about writing newly added code in English, and changing bits of codes that we touch incrementally, as it makes sense?

dmose:
As you mentioned above, the GeckoJS stuff is going to be the higher risk.  The thing that jumps out at me is wondering whether the stuff you've done for UDL is even going to be expressable in a way that the Firefox WebIDL parser will understand...

docs/manual/src/udl/enumerations.md
rfk: I filed a "discuss" issue here: https://github.com/mozilla/uniffi-rs/issues/385

docs/manual/src/internals/lifting_and_lowering.md
rfk: I'm going to try just `[Enum]` and see how I like it...

rfk:
I've got this working in a way that I'm pretty happy with for the Kotlin, Swift and Python backends. I haven't tackled the gecko_js backend yet, and similar to my comments [here](https://github.com/mozilla/uniffi-rs/pull/384#discussion_r575907962), I think we should have a little sync-up about how we'll approach that backend strategically in the short-term. I'm happy to do the work! I just don't want to get stuck down a rabbit-hole if it's not strategically important right now.

(Unlike #384, I don't really have a good idea for how to expose these through C++ and into special Gecko JS. We could try to do something similar to what the python backend is currently doing here, basically creating a separate class for each variant and packaging them up into a convenient API...but I can easily imagine that not playing nicely with all the C++ codegen and what-not).

rfk:
A note to self from trying this out in the fxa-client component: we'll need to be careful with name conflicts at different scopes.

The fxa-client declared a set of enums like this:

```
[Enum]
interface AccountEvent {
  IncomingDeviceCommand(IncomingDeviceCommand command );
  ProfileUpdated();
  AccountAuthStateChanged();
  AccountDestroyed();
  DeviceConnected(string device_name );
  DeviceDisconnected(string device_id, boolean is_local_device );
};

[Enum]
interface IncomingDeviceCommand {
  TabReceived(Device? sender, SendTabPayload payload );
};
```

Note that the name `IncomingDeviceCommand` is both a top-level enum, and one of the variants on the `AccountEvent` enum. When I generated the Kotlin code for this, we wound up with:

```
sealed class IncomingDeviceCommand {
    class TabReceived(
        val sender: Device?,
        val payload: SendTabPayload
    ) : IncomingDeviceCommand() { }

    companion object {
        internal fun read(buf: ByteBuffer): IncomingDeviceCommand {
            // details omitted
        }
    }
}

sealed class AccountEvent {
    class IncomingDeviceCommand(
        val command: IncomingDeviceCommand
    ) : AccountEvent() { ... }

    // Other cases omitted

    companion object {
        internal fun read(buf: ByteBuffer): AccountEvent {
            return when (buf.getInt()) {
                1 -> AccountEvent.IncomingDeviceCommand(
                    //
                    //  +---- this line gives a compile error
                    //  |
                    //  v
                    IncomingDeviceCommand.read(buf)
                )
                // other cases omitted
                else -> throw RuntimeException("invalid enum value, something is very wrong!!")
            }
        }
    }
}
```

This doesn't compile, because Kotlin thinks that `IncomingDeviceCommand.read(buf)` is referring to the *inner* `IncomingDeviceCommand` class defined on the `AccountEvent` enum, rather than the top-level `IncomingDeviceCommand` class.

I've worked around this for now by avoiding the name conflict, but we should figure out a better approach before landing this PR. Maybe there's a way to do the codegen so that the names are not ambiguous, or maybe we need to just check for duplicate names and error in that case.

At the very least, we need a testcase for this.

rfk:
Alrighty, I've rebased this atop latest main, and added an early bail out if we see an enum variant name that conflicts with a top-level type name.

uniffi_bindgen/src/bindings/gecko_js/templates/SharedHeaderTemplate.h
```diff
@@ -29,13 +29,17 @@ namespace dom {
...
 
 {% for e in ci.iter_enum_definitions() %}
+{% if e.has_associated_data() %}
+MOZ_STATIC_ASSERT(false, "Sorry the gecko-js backend does not yet support enums with associated data");
+{% else %}
 template <>
-struct ViaFfi<{{ e.name()|class_name_cpp(context) }}, uint32_t> {
-  [[nodiscard]] static bool Lift(const uint32_t& aLowered, {{ e.name()|class_name_cpp(context) }}& aLifted) {
-    switch (aLowered) {
+struct Serializable<{{ e.name()|class_name_cpp(context) }}> {
```
rfk, line 36: This changes the way enums are passed over the FFI, even enums without any associated data. I've done my best to update this and will see if I can try it out in mozilla-central following the instructions in [this doc](https://docs.google.com/document/d/1JnSruKgfKDcfIuiMirEIhum4NOWYlo58sCXPoxtC21g/) but I don't have high confidence in my skills in this area.

rfk: I was able to integrate this into m-c and have it compile, then fail on a linker error. So I guess that's a good sign.

uniffi_bindgen/src/bindings/python/templates/EnumTemplate.py
```diff
@@ -1,4 +1,59 @@
...
+    # For each variant, we have an `is_NAME` method for easily checking
+    # whether an instance is that variant.
+    {% for variant in e.variants() -%}
+    def is_{{ variant.name()|var_name_py }}(self):
+        return isinstance(self, {{ e.name()|class_name_py }}.{{ variant.name()|enum_name_py }})
+    {% endfor %}
+
+# Now, a little trick - we make each nested variant class by a subclass of the main
+# enum subclass, so that method calls and instance checks etc will work intuitively.
+# We might be able to do this a little more nearly with a metaclass, but this'll do.
```
mhammond, line 47: nit: neatly?

uniffi_bindgen/src/bindings/python/templates/RustBufferBuilder.py
```diff
@@ -97,7 +97,7 @@ def writeF64(self, v):
     {% when Type::Boolean -%}
 
     def writeBool(self, v):
-        self._pack_into(1, ">b", 0 if v else 1)
+        self._pack_into(1, ">b", 1 if v else 0)
```
mhammond, line 100: this is an unrelated bugfix, right?

mhammond (commented):
This looks great with the usual attention to detail we've come to expect from your PRs :) r+ from me, but I'll let James give formal approval once he has cast his eye again over the generated swift/kotlin

uniffi_bindgen/src/bindings/python/templates/RustBufferBuilder.py
rfk: Oh, yes, yes it is - but one that I only happened to discover in the process of getting all this working.

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -1,19 +1,94 @@
...
+            }
+        }
+    }
+
+    internal fun lower(): RustBuffer.ByValue {
+        return lowerIntoRustBuffer(this, {v, buf -> v.write(buf)})
+    }
+
+    internal open fun write(buf: RustBufferBuilder) {
+        throw RuntimeException("enum variant should have overridden `write` method, something is very wrong!!")
```
jhugman, line 62: On reflection, I think I'd rather put the `write` method and friends in the top level, and use pattern matching to work out what to do. 

(FWIW I don't think there's any perf difference, in fact: given it's thee idiomatic way of doing tagged-unions, I expect there's some particular optimizations we could end up losing out on by adding methods to individual variant classes.)

jhugman (commented):
I'm pretty happy with how this is coming along; I have a few questions, but this is looking really close.

uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
```diff
@@ -1,30 +1,40 @@
-public enum {{ e.name()|class_name_swift }}: ViaFfi {
+public enum {{ e.name()|class_name_swift }}: ViaFfiUsingByteBuffer, ViaFfi, Equatable {
     {% for variant in e.variants() %}
-    case {{ variant|enum_variant_swift }}
+    case {{ variant.name()|enum_variant_swift }}{% if variant.fields().len() > 0 %}(
+        {% for field in variant.fields() %}
+        {{ field.name()|var_name_swift }}: {{ field.type_()|type_swift }}{%- if loop.last -%}{%- else -%},{%- endif -%}
+        {% endfor %}
```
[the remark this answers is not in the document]
rfk: It's copy-pasted from the logic for declaring record fields. I'm going to extract them into a `swift::field_list_decl` macro for convenience.

uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
```diff
@@ -1,30 +1,40 @@
-public enum {{ e.name()|class_name_swift }}: ViaFfi {
+public enum {{ e.name()|class_name_swift }}: ViaFfiUsingByteBuffer, ViaFfi, Equatable {
```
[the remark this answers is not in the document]
rfk: Sounds good, I filed https://github.com/mozilla/uniffi-rs/issues/396 for future discussions.

uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -31,34 +31,84 @@
...
         &self.name
     }
-    pub fn variants(&self) -> Vec<&str> {
-        self.variants.iter().map(|v| v.as_str()).collect()
+    pub fn variants(&self) -> Vec<&Variant> {
+        self.variants.iter().collect()
+    }
+
+    pub fn has_associated_data(&self) -> bool {
+        self.variants.iter().any(Variant::has_fields)
```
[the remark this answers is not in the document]
rfk: Yeah, I think you're right. I'm not super happy about it because it feels like an implementation detail of the (what we hope to be temporary) IDL leaking all the way through to the generated bindings, but that's probably less bad than having a confusing user experience.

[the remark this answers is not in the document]
rfk: I'm updating this to mark enums that come from `enum foo` as "flat" and enums that come from `[Enum] interface` as "not flat", and will use that in codegen to switch between the two implementations.

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -1,19 +1,94 @@
...
+// but is a little nicer for consumers than its `sealed class` enum pattern.
+// So, we switch here, using `enum class` for enums with no associated data
+// and `sealed class` for the general case.
+#}
+
+{% if e.has_associated_data() %}
+
+sealed class {{ e.name()|class_name_kt }} {
+    {% for variant in e.variants() -%}
+    class {{ variant.name()|class_name_kt }}({% if variant.has_fields() %}
```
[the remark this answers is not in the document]
rfk: FWIW, I mildly dislike the asymmetry this produces between `EnumerationAvecDonnees.Zero` without parens and `EnumerationAvecDonnees.Un(1u)` with parens, but am happy to defer to your judgement here.

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -1,19 +1,94 @@
...
+                {% if variant.has_fields() -%}
+                {% for field in variant.fields() -%}
+                {{ field.name()|var_name_kt }} == other.{{ field.name()|var_name_kt }}{% if loop.last %}{% else %} && {% endif -%}
+                {% endfor -%}
+                {% else -%}
+                true
+                {%- endif %}
+            } else {
+                false
+            }
```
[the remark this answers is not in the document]
rfk: 🤯 

The galaxy-brain approach seems to work a treat, and I'll add a couple more testcases around equality to ensure that. Very nice!

rfk:
Updated in response to @jhugman latest feedback.

examples/rondpoint/tests/bindings/test_rondpoint.kts
```diff
@@ -6,7 +6,24 @@ assert(dico == copyDico)
...
+    "2" to EnumerationAvecDonnees.Deux(2u, "deux")
+)) == mapOf(
+    "0" to EnumerationAvecDonnees.Zero,
+    "1" to EnumerationAvecDonnees.Un(1u),
+    "2" to EnumerationAvecDonnees.Deux(2u, "deux")
+))
+
+val var1: EnumerationAvecDonnees = EnumerationAvecDonnees.Zero
+val var2: EnumerationAvecDonnees = EnumerationAvecDonnees.Un(1u)
+val var3: EnumerationAvecDonnees = EnumerationAvecDonnees.Un(2u)
```
rfk, line 21: The typed variables here are to outsmart the Kotlin compiler, which will give a compile-time error if you try to directly compare something of type `EnumerationAvecDonnees.Zero` and something of type `EnumerationAvecDonnees.Un`.

rfk:
Hrm, the `rondpoint` crate is failing to build in CI here, with what appear to be out-of-memory errors:

```
Caused by:
  process didn't exit successfully: `rustc --crate-name uniffi_rondpoint --edition=2018 examples/rondpoint/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type cdylib --crate-type lib --emit=dep-info,link -C debuginfo=2 -C metadata=053719643b56441c --out-dir /home/circleci/project/target/debug/deps -C incremental=/home/circleci/project/target/debug/incremental -L dependency=/home/circleci/project/target/debug/deps --extern uniffi=/home/circleci/project/target/debug/deps/libuniffi-622b493c8a6725e6.rlib --extern uniffi_macros=/home/circleci/project/target/debug/deps/libuniffi_macros-b7aff2897c4d9de6.so -D warnings` (signal: 9, SIGKILL: kill)
warning: build failed, waiting for other jobs to finish...
error: linking with `cc` failed: exit code: 1
  |
  = note: "cc" "-Wl,--as-needed" "-Wl,-z,noexecstack" "-m64" "-L" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib" "/home/circleci/project/target/debug/deps/arithmetical.128vv0nl8ohl5duk.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.17qde80626cmwono.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.190shw0nb0jwrvnq.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.1efkqkyw2vfvq0cq.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.1h549mnvlhu6kx62.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.1trgsh1jgtu2dzg4.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.201cn5r1kdf42vwp.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.23xtc1yr9w6voutq.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.249yaojpywzbmc58.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.26tnl9p6edsmo1vn.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.2b8zs3xp8yccjv9n.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.2dfznzhzko3hfucn.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.2o3y1jvfz588xr4j.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.2whez0kgj4a6yh6v.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.31ziggzrgykvtvvh.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.3fo6cwgekbm8eipp.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.3smn1wl22kf53084.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.42sg7z6pu41rx0oj.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.4blho4emn5w3yt1w.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.4ssytj083kstompf.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.544wkcpgqb1z1i1a.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.56t3ybnmrjh0xjzu.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.5fgy3snlmii7kl8g.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.dsblm0i1ak2u4v8.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.nqushftut1gb8rg.rcgu.o" "/home/circleci/project/target/debug/deps/arithmetical.nwud4vssfr9v9r5.rcgu.o" "-o" "/home/circleci/project/target/debug/deps/libarithmetical.so" "-Wl,--version-script=/tmp/rustcImMpM0/list" "/home/circleci/project/target/debug/deps/arithmetical.hkx95sgdce0u0cv.rcgu.o" "-Wl,--gc-sections" "-Wl,-zrelro" "-Wl,-znow" "-nodefaultlibs" "-L" "/home/circleci/project/target/debug/deps" "-L" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-Wl,-Bstatic" "/home/circleci/project/target/debug/deps/libuniffi-622b493c8a6725e6.rlib" "/home/circleci/project/target/debug/deps/libuniffi_bindgen-16e505356ba326f5.rlib" "/home/circleci/project/target/debug/deps/libweedle-aef167c1aa3dbb32.rlib" "/home/circleci/project/target/debug/deps/libnom-0fee0736e68b1069.rlib" "/home/circleci/project/target/debug/deps/libheck-e9f9f2619703ceaa.rlib" "/home/circleci/project/target/debug/deps/libunicode_segmentation-c8f494fc70b1117f.rlib" "/home/circleci/project/target/debug/deps/libaskama-537ffd130493464b.rlib" "/home/circleci/project/target/debug/deps/libaskama_shared-9451c3a778333d62.rlib" "/home/circleci/project/target/debug/deps/libtoml-31756e16259888ea.rlib" "/home/circleci/project/target/debug/deps/libsyn-1cb4f73bae3a226d.rlib" "/home/circleci/project/target/debug/deps/libnom-9b9d4b929f8fedf4.rlib" "/home/circleci/project/target/debug/deps/libmemchr-b63c81c95b3ea372.rlib" "/home/circleci/project/target/debug/deps/liblexical_core-5908b33ac15872fb.rlib" "/home/circleci/project/target/debug/deps/libarrayvec-fc27a13ca7389030.rlib" "/home/circleci/project/target/debug/deps/libbitflags-599c3a9be4aadd44.rlib" "/home/circleci/project/target/debug/deps/libbitvec-c3fcb913cff8ff19.rlib" "/home/circleci/project/target/debug/deps/libfunty-19a720bfa18dce89.rlib" "/home/circleci/project/target/debug/deps/libwyz-4577ea1179b041c6.rlib" "/home/circleci/project/target/debug/deps/libtap-fd238052b670c4b5.rlib" "/home/circleci/project/target/debug/deps/libradium-0a3a3c08ef28929c.rlib" "/home/circleci/project/target/debug/deps/libquote-7e1852a0cea0cdb2.rlib" "/home/circleci/project/target/debug/deps/libproc_macro2-815d44389fbca503.rlib" "/home/circleci/project/target/debug/deps/libunicode_xid-5bbc8b1bac57b67c.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libproc_macro-9fd29953164d86f0.rlib" "/home/circleci/project/target/debug/deps/libaskama_escape-ed1e0cc3b58d2466.rlib" "/home/circleci/project/target/debug/deps/libstatic_assertions-7b24323df9107fb0.rlib" "/home/circleci/project/target/debug/deps/libcargo_metadata-e1494bb5b6f651dd.rlib" "/home/circleci/project/target/debug/deps/libserde_json-e2226d212e2ed73f.rlib" "/home/circleci/project/target/debug/deps/libryu-65130c6d4b3381b5.rlib" "/home/circleci/project/target/debug/deps/libitoa-eca5c176e5c68ac7.rlib" "/home/circleci/project/target/debug/deps/libsemver-7b8295dac8c0b8df.rlib" "/home/circleci/project/target/debug/deps/libserde-81a5e1ceab613554.rlib" "/home/circleci/project/target/debug/deps/libsemver_parser-906f2c60081c06fb.rlib" "/home/circleci/project/target/debug/deps/libffi_support-0e7f42b523b4e0fd.rlib" "/home/circleci/project/target/debug/deps/liblazy_static-5390404b369a64d3.rlib" "/home/circleci/project/target/debug/deps/liblog-f8ad7d725fc2e075.rlib" "/home/circleci/project/target/debug/deps/libcfg_if-bba59873400967df.rlib" "/home/circleci/project/target/debug/deps/libbytes-040a8c7b9588319c.rlib" "/home/circleci/project/target/debug/deps/libanyhow-34a3b06fd061f9d1.rlib" "/home/circleci/project/target/debug/deps/libthiserror-bae65f7bcdf0be60.rlib" "-Wl,--start-group" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-7c5e456310a1373c.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libpanic_unwind-b981d9b2a408308f.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libhashbrown-43d0ea1b5ae34d0d.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/librustc_std_workspace_alloc-09e7f22e773899cd.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libbacktrace-aa74f166651adf6e.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libbacktrace_sys-22c386707b639611.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/librustc_demangle-db04c9c5cd3bcf45.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libunwind-bb27492f721492e8.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcfg_if-6b95245dbf686e20.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/liblibc-287409d75db2ecd3.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/liballoc-a93f70ee2006b6e3.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/librustc_std_workspace_core-566cdfbcc94b4360.rlib" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcore-7bb8dddc7ce34e92.rlib" "-Wl,--end-group" "/usr/local/rustup/toolchains/1.43.0-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-b117658e17259aa6.rlib" "-Wl,-Bdynamic" "-ldl" "-lrt" "-lpthread" "-lgcc_s" "-lc" "-lm" "-lrt" "-lpthread" "-lutil" "-lutil" "-shared"
  = note: /usr/bin/ld: final link failed: Cannot allocate memory
          collect2: error: ld returned 1 exit status
```

I'm kind of hoping it's a transient infrastructure issue, because I haven't made any changes to the Rust code there since the previous iteration of this PR, which compiled successfully.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -451,6 +454,29 @@ impl<'ci> ComponentInterface {
...
+        for e in self.enums.iter() {
+            for variant in e.variants.iter() {
+                if self.types.get_type_definition(variant.name()).is_some() {
+                    bail!(
+                        "Enum variant names must not shadow type names: \"{}\"",
+                        variant.name()
+                    )
+                }
+            }
+        }
```
jhugman, line 476: Ok.

jhugman (approved):
This is great.

I'm not qualified to comment about Python, or gecko-js, but Swift and Kotlin get my thumbs up.

uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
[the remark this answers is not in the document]
rfk: That makes sense. I think in practice they will probably get an error before they even try to generate swift code, because the Rust code will need to use some `Box` indirection and the generated scaffolding won't support it. But I can add a link to the issue in a few key places in the hope that any such users will find it.

PR #382, opened 2021-02-14
Hide Rust scaffolding from `cargo doc`.

rfk:
I've been experimenting with using `cargo doc` as the canonical source
of documentation for a UniFFI component, and I quite like it so far,
except for the way that it currently documents all the autogenerated
helper functions and traint impls. So, let's hide those using the
nice `#[doc(hidden)]` annotation.

### DIFF
--- a/uniffi_bindgen/Cargo.toml
+++ b/uniffi_bindgen/Cargo.toml
@@ -23,3 +23,11 @@ heck = "0.3"
 clap = { version = "2", default-features = false }
 serde = "1"
 toml = "0.5"
+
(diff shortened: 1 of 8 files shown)


rfk:
Humph. CI is failing here due to some conflicts between dependencies and a recently-stabilized new builtin field for numberic types. IIUC it's currently blocked on a new release of `nom` (https://github.com/Geal/nom/pull/1286) which gets into our dependency tree via `askama`. Fortunately `askama` doesn't seem to pin its version of `nom` so that *should* be enough to get us back up and running.

mhammond (approved):

PR #386, opened 2021-02-15
A little experiment with macros instead of build scripts.

rfk:
This commit is basically me messing around with procedural macros,
to try them out as a potential future iteration of the UniFFI
developer experience.

As a crate author, I think it would be amazing not to have to
worry about .udl files and built scripts and what-not, but
instead be able to declare my API with a macro and have Rust
just take care of the details. This commit is a tiny experiment
in that direction, based on existing features of UniFFI.

The idea is that I can wrap my Rust code in a little macro
like this:

```
[uniffi_macros::declare_interface]
pub mod my_component {
  ...my rust code goes here...
}
```

And I can provide a corresponding `my_component.udl` file that
declares the interface, and all the Rust scaffolding code just
gets taken care of automagically.

I still have to provide a `.udl` file, but I hope you can imagine
one day the possibility of generating the interface definition
directly from the Rust code that is decorated by the macro.
Not any time soon! But one day...

Curious to hear any early impressions on these shenanigans :-)

### DIFF
--- a/examples/geometry/Cargo.toml
+++ b/examples/geometry/Cargo.toml
@@ -11,8 +11,5 @@ crate-type = ["cdylib"]
 name = "uniffi_geometry"
 
 [dependencies]
-uniffi_macros = {path = "../../uniffi_macros"}
+uniffi_macros = {path = "../../uniffi_macros", features=["builtin-bindgen"]}
 uniffi = {path = "../../uniffi", features=["builtin-bindgen"]}
-
-[build-dependencies]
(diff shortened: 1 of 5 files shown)


rfk:
(I converted this to draft because we really don't need to be merging it anytime soon; the convenience macro to include generated scaffolding though I think is quite nice and have pushed separately at https://github.com/mozilla/uniffi-rs/pull/397)

rfk:
I don't intend for this to actually be merged, so I'm going to close it to avoid confusion. (I hope to find a bit of downtime to open an alternative macro-based PR for further discussion)

PR #387, opened 2021-02-17
Add support for named alternate constructors.

### LINKED ISSUE #37: Static and/or multiple constructors support

rfk:
This is an alternative approach to factory methods, based on feedback from #384. It also fixes #37.

It's a fairly common pattern for object-oriented interfaces to have
static methods that act as named "factories" for producing object
instances in ways that differ from the default constructor.
This commit adds basic support for such named alternate constructors:

* In the UDL, declare additional constructors with a [Name=Value]
  attribute to tell the tool what name should be exposed under.
* In the Rust code, implement a corresponding method on the struct
  that acts as a constructor (no `self` parameter, `Self` return type).
* In the foreign-language code, expect a static method on the
  resulting class with name matching the one provided in Rust.

We continue to reserve the name "new" to refer to the default constructor.
A followup could consider annotating it with an additional attribute
like [DefaultConstructor] in order to let the default constructor have
a different name on the Rust side.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
@@ -12,10 +12,12 @@ class {{ obj.name()|class_name_kt }}(
     handle: Long
 ) : FFIObject(AtomicLong(handle)), {{ obj.name()|class_name_kt }}Interface {
 
-    {%- for cons in obj.constructors() %}
+    {%- match obj.primary_constructor() %}
+    {%- when Some with (cons) %}
     constructor({% call kt::arg_list_decl(cons) -%}) :
         this({% call kt::to_ffi_call(cons) %})
-    {%- endfor %}
+    {%- when None %}
+    {%- endmatch %}
 
     /**
      * Disconnect the object from the underlying Rust object.
@@ -48,12 +50,21 @@ class {{ obj.name()|class_name_kt }}(
         }.let {
             {{ "it"|lift_kt(return_type) }}
         }
-    
+
     {%- when None -%}
     override fun {{ meth.name()|fn_name_kt }}({% call kt::arg_list_protocol(meth) %}) =
         callWithHandle {
-            {%- call kt::to_ffi_call_with_prefix("it", meth) %} 
+            {%- call kt::to_ffi_call_with_prefix("it", meth) %}
         }
     {% endmatch %}
     {% endfor %}
+
+    {% if obj.constructors().len() > 1 -%}
+    companion object {
+        {% for cons in obj.alternate_constructors() -%}
+        fun {{ cons.name()|fn_name_kt }}({% call kt::arg_list_decl(cons) %}): {{ obj.name()|class_name_kt }} =
+            {{ obj.name()|class_name_kt }}({% call kt::to_ffi_call(cons) %})
+        {% endfor %}
+    }
+    {%- endif %}
 }
--- a/uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
@@ -9,14 +9,20 @@ public protocol {{ obj.name() }}Protocol {
     {% endfor %}
 }
 
-public class {{ obj.name() }}: {{ obj.name() }}Protocol {
+public class {{ obj.name()|class_name_swift }}: {{ obj.name() }}Protocol {
     private let handle: UInt64
 
-    {%- for cons in obj.constructors() %}
-    public init({% call swift::arg_list_decl(cons) -%}) {% call swift::throws(cons) %} {
-        self.handle = {% call swift::to_ffi_call(cons) %}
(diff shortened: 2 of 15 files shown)


uniffi_bindgen/src/interface/object.rs
```diff
@@ -347,7 +365,13 @@ impl APIConverter<Method> for weedle::interface::OperationInterfaceMember<'_> {
...
             name: match self.identifier {
                 None => bail!("anonymous methods are not supported {:?}", self),
-                Some(id) => id.0.to_string(),
+                Some(id) => {
+                    let name = id.0.to_string();
+                    if name == "new" {
+                        bail!("the method name \"new\" is reserved for the default constructor");
+                    }
+                    name
+                }
```
jhugman, line 374: Nice.

uniffi_bindgen/src/interface/object.rs
```diff
@@ -413,4 +437,82 @@ mod test {
...
+        let cons = obj.alternate_constructors()[0];
+        assert_eq!(cons.name(), "newish");
+        assert_eq!(cons.arguments.len(), 0);
+        assert_eq!(cons.ffi_func.arguments.len(), 0);
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_the_name_new_is_reserved_for_constructors() {
```
jhugman, line 504: Whilst we're crossing the 👀 and dotting the 🍵 testing-wise, what happens if we have two constructors called `new`?

```
            namespace test{};
            interface Testing {
                constructor();
                [Name=new]
                constructor();
            };
```

uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -48,12 +50,22 @@ class {{ obj.name()|class_name_kt }}(
...
+
     {%- when None -%}
     override fun {{ meth.name()|fn_name_kt }}({% call kt::arg_list_protocol(meth) %}) =
         callWithHandle {
-            {%- call kt::to_ffi_call_with_prefix("it", meth) %} 
+            {%- call kt::to_ffi_call_with_prefix("it", meth) %}
         }
     {% endmatch %}
     {% endfor %}
+
```
jhugman, line 61: Nit: whitespace.

jhugman (approved):
I'm not really qualified to comment on the WebIDL or python, but the Rust, Kotlin and Swift parts are looking very nice. Thank you.

There are a few nits, but I won't hold up proceedings to get this moving. r+

uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
```diff
@@ -9,22 +9,38 @@ public protocol {{ obj.name() }}Protocol {
...
     deinit {
         try! rustCall(InternalError.unknown()) { err in
             {{ obj.ffi_object_free().name() }}(handle, err)
         }
     }
 
-    // TODO: Maybe merge the two templates (i.e the one with a return type and the one without)
+    static func lift(_ handle: UInt64) throws -> {{ obj.name()|class_name_swift }} {
+        {{ obj.name()|class_name_swift }}(fromRawHandle: handle)
+    }
```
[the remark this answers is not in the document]
rfk: Ah, good catch, I think this is leftover from the previous factoring.

uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
```diff
@@ -9,22 +9,38 @@ public protocol {{ obj.name() }}Protocol {
...
-    {%- for cons in obj.constructors() %}
+    private init(fromRawHandle handle: UInt64) {
+        self.handle = handle
+    }
+
+    {%- match obj.primary_constructor() %}
+    {%- when Some with (cons) %}
     public init({% call swift::arg_list_decl(cons) -%}) {% call swift::throws(cons) %} {
         self.handle = {% call swift::to_ffi_call(cons) %}
     }
```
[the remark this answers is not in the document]
rfk: Nice, thanks. Also, my word Swift is a conceptually-dense language, TIL about "conveience initializers" vs "designated initializers"...

uniffi_bindgen/src/interface/object.rs
rfk: OK, I'll fess up...I actually started writing this exact test and then deleted it when I realized that it was likely to behave in a dumb manner (allowing duplicate names in the `ComponentInterface` and then giving inscrutable compile errors later on in the generated code). I'll take a look at making this error out more cleanly.

rfk: (Surprise! Making it error out on duplicate names was actually quite straightforward and I should have just done this work the first time around 😅. Updated.)

mhammond (approved):
This is much clearer!

PR #388, opened 2021-02-18
heuristically use the cdylib produced by `pkg_dir` if there are multiple cdylibs

king6cong:


rfk:
Interesting, thanks @king6cong. Could you please say a bit more about the setup you were using the tool with when you encountered this issue, to help me make sure I understand the motivation here?

king6cong:
@rfk Sorry for not being clear 😄 

It seems like when we have a dependency which contains `cdylib` as its `crate-type`, this dependency will end up in our cdylib lists, which we don't want to to use as our ffi cdylib

Steps to produce:
```bash
git clone https://github.com/bevyengine/bevy
cd bevy/crates/bevy_log
cargo build --message-format=json --lib --target wasm32-unknown-unknown | rg -v Compiling | rg cdylib
```

Output:
```
{"reason":"compiler-artifact","package_id":"tracing-wasm 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)","target":{"kind":["cdylib","rlib"],"crate_types":["cdylib","rlib"],"name":"tracing-wasm","src_path":"/Users/cong/.cargo/registry/src/github.com-1ecc6299db9ec823/tracing-wasm-0.1.0/src/lib.rs","edition":"2018","doc":true,"doctest":true,"test":true},"profile":{"opt_level":"0","debuginfo":2,"debug_assertions":true,"overflow_checks":true,"test":false},"features":[],"filenames":["/Users/cong/repos/bevy/target/wasm32-unknown-unknown/debug/deps/tracing_wasm-a4ecb6852b53e390.wasm","/Users/cong/repos/bevy/target/wasm32-unknown-unknown/debug/deps/libtracing_wasm-a4ecb6852b53e390.rlib"],"executable":null,"fresh":true}
```

rfk:
Aha, that makes sense, thanks! I think it would be good to also add a small [regression-test case](https://github.com/mozilla/uniffi-rs/tree/main/fixtures/regressions/) to ensure we don't break the fix in any future refactorings. If you're willing to have a go at adding one I'd be happy to review, otherwise I'll try to find some time to take a deeper look in the next few days.

king6cong:
@rfk Regression test added

rfk (approved):
Thanks for the test. This looks solid and I confirmed the new test passes locally 👍

PR #389, opened 2021-02-18
Add config option to set the cdylib name for backends that use it.

rfk:
When building uniffi components in the application-services monorepo,
we know that they're all going to want to load their FFI symbols from
the combined "libmegazord.so". This adds a config option so that
the generated code can use this cdylib name by default, rather
than having to specify it as an override at runtime.

Ref https://github.com/mozilla/application-services/issues/3874

### DIFF
--- a/examples/arithmetic/uniffi.toml
+++ b/examples/arithmetic/uniffi.toml
@@ -1,2 +1,9 @@
 [bindings.kotlin]
 package_name = "org.mozilla.uniffi.example.arithmetic"
+cdylib_name = "arithmetical"
+
+[bindings.python]
+cdylib_name = "arithmetical"
+
+[bindings.swift]
+cdylib_name = "arithmetical"
--- a/uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
@@ -16,18 +16,23 @@ use crate::MergeWith;
 #[derive(Debug, Default, Clone, Serialize, Deserialize)]
 pub struct Config {
     package_name: Option<String>,
+    cdylib_name: Option<String>,
 }
 
 impl Config {
-    fn default_package_name() -> String {
-        "uniffi".into()
-    }
-
     pub fn package_name(&self) -> String {
         if let Some(package_name) = &self.package_name {
             package_name.clone()
         } else {
-            Config::default_package_name()
+            "uniffi".into()
+        }
+    }
+
+    pub fn cdylib_name(&self) -> String {
+        if let Some(cdylib_name) = &self.cdylib_name {
+            cdylib_name.clone()
+        } else {
+            "uniffi".into()
         }
     }
 }
@@ -36,6 +41,7 @@ impl From<&ComponentInterface> for Config {
     fn from(ci: &ComponentInterface) -> Self {
         Config {
             package_name: Some(format!("uniffi.{}", ci.namespace())),
+            cdylib_name: Some(format!("uniffi_{}", ci.namespace())),
         }
     }
 }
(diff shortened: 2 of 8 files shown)


examples/arithmetic/uniffi.toml
```diff
@@ -1,2 +1,9 @@
 [bindings.kotlin]
 package_name = "org.mozilla.uniffi.example.arithmetic"
+cdylib_name = "arithmetical"
+
+[bindings.python]
+cdylib_name = "arithmetical"
+
+[bindings.swift]
+cdylib_name = "arithmetical"
```
rfk, line 9: I think it would be nice to be able to specify this once, perhaps in a top-level `[bindings]` config group, but that's a bigger refactor of the way config works than I was willing to take on this afternoon.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -16,18 +16,23 @@ use crate::MergeWith;
 #[derive(Debug, Default, Clone, Serialize, Deserialize)]
 pub struct Config {
     package_name: Option<String>,
+    cdylib_name: Option<String>,
 }
 
 impl Config {
-    fn default_package_name() -> String {
-        "uniffi".into()
-    }
```
rfk, line 24: I didn't see the point in adding a `default_cdylib_name()` method and maybe more in future, so I just inlined the defaults into the methods below.

rfk:
Two things to note here for completeness:

* The swift backend doesn't locate the dylib dynamically, it assumes that the app has linked to the appropriate library at build time. So we only need to tell the build step about the custom name, not the generated code.
* The gecko_js backend doesn't locate the dylib dynamically, it assumes that the symbols will be linked in to the application when building Gecko. So there's nothing to do for that backend.

rfk:
@lougeniaC64 this seemed like a relatively friendly on-ramp PR to the uniffi codebase and it affects something you're actively working on, so I hope you don't mind me looping you in for a review here :-)

lougeniaC64:
> @lougeniaC64 this seemed like a relatively friendly on-ramp PR to the uniffi codebase and it affects something you're actively working on, so I hope you don't mind me looping you in for a review here :-)

Thanks for this--will do!

uniffi_bindgen/src/bindings/python/templates/NamespaceLibraryTemplate.py
```diff
@@ -5,19 +5,19 @@
 # E.g. we might start by looking for the named component in `libuniffi.so` and if
 # that fails, fall back to loading it separately from `lib${componentName}.so`.
 
-def loadIndirect(componentName):
+def loadIndirect():
```
mhammond, line 8: It's a shame this is done here - eg, the "win" branch isn't really correct in most scenarios - a `lib_` prefix is the exception rather than the norm and tends to reflect someone who came from a non-windows background - eg, my "system32" directory has 3500 dll files, which exactly zero starting with "lib_" :)

(I've no problem with going ahead with this though, just thought I'd mention it)

lougeniaC64 (approved):
I'm going to need to spend some time with the latest ADR, but after doing some google research everything here looks good.

uniffi_bindgen/src/bindings/python/templates/NamespaceLibraryTemplate.py
lougeniaC64: @mhammond FYI I didn't see your comment before my approval 😬

rfk: Thanks for the heads-up; I'm sure you can guess how many times I've tried running this on Windows in practice given our current need on that platform, which is a shame :-/

rfk: I'm going to push ahead with merging this, but I'm also installing the native Windows Rust build toolchain right now and will poke at it in between some other things, ref https://github.com/mozilla/uniffi-rs/issues/391

mhammond: > @mhammond FYI I didn't see your comment before my approval 😬

No probs - I didn't intend for my comment to block approval!

PR #390, opened 2021-02-18
Tweak generated Kotlin for compatibility with Android API 21.

rfk:
While trying to uniffi the fxa-client crate, I discovered some linter errors in the generated Kotlin saying "`Error: Call requires API level 24 (current min is 21)`". A bit of ducking-and-going led me to the tweaks here which silence the warnings. For good measure, I've also avoided generating helper code when it won't be used by the bindings.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
@@ -1,6 +1,7 @@
 // A handful of classes and functions to support the generated data structures.
 // This would be a good candidate for isolating in its own ffi-support lib.
 
+{% if ci.iter_object_definitions().len() > 0 %}
 abstract class FFIObject(
     private val handle: AtomicLong
 ) {
@@ -28,7 +29,9 @@ inline fun <T : FFIObject, R> T.use(block: (T) -> R) =
             // swallow
         }
     }
+{% endif %}
 
+{% if ci.iter_callback_interface_definitions().len() > 0 %}
 internal typealias Handle = Long
 internal class ConcurrentHandleMap<T>(
     private val leftMap: MutableMap<Handle, T> = mutableMapOf(),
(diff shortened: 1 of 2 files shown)


uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -41,7 +44,7 @@ internal class ConcurrentHandleMap<T>(
     fun insert(obj: T): Handle =
         lock.withLock {
             rightMap[obj] ?:
-                currentHandle.getAndAccumulate(stride) { a, b -> a + b }
+                currentHandle.getAndAdd(stride)
```
rfk, line 47: The tests pass, but I'm not 100% sure I haven't subtly messed something up here with the `.also` and what-not...

jhugman: No, this is fine.

At some point, we'll need to start writing lower level tests for all this helper code.

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -377,7 +377,10 @@ internal fun lower{{ canonical_type_name }}(m: Map<String, {{ inner_type_name }}
 
 internal fun write{{ canonical_type_name }}(v: Map<String, {{ inner_type_name }}>, buf: RustBufferBuilder) {
     buf.putInt(v.size)
-    v.forEach { k, v ->
+    // The parens on `(k, v)` here ensure we're calling the right method,
+    // which is important for compatibility with older android devices.
+    // Ref https://blog.danlew.net/2017/03/16/kotlin-puzzler-whose-line-is-it-anyways/
+    v.forEach { (k, v) ->
```
jhugman, line 383: Destructuring, FTW

jhugman (approved):
This is good!

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
rfk: Indeed. I don't have a good idea of how to structure such tests though :-/

Issue #391, opened 2021-02-18
Make things work on a native Windows build


rfk:
Rust treats Windows as a first-class platform, and so should we. I'm going to set up VSCode with the native Windows Rust tools and see whether I can get `cargo test` working without errors.

Ref https://github.com/mozilla/uniffi-rs/pull/389#discussion_r578741329 for one specific potential source of said errors.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2062)

PR #393, opened 2021-02-18
Get `cargo test py` passing on Window.

rfk:
Fixes https://github.com/mozilla/uniffi-rs/issues/391.

I have no intention of messing around with Swift or Kotlin on Windows, but this at least makes `cargo test py` run successfully.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/templates/NamespaceLibraryTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/NamespaceLibraryTemplate.py
@@ -11,7 +11,14 @@ def loadIndirect():
     elif sys.platform == "darwin":
         libname = "lib{}.dylib"
     elif sys.platform.startswith("win"):
-        libname = "lib_{}.dll"
+        # As of python3.8, ctypes does not seem to search $PATH when loading DLLs.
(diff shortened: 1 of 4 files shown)


uniffi_bindgen/src/bindings/python/templates/NamespaceLibraryTemplate.py
```diff
@@ -11,7 +11,13 @@ def loadIndirect():
     elif sys.platform == "darwin":
         libname = "lib{}.dylib"
     elif sys.platform.startswith("win"):
-        libname = "lib_{}.dll"
+        # As of python3.8, Windows does not search $PATH when looking for DLLs.
+        # It doesn't seem right to mess around with `os.add_dll_directory` and affect
+        # loading for the entire application, so for now, let's use a full path.
+        libname = os.path.join(
```
mhammond, line 17: mildly surprised that 3.8 changed how ctypes loads DLLs (I do know it changed how dynamic extension modules are loaded, but IIUC this isn't one of them) - and even then, the change was how *dependent* DLLs were located. IOW, I'm not sure this comment is 100% true (but also no evidence that it's not :) But regardless, loading a generic DLL via the full path when known is the right thing to do, so 👍

mhammond (approved):

PR #397, opened 2021-02-19
Add a small convenience macro for including generated scaffolding.

rfk:
Smaller and simpler than https://github.com/mozilla/uniffi-rs/pull/386, this is a macro to include the generated Rust scaffolding by name, without having to know the precise details of where and how it gets generated.


mhammond (approved):
👍

PR #402, opened 2021-02-24
Fix a clippy lint in the generated scaffolding for enums with associated data

rfk:
Otherwise clippy yells at me when trying to use this feature in the fxa-client crate:

```
error: the `v @ _` pattern can be written as just `v`
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
   Rust and Foreign Language tests:
     docker:
       - image: rfkelly/uniffi-ci:latest
-    resource_class: medium
+    resource_class: large
     steps:
       - run: cat ~/.profile >> $BASH_ENV
       - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -37,7 +37,7 @@ jobs:
   Rust and Foreign Language tests:
     docker:
       - image: rfkelly/uniffi-ci:latest
-    resource_class: medium
+    resource_class: large
```
rfk, line 40: Looks like `rondpoint` has started reliably exceeding the resources available to a medium instance. I'm going to bump it to a large for now, but that's probably a sign that it needs to be split up.

mhammond: ouch

mhammond (approved):
rs=me

.circleci/config.yml
rfk: yeah :-/

PR #403, opened 2021-02-24
Mark record fields as `public` in Swift bindings.

rfk:
Swift struct fields default to `internal` visibility, loosly
analogous to Rust's `pub(crate)`. That's clearly enough for them
to work fine in our tests, but when I tried out using these
across Swift module boundaries, I got errors about being able
to access to fields.

Since the whole point of record types is to make structured
data available, I think it makes sense for their fields to be
marked public.


jhugman (approved):
This looks good. Do we need to add the same to methods as well?

rfk:
> Do we need to add the same to methods as well?

It looks like methods are already marked `public`, so I think we're OK on that front.

PR #404, opened 2021-02-24
No bug - Fix up iOS example app.

jhugman:
The iOS app is not on CI, so suffers from bit rot. 

This is the occasional fix for that bit rot.

### DIFF
--- a/examples/app/ios/IOSApp-Bridging-Header.h
+++ b/examples/app/ios/IOSApp-Bridging-Header.h
@@ -6,6 +6,6 @@
 #ifndef IOSApp_Bridging_Header_h
 #define IOSApp_Bridging_Header_h
 
-#import "todolist-Bridging-Header.h"
+#import "uniffi_todolist-Bridging-Header.h"
 
 #endif /* IOSApp_Bridging_Header_h */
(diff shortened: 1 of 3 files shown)


examples/app/ios/IOSApp-Bridging-Header.h
```diff
@@ -6,6 +6,6 @@
 #ifndef IOSApp_Bridging_Header_h
 #define IOSApp_Bridging_Header_h
 
-#import "todolist-Bridging-Header.h"
+#import "uniffi_todolist-Bridging-Header.h"
```
rfk, line 9: I don't recall what lead us to changing the name here, bit as I was messing around with an iOS integration yesterday, I thought to myself "what is the point of this ugly little `uniffi_` prefix on the bridging header..?". Do you have a strong preference here either way @jhugman?

examples/app/ios/README.md
```diff
@@ -26,14 +26,14 @@ This will not be a complete tutorial on how to use `uniffi`, just the bits to ge
 4. Use `uniffi-bindgen` on each file to generate the headers and swift scaffolding.
 
 ```sh
-uniffi-bindgen generate "$INPUT_FILE_PATH" --language swift --out-dir "$DERIVED_FILE_DIR"
+$HOME/.cargo/bin/uniffi-bindgen generate "$INPUT_FILE_PATH" --language swift --out-dir "$DERIVED_FILE_DIR"
```
rfk, line 29: Hrm, it's a shame we can't just pick this up from `$PATH`, but I noticed the same thing. Do you know if there's a way we could tell XCode about the `$PATH` somehow?

rfk (approved):
Looks good, thanks! I had some questions but they're very much for future followups.

PR #405, opened 2021-02-25
Flesh out the docs on consuming Swift bindings.

rfk:
@jhugman here are my notes from trying out uniff with the fxa-client crate. I may not be using the correct technical terms in places here since I'm far from a Swift expert, but hopefully this gets the gist across.


docs/manual/src/swift/module.md
```diff
@@ -0,0 +1,34 @@
...
+module example {
+    header "path/to/uniffi_example-Bridging-Header.h"
+    export *
+}
+```
+
+Then use `swiftc` to combine the cdylib from your Rust crate with the generated
+Swift bindings:
+
+```
```
rfk, line 21: This is based on what our tests do, and it felt weird not to have it documented anywhere.

rfk:
I should add, the xcode project file in https://github.com/mozilla/application-services/pull/3876 seems to be working correctly to build the fxa-client iOS bindings, to the point that I'm able to successfully incorporate the build into the firefox-ios app. So I guess that means I must be doing things approximately right...right? :-)

docs/manual/src/swift/module.md
```diff
@@ -0,0 +1,34 @@
+# Compiling a Swift module
```
jhugman, line 1: We do generate a module map, but only while running tests.

jhugman (commented):
At some point, we should move/adapt the docs from the [iOS README.md][1] to here.

[1]: https://github.com/mozilla/uniffi-rs/blob/main/examples/app/ios/README.md

rfk:
> we should move/adapt the docs from the iOS README.md to here.

I'm ashamed to admit, I did not read these docs in any detail before setting out here. I wonder if it would be more useful to link *out* to them more explicitly from the manual, rather than trying to pull them in?

docs/manual/src/swift/module.md
rfk: We do, but I don't think there's a useful way for the consumer to have us generate one for them. Is it worth us trying to expose such a thing for convenience?

docs/manual/src/swift/xcode.md
```diff
@@ -1,8 +1,52 @@
...
-$HOME/.cargo/bin/uniffi-bindgen generate $INPUT_FILE_PATH --language swift --out-dir $DERIVED_FILE_DIR\n
 ```
+#include "uniffi_example-Bridging-Header.h"
+```
+
+For this to work without complaint from XCode, you also need to add the
+generated header file as a **Public** header in the "Headers" build phase
+of your project (which is why it's useful to generate this file somewhere in
+your source tree, rather than in temporary build directory).
+
```
jhugman, line 52: We should probably mention not explicitly compiling the generated swift code here as well.

docs/manual/src/swift/xcode.md
```diff
@@ -1,8 +1,52 @@
...
+3. Include the generated bridging header into your overall bridging header.
+
+There is also an example app in the UniFFI project repo that may be helpful.
+
+## Compiling the Rust crate.
+
+Sorry, configuring XCode to compile the Rust crate into a staticlib
+is beyond the scope of this document. However you do so, make sure you
+include the resulting `libexample.a` file in the "Link Binary with Libraries"
+build phase for your framework.
```
jhugman, line 19: We have an example in the repo, in `examples/app/ios`.

jhugman (approved):
Ryan: I'm really sorry this took so long to get a review. I have been getting halfway through your documentation PRs, and then getting distracted, only for someone else to have reviewed them and being merged.

This one looks good. I have added a couple of final comments, but nothing to hold for another review cycle.

rfk:
> I have been getting halfway through your documentation PRs, and then getting distracted, only for someone else
> to have reviewed them and being merged.

Thanks @jhugman. I in turn have been more trigger-happy on the "merge" button than usual because I've been trying not to leave too many open loops when I head off for PTO, so I'm sorry if this has sent a signal that I don't appreciate your input; I very much do!

PR #406, opened 2021-02-25
Export the uniffi-bindgen `main` for possible re-use.

rfk:
I want to experiment with some changes to the UniFFI development flow,
and being and to import and use this from other crates will be
very helpful in doing so.

(It was a literal copy-paste job from one file to another)


mhammond (approved):
Bit of a shame you can't just export it from main.rs, but that seems impossible (and I guess it makes sense you can't do that) so 👍

rfk:
Heh, yeah, I tried that and was disappointed that it didn't seem to work

PR #407, opened 2021-02-25
(cargo-release) version v0.7.2

rfk:
(A friendly reminder that if you merge this, please create a merge commit rather than squashing or rebasing, since the release tag is associated with the exact commit staged in this PR).

### DIFF
--- a/examples/arithmetic/Cargo.toml
+++ b/examples/arithmetic/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "uniffi-example-arithmetic"
 edition = "2018"
-version = "0.7.1"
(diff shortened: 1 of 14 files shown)


dmose (approved):
r=dmose

I mean, really, feel-free to use rs=dmose for any such version-bumping commit in the future.

rfk:
> I mean, really, feel-free to use rs=dmose for any such version-bumping commit in the future.

Thanks. I agree that having a PR for this does seem a bit pointless, especially since I've already pushed the release to crates.io. Perhaps we should all just have a gentlepersons' agreement to use administrative privileges to push release tags without a review.

PR #409, opened 2021-02-26
Eliminate duplicate declarations of helper structs and extensions in Swift

### LINKED ISSUE #408: Eliminate re-declaration of helper structs and protocols in Swift bindings
Writing the docs in https://github.com/mozilla/uniffi-rs/pull/405 made me realize, I think we're going to face a build issue when we come to incorporate multipe uniffied components into a single project. We should avoid this by generating unique names for all helper structs, using the component namespace as a...yeah, a namespace.

The Swift bindings for a uniffied component generate some helper C structs and declarations in a file named `uniffi_<namespace>-Bridging-Header.h`, which we expect to be `#include`d into the overall header file of the appservices megazord. If we have more than one uniffied component, then any symbol name conflicts in the generated header files risk causing errors when processing the combined header.

The template for this file is [here](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h) and it defines structs named `RustBuffer`, `ForeignBytes`, etc. I suspect we'll encounter problems with duplicating these definitions.

One solution would be to name them `<namespace>_RustBuffer` etc to avoid conflicts, but I'm open to other suggestions!

(The gecko-js backend has a similar problem, which IIUC it solves through a use of [C++ namespaces](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/gecko_js/templates/RustBufferHelper.h) and an [optional global name prefix](https://github.com/mozilla/uniffi-rs/blob/78af31928fa1c34a3c628bc3fd1a274313e34cc8/uniffi_bindgen/src/bindings/gecko_js/gen_gecko_js.rs#L25)).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2078)

jhugman:
Fixes #408.

This PR:

* hides all helpers in swift protocols and extensions behind `fileprivate` access modifiers.
* protects redefinition of C structs with header guards.

### DIFF
--- a/examples/app/ios/IOSApp/ContentView.swift
+++ b/examples/app/ios/IOSApp/ContentView.swift
@@ -10,7 +10,8 @@ struct ContentView: View {
     }
 
     var todoList: TodoList
-    @State private var clicked = 0
+    let stride: UInt64 = UInt64(1)
+    @State private var clicked = UInt64.zero
     @State private var text = ""
     var body: some View {
         VStack {
@@ -19,7 +20,7 @@ struct ContentView: View {
                 TextField("New task", text: $text, onCommit:  {
                     try! self.todoList.addEntry(entry: TodoEntry(text: "\(clicked) \(text)"))
                     text = ""
-                    clicked = clicked + 1
+                    clicked = try! add(a: clicked,  b: stride)
                 }).padding()
             }
 
@@ -32,7 +33,7 @@ struct ContentView: View {
                         if let i = index.first {
                             let entry = todoList.getEntries()[i]
                             try! todoList.clearItem(todo: entry.text)
-                            clicked = clicked - 1
+                            clicked = try! sub(a: clicked,  b: stride)
                         }
                     }
                 }
--- a/examples/arithmetic/Cargo.toml
+++ b/examples/arithmetic/Cargo.toml
@@ -7,7 +7,7 @@ license = "MPL-2.0"
 publish = false
 
 [lib]
-crate-type = ["cdylib", "lib"]
+crate-type = ["staticlib", "cdylib", "lib"]
 name = "arithmetical"
 
 [dependencies]
--- a/uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
@@ -5,19 +5,19 @@
 
 #}
 
-protocol RustError: LocalizedError {
+fileprivate protocol RustError: LocalizedError {
     static func fromConsuming(_ rustError: NativeRustError) throws -> Self?
 }
 
 // An error type for FFI errors. These errors occur at the UniFFI level, not
 // the library level.
-enum InternalError: RustError {
+fileprivate enum UniffiInternalError: RustError {
     case bufferOverflow
     case incompleteData
     case unexpectedOptionalTag
     case unexpectedEnumCase
     case emptyResult
-    case unknown(message: String = "")
+    case unknown(_ message: String)
 
     public var errorDescription: String? {
         switch self {
@@ -30,18 +30,18 @@ enum InternalError: RustError {
         }
     }
 
-    static func fromConsuming(_ rustError: NativeRustError) throws -> Self? {
+    fileprivate static func fromConsuming(_ rustError: NativeRustError) throws -> Self? {
         let message = rustError.message
         defer {
             if message != nil {
-                try! rustCall(InternalError.unknown()) { err in
+                try! rustCall(UniffiInternalError.unknown("UniffiInternalError.fromConsuming")) { err in
                     {{ ci.ffi_string_free().name() }}(message!, err)
                 }
             }
         }
         switch rustError.code {
         case 0: return nil
-        default: return .unknown(message: String(cString: message!))
+        default: return .unknown(String(cString: message!))
         }
     }
 }
@@ -69,11 +69,11 @@ public enum {{e.name()}}: RustError {
     // The name is attempting to indicate that we free message if it
     // existed, and that it's a very bad idea to touch it after you call this
     // function
-    static func fromConsuming(_ rustError: NativeRustError) throws -> Self? {
+    fileprivate static func fromConsuming(_ rustError: NativeRustError) throws -> Self? {
         let message = rustError.message
         defer {
             if message != nil {
-                try! rustCall(InternalError.unknown()) { err in
+                try! rustCall(UniffiInternalError.unknown("{{e.name()}}.fromConsuming")) { err in
                     {{ ci.ffi_string_free().name() }}(message!, err)
                 }
             }
@@ -92,28 +92,28 @@ public enum {{e.name()}}: RustError {
 }
 {% endfor %}
 
-func rustCall<T, E: RustError>(_ err: E, _ cb: (UnsafeMutablePointer<NativeRustError>) throws -> T?) throws -> T {
+private func rustCall<T, E: RustError>(_ err: E, _ cb: (UnsafeMutablePointer<NativeRustError>) throws -> T?) throws -> T {
     return try unwrap(err) { native_err in
         return try cb(native_err)
     }
 }
 
-func nullableRustCall<T, E: RustError>(_ err: E, _ cb: (UnsafeMutablePointer<NativeRustError>) throws -> T?) throws -> T? {
+private func nullableRustCall<T, E: RustError>(_ err: E, _ cb: (UnsafeMutablePointer<NativeRustError>) throws -> T?) throws -> T? {
     return try tryUnwrap(err) { native_err in
         return try cb(native_err)
     }
 }
 
 @discardableResult
- func unwrap<T, E: RustError>(_ err: E, _ callback: (UnsafeMutablePointer<NativeRustError>) throws -> T?) throws -> T {
+private func unwrap<T, E: RustError>(_ err: E, _ callback: (UnsafeMutablePointer<NativeRustError>) throws -> T?) throws -> T {
     guard let result = try tryUnwrap(err, callback) else {
-        throw InternalError.emptyResult
+        throw UniffiInternalError.emptyResult
     }
     return result
 }
 
 @discardableResult
-func tryUnwrap<T, E: RustError>(_ err: E, _ callback: (UnsafeMutablePointer<NativeRustError>) throws -> T?) throws -> T? {
+private func tryUnwrap<T, E: RustError>(_ err: E, _ callback: (UnsafeMutablePointer<NativeRustError>) throws -> T?) throws -> T? {
     var native_err = NativeRustError(code: 0, message: nil)
     let returnedVal = try callback(&native_err)
     if let retErr = try E.fromConsuming(native_err) {
--- a/uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift
@@ -1,4 +1,4 @@
-public struct {{ rec.name()|class_name_swift }}:  ViaFfiUsingByteBuffer, ViaFfi, Equatable, Hashable {
+public struct {{ rec.name()|class_name_swift }}: Equatable, Hashable {
     {%- for field in rec.fields() %}
     public let {{ field.name()|var_name_swift }}: {{ field.type_()|type_swift }}
     {%- endfor %}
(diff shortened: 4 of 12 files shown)


uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h
```diff
@@ -6,6 +6,14 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+// The following structs are used to implement the lowest level
+// of the FFI, and thus useful to multiple uniffied crates.
+// We ensure they are declared exactly once, with a header guard, UNIFFI_SHARED_H.
+// This may become a potential footgun if multiple versions of uniffi are being used
+// in the same project.
+#ifndef UNIFFI_SHARED_H
+#define UNIFFI_SHARED_H
```
jhugman, line 15: Note: foot-gun potential

jhugman (commented):
I was hoping to get a test up and running, but it turned in to a battle with xcodebuild.

I'll get this running with Nimbus before formally requesting a review.

uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h
rfk: This is making me realize that I don't have a fully fleshed-out mental model of how the uniffi helper code works across versions, and it's probably worth us discussing a bit at a high-level to get aligned.

In the initial commits for the project, I deliberately generated all of the helper code alongside the module's own code, rather than e.g. making a "uniffi_runtime" Kotlin package that the generated code could depend on. This was partly for simplicity, and partly from a half-baked notion that you might be able to safely depend on multiple uniffied packages that were using multiple different versions of uniffi. But it's not clear whether we actually need that in practice, or what the costs of supporting that are going to be.

I'm fine with carefully moving away from that model, and assuming that all the uniffied components in a build are using the same version of uniffi, as long as we can mitigate the footgun potential.

What do you think about trying to put a little version check in here? Something like (untested!):

```
#ifdef UNIFFI_SHARED_H
  #ifndef UNIFFI_VERSION_1_2_3
  #error Combining helper code from multiple versions of uniffi is not supported
  #endif
#else
  #define UNIFFI_SHARED_H
  #define UNIFFI_VERSION_1_2_3
  // ...the generated header code
#endif
```

We'd have to do a bit of plumbing to expose the current version number in a suitable form, but that doesn't seem too bad.

jhugman: Your `(untested!)` scheme is stricter than I was thinking, but probably more sensible. 

I'll leave the scheme in my head here, only to get it out of my head.

```h
#ifndef UNIFFI_SHARED_0_7_2
#define UNIFFI_SHARED_0_7_2
// everthing defined up until now.
#endif
#ifndef UNIFFI_SHARED_0_8_0
#define UNIFFI_SHARED_0_8_0
// A theoretical non-breaking addition in 0.8.0
#endif
```

But I only have a partial solution for breaking changes.

It strikes me that this is very much bread and butter for more experienced C developers than I.

@dmose may be able to point to prior art here.

dmose: I don't have much experience in the space of handling use of multiple SDK versions of the same thing in an executable at compile time.

That said, I'm a fan of the "not supported" approach, because I think it keeps us from expanding our problem space to include complexity that we don't currently have reason to believe we'll need.

If you really want more C/C++ thoughts, there are some C/C++ experts at Mozila (eg Botond) that one could reach out to.

uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift
```diff
@@ -1,4 +1,4 @@
-public struct {{ rec.name()|class_name_swift }}:  ViaFfiUsingByteBuffer, ViaFfi, Equatable, Hashable {
+public struct {{ rec.name()|class_name_swift }}: Equatable, Hashable {
```
jhugman, line 1: The general pattern here is: 

1. Make the uniffi protocol `fileprivate`, so we can't see it outside of this file.
2. Split the implementing `extension` off from the public extensions/declaration. This is optional, but is more swifty.
3. Implement the `fileprivate` protocol in a vanilla extension, but make all members `fileprivate`.

Neither the methods nor the protocols should now be visible outside this file.

examples/arithmetic/Cargo.toml
```diff
@@ -7,7 +7,7 @@ license = "MPL-2.0"
 publish = false
 
 [lib]
-crate-type = ["cdylib", "lib"]
+crate-type = ["staticlib", "cdylib", "lib"]
```
jhugman, line 10: `staticlib`s because iOS doesn't deal with anything else.

examples/app/ios/IOSApp/ContentView.swift
```diff
@@ -32,7 +33,7 @@ struct ContentView: View {
                         if let i = index.first {
                             let entry = todoList.getEntries()[i]
                             try! todoList.clearItem(todo: entry.text)
-                            clicked = clicked - 1
+                            clicked = try! sub(a: clicked,  b: stride)
```
jhugman, line 36: `examples/app/ios` now deals with two un-zorded uniffied files.

uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h
```diff
@@ -31,6 +47,7 @@ typedef struct NativeRustError {
     char *_Nullable message;
 } NativeRustError;
 
+#endif // def UNIFFI_SHARED_H
```
jhugman, line 50: 1. We need to version bump and a release, so I picked a minor release number.
2. I'm not sure to do that here in this PR or in a separate release PR.

jhugman (commented):
A version bump is needed for this to be usable in application services.

uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h
rfk: > // A theoretical non-breaking addition in 0.8.0

FWIW, I don't really trust us (including myself!) to reason carefully about whether or not each change is going to be sufficiently non-breaking, or even to have the tools to really understand the ways breaking things could happen. We could get there with deliberate effort, but I just don't think we're operating in that mode right now, or that it's a useful way to spend our limited energies at this time.

Rephrased: I think we should deliberately choose not to expend effort on layers of b/w compat right now, it's low value and high risk for the current state of the project. Like so many of our choices here, we should also deliberately plan to revisit that question in future.

uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
```diff
@@ -5,19 +5,19 @@
 
 #}
 
-protocol RustError: LocalizedError {
+fileprivate protocol RustError: LocalizedError {
```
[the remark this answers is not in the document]
rfk: Indeed; I hit this very name conflict over in https://github.com/mozilla/application-services/pull/3900, and also with `RustBuffer` and `rustCall`.

uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift
rfk: This is going to lightly interfere with what I was trying to do over in https://github.com/mozilla/application-services/pull/3900, but since that PR is just a hypothetical experiment, I don't think we should change things in UniFFI to support it. Your proposal here sounds good.

examples/app/ios/IOSApp/ContentView.swift
```diff
@@ -19,7 +20,7 @@ struct ContentView: View {
                 TextField("New task", text: $text, onCommit:  {
                     try! self.todoList.addEntry(entry: TodoEntry(text: "\(clicked) \(text)"))
                     text = ""
-                    clicked = clicked + 1
+                    clicked = try! add(a: clicked,  b: stride)
```
rfk, line 23: Haha, nice!

uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h
```diff
@@ -6,6 +6,22 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+// The following structs are used to implement the lowest level
+// of the FFI, and thus useful to multiple uniffied crates.
+// We ensure they are declared exactly once, with a header guard, UNIFFI_SHARED_H.
+#ifdef UNIFFI_SHARED_H
+    // We also try to prevent mixing versions of shared uniffi header structs
```
rfk, line 13: nano-nit: period at the end of this sentence.

uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h
rfk: I support `v0.8.0` and I support doing it in a separate PR 👍

uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h
```diff
@@ -6,6 +6,22 @@
...
+    // If you add anything to the #else block, you must increment the version in UNIFFI_VERSION_x_y_z.
+    #ifndef UNIFFI_VERSION_0_8_0
+        #error Combining helper code from multiple versions of uniffi is not supported
+    #endif // ndef UNIFFI_VERSION_0_8_0
+#else
+    #define UNIFFI_SHARED_H
+    #define UNIFFI_VERSION_0_8_0
+    // ⚠️ Attention: If you change this #else block (ending in `#endif // def UNIFFI_SHARED_H`) ⚠️
+    // ⚠️ you must increment the version in all instance of UNIFFI_VERSION_ in this file to the ⚠️
+    // ⚠️ current uniffi version.                                                               ⚠️
```
rfk, line 23: Hrm. I'm trying to mentally play out the potential ways this could go wrong versus the value it will provide us.

IIUC we don't have any tests to ensure that this version-bump gets done, so future-me could conceivably add something further down in the `#else` block without really registering this warning. The downside consequences if that does happen could be catastrophic, e.g. arbitrary memory unsafety security problems.

A different approach could have this constant change automatically on every UniFFI version bump, perhaps because we plumb in the current version number through the template, or because we configure `cargo release` to change it for us automatically. The downside consequences of that approach seem small by comparison: two uniffied components using two different versions of UniFFI that *could* have operated perfectly happily together, will refuse to do so.

In practice, I don't think we're going to actually have Swift code generated by two different versions of UniFFI that we want compiled together on purpose. We do strict version checks when building the Rust code, so the megazord won't build. We also generate the Swift bindings on demand so they *should* always use the same version of UniFFI.

So...on balance I think we should treat this as a build-misconfiguration-produces-catastrophic-security-issue prevention mechanism, and automatically use the actual UniFFI release version for this constant. Plumbing it in as a template variable seems like the easiest way to achieve that, although I haven't thought about the "how" in too much detail.

Alternately, if you feel strongly about decoupling this from the UniFFI release version, then I think we should go even further in that direction: name the constant something like `UNIFFI_SHARED_HEADER_v1` and create an entirely separate incrementing version space that we change on its own schedule, and that can't be accidentally confused for the crate release version.

rfk (approved):
Thanks @jhugman!

I'm in favour of everything here except for the fact that it needs an `⚠️ Attention` block, which I've left some comments on below. I'm going to approve this because timezones make our feedback cycle long and I want you to keep your forward momentum here, but please do respond to my feedback below before merge (even if it's just to say "no I strongly disagree" 😅).

uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h
jhugman: > I think we should deliberately choose not to expend effort on layers of b/w compat right now

:100: I agree.

uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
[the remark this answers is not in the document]
jhugman: Hmm. Let's land this, and we can talk about it.

uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h
jhugman: I think there are two sources of incompatibilities here:

1. two different libraries (e.g. `fxa-client` and `nimbus`) having bindings and scaffolding prepared by different versions of uniffi.
2. the same library having bindings prepared by one version, and scaffolding prepared by another version.

I think whatever we choose, the way megazording is working in Xcode, with `embedded-uniffi-bindgen`, 1. is effectively eliminated, and 2. is addressed by including the uniffi version in the `ComponentInterface` hash.

> I think we should deliberately choose not to expend effort on layers of b/w compat right now, […]
> we should also deliberately plan to revisit that question in future

I think I prefer you're final suggestion of using a different version space, until we revisit this in the future.

rfk: > is addressed by including the uniffi version in the ComponentInterface hash.

This is a good point, we have several other layers of protection here. Thanks for talking it through!

PR #414, opened 2021-03-05
Use static assertions to check uniffi/uniffi-bindgen version compatibility.

### LINKED ISSUE #413: `ensure_versions_compatibility` may not check the right version
Filing as a follow-up to https://github.com/mozilla/application-services/pull/3904#issuecomment-790224056, which I'll copy here for clarity. We managed to trigger a CI failure in appservices simply be releasing a new version of the `uniffi` crate.

* Nimbus depends on uniffi "0.7", that is, on the latest release in the "0.7.x" series.
* The appservices megazord depends on uniffi "^0.7.1", that is, on any release in the "0.7.x" series that descends from "0.7.1".
* The appservices workspace has a `Cargo.lock` file which pins its release of uniffi specifically to "0.7.1".
* Nimbus does not have a `Cargo.lock` file, and is not in the appservices workspace.

This was all fine and good when we cut the previous release of appservices, because v0.7.1 was the latest release of UniFFI. But now that there is a v0.7.2 release available, AFAICT here's what happens:

* CI installs uniffi `v0.7.1`, which is the version we expect to use and the version that is pinned in `Cargo.lock`.
* CI tries to build nimbus-sdk, invoking its `build.rs` to generate the uniffi bindings.
* UniFFI executes its [`ensure_version_compatibility`](https://github.com/mozilla/uniffi-rs/blob/56e463fe44286685276d7efd8657ad23d73639f3/uniffi_bindgen/src/lib.rs#L153) function, which calls `cargo metadata` on the target crate to check what version of uniffi it depends on.
* `cargo metadata` executes on the nimbus-sdk `Cargo.toml`. It's not part of a workspace and it doesn't have a `Cargo.lock`, so *when looked at in isolation* it appears to depend on the latest release of the uniffi "0.7.x" series, which is now `v0.7.2`.
* UniFFI correctly refuses to generate bindings for what appears to be an incompatible crate.

To problem here is that we call `cargo metadata` to see what version of `uniffi` is depended on by the target crate. But this doesn't take into account the dependency tree of the build that's currently in progress, it looks at what version the dependency *would* resolve to if the target crate were built in isolation.

I'm not really sure how to fix this, but will do a bit of digging...

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2087)

rfk:
(I know I said over in https://github.com/mozilla/uniffi-rs/issues/413#issuecomment-790259989 that this may not be too much of a problem
in practice, but I suddenly realised what a good fix would look like and decided to
try it out).

Fixes #413.

Prior to this commit, the `uniffi-bindgen` command would use `cargo metadata`
to discover the version of uniffi being used by the target crate, and would
error out if did not match the version of uniffi-bindgen itself.

Unfortunately, this approach is not reliable when the target crate is being
compiled as a dependency rather than as the primary crate. The build process
may pick a version of uniffi based on the `Cargo.lock` of the primary crate
or based on resolving compatible versions with other dependencies, while
while `cargo metadata` will report the verison of uniffi that would be used
if you built the target crate in isolation.

To avoid these potential sources of error, this commit switches to an approach
based on static assertions, in two halves:

* When generating the Rust scaffolding, we emit a call to a new
  macro `uniffi::assert_compatible_version!(v)` and pass it a static
  string constant representing the version of uniffi_bindgen that
  generated the code.

* When compiling the resulting code as part of a uniffied crate, the
  `assert_compatible_version!` macro will check the static string against
  the version of uniffi being used in the build, producing a compile-time
  failure if they do not match.

This is a slightly worse consumer experience because the error message
isn't as good, but it's more reliable, so I think it's a win overall.

Be warned, I had to do some terrible things to perform this check in
a compile-time const context on stable Rust :-/

I've also taken the opportunity to try out the `trybuild` crate to
write a testcase for the actual compiler error message generated
by the macro. As we come to iterate more on the developer experience,
I expect `trybuild` or something like it will become pretty important.

### DIFF
--- a/docker/Dockerfile-build
+++ b/docker/Dockerfile-build
@@ -36,8 +36,8 @@ RUN sudo apt-get update -qq \
 
 RUN rustup update \
     # This specific version of Rust is required for compatibility with mozilla-central
-    && rustup install 1.43.0 \
-    && rustup default 1.43.0
+    && rustup install 1.50.0 \
+    && rustup default 1.50.0
 
 RUN mkdir -p /tmp/setup-swift \
     && cd /tmp/setup-swift \
--- a/uniffi/src/lib.rs
+++ b/uniffi/src/lib.rs
@@ -50,6 +50,74 @@ pub mod deps {
     pub use static_assertions;
 }
 
+const PACKAGE_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+// For the significance of this magic number 10 here, and the reason that
+// it can't be a named constant, see the `check_compatible_version` function.
+static_assertions::const_assert!(PACKAGE_VERSION.as_bytes().len() < 10);
+
+/// Check whether the uniffi runtime version is compatible a given uniffi_bindgen version.
+///
+/// The result of this check may be used to ensure that generated Rust scaffolding is
+/// using a compatible version of the uniffi runtime crate. It's a `const fn` so that it
+/// can be used to perform such a check at compile time.
+pub const fn check_compatible_version(bindgen_version: &'static str) -> bool {
+    // While UniFFI is still under heavy development, we require that
+    // the runtime support crate be precisely the same version as the
+    // build-time bindgen crate.
+    //
+    // What we want to achieve here is checking two strings for equality.
+    // Unfortunately Rust doesn't yet support calling the `&str` equals method
+    // in a const context. We can hack around that by doing a byte-by-byte
+    // comparison of the underlying bytes.
+    let package_version = PACKAGE_VERSION.as_bytes();
+    let bindgen_version = bindgen_version.as_bytes();
+    // What we want to achieve here is a loop over the underlying bytes,
+    // something like:
+    // ```
+    //  if package_version.len() != bindgen_version.len() {
+    //      return false
+    //  }
+    //  for i in 0..package_version.len() {
+    //      if package_version[i] != bindgen_version[i] {
+    //          return false
+    //      }
+    //  }
+    //  return true
+    // ```
+    // Unfortunately stable Rust doesn't allow `if` or `for` in const contexts,
+    // so code like the above would only work in nightly. We can hack around it by
+    // statically asserting that the string is shorter than a certain length
+    // (currently 10 bytes) and then manually unrolling that many iterations of the loop.
+    //
+    // Yes, I am aware that this is horrific, but the externally-visible
+    // behaviour is quite nice for consumers!
+    package_version.len() == bindgen_version.len()
+        && (package_version.len() == 0 || package_version[0] == bindgen_version[0])
+        && (package_version.len() <= 1 || package_version[1] == bindgen_version[1])
+        && (package_version.len() <= 2 || package_version[2] == bindgen_version[2])
+        && (package_version.len() <= 3 || package_version[3] == bindgen_version[3])
+        && (package_version.len() <= 4 || package_version[4] == bindgen_version[4])
+        && (package_version.len() <= 5 || package_version[5] == bindgen_version[5])
+        && (package_version.len() <= 6 || package_version[6] == bindgen_version[6])
+        && (package_version.len() <= 7 || package_version[7] == bindgen_version[7])
+        && (package_version.len() <= 8 || package_version[8] == bindgen_version[8])
+        && (package_version.len() <= 9 || package_version[9] == bindgen_version[9])
+        && package_version.len() < 10
+}
+
+/// Assert that the uniffi runtime version matches an expected value.
+///
+/// This is a helper hook for the generated Rust scaffolding, to produce a compile-time
+/// error if the version of `uniffi_bindgen` used to generate the scaffolding was
+/// incompatible with the version of `uniffi` being used at runtime.
+#[macro_export]
+macro_rules! assert_compatible_version {
+    ($v:expr $(,)?) => {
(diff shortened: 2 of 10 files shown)


uniffi/tests/ui/version_mismatch.stderr
```diff
@@ -0,0 +1,7 @@
+error[E0080]: evaluation of constant value failed
+ --> $DIR/version_mismatch.rs:2:1
+  |
+2 | uniffi::assert_compatible_version!("0.0.1"); // An error message would go here.
+  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `0_usize - 1_usize`, which would overflow
+  |
+  = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
rfk, line 7: This is the error experience that consumers would get if they don't have aligned versions. It's...not awesome, but I reckon it would be good enough, and is certainly less confusing than getting a false-positive compilation error like I got over in https://github.com/mozilla/application-services/pull/3904#issuecomment-790224056

docker/Dockerfile-build
```diff
@@ -36,8 +36,8 @@ RUN sudo apt-get update -qq \
 
 RUN rustup update \
     # This specific version of Rust is required for compatibility with mozilla-central
-    && rustup install 1.43.0 \
-    && rustup default 1.43.0
+    && rustup install 1.50.0 \
+    && rustup default 1.50.0
```
rfk, line 40: Rust `1.43` doesn't even have `&&` in const expressions, but we're moving to a min of `1.50` in appservices for unrelated reasons, so this bump seems fine to me.

uniffi/src/lib.rs
```diff
@@ -406,3 +460,12 @@ unsafe impl<V: ViaFfi> ViaFfi for HashMap<String, V> {
         Ok(map)
     }
 }
+
+#[cfg(test)]
+mod test {
+    #[test]
+    fn trybuild_ui_tests() {
+        let t = trybuild::TestCases::new();
+        t.compile_fail("tests/ui/*.rs");
```
jhugman, line 469: TIL about `trybuild` and `ui` tests.

I guess any non-failing tests would just be 'tests' not, UI tests.

uniffi_bindgen/src/templates/scaffolding_template.rs
```diff
@@ -2,6 +2,12 @@
 // Trust me, you don't want to mess with it!
 {% import "macros.rs" as rs %}
 
+// Check for compatibility between `uniffi` and `uniffi_bindgen` versions.
+// Note that we have an error message on the same line as the assertion.
+// This is important, because if the assertion fails, the compiler only
+// seems to show that single line as context for the user.
+uniffi::assert_compatible_version!("{{ uniffi_version }}"); // Please check that you depend on version {{ uniffi_version }} of the `uniffi` crate.
```
jhugman, line 9: To check my understanding: this is about checking that the `uniffi_bindgen`-generated scaffolding is being compiled with the same version of `uniffi`-runtime, and that the later compiled when the generated scaffold and runtime is being compiled. Is this correct?

jhugman (approved):
As far as I understand it, this looks good to me!

uniffi/src/lib.rs
rfk: Right.

uniffi_bindgen/src/templates/scaffolding_template.rs
rfk: > and that the later compiled when the generated scaffold and runtime is being compiled. 

I don't fully parse this part TBH, but what you're saying here otherwise sounds right to me. If I were to try to round-trip that you're saying here through my own parser and renderer it would come out as:

"""
this is about checking that the `uniffi_bindgen`-generated scaffolding is being compiled against the same version of the `uniffi` runtime crate as was used when the `uniffi_bindgen` executable itself was compiled.
"""

Or put another way, if `uniffi-bindgen --version` reports `v1.2.3`, then any bindings it generates will generate a compile-time error if they're being compiled against a version of `uniffi` other than `v1.2.3`.

rfk: (I'll also note that this problem would go away entirely if we switched to using macros rather than an external bindgen command, but that's something to discuss for the future)

uniffi/src/lib.rs
```diff
@@ -50,6 +50,60 @@ pub mod deps {
...
+        && (package_version.len() == 0 || package_version[0] == bindgen_version[0])
+        && (package_version.len() <= 1 || package_version[1] == bindgen_version[1])
+        && (package_version.len() <= 2 || package_version[2] == bindgen_version[2])
+        && (package_version.len() <= 3 || package_version[3] == bindgen_version[3])
+        && (package_version.len() <= 4 || package_version[4] == bindgen_version[4])
+        && (package_version.len() <= 5 || package_version[5] == bindgen_version[5])
+        && (package_version.len() <= 6 || package_version[6] == bindgen_version[6])
+        && (package_version.len() <= 7 || package_version[7] == bindgen_version[7])
+        && (package_version.len() <= 8 || package_version[8] == bindgen_version[8])
+        && (package_version.len() <= 9 || package_version[9] == bindgen_version[9])
```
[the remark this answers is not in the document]
rfk: yes, good idea, I will unroll my thought process here via a series of comments on the increasing horrificity of the final code.

PR #416, opened 2021-03-08
Experiment: Use Rust code directly as the interface definition.

rfk:
This is a fun little something I've been messing around with,
partly just to learn more about Rust macros and `syn`, and partly
to see if they could make a good replacement for our current use
of WebIDL for an interface definition language.

Key ideas:

* Declare the UniFFI component interface using a restricted subset
  of Rust syntax, directly as part of the code that implements the
  Rust side of the component. There is no separate .udl file.

* Use a procedural macro to insert the Rust scaffolding directly
  at crate build time, rather than requiring a separate `build.rs`
  setup to generate and include it.

* When generating the foreign-language bindings, have `uniffi-bindgen`
  parse the Rust source directly in order to find the component
  interface.

This seems to work surprisingly well so far. If we do decide to
go this route, the code here will need a lot more polish before
it's ready to land...but it works! And it works in a way that's
not too conceptually different from what we're currently doing
with a separate `.udl` file.


uniffi_bindgen/src/bindings/udl/gen_udl.rs
```diff
@@ -0,0 +1,69 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk, line 3: Heh, so, this is a litle odd, but I also added a UDL-generating bindings backend. That is: you can use this to *parse* a `ComponentInterface` from a Rust source file and then *generate* a corresponding `.udl` file. This was entirely in support of https://github.com/mozilla/application-services/pull/3876, because I didn't want to keep the interface and its doc in sync between two different files. Consider this a way to experiment with the parsing-rust-code stuff without necessarily committing to it right away - I was able to write my crate *as though* we had all this fun macro stuff up and running, then automagically generate a corresponding `.udl` file so I could use it with the current release version of UniFFI.

I expect we would not want to actually land this or maintain it long term.

uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -87,12 +88,13 @@ use super::{APIConverter, ComponentInterface};
...
 /// Enums are passed across the FFI by serializing to a bytebuffer, with a
 /// i32 indicating the variant followed by the serialization of each field.
-#[derive(Debug, Clone, Hash)]
+#[derive(Debug, Clone, Hash, Default)]
 pub struct Enum {
     pub(super) name: String,
     pub(super) variants: Vec<Variant>,
     // "Flat" enums do not have, and will never have, variants with associated data.
     pub(super) flat: bool,
+    pub(super) docs: Vec<String>,
```
rfk, line 97: One neat thing that we get pretty much for free from parsing Rust code, is *excellent* support for inline docs on all items. Here, I've given each `ComponentInterface` item a simple `docs` attribute to let it hold any docs parsed from the Rust code, and have spat the docs back out again when converting it into a `.udl`. file. You can imagine a future where we carry the docs through to the generated Kotlin, Swift etc files as well.

uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -154,6 +161,25 @@ impl APIConverter<Enum> for weedle::InterfaceDefinition<'_> {
...
+                .variants
+                .iter()
+                .map(|v| v.convert(ci))
+                .collect::<Result<Vec<_>>>()?;
+        let flat = ! variants.iter().any(|v| v.fields().len() > 0);
+        Ok(Enum {
+            name: self.ident.to_string(),
+            variants,
+            flat,
+            docs: attrs.docs,
```
rfk, line 182: Here, I've taken the same `APIConverter` pattern that we've been using for `weedle` parse nodes and have implemented it for `syn` parse nodes as well. It more-or-less Just Worked. Like what we currently do for weedle, the trickiest part is looking for all the extra bits of syntax that we don't want to support and erroring out if we see any.

For example, this takes a parsed Rust `enum` definition and converts it into a definition for a UniFFI component `enum`.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -119,6 +121,83 @@ impl<'ci> ComponentInterface {
...
 
+    /// Parse a `ComponentInterface` from a string containing a Rust source file.
+    pub fn from_rust(rs: &str) -> Result<Self> {
+        let mut ci = Self::default();
+        ci.uniffi_version = env!("CARGO_PKG_VERSION").to_string();
+        // There's some lifetime thing with the errors returned from parsing
+        // that my own lifetime is too short to worry about figuring out; unwrap and move on.
+        let file_src = syn::parse_file(rs).expect("Failed to parse Rust code");
+        // First, let's see what sort of style of file we're dealing with.
+        // It might be `include_scaffolding!` or it might be `#[declare_interface]`.
```
rfk, line 132: This is currently a bit weird. I'm trying to support parsing both the new `#[declare_interface]` syntax proposed in this PR, and also the existing `include_scaffolding!()` syntax that is offered by the current release version of UniFFI. That's probably not worthwhile except to support short-term experiments with the parsing logic here.

docs/manual/src/interface_definition.md
```diff
@@ -0,0 +1,80 @@
...
+//!
+
+// It can use things from stdlib, submodules, or other Rust crates.
+use std::io::prelude::*;
+
+// But its public interface should be defined inside the `declare_interface` macro.
+#[uniffi_macros::declare_interface]
+mod sprites {
+
+  // The interface can define "records" as structs with public fields.
```
mhammond, line 23: I don't think this comment is as clear as it could be either - it says `"records"` but doesn't state what records are. Saying "they do not have any public method impls" contradicts below where one does.

Maybe you were trying to say something like "A simple struct without an public method impls can be considered a kind of simple 'record'" or similar?

docs/manual/src/tutorial/Rust_scaffolding.md
```diff
@@ -1,50 +1,33 @@
...
-Now we generate some Rust helper code to make the `add` method available to foreign-language bindings.  
-
-First, add `uniffi` to your crate dependencies: it is the runtime support code that powers uniffi's serialization of data types across languages:
+First, add `uniffi` and `uniffi_macros` to your crate dependencies:
 
 ```toml
 [dependencies]
-uniffi = "0.5"
+uniffi = "0.7"
+uniffi_macros = "0.7"
```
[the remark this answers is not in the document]
mhammond: Yeah, I think serde does this - IIRC it used to be necessary to explicitly use `serde_derive` alot more than it is today, and grepping in a-s seems to support that (eg, the only reference to `serde_derive` in autofill is in `Cargo.toml`, but older crates use it everywhere.)

But that probably implies they *will* need to know about it in the `[dependencies]` list but you might be able to change the decorations to, say, `#[uniffi::declare_interface]`, which is a subtly different thing than you are saying here?

docs/manual/src/tutorial/foreign_language_bindings.md
```diff
@@ -6,18 +6,28 @@ As stated in the [Overview](../Overview.md), this library and tutorial does not
...
-This is it, you have an MVP integration of uniffi in your project.
+This is it, you have an MVP integration of UniFFI in your project!
+
+TODO: I think it would be neat to turn this into a cargo subcommand, like:
+
+```
+cargo uniffi generate --language kotlin
+```
+
+The advantage of this UI would be that you don't have to specify the path
```
mhammond, line 31: I think the word 'UI' isn't valuable here.

uniffi_bindgen/src/bindings/udl/gen_udl.rs
mhammond: While that's very cool, I agree we probably don't want it - I was very confused above when I saw a .udl back-end :)

uniffi_bindgen/src/interface/enum_.rs
mhammond: The fact the ffi is bound much tighter to the actual impl really is the key feature here, and doc generation is just the most obvious manifestation of that.

uniffi_bindgen/src/interface/mod.rs
mhammond: agreed!

mhammond (commented):
I skimmed over some of the implementation, but I think this has real value and we should push on it.

docs/manual/src/Overview.md
```diff
@@ -1,13 +1,20 @@
...
-Uniffi is a tool that automatically generates foreign-language bindings targeting Rust libraries.  
+UniFFI is a tool that automatically generates foreign-language bindings targeting Rust libraries.  
 It fits in the practice of consolidating business logic in a single Rust library while targeting multiple platforms, making it simpler to develop and maintain a cross-platform codebase.  
 Note that this tool will not help you ship a Rust library to these platforms, but simply not have to write bindings code by hand [[0]](https://i.kym-cdn.com/photos/images/newsfeed/000/572/078/d6d.jpg).
 
 ## Design
 
-uniffi requires to write an Interface Definition Language (based on [WebIDL](https://heycam.github.io/webidl/)) file describing the methods and data structures available to the targeted languages.  
-This .udl (Uniffi Definition Language) file, whose definitions must match with the exposed Rust code, is then used to generate Rust *scaffolding* code and foreign-languages *bindings*. This process can take place either during the build process or be manually initiated by the developer.
+UniFFI requires to declare the interface you want to expose to other languages using a restricted
```
lougeniaC64, line 9: nit: requires _declaring_ the interface

docs/manual/src/Overview.md
```diff
@@ -16,4 +23,4 @@ This .udl (Uniffi Definition Language) file, whose definitions must match with t
 - Kotlin
 - Swift
 - Python
-- [Gecko](https://en.wikipedia.org/wiki/Gecko_(software)) C++
+- [Gecko](https://en.wikipedia.org/wiki/Gecko_(software)) JavaScript
```
lougeniaC64, line 26: clarifying question: Is this a necessary part of this change or tangentially related?

docs/manual/src/interface/errors.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+Note that you cannot currently use a typedef for the `Result` type, UniFFI only supports
+the full `Result<T, E>` syntax. This limitation may be removed in future.
+
+The resulting `add` function would be exposed to the foreign-language code using
+its native error-handling mechanism. For example:
+
+* In Kotlin, there would be an `ArithmeticErrorException` class with an inner class
+  for each ariant, and the `add` function would throw it.
```
lougeniaC64, line 38: nit: variant

docs/manual/src/Overview.md
rfk: completely unnecessary tangent

docs/manual/src/tutorial/Rust_scaffolding.md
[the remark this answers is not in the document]
rfk: Having the consumer write `#[uniffi::declare_interface]` seems reasonable to me, with the separate "macros" crate being an implementation detail.

examples/geometry/tests/test_generated_bindings.rs
```diff
@@ -1,5 +1,5 @@
 uniffi_macros::build_foreign_language_testcases!(
-    "src/geometry.udl",
+    "src/lib.rs",
```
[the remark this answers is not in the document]
rfk: > I think it might be better if we passed a reference to the crate itself rather than to a specific file,

Months later, I still have a pretty strong gut feeling that treating the crate as the unit of UniFFIcation is the right way to go.

rfk:
I've rebased this atop latest `main` and simplified it a bit in the hope of making it more approachable. In particular:

* I've removed the UDL-generating backend, which was a fun hack but isn't something we'd want to actually land.
* I've removed support for parsing a Rust file with the `include_scaffolding!` syntax; we now only support the `uniffi::declare_interface` macro.

I think a great next step here for interested folks, could be to try porting more of the examples and/or tests over to use the new macro syntax. Probably this won't work for any but the simplest examples, because of missing features! But it will be an interesting exercise in trying to use the new approach.

rfk:
Architecturally, I think the parsing-related logic in the `interface` module is getting a bit out of hand here. If we decide we want to keep both the UDL parsing for b/w compat and the new syn-based parsing, I think we should:

* Keep the base `interface` module as just defining the structs and traits, without any parsing logic.
* Move the weedle-based parsing impls into a submodule like `interface/parse/udl`.
* Move the syn-based parsing impls into a submodule like `interface/parse/syn`.

This would help ensure that you can look at just one piece of the parsing logic and a time, and also make it easier to eventually delete the deprecated UDL syntax.

docs/manual/src/interface/errors.md
```diff
@@ -0,0 +1,40 @@
...
+}
+```
+
+Note that you cannot currently use a typedef for the `Result` type, UniFFI only supports
+the full `Result<T, E>` syntax. This limitation may be removed in future.
+
+The resulting `add` function would be exposed to the foreign-language code using
+its native error-handling mechanism. For example:
+
+* In Kotlin, there would be an `ArithmeticErrorException` class with an inner class
```
tarikeshaq, line 37: With the https://github.com/mozilla/uniffi-rs/pull/489

docs/manual/src/interface/objects.md
```diff
@@ -0,0 +1,281 @@
...
+## Managing Shared References
+
+To the foreign-language consumer, UniFFI object instances are designed to behave as much like
+regular language objects as possible. They can be freely passed as arguments or returned as values,
+like this Kotlin API:
+
+```kotlin
+interface TodoListInterface {
+    // Copy the items from another TodoList into this one.
+    // TODO: hrm, actually we need to pass a concrete `TodoList` instance here,
```
lougeniaC64, line 106: I'm a bit confused about this comment. I assumed that in the swift example above, it is recommended that the protocol (`TodoListProtocol`) be passed to a function instead of the concretion (`TodoList`) because it allows for increased flexibility so I'm not sure why it is a bad idea here.

uniffi_macros/src/lib.rs
```diff
@@ -112,3 +119,100 @@ pub fn include_scaffolding(component_name: proc_macro::TokenStream) -> proc_macr
...
+        .map(|name| {
+            let name = syn::Ident::new(name, module.span());
+            quote! {
+                pub use #namespace::#name;
+            }
+        })
+        .collect::<Vec<_>>();
+    proc_macro::TokenStream::from(quote! {
+        #module
+        #(#imports)*
```
tarikeshaq, line 216: For my own understanding, any reason we're exposing those functions at the top level? Is it for other rust consumers, in which case wouldn't it be better to just keep them inside the module where consumers can see them in the code file?

uniffi_bindgen/src/interface/function.rs
```diff
@@ -141,8 +146,50 @@ impl APIConverter<Function> for weedle::namespace::OperationNamespaceMember<'_>
...
+            syn::ReturnType::Default => None,
+            syn::ReturnType::Type(_, type_) => Some({
+                let (throws, returns) = super::synner::destructure_if_result_type(type_)?;
+                match (throws, returns) {
+                    (None, t) => t,
+                    (Some(err), t) => {
+                        attributes.push(super::attributes::Attribute::Throws(err));
+                        t
+                    }
+                }
```
tarikeshaq, line 173: nit:
```suggestion
                if let Some(err) = throws {
                    attributes.push(super::attributes::Attribute::Throws(err));
                }
                returns
```

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -132,6 +129,69 @@ impl<'ci> ComponentInterface {
...
+        }
+        if interface_modules.len() > 1 {
+            bail!("Multiple uses of `uniffi::declare_interface` found in Rust source file")
+        }
+        Self::from_rust_module(interface_modules[0])
+    }
+
+    /// Parse a `ComponentInterface` from a string containing a pre-parsed Rust module,
+    /// of the sort you might encounter if you're a macro.
+    pub fn from_rust_module(rsmod: &syn::ItemMod) -> Result<Self> {
```
tarikeshaq, line 175: Wow, I am very pleasantly surprised how elegant this function ended up being 🎉

uniffi_bindgen/src/lib.rs
```diff
@@ -176,16 +185,23 @@ pub fn generate_bindings<P: AsRef<Path>>(
...
     let cdylib_dir = cdylib_dir.as_ref();
-    let udl_file = udl_file.as_ref();
+    let interface_file = interface_file.as_ref();
     let config_file_override = config_file_override.as_ref().map(|p| p.as_ref());
+    println!("RUNNING TESTS {:?}", interface_file);
 
-    let component = parse_udl(udl_file)?;
-    let config = get_config(&component, udl_file, config_file_override)?;
+    let component = if matches!(interface_file.extension(), Some(ext) if ext == "rs") {
+        println!("FROM RUST");
```
tarikeshaq, line 198: 😄

tarikeshaq (commented):
OK finally was able to do a full go-through of this PR, I'm very happy with how we were able to almost plug-in into the existing patterns that we used for the udl file 💯 Excited to see where this goes

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -654,6 +771,108 @@ impl APIBuilder for weedle::Definition<'_> {
...
+        let attrs = synner::Attributes::try_from(&self.attrs)?;
+        ci.docs.extend(attrs.docs);
+        for item in &self.items {
+            item.process(ci)?
+        }
+        Ok(())
+    }
+}
+
+impl APIBuilder for &syn::ItemMod {
```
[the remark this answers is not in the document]
rfk: The reason is almost certainly "the compiler was complaining at me and I randomly added/removed ampersands until it stopped"

docs/manual/src/interface/functions.md
```diff
@@ -0,0 +1,70 @@
...
+    fn hello_world() -> MyString {
+        "Hello World!".to_owned()
+    }
+
+    // Error: UniFFI doesn't know about the `NotPartOfTheInterface` type.
+    fn example(v: NotPartOfTheInterface) {
+        println!("value: {}", v.value);
+    }
+
+    // Error: UniFFI doesn't understand the `std::vec::` path; just use `Vec<T>`.
```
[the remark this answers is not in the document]
rfk: Yeah, it's tricky. The Rust language has so much complexity/power that I think it's hard to inhabit a middle ground where you parse the Rust code independently of the compiler and correctly infer its semantics. In order to be safe, you need to either:

* Let the Rust compiler actually compile the code and output the information you want as a side-effect. This is how `wasm-bindgen` works for example, it makes Rust actually build the code and then arranges for the resulting library to have an extra data section with details about the types it encountered.
* Parse a subset of Rust that is so restricted that you can't possibly misinterpret what it means. That's what I was going for here. In fact, I think if we want to do this for real, we might even want to forcibly inject e.g. a `use std::vec::Vec` into the code so that we know that when we see a type named `Vec`, it is 100% definitely the `Vec` from the stdlib and some smart-alec hasn't done `use std::option::Option as Vec` at the top of the file for a laugh.

docs/manual/src/interface/objects.md
rfk: What I really want here is a `Self` type parameter. If you have some Swift-level mock implementation of the `TodoListInterface` then you absolutely cannot pass that in to the Rust code and expect it to work out well. The Rust code will only accept instances of the concrete `TodoList` type.

docs/manual/src/Getting_started.md
```diff
@@ -9,4 +9,4 @@ fn add(a: u32, b: u32) -> u32 {
 ```
 
 And top brass would like you to expose this *business-critical* operation to Kotlin and Swift.  
-**Don't panic!** We will show you how do that using UniFFI.
+**Don't panic!** We will show you how to do that using uniffi.
```
mhammond, line 12: Why the case change of `UniFFI`?

rfk: oh no 😱 

This is probably a rebase artifact, my original macros experiment predates my attempts to force everyone to capitalize UniFFI a certain way

docs/manual/src/interface/functions.md
```diff
@@ -0,0 +1,70 @@
...
+## Optional arguments & default values
+
+Function arguments can be marked as optional with a default value specified.
+TODO: what will the Rust syntax for this be?
+
+The Rust code will declare this using a macro annotation:
+
+```rust
+#[uniffi::defaults(name="World")] // TODO: not even sure what syntax is possible here...
+fn hello_name(name: String) -> String {
```
mhammond, line 56: Probably not a great suggestion, but we might consider dropping this as a feature? Even in today's UDL I get a bit confused with the `optional type?` thing and it never does quite what I expect on the rust side. For example, let's say a new binding comes along which doesn't support this (eg, even the "other side" of Rust), it would leave the authors in a bit of a bind.

docs/manual/src/interface/namespace.md
```diff
@@ -0,0 +1,19 @@
...
+```rust
+#[uniffi::declare_interface]
+mod example_namespace {
+  // The resulting interface definition will have a namespace
+  // of "example_namespace".
+}
+```
+
+The namespace servces multiple purposes:
+- It defines the default package/module/etc namespace in generated foreign-language bindings.
```
mhammond, line 15: s/namespace/name/ in this line?

docs/manual/src/interface/namespace.md
```diff
@@ -0,0 +1,19 @@
...
+- It is used to ensure uniqueness of names in the generated C FFI.
```
mhammond, line 16: I initially thought "what does this mean?" then I saw the following sentence - consider combining that sentence into this one?

docs/manual/src/interface/objects.md
```diff
@@ -0,0 +1,281 @@
+# Objects
+
+Objects are represented in the Rust world as a struct with an `impl` block containing methods.
+In the foreign-language bindings they would typically correspond to a class.
+
+In the interface definition, objects may have public methods but *must not have public fields*,
+because they are intended to be opaque to the foreign-language code. Like this:
```
mhammond, line 7: I initially wrote "I don't understand the justification for this limitation", but down in record.md I see that the public fields would make uniffi treat this as a record - which makes far more sense to me. Maybe update this para to reflect that?

docs/manual/src/interface/objects.md
```diff
@@ -0,0 +1,281 @@
...
+
+```rust
+#[uniffi::declare_interface]
+mod example {
+    struct TodoList {
```
mhammond, line 12: This might make the docs more unwieldy than you like, but to help with the mental model, maybe consider:

* Having the `struct` definition outside of the module, which a comment explaining it need not be inside.
* Another `impl TodoList` block outside the module, explaining that these can be thought of as "uniffi-private"

(edit: now that I get down to record.md I'm not actually sure about this)

docs/manual/src/interface/objects.md
mhammond: I'm inclined to think that this "shared references" section could move to a more "advanced" section - this also confused me, and the `Arc<>` discussion still made me stop and think, so someone new to UniFFI might be best served by not encountering this here.

(OTOH, the following "concurrent access" section did feel OK in this doc, mainly because it's something you really can't ignore - whereas this section *can* be ignored until you actually need it?

docs/manual/src/interface/records.md
```diff
@@ -1,54 +1,36 @@
...
+In the interface definition, they are distinguished from Object structs by the presence of
+public fields:
+
+```rust
+#[uniffi::declare_interface]
+mod todolist {
+    struct TodoEntry {
+        pub done: bool,
+        pub due_date: u64,
+        pub text: String,
```
mhammond, line 15: what happens to non-pub fields?

docs/manual/src/internals/object_references.md
```diff
@@ -32,8 +32,8 @@ this is so.
 ## Lifetimes
 
 In order to allow for instances to be used as flexibly as possible from foreign-language code,
```
mhammond, line 34: This duplicates some of the docs above - I'd suggest maybe merging the above with this - ie, this can be the "advanced" section I referred to.

mhammond (commented):
I've only got to the docs so far, but took some notes on the way. Feel free to reject any suggestions - I thought I'd leave even the dubious ones as they accurately reflect my first reading of those docs, but I understand it might not be at all clear how to resolve.

docs/manual/src/interface/functions.md
badboy: Dropping the feature of specifying default values?
Glean would be very sad about that, we rely on default values in our Kotlin/Swift/Python APIs.
Rust already doesn't have default values and that's a bit annoying.
It's worth making it very clear that this might not be supported everywhere.

mhammond: Fair enough - I agree they are useful in languages that support them, and I have no interest in making glean sad, so 👍 . I'll leave this conversation unresolved for now so Ryan can decide whether to add a note or not about this.

badboy:
Today I took this branch for a test run against my minimal Glean+UniFFI fork to see how far I get.
The Glean UDL is here: https://github.com/badboy/glean/blob/13e765da0fccdfb376c995888b4fdc123419f6c1/glean-core/uniffi/src/glean_core.udl

One big issue with the current implmementation is a bug, see this commit: https://github.com/mozilla/uniffi-rs/commit/9a23722f6d5f6db095e0ae466ca104b5b27f6939
tl;dr: It doesn't handle objects yet properly and generates broken code.

I have some more comments:

I don't think the auto-insertion of `pub use $namespace::$thing` is a good idea. See this commit: https://github.com/mozilla/uniffi-rs/commit/87448dab9f450564ab3ded40f14c3ed29fb1207f

In the current implementation a lot of errors are hidden. That includes some compile errors, import errors, etc., because the macro happens before the Rust compiler sees the code and can compile it.
I had to remove the `uniffi::declare_interface` a couple of times to fix compile errors first, then add it back and fix the remaining uniffi errors.

Cramming everything of the exposed API into a single module is ... not good.
Glean's API is huge (General API + 15+ metric types, each with 3-5 methods).
It's unwieldy to keep all these structs, functions _and the full function implementations_ in a single file.
If it's kept like that we might resort to hacks like concatenating files ahead of the build to keep proper code organization possible.

---

Initially I thought the macro approach can work really well and with this PR up I thought we might be close to a working implementation we could ship.
After spending this morning on it I'm not so convinced anymore.
Some of this boils down to needing an improved implementation (bug fixes + better error reporting). And that's just engineering effort.

But IMO right now the requirement of having it all in one module is a showstopper for me and I don't have a good idea how we can work around that.
Annotate each `enum`, `struct` and `impl` with `#[uniffi::interface]`?
`cargo expand` the whole codebase, then parse that to find the right things to uniffi-export?
(I now understand why `wasm-bindgen` does it that way)

uniffi_macros/src/lib.rs
rfk: In my head, I really want the UniFFI component interface to *also* be the public interface exposed by the crate, which is why I'm lifting them to the top-level here. But I agree that's probably weird/confusing for Rust consumers since the names magically appear in a scope other than where they're defined.

rfk:
> But IMO right now the requirement of having it all in one module is a showstopper for me
> and I don't have a good idea how we can work around that.

That's fair. One other data point on this front is the fxa-client crate, which is the crate for which I was using this little experiment in order to autogenerate the UDL from the Rust code. It's a bit different syntax but I was actually parsing the top-level `lib.rs` to determine the interface:

* https://github.com/mozilla/application-services/blob/main/components/fxa-client/src/lib.rs

In that case, most of the functions/methods are just little stub wrappers around implementations that live elsewhere in the crate. This helped keep the code organization under control while still exposing all the function signatures in the one Rust file, albeit at the cost of some duplication of the function signatures.

It's probably not a pattern we'd want to force on all consumers, though.

badboy:
This is now fully superseded by the existing macro approach.

Issue #417, opened 2021-03-16
Allow [Threadsafe] interfaces access to `Arc<Self>`


mhammond:
We've a use-case where it would be ideal to have access to a clone of the `Arc<Self>` stored in the `ArcHandleMap`.

The motivation here is for an app-services component to register with the "sync manager". Without going into too much detail about the "spelling", think something like:

```
[Threadsafe]
interface Foo {
  [SomethingSomething]
  void register_with_something_external();
}
```
makes sense, where we call something with the signature:

`fn register_with_something_external(self: Arc<Self>) { ... }`

which leverages ["method receivers"](https://github.com/rust-lang/rust/pull/56805). The intent is that we can avoid something gross like an `<Arc<Arc<T>>` and instead leverage the fact that uniffi already manages the `Arc` we need.

I'm opening this as a place-holder issue, trying to avoid too much bikeshedding at this stage (hence `[SomethingSomething]` :) - but I welcome it in general.

cc @rfk @lougeniaC64 @jhugman @dmose

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2107)

rfk:
> The intent is that we can avoid something gross like an <Arc<Arc<T>> and instead leverage the fact
> that uniffi already manages the Arc we need.

I haven't thought about this very deeply, but high-level, I'm concerned about leaking an incidental implementation of uniffi in this way. We *happen* to currently use an `Arc<T>` and a handlemap for `[Threadsafe]` interfaces, but we might want to try changing to something like a `Box<T>` in future for performance reasons.

Again without thinking very deeply, I wonder if it would be possible to invert the responsibility here, and have a way for you to declare that your interface is already inside an `Arc<T>` and you don't need uniffi to add a second one on for you. That would solve the double-arc problem, but I guess it wouldn't solve the "have to manually wrap your object in an Arc rather than having uniffi do it for you" problem...

mhammond:
> I'm concerned about leaking an incidental implementation of uniffi in this way.

Yes, good point. Off the top of my head, maybe unifying (ha!) what constructors return and what I'm trying to achieve here - ie, defining what `Self` is? That might be somewhat orthogonal to `ThreadSafe` bug might scale better...

rfk:
> I'm concerned about leaking an incidental implementation of uniffi in this way.

[...]

PR #418, opened 2021-03-16
A strawfox for #417 to allow an interface to have a method which take…

mhammond:
…s an `Arc::clone(self)`

[Had to do a last-minute rebase based on James's recent improvements and hope I didn't do anything even dumber there :]

Intent is a new attribute is added to such methods (and the method must be on
a `[ThreadSafe]` interface). I've chosen `[SomethingSomethingArc]` as the
attribute name :)

I attempted to change things so handle_map.rs no longer does the
ref-deref `&*` magic but instead delegates that to the macros, which
only does it if the new attribute exists. This *nearly* works, but
has unintended consequences I don't yet understand (ie, the
non-threadsafe generated code is using `&*obj` which makes no sense
at this hour).

### DIFF
--- a/examples/threadsafe/src/lib.rs
+++ b/examples/threadsafe/src/lib.rs
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use std::sync::atomic::{AtomicBool, AtomicI32, Ordering};
+use std::sync::Arc;
 
 /// Simulation of a task doing something to keep a thread busy.
 /// Up until now, everything has been synchronous and blocking, so
@@ -53,6 +54,7 @@ impl Counter {
 // It relies on its own locking mechanisms, and uses the `Threadsafe`
 // attribute to tell uniffi to use the `ArcHandleMap`, so avoid the default
 // locking strategy.
+#[derive(Debug)]
 struct ThreadsafeCounter {
     is_busy: AtomicBool,
     count: AtomicI32,
@@ -61,6 +63,7 @@ struct ThreadsafeCounter {
 // Interface structs labelled `Threadsafe` should (safely) implement
 // `Send` and `Sync`.
 static_assertions::assert_impl_all!(ThreadsafeCounter: Send, Sync);
+
 
 impl ThreadsafeCounter {
     fn new() -> Self {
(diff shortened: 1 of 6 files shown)


examples/threadsafe/src/lib.rs
```diff
@@ -83,6 +86,11 @@ impl ThreadsafeCounter {
             self.count.load(Ordering::SeqCst)
         }
     }
+
+    fn cloneable(self: Arc<Self>) {
```
mhammond, line 90: to be clear, I think we want a new example rather than abusing this, but this is OK for demo purposes

uniffi_bindgen/src/templates/macros.rs
```diff
@@ -66,12 +66,20 @@ use uniffi::UniffiMethodCall;
 {% match meth.throws() -%}
 {% when Some with (e) -%}
 {{ this_handle_map }}.method_call_with_result(err, {{ meth.first_argument().name() }}, |obj| -> Result<{% call return_type_func(meth) %}, {{e}}> {
+    {%- if meth.something_something_arc() -%}
     let _retval = {{ obj.name() }}::{%- call to_rs_call_with_prefix("obj", meth) -%}?;
+    {%- else -%}
+    let _retval = {{ obj.name() }}::{%- call to_rs_call_with_prefix("&*obj", meth) -%}?;
+    {%- endif -%}
```
jhugman, line 73: The `method_call_with_result` is doing the work of choosing between the two `handle_map` implementations, so this `else` block is being seen by both `Threadsafe` and non-`Threadsafe` objects.

uniffi_bindgen/src/interface/attributes.rs
```diff
@@ -23,12 +23,13 @@ use anyhow::{bail, Result};
...
 /// any unsupported ones. These don't convert directly into parts of a `ComponentInterface`, but
 /// may influence the properties of things like functions and arguments.
-#[derive(Debug, Clone, Hash)]
+#[derive(Debug, Clone, Hash, PartialEq)]
 pub(super) enum Attribute {
     ByRef,
     Error,
     Threadsafe,
     Throws(String),
+    SomethingSomethingArc,
```
jhugman, line 32: I don't really understand the need for this, so I'm not going to much help naming this.

jhugman (commented):
Quick drive by. This looks straightforward enough. I'm sorry I'm not able to suggest any better user-facing names.

> (i.e. the non-threadsafe generated code is using &*obj which makes no sense
> at this hour).

I left a reason why this happening, but not a solution. :(

mhammond:
I think #419 has longer legs than this, so I don't plan on taking this further.

Issue #419, opened 2021-03-17
[discuss] A proposal for passing around object instances


rfk:
Riffing on some conversations today with @mhammond, here are some partly-formed thoughts on how we might work towards enabling object instances to be passed as arguments and returned from functions/methods, resolving #40 and #197.

Broadly, we have two big complexities to grapple with here - the way that handlemaps make lifting/lowering codegen for object instances difficult, and the management of lifetimes and object identity across the ffi boundary. I think we may be able to resolve both of these with some judicious use of `Arc`.

### Step 1: Remove Handlemaps

First, we replace our current use of handlemaps with `Arc`. I expect this will be more performant, but really I'm interested in simplifying codegen in the bindings.

For a threadsafe interface `T`, we put it in an `Arc<T>` just like we do today. But instead of putting the `Arc<T>` into a handlemap to turn it into an integer handle, we use [`Arc::into_raw`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.into_raw) to turn it directly into a pointer, and pass the pointer over the ffi to the foreign-language code. When the foreign-language code wants to call a method on the instance, it passes the pointer back to the Rust code and we use [`Arc::from_raw`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.from_raw) to reconstitute the `Arc`, and can use it like we currently do with the one in the handlemap. (We have to be careful about not dropping the `Arc` after reconstituting it, but this seems tractable).

For a non-threadsafe interface `T`, we first make it synchronous by putting it in a `Mutex<T>`, then putting that inside an `Arc<Mutex<T>>`. We can then treat it the same way as for threadsafe instances above, converting to and from a raw pointer to pass over the FFI.

I need to dig more into the memory-safety and thread-safety requirements of the above, but it seems like it should be tractable.

The aim of this is to remove the use of handlemap closures in order to get object references in Rust. Rather than generating code like:

```
HANDLE_MAP.call_with_result(obj_handle, |obj| -> obj.method())
```

We'd generate code like:

```
let obj = unsafe { Arc::from_raw(obj_ptr) } // TODO: some manually-drop stuff in here somewhere probably
let obj = obj.clone()  // get our own copy of the Arc, to make sure it stays alive
let mut obj = obj.lock() // if we're using a Mutex, we can lock it here 
obj.method()
```

[...]

PR #420, opened 2021-03-30
POC for step 1 of #419 - replace Handlemaps with Arcs.

mhammond:
The context for this is Ryan's [proposal for passing around object instances]( https://github.com/mozilla/uniffi-rs/issues/419) - which has steps 1 and 2 in a rough way - replacing Handlemaps with `Arc<>` (or to-be-deprecated `Arc<Mutex<>>`) and allowing them to be args and results.

This is turning out better than I hoped - for example, [see this new test](https://github.com/mhammond/uniffi-rs/blob/419-step1/fixtures/coverall/src/lib.rs#L135-L151) which exercises most of what this does.

I think the Kotlin generation works, but the Swift generation most certainly does not. @jhugman now you are back from PTO I'd love your feedback and to gauge your interest in collaborating on this? Note that my intent is to iterate slowly, just in time for Ryan to get back and review.

### DIFF
--- a/fixtures/coverall/src/lib.rs
+++ b/fixtures/coverall/src/lib.rs
@@ -34,6 +34,7 @@ pub struct SimpleDict {
     maybe_float32: Option<f32>,
     float64: f64,
     maybe_float64: Option<f64>,
+    coveralls: Option<Arc<Coveralls>>,
 }
 
 fn create_some_dict() -> SimpleDict {
@@ -54,6 +55,7 @@ fn create_some_dict() -> SimpleDict {
         maybe_float32: Some(22.0 / 7.0),
         float64: 0.0,
         maybe_float64: Some(1.0),
+        coveralls: Some(Arc::new(Coveralls::new("some_dict".to_string()))),
     }
 }
 
@@ -75,6 +77,7 @@ fn create_none_dict() -> SimpleDict {
         maybe_float32: None,
         float64: 0.0,
         maybe_float64: None,
+        coveralls: None,
     }
 }
 
@@ -129,12 +132,18 @@ impl Coveralls {
         panic!("{}", message);
     }
 
-    fn take_other(self: Arc<Self>) {
-        *self.other.lock().unwrap() = Some(Arc::clone(&self));
+    fn take_other(&self, other: Option<Arc<Self>>) {
+        *self.other.lock().unwrap() = match other {
+            None => None,
+            Some(arc) => Some(Arc::clone(&arc)),
+        }
     }
 
-    fn drop_other(&self) {
-        *self.other.lock().unwrap() = None;
+    fn get_other(&self) -> Option<Arc<Self>> {
+        match &*self.other.lock().unwrap() {
+            Some(other) => Some(Arc::clone(other)),
+            None => None,
+        }
     }
 
     fn strong_count(self: Arc<Self>) -> u64 {
@@ -148,6 +157,16 @@ impl Coveralls {
     fn take_other_panic(self: Arc<Self>, message: String) -> () {
         panic!("{}", message);
     }
+
+    fn clone_me(&self) -> Arc<Self> {
+        let other = self.other.lock().unwrap();
+        let new_other = Mutex::new(other.clone());
+        *NUM_ALIVE.write().unwrap() += 1;
+        Arc::new(Self {
+            name: self.name.clone(),
+            other: new_other,
+        })
+    }
 }
 
 impl Drop for Coveralls {
@@ -155,4 +174,27 @@ impl Drop for Coveralls {
         *NUM_ALIVE.write().unwrap() -= 1;
     }
 }
+
+#[derive(Debug, Clone, Copy)]
+enum Color {
+    Red,
+    Blue,
+    Green,
+}
+
+#[derive(Debug)]
+struct Patch {
+    color: Color,
+}
+
+impl Patch {
+    fn new(color: Color) -> Self {
+        Self { color }
+    }
+
+    fn get_color(&self) -> Color {
+        self.color
+    }
+}
+
 include!(concat!(env!("OUT_DIR"), "/coverall.uniffi.rs"));
--- a/fixtures/coverall/src/lib.rs
+++ b/fixtures/coverall/src/lib.rs
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use std::sync::RwLock;
+use std::sync::{Arc, Mutex, RwLock};
 
 lazy_static::lazy_static! {
     static ref NUM_ALIVE: RwLock<u64> = {
@@ -87,12 +87,18 @@ type Result<T, E = CoverallError> = std::result::Result<T, E>;
 #[derive(Debug)]
 pub struct Coveralls {
     name: String,
+    // A reference to another Coveralls. Currently will be only a reference
+    // to `self`, so will create a circular reference.
+    other: Mutex<Option<Arc<Self>>>,
 }
 
 impl Coveralls {
     fn new(name: String) -> Self {
         *NUM_ALIVE.write().unwrap() += 1;
-        Self { name }
+        Self {
+            name,
+            other: Mutex::new(None),
+        }
     }
 
     fn fallible_new(name: String, should_fail: bool) -> Result<Self> {
@@ -120,6 +126,26 @@ impl Coveralls {
     }
 
     fn panic(&self, message: String) {
+        panic!("{}", message);
+    }
+
+    fn take_other(self: Arc<Self>) {
+        *self.other.lock().unwrap() = Some(Arc::clone(&self));
+    }
+
+    fn drop_other(&self) {
+        *self.other.lock().unwrap() = None;
+    }
+
+    fn strong_count(self: Arc<Self>) -> u64 {
+        Arc::strong_count(&self) as u64
+    }
+
+    fn take_other_fallible(self: Arc<Self>) -> Result<()> {
+        Err(CoverallError::TooManyHoles)
+    }
+
+    fn take_other_panic(self: Arc<Self>, message: String) -> () {
         panic!("{}", message);
     }
 }
--- a/fixtures/coverall/tests/bindings/test_coverall.py
+++ b/fixtures/coverall/tests/bindings/test_coverall.py
@@ -103,6 +103,8 @@ class TestCoverall(unittest.TestCase):
         self.assertEqual(get_num_alive(), 1)
         # and check it's the correct object.
         self.assertEqual(coveralls.get_other().get_name(), "test_arcs")
+        # even better - it should be the same Python object
+        self.assertTrue(coveralls.get_other() is coveralls)
 
         with self.assertRaises(CoverallError.TooManyHoles):
             coveralls.take_other_fallible()
@@ -129,6 +131,9 @@ class TestCoverall(unittest.TestCase):
         self.assertEqual(get_num_alive(), 2)
         self.assertEqual(coveralls.strong_count(), 2)
         self.assertEqual(c2.strong_count(), 3)
+
+        # and get_other() should return the exact same instance.
+        self.assertTrue(coveralls.get_other() is c2)
 
         # We can drop Python's reference to `c2`, but the rust struct will not
         # be dropped as coveralls hold an `Arc<>` to it.
--- a/fixtures/coverall/tests/bindings/test_coverall.py
+++ b/fixtures/coverall/tests/bindings/test_coverall.py
@@ -6,6 +6,11 @@ import unittest
 from coverall import *
 
 class TestCoverall(unittest.TestCase):
+    # Any test not terminating with zero objects alive will cause others to
+    # fail - this helps us work out which test kept things alive.
+    def tearDown(self):
+        self.assertEqual(get_num_alive(), 0)
+
     def test_some_dict(self):
         d = create_some_dict()
         self.assertEqual(d.text, "text")
@@ -83,6 +88,27 @@ class TestCoverall(unittest.TestCase):
 
         with self.assertRaisesRegex(InternalError, "expected panic: oh no"):
             coveralls.panic("expected panic: oh no")
+
+    def test_arcs(self):
+        coveralls = Coveralls("test_arcs")
+        self.assertEqual(get_num_alive(), 1)
+        self.assertEqual(coveralls.strong_count(), 2)
+        coveralls.take_other()
+        # should now be a new strong ref.
+        self.assertEqual(coveralls.strong_count(), 3)
+        # but the same number of instances.
+        self.assertEqual(get_num_alive(), 1)
+
+        with self.assertRaises(CoverallError.TooManyHoles):
+            coveralls.take_other_fallible()
+
+        with self.assertRaisesRegex(InternalError, "expected panic: with an arc!"):
+            coveralls.take_other_panic("expected panic: with an arc!")
+
+        coveralls.drop_other()
+        self.assertEqual(coveralls.strong_count(), 2)
+        coveralls = None
+        self.assertEqual(get_num_alive(), 0)
 
 if __name__=='__main__':
     unittest.main()
--- a/fixtures/coverall/tests/bindings/test_coverall.py
+++ b/fixtures/coverall/tests/bindings/test_coverall.py
@@ -25,6 +25,7 @@ class TestCoverall(unittest.TestCase):
         self.assertEqual(d.maybe_signed8, 0)
         self.assertEqual(d.signed64, 9223372036854775807)
         self.assertEqual(d.maybe_signed64, 0)
+        self.assertEqual(d.coveralls.get_name(), "some_dict")
 
         # floats should be "close enough" - although it's mildly surprising that
         # we need to specify `places=6` whereas the default is 7.
@@ -52,6 +53,7 @@ class TestCoverall(unittest.TestCase):
         self.assertIsNone(d.maybe_float32)
         self.assertAlmostEqual(d.float64, 0.0)
         self.assertIsNone(d.maybe_float64)
+        self.assertIsNone(d.coveralls)
 
     def test_constructors(self):
         self.assertEqual(get_num_alive(), 0)
@@ -72,7 +74,7 @@ class TestCoverall(unittest.TestCase):
         with self.assertRaisesRegex(InternalError, "expected panic: woe is me"):
             Coveralls.panicing_new("expected panic: woe is me")
 
-        # in the absence of cycles Python is deterministic killing refs
+        # in the absence of cycles Python is deterministic in killing refs
         coveralls2 = None
         self.assertEqual(get_num_alive(), 1)
         coveralls = None
@@ -93,11 +95,14 @@ class TestCoverall(unittest.TestCase):
         coveralls = Coveralls("test_arcs")
         self.assertEqual(get_num_alive(), 1)
         self.assertEqual(coveralls.strong_count(), 2)
-        coveralls.take_other()
+        self.assertIsNone(coveralls.get_other())
+        coveralls.take_other(coveralls)
         # should now be a new strong ref.
         self.assertEqual(coveralls.strong_count(), 3)
         # but the same number of instances.
         self.assertEqual(get_num_alive(), 1)
+        # and check it's the correct object.
+        self.assertEqual(coveralls.get_other().get_name(), "test_arcs")
 
         with self.assertRaises(CoverallError.TooManyHoles):
             coveralls.take_other_fallible()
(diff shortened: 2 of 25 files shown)


uniffi_bindgen/src/interface/object.rs
```diff
@@ -232,6 +232,7 @@ impl Constructor {
         self.ffi_func.name.push_str("_");
         self.ffi_func.name.push_str(&self.name);
         self.ffi_func.arguments = self.arguments.iter().map(Into::into).collect();
+        // XXX - we need this to be some kind of pointer abstraction - `usize`?
```
rfk, line 235: If we're going to work with raw pointers, I think the right thing to do here is add an `FFIType::Pointer` variant and then map it explicitly to the corresponding pointer abstractions in each foreign language. (The existing `FFITypeRustCString` variant actually already is *sort* of this, but specifically for a pointer-to-char).

uniffi_bindgen/src/templates/ObjectTemplate.rs
```diff
@@ -1,36 +1,36 @@
 // For each Object definition, we assume the caller has provided an appropriately-shaped `struct`
-// with an `impl` for each method on the object. We create a `ConcurrentHandleMap` for safely handing
+// with an `impl` for each method on the object. We create an `Arc` in
+// `[Threadsafe]` interfaces or `Arc<Mutex<>>` otherwise, for for "safely" handing
```
rfk, line 3: "for for"; also lol at the scare-quotes around "safely" :-P

uniffi_bindgen/src/templates/ObjectTemplate.rs
```diff
@@ -1,36 +1,36 @@
...
-    pub extern "C" fn {{ ffi_free.name() }}(handle: u64) {
-        let _ = {{ handle_map }}.delete_u64(handle);
+{% let ffi_free = obj.ffi_object_free() -%}
+#[doc(hidden)]
+#[no_mangle]
+pub extern "C" fn {{ ffi_free.name() }}(ptr: u64) {
+    if let Err(e) = std::panic::catch_unwind(|| {
+        assert_ne!(ptr, 0);
+        // turn it into an Arc and let the value naturally drop.
+        unsafe { std::sync::Arc::from_raw(ptr as *const {{ obj.name() }}) };
```
rfk, line 22: As a purely stylistic nit, I think I'd do an explicit `drop()` here just to clearly communicate intent.

rfk: Also I wonder if this should be `*mut` rather than `*const` if we're going to drop it, and whether that actually matters in practice. It probably doesn't.

uniffi_bindgen/src/templates/macros.rs
```diff
@@ -6,9 +6,14 @@
...
 
-{%- macro to_rs_call_with_prefix(prefix, func) -%}
+{%- macro to_rs_call_with_prefix(arg_name, func, obj) -%}
     {{ func.name() }}(
-    {{- prefix }}{% if func.arguments().len() > 0 %}, {% call _arg_list_rs_call(func) -%}{% endif -%}
+    {% if obj.threadsafe() %}
+    &*{{- arg_name }}
+    {% else %}
+    &mut *{{- arg_name }}.lock().unwrap()
+    {% endif %}
```
rfk, line 15: This dereferencing/unwrapping logic...I wonder if it makes sense to move some or all of that into the definition of the `lift_rs` filter. It seems pretty analogous to me - this is the stuff that you have to do to go from the value that you get over the FFI, to the value that the Rust code needs to use at runtime.

uniffi_bindgen/src/templates/macros.rs
```diff
@@ -47,33 +52,63 @@
...
+        },
+        Err(e) => {
+            *err = e.into();
+            0 as usize /*`std::ptr::null()` is a compile error */
+        }
+    }
 {% endmatch %}
 {% endmacro %}
 
+{% macro get_arc(obj) -%}
```
rfk, line 81: nit: I think calling this `clone_arc` would be clearer about your intent.

uniffi_bindgen/src/templates/macros.rs
```diff
@@ -47,33 +52,63 @@
...
+    {% if obj.threadsafe() %}
+    let _arc = unsafe { std::sync::Arc::from_raw(ptr as *const {{ obj.name() }}) };
+    {% else %}
+    let _arc = unsafe { std::sync::Arc::from_raw(ptr as *const std::sync::Mutex<{{ obj.name() }}>) };
+    {% endif %}
+    // This arc now "owns" the reference but we need an outstanding reference still.
+    std::sync::Arc::into_raw(std::sync::Arc::clone(&_arc));
```
rfk, line 88: I suggest using [`ManuallyDrop`](https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html) rather than leaking it back via `into_raw`, as I think it better captures the intent. If I had to write this code out by hand with lots of comments it would look something like:

```
// The pointed-to Arc is owned by the foreign-language code, so we mustn't drop it.
let _foreign_arc = ManuallyDrop::new(unsafe { Arc::from_raw(ptr as *const {{ obj.name() }}) });
// We want an owned copy of the Arc to pass internally to the Rust code.
let _arc = Arc::clone(&*_foreign_arc);

```

uniffi_bindgen/src/templates/macros.rs
```diff
@@ -47,33 +52,63 @@
...
 {% match meth.throws() -%}
 {% when Some with (e) -%}
-{{ this_handle_map }}.method_call_with_result(err, {{ meth.first_argument().name() }}, |obj| -> Result<{% call return_type_func(meth) %}, {{e}}> {
-    let _retval = {{ obj.name() }}::{%- call to_rs_call_with_prefix("obj", meth) -%}?;
-    Ok({% call ret(meth) %})
-})
+    uniffi::deps::ffi_support::call_with_result(
+        err,
+        || -> Result<_, {{ e }}> {
+            {% call get_arc(obj) %}
```
rfk, line 98: Instead of using a macro for this, what if this was done via the existing "lift" mechanics, something like:

```
let _arc = {{meth.first_argument().name()|lift_rs(meth.first_argument().type_()) }}
```

Broadly, I think the more we can treat these object-typed arguments in the same way as we treat other arguments, the better.

uniffi_bindgen/src/templates/macros.rs
```diff
@@ -47,33 +52,63 @@
...
-    {{ obj.name() }}::{% call to_rs_call(cons) %}
-})
+    match std::panic::catch_unwind(|| {
+        {%- if obj.threadsafe() %}
+        let _new = {{ obj.name() }}::{% call to_rs_call(cons) %};
+        {%- else %}
+        let _new = std::sync::Mutex::new({{ obj.name() }}::{% call to_rs_call(cons) %});
+        {%- endif %}
+        let _arc = std::sync::Arc::new(_new);
+        std::sync::Arc::into_raw(_arc)
```
rfk, line 67: (I know this isn't intended for merge, but I'm commenting to share my mental model of how the final code might look).

This needs a comment along the lines of "This copy of the Arc is leaked, and is conceptually now owned by the foreign-language code. It will only be dropped when the foreign-language code calls the matching desctructor".

uniffi_bindgen/src/templates/ObjectTemplate.rs
```diff
@@ -1,36 +1,36 @@
...
 
-    {% let ffi_free = obj.ffi_object_free() -%}
-    #[doc(hidden)]
-    #[no_mangle]
-    pub extern "C" fn {{ ffi_free.name() }}(handle: u64) {
-        let _ = {{ handle_map }}.delete_u64(handle);
+{% let ffi_free = obj.ffi_object_free() -%}
+#[doc(hidden)]
+#[no_mangle]
+pub extern "C" fn {{ ffi_free.name() }}(ptr: u64) {
```
rfk, line 18: Yeah, I really think we want to update the `FFIType` enum to let us use a pointer type in the argument here.

rfk (commented):
This looks pretty neat! I'm kind of strangely glad to see a SIGSEGV in the circle failures here, means we're definitely in raw pointer territory :-)

I'd like to consider whether we can embed the use of Arc here more thoroughly in the mechanics of how we work with object instances, so will share some rambling thoughts about how that might look architecturally.

First, observe that when we generate the [support code for records](https://github.com/mozilla/uniffi-rs/blob/67fcfaf4d1bc1e6aad9ed814846a042fac8eeccb/uniffi_bindgen/src/templates/RecordTemplate.rs#L8), we generate an implementation of the `ViaFfi` trait that knows how to read/write then in a buffer.  We currently do not implement `ViaFfi` for object types at all.

What if we did update [`ObjectTemplate.rs`](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/templates/ObjectTemplate.rs) to produce a `ViaFfi` implementation for object types? It might look something like this:

```
// This won't work as written IIRC, because this code doesn't define `ViaFfi` and
// doesn't define `Arc`, so Rust's trait coherence rules will stop us from impling
// the trait on `Arc<ObjectName>`. But it's interesting to think about architecturally.
// Maybe we could define a newtype wrapper that would let this work?

unsafe impl uniffi::ViaFfi for Arc<ObjectName> {

    // When passing object instances over the FFI, we pass a pointer
    // that is the raw form of an Arc<ObjectName> owned by the foreign-language code.
    type FfiType = *const ObjectName;

    // If we return an `Arc<ObjectName>` to the foreign-language code, then it
    // owns the reference and is responsible for freeing it.
    fn lower(self) -> Self::FfiType {
        self.into_raw()
    }

    // If we want to receive an `Arc<ObjectName>` from the foreign-language code,
    // we have to make our own clone of it.
    fn try_lift(v: Self::FfiType) -> Result<Self> {
        // The received pointer is an Arc owned by foreign-language code, don't drop it.
        let foreign_arc = ManuallyDrop::new(unsafe { Self::from_raw(v) });
        // Take a (cheap) clone for our own use.
        Ok(Arc::clone(foreign_arc))
    }

    // If we want to send a data structure with an `Arc<ObjectName>` in one of its
    // fields, we pass ownership of a clone by passing the pointer.
    fn write<B: bytes::BufMut>(&self, buf: &mut B) {
        let foreign_arc = Arc::clone(self);
        buf.put_u64(foreign_arc.into_raw() as u64) // Probably need some sizeof() checks here :-/
    }

    // If we receive a serialized data structure with an `Arc<ObjectName>` in one of its
    // fields, we need to take a clone for our own purposes.
    fn try_read<B: bytes::Buf>(buf: &mut B) -> Result<Self> {
      Self::try_lift(buf.get_u64() as const* ObjectName)
    }
}
```

I wonder if something like this would help make the generated code any cleaner?

rfk:
> // This won't work as written IIRC, because this code doesn't define `ViaFfi` and
> // doesn't define `Arc`, so Rust's trait coherence rules will stop us from impling
> // the trait on `Arc<ObjectName>`. But it's interesting to think about architecturally.
> // Maybe we could define a newtype wrapper that would let this work?

Actually, if this seems useful then we might be able to have uniffi provide a default `impl<T> ViaFfi for Arc<T>` in the same way that it provides one for `Vec<T>`.

mhammond:
Callbacks are broken - possibly for many reasons, but at least one is due to the constructors creating a `Arc<Box<T>>` but the dtor assuming a simple `Arc<T>`. It's not yet obvious to me why callbacks use a `Box<>` and whether an `Arc` might work ('cos if it did, I imagine the extra symmetry might be worthwhile).

I'm going to put this down for a few days, so let me know if anyone is interested in picking any of this up.

mhammond:
Now with an "identity map" for Python :) Assuming #431 and #429 aren't too controversial, once they land I'll rebase this to help make this simpler to review.

fixtures/coverall/src/lib.rs
```diff
@@ -122,11 +131,70 @@ impl Coveralls {
     fn panic(&self, message: String) {
         panic!("{}", message);
     }
+
+    fn take_other(&self, other: Option<Arc<Self>>) {
+        *self.other.lock().unwrap() = match other {
+            None => None,
+            Some(arc) => Some(Arc::clone(&arc)),
```
rfk, line 138: IIUC, we already own the `Arc<Self>` in `other` at this point, so I don't think we need to clone here, and could just `*self.other.lock().unwrap() = other`. The current setup will increment the refcount for the `Arc::clone`, then immediately decrement it when `other` is dropped at the end of the function.

fixtures/coverall/tests/bindings/test_coverall.py
```diff
@@ -6,6 +6,11 @@
 from coverall import *
 
 class TestCoverall(unittest.TestCase):
+    # Any test not terminating with zero objects alive will cause others to
+    # fail - this helps us work out which test kept things alive.
+    def tearDown(self):
+        self.assertEqual(get_num_alive(), 0)
```
rfk, line 12: 👍🏻 nice!

uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
```diff
@@ -14,17 +20,37 @@ def __del__(self):
...
     @classmethod
-    def {{ cons.name()|fn_name_py }}(cls, {% call py::arg_list_decl(cons) %}):
-        {%- call py::coerce_args_extra_indent(cons) %}
-        # Call the (fallible) function before creating any half-baked object instances.
-        handle = {% call py::to_ffi_call(cons) %}
+    def _get_or_make_instance_(cls, handle):
+        {# Look in our weak map for an instance already associated with this
+           pointer. If it exists we can return it, but the pointer we have
+           is a clone of an `Arc<>` so has a reference we need to destroy.
+        #}
```
rfk, line 30: I feel like we probably want a mutex or other concurrency-control here around the updates to the identity map.

uniffi/src/lib.rs
```diff
@@ -476,6 +475,74 @@ unsafe impl<V: ViaFfi> ViaFfi for HashMap<String, V> {
...
+/// by reference must be encapsulated in an `Arc`, and must be safe to share
+/// across threads.
+unsafe impl<T: Sync + Send> ViaFfi for std::sync::Arc<T> {
+    // Don't use a pointer to <T> as that requires a `pub <T>`
+    type FfiType = *const std::os::raw::c_void;
+
+    /// When lowering, we have an owned `Arc<T>` and we transfer that ownership
+    /// to the foreign-language code, "leaking" it out of Rust's ownership system
+    /// as a raw pointer. The foreign-language code is responsible for freeing
+    /// this via TODO calling a special destructor for `T`.
```
rfk, line 490: Oh yeah! We already have one of these in `obj.ffi_object_free().name()` and it just needs to be updated, so there's not really anything "TODO" about this except for document it here.

uniffi_bindgen/src/bindings/python/templates/RustBufferBuilder.py
```diff
@@ -108,10 +108,12 @@ def writeString(self, v):
 
     {% when Type::Object with (object_name) -%}
     # The Object type {{ object_name }}.
-    # Objects cannot currently be serialized, but we can produce a helpful error.
+    # We write the pointer value directly - what could possibly go wrong?
```
rfk, line 111: 🤞🏻😅🤞🏻

uniffi_bindgen/src/interface/attributes.rs
```diff
@@ -190,24 +196,29 @@ impl<T: TryInto<FunctionAttributes, Error = anyhow::Error>> TryFrom<Option<T>>
 
 /// Represents UDL attributes that might appear on a function argument.
 ///
-/// This supports the `[ByRef]` attribute for arguments that should be passed
-/// by reference in the generated Rust scaffolding.
+/// This supports the `[ByRef]` and `[ByArc]` attribute for arguments that
+/// should be passed by reference in the generated Rust scaffolding.
```
rfk, line 200: nit: the "passed by reference" part of this comment will need re-wording.

uniffi_bindgen/src/interface/attributes.rs
```diff
@@ -190,24 +196,29 @@ impl<T: TryInto<FunctionAttributes, Error = anyhow::Error>> TryFrom<Option<T>>
...
+        self.0.iter().any(|attr| matches!(attr, Attribute::ByArc))
+    }
 }
 
 impl TryFrom<&weedle::attribute::ExtendedAttributeList<'_>> for ArgumentAttributes {
     type Error = anyhow::Error;
     fn try_from(
         weedle_attributes: &weedle::attribute::ExtendedAttributeList<'_>,
     ) -> Result<Self, Self::Error> {
+        // XXX - how do we ensure both `ByRef` and `ByArc` aren't specified?
```
rfk, line 218: FWIW, the `InterfaceAttributes` struct has an extremely dumb check to ensure that `[Threadsafe]` and `[Enum]` are not applied to the same interface.

uniffi_bindgen/src/interface/function.rs
```diff
@@ -184,10 +188,12 @@ impl APIConverter<Argument> for weedle::argument::SingleArgument<'_> {
             Some(v) => Some(convert_default_value(&v.value, &type_)?),
         };
         let by_ref = ArgumentAttributes::try_from(self.attributes.as_ref())?.by_ref();
+        let by_arc = ArgumentAttributes::try_from(self.attributes.as_ref())?.by_arc();
```
rfk, line 191: I'm actually wondering if `[ByArc]` is really necessary on arguments, given that we've simplified the proposal down quite a bit. If you're passing something that's not an `Object` instance then `[ByArc]` is invalid. If you're passing something that is an `Object` instance then you *have* to be passing it in an `Arc`. So I'm not sure what value the annotation provides, except maybe for reminding the component author that they need to accept an `Arc<T>` instead of a `T`. (But even then, the Rust compiler will also remind them of this fact).

uniffi_bindgen/src/interface/object.rs
```diff
@@ -327,6 +332,10 @@ impl Method {
         self.attributes.get_throws_err()
     }
 
+    pub fn self_by_arc(&self) -> bool {
```
rfk, line 335: The parsing center of my brain wants the name of this to be worded as a predicate about the object, e.g. `takes_self_by_arc` or similar. But then I can't really articulate a cohesive style guide for the way things are named in here, so I'm just going to note it and move on.

rfk (commented):
I *really* like how this is coming together, thanks for pushing on it! In particular I'm really happy with how small the diff seems to be, given the conceptual impact of the change. (Other language bindings still to come, obviously, but still...).

I'll try to find some time to try my hand at the Kotlin bindings to help move this forward and shake out any other issues.

How should our work here relate to the ADR in #430? Right now I can convince myself that we're just prototyping to help inform a decision in that ADR, but if we keep going we'll eventually get to a point where we're actually assuming the outcome of that discussion.

mhammond:
> How should our work here relate to the ADR in #430?

Funny you should ask - I was just making note about that. When we chatted in person, I mentioned that I thought it would be possible to still use a `HandleMap` in this world (ie, we'd not strictly need #430) - we'd still use `Arc` as the values in the map, but could hand out handles to consumers. I think we were both in agreement that there's not a huge amount of value there, but left the door open.

Thinking more though, I think that would work OK right until the last commit in the set - the one that provides "identity" for Python. That relies on the exact same value for the same objects, and the existing handle maps do not offer that - it probably *could* with more complex hoops, but I don't think that's worthwhile. So assuming we want to lean in on offering identity for languages which make it possible, I'm inclined to suggest we use this justification to push forward with #430.

uniffi_bindgen/src/interface/function.rs
rfk: > If you're passing something that is an Object instance then you have to be passing it in an Arc.

Hrm, I guess this isn't necessarily true - the bindings could take an `&T` out of the `Arc<T>` and pass it to the hand-written Rust code. But we could consider having `f(T arg)` mean "receives an owned value in an Arc" and `f([ByRef] T arg)` mean "receives a borrowed reference", which is how it works for non-Object types.

rfk:
> So assuming we want to lean in on offering identity for languages which make it possible,
> I'm inclined to suggest we use this justification to push forward with #430.

Sold! I'll take another look at that ADR next, and if it's welcome, perhaps put up a PR to your PR with some more thoughts.

mhammond:
> Sold! I'll take another look at that ADR next, and if it's welcome, perhaps put up a PR to your PR with some more thoughts.

Obviously welcome! We could collaborate on the same branch, but I guess PRs on PRs gives us some kind of audit-trail for reviews - I review your PRs into my PR, you review my PR into main :)

uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
```diff
@@ -14,17 +20,37 @@ def __del__(self):
...
+        existing = cls._instances.get(handle)
+        if existing is not None:
+            rust_call_with_error(
+                InternalError,
+                _UniFFILib.{{ obj.ffi_object_free().name() }},
+                handle
+            )
```
rfk, line 37: Oh! I just understood what you're doing here, and it's neat.

In my head I was worried about the asymmetry where we return a fresh clone of the `Arc<T>` on the Rust side but collapse them all into a single instance on the foreign-language side. I thought we might have to keep track of the number of `Arc<T>` that had been merged into this instance, so that we could call `ffi_object_free` the appropriate number of times. But immediately destroying the clones is much simpler (and really reinforces the "The Prestige" metaphor that @jhugman used to describe this identity map idea to me at one point...).

In the future we might be motivated to keep a count and destroy them all at once for performance reasons, but that seems likely to be a long ways off.

fixtures/coverall/tests/bindings/test_coverall.py
```diff
@@ -84,5 +91,65 @@ def test_errors(self):
         with self.assertRaisesRegex(InternalError, "expected panic: oh no"):
             coveralls.panic("expected panic: oh no")
 
+    def test_arcs(self):
+        coveralls = Coveralls("test_arcs")
+        self.assertEqual(get_num_alive(), 1)
+        self.assertEqual(coveralls.strong_count(), 2)
```
rfk, line 97: To check my understanding: the strong count is 2 because 1 reference is held by the foreign-language code, and one reference is held by the Rust code for the duration of the call?

mhammond: yes, in this specific example, 1 is in `coveralls._handle` and the other is a new Arc created for `self: Arc<Self>`

uniffi_bindgen/src/bindings/python/templates/RustBufferBuilder.py
rfk: "What could possibly go wrong?" they ask, tempting fate...https://github.com/mozilla/uniffi-rs/issues/457

rfk:
I'm going to close this in favour of a rebased and squashed commit in #462, which incorporates some parts of this that have already been taken onto `main`.

PR #421, opened 2021-03-31
Add an ADR to remove support for non-threadsafe interfaces.

mhammond:
### [Click for Rendered View](https://github.com/mhammond/uniffi-rs/blob/adr-0004/docs/adr/0004-only-threadsafe-interfaces.md)

This attempts to the document the reasons why we think removing
support for uniffi managing thread-safety is a bad idea and
formalizes a decision to remove that support.

### DIFF
--- a/docs/adr/0004-only-threadsafe-interfaces.md
+++ b/docs/adr/0004-only-threadsafe-interfaces.md
@@ -1,9 +1,10 @@
 # Remove support for non-`Send+Sync` interfaces
 
-* Status: proposed
-* Deciders: mhammond, rfkelly, travis, jhugman, dmose; proposed: janerik
-* Date: 2021-04-13
+* Status: accepted
+* Deciders: mhammond, rfkelly, travis, jhugman, dmose
+* Date: 2021-04-19
 
+Discussion and approval: [PR 421](https://github.com/mozilla/uniffi-rs/pull/421)
 Technical Story: [Issue 419](https://github.com/mozilla/uniffi-rs/issues/419)
 
 ## Context and Problem Statement
--- a/docs/adr/0004-only-threadsafe-interfaces.md
+++ b/docs/adr/0004-only-threadsafe-interfaces.md
@@ -1,4 +1,4 @@
-# Remove support for non-threadsafe interfaces
+# Remove support for non-`Send+Sync` interfaces
 
 * Status: proposed
 * Deciders: mhammond, rfkelly, travis; proposed: jhugman, dmose, janerik
@@ -10,7 +10,7 @@ Technical Story: [Issue 419](https://github.com/mozilla/uniffi-rs/issues/419)
 
 [ADR-0003](0003-threadsafe-interfaces.md) introduced support for "thread-safe
 interfaces" - possibly leading to the impression that there is such a thing as
-non-threadsafe interfaces.
+non-threadsafe interfaces and confusion about exactly what the attribute means.
 
 However, the entire concept of non-threadsafe interfaces is a mismomer -
 everything wrapped by uniffi is thread-safe - the only question is who manages
@@ -20,12 +20,20 @@ therefore not obvious to the casual reader.
 
 The [Threadsafe] marker acts as a way for the component author to opt out of
 the overhead and blocking behaviour of this mutex, at the cost of opting in to
-managing their own locking internally. This ADR proposes that managing that
-locking internally be the only locking mechanism supported by uniffi.
+managing their own locking internally. This ADR proposes that uniffi forces
+component authors to explicitly manage that locking in all cases - or to put
+this in Rust terms, that all structs supported by uniffi must already be
+`Send+Sync`
+
+Note that this ADR will hence-forth use the term `Send+Sync` instead of
+"Threadsafe" because it more accurately describes the actual intent and avoids
+any misunderstandings that might be caused by using the somewhat broad and
+generic "Threadsafe".
 
 ## Decision Drivers
 
-* Non-threadsafe structs are considered a "foot-gun", leading to accidentally
+* Supporting non-`Send+Sync` structs means uniffi must add hidden locking to make
+  them `Send+Sync`. We consider this a "foot-gun" as it may lead to accidentally
   having method calls unexpectedly block for long periods, such as
   [this Fenix bug](https://github.com/mozilla-mobile/fenix/issues/17086)
   (with more details available in [this JIRA ticket](https://jira.mozilla.com/browse/SDK-157).)
@@ -37,22 +45,22 @@ locking internally be the only locking mechanism supported by uniffi.
 
 ## Considered Options
 
-* [Option 1] Continue supporting non-threadsafe interfaces while also working
-  on the enhancements listed above, but exclude non-threadsafe interfaces from
+* [Option 1] Continue supporting non-`Send+Sync` interfaces while also working
+  on the enhancements listed above, but exclude non-`Send+Sync` interfaces from
   such enhancements.
 
 * [Option 2] Immediately deprecate, then remove entirely, support for
-  non-threadsafe interfaces.
+  non-`Send+Sync` interfaces.
 
 ## Decision Outcome
 
 Chosen option:
 
 * **[Option 2] Immediately deprecate, then remove entirely, support for
-  non-threadsafe interfaces.**
+  non-`Send+Sync` interfaces.**
 
 This decision was taken because our real world experience tells us that
-non-threadsafe interfaces are only useful in toy or example applications (eg,
+non-`Send+Sync` interfaces are only useful in toy or example applications (eg,
 the nimbus and autofill projects didn't get very far before needing these
 capabilities), so the extra ongoing work in supporting these interfaces can not
 be justified.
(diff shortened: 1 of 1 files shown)


docs/adr/0004-only-threadsafe-interfaces.md
```diff
@@ -0,0 +1,118 @@
...
+
+[ADR-0003](0003-threadsafe-interfaces.md) introduced support for "thread-safe
+interfaces" - possibly leading to the impression that there is such a thing as
+non-threadsafe interfaces.
+
+However, the entire concept of non-threadsafe interfaces is a mismomer -
+everything wrapped by uniffi is thread-safe - the only question is who manages
+this thread-safety. Interfaces which are not marked as thread-safe cause uniffi
+to wrap the interface in a mutex which is hidden in the generated code and
+therefore not obvious to the casual reader.
```
rfk, line 19: Additional context here may be something like: The `[Threadsafe]` marker acts as a way for the component author to opt out of the overhead and blocking behaviour of this mutex, at the cost of opting *in* to managing their own locking internally".

docs/adr/0004-only-threadsafe-interfaces.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Decision Drivers
+
+* Non-threadsafe structs are considered a "foot-gun", leading to accidentally
```
rfk, line 23: nit: grammer here..."leading us to" or "leading to accidental introduction of" or similar?

rfk: I think it may be worth saying explicitly that the issue it causes is "method calls that can unexpectedly block for long periods" or similar.

docs/adr/0004-only-threadsafe-interfaces.md
```diff
@@ -0,0 +1,118 @@
...
+
+This decision was taken because our real world experience tells us that
+non-threadsafe interfaces are only useful in toy or example applications (eg,
+the nimbus and autofill projects didn't get very far before needing these
+capabilities), so the extra ongoing work in supporting these interfaces can not
+be justified.
+
+### Positive Consequences
+
+* The locking in all uniffi supported applications will be clear.
```
rfk, line 56: I endorse "clear", but I also like to describe it as "easily discoverable" - that is, you can look at your code and its type signatures and correctly understand what the locking/blocking behaviour will be without have to understand anything about the internals of uniffi.

docs/adr/0004-only-threadsafe-interfaces.md
```diff
@@ -0,0 +1,118 @@
...
+* All consumers (both inside Mozilla and external) will need to change their
+  interfaces to support thread-safety.
+
+* Simple, toy applications will be more difficult to wrap - consumers will not
+  be able to defer decisions about thread-safety.
+
+* Existing applications that are yet to consider thread-safety can not be
+  wrapped until they have.
+
+* The examples will become more complex as they will all need to implement and
```
rfk, line 75: I think this will be useful to us actually, having to work through the updates to those examples and try to explain it. (It's still work that we wouldn't otherwise have to do, but I'm glad we have it to do).

docs/adr/0004-only-threadsafe-interfaces.md
```diff
@@ -0,0 +1,118 @@
...
+  as possible.
+
+* Noisily deprecate non-threadsafe interfaces so existing consumers are likely
+  to see warnings and a link to our documentation when they upgrade.
+
+* Upgrade all internal mozilla consumers as soon as possible so they do not
+  issue deprecation warnings.
+
+* Perform the actual removal as late as possible (ie, until support for non
+  threadsafe interfaces actually inhibits our ability to add new features)
```
rfk, line 114: This seems well-considered, thanks for layout out the steps like this!

rfk (approved):
Mark and I discussed this synchronously, and I am quite in favour of the proposal.

To be honest, I feel like making the code author explicitly aware of these details and having them encode their choices rather than hiding the complexity behind a costly abstraction...well, it seems very on-brand for a Rust tool, so I'm not even that worried about it being a big hindrance to adoption of the tool in practice 😅 

I have this idea that one should be able to build a macro that lets users opt back in to the previous behaviour with a single explicit line, something like:

```
#[synchronized]
pub struct MyToyExampleWithNoLocking {
...
}
```

Mark has explained to me various ways that this is complicated in practice, but still, if we *wanted* to have the same kind of ease-of-getting started then I think that's approximately the developer experience we should aim for - an annotation on the struct that sits between the user's code and uniffi's FFI scaffolding and applies the locking logic at that layer.

travis79 (approved):
Well and clearly explained!  I agree that this will make us better off in the long run.  You have my r+

docs/adr/0004-only-threadsafe-interfaces.md
```diff
@@ -0,0 +1,118 @@
...
+### Negative Consequences
+
+* All consumers (both inside Mozilla and external) will need to change their
+  interfaces to support thread-safety.
+
+* Simple, toy applications will be more difficult to wrap - consumers will not
+  be able to defer decisions about thread-safety.
+
+* Existing applications that are yet to consider thread-safety can not be
+  wrapped until they have.
```
dmose, line 73: In some ways, this sounds like a positive consequence to me.  I think the current state of affairs effectively makes a promise that it can't keep by appearing simpler than it actually is.

docs/adr/0004-only-threadsafe-interfaces.md
dmose: +1

I think doing this work will also help us understand the edges and consequences of this decision better, and I suspect we'll end up wanting to continue iterating on this document until that work is done.

docs/adr/0004-only-threadsafe-interfaces.md
rfk: > Which is to say that I'd vote for also dropping the [Threadsafe] UDL attribute as well (without changing any
> other semantics).

I agree, and in fact I had implicitly assumed that this is what would happen - the `[Threadsafe]` attribute would disappear and its requirement that the underlying Rust struct be `Sync + Send` would apply by default to all interfaces exposed via UniFFI. It certainly seems worth making that more explicit in the doc here.

mhammond: > I agree, and in fact I had implicitly assumed that this is what would happen - the `[Threadsafe]` attribute would disappear and its requirement that the underlying Rust struct be `Sync + Send` would apply by default to all interfaces exposed via UniFFI. It certainly seems worth making that more explicit in the doc here.

I'm not sure where that would naturally live TBH - I don't think here is the right place. I actually thought "remove entirely, support for non-threadsafe interfaces" is fairly clear in that regard.

mhammond: > I'm not convinced that this will, by itself, cause most casual SDK users to reason correctly about what's likely to block and what's not.

Yes, fair point - I'll reword this to say something like "less hidden" which is actually the point being made.

rfk: We could say a few more explicit words about it under the "implications" section maybe, describing the intended final state along the lines of the above "`[Threadsafe]` disappears, all interfaces `Sync + Send`" as above.

mhammond: heh - actually, `"[Threadsafe] disappears` probably isn't correct as that would break existing code in ways we don't want to. It will be more like "Threadsafe becomes the default, and eventually the attribute becomes deprecated" or something. I'll try and come up with something suitable in the "implications" section.

mhammond:
Thanks Dan, those comments were very valuable. I think I've addressed all your comments, so I'm going to re-flag you for review again.

mhammond:
Based on feedback from Dan, I tweaked the ADR to try and avoid the term "Threadsafe" where possible and instead use `Send+Sync` - I added that change as a separate commit, so I'd appreciate feedback on whether that is an improvement or not - I think it is, but it's worth calling out.

It hasn't changed the suggested outcomes though, so I'm not going to re-request review from anyone.

docs/adr/0004-only-threadsafe-interfaces.md
```diff
@@ -0,0 +1,156 @@
...
+* Communicating both this decision and how consumers can work around it as soon
+  as possible.
+
+* Noisily deprecate non-`Send+Sync` interfaces so existing consumers are likely
+  to see warnings and a link to our documentation when they upgrade.
+
+* Upgrade all internal mozilla consumers as soon as possible so they do not
+  issue deprecation warnings. As an example of what this entails,
+  see [this PR](https://github.com/mozilla/uniffi-rs/commit/454dfff6aa560dffad980a9258853108a44d5985)
+  which converts the `todolist` example to be `Send+Sync`.
```
jhugman, line 139: This migration plan might be useful to flesh out a bit. 

Consumers can get out ahead of these changes by making their interfaces `Threadsafe` before this change ships.
Do we need to warn consumers (e.g. having a transitionory release) before the switch?

docs/adr/0004-only-threadsafe-interfaces.md
```diff
@@ -0,0 +1,156 @@
...
+
+Technical Story: [Issue 419](https://github.com/mozilla/uniffi-rs/issues/419)
+
+## Context and Problem Statement
+
+[ADR-0003](0003-threadsafe-interfaces.md) introduced support for "thread-safe
+interfaces" - possibly leading to the impression that there is such a thing as
+non-threadsafe interfaces and confusion about exactly what the attribute means.
+
+However, the entire concept of non-threadsafe interfaces is a mismomer -
```
jhugman, line 15: Nit: misnomer.

Is it misnamed? or is it a misconception?

jhugman (approved):
I approve of this decision.

docs/adr/0004-only-threadsafe-interfaces.md
dmose: Do we know who all the consumers are?  I personally know of one non-Mozilla project using UniFFI (or at least did a month or two), and UniFFI was featured in "This Week in Rust" some time back...

dmose (approved):
This looks great, modulo the one framing that I think is worth changing.  I don't think I need to see it again, thought, so r=dmose with that addressed in some way.

docs/adr/0004-only-threadsafe-interfaces.md
```diff
@@ -0,0 +1,156 @@
...
+everything wrapped by uniffi is thread-safe - the only question is who manages
+this thread-safety. Interfaces which are not marked as thread-safe cause uniffi
+to wrap the interface in a mutex which is hidden in the generated code and
```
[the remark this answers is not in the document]
mhammond: No, I think it's accurate as written. If it's not marked as "threadsafe" then we wrap it in a mutex, which is threadsafe. If it is marked as [threadsafe] but is not Send+Sync it will not compile.

I'll clarify this though by changing that line to "The Rust compiler insists that everything wrapped by uniffi is thread-safe."

docs/adr/0004-only-threadsafe-interfaces.md
mhammond: changed to "misconception"

docs/adr/0004-only-threadsafe-interfaces.md
mhammond: I've tried to clarify this a little more to indicate there will at *at least* 1 release where deprecation warnings will be issued but where non-threadsafe-marked interfaces are supported.

(When I wrote this I was trying to hedge my bets a little as I had a concern that supporting non-threadsafe interfaces would make doing some of this work very difficult, but that turns out to not be the case - we can probably keep non-threadsafe interfaces around for as long we feel reasonable for people to upgrade)

docs/adr/0004-only-threadsafe-interfaces.md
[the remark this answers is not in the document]
dmose: Ah, that's great.  Thanks!

mhammond:
I've marked this as "accepted" - thanks all!

PR #422, opened 2021-04-01
Convert the todolist example to be threadsafe.

mhammond:
For demonstration purposes we mark the interface as threadsafe,
meaning we need to drop all `&mut self` functions and wrap
`self.items` in a suitable wrapper - `RwLock` in this case.

This is done in support of [ADR-0004](
https://github.com/mozilla/uniffi-rs/blob/main/docs/adr/0004-only-threadsafe-interfaces.md),
which will contain a link to this PR/commit to show an example of the
kind of modifications interfaces  need to become threadsafe.

This seems worthwhile even if we don't adopt ADR-0004, as it becomes a reasonably real-world threadsafe example which we don't really have at the moment.

### DIFF
--- a/examples/todolist/src/lib.rs
+++ b/examples/todolist/src/lib.rs
@@ -2,6 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use std::sync::RwLock;
+
 #[derive(Debug, Clone)]
 pub struct TodoEntry {
     text: String,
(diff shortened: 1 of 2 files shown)


examples/todolist/src/lib.rs
```diff
@@ -33,57 +35,70 @@ fn create_entry_with<S: Into<String>>(item: S) -> Result<TodoEntry> {
 
 type Result<T, E = TodoError> = std::result::Result<T, E>;
 
-// I am a simple Todolist
-#[derive(Debug, Clone)]
+// I am a simple Todolist. Our .udl defines us as "Threadsafe", meaning it's up
```
rfk, line 38: nit: the switch from singular "I" to the Royal "us/our/we" here is a little jarring, although I'm not sure I can offer you a concrete suggestion to make it better :-)

rfk (approved):
Makes sense to me, and I agree it's valuable even before the linked ADR lands 👍🏻

examples/todolist/src/lib.rs
mhammond: Ended up with just "A simple Todolist."

PR #423, opened 2021-04-01
Mark ADR-0003 as formally accepted

mhammond:


jhugman (approved):
I accept this decision.

PR #424, opened 2021-04-01
Make Swift enums Hashable

sammygutierrez:
Hey!

First, thanks for your work on this project. I've been having fun trying to integrate it into a personal app I've been working on.

Hopefully this is a minor thing. I'm trying to use a sequence of dictionaries that contain enums with associated values. Tests will fail for this case for Swift because those enums do not conform to `Hashable`. 

```
error: instance method 'combine' requires that 'Relationship' conform to 'Hashable'
        hasher.combine(relationship)
        ^
Swift.Hasher:3:37: note: where 'H' = 'Relationship'
    @inlinable public mutating func combine<H>(_ value: H) where H : Hashable
```

It looks like record types conform to `Hashable` by default so perhaps the same could be done for enums.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
@@ -1,7 +1,7 @@
 
 // Note that we don't yet support `indirect` for enums.
 // See https://github.com/mozilla/uniffi-rs/issues/396 for further discussion.
-public enum {{ e.name()|class_name_swift }}: Equatable {
+public enum {{ e.name()|class_name_swift }}: Equatable, Hashable {
     {% for variant in e.variants() %}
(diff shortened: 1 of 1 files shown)


jhugman (approved):
It took me a while to reason through this, sorry for the delay.

If all associated types are `Hashable`, then we're ok.

Are all the associated types `Hashable`?

 * Primitive types are all hashable, 
 * Records as combinations of primitives and records are hashable.
 * Significant objects (`interface`s in UDL terms) aren't hashable, but don't travel across the FFI in `enum`s, so can't be a part of an enum.
 * Enums would be hashable.

Thank you so much for your contribution! Is there anything else you'd like to do? or have done in the uniffi project?

sammygutierrez:
Hey @jhugman.

Thanks for taking a look at this.

I'm still learning Rust and I'll admit this project is a bit over my head. For my project, I have a simple crate that wraps some queries through `rusqlite` which is then imported into iOS (with Swift Package Manager) and Android.

I'd love to help out with this project. If there's something you think that would be a good fit, please let me know!

PR #426, opened 2021-04-09
re-enable python tests for the todolist example.

mhammond:
Not clear why is was disabled, but it passes locally and python tests remain enabled for all other ones. Let's see what CI says.


mhammond:
There were actually 2 disabled tests - one was hiding in my patch queue.

mhammond:
Turns out there was a good reason it was disabled in callbacks - callbacks are not implemented for python :)

jhugman (approved):
I'm not sure there's a reason we shouldn't do this.

Callbacks aren't needed for the TODO list example. Is anything else that isn't supported by the python backend?

mhammond:
Thanks!

> Is anything else that isn't supported by the python backend?

Not that I can recall.

PR #428, opened 2021-04-16
Fix swiftformat issues with Swift extension template

travis79:
This reduces swiftformat noise in generated files by moving the extension access modifier to the extension declaration and removing it from the function declarations. The functions should inherit the fileprivate access from the extension declaration.


jhugman (approved):
This is a nice cleanup, thank you.

travis79:
Hrm, using `fileprivate` with protocol conformance seems to have a conflict of access modifiers. Interesting that this doesn't fail locally. Looking into this to find an acceptable solution.

PR #429, opened 2021-04-19
Issue deprecation warning for non-Threadsafe interfaces

mhammond:
Now that [adr-0004](https://github.com/mozilla/uniffi-rs/blob/main/docs/adr/0004-only-threadsafe-interfaces.md)
has been approved, we should deprecate non-threadsafe interfaces ASAP,
to ensure that consumers have as much time as possible to migrate
before we drop support for non-threadsafe interfaces entirely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,10 @@ This is a little experiment in building cross-platform components in Rust, based
 we've learned in the [mozilla/application-services](https://github.com/mozilla/application-services)
 project.
 
-It's currently at the "basic prototype" stage - the core idea seems to work, but there are a lot of
-missing details and it needs a lot of polish. We're currently focussed on making it useable for a
-real shipping component.
+It's currently at the "minimally useful" stage - the core idea works and is being used in
+a number of shipping products. However, there are still a lot of missing details and it needs a lot of polish.
+We're currently focussed on removing some of the paper-cuts we feel in these shipping products, but believe
+this tool has a bright future for many FFI requirements.
 
 If this sounds interesting to you, please dive in! You can:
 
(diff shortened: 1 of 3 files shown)


mhammond:
FTR, every non-threadsafe struct will issue a message similar to this one when running the 'rondpoint' example:

```
warning: use of deprecated function `note_nonthreadsafe_deprecation_Stringifier`: Non thread-safe objects are deprecated - you should upgrade `Stringifier` to be threadsafe, then add the `[ThreadSafe]` annotation. See https://github.com/mozilla/uniffi-rs/#thread-safety for more details
   --> C:\src\moz\uniffi-rs\target\debug\build\uniffi-example-rondpoint-ba23e33efe0193e7\out/rondpoint.uniffi.rs:875:5
    |
875 |     note_nonthreadsafe_deprecation_Stringifier();
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

README.md
```diff
@@ -6,9 +6,10 @@ This is a little experiment in building cross-platform components in Rust, based
 we've learned in the [mozilla/application-services](https://github.com/mozilla/application-services)
 project.
 
-It's currently at the "basic prototype" stage - the core idea seems to work, but there are a lot of
-missing details and it needs a lot of polish. We're currently focussed on making it useable for a
-real shipping component.
+It's currently at the "minimally useful" stage - the core idea works and is being used in
+a number of shipping products. However, there are still a lot of missing details and it needs a lot of polish.
+We're currently focussed on removing some of the paper-cuts we feel in these shipping products, but believe
+this tool has a bright future for many FFI requirements.
```
rfk, line 12: ❤️ this update, thank you!

README.md
```diff
@@ -71,6 +72,32 @@ This is all very experimental and incomplete, but we do have some basic examples
 functions and datatypes in Rust and manipulating them from Kotlin, Swift, and Python.
 Take a look in the [`./examples/`](./examples/) directory to see them in action.
```
rfk, line 73: The `Status` section above here feels like it could maybe just be removed, in favour of the updated status that you added earlier in the doc.

uniffi_bindgen/src/templates/ObjectTemplate.rs
```diff
@@ -8,6 +8,22 @@
...
+// attribute, then generate call to that function in the object constructors.
+#[deprecated(
+    since = "0.8.0",
+    note = "Non thread-safe objects are deprecated - you should upgrade \
+            `{{ obj.name() }}` to be threadsafe, then add the `[ThreadSafe]` \
+            annotation. \
+            See https://github.com/mozilla/uniffi-rs/#thread-safety for more details"
+)]
+#[allow(non_snake_case)]
+fn note_nonthreadsafe_deprecation_{{ obj.name() }}() {}
```
rfk, line 24: nit: please prefix this with `uniffi_` or similar; I don't think we're 100% in the habit of doing this for our internal-use symbols, but it seems worthwhile to try to converge towards a common pattern there when we can.

rfk (approved):
That's quite a hack to make this work right, but I like it. r+ with a couple of nits.

PR #430, opened 2021-04-20
Add an ADR which calls for replacing `HandleMap`s with `Arc`s.

mhammond:
Combined with [ADR-0004](https://github.com/mozilla/uniffi-rs/blob/main/docs/adr/0004-only-threadsafe-interfaces.md
) this is the last significant decision to be made on the path to
allowing us to [pass around object instances](https://github.com/mozilla/uniffi-rs/issues/419).


docs/adr/0005-arc-pointers.md
```diff
@@ -0,0 +1,105 @@
...
+* Bad, because handing raw pointers around means bugs in the generated code or
+  intentional misuse of the bindings might cause vulnerabilities.
+
+## Links
+
+* Thom discusses this a but in [this issue](https://github.com/mozilla/uniffi-rs/issues/244)
+  and agrees with the assertion that raw pointer make sense when all
+  the relevant code is generated.
+
+* Ryan discusses his general approval for this approac in [this issue](https://github.com/mozilla/uniffi-rs/issues/419)
```
rfk, line 105: sp: approach

docs/adr/0005-arc-pointers.md
```diff
@@ -0,0 +1,105 @@
...
+Technical Story: [Issue 419](https://github.com/mozilla/uniffi-rs/issues/419)
+
+## Context and Problem Statement
+
+uniffi currently leans heavily on the HandleMap struct in the ffi-support crate.
+This means that external consumers of uniffi-wrapped interfaces never see
+any pointers to structs - instead, they get what is (roughly) an index into
+an array, with the pointer being in that array.
+
+This has a number of safety characteristics which is particularly important for
```
rfk, line 17: sp: "which is" -> "which are"

docs/adr/0005-arc-pointers.md
```diff
@@ -0,0 +1,105 @@
...
+## Decision Drivers
+
+* We desire the ability to have more flexible life-times for our interfaces so
+  they can be stored in dictionaries or other interfaces, and be returned by
+  functions other than constructors.
+
+* We would like to keep the uniffi implementation as simple as possible while
+  providing a suitable degree of safety - in particular, a promise that it
+  should be impossible to misuse the generated bindings in a way that causes
+  us into Rust's "undefined behavior" (and in particular, avoiding things like
```
rfk, line 35: awkward phrase "causes us into" here, do you mean "to fall into" or similar?

rfk (commented):
Just catching up on issues, and had a few comments while I was in here.

docs/adr/0005-arc-pointers.md
mhammond: thanks for calling out my hand-waving :) I mean something like "generated bindings will be kinda impossible to misuse in a way that causes bad things, like use-after-free" but not sure how to phrase that in a way that acknowledges the reality of the `unsafe` stuff we generate - I'll have another poke 👍

rfk:
@mhammond I put up some suggested edits and additions over in https://github.com/mhammond/uniffi-rs/pull/2

rfk:
Recording my input here for completeness: I've been trying out the prototype from https://github.com/mozilla/uniffi-rs/pull/420 and I really like it, I think the consumer experience is good and the uniffi codebase will be easier to maintain as a result of removing some concepts in favour of simpler things from the stdlib. I'm keen to push this forward, but also don't want to get too carried away if anyone else has concerns.

Related question for @mhammond - do you think we should push through with the outright removal of non-threadsafe interfaces before trying to land this conversation to using raw Arcs? My read of the PoC PR is that there's currently some complexity in order to handle non-threadsafe cases, and it'd be nice if we didn't need to land that, but also maybe we would need to coordinate a series of releases to get the deprecation notice out the door?

mhammond:
> do you think we should push through with the outright removal of non-threadsafe interfaces before trying to land this conversation to using raw Arcs? My read of the PoC PR is that there's currently some complexity in order to handle non-threadsafe cases, and it'd be nice if we didn't need to land that, but also maybe we would need to coordinate a series of releases to get the deprecation notice out the door?

I think what you outline in #449 is a fine approach, although I guess that doesn't answer the question. The extra complexity isn't *too* bad and relatively self-contained, so if supporting both means I can get my hands on Uniffi's `<Arc> sooner, then that would make me happy. OTOH, as soon as it started to get painful I'd be happy to wait (there's literally nothing that is blocked by not having the `Arc` available, it will just enable a few cleanups.

(Did I answer the question? :)

rfk:
> (Did I answer the question? :)

Yes, thank you :-)

docs/adr/0005-arc-pointers.md
mhammond: I had a poke at this before I pulled your PR and we came very close to the same thing - I left a few additional words in, but happy to take them back out if you think it's getting too verbose.

mhammond:
okie dokie, I think this is ready for final review.

docs/adr/0005-arc-pointers.md
```diff
@@ -0,0 +1,167 @@
+# Use raw `Arc<T>` pointers to pass objects across the FFI
+
+* Status: proposed
+* Deciders: mhammond, rfkelly
+* Consulted: travis, jhugman, dmose
```
rfk, line 5: @travis79, @jhugman, @dmose do you feel like you have been sufficiently consulted on this change, or would you like additional details or time or otherwise to understand the implications here in more detail?

docs/adr/0005-arc-pointers.md
```diff
@@ -0,0 +1,167 @@
...
+an array, with the struct being stored in (and owned by) that array.
+
+This has a number of safety characteristics which are particularly important for
+hand-written FFI interfaces, but it does cause some issues in evolving UniFFI in
+directions we consider important. In addition to the slight performance overhead,
+the use of `HandleMap`s makes it difficult to support:
+
+* Passing object instances as arguments ([#40](https://github.com/mozilla/uniffi-rs/issues/40)).
+  Getting objects out of a `HandleMap` involves a closure, so accepting multiple
+  object-typed arguments would involve code-generating nesture closures.
```
rfk, line 28: I'm sure this is my typo, but: "nesture" -> "nested"

docs/adr/0005-arc-pointers.md
```diff
@@ -0,0 +1,167 @@
...
+* Returning object instances from functions ([#197](https://github.com/mozilla/uniffi-rs/issues/197)).
+  Does the returned object already exist in the handlemap? If so, what is its handle?
+  How will we manage the lifetime of multiple references to the object?
```
rfk, line 31: We could add another bullet point here about having object-typed fields in records, which AFAICT is going to Just Work as part of this approach.

rfk (approved):
Looks great! I'm r+ on this, and excited enough about it to push forward with the implementation, but let's leave it until early next week so that there's time for any further input.

docs/adr/0005-arc-pointers.md
rfk, line 28:
```suggestion
  object-typed arguments would involve code-generating nested closures.
```

PR #431, opened 2021-04-21
New 'coverall' fixture to test coverage and edge-cases.

mhammond:
This is similar to our examples, but it's intended to be contrived and to
ensure we get good test coverage of all possible options - particularly for
future changes.

It's in the "fixtures" directory so it doesn't even need to make a cursory
effort to be a "good example"; it intentionally panics, asserts params are
certain values, has no-op methods etc. If you're trying to get your head
around uniffi then the "examples" directory will be a much better bet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ members = [
   "examples/sprites",
   "examples/todolist",
   "examples/threadsafe",
+  "fixtures/coverall",
   "fixtures/regressions/enum-without-i32-helpers",
   "fixtures/regressions/cdylib-crate-type-dependency/ffi-crate",
   "fixtures/regressions/cdylib-crate-type-dependency/cdylib-dependency",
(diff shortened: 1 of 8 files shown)


jhugman (approved):
This is an excellent complement to `rondpoint`, which is getting a bit too big and French.

So far, testing is just in Python. Should we file bugs to get them into Kotlin and Swift?

mhammond:
> This is an excellent complement to `rondpoint`, which is getting a bit too big and French.

Actually, one of the things I don't like about coveralls is that is has too many "simple type" checks - that's what rondpoint is! So we should consider fomalizing this - `rondpoint` == "simple type/dict roundtripping", `coverall` == "higher level/interfaces". Move `rondpoint` to `fixtures`?

jhugman:
I think: let's do it, but after https://github.com/mozilla/uniffi-rs/pull/440 lands!

PR #432, opened 2021-04-22
Define our rust version policy

mhammond:
Clippy is getting upset because it's running 1.5x, but we have an [informal policy](https://github.com/mozilla/uniffi-rs/blob/a7ddeed2ab555f684d521dded9d049145901dcee/uniffi_bindgen/src/interface/literal.rs#L67-L68) of tracking mozilla-central's version - let's formalize that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,7 @@ jobs:
   Check Rust formatting:
     docker:
       # Our policy for updating this rust version is at https://github.com/mozilla/uniffi-rs/tree/main/docs/policies/rust-versions.md
+      # See also rust-toolchain.toml in the root of this repo, which is used to ensure the local version of rust matches.
       - image: circleci/rust:1.47.0
     resource_class: small
     steps:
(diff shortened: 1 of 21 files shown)


mhammond:
So with no ` -- -D warnings` this seems green, and I think an improvement. WDYT?

mhammond:
Re a discussion Dan and I had: I asked on #build in element and they tell me `MINIMUM_RUST_VERSION` is what we should be tracking and not the version our builders currently use (which at this time is 1.47 vs 1.51, so not an insignificant difference)

mhammond:
I knew that app-services had a policy but couldn't work out where it was enforced - and then I remembered [rust-toolchain](https://github.com/mozilla/application-services/blob/main/rust-toolchain). Somewhat surprisingly, this pins app-services to 1.50.0 and all relevant documentation just says somewhat vaguely "whatever version is used by mozilla-central".

It also seems unfortunate that we now have this version specified *3* times.

So I'm not sure where this leaves us. I'm considering splitting the actual fixes from the policy part of this PR - WDYT?

dmose (approved):
I actually disagree with the logic of tracking the minimum rust version in the tree, but it's not that important to me, and I expect that relitigating with the build team would be painful and might not convince them, so, sure, go for it.    

This all looks fine to me.  I assume the Rust changes are because the older versions of clippy and rustfmt have different desires than 1.50 does?

Issue #374, opened 2021-01-20
Use an ordered set of types in TypeUniverse


sysrqb:
Currently, `types::TypeUniverse:all_known_types` is a `HashSet`, unfortunately this breaks reproducible builds because uniffi_bindgen iterates over that set (in an undefined order) when generating bindings resulting in different ordering in the generated files (see https://gitlab.torproject.org/tpo/applications/tor-browser-build/-/issues/40208 for some context, starting at Fenix and working down to uniffi_bindgen :) )

As one possible solution, substituting `HashSet` for `BTreeSet`, and adding `PartialOrd` and `Ord` as derived traits seems to solve this problem.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2014)

sysrqb:
Specifically, this is the patch I'm currently testing on v0.5.0

```diff
diff --git a/uniffi_bindgen/src/interface/types.rs b/uniffi_bindgen/src/interface/types.rs
index 6448d58..b7efd22 100644
--- a/uniffi_bindgen/src/interface/types.rs
+++ b/uniffi_bindgen/src/interface/types.rs
@@ -26,7 +26,7 @@
 use anyhow::bail;
 use anyhow::Result;
 use std::convert::TryFrom;
-use std::{collections::hash_map::Entry, collections::HashMap, collections::HashSet};
+use std::{collections::hash_map::Entry, collections::HashMap, collections::BTreeSet};
 
 use super::Attributes;
 
@@ -71,7 +71,7 @@ pub enum FFIType {
 /// Represents all the different high-level types that can be used in a component interface.
 /// At this level we identify user-defined types by name, without knowing any details
 /// of their internal structure apart from what type of thing they are (record, enum, etc).
-#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash, PartialOrd, Ord)]
 pub enum Type {
     // Primitive types.
     UInt8,
@@ -195,7 +195,7 @@ pub(crate) struct TypeUniverse {
     // Named type definitions (including aliases).
     type_definitions: HashMap<String, Type>,
     // All the types in the universe, by canonical type name.
-    all_known_types: HashSet<Type>,
+    all_known_types: BTreeSet<Type>,
 }
 
 impl TypeUniverse {
```

mhammond:
[...]

PR #433, opened 2021-04-28
Store the universe of known types in an ordered BTreeSet (#374)

sysrqb:
The set of types is used in multiple places during FFI creation. When
the data structure does not guarantee the order of elements, such as
HashSet, the resulting generated source code is not deterministic. Using
a BTreeSet resolves this issue by providing a well-defined element order
over which the the set may be iterated.

### DIFF
--- a/uniffi_bindgen/src/interface/types/mod.rs
+++ b/uniffi_bindgen/src/interface/types/mod.rs
@@ -21,7 +21,7 @@
 //! about how these API-level types map into the lower-level types of the FFI layer as represented
 //! by the [`ffi::FFIType`] enum, but that's a detail that is invisible to end users.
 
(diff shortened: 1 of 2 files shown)


sysrqb:
I thought adding too much detail in the code regarding choosing BTreeSet over HashSet was too verbose and seemed like it complicated the existing concise comments. I can move some (or all) of the reasoning provided in the commit message into the code, if that is preferred.

mhammond:
Thanks! This looks good to me - is a test possible? On one hand I think the comments really don't need to spell out exactly *why* the order must be stable, but a test that they are actually stable seems possible and might be a nice place to add a note as to why.

mhammond (approved):
Perfect, thanks very much

Issue #348, opened 2020-11-10
Support for more common data types


jhugman:
This is more an idea than a need, capturing here to not forget it.

We currently are able to support Strings, numbers and booleans.

It would be nice to be able to transfer types like:

 * URL/URIs
 * Date/Timestamps
 * Filepaths

I expect there is some complexity around giving the client some choice of implementations (e.g. `java.net.URL` vs `android.net.Uri`, [the various date libraries available to android][1]), but for a first cut this wouldn't be necessary.

[1]: https://www.rockandnull.com/java-time-android/

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-36)
┆Issue Number: UNIFFI-36

rfk:
Also JSON-related datatypes, ref https://github.com/mozilla/uniffi-rs/issues/411

rfk:
Cross-referencing #440 which is adding a `JSONObject` type, and in particular my comments at https://github.com/mozilla/uniffi-rs/pull/440#pullrequestreview-662789411 about whether we should be adding lots of builtin datatypes, or working on an extension mechanism. (I don't really know how an extension mechanism would work yet, but it's good to think about).

mhammond:
> (I don't really know how an extension mechanism would work yet, but it's good to think about

Assuming #416, I could imagine it could be exposed procedurally (and macros can add sugar for it). I don't have a concrete suggestion yet, but alot of the "blockers" I see to that approach tend to be due to the fact the interface is in the .idl.

mhammond:
I think we are very happy with custom types here, and are ideally removing types (Duration, anyone? ;)

PR #434, opened 2021-04-30
Support Timestamp and Duration types

npars:
- Implement Timestamp and Duration types (See #348)
- For Rust, map Timestamp to Rust SystemTime type and Duration to Rust
  Duration type
- For Kotlin, map Timestamp to Java Instant type and Duration to Java
  Duration type
- For Python, map Timestamp to naive Python datetime type and Duration
  to Python timedelta type
- [x] Implement bindings for Swift
- [ ] ~Implement bindings for Gecko~
- [x] Update documentation


npars:
This PR adds the common Timestamp and Duration types to Uniffi, before I proceed any further there were a few open questions I wanted to discuss:

1. The current implementation for Python loses precision because Python's internal [`datetime`](https://docs.python.org/3/library/datetime.html) type uses a floating point number to store the timestamp internally (or at least for doing some of it's math). This could be a gotcha for those expecting to have the exact nanoseconds as kept by Rust. 

   I there think is a trade-off we need to decide on here: 
   * A) Always use the internal types for each language, warning users that they may experience some data loss
   * B) Roll our own types for each language (like gRPC does for Timestamp) and force the user to do make their own decisions on how to handle any conversions that may result in loss of precision
   
2. The Duration type for Swift ([`TimeInterval`](https://developer.apple.com/documentation/foundation/timeinterval)) is just a typealias on `Double`. Since we already have bindings for `Double` in our templates this causes a namespace collision if I were to try to add any for `TimeInterval`. Would appreciate some suggestions on the best way forward here, should I try rewriting the Swift templates so that the function names are explicit? IE: liftDouble, liftTimeInterval, etc.

mhammond:
Thanks for this PR. This looks reasonable to me, and the fact that Python is going to lose nanoseconds doesn't seem that much of a big deal to me. However, for various reasons the people who are best suited to dig in to this aren't going to be fully available for a couple of weeks, so please don't see any delays here as representing a lack of interest!

npars:
@mhammond do you think it would be acceptable to leave the Gecko bindings unimplemented in this PR? It doesn't seem like we have standardized on testing in the examples like we have with Swift, Kotlin, and Python, which make me a bit wary about implementing.

mhammond:
> @mhammond do you think it would be acceptable to leave the Gecko bindings unimplemented in this PR?

Yes, I think that would be fine.

examples/time/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "uniffi-example-time"
```
rfk, line 2: In the early days of this project, the only way we had of running tests was to create example crates, so we added things in our examples that didn't really make narrative sense, but which exercised important parts of the code. The "rondpoint" example is a good example of this - anyone trying to actually learn UniFFI by example using that crate will probably have a bad time.

These days, we have a separate `fixtures` directory for things that are designed to exercise the code itself, rather than to provide examples of how to use the code, and I think it helps keep the repo a bit easier to navigate around in.

How would you feel about moving this `uniffi-example-time` crate into the `./fixtures` directory?  I think you should be able to keep it mostly intact, perhaps just renaming it to `uniffi-fixture-time` or similar.

examples/time/uniffi.toml
```diff
@@ -0,0 +1,9 @@
+[bindings.kotlin]
+package_name = "org.mozilla.uniffi.example.chronological"
+cdylib_name = "chronological"
+
+[bindings.python]
+cdylib_name = "chronological"
+
+[bindings.swift]
+cdylib_name = "chronological"
```
rfk, line 9: Unless you need it for something, I think you should be able to get away without this `uniffi.toml` and just use the uniffi-provided defaults. Some of the existing examples use a `uniffi.toml` just so that we have some test coverage of how `uniffi.toml` works, but as mentioned above, I'm trying to move us away from that habit for new work.

uniffi/src/lib.rs
```diff
@@ -354,6 +358,59 @@ unsafe impl ViaFfi for String {
     }
 }
 
+/// Support for passing timestamp values via the FFI.
+///
+/// Timestamps prior to the UNIX EPOCH will cause a panic.
```
rfk, line 363: Hrm...I wonder how likely we are to encounter these panics in practice.

For most other datatypes, IIUC we should *never* encounter an invalid value, which is why we're happy to panic here. But reading e.g. the docs for [`java.time.Instant`](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html) it seems like API consumers could easily create a pre-epoch Instant on the Kotlin side and try to pass it into the Rust code.

That's probably fine, so long as the Kotlin wrapper takes responsibility for guarding against it and turning it into an exception instead of a crash. More generally: a panic should only happen due to some internal implementation bug in UniFFI, not due to the API consumer mis-using the generated code.

I don't have any concrete change to suggest here, just thinking it through out loud...

uniffi_bindgen/src/bindings/gecko_js/webidl.rs
```diff
@@ -111,6 +111,8 @@ impl From<Type> for WebIDLType {
                 // https://github.com/mozilla/uniffi-rs/issues/295.
                 panic!("[TODO: From<Type>({:?})]", type_)
             }
+            Type::Timestamp => panic!("Timestamp unimplemented"),
+            Type::Duration => panic!("Duration unimplemented"),
```
rfk, line 115: 👍🏻 ; frankly I'd like to find a way for us to not have the `gecko_js` bindings in this repo right now, because they're slowly bit-rotting, but it's definitely not something to worry about in this PR.

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -282,6 +282,70 @@ internal fun String.write(buf: RustBufferBuilder) {
...
+
+internal fun lift{{ canonical_type_name }}(rbuf: RustBuffer.ByValue): {{ type_name }} {
+    return liftFromRustBuffer(rbuf) { buf ->
+        read{{ canonical_type_name }}(buf)
+    }
+}
+
+internal fun read{{ canonical_type_name }}(buf: ByteBuffer): {{ type_name }} {
+    val seconds = buf.getLong()
+    val nanoseconds = buf.getInt().toLong()
```
rfk, line 296: Ah, I see, so Java uses signed integers here while Rust uses unsigned. That's unfortunately, but par for the course here with the Kotlin bindings :-/

Could you please add a comment here noting that the data was written to the buffer as an unsigned integer, but we're deliberately reading as a signed integer because that's what Java/Kotlin need? Just to make it clear that we're doing it on purpose, because we've had bugs in this code in the past where we were doing it by accident.

Alternately - what do you think about enforcing this at the Rust layer, so that all consumers get consistent behaviour? For example, our String type [limits everyone to an i32 string length](https://github.com/mozilla/uniffi-rs/blob/b5a54b162c1af4989c9092761d211f5b942c0ded/uniffi/src/lib.rs#L313) because Java/Kotlin strings have this limit, to ensure consistency for all consumers. In practice that would mean serializing durations as signed rather than unsigned integers, similar to what we do for string length there.

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -282,6 +282,70 @@ internal fun String.write(buf: RustBufferBuilder) {
...
+}
+
+internal fun write{{ canonical_type_name }}(v: {{ type_name }}, buf: RustBufferBuilder) {
+    if (v.epochSecond < 0) {
+        throw IllegalArgumentException("Invalid timestamp, must be greater than UNIX EPOCH")
+    }
+
+    if (v.nano < 0) {
+        throw IllegalArgumentException("Invalid timestamp, nano value must be non-negative")
+    }
```
rfk, line 319: Ah, excellent, this was the "guard against panicking values in the Kotlin layer" logic I was hoping to find based on my earlier comment :-)

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -282,6 +282,70 @@ internal fun String.write(buf: RustBufferBuilder) {
...
+
+    buf.putLong(v.epochSecond)
+    buf.putInt(v.nano)
```
rfk, line 322: As above, please leave a comment here that we're mismatching signed-vs-unsigned on purpose.

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt
```diff
@@ -282,6 +282,70 @@ internal fun String.write(buf: RustBufferBuilder) {
...
+
+internal fun lower{{ canonical_type_name }}(v: {{ type_name }}): RustBuffer.ByValue {
+    return lowerIntoRustBuffer(v) { v, buf ->
+        write{{ canonical_type_name }}(v, buf)
+    }
+}
+
+internal fun write{{ canonical_type_name }}(v: {{ type_name }}, buf: RustBufferBuilder) {
+    buf.putLong(v.seconds)
+    buf.putInt(v.nano)
```
rfk, line 346: Careful, I think that java `Duration` instances [can also be negative](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#isNegative--) so you'll need to perform similar checks to those for `Instant`.

uniffi_bindgen/src/bindings/swift/gen_swift.rs
```diff
@@ -220,23 +222,46 @@ mod filters {
...
     ///
     /// This is used to pass arguments over the FFI, from Swift to Rust.
-    pub fn lower_swift(name: &dyn fmt::Display, _type_: &Type) -> Result<String, askama::Error> {
-        Ok(format!("{}.lower()", var_name_swift(name)?))
+    pub fn lower_swift(name: &dyn fmt::Display, type_: &Type) -> Result<String, askama::Error> {
+        match type_ {
+            Type::Duration => Ok(format!(
+                "{}.lower{}()",
+                var_name_swift(name)?,
+                type_swift(type_)?
```
rfk, line 230: Stylistically, I think we should use `type_.canonical_name()` rather than `type_swift(type_)` here when including the name of the type as part of the name of something else. It won't matter in this case because `Duration` is a primitive type, but it would more clearly communicate the intent of the code, since that's what the canonical type name is for. So that would be something like:

```
format!(
    "{}.lower{}()",
    var_name_swift(name)?,
    type_.canonical_type_name()
)
```

uniffi_bindgen/src/bindings/swift/gen_swift.rs
```diff
@@ -220,23 +222,46 @@ mod filters {
...
     /// Lift a Swift type from an FFI type.
     ///
     /// This is used to receive values over the FFI, from Rust to Swift.
     pub fn lift_swift(name: &dyn fmt::Display, type_: &Type) -> Result<String, askama::Error> {
-        Ok(format!("{}.lift({})", type_swift(type_)?, name))
+        match type_ {
+            Type::Duration => Ok(format!(
+                "{}.lift{}({})",
+                type_swift(type_)?,
+                type_swift(type_)?,
```
rfk, line 244: Similarly, please make this something like:

```
format!(
    "{}.lift{}({})",
    type_swift(type_)?,
    type_.canonical_type_name(),
    name,
)
```

The stylistic difference is more obvious in this case - you want one reference to the type as an actual type expression, and one to a string naming the type to be included as part of the method name.

uniffi_bindgen/src/bindings/swift/gen_swift.rs
```diff
@@ -220,23 +222,46 @@ mod filters {
...
     ///
     /// This is used to receive values over the FFI, when they're part of a complex type
     /// that is passed by serializing into bytes.
     pub fn read_swift(name: &dyn fmt::Display, type_: &Type) -> Result<String, askama::Error> {
-        Ok(format!("{}.read(from: {})", type_swift(type_)?, name))
+        match type_ {
+            Type::Duration => Ok(format!(
+                "{}.read{}(from: {})",
+                type_swift(type_)?,
+                type_swift(type_)?,
```
rfk, line 260: As above re: `canonical_type_name`.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -225,6 +225,57 @@ extension Bool: ViaFfi {
...
+        }
+        let seconds = UInt64(delta)
+        let microseconds = UInt32((delta - Double(seconds)) * 1.0e9)
+        buf.writeInt(seconds)
+        buf.writeInt(microseconds)
+    }
+}
+
+extension TimeInterval {
+    fileprivate static func liftTimeInterval(_ buf: RustBuffer) throws -> Self {
```
rfk, line 249: You may need to change the name of this method to match suggested use of `canonical_type_name` above.

rfk (changes requested):
@npars thanks for diving in! This is a really well executed PR. I've left detailed comments below, but in summary, the "request" changes here is mainly for:

* Please consider moving the test coverage from `./examples` to `./fixtures`.
* I think at least the Kotlin code needs to be more careful around handling negative durations.

Also, a broader thought that I figured I'd share....some [experimenting](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f033d1ece3d6e17fb48b94e7e1ed2f69) suggests that Rust is quite happy to represent `SystemTime` instances that are before the unix epoch. Given that all three of the target languages here require code to explicitly disallow pre-epoch times, I wonder if it's worth doing the opposite and explicitly supporting them, making the Rust code a bit more complicated but the code in each of the binding languages a bit simpler.

For example, we might serialize a `timestamp` as a *signed* number of seconds and unsigned nanoseconds, then in the rust code, check for positive or negative duration and calculate the resulting `SystemTime` via `UNIX_EPOCH + duration` or `UNIX_EPOCH - duration` as appropriate.

What do you think?

npars:
> Also, a broader thought that I figured I'd share....some [experimenting](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f033d1ece3d6e17fb48b94e7e1ed2f69) suggests that Rust is quite happy to represent `SystemTime` instances that are before the unix epoch. Given that all three of the target languages here require code to explicitly disallow pre-epoch times, I wonder if it's worth doing the opposite and explicitly supporting them, making the Rust code a bit more complicated but the code in each of the binding languages a bit simpler.
> 
> For example, we might serialize a `timestamp` as a _signed_ number of seconds and unsigned nanoseconds, then in the rust code, check for positive or negative duration and calculate the resulting `SystemTime` via `UNIX_EPOCH + duration` or `UNIX_EPOCH - duration` as appropriate.
> 
> What do you think?

That's a good catch, I mistakenly made an assumption based on the Rust `Duration` type that `SystemTime` was stored internally as a `u64` as well. After digging a bit it looks like [`SystemTime`'s implementation](https://doc.rust-lang.org/src/std/time.rs.html#185) is actually specific to each target environment. In the playground you posted I verified that the max value for that env is:
```rs
std::time::UNIX_EPOCH + Duration::from_secs(i64::MAX as u64)
```
As a side note, it looks like there is an [existing issue around this behaviour](https://github.com/rust-lang/rust/issues/44394).

I think it makes sense to support pre-epoch times. That change will still leave us with positive only `Duration`, but that's perhaps a bit easier to swallow.

rfk:
> That change will still leave us with positive only Duration, but that's perhaps a bit easier to swallow.

Agreed, I don't think there's much we can do about that for `Duration`, and also it seems reasonable to reflect the Rust code's semantics in that case.

npars:
@rfk I think I've addressed all of your comments. There's probably some room for optimization in the conversion code for each language, but I think it's ok for a first pass.

rfk:
> There's probably some room for optimization in the conversion code for each language

This could be said about many things in the current state of this repo, I don't think it will be a blocker for merge 👍🏻

docs/manual/src/internals/lifting_and_lowering.md
```diff
@@ -59,6 +59,8 @@ Calling this function from foreign language code involves the following steps:
 | `f64`/`double` | `double` |
 | `boolean` | `int8_t`, either `0` or `1` |
 | `string` | `RustBuffer` struct pointing to utf8 bytes |
+| `timestamp` | `RustBuffer` struct pointing to a u64 representing seconds and a u32 representing nanoseconds |
```
rfk, line 62: For timestamps, this is now an `i64` representing seconds, not a `u64`.

uniffi/src/lib.rs
```diff
@@ -354,6 +358,76 @@ unsafe impl ViaFfi for String {
     }
 }
 
+/// Support for passing timestamp values via the FFI.
+///
+/// Timestamps values are currently always passed by serializing to a buffer.
```
rfk, line 363: Could you please add a brief comment here about the use of an `i64` to represent times before the unix epoch? It's a detail that's otherwise easy to miss for casual readers.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -225,6 +225,75 @@ extension Bool: ViaFfi {
...
+extension Date: ViaFfiUsingByteBuffer, ViaFfi {
+    fileprivate static func read(from buf: Reader) throws -> Self {
+        let seconds: Int64 = try buf.readInt()
+        let microseconds: UInt32 = try buf.readInt()
+        if seconds >= 0 {
+            let delta = Double(seconds) + (Double(microseconds) / 1.0e9)
+            return Date.init(timeIntervalSince1970: delta)
+        } else {
+            let delta = Double(-seconds) + (Double(microseconds) / 1.0e9)
+            return Date.init(timeIntervalSince1970: -delta)
```
rfk, line 237: I don't understand the double negation here (once in `Double(-seconds)` and then again on the result in `-delta`. Is it necessary, or does it help with precision or something like that?

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -225,6 +225,75 @@ extension Bool: ViaFfi {
...
+        if delta < 0 {
+            sign = -1
+            delta = -delta
+        }
+        if delta.rounded(.down) > Double(Int64.max) {
+            fatalError("Timestamp overflow, exceeds max bounds supported by Uniffi")
+        }
+        let seconds = Int64(delta)
+        let microseconds = UInt32((delta - Double(seconds)) * 1.0e9)
+        buf.writeInt(sign * seconds)
```
rfk, line 253: Similarly here, I don't understand why we need to extract `sign` from `delta` and the multiple them back together. Is it so that we can do the microseconds calculation more easily? Please consider adding a comment to help future readers (aka myself in 6 months from now when I forget what this was all about 😅)

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -225,6 +225,75 @@ extension Bool: ViaFfi {
...
+
+    fileprivate func lowerDuration() -> RustBuffer {
+      let writer = Writer()
+      self.writeDuration(into: writer)
+      return RustBuffer(bytes: writer.bytes)
+    }
+
+    fileprivate static func readDuration(from buf: Reader) throws -> Self {
+        let seconds: UInt64 = try buf.readInt()
+        let microseconds: UInt32 = try buf.readInt()
```
rfk, line 277: IIUC the value read from the buffer is `nanoseconds`, not `microseconds`.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -225,6 +225,75 @@ extension Bool: ViaFfi {
...
+        }
+
+        if self < 0 {
+            fatalError("Invalid duration, must be non-negative")
+        }
+
+        let seconds = UInt64(self)
+        let microseconds = UInt32((self - Double(seconds)) * 1.0e9)
+        buf.writeInt(seconds)
+        buf.writeInt(microseconds)
```
rfk, line 293: As above, the value written to the buffer should be named `nanoseconds`, not `microseconds`.

rfk (approved):
Looks great, thank you!

I left a couple of requests for more comments below, and one place where I think the name of a variable is misleading, but otherwise this is in really good shape. I'm going to tick "approve" here because I don't need to look at this again in any detail, please just adjust for those small comments and let me know when it's pushed, and I'll go ahead and merge.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
npars: I was trying to match how the math works in the other languages, but it just makes for a confusing read in this case. Will update to simplify.

uniffi/src/lib.rs
npars: I gave it a shot :sweat_smile:

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
npars: Added a comment a few lines up

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -228,30 +228,33 @@ extension Bool: ViaFfi {
...
         }
     }
 
     fileprivate func write(into buf: Writer) {
         var delta = self.timeIntervalSince1970
         var sign: Int64 = 1
         if delta < 0 {
+            // The nanoseconds portion of the epoch offset must always be
+            // positive, to simplify the calculation we will use the absolute
+            // value of the offset.
```
rfk, line 247: 👍🏻

uniffi/src/lib.rs
```diff
@@ -361,6 +361,17 @@ unsafe impl ViaFfi for String {
...
+/// Timestamps are represented on the buffer by an i64 that indicates the
+/// direction and the magnitude in seconds of the offset from epoch, and a
+/// u32 that indicates the nanosecond portion of the offset magnitude. The
+/// nanosecond portion is expected to be between 0 and 999,999,999.
+///
+/// To build an epoch offset the absolute value of the seconds portion of the
+/// offset should be combined with the nanosecond portion. This is because
+/// the sign of the seconds portion represents the direction of the offset
+/// overall. The sign of the seconds portion can then be used to determine
+/// if the total offset should be added to or subtracted from the unix epoch.
```
rfk, line 374: Perfect, thanks so much for adding these!

rfk (approved):

PR #436, opened 2021-05-06
Add ruby adapter

saks:
Hey!

First of all, thanks for your hard work on this project. I really like the idea and the code quality that it produces for mobile platforms.

I was wondering if I can use `uniffi-rs` to generate ruby library in addition to swift and kotlin. While experimenting I have reached the point where the generated ruby code is good enough as a direct replacement to my FFI layer that was written by hand.

This adapter is by no mean on par with all features but it does provide basic integration that can be useful for starters. I happy to continue improving and maintaining this code.

FYI: this PR is heavily influenced and based on python integration.

Any feedback is very much appreciated!


.circleci/config.yml
```diff
@@ -47,6 +47,12 @@ jobs:
           name: "Set RUSTFLAGS to fail the build if there are warnings"
           command: echo 'export RUSTFLAGS="-D warnings"' >> $BASH_ENV
       - checkout
+      - run:
+          name: "Install ruby"
```
saks, line 51: This step can certainly be added to the docker image if necessary

uniffi_bindgen/src/bindings/ruby/gen_ruby.rs
```diff
@@ -0,0 +1,256 @@
...
+    "until", "when", "while", "yield", "__FILE__", "__LINE__",
+];
+
+fn is_reserved_word(word: &str) -> bool {
+    RESERVED_WORDS.contains(&word)
+}
+
+// Some config options for it the caller wants to customize the generated ruby.
+// Note that this can only be used to control details of the ruby *that do not affect the underlying component*,
+// sine the details of the underlying component are entirely determined by the `ComponentInterface`.
```
mhammond, line 26: typo - since

mhammond (commented):
This looks good to me with a quick glance by someone who has never written Ruby :) It does raise a good question re CI and the level of support - I'm not sure that it should be a hard requirement that we never break the Ruby bindings. Or if we consider a future where more bindings are supported, there certainly will be some point where we need to declare that some bindings are "tier 1" and others are lower.

I'd welcome any thoughts on that.

Note that for various reasons, there might not be much movement on this PR for a week or few, but don't take that as a lack of interest.

uniffi_bindgen/src/bindings/ruby/gen_ruby.rs
saks: Thanks!

mhammond:
Might also be worth adding support for the new "coverall" fixture (#431) - there are significant changes coming (see the other various open PRs) which will cause some churn here, so getting base-level support in early might help you later :)

.circleci/config.yml
rfk: Yep, I'd be happy to add it to the docker image, and that would make each individual test run go a little faster.

uniffi_bindgen/src/bindings/ruby/gen_ruby.rs
```diff
@@ -0,0 +1,254 @@
...
+const RESERVED_WORDS: &[&str] = &[
+    "alias", "and", "BEGIN", "begin", "break", "case", "class", "def", "defined?", "do", "else",
+    "elsif", "END", "end", "ensure", "false", "for", "if", "module", "next", "nil", "not", "or",
+    "redo", "rescue", "retry", "return", "self", "super", "then", "true", "undef", "unles",
+    "until", "when", "while", "yield", "__FILE__", "__LINE__",
+];
+
+fn is_reserved_word(word: &str) -> bool {
+    RESERVED_WORDS.contains(&word)
+}
```
rfk, line 22: Ah yes, we've put off dealing with reserved words for far too long. We might be able to find a way to push some of the logic for handling this down into shared code.

.circleci/config.yml
saks: Thanks!

uniffi_bindgen/src/bindings/ruby/gen_ruby.rs
saks: > We might be able to find a way to push some of the logic for handling this down into shared code.

@rfk, sounds good. Do you think we should look into that as part of this PR?

rfk: > Do you think we should look into that as part of this PR?

No, unless you have any good ideas you'd like to explore. I think it's fine to start with figuring out how to handle it here in a specific language backend, and evolve the approach as we understand it better.

saks: Sounds good @rfk, I'd prefer to avoid scope creep and deal with one issue at the time. I'm open to look at it later as a separate PR.

saks:
hey @rfk! Is there anything that I can do to facilitate code review of this PR?

rfk:
> Is there anything that I can do to facilitate code review of this PR?

@saks hey, thanks for asking - nothing specific yet, I have it on my stack to look at in detail this week.

One thing that's on my mind is how this will interact with some changes to the way we handle object references, which I'm hoping to also start landing this week (ref [ADR-0004](https://github.com/mozilla/uniffi-rs/blob/main/docs/adr/0004-only-threadsafe-interfaces.md) and the proposed [ADR-0005](https://github.com/mozilla/uniffi-rs/pull/430)). These will mean that the handlemap-related parts of the Ruby backend will need to change, and I'm thinking it may be worth holding off on merging this until that work is complete, and asking you to rebase this PR on top of those changes. (I'll cross-reference the active PRs for that work from here when they're ready to be consumed).

Of course, I can still review the PR in its current state in order to give broader feedback, pending those changes.

Does that sound OK?

saks:
hey @rfk, thanks for getting back to me. I'm glad to see how UniFFI is evolving and I'm looking forward to new opportunities that both ADR-0004 and ADR-0005 would enable.

As for the ruby adapter, the code was prepared and tested with the current UniFFI architecture. Landing this PR before implementing upcoming changes can be beneficial since I can start migrating my projects from hand-written FFI layers into UniFFI and unblock implementation of new functionality.

At the same time I'm sympathetic to development plans for UniFFI and respect decisions made according to it. I'm happy to see what changes are going to be necessary to make ruby adapter work with new pointers handling and thread safety rules. 

Thanks!

rfk (approved):
With the *very large* caveat that I don't really know much Ruby, let alone idiomatic Ruby, this looks solid to me 👍🏻. I'm really happy to see that you went through and made Ruby versions of all the tests. Thanks again for your interest and for diving in here!

I think we can merge this, but I want to reinforce again that we'll have to treat it as a "best effort" maintenance situation. We'll try to keep it running while we work on new features for UniFFI, but if we need to get a new feature released and we can't figure out how to make it work for the Ruby backend, then we won't be able to block the release on getting it working in Ruby. (This is basically how we treat the current `gecko_js` backend as well, FWIW).

That said, I hope you'll stick around to help keep the Ruby backend running, and feel welcome to dive in to any discussions about new features! Please don't hesitate to open an issue or email me if there are things I can do to help you stay involved on an ongoing basis.

Longer term, I hope that we can find a way to decouple the development of specific language backends from all being in the same repo. Assuming that we do, I will probably advocate for the Ruby backend to move into its own separate repository (and the Python backend as well).

What I think we should do from here:

* If you're happy to, please make a separate PR that just adds Ruby to the Docker image. I'll get that pushed and then we can update this PR to use it.
* Please rebase this PR on top of current `main`, which I think will involve some minor conflicts in the test setup code.
* I will finish up my current batch of work on replacing handlemaps, and see how much that's likely to affect the code here. Having had a closer look, I'm actually fairly hopeful that it won't conflict too badly.
* With luck, I'll be able to apply my changes cleanly on top of this PR, in which case I'll go ahead and merge it while we keep working on those handlemap changes.

.circleci/config.yml
saks: Sure thing! (https://github.com/mozilla/uniffi-rs/pull/461)

fixtures/coverall/tests/bindings/test_coverall.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+
+    assert_raise Coverall::InternalError, 'expected panic: oh no' do
+      coveralls.panic 'expected panic: oh no'
+    end
+
+    assert_raise_message /expected panic: oh no/ do
+      coveralls.panic 'expected panic: oh no'
+    end
+  end
```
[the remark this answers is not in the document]
saks: Done! Thanks for pointing that out :)

.circleci/config.yml
rfk: Great; I've pushed the updated docker image, so in theory we can remove this from the CircleCI config and everything should still run and pass 🤞🏻

saks: Done 👍 

FYI: I've noticed that docker image is built with rust `1.50.0` but then we install a different version based on `rust-toolchain.toml` (`1.47.0` at the moment). I've been following why this repo uses a specific rust version, but I guess we can cut a bit of time from each test if we build docker image with rust `1.50.0`. I can prepare a PR if this is useful :)

rfk: > FYI: I've noticed that docker image is built with rust 1.50.0 but then we install a different version based on
> rust-toolchain.toml (1.47.0 at the moment). 

Huh, so we do! I guess the scripting for the docker image there may pre-date our decision to pin to an earlier Rust. If you have a moment to prep a PR, that would be helpful, please do.

saks: hey @rfk, https://github.com/mozilla/uniffi-rs/pull/463

rfk:
This branch needs a couple of tweaks before final landing, for compatibility with other things that have merged to `main`. I'm going to go ahead and so those myself in a fresh PR over in https://github.com/mozilla/uniffi-rs/pull/464 so that we can get this into a release for you to try out. Thanks again!

rfk:
Landed in https://github.com/mozilla/uniffi-rs/commit/962320b7bfb9d843b30ec2c0b9ea82ef9160d8a8

rfk:
The ruby adapter is now availalbe in the `v0.10.0` release of UniFFI 🎉

PR #440, opened 2021-05-07
Adds support for JSON Objects

jhugman:
This PR introduces support for passing JSON objects back and forth across the FFI.

It fixes [SDK-267][1].

This maps `serde_json::Value::Object` from Rust to an untyped dictionary on the foreign language side.

This corresponds to `org.json.JSONObject` in Kotlin, `[String: Any]` (via `JSONSerialization`) in swift and `dict`s with `String` keys in Python.

Serialization and de-serialization is necessarily dumb: uniffi serializes into Strings and then passes it across the FFI. On the other side, 
the object is deserialized from the passed String. This is inefficient, but correct. Clients requiring efficient transport across the FFI would be encouraged to use the existing types `record<string, T>` if they can. 

[1]: https://jira.mozilla.com/browse/SDK-276

### DIFF
--- a/docker/Dockerfile-build
+++ b/docker/Dockerfile-build
@@ -61,12 +61,13 @@ RUN mkdir -p /tmp/setup-kotlin \
     && cd ../ \
     && rm -rf ./setup-kotlin
 
-RUN mkdir -p /tmp/setup-jna \
-    && cd /tmp/setup-jna \
+RUN mkdir -p /tmp/setup-kotlin-env \
+    && cd /tmp/setup-kotlin-env \
     && curl -o jna.jar https://repo1.maven.org/maven2/net/java/dev/jna/jna/5.6.0/jna-5.6.0.jar \
+    && curl -o json.jar https://repo1.maven.org/maven2/org/json/json/20190722/json-20190722.jar \
     # XXX TODO: should check a sha256sum or something here...
-    && sudo mv jna.jar /opt \
-    && echo "export CLASSPATH=\"\$CLASSPATH:/opt/jna.jar\"" >> /home/circleci/.bashrc \
-    && echo "export CLASSPATH=\"\$CLASSPATH:/opt/jna.jar\"" >> /home/circleci/.profile \
+    && sudo mv *.jar /opt \
+    && echo "export CLASSPATH=\"\$CLASSPATH:/opt/jna.jar:/opt/json.jar\"" >> /home/circleci/.bashrc \
+    && echo "export CLASSPATH=\"\$CLASSPATH:/opt/jna.jar:/opt/json.jar\"" >> /home/circleci/.profile \
     && cd ../ \
-    && rm -rf ./setup-jna
+    && rm -rf ./setup-kotlin-env
--- a/examples/rondpoint/src/lib.rs
+++ b/examples/rondpoint/src/lib.rs
@@ -121,6 +121,9 @@ impl Retourneur {
     fn identique_string(&self, value: String) -> String {
         value
     }
+    fn identique_json_value(&self, value: Value) -> Value {
+        value
+    }
     fn identique_nombres_signes(
         &self,
         value: DictionnaireNombresSignes,
@@ -179,7 +182,7 @@ impl Stringifier {
     fn to_string_boolean(&self, value: bool) -> String {
         value.to_string()
     }
-    fn to_string_json_value(&self, value: serde_json::Value) -> String {
+    fn to_string_json_value(&self, value: Value) -> String {
         value.to_string()
     }
     fn well_known_string(&self, value: String) -> String {
--- a/examples/rondpoint/src/lib.rs
+++ b/examples/rondpoint/src/lib.rs
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use serde_json::Value;
 use std::collections::HashMap;
 
 #[derive(Debug, Clone)]
@@ -176,6 +177,9 @@ impl Stringifier {
         value.to_string()
     }
     fn to_string_boolean(&self, value: bool) -> String {
+        value.to_string()
+    }
+    fn to_string_json_value(&self, value: serde_json::Value) -> String {
         value.to_string()
     }
     fn well_known_string(&self, value: String) -> String {
--- a/examples/rondpoint/src/lib.rs
+++ b/examples/rondpoint/src/lib.rs
@@ -121,7 +121,7 @@ impl Retourneur {
     fn identique_string(&self, value: String) -> String {
         value
     }
-    fn identique_json_value(&self, value: Value) -> Value {
+    fn identique_json_object(&self, value: Value) -> Value {
         value
     }
     fn identique_nombres_signes(
@@ -182,7 +182,7 @@ impl Stringifier {
     fn to_string_boolean(&self, value: bool) -> String {
         value.to_string()
     }
-    fn to_string_json_value(&self, value: Value) -> String {
+    fn to_string_json_object(&self, value: Value) -> String {
         value.to_string()
     }
     fn well_known_string(&self, value: String) -> String {
--- a/examples/rondpoint/src/lib.rs
+++ b/examples/rondpoint/src/lib.rs
@@ -76,6 +76,10 @@ fn switcheroo(b: bool) -> bool {
     !b
 }
 
+struct DictionnaireAvecJSON {
+    value: Value,
+}
+
 // Test that values can traverse both ways across the FFI.
 // Even if roundtripping works, it's possible we have
 // symmetrical errors that cancel each other out.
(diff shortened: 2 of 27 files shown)


uniffi/Cargo.toml
```diff
@@ -22,6 +22,7 @@ cargo_metadata = "0.11"
 paste = "1.0"
 uniffi_bindgen = { path = "../uniffi_bindgen", optional = true, version = "= 0.8.0"}
 static_assertions = "1.1.0"
+serde_json = "1.0.64"
```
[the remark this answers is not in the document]
jhugman: I'd like to propose we land this first, then sort it out later.

uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift
```diff
@@ -13,16 +13,20 @@ public struct {{ rec.name()|class_name_swift }}: Equatable, Hashable {
...
+        {% endif %}
         {%- endfor %}
         return true
     }
 
     public func hash(into hasher: inout Hasher) {
         {%- for field in rec.fields() %}
+        {%- if field.type_() != Type::JSONObject %}
         hasher.combine({{ field.name()|var_name_swift }})
+        {% endif -%}
```
jhugman, line 29: I'm not altogether happy about this: since `JSONSerialization` gives us `Dictionary<String, Any>`, we can't do much except ignore it, or stringify and uses that to compare and hash. Neither are _great_ solutions.

jhugman (commented):
Self review.

uniffi/src/lib.rs
```diff
@@ -476,6 +476,66 @@ unsafe impl<V: ViaFfi> ViaFfi for HashMap<String, V> {
...
+}
+
+fn to_supported_string(value: &Value) -> Result<String> {
+    match value {
+        Value::Object(_) => Ok(serde_json::to_string(value)?),
+        _ => {
+            // We can either cause a panic! when unsupported/unsanitized JSON is
+            // accidently given to us, or we can wrap it up as an object with an error message
+            // as a key. Neither is ideal, but this should cause failure down stream quick enough
+            // for the developer to catch it.
```
mhammond, line 529: sorry, but I'm not a fan of this. I think we should panic. I think there are a number of theoretical panics in even the simple type operations. Regardless though, this gets to the heart of the risks in introducing a "fallible" type (which I'm not against, but opens some interesting questions)

uniffi_bindgen/src/bindings/python/gen_python.rs
```diff
@@ -141,9 +141,12 @@ mod filters {
...
             Type::Boolean => format!("bool({})", nm),
-            Type::String | Type::Object(_) | Type::Enum(_) | Type::Error(_) | Type::Record(_) => {
-                nm.to_string()
-            }
+            Type::String
+            | Type::Object(_)
+            | Type::Enum(_)
+            | Type::Error(_)
+            | Type::Record(_)
+            | Type::JSONObject => nm.to_string(),
```
mhammond, line 149: I'm mildly surprised that given the other comments above, JSONObject isn't kinda disguised more as an actual string. I'm sure a longer look will make it clearer though.

mhammond (commented):
🚀

uniffi/src/lib.rs
```diff
@@ -476,6 +476,66 @@ unsafe impl<V: ViaFfi> ViaFfi for HashMap<String, V> {
...
+/// Only top-level objects are supported: i.e. the value must be a `Value::Object`.
+/// If the Rust side tries to pass back a `JSONArray` or a scalar, and the foreign language side
+/// is expecting a dictionary shaped value, it would be a failure.
+///
+/// In this implementation, the value is wrapped in an object with the error message as the key.
+unsafe impl ViaFfi for Value {
+    type FfiType = RustBuffer;
+
+    fn lower(self) -> Self::FfiType {
+        match to_supported_string(&self) {
```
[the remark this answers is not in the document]
mhammond: `to_supported_string(&self).map(String::lower).expect("Unser...")` maybe?

uniffi/src/lib.rs
jhugman: Good to get a second opinion on this. I'll change it to a runtime check and panic, and add some "sanitize your inputs" warnings in the docs.

jhugman (commented):
Addressed your comments. Thank you for the fast turn around.

examples/rondpoint/src/lib.rs
```diff
@@ -178,6 +192,9 @@ impl Stringifier {
     fn to_string_boolean(&self, value: bool) -> String {
         value.to_string()
     }
+    fn to_string_json_object(&self, value: Value) -> String {
```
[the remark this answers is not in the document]
mhammond: Note we have a fn named `_json_object()` taking a `serde_json::Value` arg - and `serde_json::Value::Object` is a thing. So either `_json_value()` or `_json()`? I'm voting for the latter.

> We might want to support a top level JSONArray one day.

Exactly - `Value::Array` vs `Value::Object` - which is why I think the 'object' part here isn't correct - should be 'value' or nothing?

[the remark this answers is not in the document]
jhugman: Very happy to rename it, in the interests of landing and releasing quickly.

docker/Dockerfile-build
```diff
@@ -61,12 +61,13 @@ RUN mkdir -p /tmp/setup-kotlin \
...
     # XXX TODO: should check a sha256sum or something here...
-    && sudo mv jna.jar /opt \
-    && echo "export CLASSPATH=\"\$CLASSPATH:/opt/jna.jar\"" >> /home/circleci/.bashrc \
-    && echo "export CLASSPATH=\"\$CLASSPATH:/opt/jna.jar\"" >> /home/circleci/.profile \
+    && sudo mv *.jar /opt \
+    && echo "export CLASSPATH=\"\$CLASSPATH:/opt/jna.jar:/opt/json.jar\"" >> /home/circleci/.bashrc \
+    && echo "export CLASSPATH=\"\$CLASSPATH:/opt/jna.jar:/opt/json.jar\"" >> /home/circleci/.profile \
     && cd ../ \
-    && rm -rf ./setup-jna
+    && rm -rf ./setup-kotlin-env
```
jhugman, line 73: Unfortunately, I can't rebuild the docker image.

uniffi/src/lib.rs
```diff
@@ -476,6 +476,60 @@ unsafe impl<V: ViaFfi> ViaFfi for HashMap<String, V> {
...
+
+#[cfg(feature = "serde_json")]
+fn to_supported_string(value: &Value) -> Result<String> {
+    match value {
+        Value::Object(_) => Ok(serde_json::to_string(value)?),
+        _ => {
+            // Currently we don't support JSONArrays or scalar values
+            // at the top level. We'd like to catch this at compile time,
+            // but because serde can load everything into a single enum,
+            // we have to rely on runtime checking.
```
mhammond, line 527: I'm sorry I'm looking at this too quickly, but given that comment, maybe you want your API to be truly `Value::Object` (which means `Map<String, Value>`)?

jhugman:
I think widening out support for all JSON types is probably a good aim for
another issue and release of uniffi.

Let's land this, and see where the ergonomic problems are before more
engineering is spent on this.

jhugman:
@mhammond I took your suggestion re: `Map<String, Value>` and renamed the rondpoint methods back to use `json_object`. 

I separated the rename commit off, so if you don't like it, then we can easily get rid of it.

Tests on CI won't pass until we update the docker image, which is likely soonest when @rfk returns, but I'd like to get this landed and released early this week.

docker/Dockerfile-build
rfk: @jhugman I've pulled your branch, built a new docker image, and pushed it, so hopefully this will work now.

I'll also take a look at what our options are for making a team-owned image rather than one under my personal dockerhub account.

rfk: I filed https://github.com/mozilla/uniffi-rs/issues/443 to follow up on the question of access to this docker image.

rfk:
@jhugman I re-triggered the CI here to test out the new docker image. It seems to be working as expected - there's a new and different CI failure which I believe will be resolved if you rebase atop latest `main`.

rfk:
(I've been holding off on diving in to the code in too much detail here until after we manage to sync up in person, but it's on my list for this week)

rfk (changes requested):
The "request changes" here is specifically because I don't think we should merge the current approach to Swift equality where it ignores any JSON fields. I'd prefer we not implement it at all in that case, rather than implement it incorrectly.

Some higher-level discussion below that doesn't need to block anything here.

---

High-level, seeing this PR made me realize that I don't have a very good mental model or set of principles for what sort of functionality should live in core UniFFI, versus what sort of functionality we should try to expose via some kind of extension mechanism.

On the one hand, I'm concerned about the maintenance and cognitive-load burden of taking a "Kitchen Sink" approach and adding in tons of datatypes likes JSON, filepaths, URLs, guids, dates, etc etc. Especially when those datatypes come with extra dependencies for foreign-language consumers, such as the Kotlin JSON bindings where we are basically picking a particular JSON package for consumers to use.

On the other hand, we're a long way away from being able to define extra data types via any sort of extension mechanism, and there are clear short-term advantages of being able to have these types integrated close to the core of UniFFI rather than having to layer them on top of the generated code by hand.

So...on balance, I think I support this if we have a clear short-term consumer for it, but I'd like us to keep it in mind as a motivating use-case for a potential future extensibility mechanism, if and when we decide we want such a thing.

Issue #444, opened 2021-05-18
Propagate @ExperimentalUnsignedTypes annotations for APIs that use unsigned types


rfk:
Ref https://github.com/mozilla/application-services/issues/3957 for the source of this request.

Kotlin's unsigned types are currently experimental, and code that uses them needs to either:

* Explicitly opt-in to using the experimental feature, accepting the risk of bustage if if changes.
* Explicitly propagate the "this uses an experimental feature" flag so that downstream consumers can decide whether to opt in.

Since UniFFI generates library code that is intended for use by others, the right thing for us to do is the latter, allowing consuming code to make the final decision. In practice this means that anywhere our generated Kotlin code uses an unsigned type, should be annotated with `@ExperimentalUnsignedTypes`. If we don't, then consumers will get warnings about using the experimental unsigned types without an explicit opt-in.

In fact we already do this a little bit, as of https://github.com/mozilla/uniffi-rs/issues/216). See for example the [Kotlin RustBuffer helpers](https://github.com/mozilla/uniffi-rs/blob/9aee2083ae51fd8e3e87682a2c10370aa2213617/uniffi_bindgen/src/bindings/kotlin/templates/RustBufferHelpers.kt#L128), which explicitly tag each method for an unsigned type with the `@ExperimentalUnsignedTypes` annotation. But the current implementation does not go far enough.

As an example, for a UDL definition like this:

```
dictionary DirectlyUsesU8 {
  u8 member;
  string other;
};

dictionary RecursivelyUsesU8 {
  DirectlyUsesU8 member;
};
```

We currently generate a Kotlin wrapper that looks like:

```
data class DirectlyUsesU8 (
    val member: UByte,
    val other: String
) {
    // Omitted
}

data class RecursivelyUsesU8 (
    val member: DirectlyUsesU8 
) {
    // Omitted
}
```

But to avoid warnings when compiling the resulting Kotlin code, it should really look like:

```
@ExperimentalUnsignedTypes
data class DirectlyUsesU8 (
    val member: UByte,
    val other: String
) {
    // Omitted
}

@ExperimentalUnsignedTypes
data class RecursivelyUsesU8 (
    val member: DirectlyUsesU8 
) {
    // Omitted
}
```

This issue is about generating `@ExperimentalUnsignedTypes` annotations in the right places in the generated Kotlin code, so that:

* Components that use unsigned types will compile without any warnings about experimental types.
[...]

PR #445, opened 2021-05-18
Add a test fixture for Kotlin "experimental unsigned types" warnings.

rfk:
Previously, our test running infrastructure would specify the Kotlin option
 "-Xopt-in=kotlin.ExperimentalUnsignedTypes" when compiling the uniffi bindings,
in order to silence warnings about using the experimental unsigned types.

This is *not* how we intend to compile the generated bindings in practice.
Rather, we want to only opt in to experimental types at the level of the final
consumer, with all intermediate layers propagating the experimental types
annotation to the layer above them.

This commit removes the unsigned types opt-in when compiling bindings for
tests, and replaces it with `-Werror` so that any warnings when compiling
the bindings will fail the corresponding test.  (It keeps the opt-in when
running the actual test script, since that is the uppermost layer and is
where we intend real consumers to specify the opt-in behaviour).

It also adds a new test fixture that deliberately uses a bunch of unsigned
types, and tests that the Kotlin bindings compile and run successfully.

This test currently fails, because we're not propagating the annotations
correctly. Making the test pass is issue #444.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
   "examples/threadsafe",
   "fixtures/coverall",
   "fixtures/regressions/enum-without-i32-helpers",
+  "fixtures/regressions/kotlin-experimental-unsigned-types",
   "fixtures/regressions/cdylib-crate-type-dependency/ffi-crate",
   "fixtures/regressions/cdylib-crate-type-dependency/cdylib-dependency",
 ]
--- a/fixtures/regressions/kotlin-experimental-unsigned-types/Cargo.toml
+++ b/fixtures/regressions/kotlin-experimental-unsigned-types/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "kotlin-experimental-unsigned-types"
+edition = "2018"
+version = "0.8.0"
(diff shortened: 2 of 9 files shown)


rfk:
@skhamis no particular urgency to look at this, just sharing it based on our quick discussion about these annotations over in https://github.com/mozilla/application-services/pull/4068. Does the eventual fix implied by this test fixture, match your expectations of how UniFFI should work as a consumer?

skhamis (commented):
I think this is definitely what is needed for this! This looks like a really elegant fix! So I would be able to remove that opt-in code but will have to include some form of it on the final consumer side?

rfk:
> So I would be able to remove that opt-in code but will have to include some form of it on the final consumer side?

Right, you'd remove the opt-in code from the logins component build, and instead rely on the ultimate consumer of that code (e.g. the Fenix app) to perform the opt-in during its build process.

skhamis (approved):

skhamis:
We can safely close this @rfk due to this landing: https://github.com/mozilla/uniffi-rs/pull/977

rfk:
👍🏻 thanks for pushing this forward!

PR #446, opened 2021-05-18
Adjust expected `trybuild` error output for new rust version.

rfk:
I recently updated the docker image that we use to run CI, which
seems to have perturbed the version of Rust being used, which has
subtly changed the expected error output from `trybuild`. This
should get CI passing again.


rfk:
Frustratingly, I think it's the *newer* versions of Rust that include the comma in the error message, so there's a change this will make the tests stop passing on local development machines. I'm not entirely sure how to resolve that mismatch though, and it seems better to have things passing reliably in CI.

mhammond (approved):

PR #447, opened 2021-05-19
Make all examples, other than Threadsafe, Threadsafe.

mhammond:
We are leaving the Threadsafe example as the only one
which supports a non-threadsafe interface until we actually
remove support for non-theadsafe interfaces.

I guess this is the next logical step after #429? Note I can't yet remove the `-A deprecated` because one non-threadsafe interface remains.

### DIFF
--- a/examples/callbacks/src/lib.rs
+++ b/examples/callbacks/src/lib.rs
@@ -63,8 +63,9 @@ impl Default for RustGetters {
     }
 }
 
-// Use Send if we want to store the callback in an exposed object.
-trait StoredForeignStringifier: Send + std::fmt::Debug {
+// Use `Send+Send` because we want to store the callback in an exposed
+// `Send+Sync` object.
+trait StoredForeignStringifier: Send + Sync + std::fmt::Debug {
     fn from_simple_type(&self, value: i32) -> String;
     fn from_complex_type(&self, values: Option<Vec<Option<f64>>>) -> String;
 }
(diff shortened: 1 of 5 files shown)


examples/callbacks/src/lib.rs
```diff
@@ -84,4 +84,7 @@ impl RustStringifier {
     }
 }
 
+unsafe impl Send for RustStringifier {}
+unsafe impl Sync for RustStringifier {}
```
rfk, line 88: I'm lightly concerned about this being in an example, where new users might see it and copy-paste it without really understanding the implications. Could you please add a comment about why we need to declare this explicitly rather than Rust figuring it out for us, and why it's safe for us to do so?

examples/sprites/src/lib.rs
```diff
@@ -27,34 +29,35 @@ pub fn translate(p: &Point, v: Vector) -> Point {
 
 // An entity in our imaginary world, which occupies a position in space
 // and which can move about over time.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct Sprite {
-    current_position: Point,
+    current_position: RwLock<Point>,
```
rfk, line 34: For educational purposes, adding a short comment here about "All interfaces must be threadsafe, so we need to wrap this in a lock" may help new users for the correct mental model of how we expect threadsafety to work.

rfk (approved):
👍🏻; I requested some comments in a couple of places in order to help the examples function properly as examples, but this LGTM

examples/callbacks/src/lib.rs
mhammond: err - somewhat embarrassingly, I failed to add an `XXX` comment for myself to remember to come back to this, then a month passed and 🤷 . Rust *can* figure it out - I just needed to add a `Sync` bound to the `StoredForeignStringifier` trait - much cleaner. I updated the comment accordingly, but there's less to hand-wave now!

PR #448, opened 2021-05-20
Try to consistently capitalize "UniFFI" in consumer docs.

rfk:
We don't currently have much consistency around how we capitalize the name of this project. I'd like to propose "UniFFI" as the preferred rendering, as both a Proper Noun and containing a correctly capitalized embedded acronym "FFI". My preference for this particular rendering is weakly held, I just want to pick *something* that we can use consistently :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# uniffi - a multi-language bindings generator for rust
+# UniFFI - a multi-language bindings generator for rust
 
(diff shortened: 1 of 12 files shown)


mhammond (approved):
TBH, I mildly prefer the status quo, but `FFI` is what rust uses for that concept and the leading `U` does work as a proper noun, so it works for me.

PR #451, opened 2021-05-21
Fix Kotlin codegen for writing optional integers.

rfk:
Seems like you can't use `v=== null` on when `v` is an optional primitive type like `UByte?`. I noticed this while working on something else, but figured it would be simpler as a standalone PR. Also, a good excuse to get started on a Kotlin coveralls file.

### DIFF
--- a/fixtures/coverall/tests/bindings/test_coverall.kts
+++ b/fixtures/coverall/tests/bindings/test_coverall.kts
@@ -1,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import uniffi.coverall.*
+
+// Test some_dict().
+// TODO: use an actual test runner.
+
+val d = createSomeDict();
+assert(d.text == "text");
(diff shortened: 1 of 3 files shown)


rfk:
So, I have a fix that I think will address #457 and that I'm *mostly* happy with, but I don't fully trust myself to hit the right complexity tradeoff here. @jhugman @mhammond would you mind giving me a gutcheck on what's proposed here (as well as on the inline commentary about some considered alternatives)?

fixtures/coverall/tests/bindings/test_coverall.kts
```diff
@@ -0,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import uniffi.coverall.*
+
+// Test some_dict().
```
rfk, line 7: FYI, these were copied directly from the python test of the same name: https://github.com/mozilla/uniffi-rs/blob/e6bc3ef633929cba65250f36a50c4cb997bbc128/fixtures/coverall/tests/bindings/test_coverall.py#L9

mhammond (approved):

PR #452, opened 2021-05-21
Remove support for non-threadsafe interfaces.

### LINKED ISSUE #449: Remove support for non-threadsafe interfaces
In #429 we officially deprecated the use of non-threadsafe interfaces; currently all interfaces are expected to have an explicit `[Threadsafe]` attribute and we'll issue a warning if they don't.

Per [ADR-0004](https://github.com/mozilla/uniffi-rs/blob/main/docs/adr/0004-only-threadsafe-interfaces.md), the intended end state is for all interfaces to be thread-safe by default, regardless of the presence of a `[Threadsafe]` attribute. This issue tracks the remaining work of getting to that state, which includes:

* [x] Cut at least one release with the deprecation notice for non-threadsafe interfaces in place.
* [ ] Audit all known internal consumers to make sure they've updated to using `[Threadsafe]`.
* [ ] Implement the removal, which will involve:
  * [ ] Removing the existing deprecation notice on non-threadsafe interfaces.
  * [ ] Deleting the `Object::threadsafe` method.
  * [ ] Finding all the places where we branch on `obj.threadsafe()`, and deleting the falsey branch.
  * [ ] Adding a new deprecation notice that warns when `[Threadsafe]` is present, since it's now redundant and confusing.
  * [ ] Simplifying the current "threadsafe" example and moving it into the test-fixtures directory
  * [ ] Updating our other examples so they don't give deprecation warnings.
* [ ] Cut at least one release with the new deprecation notice in place.
* [ ] Eventually, entirely remove support the for the `[Threadsafe]` attribute, making it a syntax error.

I'd like to treat this removal as a blocker for landing the Arc-pointer conversation proposed in [ADR-0005](https://github.com/mozilla/uniffi-rs/pull/430), so that we don't need to add code for wrapping non-threadsafe interfaces in a `Mutex` before putting them in an `Arc`.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2218)

rfk:
Fixes #449. Depends on #451. This will be a draft until we've given the v0.9.0 release a chance to roll out to consumers.

With the v0.9.0 release of UniFFI, non-threadsafe interfaces are explicitly deprecated. This commit removes support for non-threadsafe interfaces entirely, making threadsafety the default and only supported mode. In practice, for any `interface` declared in the UDL:

* Its implementation must be `Sync` and `Send`.
* Its methods cannot accept a mutable `&self` parameter.

The `[Threadsafe]` attribute may still be specified explicitly in the UDL file, but will result in a deprecation warning and is otherwise ignored.

We don't have a good way to write tests for the following, but I've manually verified that:

* A non-`Send+Sync` interface implementation produces a compile-time error.
* Use of the `[Threadsafe]` attribute results in a deprecation warning.


fixtures/coverall/tests/bindings/test_coverall.kts
```diff
@@ -28,4 +30,37 @@ fun Double.almostEquals(other: Double) = Math.abs(this - other) < 0.000001
...
+assert(d.maybeFloat64!!.almostEquals(1.0))
+
+// This tests that the UniFFI-generated scaffolding doesn't introduce any unexpected locking.
+// We have one thread busy-wait for a some period of time, while a second thread repeatedly
+// increments the counter and then checks if the object is still busy. The second thread should
+// not be blocked on the first, and should reliably observe the first thread being busy.
+// If it does not, that suggests UniFFI is accidentally serializing the two threads on access
+// to the shared counter object.
+
+ThreadsafeCounter().use { counter ->
```
rfk, line 42: I removed some layers of abstraction from this test, and moved it from "examples" to "fixtures". The abstractions seemed less necessary now that it's not trying to test both the threadsafe and non-threadsafe scenarios.

rfk:
This is ready for review, but I don't plan to land it with any urgency. My main motivation here is to simplify work on the Arc-pointers effort, and in particular, to simplify the *docs* will that will need to be written for that effort. I'm not really interested in trying to explain a different version of the automagic-mutex behaviour in addition to rambling on about Arcs.

docs/manual/src/internals/object_references.md
```diff
@@ -10,21 +10,35 @@ including:
...
+that foreign-language callers will be "well behaved".
+
+UniFFI's hands-off approach means that all object instances exposed by UniFFI must be safe to
+access concurrently. In Rust terminology, they must be `Send+Sync` and must be useable
+without taking any `&mut` references.
+
+Typically this will mean that the Rust implementation of an object uses some of Rust's
+data structures for thread-safe interior mutability, such as a `Mutex` or `RwLock` or
+the types from `std::atomic`. The precise details are completely up to the author
+of the component - as much as possible, UniFFI tries to stay out of your way, simply requiring
```
mhammond, line 26: 👍

docs/manual/src/internals/object_references.md
```diff
@@ -85,21 +99,4 @@ This indirection gives us some important safety properties:
...
-
-For instances that are explicited tagged with the `[Threadsafe]` attribute, uniffi instead uses
-a custom `ArcHandleMap` struct. This replaces the run-time `Mutex` with compile-time assertions
-about the safety of the underlying Rust struct. Specifically:
-
-* The `ArcHandleMap` will never give out a mutable reference to an instance, forcing the
-  underlying struct to use interior mutability and manage its own locking.
-* The `ArcHandleMap` can only contain structs that are `Sync` and `Send`, ensuring that
-  shared references can safely be accessed from multiple threads.
+  by enforcing that the contained type is `Send+Sync`, and by refusing to hand out any mutable references.
```
mhammond: s/enforcing that the contained type is/specifying that the contained type must be/ or something? "enforcing" makes it sound like some kind of check the map itself is doing, but it's the compiler that is actually enforcing.

docs/manual/src/udl/interfaces.md
```diff
@@ -40,11 +40,9 @@ interface TodoList {
 
 By convention, the `constructor()` calls the Rust's `new()` method.
 
-Conceptually, these `interface` objects are live Rust objects that have a proxy on the foreign language side; calling any methods on them, including a constructor or destructor results in the corresponding methods being call in Rust.
+Conceptually, these `interface` objects are live Rust structs that have a proxy object on the foreign language side; calling any methods on them, including a constructor or destructor results in the corresponding methods being call in Rust.
```
mhammond, line 43: typo: called

docs/manual/src/udl/interfaces.md
```diff
@@ -40,11 +40,9 @@ interface TodoList {
...
 
-`uniffi` will generate these proxies of live objects with an interface or protocol.
-
-e.g. in Kotlin.
+UniFFI will generate these proxies of live objects with an interface or protocol to help with testing in the foreign-language code. For example in Kotlin, the `TodoList` would generate:
```
mhammond, line 45: nit: "of live objects" can probably be dropped from this sentence?

examples/sprites/src/lib.rs
```diff
@@ -31,8 +31,8 @@ pub fn translate(p: &Point, v: Vector) -> Point {
 // and which can move about over time.
 #[derive(Debug)]
 pub struct Sprite {
-    // All interfaces must be threadsafe, meaning `Send+Sync`, and our naive
-    // `Point` struct isn't - so we wrap it in a `RwLock` to make it so.
+    // All interfaces must be `Send + Sync`, and our naive `Point` struct
```
mhammond, line 34: nit: almost everywhere else has no space between Send and Sync. Not sure which one is actually preferred, but we might as well be consistent.

docs/manual/src/internals/object_references.md
```diff
@@ -10,21 +10,35 @@ including:
...
 * Guarding against data races.
 
-UniFFI aims to maintain these guarantees even when the Rust code is being invoked
-from a foreign language, at the cost of turning them into run-time checks rather
-than compile-time guarantees.
+The very nature of the problems UniFFI tries to solve is that calls may come
+from foreign languages on any thread, outside of the control of Rust's ownership
+system. UniFFI itself tries to take a hands-off approach as much as possible and
+depends on the Rust compiler itself to uphold safety guarantees, without assuming
+that foreign-language callers will be "well behaved".
```
rfk, line 17: @mhammond FYI I've liberally appropriated the docs updates from your #420.

mhammond: great - so the nits below are probably mine :)

mhammond (approved):
Lovely!

README.md
```diff
@@ -68,27 +68,27 @@ starting with [this motivational document](./docs/adr/0000-whats-the-big-idea.md
...
-commit](https://github.com/mozilla/uniffi-rs/commit/454dfff6aa560dffad980a9258853108a44d5985)
+Version 0.9.0 took this further, and interfaces not marked as `[Threadsafe]` 
+started issuing a deprecation warning. If you are seeing these deprecation warnings,
+you should upgrade your component as soon as possible. For an example of what kind of
+effort is required to make your interfaces thread-safe, you might like to see
+[this commit](https://github.com/mozilla/uniffi-rs/commit/454dfff6aa560dffad980a9258853108a44d5985)
 where we made one the examples thread-safe.
 
+As of version 0.10.0, all interfaces will be required to be  `Send+Sync`, and the
+`[Threadsafe]` attribute is will be deprecated and ignored.
```
npars, line 90:
```suggestion
`[Threadsafe]` attribute will be deprecated and ignored.
```

rfk: Nice catch, thanks 👍🏻

rfk:
Let's push ahead with this; if we feel like we need further releases in the `v0.10` series we can manage them on a branch.

PR #454, opened 2021-05-22
Update dependencies

r-ml:
Was taking a look my dependency tree and noticed uniffi is still on `bytes 0.5`.

Hopefully this fixes that :)

### DIFF
--- a/uniffi/src/lib.rs
+++ b/uniffi/src/lib.rs
@@ -347,7 +347,7 @@ unsafe impl ViaFfi for String {
         check_remaining(buf, 4)?;
         let len = usize::try_from(buf.get_i32())?;
         check_remaining(buf, len)?;
-        let bytes = &buf.bytes()[..len];
+        let bytes = &buf.chunk()[..len];
         let res = String::from_utf8(bytes.to_vec())?;
         buf.advance(len);
         Ok(res)
--- a/uniffi/src/testing.rs
+++ b/uniffi/src/testing.rs
@@ -120,17 +120,12 @@ pub fn ensure_compiled_cdylib(pkg_dir: &str) -> Result<String> {
     let cdylib_files: Vec<_> = cdylib
         .filenames
         .iter()
-        .filter(|nm| {
-            matches!(
-                nm.extension().unwrap_or_default().to_str(),
-                Some(std::env::consts::DLL_EXTENSION)
-            )
-        })
+        .filter(|nm| matches!(nm.extension(), Some(std::env::consts::DLL_EXTENSION)))
         .collect();
     if cdylib_files.len() != 1 {
         bail!("Failed to build exactly one cdylib file, it must not be a uniffi component");
     }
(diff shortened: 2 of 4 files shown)


uniffi/src/lib.rs
```diff
@@ -347,7 +347,7 @@ unsafe impl ViaFfi for String {
         check_remaining(buf, 4)?;
         let len = usize::try_from(buf.get_i32())?;
         check_remaining(buf, len)?;
-        let bytes = &buf.bytes()[..len];
+        let bytes = &buf.chunk()[..len];
```
rfk, line 350: Hrm...the docs for the [`chunk` method](https://docs.rs/bytes/1.0.1/bytes/trait.Buf.html#method.chunk) note that "this can return shorter slice (this allows non-continuous internal representation)" which IIUC would risk a panic here. In practice though we only ever use this with an underlying `[u8]` which doesn't have a non-continuous internal representation. So I think this is fine, but I'll make a followup to try to make that a little more explicit in the code. Thanks!

uniffi/src/testing.rs
```diff
@@ -120,17 +120,12 @@ pub fn ensure_compiled_cdylib(pkg_dir: &str) -> Result<String> {
     let cdylib_files: Vec<_> = cdylib
         .filenames
         .iter()
-        .filter(|nm| {
-            matches!(
-                nm.extension().unwrap_or_default().to_str(),
-                Some(std::env::consts::DLL_EXTENSION)
-            )
-        })
+        .filter(|nm| matches!(nm.extension(), Some(std::env::consts::DLL_EXTENSION)))
```
rfk, line 123: Out of curiosity, were these simplifications done by clippy or similar, or something that you discovered by hand?

rfk (approved):
Thanks!

uniffi/src/testing.rs
r-ml: Neither, these were `rustc` errors. It appears these returned `Option<Cow<str>>` before and now just `Option<&str>`

uniffi/src/lib.rs
r-ml: Is it not the same as before? \
The [previous docs](https://docs.rs/bytes/0.5.6/bytes/trait.Buf.html#tymethod.bytes) have the same warning.

r-ml: I'll checkout your PR and try to decipher your statement, thanks! :)

rfk: Thanks for checking - it seems very likely that the old version also had this behavior and I just didn't notice!

rfk: FWIW, if you do take a look, I'm curious whether the commentary I added in https://github.com/mozilla/uniffi-rs/pull/455 made any sense.

r-ml: After navigating a bit through the rest of the library, it does. \
Thanks!

PR #455, opened 2021-05-24
Implement try_read/write for concrete types, not Buf/BufMut generics.

rfk:
This is a followup to the upgrade of our `bytes` dependency from #454.

Our use of the `bytes::Buf` trait assumes that `Buf::chunks` will return
all the remaining data. This is not guaranteed by the trait, but it is
guaranteed in practice by its implementation on the concrete types that
we use.

Since we don't really work with any generic `Buf` impl anyway, this commit
removes the generics from `ViaFfi::try_read` and `ViaFii:write`, replacing
them with the single concrete type that we actually ues in practice. This
should reduce potential for future confusion, and also make compilation
slightly easier on the Rust compiler.

### DIFF
--- a/uniffi/src/lib.rs
+++ b/uniffi/src/lib.rs
@@ -179,7 +179,7 @@ pub unsafe trait ViaFfi: Sized {
     /// This trait method can be used for sending data from rust to the foreign language code,
     /// in cases where we're not able to use a special-purpose FFI type and must fall back to
     /// sending serialized bytes.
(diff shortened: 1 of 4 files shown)


rfk:
@lougeniaC64 could you please take a look at this one? I'm trying to spread out the review load of my recent UniFFI patches a little more evenly.

mhammond (approved):
this makes sense to me - generics that aren't used in practice just cause confusion.

PR #456, opened 2021-05-24
Implement `[Self=ByArc]` annotation for methods on threadsafe objects.

rfk:
As of ADR-0005, the use of `Arc<T>` for passing object instances will
be part of the official Rust-facing  API of UniFFI rather than being
a private implementation detail. A full implementation of that ADR is
still in progress, but in the meantime we can get a relatively easy
implementation of one small part of it: support for `Arc<Self>` as
method receiver.

This commit implements support for the `[ByArc]` annotation on method
definitions, indicating they the underlying Rust code will accept
`self: Arc<Self>` rather than `self: &Self` as method receiver. This
only works for `[Threadsafe]` interfaces since they are the only ones
that use `Arc<Self>` internally, but that's OK because we intend to
require that all interfaces are threadsafe in the near future.

The implementation is small, but not particularly pretty. I think
it's fine to push forward with given our broader plans to simplify
away a lot of the branching in this code.

### DIFF
--- a/fixtures/coverall/src/coverall.udl
+++ b/fixtures/coverall/src/coverall.udl
@@ -47,4 +47,10 @@ interface Coveralls {
     boolean maybe_throw(boolean should_throw);
 
     void panic(string message);
+
+    // *** Test functions which take either `self` or other params as `Arc<Self>` ***
+
+    /// Calls `Arc::strong_count()` on the `Arc` containing `self`.
+    [Self=ByArc]
+    u64 strong_count();
 };
--- a/uniffi_bindgen/src/interface/attributes.rs
+++ b/uniffi_bindgen/src/interface/attributes.rs
@@ -29,6 +29,7 @@ pub(super) enum Attribute {
     Enum,
     Error,
     Name(String),
+    SelfType(SelfType),
     Threadsafe,
     Throws(String),
 }
@@ -63,6 +64,7 @@ impl TryFrom<&weedle::attribute::ExtendedAttribute<'_>> for Attribute {
                 match identity.lhs_identifier.0 {
                     "Name" => Ok(Attribute::Name(name_from_id_or_string(&identity.rhs))),
                     "Throws" => Ok(Attribute::Throws(name_from_id_or_string(&identity.rhs))),
+                    "Self" => Ok(Attribute::SelfType(SelfType::try_from(&identity.rhs)?)),
                     _ => anyhow::bail!(
                         "Attribute identity Identifier not supported: {:?}",
                         identity.lhs_identifier.0
@@ -170,7 +172,7 @@ impl TryFrom<&weedle::attribute::ExtendedAttributeList<'_>> for FunctionAttribut
     ) -> Result<Self, Self::Error> {
         let attrs = parse_attributes(weedle_attributes, |attr| match attr {
             Attribute::Throws(_) => Ok(()),
-            _ => bail!(format!("{:?} not supported for functions or methods", attr)),
+            _ => bail!(format!("{:?} not supported for functions", attr)),
         })?;
         Ok(Self(attrs))
     }
@@ -272,15 +274,10 @@ impl<T: TryInto<InterfaceAttributes, Error = anyhow::Error>> TryFrom<Option<T>>
     }
 }
 
-// There may be some divergence between Methods and Functions at some point,
-// but not yet.
-pub(super) type MethodAttributes = FunctionAttributes;
-
-// Constructors can have a `Name` attribute, so need their own implementation.
-/// Represents UDL attributes that might appear on a function.
+/// Represents UDL attributes that might appear on a constructor.
 ///
-/// This supports the `[Throws=ErrorName]` attribute for functions that
-/// can produce an error.
+/// This supports the `[Throws=ErrorName]` attribute for constructors that can produce
+/// an error, and the `[Name=MethodName]` for non-default constructors.
 #[derive(Debug, Clone, Hash, Default)]
 pub(super) struct ConstructorAttributes(Vec<Attribute>);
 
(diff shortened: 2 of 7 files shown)


fixtures/coverall/src/coverall.udl
```diff
@@ -47,4 +47,10 @@ interface Coveralls {
     boolean maybe_throw(boolean should_throw);
 
     void panic(string message);
+
+    // *** Test functions which take either `self` or other params as `Arc<Self>` ***
+
+    /// Calls `Arc::strong_count()` on the `Arc` containing `self`.
+    [ByArc]
```
rfk, line 54: 🚨🚲🏡🚨  **bikeshed alert** 🚨🏡🚲🚨

I'm not 100% happy with this name in this context - unless you already know what it is, it's hard to tell by reading just exact *what* will be dealt with "by arc". Something like `[SelfAsArc]` or `[Self=ByArc]` or similar might help on that front...but OTOH those are both kind of an eyesore...

I don't feel very strongly about it, and maybe it'll be rare enough that it doesn't really matter anyway. Just kinda thinking out loud about it.

uniffi_bindgen/src/interface/attributes.rs
```diff
@@ -272,15 +274,10 @@ impl<T: TryInto<InterfaceAttributes, Error = anyhow::Error>> TryFrom<Option<T>>
     }
 }
 
-// There may be some divergence between Methods and Functions at some point,
-// but not yet.
```
rfk, line 276: Turns out this is the point at which there is divergence between Methods and Functions.

uniffi_bindgen/src/templates/macros.rs
```diff
@@ -65,12 +65,12 @@ use uniffi::UniffiMethodCall;
 {%- endif -%}
 {% match meth.throws() -%}
 {% when Some with (e) -%}
-{{ this_handle_map }}.method_call_with_result(err, {{ meth.first_argument().name() }}, |obj| -> Result<{% call return_type_func(meth) %}, {{e}}> {
+{{ this_handle_map }}.method_call_{% if meth.takes_self_by_arc() %}by_arc_{% endif %}with_result(err, {{ meth.first_argument().name() }}, |obj| -> Result<{% call return_type_func(meth) %}, {{e}}> {
```
rfk, line 68: 🤮 at my own work here, but if it's useful to get this in early, I think we can tolerate this for a short while.

fixtures/coverall/src/coverall.udl
mhammond: Yeah, I had the same response. I didn't really like the variants with `Self` in the name because I kinda pictured it might be used in contexts other than `self` (eg, once we start allowing interfaces as params etc). OTOH though, it seems unlikely we will ever support these params etc being anything other than as `Arc`, and in my WIP no such annotation was necessary for those contexts.

That said though, I guess I'm vaguely preferring `[Self=ByArc]` and if we did need annotations on params in the future, we'd reuse `ByArc` there.

(It does seem worth bike shedding this now)

uniffi_bindgen/src/templates/macros.rs
mhammond: heh

mhammond (approved):
This is great! As mentioned though, future us is likely to thank is if we bikeshed a little on the attribute.

fixtures/coverall/src/coverall.udl
mhammond: > I guess I'm vaguely preferring `[Self=ByArc]`

OTOH, I thought that before, so went looking for things other than `ByArc` that might be on `Self=` and came up blank - and thought that `[Self=ByArc]` when it's almost certain there will be no other value allowed there seemed dumb too. So I ended up paralyzed making a decision and moved on. And here we are again :)

rfk: > (It does seem worth bike shedding this now)

Lol this is all the encouragement I need ;-)

> I guess I'm vaguely preferring [Self=ByArc] and if we did need annotations on params in the future, we'd reuse ByArc there.

I think this is where I've ended up as well - it's not *great*, but it reads more descriptively, it kind of fits with the ad-hoc convention we invented with `ByRef`, and would extend naturally to adding `ByArc` in other places. Might be the best of an uninspiring bunch. I think I'm willing to accept the weirdness of only having a single allowed value, just because of the way it reads better.

Perhaps our future selves will not have this problem because we're just reading `self: Arc<Self>` directly out of the underlying Rust source code 😁

rfk:
Alrighty, I've updated with support for `[Self=ByArc]` syntax, which I do find myself liking better. I also went ahead and added support for `[Self=ByRef]` for symmetry, although it's currently parsed and ignored because that's the default behaviour. @mhammond want to give it one quick final look?

uniffi_bindgen/src/interface/attributes.rs
```diff
@@ -315,6 +313,78 @@ impl TryFrom<&weedle::attribute::ExtendedAttributeList<'_>> for ConstructorAttri
...
+        }
+    }
+}
+
+/// Represents the different possible types of method call receiver.
+///
+#[derive(Debug, Clone, Hash)]
+pub(super) enum SelfType {
+    ByArc, // Method receiver is `Arc<Self>`.
+    ByRef, // Method receiver is `&self`.
```
mhammond, line 369: I don't feel too strongly about this, but I fear supporting `ByRef` here might just cause confusion. IIRC, there will never be a reason to specify it, which will confuse when someone specifies it anyway. Then some people will specify it everywhere as a convention, then get confused when *not* specifying doesn't mean the receiver consumes self (ie, they might expect support for  `self: Self`)

mhammond (approved):
👍

uniffi_bindgen/src/interface/attributes.rs
rfk: That's fair; I'll remove it before merge, it's easy to bring back later if we need it.

Issue #457, opened 2021-05-24
Kotlin method calls and object destcruction can race, with crashy results


rfk:
Unlike our other foreign-language bindings, Kotlin objects need to be explicitly destroyed rather than being freed automatically by the garbage collector. This means that if calling code is not well-behaved, we might end up with a method call in flight on one thread while the object is being destroyed on a different thread [1].

The way we [use an `AtomicLong`](https://github.com/mozilla/uniffi-rs/blob/307ad92274522f1845beb8d4e2c6461cb16d3870/uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt#L6) for object handles is not sufficient to guard against races here. It's possible for us to execute the following sequence of events:

* Thread A reads the object handle from the `AtomicLong`, then goes to sleep.
* Thread B calls `destroy()` on the Kotlin object, which frees the underlying Rust object and sets the `AtomicLong` handle to zero.
* Thread A wakes up and proceeds with its method call, using the now-stale handle to try to operate on the now-freed Rust object.

With our current implementation, this will cause the Rust code to bail out with a "stale handle" error. With the proposed move to using raw Arc pointers, this would be a use-after-free.

This is unlikely to cause us any issues in practice right now, because the way consumers currently use our objects means they don't ever really get destroyed. Still, we should fix it sooner rather than later, because it should never be possible to trigger unexpected behaviour in the Rust code by calling the generated bindings.

[1] Actually I think we'd have the same bug even if we were using finalizers to destroy objects automatically; the JVM apparently gets super aggressive about finalization and will [happily finalize an object while there's an active call to one of its methods](https://stackoverflow.com/questions/24376768/can-java-finalize-an-object-when-it-is-still-in-scope/24380219), as long as it can prove that the method call will not make any further references to the object itself.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2220)

rfk:
I was able to pretty easily confirm this bug on my machine, using the test case in https://github.com/mozilla/uniffi-rs/pull/458. Basically, it runs a tight loop of:

* Create an object on Thread A.
* Spawn a method call on the object on Thread B.
* Destroy the object on Thread B.

This reliably fails on my machine with:

```
java.util.concurrent.ExecutionException: uniffi.threadsafe.InternalException: Handle has stale version number
        at java.util.concurrent.FutureTask.report(FutureTask.java:122)
[...]

PR #458, opened 2021-05-24
Guard against race between Kotlin method calls and object destruction

rfk:
Currently just a test case that reproduces the issue on my machine, but I'm thinking about a fix.

(Depends on the Kotlin codegen fix from https://github.com/mozilla/uniffi-rs/pull/451)

### DIFF
--- a/fixtures/coverall/tests/bindings/test_handlerace.kts
+++ b/fixtures/coverall/tests/bindings/test_handlerace.kts
@@ -1,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import java.util.concurrent.*
+import uniffi.coverall.*
+
+// This tests the thread-safety of operations on object handles.
+// The idea is to have a method call running on one thread while
+// we try to destroy the object on another thread. This should
+// be handled correctly at the Kotlin level, rather than triggering
+// any fail-safe checks (or worse, memory unsafety!) in the underlying
+// Rust code.
+//
+// This test was able to reproduce the concurrency issue in
+// https://github.com/mozilla/uniffi-rs/issues/457, and is kept
+// to prevent regressions.
+
+// Give ourselves enough opportunity to trigger concurrency issues.
+for (i in 1..1000) {
+    // So we can operate concurrently.
+    val executor = Executors.newFixedThreadPool(2)
+    try {
+        // On the main thread here, we're going to create and then destroy a `Coveralls` instance.
+        val concurrentMethodCall: Future<String> = Coveralls("test_handlerace").use { coveralls ->
+            // Make a method call on a separate thread thread.
+            val concurrentMethodCall = executor.submit(Callable {
+                coveralls.getName()
+            })
+            // Sleep a little, to give the background thread a chance to start operating.
+            Thread.sleep(1)
+            concurrentMethodCall
+        }
+        // At this point the object has been destroyed.
+        // The concurrent method call should either have failed to run at all (if the object was destroyed
+        // before it started) or should have completely succeeded (if it started before the object was
+        // destroyed). It should not fail in the Rust code (which currently fails cleanly on handlemap
+        // safety checks, but in future might be a use-after-free).
+        try {
+            concurrentMethodCall.get()
+        } catch (e: ExecutionException) {
+            if (e.cause is IllegalStateException && e.message!!.endsWith("has already been destroyed")) {
+                // This is fine, we rejected the call after the object had been destroyed.
+            } else {
+                // Any other behaiour signals a possible problem.
+                throw e
+            }
+        }
+    } finally {
+        executor.shutdown()
+    }
+}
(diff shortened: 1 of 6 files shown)


rfk:
(I'm happy to note that the testcase reproduces in CI)

fixtures/coverall/tests/bindings/test_coverall.kts
```diff
@@ -0,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk, line 3: (note that these tests are from the commit in https://github.com/mozilla/uniffi-rs/pull/451, I just needed to get that fix in order to write the new test for this PR)

fixtures/coverall/tests/bindings/test_handlerace.kts
```diff
@@ -0,0 +1,52 @@
...
+// This tests the thread-safety of operations on object handles.
+// The idea is to have a method call running on one thread while
+// we try to destroy the object on another thread. This should
+// be handled correctly at the Kotlin level, rather than triggering
+// any fail-safe checks (or worse, memory unsafety!) in the underlying
+// Rust code.
+//
+// This test was able to reproduce the concurrency issue in
+// https://github.com/mozilla/uniffi-rs/issues/457, and is kept
+// to prevent regressions.
```
rfk, line 17: Ref https://github.com/mozilla/uniffi-rs/issues/457#issuecomment-846747337 for the failure that was previously being demonstrated by this test.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -2,21 +2,130 @@
...
+// possible for the JVM to finalize an object while there is an in-flight call to one of its methods [1],
+// so we would not be able to avoid all the complexity here).
+//
+// Sigh...all of this for want of a robust finalization mechanism.
+//
+// [1] https://stackoverflow.com/questions/24376768/can-java-finalize-an-object-when-it-is-still-in-scope/24380219
+//
 abstract class FFIObject(
-    private val handle: AtomicLong
+    protected val handle: Long
```
rfk, line 87: I didn't see any value in keeping this in an `AtomicLong` since it's set at initialization time and never changed, and we're no longer using it for coordination.

rfk (commented):
A self-review with some extra commentary:

rfk:
So, I have a fix that I think will address #457 and that I'm *mostly* happy with, but I don't fully trust myself to hit the right complexity tradeoff here. @jhugman @mhammond would you mind giving me a gutcheck on what's proposed here (as well as on the inline commentary about some considered alternatives)?

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -2,21 +2,131 @@
...
+//      before it can pass the handle over the FFI to Rust.
+//    * Thread B calls `destroy` and frees the underlying Rust struct.
+//    * Thread A resumes, passing the already-read handle value to Rust and triggering
+//      a use-after-free.
+//
+// One possible solution would be to use a `ReadWriteLock`, with each method call taking
+// a read lock (and thus allowed to run concurrently) and the special `destroy` method
+// taking a write lock (and thus blocking on live method calls). However, we aim not to
+// generate methods with any hidden blocking semantics, and a `destroy` method that might
+// block if called incorrectly seems to meet that bar.
```
rfk, line 48: See here ^ for an alternative approach I considered.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -2,21 +2,131 @@
...
+
+    open protected fun freeHandle() {
+        // To be overridden in subclasses.
+    }
+
     open fun destroy() {
-        this.handle.set(0L)
+        // Only allow a single call to this method.
+        // TODO: maybe we should log a warning if called more than once?
+        if (this.wasDestroyed.compareAndSet(false, true)) {
```
rfk, line 100: If we wanted to, we could keep `this.handle` in an `AtomicLong` and use that as the "was destroy already called" flag, setting it to zero on the first call to `destroy()`. I opted for a separate flag here in order to have only a single atomic in the method call logic, but am open to changing it if you think fewer variables would make things simpler over all.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -2,21 +2,131 @@
...
+    val callCounter = AtomicLong(1)
+
+    open protected fun freeHandle() {
+        // To be overridden in subclasses.
+    }
+
     open fun destroy() {
-        this.handle.set(0L)
+        // Only allow a single call to this method.
+        // TODO: maybe we should log a warning if called more than once?
```
mhammond, line 99: TBH, I'd consider throwing an `IllegalStateException` exception.

mhammond (approved):
It's a bit hairy, but it seems sound.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
rfk: Yeah. IIUC the motivation for the current behaviour is that we'd much rather than `destroy()` be called more than once, then not be called at all (and we even have some tests to assert this behaviour).

rfk:
I didn't get a chance to sync up with James about this, but I'm going to go ahead and land it based on Mark's review so that it doesn't bitrot.

PR #459, opened 2021-05-25
Start keeping a changelog.

rfk:
If we're going to be issuing deprecation notices and such, we really ought to
be keeping a changelog. This commit adds one, with some retroactive notes from
the `v0.9.0` release.

I had hoped to be able to automatically manage updating the section headers
in the changelog using `cargo release`, but that doesn't seem to work as I
hoped when releasing multiple crates in a single workspace. Still, let's
organize it as though we'll be using `cargo release` for this one day.

### DIFF
--- a/release.toml
+++ b/release.toml
@@ -7,3 +7,13 @@ consolidate-commits = true
 # So we have to tag manually for now.
 disable-tag = true
 disable-push = true
+disable-publish = true
+
+# This is how we'd *like* to manage the sections in CHANGELOG.md, but it doesn't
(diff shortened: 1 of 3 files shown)


mhammond (approved):

release.toml
```diff
@@ -7,3 +7,13 @@ consolidate-commits = true
 # So we have to tag manually for now.
 disable-tag = true
 disable-push = true
+disable-publish = true
```
rfk, line 10: whoops 😬

PR #462, opened 2021-05-25
Allow passing object instances as args, fields and return values.

rfk:
This implements the proposal from [ADR-0005](https://github.com/mozilla/uniffi-rs/pull/430),
following discussion in #419. It depends on the removal on non-threadsafe interfaces
from #452.

This is a significant re-working of how object instances are handled,
replacing our previous use of handlemaps with direct use of an `Arc<T>`
pointer. This both reduces overhead and allows us to must more easily
generate code for dealing with object references.

On the Rust side of the world, code that needs to deal with an
object reference will typically take an `Arc<T>`, in the spirit
of being as explicit as possible. The one exception is when passing
objects as arguments, where (like other types) you can annotate the
argument with `[ByRef]` and have UniFFI hand you an `&T`.

For an example of how this might be used in practice, the "todolist"
example has grown the ability to manage a global default list by
setting and retreiving a shared object reference.


examples/todolist/src/lib.rs
```diff
@@ -2,13 +2,19 @@
...
 #[derive(Debug, Clone)]
 pub struct TodoEntry {
     text: String,
 }
 
+lazy_static::lazy_static! {
+    // There is a single "default" TodoList that can be shared
+    // by all consumers of this component.
+    static ref DEFAULT_LIST: RwLock<Option<Arc<TodoList>>> = RwLock::new(None);
+}
```
rfk, line 16: I've updated to "todolist" example to demonstrate how I think we could use this in something like Glean or Nimbus in the future, managing a global singleton that's initialized by the application code but can be accessed by multiple consumers.

examples/todolist/tests/bindings/test_todolist.kts
```diff
@@ -50,6 +50,29 @@ todo.addItems(listOf("bobo", "fofo"))
...
+    }
+
+    todo.addItem("Test liveness after being demoted from default")
+    assert(todo.getLast() == "Test liveness after being demoted from default")
+
+    todo2.addItem("Test shared state through local vs default reference")
+    getDefaultList()!!.use { default ->
+        assert(default.getLast() == "Test shared state through local vs default reference")
+    }
+}
```
rfk, line 74: This test shows a bit of the awkwardness of working with object references in Kotlin, where we have to make sure each instance is explicitly destroyed when we're finished with it. It's manageable, but easy to forget about. I'd like to see us take a fresh look at e.g. weak references for doing this more automatically, but that's well outside the scope of this PR.

examples/todolist/tests/bindings/test_todolist.py
```diff
@@ -23,3 +23,28 @@
...
+assert(todo != todo2)
+assert(todo is not todo2)
+
+assert(get_default_list() is None)
+
+set_default_list(todo)
+assert(todo.get_items() == get_default_list().get_items())
+# In the future, we'll maintain object identity over the FFI.
+#assert(get_default_list() is todo)
+#assert(get_default_list() is not todo2)
```
rfk, line 37: Currently, every object instance returned over the FFI is a fresh foreign-language object. Mark's [prototype PR](https://github.com/mozilla/uniffi-rs/pull/420) implemented a reverse-identity-map for the Python backend using weak references and I think we could do something similar for the other languages, but on balance I think it will be better for us to try out the basic functionality without blocking on that extra complexity. Happy to work on it in a followup PR, and I'm open to the suggestion that we shouldn't land without it because this "every object is a new one" behaviour could be confusing for consumers.

(There's also the question of how a reverse-identity-map would interact with the Kotlin `destroy()`/`.use` pattern, which I really don't have my head around just yet.)

fixtures/coverall/src/coverall.udl
```diff
@@ -22,6 +22,13 @@ dictionary SimpleDict {
     float? maybe_float32;
     double float64;
     double? maybe_float64;
+    Coveralls? coveralls;
+};
+
+[Enum]
+interface MaybeSimpleDict {
+    Yeah(SimpleDict d);
+    Nah();
```
rfk, line 31: This tests codegen for nested object references, I don't have any tests that actually exercise it yet.

fixtures/coverall/tests/bindings/test_coverall.kts
```diff
@@ -6,31 +6,100 @@ import java.util.concurrent.*
...
-fun Float.almostEquals(other: Float) = Math.abs(this - other) < 0.000001
-fun Double.almostEquals(other: Double) = Math.abs(this - other) < 0.000001
-
-assert(d.float32.almostEquals(1.2345F))
-assert(d.maybeFloat32!!.almostEquals(22.0F/7.0F))
-assert(d.float64.almostEquals(0.0))
-assert(d.maybeFloat64!!.almostEquals(1.0))
+// Test some_dict().
+// N.B. we need to `use` here to clean up the contained `Coveralls` reference.
+createSomeDict().use { d ->
```
rfk, line 13: Because the `SimpleDict` type contains an object reference, it too provides helpers for manual management of the object lifecycle.

fixtures/coverall/tests/bindings/test_coverall.rb
```diff
@@ -52,6 +55,7 @@ def test_none_dict
   end
 
   def test_constructors
+    GC.start
```
rfk, line 58: AFAICT the Ruby garbage collector is not deterministic in freeing objects when they go out of scope, so we have to manually trigger collections at key points where we want to test whether things have been cleaned up correctly.

fixtures/coverall/tests/bindings/test_coverall.rb
```diff
@@ -109,4 +113,93 @@ def test_self_by_arc
...
+    assert_equal 3, coveralls.strong_count
+    # but the same number of instances.
+    assert_equal 1,  Coverall.get_num_alive
+    # and check it's the correct object.
+    assert_equal "test_arcs",  coveralls.get_other.get_name
+    # TODO: when we implement an identity map, it will even be the same Ruby object
+    # self.assertTrue(coveralls.get_other() is coveralls)
+
+    # TODO: using `assert_raise` seems to keep a reference to `coveralls` alive,
+    # breaking the reference-count checks further down.
```
rfk, line 136: @saks if you have a chance to investigate this, it'd be appreciated - if I try to write the test cases here with `assert_raise` then it seems to keep the value from `coveralls` alive, interfering with the attempt to test whether its references are being cleaned up. I suspect maybe `assert_raises` is stashing the raised exception somewhere, which has a reference to the local variables including `coveralls`?

fixtures/coverall/tests/bindings/test_coverall.swift
```diff
@@ -0,0 +1,103 @@
...
+    } catch CoverallError.TooManyHoles {
+        // It's okay!
+    }
+    // TODO: kinda hard to test this, as it triggers a fatal error.
+    // coveralls!.takeOtherPanic(message: "expected panic: with an arc!")
+    coveralls.takeOther(other: nil);
+    assert(coveralls.strongCount() == 2);
+}
+// Swift GC is deterministic, `coveralls` is freed when it goes out of scope.
+assert(getNumAlive() == 0);
```
rfk, line 73: By contrast, the Swift GC is based on reference-counting so this is all totally deterministic \o/

uniffi/Cargo.toml
```diff
@@ -14,7 +14,7 @@ keywords = ["ffi", "bindgen"]
 # Re-exported dependencies used in generated Rust scaffolding files.
 anyhow = "1"
 bytes = "1.0"
-ffi-support = "~0.4.2"
+ffi-support = "~0.4.3"
```
rfk, line 17: This gets us an automatic `impl IntoFfi for c_void`, ref https://github.com/mozilla/ffi-support/pull/4

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -78,11 +78,28 @@ sealed class {{ e.name()|class_name_kt }} {
...
-                {% endfor -%}
+                {% endfor %}
             }
             {%- endfor %}
         }.let { /* this makes the `when` an expression, which ensures it is exhaustive */ }
     }
+
+    {% if e.contains_object_references(ci) %}
+    @Suppress("UNNECESSARY_SAFE_CALL") // codegen is much simpler if we unconditionally emit safe calls here
+    override fun destroy() {
```
rfk, line 89: The new helper for things that might recursively contain object references.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -32,13 +56,13 @@
...
 // possibility of a race between a method call and a concurrent call to `destroy`:
 //
-//    * Thread A starts a method call, reads the value of the handle, but is interrupted
-//      before it can pass the handle over the FFI to Rust.
+//    * Thread A starts a method call, reads the value of the pointer, but is interrupted
+//      before it can pass the pointer over the FFI to Rust.
 //    * Thread B calls `destroy` and frees the underlying Rust struct.
-//    * Thread A resumes, passing the already-read handle value to Rust and triggering
+//    * Thread A resumes, passing the already-read pointer value to Rust and triggering
 //      a use-after-free.
```
rfk, line 66: This change will make the Rust panic from https://github.com/mozilla/uniffi-rs/pull/458 be a bona fide use-after-free memory unsafety issue, so I've based this PR on top of that one. Please review it first :-)

uniffi_bindgen/src/bindings/ruby/templates/ObjectTemplate.rb
```diff
@@ -1,22 +1,46 @@
...
-  {%- match obj.primary_constructor() %}
-  {%- when Some with (cons) %}
-  def initialize({% call rb::arg_list_decl(cons) -%})
-    {%- call rb::coerce_args_extra_indent(cons) %}
-    handle = {% call rb::to_ffi_call(cons) %}
-    ObjectSpace.define_finalizer(self, self.class.define_finalizer_by_handle(handle))
-    @handle = handle
+
+  # A private helper for initializing instances of the class from a raw pointer,
+  # bypassing any initialization logic and ensuring they are GC'd properly.
```
rfk, line 4: @saks I pulled out some "private" helpers here to make codegen a bit easier. Is there a pattern we should use to discourage consumers from calling this, similar to e.g. the leading-underscore pattern from Python?

uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
```diff
@@ -9,30 +9,43 @@ public protocol {{ obj.name() }}Protocol {
...
 
+    // Object equality is pointer identity.
+    // TODO: perhaps we should delegate to Rust for this sort of thing?
+    public static func ==(lhs: {{ obj.name()|class_name_swift }}, rhs: {{ obj.name()|class_name_swift }}) -> Bool {
+        return lhs.pointer == rhs.pointer
+    }
+
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(self.pointer)
+    }
```
rfk, line 44: Here I've implemented `Equatable` and `Hashable` for Object classes, mostly so that they can appear as fields in Records, which are in turn `Equatable` and `Hashable`. An alternative would be to only implement `Equatable` and `Hashable` for Records that do not contain any Object instances.

uniffi_bindgen/src/interface/function.rs
```diff
@@ -61,12 +61,19 @@ impl Function {
...
         &self.name
     }
+
     pub fn arguments(&self) -> Vec<&Argument> {
         self.arguments.iter().collect()
     }
+
+    pub fn all_arguments(&self) -> Vec<Argument> {
+        self.arguments.to_vec()
+    }
```
rfk, line 71: This is a little trick to simplify codegen, more below...

uniffi_bindgen/src/interface/object.rs
```diff
@@ -313,22 +321,29 @@ impl Method {
...
             default: None,
         }
     }
 
+    pub fn all_arguments(&self) -> Vec<Argument> {
+        vec![self.first_argument()]
+            .into_iter()
+            .chain(self.arguments.iter().cloned())
+            .collect()
+    }
```
rfk, line 339: So here for methods, we're making a distinction between the "visible to the user" `arguments()` list and the "what you see in the actual call" `all_arguments()` list...

uniffi_bindgen/src/templates/macros.rs
```diff
@@ -6,14 +6,8 @@
 {{ func.name() }}({% call _arg_list_rs_call(func) -%})
 {%- endmacro -%}
 
-{%- macro to_rs_call_with_prefix(prefix, func) -%}
-    {{ func.name() }}(
-    {{- prefix }}{% if func.arguments().len() > 0 %}, {% call _arg_list_rs_call(func) -%}{% endif -%}
-)
-{%- endmacro -%}
```
rfk, line 13: ...and now we can get rid of this "call with prefix thing" in the Rust codegen macros! Because the `Object` type now lifts and lowers in the same way as the other types, we don't need to do any specially-structured codegen for it, and can just use the plain `to_rs_call` macro as long as we pass it the full list of arguments.

rfk (commented):
So, this is a big change, but I think it has come together quite nicely and is just about ready for review!

@mhammond, thanks for getting this off to such a great start with your earlier PR. Much of what's here should look familiar, although I've removed the identity-map stuff from the Python backend because I wasn't yet prepared to tackle that in all languages, and it seemed worth trying to be consistent.

@jhugman, I'm not sure how closely you've been following this discussion, but would love to get your take on this, especially from the consumer perspective. Some self-review notes below on things to watch out for.

@saks, could you please take a look at what I've done with the Ruby stuff here? This was certainly an interesting way for me to start learning more about Ruby! 😅

CHANGELOG.md
```diff
@@ -19,6 +19,15 @@
   will be removed in a future release. More details on the motivation for this change can be found
   in [ADR-0004](https://github.com/mozilla/uniffi-rs/blob/main/docs/adr/0004-only-threadsafe-interfaces.md).
 
+### What's New
+
+- It is now possible to use Object instances as fields in Records or Enums, to pass them as arguments,
+  and to return them from function and method calls. They should for the most part behave just like
+  a host language object, and their lifecycle is managed transparently using Rust's `Arc<T>` type.
+    - Cross-language reference cycles will not be garbage collected.
```
mhammond, line 27: The "cross language" part here makes it sound like some cycles will be collected - don't have a concrete suggestion here because we don't want to give the impression we will somehow break existing GC, but thought I'd mention it in case you have inspiration.

docs/manual/src/udl/structs.md
```diff
@@ -10,15 +12,52 @@ struct TodoEntry {
...
     boolean done;
     u64 due_date;
     string text;
 };
+```
+
+The fields in a dictionary can be of almost any type, including objects or other dictionaries.
+The current limitations are:
+
+* They cannot recursively contain another intance of the *same* dictionary.
```
mhammond, line 28: typo - intance

examples/todolist/src/lib.rs
mhammond: I wonder if using a `Weak` here might be even more informative? If this was a "real" app, I imagine the default would be a strong ref, but it's an example app. Maybe just a comment along the lines of "depending on your requirements, you could use a `Weak` here and avoid the possible circular reference."?

examples/todolist/tests/bindings/test_todolist.kts
mhammond: It might be worth summarizing this comment in the example - a casual reader might not realize the importance of the pattern used.

examples/todolist/tests/bindings/test_todolist.kts
```diff
@@ -50,6 +50,29 @@ todo.addItems(listOf("bobo", "fofo"))
...
+
 // Ensure the kotlin version of deinit doesn't crash, and is idempotent.
 todo.destroy()
 todo.destroy()
```
mhammond, line 78: and I guess if you are touching this file you might as well add the trailing \n :)

examples/todolist/tests/bindings/test_todolist.py
mhammond: SGTM - maybe even just take that "promise" and commented code out entirely? It sounds like a commitment that I don't think we need to make. Alternatively, water it down "we might..." and link to an issue?

fixtures/coverall/tests/bindings/test_coverall.py
```diff
@@ -89,5 +96,65 @@ def test_self_by_arc(self):
...
+        self.assertEqual(coveralls.strong_count(), 2)
+        self.assertIsNone(coveralls.get_other())
+        coveralls.take_other(coveralls)
+        # should now be a new strong ref.
+        self.assertEqual(coveralls.strong_count(), 3)
+        # but the same number of instances.
+        self.assertEqual(get_num_alive(), 1)
+        # and check it's the correct object.
+        self.assertEqual(coveralls.get_other().get_name(), "test_arcs")
+        # TODO: when we implement an identity map, it will even be the same Python object
```
mhammond, line 111: ditto here re future promises.

fixtures/coverall/tests/bindings/test_coverall.swift
```diff
@@ -0,0 +1,103 @@
...
+        assert(c2.strongCount() == 2)
+
+        coveralls.takeOther(other: c2)
+        // same number alive but `c2` has an additional ref count.
+        assert(getNumAlive() == 2)
+        assert(coveralls.strongCount() == 2)
+        assert(c2.strongCount() == 3)
+
+        assert(coveralls.getOther() == c2)
+        // TODO: in future get_other() should return the exact same instance.
```
mhammond, line 95: ditto re promises

fixtures/coverall/tests/bindings/test_coverall.swift
```diff
@@ -0,0 +1,103 @@
...
+    // TODO: ideally, it would return the same Swift-level object.
+    // assert(coveralls.getOther()!! === coveralls);
+    do {
+        try coveralls.takeOtherFallible()
+        fatalError("Should have thrown")
+    } catch CoverallError.TooManyHoles {
+        // It's okay!
+    }
+    // TODO: kinda hard to test this, as it triggers a fatal error.
+    // coveralls!.takeOtherPanic(message: "expected panic: with an arc!")
```
mhammond, line 68: hrmph - that's surprising to me, although I guess it shouldn't be.

mhammond (approved):
I'm glad it was you and not me who had to work through the Kotlin thorns! It's exciting to see this ready to land 🚀

uniffi_bindgen/src/bindings/ruby/templates/ObjectTemplate.rb
saks: @rfk The best way to hide it would be to mark these methods as `private_class_method`, but this is not quite feasible in our case. Leading-underscore is also pretty common in ruby community, so employing it here is rather a good idea.

fixtures/coverall/tests/bindings/test_coverall.rb
saks: Unfortunately, this is accurate 👍

examples/todolist/tests/bindings/test_todolist.py
rfk: You're right, we should just delete the promise; if it's important to us we'll actually do it, and if it isn't important we don't need to make any promises about it.

fixtures/coverall/tests/bindings/test_coverall.rb
saks: @rfk I've experimented with it and my best guess about what happens is:
1. we define local variables (such as `coveralls`)
2. when calling `assert_raise` we pass it a block that should raise an exception
3. that block creates a closure that captures all visible local variables
4. closure holds local vars and I found no way to GC it as an object

I was able to confirm this by creating a closure before and after defining `coveralls` variable. And AFAIK this has nothing to do with `assert_raise` implementation. I was able to find lots of complaints regarding closures preventing GC from doing it's job.

saks:
I'd like give a shout-out to @rfk for spending time to make these changes work for ruby adapter. Big thank you! 👍

fixtures/coverall/tests/bindings/test_coverall.rb
rfk: Thank you! I've updated the comment in the code to explain this a little more clearly.

uniffi_bindgen/src/bindings/ruby/templates/ObjectTemplate.rb
rfk: Great, I'll by some leading underscores in here to more clearly communicate intent.

rfk:
I've addressed the feedback here, rebased and squashed ready for merge.

uniffi_bindgen/src/templates/macros.rs
jhugman: 😎

jhugman (commented):
I am liking this a lot. I'm not sure I'm qualified to give you a Big Green Button, but I've left some general feedback and some kotlin/swift specific.

I am very excited that we have a clear path to being able to support [static constructor methods][1] and singletons.

[1]: https://github.com/mozilla/uniffi-rs/issues/37
#

docs/manual/src/udl/structs.md
```diff
@@ -2,6 +2,8 @@
 
 Dictionaries can be compared to POJOs in the Java world: just a data structure holding some data.
```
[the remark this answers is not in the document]
rfk: Oh dear 😅 

Honestly, "compared to POJO" and "compared to JavaBeans" both do absolutely nothing to help my personal understanding of what's happening here, which is maybe a signal that we should just reword this to use a general description rather than a language-specific analogy. Thanks for pointing it out.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -2,20 +2,44 @@
...
+//
+// Such types expose a `destroy()` method that must be called to cleanly
+// dispose of the contained objects. Failure to call this method may result
+// in memory leaks.
+//
+// The easiest way to ensure this method is called is to use the `.use`
+// helper method to execute a block and destroy the object at the end.
+interface Destroyable {
+    fun destroy()
+}
```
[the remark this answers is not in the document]
rfk: > we should call this Disposable

This name makes sense to me; should the corresponding method be called `dispose()` for symmetry rather than `destroy()`?

uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
rfk: > Recommend we don't implement Equatable and Hashable until we have a better sense of demand for this.

Aye, thanks. I think I knew that this was most likely the right thing to do but just didn't want to deal with it in this PR. I'll update.

uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
```diff
@@ -42,14 +55,45 @@ public class {{ obj.name()|class_name_swift }}: {{ obj.name() }}Protocol {
...
+    fileprivate static func lift(_ pointer: UnsafeMutableRawPointer) throws -> Self {
+        return Self(unsafeFromRawPointer: pointer)
+    }
+
+    fileprivate func lower() -> UnsafeMutableRawPointer {
+        return self.pointer
+    }
+}
+
+extension {{ obj.name()|class_name_swift }} : ViaFfi, Serializable {}
```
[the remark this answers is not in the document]
rfk: I wasn't able to make it so, for reasons that I don't really understand to do with Swift's object model. Perhaps I'll take a fresh look at it now that I've been away from the code for a little while. At the very least, I can leave a comment here about *why* it can't be private.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -204,10 +204,43 @@ impl<'ci> ComponentInterface {
...
+                    e.variants().iter().any(|v| {
+                        v.fields()
+                            .iter()
+                            .any(|f| self.type_contains_object_references(&f.type_))
+                    })
+                })
+                .unwrap_or(false),
+            _ => false,
+        }
+    }
```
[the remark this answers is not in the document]
rfk: *cough* if only there were some non-enum-based abstraction over all these bits of a `ComponentInterface` that we could use to simplify this...

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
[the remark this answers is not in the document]
rfk: Noting for completeness, that in my local copy I have updated this to implement `AutoCloseable` for objects.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -2,20 +2,44 @@
...
+
+inline fun <T : Destroyable?, R> T.use(block: (T) -> R) =
+    try {
+        block(this)
+    } finally {
+        try {
+            this?.destroy()
```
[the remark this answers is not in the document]
rfk: It's because we implement it on `Destroyable?` in order to be able to use it on `Option<T>` as well as just `T`.

uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
```diff
@@ -9,30 +9,43 @@ public protocol {{ obj.name() }}Protocol {
...
 
     deinit {
         try! rustCall(UniffiInternalError.unknown("deinit")) { err in
-            {{ obj.ffi_object_free().name() }}(handle, err)
+            {{ obj.ffi_object_free().name() }}(pointer, err)
         }
     }
 
+    // Object equality is pointer identity.
+    // TODO: perhaps we should delegate to Rust for this sort of thing?
```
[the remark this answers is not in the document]
rfk: Hrm, I couldn't figure out how to override the default implementation of `===`, so I think I'm going to have to punt on this. (Swift was happy for me to provide my own implementation of the `===` operator but it doesn't seem to be called when I expect it to be).

uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
rfk: Specifically, what I've done is:

* Do not implement `Equatable/Hashable` for `Object` types.
* Update `Record` and `Enum` types so that they only implement `Equatable/Hashable` if they do not contain any `Object` type fields.

uniffi_bindgen/src/interface/object.rs
rfk: In fact, `first_argument` can be removed completely, because we only ever used it in order to build up the list that is now returned by `all_arguments`.

rfk:
I've tweaked the Kotlin and Swift bindings to accommodate James's feedback, and I think this is good to go!

While this is not *technically* a breaking change, it feels significant enough that I'm going to give it its own minor version increment.

PR #463, opened 2021-05-25
Use rust version from rust-toolchain.toml file.

saks:
FYI: I'd prefer to be able to tell `rustup` to look at `rust-toolchain.toml` and install everything, but it looks like `rustup` is [not quite there yet](https://github.com/rust-lang/rustup/issues/2686).


rfk (approved):
Thanks! We're having some broader discussions about how to handle Rust versions in our collection of github projects (e.g. https://github.com/mozilla/application-services/issues/4140) but until we find something clever to do there, it will be valuable to be consistent.

PR #464, opened 2021-05-26
Add codegen backend for Ruby.

rfk:
This is #436, with some adjustments for landing:
* rebased against latest `main`
* squashed to a single commit,
* added changelog entry
* cleanly error out on the new datatypes from #434.

Currently "draft" while I see if I can get `cargo test` passing cleanly on my local machine, and document and non-obvious steps involved in doing so.


rfk:
> Currently "draft" while I see if I can get cargo test passing cleanly on my local machine

Seems like the builtin `ruby` on Mac is too old, but the `ruby` installed from homebrew works fine.

@mhammond I'm r+ on this over in #436, but could you please take a glance at the consolidated commit here for final merge?

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@
 
 ### What's New
 
+- A new **Ruby** codegen backend has been added. You can now call `uniffi-bindgen -l ruby` to
+  generate a Ruby module that wraps a UniFFI Rust component. Thanks to @saks for contributing
+  this backend!
+    - When running `cargo test` locally, you may need to install Ruby and the `ffi` gem
```
mhammond, line 20: some time later we should try and improve this a little - It can be tiresome to filter through test failure output each time and go "yep, that just the lack of swift failing" etc - especially when running from different environments, each of which has a different set of test deps installed.

mhammond (approved):
github has no :stamp emoji, which would be appropriate for me reviewing a ruby patch. WCGW?

CHANGELOG.md
rfk: Yeah, probably it makes more sense to log a message about skipping the test than the fail it; filed https://github.com/mozilla/uniffi-rs/issues/466 for followup.

PR #465, opened 2021-05-26
Fix flaky timestamp test, and add a changelog entry for timestamps.

rfk:


rfk:
@skhamis would you mind taking a look at this one, to distribute the review load away from Mark for a little bit? :-)

fixtures/uniffi-fixture-time/tests/bindings/test_chronological.kts
```diff
@@ -38,6 +38,11 @@ try {
...
 val rustNow = now()
 val kotlinAfter = Instant.now()
-
 assert(kotlinBefore.isBefore(rustNow) || kotlinBefore.equals(rustNow))
-assert(kotlinAfter.isAfter(rustNow) || kotlinAfter.equals(rustNow))
\ No newline at end of file
+assert(
+        kotlinAfter.isAfter(rustNow) ||
+        kotlinAfter.equals(rustNow) ||
+        // Unfortunately the JVM clock was be lower resolution than the Rust clock,
```
skhamis, line 45: nit: Is this meaning to say `... JVM clock  *is*  a lower resolution then the Rust clock`? If so I think swapping the `was be` for `is a` might offer a tad bit more clarity!

For that actual test, I tried to map out all scenarios here and think this covers the ones we "want to care about"

skhamis (approved):
Nice! Excited for Duration to be a type we will be able to use :) LGTM!

fixtures/uniffi-fixture-time/tests/bindings/test_chronological.kts
rfk: Good catch; I think I meant "JVM clock *may* be a lower resolution" (equivocating because the test worked fine in CI but failed consistently on my local machine).

PR #468, opened 2021-05-26
Re-enable publishing in cargo-release configuration.

rfk:
This was accidentally left in as a result of my testing
from https://github.com/mozilla/uniffi-rs/pull/459


rfk:
I've published the `v0.10.0` release by hand, but this will help out release-cutting instructions actually work for future versions.

mhammond (approved):
WCGW?

PR #470, opened 2021-06-02
Update our rust policy to match mozilla-central's 'uses' version

mhammond:
[Rendered version](https://github.com/mhammond/uniffi-rs/blob/update-rust-policy/docs/policies/rust-versions.md)

Ryan and I have come up with this policy for our rust components. We want uniffi and app-services to have the same policy, and because this repo already had a documented I figured I'd start here, then once approved, copy it into the app-services repo.

This updates #432 - @dmose reviewed/approved that, but he's out, so I'm going with just @rfk

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,10 +75,10 @@ jobs:
       - run:
           name: "Print the Rust version, to help with debugging"
           command: rustc --version
-      # some tests fail on MSR:
+      # some tests fail on earlier rust versions and we want to ignore them:
       # * trybuild_ui_tests - error message had a comma inserted! Can probably
       #   be re-enabled after the next rust version bump.
-      - run: cargo test --skip trybuild_ui_tests
+      - run: cargo test -- --skip trybuild_ui_tests
   Deploy website:
     docker:
       - image: circleci/node:latest
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,7 +64,7 @@ jobs:
           name: "Set RUSTFLAGS to fail the build if there are warnings"
           command: echo 'export RUSTFLAGS="-D warnings"' >> $BASH_ENV
       - run: cargo test
-  Rust and Foreign Language tests - MSR:
+  Rust and Foreign Language tests - min. supported rust:
     docker:
       - image: rfkelly/uniffi-ci:latest
     resource_class: large
@@ -75,7 +75,10 @@ jobs:
       - run:
           name: "Print the Rust version, to help with debugging"
           command: rustc --version
-      - run: cargo test
+      # some tests fail on MSR:
+      # * trybuild_ui_tests - error message had a comma inserted! Can probably
+      #   be re-enabled after the next rust version bump.
+      - run: cargo test --skip trybuild_ui_tests
   Deploy website:
     docker:
       - image: circleci/node:latest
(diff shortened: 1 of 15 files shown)


docker/Dockerfile-build
```diff
@@ -36,10 +36,8 @@ RUN sudo apt-get update -qq \
         python3-pip \
     && sudo apt-get clean
 
-RUN rustup update \
-    # This specific version of Rust is required for compatibility with mozilla-central
-    && rustup install 1.47.0 \
-    && rustup default 1.47.0
+# This should automatically install the version specified in rust-toolchain.toml
+RUN rustup update
```
rfk, line 40: Hrm it's not obvious to me that this will execute with the correct context to see the toolchain file by default, I'll try it out and see...

docs/policies/rust-versions.md
```diff
@@ -5,52 +5,53 @@ it be used by external projects. If uniffi always uses Rust features available
 in only the very latest Rust version, this will cause problems for projects
 which aren't always able to be on that latest version.
 
-This is particularly important while uniffi is new and young - it's not always
-possible for our policy to be something like "defer updating uniffi until your
-project is on the same version of rust", because there's a good chance these
-projects *need* the latest version of uniffi for it to be a useful solution.
-
 Given uniffi is currently developed and maintained by Mozilla staff, it should
```
rfk, line 8: Nit: consistent capitalization of "UniFFI" and "Rust" throughout this document

rust-toolchain.toml
```diff
@@ -1,7 +1,7 @@
 # See https://rust-lang.github.io/rustup/overrides.html for details on how this file works
 # and how you can override the choices made herein.
 [toolchain]
-channel = "1.47.0"
+channel = "1.52.1"
```
rfk, line 4: We could consider adding  a note here to reference the circleci config, similar to how the circleci config references this file.

.circleci/config.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     docker:
       # Our policy for updating this rust version is at https://github.com/mozilla/uniffi-rs/tree/main/docs/policies/rust-versions.md
       # See also rust-toolchain.toml in the root of this repo, which is used to ensure the local version of rust matches.
-      - image: circleci/rust:1.47.0
+      - image: circleci/rust:1.52.1
```
rfk, line 34: Actually, I wonder if it would be simpler for us to just use the `uniffi-ci` docker image for running these tests as well, then we don't have to specify an explicit version number here. Not sure how this would affect our CircleCI costs or anything like that, but it'd be simpler.

rfk (commented):
I am 100% on board with this policy change; holding off on r+ purely because I need to get my head around the docker-image build stuff.

rfk (approved):
r+ with the Docker tweak from https://github.com/mhammond/uniffi-rs/pull/4, if they make sense to you. Sadly I think there are also some clippy issues with the new Rust version 😬

.circleci/config.yml
```diff
@@ -13,28 +13,39 @@ commands:
...
+      - run: rustup override set 1.47.0
+      - run: rustup update
+
 orbs:
   gh-pages: sugarshin/gh-pages@0.0.6
 jobs:
   Check Rust formatting:
     docker:
       # Our policy for updating this rust version is at https://github.com/mozilla/uniffi-rs/tree/main/docs/policies/rust-versions.md
       # See also rust-toolchain.toml in the root of this repo, which is used to ensure the local version of rust matches.
```
rfk, line 33: This comment no longer refers to anything, as there is no "this rust version" in the lines that follow it.

rfk (approved):
Looks great! I like the way you've factored the Rust versions in CI here, much cleaner than how we currently do the beta tests in appservices.

Issue #471, opened 2021-06-02
Proposal: Adding Mutability to Swift structs


skhamis:
## Problem

Uniffi currently generates structs in Swift with the keyword `let` which evaluates as a constant and `var` which is the mutable counterpart. [Swift documentation on var vs let](https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html#ID310)

This is currently happening in the current uniffi work https://github.com/mozilla-mobile/firefox-ios/pull/8602 where I need to construct an entirely new struct to change a couple of variables inside. While not terrible, it does leave potential room for improvement. The actual "fix": https://github.com/mozilla-mobile/firefox-ios/pull/8602/commits/54cee20b10a249f37167b8e2fba6c4a69ee68317

## Solution

Since this is swift specific it might be harder to propose any solution in the UDL without mucking things up. However there seems to be two paths:

1. In the UDL have a decorator that can specify potential language-specific needs while keeping the verbose-ness of the current UDL style:
    - something like `[Mutable]` since almost all supported languages have equivalent of mutable/immutable

2. After briefly discussing this issue with @rfk, another (probably much better) option is to default swift structs to be `var` instead of `let` to allow mutability by default. However that comes with it's own potential paths to think through.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2234)

rfk:
FWIW the Kotlin bindings have a similar property, using `val` instead of `var` and thus creating immutable properties.

I think I'm lightly +1 on allowing these properties to be mutable by default. One risk could be if consumers expect the mutations to somehow be automatically reflected in an underlying object in the Rust code, but given the current shape of our APIs that doesn't seem too likely.

rfk:
If you wanted to play around with this in a local build, I *think* it could be as small as changing `let` to `var` in the Swift Record template [here](https://github.com/mozilla/uniffi-rs/blob/66de08450b9df8447278c7106dfa99f863f9c2ac/uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift#L3) and `val` to `var` in the Kotlin Record template [here](https://github.com/mozilla/uniffi-rs/blob/66de08450b9df8447278c7106dfa99f863f9c2ac/uniffi_bindgen/src/bindings/kotlin/templates/RecordTemplate.kt#L3).

skhamis:
[...]

PR #474, opened 2021-06-03
Change swift and kotlin bindings to be mutable by default

skhamis:
This is the result from the discussion in Issue https://github.com/mozilla/uniffi-rs/issues/471


rfk (approved):
Thanks! Could you please also add a brief changelog entry for completeness?

PR #479, opened 2021-06-12
Remove lots of boilerplate from bindings by using askama's 'else'

mhammond:
There's a fair bit of copy/pasted "this type not supported" boilerplate in the bindings that would be simplified using `{% else %}` - when I found myself adding copy/pasting the same boilerplate in these locations I figured I'd see if it could be avoided.


mhammond:
Turns out the ruby and python bindings both had errors in that copy/pasta too - the `write` functions didn't have a param for the value.

rfk (approved):
Thanks! When I initial structured it this way, I think I was trying to take advantage of exhaustive matching to make sure we didn't accidentally forget to handle a particular type in a backend. But that doesn't really work with the current approach of just stubbing out unimplemented types, so using `else` makes more sense 👍🏻

PR #480, opened 2021-06-14
Publish auto-generated Rust API docs as part of the UniFFI manual.

rfk:
Rust has great tooling for API docs, and I think we should try to
get into the habit of using it. This commit generates API docs for
our main crates and includes them in the published UniFFI manual,
which will hopefully help folks onboarding into the codebase.


mhammond (approved):
These docs are really coming together

Issue #475, opened 2021-06-03
Support "local" types which know how to convert to and from primitive types - eg, Guid, Url


mhammond:
In app-services, we've had to compromise on the interfaces we expose, particularly around `Guid` and `Url`. Here's my latest take on this:

How about we "just" provide a way for consumers to provide a list of "external types" - a list of `[(name, raw_type)]` - eg, `('Guid', String)`

Doing this allows us to write in the .udl:

```
  Guid get_guid(optional Guid? s);
```

and then it's fairly easy to have:

1) the Rust side of the generation just calls `fn get_guid(guid: Option<Guid>) -> Guid` (assuming that type exists and implements the uniffi traits)
2) the "other" side of the generation world to just use strings - in Python, `get_guid(None))` works and returns a string.

and you can imagine (2) being optionally smarter - eg, we can probably arrange for projects to also arrange for a suitable `Guid` type to be used in the bindings.

I haven't thought too much about how to spell these types, but for the sake of example, let's assume in `uniffi.toml` with something like:

```
[local_types]

[local_types.Guid]
raw = "String"
# Python gets special support, other languages get a string.
python = { lower = "some_package.GuidType.to_string", lift = "some_package.GuidType.from_string" }

[local_types.JSONObject]
raw = "String"
python = {...}
kotlin = {...}
# etc!
```

This would probably allow a consumer to fully implement #440 without any help from us and probably even backout timestamp/duration support.

(To be clear, I'm not saying we *should* exclude JSON support or back timestamps out, but we *could*)

FWIW, I've got my `get_guid()` example above working in a very very hacky way, including painfully hand-written implemention of `ViaFfi` for `Guid` - but close enough to convince me it's doable.

WDYT?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-65)
┆Issue Number: UNIFFI-65

rfk:
One thing to be aware of here is that this can interfere with the use of generics in function signatures, like [this example](https://github.com/mozilla/uniffi-rs/blob/d58e92f58a48e65b0c6ec10c0cb22d103a266e50/examples/todolist/src/lib.rs#L53) from the TodoList crate:

```
fn add_item<S: Into<String>>(&self, item: S) -> Result<()> {
     ...
} 
```

[...]

Issue #478, opened 2021-06-09
Improve situation around sharing types between crates.


mhammond:
It's currently impossible to expose a struct etc defined in another crate - uniffi tries to define `uniffi::ViaFfi` for the type and the rust compiler doesn't let that happen (`impl doesn't use only types from inside the current crate`).

Even if we came up with a way to get around that, we'd need to take some care as the next obvious step would be that multiple crates all expose the same type and would expect to be able to pass these types between each other - which is another can of worms if each crate generates its own version of the type.

For a concrete example, grep for the `DeviceType` enum in application-services - the "real" type should be owned by sync15-traits, but fxa-client, sync-manager and tabs have clones of the enum, and at least the fxa and sync15 versions already diverge in how they serialize to json.

This might be tricky, so the least we can do is document this.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-66)
┆Issue Number: UNIFFI-66

PhilippGackstatter:
I just ran into this issue while trying to setup uniffi for a multi-crate project. Is there a typical way to work around this? It's not possible to write a `udl` file for each crate and combine the results somehow, right? Thanks.

rfk:
> Is there a typical way to work around this?

We don't have a good way to deal with it yet, but we've started running into this issue in our production code so we're actively working on a solution. Are you able to describe your multi-crate project setup a little bit, so we can use it as another reference point when discussing solutions?

PhilippGackstatter:
Sure. Although, I just got started with uniffi and considered it to implement our bindings. Among others, there are the `identity-core` crate, and the `identity-iota` crate. There is a `identity_core::crypto::KeyPair` struct and that is used in one of the methods of `identity_iota::IotaDocument`. We also have an `identity` crate that re-exports all these types for easier usage.

Writing two separate `udl` files for `identity_core` and `identity_iota` doesn't allow the scaffolding to succeed, as `identity_iota`'s `udl` file doesn't contain the `KeyPair`, because it is not defined in that crate. Importing the `KeyPair` also fails, since we cannot implement `ViaFfi` for such a type, as mentioned above. I could see one solution involving an import mechanism in `udl` files, to combine multiple crates into one "ffi library", though I'm sure that comes with is own caveats.

[...]

PR #482, opened 2021-06-14
Experiment with what "external types" might look like.

mhammond:
An experiment for what I described in #475.

The most interesting part is [a new fixture](https://github.com/mhammond/uniffi-rs/tree/typedefs/fixtures/external-types). It implements 2 types - a `Guid` and a `JSONObject`, with the idea that all languages except rust could use Guid as a String, but `JSONObject` would have all languages have a custom type. This support for foreign types is defined in [uniffi.toml](https://github.com/mhammond/uniffi-rs/blob/typedefs/fixtures/external-types/uniffi.toml) and only exists for Python, but I think should work fine in the others.

The JSONObject is a bit clunky - it would be perfect if rust could use serde::Value directly, but the "no traits on external types" means we need to wrap it. If we could use serde directly, the ergonomics for both sides of the FFI here seem almost identical to what we'd end up with otherwise.

Error handling will need some thought - if we are wrapping a function that doesn't Err<>, then do we allow use of custom types which might fail, or do we just unwrap for them? Everything currently unwraps, which is bad.

Quite a few other details you'll find too :)

### DIFF
--- a/fixtures/external-types/src/external-types.udl
+++ b/fixtures/external-types/src/external-types.udl
@@ -1,0 +1,23 @@
+
+[External]
+typedef string Guid;
+
+[External]
+typedef string JSONObject;
+
+dictionary ExtTypes {
+    Guid guid;
+    sequence<Guid> guids;
+    JSONObject json;
+    sequence<JSONObject> jsons;
+    // should get all other recursive types.
+};
+
+namespace external_types {
+    string get_string(optional string? val);
+    Guid get_guid(optional Guid? val);
+
+    JSONObject? get_json_object(optional JSONObject? val);
+
+    ExtTypes get_ext_types(optional ExtTypes? vals);
+};
--- a/fixtures/external-types/src/lib.rs
+++ b/fixtures/external-types/src/lib.rs
@@ -1,0 +1,99 @@
+// A trivial Guid implementation.
+struct Guid(String);
+
+// A simple `type JSONObject = serde_json::Value;` would work, except that
+// we can't impl ViaFfi `impl doesn't use only types from inside the current crate`
+struct JSONObject(serde_json::Value);
+
+fn get_guid(guid: Option<Guid>) -> Guid {
+    match guid {
+        Some(guid) => guid,
+        None => Guid("NewGuid".to_string()),
+    }
+}
+
+fn get_string(s: Option<String>) -> String {
+    match s {
+        Some(s) => s,
+        None => "NewString".to_string(),
+    }
+}
+
+fn get_json_object(v: Option<JSONObject>) -> Option<JSONObject> {
+    match v {
+        Some(v) => Some(v),
+        None => Some(JSONObject(serde_json::json!({"foo": "bar"}))),
+    }
+}
+
+struct ExtTypes {
+    guid: Guid,
+    guids: Vec<Guid>,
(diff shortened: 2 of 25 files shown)


uniffi_bindgen/src/bindings/python/gen_python.rs
```diff
@@ -206,6 +230,21 @@ mod filters {
             Type::Boolean => format!("(True if {} else False)", nm),
             Type::String => format!("{}.consumeIntoString()", nm),
             Type::Object(name) => format!("{}._make_instance_({})", class_name_py(name)?, nm),
+            Type::ExternalType(ext_name, primitive) => {
+                let prim_lift = format!(
```
mhammond, line 234: an alternative strategy here might be to generate custom global helpers for each type and this can shrink to identical boilerplate regardless of implementation. That's just shifting the code to a more dedicated "setup" step, but that might be worthwhile.

fixtures/external-types/src/external-types.udl
```diff
@@ -0,0 +1,23 @@
+
+[External]
+typedef string Guid;
+
+[External]
+typedef string JSONObject;
```
jhugman, line 6: Oh, I like that. I was wondering how to get these in to the frontend.

fixtures/external-types/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+                Ok($from_string(s)) // XXX - wire up errors
+            }
+        }
+    }
+}
+
+viaffi_simple_string!(JSONObject, self, self.0.to_string(), |s: String| Self(
+    serde_json::from_str(&s).unwrap()
+));
+viaffi_simple_string!(Guid, self, self.0, |s| Self(s));
```
jhugman, line 97: Does this mean that each crate that needs a JSON needs to implement this? Or can these be put into a uniffi-runtime addendum crate(s)?

fixtures/external-types/uniffi.toml
```diff
@@ -0,0 +1,7 @@
+[bindings.python]
+
+[bindings.python.external_types]
+[bindings.python.external_types.JSONObject]
+setup = "import json"
+lift_from_primitive = "json.loads"
+lower_to_primitive = "json.dumps"
```
jhugman, line 7: This looks interesting: I wonder how multi-lines and error handling fit in.

jhugman (commented):
This is interesting work. I love the idea of moving the type transforms into their own crates.

I'm not sure about the client code inheriting uniffi-runtime implmentation details (the JSONObject wrapper struct).

Could you/would see this generalize further? `ffi_via_i32!` for example? Could we get rid of some of these match expressions and replace them with lookups?

fixtures/external-types/src/lib.rs
mhammond: Thanks James. That's a great question! I guess I'm hoping for a middle ground, so that both of these can be true:

* uniffi itself has no opinion about what a `JSONObject` is, and no opinion on how it should be exposed to each language.
* projects like app-services don't need to specify this for every crate - and better, should be able to pass `JSONObject`s around freely.

How to spell that is unclear. Some kind of `include` mechanism? Something else? This intersects very closely with the problem described in #478.

PR #483, opened 2021-06-15
Use rusty docker image for building API docs in CI.

rfk:
The docs-publishing CircleCI task is using a node-based docker image,
which means it doesn't have the necessary tooling to build the API
docs. Let's use the same docker image as the other tasks for simplicity.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,9 @@ commands:
     steps:
       - run:
           name: Build API Docs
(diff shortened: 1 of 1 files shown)


rfk:
I temporarily configured Circle to deploy the website from this PR to confirm that it works, and it seems to:

* https://mozilla.github.io/uniffi-rs/internals/crates.html

mhammond (approved):

PR #484, opened 2021-06-15
Fix U16 for python and ruby

saks:


rfk (approved):
Nice catch, thank you! Could you please also add a brief note in the changelog?

PR #486, opened 2021-06-16
Fixes mismatch in kotlin panic exceptions

### LINKED ISSUE #485: Error returning functions will throw a different exception when panic'ing than non-error returning functions in kotlin
So looking back at some of the binding generation code, and trying to reason about the error handling - got a personal investment in that code 😄 

[It seems that non-throwing functions are set to return an `InternalException`](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt#L116) which is great to catch panics! Because a panic will be mapped to an Error on the rust side. [Eventually, this InternalExpection error is thrown up](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt#L94)

The issue comes in because throwing functions (functions that already are expecting to be able to throw errors) will not throw the same error if they see a panic - [they instead would throw a `RuntimeException`](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt#L116).

I think this is okay consumer-wise, a panic being an `InternalException` or a `RuntimeException` is OK (I think? they're not meant to be caught anyway) but it can cause problems in a development experience for developers consuming our APIs.

### How to fix it
A very simple fix is probably to simply throw an `InternalException` instead of a `RuntimeException` [when a code is not recognized](ttps://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt#L116)

This could be a nice first issue for me to take on coming back, so I'll pick it up!
I'll take a look at the other bindings after, and file other issues if I see something similar

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2256)

tarikeshaq:
fixes #485 

As a follow-up I'll take a look at the other bindings and make sure this behaviour is consistent. I'll file an issue to add a coverall.falliblePanic test for the other bindings (which will include figuring out if this behaviour is also wrong there), ref for follow up: #487

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
@@ -113,7 +113,8 @@ internal open class {{e.name()}} : RustError() {
             {% for value in e.values() -%}
             {{loop.index}} -> return {{e.name()}}Exception.{{value}}(message) as E
             {% endfor -%}
-            else -> return InternalException(message) as E
+            -1 -> return InternalException(message) as E
+            else -> throw RuntimeException("invalid error code passed across the FFI")
         }
     }
 }
--- a/uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
@@ -113,7 +113,7 @@ internal open class {{e.name()}} : RustError() {
             {% for value in e.values() -%}
             {{loop.index}} -> return {{e.name()}}Exception.{{value}}(message) as E
             {% endfor -%}
-            else -> throw RuntimeException("Invalid error received: $code, $message")
+            else -> return InternalException(message) as E
         }
     }
 }
--- a/uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
@@ -83,12 +83,16 @@ public enum {{e.name()}}: RustError {
         switch rustError.code {
             case 0:
                 return nil
+            case -1:
+            // TODO: Return a Uniffi defined error here
+            // to indicate a panic
+                fatalError("Panic detected!")
             {% for value in e.values() %}
             case {{loop.index}}:
                 return .{{value}}(message: String(cString: message!))
             {% endfor %}
             default:
-                return nil
+                fatalError("Invalid error")
         }
     }
 }
--- a/fixtures/coverall/src/coverall.udl
+++ b/fixtures/coverall/src/coverall.udl
@@ -56,6 +56,9 @@ interface Coveralls {
 
     void panic(string message);
 
+    [Throws=CoverallError]
+    void fallible_panic(string message);
+
     // *** Test functions which take either `self` or other params as `Arc<Self>` ***
 
     /// Calls `Arc::strong_count()` on the `Arc` containing `self`.
--- a/fixtures/coverall/src/lib.rs
+++ b/fixtures/coverall/src/lib.rs
@@ -125,6 +125,10 @@ impl Coveralls {
         }
     }
 
+    fn fallible_panic(&self, message: String) -> Result<()> {
+        panic!("{}", message);
+    }
+
     fn get_name(&self) -> String {
         self.name.clone()
     }
--- a/fixtures/coverall/tests/bindings/test_coverall.kts
+++ b/fixtures/coverall/tests/bindings/test_coverall.kts
@@ -68,6 +68,13 @@ Coveralls("test_arcs").use { coveralls ->
     } catch (e: InternalException) {
         // No problemo!
     }
+
+    try {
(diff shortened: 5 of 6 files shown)


uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -113,7 +113,7 @@ internal open class {{e.name()}} : RustError() {
             {% for value in e.values() -%}
             {{loop.index}} -> return {{e.name()}}Exception.{{value}}(message) as E
             {% endfor -%}
-            else -> throw RuntimeException("Invalid error received: $code, $message")
+            else -> return InternalException(message) as E
```
rfk, line 116: So, this is an underspecified part of UniFFI. Did we actually catch a panic (which has [a specific error code](https://docs.rs/ffi-support/0.4.3/ffi_support/struct.ErrorCode.html#associatedconstant.PANIC)) or did we read garbage data and try to interpret it as an `ExternError`? And how much does the distinction matter?

Since the Kotlin bindings have other places where they check internal invariants and raise `RuntimeException` if they're violated, what I think I'd like to see here is:

* If `code` is `-1` (the special value for a caught Rust panic) then throw `InternalException`
* If it is any other unexpected value, continue throwing `RuntimeException` as before

rfk (changes requested):
Thanks for digging back in here! Since we're thinking about how to make the behaviour here predictable, a thought below on the overall mental model of the behaviour of the bindings:

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
tarikeshaq: Hmm good point.. and from there, I can probably create a follow-up to rename the `InternalException` to a `PanicException` or something similar across the bindings.

tarikeshaq:
@rfk I also added a change to the swift side of things. I noticed that it was a more serious bug so I opted to add it here - fallible functions didn't return **any** error when they panic'ed...

uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
```diff
@@ -83,12 +83,16 @@ public enum {{e.name()}}: RustError {
         switch rustError.code {
             case 0:
                 return nil
+            case -1:
+            // TODO: Return a Uniffi defined error here
+            // to indicate a panic
+                fatalError("Panic detected!")
```
rfk, line 89: Aye, this is a broader thing we have to think about in the Swift bindings - should it be possible to catch and handle Rust panics? And does it make sense that you can do this only for functions that otherwise return an error? I'm not familiar enough with consuming the swift bindings to know what a reasonable set of expectations are.

uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
```diff
@@ -83,12 +83,16 @@ public enum {{e.name()}}: RustError {
...
             {% for value in e.values() %}
             case {{loop.index}}:
                 return .{{value}}(message: String(cString: message!))
             {% endfor %}
             default:
-                return nil
+                fatalError("Invalid error")
```
rfk, line 95: Oof, nice catch, thanks!

rfk (approved):

rfk:
@tarikeshaq please feel free to test out your write access to this repo by merging this PR :-)

uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
tarikeshaq: Hmm it seems functions that don't return errors capture the errors, but instead of returning them to the consumer, [they cause a fatal error because of the `try!` here](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/swift/templates/macros.swift#L100) So that adds a bit of context, but still isn't exposed to the consumer... So maybe the same approach? But ya, other than messing around with it in this project, my Swift knowledge is nil 🙃

rfk: Right, and this is kind of unfortunately necessary because of Swift's explicit checked-exceptions kind of model. We either force *every* call to a Swift API function to have to explicitly handle the possibility of an error, or we allow functions that "should" never fail to be called without a `try` handler and are forced to fatally error out if they do fail in practice.

(It's similar in a lot of ways to Rust's `Result` type; if we were generating Rust bindings, would we want to force every single exposed function and method to return a `Result` just to be able to handle the case of a panic?)

tarikeshaq: Yeah I'm a fan of the "let them fatally error" approach, seems the most "true" to Rust's approach. Because those types of errors should only happen if we observed a panic (which would have fatally errored anyways) or something went awfully wrong. I'll add this bit of context to #487

PR #488, opened 2021-06-17
Updating the examples documentiation

bendk:
Couple documentation updates based on my trying to get started with uniffi coding.

- Updated ruby dependencies
- Updated cargo commands
- Link to the examples dir from the manual.  These were really helpful
  for me and I think it would be great to link for others.  Is there a
  better way that a direct github link?


rfk (approved):
Great, thank you!

> Is there a better way that a direct github link?

Hrm...I can't think of a better way than this for now. In the future we may decide to actually publish some of these examples on crates.io, and if we did it would make sense to link to them there. But even highlighting the github repo is a substantial improvement 👍🏻

Issue #442, opened 2021-05-14
Kotlin errors being `FooErrorException` is awkward. 


mhammond:
eg, [`ArithmeticErrorException` here](https://github.com/mozilla/uniffi-rs/blob/9aee2083ae51fd8e3e87682a2c10370aa2213617/examples/arithmetic/tests/bindings/test_arithmetic.kts#L9) - all other bindings have the more natural `ArithmeticError` as defined in the udl.

It's grating on me in https://github.com/mozilla/application-services/pull/4068 so thought I'd ask if this really is idiomatic?

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-2208)

rfk:
I agree this is pretty yuck :-/

Naively, it feels like the right thing to do would be to have the exception class named `ArithmeticError` just like in other bindings. If calling it "FooException" is really important for idiomatic Kotlin reasons, we could also consider translating error classes named `FooError` into `FooException` automatically in the bindings.

@grigoryk maybe you can give us a bit of a gut-check here from the PoV of a consumer? Looking at the FxA error class wrappers [here](https://github.com/mozilla-mobile/android-components/blob/bfaf514793dd20f9f12d3f551f9ff3e470f52727/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Exceptions.kt#L10), would it be appropriate for us to rename `fxaclient.FxaErrorException` to just `fxaclient.FxaError`?

npars:
In java a "FooError" would typically be expected to extend [Error](https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html) and would be considered fatal. Meanwhile "FooException" would be expected to extend [Exception](https://docs.oracle.com/javase/7/docs/api/java/lang/Exception.html) and be non-fatal / catchable. I assume that was the original intention behind tacking on "Exception" at the end.

jhugman:
@npars is correct: from a JVM point of view, Error means something very specific. Removing an `Error` suffix before appending `Exception` may be the write thing to do here, even if low priority.

Stymergic hint: the quickest dirtiest path I can think of would be doing this in `class_name_kt` in `gen_kotlin.rs`.

grigoryk:
ditto JVM Error meaning.

I agree with @jhugman's suggestion RE dropping `Error` before the name conversion, so we end up w/ `ArithmeticException`, not `ArithmeticErrorException`.

rfk:
[...]

PR #489, opened 2021-06-17
Rename kotlin errors (#442)

bendk:


examples/arithmetic/tests/bindings/test_arithmetic.kts
```diff
@@ -6,7 +6,7 @@ assert(add(4u, 8u) == 12uL)
 try {
     sub(0u, 2u)
     throw RuntimeException("Should have thrown a IntegerOverflow exception!")
-} catch (e: ArithmeticErrorException) {
+} catch (e: ArithmeticException) {
```
rfk, line 9: OK yes, I'm one line in to the PR and already this seems much nicer 👍🏻

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -280,4 +280,20 @@ mod filters {
             _ => format!("{}.read({})", type_kt(type_)?, nm),
         })
     }
+
+    /// Generate a Kotlin exception name
+    ///
+    /// This replaces "Error" at the end of the name with "Exception".  Rust code typically uses
+    /// "Error" for any type of error but in the Java word, "Error" means a non-recoverable error
```
rfk, line 287: sp: "word" => "world"

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -280,4 +280,20 @@ mod filters {
...
+    /// and is distinguished from an "Exception".
+    pub fn exception_name_kt(error: &Error) -> Result<String, askama::Error> {
+        match error.name().strip_suffix("Error") {
+            None => Ok(error.name().to_owned()),
+            Some(trimmed) => {
+                let mut rv = trimmed.to_owned();
+                rv.push_str("Exception");
+                Ok(rv)
```
rfk, line 295: Purely as a style-consistency nit, I'm not 100% sure how to parse "rv" here - is it "return value"? I think I'd lightly prefer a more specific name like `err_name` or similar, but I don't feel strongly about it (and I'm sure there are plenty of things in this codebase that are not consistent with variable names 😅)

rfk (approved):
Great, thanks for diving in! I left a couple of style nits, feel free to take 'em or leave 'em.

Please also add a pending entry at the top of the changelog, calling out in particular that this is a breaking change for consumers.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
bendk: Yeah, "rv" was supposed to mean return value.  I just pushed a new commit that has something more descriptive.

PR #491, opened 2021-06-17
Fixed timestamp conversion during DST

bendk:
I'm not totally sure about this one, but I believe it's an improvement.  Here's snippet showing how the two pieces of code are different (NOTE, the same snippet may give you a different result if you're not in the Eastern timezone):

```
>>> print(datetime.datetime.fromtimestamp(1623965559))
2021-06-17 17:32:39
>>> print(datetime.datetime.fromtimestamp(0) + datetime.timedelta(seconds=1623965559))
2021-06-17 16:32:39
```
 
OTOH, I don't like the way we have to catch all ValueErrors and convert them to OverflowErrors.  [The python docs say that this was changed in version 3.3](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp).  They also say that there's some errors that get classified as OSErrors.  With this code, they will get misclassified as OverflowErrors.  But it seems to me the DST issue is more likely to cause problems than failures in `localtime()`

BTW, what version of python are we targetting with uniffi?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 ### What's Changed
 
 - Both python and ruby backends now handle U16 correctly.
+- Python timestamps will now be in UTC and timezone-aware rather than naive.
 
 ## v0.12.0 (2021-06-14)
 
--- a/uniffi_bindgen/src/bindings/python/templates/RustBufferStream.py
+++ b/uniffi_bindgen/src/bindings/python/templates/RustBufferStream.py
@@ -113,10 +113,13 @@ class RustBufferStream(object):
     def read{{ canonical_type_name }}(self):
         seconds = self._unpack_from(8, ">q")
         microseconds = self._unpack_from(4, ">I") / 1000
+        # Use fromtimestamp(0) then add the seconds using a timedelta.  This
+        # ensures that we get OverflowError rather than ValueError when
+        # seconds is too large.
         if seconds >= 0:
-            return datetime.datetime.fromtimestamp(0, tz=None) + datetime.timedelta(seconds=seconds, microseconds=microseconds)
+            return datetime.datetime.fromtimestamp(0, tz=datetime.timezone.utc) + datetime.timedelta(seconds=seconds, microseconds=microseconds)
         else:
-            return datetime.datetime.fromtimestamp(0, tz=None) - datetime.timedelta(seconds=-seconds, microseconds=microseconds)
+            return datetime.datetime.fromtimestamp(0, tz=datetime.timezone.utc) - datetime.timedelta(seconds=-seconds, microseconds=microseconds)
 
     {% when Type::Duration -%}
     # The Duration type.
--- a/fixtures/uniffi-fixture-time/tests/bindings/test_chronological.py
+++ b/fixtures/uniffi-fixture-time/tests/bindings/test_chronological.py
@@ -3,39 +3,40 @@ from datetime import datetime, timedelta, timezone, MAXYEAR
 import sys
 
 # Test passing timestamp and duration while returning timestamp
-assert add(datetime.fromtimestamp(100.000001), timedelta(seconds=1, microseconds=1)).timestamp() == 101.000002
+assert add(datetime.fromtimestamp(100.000001, timezone.utc), timedelta(seconds=1, microseconds=1)).timestamp() == 101.000002
 
 # Test passing timestamp while returning duration
-assert diff(datetime.fromtimestamp(101.000002), datetime.fromtimestamp(100.000001)) == timedelta(seconds=1, microseconds=1)
+assert diff(datetime.fromtimestamp(101.000002, timezone.utc), datetime.fromtimestamp(100.000001, timezone.utc)) == timedelta(seconds=1, microseconds=1)
 
 # Test pre-epoch timestamps
-assert add(datetime.fromisoformat('1955-11-05T00:06:00.283001'), timedelta(seconds=1, microseconds=1)) == datetime.fromisoformat('1955-11-05T00:06:01.283002')
+assert add(datetime.fromisoformat('1955-11-05T00:06:00.283001+00:00'), timedelta(seconds=1, microseconds=1)) == datetime.fromisoformat('1955-11-05T00:06:01.283002+00:00')
 
 # Test exceptions are propagated
 try:
-    diff(datetime.fromtimestamp(100), datetime.fromtimestamp(101))
+    diff(datetime.fromtimestamp(100, timezone.utc), datetime.fromtimestamp(101, timezone.utc))
     assert(not("Should have thrown a TimeDiffError exception!"))
 except ChronologicalError.TimeDiffError:
     # It's okay!
     pass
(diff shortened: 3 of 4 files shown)


rfk:
> BTW, what version of python are we targetting with uniffi?

 I don't think we have declared an official minimum version.

uniffi_bindgen/src/bindings/python/templates/RustBufferStream.py
```diff
@@ -113,10 +113,13 @@ def readString(self):
     def read{{ canonical_type_name }}(self):
         seconds = self._unpack_from(8, ">q")
         microseconds = self._unpack_from(4, ">I") / 1000
-        if seconds >= 0:
-            return datetime.datetime.fromtimestamp(0, tz=None) + datetime.timedelta(seconds=seconds, microseconds=microseconds)
-        else:
-            return datetime.datetime.fromtimestamp(0, tz=None) - datetime.timedelta(seconds=-seconds, microseconds=microseconds)
+        try:
+            if seconds >= 0:
+                return datetime.datetime.fromtimestamp(seconds, tz=None) + datetime.timedelta(microseconds=microseconds)
```
rfk, line 118: Hrm, interesting. According to [the docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp), specifying `tz=None` will make a "naive" datetime object that doesn't know what timezone it is in, which doesn't feel right to me. We may be better off specifying `tz=datetime.timezone.utc` here explicitly:

```
return datetime.datetime.fromtimestamp(0, tz=datetime.timezone.utc) + datetime.timedelta(seconds=seconds, microseconds=microseconds)
```

@bendk does that make a difference in consistency for DST time for you, and would it avoid the need for `ValueError` catching?

bendk: On the practical question Using UTC makes the time `21:32`, we'd need to then convert it to the user's local timezone, but I don't think we can do that -- it's especially hard without the `pytz` package

On the meta point, I think having a naive datetime could be correct.  The underlying data is just an integer, so the timezone should be ambiguous.

rfk: > The underlying data is just an integer, so the timezone should be ambiguous.

My mental model of the underlying data is that it is unambiguously a duration relative to 1970-01-01 00:00:00 UTC; in the language of the python `datetime` docs it "represents a specific moment in time that is not open to interpretation."

Then again, I do not claim to be an expert on these things, and TBH I don't really understand why the two snippets in your original example produce a different result 😬 

/cc @npars who worked on the initial implementation here.

bendk: Time is so weird.  You're right that the timestamp is unambiguous.  What's tricky is to convert a UTC time to a local time.  It seems that `fromtimestamp()` is smart and leverages the locale information to figure out if you're in DST or not.   timedelta addition can't really do that.  So with the current information, it always assumes that you're not in DST because the unix epoch wasn't (or maybe it's the other way).

mhammond: > What's tricky is to convert a UTC time to a local time.

Do you mean here that you simply want to return the exact same time but in the user's local timezone? If so, I don't see why that's necessary. However, IIUC, SystemTime itself doesn't carry timezone information, so a naive datetime object seems right as sad as that sounds. `chrono::DateTime`, for example, would probably have allowed a timezone

npars: @mhammond is correct for my original justification, SystemTime is not based in a specific timezone, so it should be naive.

npars: I would be interested in seeing a test for this. I wonder if we could verify that printing an iso8601 version of the timestamp produces the same result in both Python and Rust

bendk: I think there's 2 directions to go with this one:

1) If we want to use a naive datetime that has the system time, I think I could improve the code by trying to do the datetime addition version in the `except:` handler.  If the addition produces an OverflowError, then let that bubble up, otherwise raise an OS error.  I think we should also add an hour to the time, to compensate for potential DST changes.

2) I think that rfk convinced me on this one that SystimeTime is in UTC.  I'm pretty sure that if I simply added the seconds to 1970, I would get end at `21:32` instead of `17:32`.  So it seems like there's an attempt to convert it into my timezone.  https://doc.rust-lang.org/std/time/struct.SystemTime.html#impl also lists UTC in the `UNIX_EPOC` constant.  If that's true, maybe it makes sense to return a tz-aware datetime in UTC time -- the one TZ that we can handle without pytz.  That could be an issue for existing code, but it seems the most correct to me.  It also might help out future consumer code, it can be really painful trying to deal with datetimes stored in a database without a timezone.

rfk: >  maybe it makes sense to return a tz-aware datetime in UTC time

I support this FWIW; it may not be any easier to explain this choice of default to consumers, but it seems like the one that's least likely to be a silent footgun for them in practice. (It might be a *noisy* footgun if they then try to combine it with naive datetime objects, but that's much better IMHO).

bendk: Just pushed a changeset that makes the datetimes in UTC and tz-aware.

This requires python 3.2.  If we want to support earlier versions it should be simple to implement it ourselves.

rfk: > This requires python 3.2. If we want to support earlier versions (including 2.x) 

That sounds reasonable to me; in particular I think supporting python2 should be an explicit non-goal, and IIRC the first few releases in the 3.x series don't have a significant install base (but that's just a vague impression, I'm not an expert on the matter).

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 ### What's Changed
 
 - Both python and ruby backends now handle U16 correctly.
+- Python timestamps will now be in UTC and timezone-aware instead of than naive.
```
rfk, line 18: sp: "instead of than" -> "rather than" or similar

uniffi_bindgen/src/bindings/python/templates/RustBufferStream.py
```diff
@@ -113,10 +113,13 @@ def readString(self):
     def read{{ canonical_type_name }}(self):
         seconds = self._unpack_from(8, ">q")
         microseconds = self._unpack_from(4, ">I") / 1000
+        # Use fromtimestamp(0) then add the seconds using a timedelta.  This
+        # ensures that we get OverflowError rather than ValueError when
+        # seconds is too large.
```
rfk, line 118: nice, thanks for adding this comment

rfk (approved):
Looks 👍🏻 to me, thanks for taking to time to talk this through everyone!

PR #492, opened 2021-06-21
Adds ktlint to kotlin requirements in contributing.md

tarikeshaq:
~~I remember having to install ktlint to have the kotlin tests pass on my local environment~~ Looks like it just emits a message saying that that the files couldn't be formated, nevertheless, we should probably add it to our docs so adding that to the `contributing.md`

(as a side point, when picking who to review this, should I be picking `uniffi-devs` or `uniffi-review`? I notice `uniffi-devs` is the more complete list?)

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -47,7 +47,8 @@ recommend installing the language dependencies on your local machine. In order t
 test suite you will need:
 
 * Kotlin:
-  * `kotlinc`, the [Kotlin command-line compiler](https://kotlinlang.org/docs/command-line.html)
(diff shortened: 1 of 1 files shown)


bendk (approved):
looks good to me.

rfk:
> as a side point, when picking who to review this, should I be picking uniffi-devs or uniffi-review?

I don't quite understand the state of the `uniffi-review` team, it has only two members and no maintainers. I think I will make a github admin request to have it deleted and we can just keep `uniffi-devs`.

PR #493, opened 2021-06-21
Generates cargo clippy safe code

### LINKED ISSUE #257: Generate clippy correct code
Follow up after #256 

We disabled clippy for our generated functions. A better approach would be to have our generated code make clippy happy.

For reference of a log of clippy errors check #254

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/SYNC-1717)

tarikeshaq:
fixes #257 

We had two main issues with the generated code clippy-wise:
- We had a `-> ()` on ffi functions that don't have a return value, clippy prefers if you simply omit that. For example:
```rust
pub exter "C" fn func() -> () {  // clippy prefers if we remove the "-> ()"
.
.
}
```
- We had a `let` binding to functions that don't return `Result`, clippy wanted us to simply return the called library function. For example:
```rust
uniffi::deps::ffi_support::call_with_output(err, || {
let _retval = rust_library_func();
_retval // clippy prefers if instead of those two lines, we just did "rust_library_func()" and returned that
}
```

Before merging, I'll try this out on `fxa_client` and make sure that it indeed generates clippy safe code. I manually installed `uniffi-bindgen` from local source code and ran `cargo clippy` in Application Services and it seems fine, but for some reason I can't locate the generated bindings and I wanna double check that

### DIFF
--- a/uniffi_bindgen/src/scaffolding/templates/ObjectTemplate.rs
+++ b/uniffi_bindgen/src/scaffolding/templates/ObjectTemplate.rs
@@ -40,7 +40,6 @@ pub extern "C" fn {{ ffi_free.name() }}(ptr: *const std::os::raw::c_void) {
 }
 
 {%- for cons in obj.constructors() %}
-    #[allow(clippy::all)]
     #[doc(hidden)]
     #[no_mangle]
     pub extern "C" fn {{ cons.ffi_func().name() }}(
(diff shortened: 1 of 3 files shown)


uniffi_bindgen/src/scaffolding/templates/ObjectTemplate.rs
```diff
@@ -57,12 +56,11 @@ pub extern "C" fn {{ ffi_free.name() }}(ptr: *const std::os::raw::c_void) {
 {%- endfor %}
 
 {%- for meth in obj.methods() %}
-    #[allow(clippy::all)]
     #[doc(hidden)]
     #[no_mangle]
     pub extern "C" fn {{ meth.ffi_func().name() }}(
         {%- call rs::arg_list_ffi_decl(meth.ffi_func()) %}
-    ) -> {% call rs::return_type_func(meth) %} {
+    ) {% call rs::return_value(meth) %} {
```
rfk, line 63: naming nit: this doesn't really represent a return *value*, but then again I don't have a good name for the thing it does represent. Maybe "return signature"?

rfk (approved):
LGTM 👍🏻

tarikeshaq:
Verified against fxa-client locally. I pointed all the uniffi dependencies to my local uniffi source code, made sure the produced scaffolding had the changes implemented here. Then ran `cargo clippy` and it didn't complain!

PR #972, opened 2021-06-23
Replace lower_into_buffer and try_lift_from_buffer with a trait

bendk:
I'm not totally sure about this one, but it seems like it could be an improvement.  What do you all think?

The main motivation is to eliminate the possibility of calling
`try_lift_from_buffer()` on a `RustBuffer` that wasn't created with
`lower_into_buffer()`.  For example, this code seems reasonable, but it
totally fails:

```
let buf: RustBuffer = ViaFfi::lower("Test".to_string());
assert_eq!(try_lift_from_buffer::<String>(buf).unwrap(), "Test")
```

This is because Strings have an asymmetry between `lower()` and
`write()`.  `write()` needs to write out the string length, but
`lower()` can optimize to avoid that and just use `RustBuffer.len`.

It also reduces some boilerplate code, which is nice too.

### DIFF
--- a/uniffi/src/lib.rs
+++ b/uniffi/src/lib.rs
@@ -196,32 +196,6 @@ pub unsafe trait ViaFfi: Sized {
     fn try_read(buf: &mut &[u8]) -> Result<Self>;
 }
 
-/// A helper function to lower a type by serializing it into a buffer.
-///
-/// For complex types were it's too fiddly or too unsafe to convert them into a special-purpose
-/// C-compatible value, you can use this helper function to implement `lower()` in terms of `write()`
-/// and pass the value as a serialized buffer of bytes.
-pub fn lower_into_buffer<T: ViaFfi>(value: T) -> RustBuffer {
-    let mut buf = Vec::new();
-    ViaFfi::write(&value, &mut buf);
-    RustBuffer::from_vec(buf)
-}
-
-/// A helper function to lift a type by deserializing it from a buffer.
-///
-/// For complex types were it's too fiddly or too unsafe to convert them into a special-purpose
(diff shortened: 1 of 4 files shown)


rfk (approved):
I like it! It's a nice cleanup, and also reduces potential for confusion 👍🏻.

uniffi/src/lib.rs
```diff
@@ -196,32 +196,6 @@ pub unsafe trait ViaFfi: Sized {
     fn try_read(buf: &mut &[u8]) -> Result<Self>;
 }
 
-/// A helper function to lower a type by serializing it into a buffer.
-///
-/// For complex types were it's too fiddly or too unsafe to convert them into a special-purpose
-/// C-compatible value, you can use this helper function to implement `lower()` in terms of `write()`
-/// and pass the value as a serialized buffer of bytes.
-pub fn lower_into_buffer<T: ViaFfi>(value: T) -> RustBuffer {
```
rfk, line 204: I guess *technically* this is a breaking change, swapping some `pub` functions for a `pub` trait, and as such should be called out in the changelog.

uniffi/src/lib.rs
```diff
@@ -364,6 +338,44 @@ unsafe impl ViaFfi for String {
     }
 }
 
+/// A helper trait to implement lowering/lifting using a `RustBuffer`
+///
+/// For complex types were it's too fiddly or too unsafe to convert them into a special-purpose
```
mhammond, line 343: typo - "where" (that typo already existed :)

mhammond (commented):
looks great - only adding this comment as I noticed a stray typo that you didn't even create, just moved :)

PR #975, opened 2021-07-06
Document optional fields & arguments

badboy:
While playing around with UniFFI, in order to figure out how we can use it in Glean, I came across these essential things for us: default values.

While implemented it lacks documentation, so I went ahead and added that.

### DIFF
--- a/docs/manual/src/udl/structs.md
+++ b/docs/manual/src/udl/structs.md
@@ -63,3 +63,72 @@ destroyed to avoid leaking the underlying memory, and this also applies to Objec
 in record fields.
 
 You can read more about managing object references in the section on [interfaces](./interfaces.md).
+
+## Default values for fields
+
+Fields can be specified with a default value:
+
+```idl
+dictionary TodoEntry {
+    boolean done = false;
+    string text;
+};
+```
+
+The corresponding generated Kotlin code would be equivalent to:
+
+```kotlin
+data class TodoEntry (
+    var done: Boolean = false,
+    var text: String
+)  {
+    // ...
+}
+```
+
+This works for Swift and Python targets too.
+If not set otherwise the default value for a field is passed to the Rust constructor.
+
+## Optional fields and default values
+
+Fields can be made optional using a `T?` type.
+
+```idl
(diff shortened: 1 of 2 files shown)


docs/manual/src/udl/structs.md
```diff
@@ -57,9 +57,78 @@ struct TodoEntry {
...
+    // ...
+}
+```
+
+This works for Swift and Python targets too.
+If not set otherwise the default value for a field is passed to the Rust constructor.
+
+## Optional fields and default values
+
+Fields can be made optional using a `T?` type.
```
badboy, line 94: I could be convinced that this is not necessary, as the `T?` thing is already document in Types and thus having default values for it with `null` is covered by the above paragraph.
I still figured it's worth calling out specifically here.

Just let me know what you think, I can change that.

audkar:
Nice! Just  found out about `uniffi` today, and was looking at default values myself. What a coincident. 

Also would be nice to have working example showcasing this functionality. In some cases (if CI includes examples) code example are easier to maintain then static docs which can go out-of-date and bring more frustration then benefits to users...

badboy:
> Also would be nice to have working example showcasing this functionality. In some cases (if CI includes examples) code example are easier to maintain then static docs which can go out-of-date and bring more frustration then benefits to users...

Sure, but let's start simple

docs/manual/src/udl/structs.md
mhammond: I don't think it hurts to mention it here too. However, I suggest maybe adding what the rust receiver looks like too (ie, takes an Option<T>), just like you've done above for the defaults.

mhammond (commented):
This looks great to me, but we might as well wait a few days for Ryan to weigh in.

docs/manual/src/udl/structs.md
rfk: I agree it seems worth calling out here as well, it may be easier to understand with these concrete examples.

rfk (approved):
Thanks for diving in! I'll be excited to hear how your explorations went for Glean, whether positive or negative :-)

PR #977, opened 2021-07-08
Add @ExperimentalUnsignedTypes to unsigned types in kotlin

skhamis:
This PR is building off of what @rfk started in this PR https://github.com/mozilla/uniffi-rs/pull/445

resolves https://github.com/mozilla/uniffi-rs/issues/444

Instead of using the compiler option of `-Xopt-in=kotlin.ExperimentalUnsignedTypes` to opt-in on the whole crate which forces any consumers to also use the compiler directive, we annotate with `@ExperimentalUnsignedTypes` if we detect any unsigned types. This will allow the consumer to either use the compiler option or annotate depending on their use case 

**Note:** This is not needed after [kotlin version 1.5 which now stabilizes unsigned types](https://kotlinlang.org/docs/whatsnew15.html#stable-unsigned-integer-types)


tarikeshaq (commented):
Adding a quick review just because I got a bit of time and was curious, it looks good to me (not that I would be able to catch any big red flags lol)

I am curious about the whole Kotlin 1.5 thou, like at what point will we be able to get rid of those annotations? I'm not quite sure about how many Android versions backward our consumers support, I wonder if we should put a disclaimer somewhere (or at least for our own sanity) what the lowest version of Kotlin we support

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -241,6 +241,40 @@ impl<'ci> ComponentInterface {
...
+    ///
+    /// In Kotlin, unsigned types are expiremental and will cause a warning if used without annotations
+    /// This should be removed when kotlin version
+    ///  > v1.5 https://kotlinlang.org/docs/whatsnew15.html#stable-unsigned-integer-types
+    pub fn contains_unsigned_type(&self, type_: &Type) -> bool {
+        match type_ {
+            Type::UInt8 => true,
+            Type::UInt16 => true,
+            Type::UInt32 => true,
+            Type::UInt64 => true,
```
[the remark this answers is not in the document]
skhamis: > Adding a quick review just because I got a bit of time and was curious, it looks good to me (not that I would be able to catch any big red flags lol)
> 
> I am curious about the whole Kotlin 1.5 thou, like at what point will we be able to get rid of those annotations? I'm not quite sure about how many Android versions backward our consumers support, I wonder if we should put a disclaimer somewhere (or at least for our own sanity) what the lowest version of Kotlin we support

Good question, I think it's tricky since the major downstream consumers for kotlin is a-c & Fenix and we probably won't be able to update to kotlin 1.5 until Fenix itself does to prevent any incompatibility.

I do think that's a good idea to keep a list of version compatibilities! It'll def help give context why certain things are done certain ways  -- I was also thinking about making a task to revisit this once Fenix updates their kotlin version but we'll see if this PR is even a good idea currently

uniffi_bindgen/src/bindings/kotlin/templates/TopLevelFunctionTemplate.kt
```diff
@@ -1,6 +1,9 @@
 {%- match func.return_type() -%}
 {%- when Some with (return_type) %}
 
+// SAM_TODO: There is one more error I need to figure out how to also catch
+// If any arguments have a long type they will need the annotation
```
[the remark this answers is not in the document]
skhamis: ah yes definitely unsigned! I somehow missed this TODO trying to get this PR up lol yes iterating through the args would be the way to go I think I found a nice neat little solution with only one extra method and it looks like that works so I'll get this fixed up ASAP!

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -241,6 +241,42 @@ impl<'ci> ComponentInterface {
         }
     }
 
+    /// Identify if we are using an unsigned type in kotlin
```
rfk, line 244: Nit: there's nothing Kotlin-specific about it at this level, so the "in kotlin" here risks confusion. (Explaining that we need this for the Kotlin bindings makes sense, but it does the same thing regardless of which bindings are being generated).

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -241,6 +241,42 @@ impl<'ci> ComponentInterface {
...
+                })
+                .unwrap_or(false),
+            _ => false,
+        }
+    }
+
+    pub fn args_contains_unsigned(&self, args: &[&Argument]) -> bool {
+        args.iter()
+            .any(|&arg| self.contains_unsigned_type(&arg.type_()))
+    }
```
rfk, line 278: As mentioned above, I think this would be better situated as part of a `contains_unsigned_type` method on the `Function`, `Constructor` and `Method` structs. It's more code but IMHO would make things more predictable as a whole.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -241,6 +241,42 @@ impl<'ci> ComponentInterface {
         }
     }
 
+    /// Identify if we are using an unsigned type in kotlin
+    ///
+    /// In Kotlin, unsigned types are expiremental and will cause a warning if used without annotations
+    /// This should be removed when kotlin version
+    ///  > v1.5 https://kotlinlang.org/docs/whatsnew15.html#stable-unsigned-integer-types
+    pub fn contains_unsigned_type(&self, type_: &Type) -> bool {
```
rfk, line 249: Another naming consistency nit: we have `type_contains_object_references` as plural with a leading `type_` prefix, and `contains_unsigned_type` as singular with no prefix. I think it's worth striving for consistency in the names of methods that do similarly-shaped things, especially given how much API surface this `ComponentInterface` struct exposes.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -241,6 +241,42 @@ impl<'ci> ComponentInterface {
...
+        match type_ {
+            Type::UInt8 | Type::UInt16 | Type::UInt32 | Type::UInt64 => true,
+            Type::Optional(t) | Type::Sequence(t) | Type::Map(t) => self.contains_unsigned_type(t),
+            Type::Record(name) => self
+                .get_record_definition(name)
+                .map(|rec| {
+                    rec.fields()
+                        .iter()
+                        .any(|f| self.contains_unsigned_type(&f.type_))
+                })
```
rfk, line 259: I find myself wondering if this "iterate over each field" logic should live in `Record::contains_unsigned_type` struct and we just call `rec.contains_unsigned_type(self)` here, rather than having this `ComponentInterface::contains_unsigned_type` method know too many details about the internals of the `Record` struct. But I note that the existing `ComponentInterface::type_contains_object_references` uses the same pattern as you've got here, so my past self clearly didn't think it was too big of a deal 😅.

If you do decide to add similar helper methods to the `Function`/`Constructor`/`Method` structs in addition to `Enum` and `Record`, then it's probably worth pushing the implementation details for `Enum` and `Record` out into those structs for consistency, rather than having them inline here.

rfk (changes requested):
This looks like it came together pretty well! The "request changes" is specifically about:

* I think there are some missing cases, such as functions and methods without return values; we should add tests for cover those as well.
* I suspect the generated code was checked in by accident (or, if it was on purpose, we should document the reasons why).

I'm a little surprised at just how many places need the annotation added, but I think it'll be worthwhile (and pretty easy to remove once we get to the appropriate version of Kotlin).

uniffi_bindgen/src/bindings/kotlin/templates/RecordTemplate.kt
```diff
@@ -1,5 +1,6 @@
 data class {{ rec.name()|class_name_kt }} (
     {%- for field in rec.fields() %}
+    {% if rec.contains_unsigned_type(ci) %}@ExperimentalUnsignedTypes{% endif %}
```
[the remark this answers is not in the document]
skhamis: ah good catch. I actually had a separate method `rec.contains_unsigned_type(field)` that would do it correctly based on the type but was trying to consolidate the methods as much as I  can and missed that this would happen. I might quickly see how adding it to the whole data class would be vs individual fields though it'd be a little less convenient for the consumer

uniffi_bindgen/src/interface/mod.rs
skhamis: hmm that's an interesting pattern -- I definitely modeled it after the `type_contains_object_references` in which the structs are actually simply calling the `ComponentInterface` under the hood (rather than CI calling the respective structs itself). However I think that may be the way to approach it as you're right it could abstract away the implementation of these different types, not to mention adding new types is more straightforward. I think I'll give this a shot.

uniffi_bindgen/src/interface/mod.rs
skhamis: I think i was trying to avoid a semi-weird name like `type_contains_unsigned_types` but I think I'll move to that for consistency

skhamis:
> This looks like it came together pretty well! The "request changes" is specifically about:
> 
>     * I think there are some missing cases, such as functions and methods without return values; we should add tests for cover those as well.
> 
>     * I suspect the generated code was checked in by accident (or, if it was on purpose, we should document the reasons why).
> 
> 
> I'm a little surprised at just how many places need the annotation added, but I think it'll be worthwhile (and pretty easy to remove once we get to the appropriate version of Kotlin).

Appreciate the thorough review! Learned quite a bit of the internals just trying to use the macros and trying to get the `ComponentIntertface` using the other types to get those proposed changes in. I think this iteration is much improved and much more consistent and imo, easier to follow as well!

uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -113,6 +113,14 @@ impl Enum {
         // and its contained types could use a bit of a cleanup.
         ci.type_contains_object_references(&Type::Enum(self.name.clone()))
     }
+
+    pub fn type_contains_unsigned_types(&self, ci: &ComponentInterface) -> bool {
```
rfk, line 117: Reverse of the previous naming nit 😅 - on these structs we've now got `contains_object_references` without a `type_` prefix and `type_contains_unsgined_types` with one. I lightly prefer to remove the `type_` prefix on the name here because the subject of the predicate is the method receiver itself, so it'll get called like `e.contains_unsigned_types` which I think reads succinctly.

uniffi_bindgen/src/interface/function.rs
```diff
@@ -90,6 +90,21 @@ impl Function {
         self.ffi_func.return_type = self.return_type.as_ref().map(|rt| rt.into());
         Ok(())
     }
+
+    // Intentionally exactly the same as the Method version
```
rfk, line 94: Aye, it'd be cool to factor some of the duplication out between `Function`/`Method`/`Constructor`, maybe as a `Callable` trait with default implementations. But, not for this PR.

uniffi_bindgen/src/interface/function.rs
```diff
@@ -90,6 +90,21 @@ impl Function {
...
+    pub fn type_contains_unsigned_types(&self, ci: &ComponentInterface) -> bool {
+        let check_return_type = {
+            match self.return_type() {
+                None => false,
+                Some(t) => ci.type_contains_unsigned_types(t),
+            }
+        };
```
rfk, line 101: A fun Rust thing you could try here, is to turn `return_type()` into an iterator and chain it with the arguments to check them all at once. Something like:

```
self.return_type().iter()
    .chain(self.arguments().iter().map(|arg|, arg.type_))
    .any(|t| ci.type_contains_unsigned_type(t))
```

It might turn out to be a bit awkward with the types because `return_type()` returns a borrow but `arg.type_()` returns a clone, but it's be interesting to try.

(To be clear: the existing code you have here is also fine to merge as-is, this is purely a "hey, good opportunity to learn more about Rust iterators if you're interested!" situation).

rfk (approved):
Good stuff! I left a couple of nits from my final read-through which you're welcome to take or leave, but I think this came together well. Moving the annotations up to the class/enum level seems to have helped simplify things quite a bit.

fixtures/regressions/kotlin-experimental-unsigned-types/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+    member: u8,
+    member_two: u16,
+    member_three: u32,
+    member_four: u64,
+    other: String,
+}
+
+pub struct RecursivelyUsesU8 {
+    member: DirectlyUsesU8,
+}
```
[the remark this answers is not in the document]
skhamis: Thanks for this, I actually ended up catching a few more areas from a more robust UDL file.

uniffi_bindgen/src/interface/function.rs
skhamis: oh interesting! I couldn't seem to get it to properly compile after time-boxing however feel pretty good about using chain in the future with how much I tried to wrangle with it lol

Issue #460, opened 2021-05-25
Support rich structured Enums for reporting errors


skhamis:
## Problem 
Currently Uniffi does not allow too much flexibility in terms of having nested enums which makes things a bit tricky when trying to create richer structured error classes.

In the current effort to [uniffi logins component in application-services](https://github.com/mozilla/application-services/pull/4068) we've run into a bit of limitations in porting over some of the errors.

## Proposed solution
Having a structure similar to dictionaries/interfaces in the `.udl` file might allow the generation of these errors to be more robust. Instead of:
```
[Error]
enum LoginsStorageError {
    "UnexpectedLoginsStorageError",
    "SyncAuthInvalid",
    "MismatchedLock",
    "NoSuchRecord",
    "IdCollision",
    "InvalidRecord",
    "InvalidKey",
    "RequestFailed",
    "Interrupted",
};
```

We could do something like
```
[Error]
enum LoginsStorageError {
   String UnexpectedLoginsStorageError,
   InvalidReason InvalidRecord,
   ...
};

enum InvalidReason {
  "EmptyPassword",
  "EmptyOrigin",
   ...
}

```

This will allow having enum (or other) types be expected and consequently allow for richer errors to be passed over.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-60)
┆Epic: UniFFI
┆Issue Number: UNIFFI-60
┆Sprint End Date: 2021-08-06

rfk:
I really like the gist of this proposal, and see a lot of value in being able to pass richer structured error information. We're probably going to have to fiddle around to find a good syntax that fits within the constraints of our current IDL. Perhaps along the lines of the current rich enum syntax, which is:

```
[Enum]
interface {
  Variant(Type field, ...);
  ....
}
```

Allowing `Error` to use the same structure might work well for our needs here:

```
[Error]
interface {
  Variant(Type field, ...);
  ...
}
```

I think maintaining the symmetry of errors being a special kind of enum makes sense.

For context for @skhamis and others: UniFFI currently (ab)uses an off-the-shelf WebIDL parser, which really helped us get up and running quickly, but means we're syntactically limited to what's supported by WebIDL. There's some discussion of lightly hacking our way around this limitation for enums in https://github.com/mozilla/uniffi-rs/pull/381.

skhamis:
[...]

PR #980, opened 2021-07-13
WIP: Rich structured error handling (#460)

bendk:
This is definitely not ready to merge, but I think it's close enough to support a discussion on the general approach.  The Rust, Python, Ruby, and Kotlin code are working well enough for the coverall tests to pass.  The next step would be to implement swift.  I'm also hoping to simplify the template code far creating function calls, like @rfk suggests in the issue.

### DIFF
--- a/fixtures/coverall/tests/bindings/test_coverall.py
+++ b/fixtures/coverall/tests/bindings/test_coverall.py
@@ -75,7 +75,7 @@ class TestCoverall(unittest.TestCase):
         with self.assertRaises(CoverallError.TooManyHoles):
             Coveralls.fallible_new("", True)
 
-        with self.assertRaisesRegex(InternalError, "expected panic: woe is me"):
+        with self.assertRaisesRegex(InternalError, "Rust panic"):
             Coveralls.panicing_new("expected panic: woe is me")
 
         # in the absence of cycles Python is deterministic in killing refs
@@ -85,15 +85,35 @@ class TestCoverall(unittest.TestCase):
         self.assertEqual(get_num_alive(), 0)
 
 
-    def test_errors(self):
+    def test_simple_errors(self):
         coveralls = Coveralls("test_errors")
         self.assertEqual(coveralls.get_name(), "test_errors")
 
         with self.assertRaises(CoverallError.TooManyHoles):
             coveralls.maybe_throw(True)
 
-        with self.assertRaisesRegex(InternalError, "expected panic: oh no"):
+        with self.assertRaisesRegex(InternalError, "Rust panic"):
             coveralls.panic("expected panic: oh no")
+
+    def test_complex_errors(self):
+        coveralls = Coveralls("test_complex_errors")
+
+        # Test success
+        self.assertEqual(True, coveralls.maybe_throw_complex(0))
+
+        # Test errors
+        with self.assertRaises(ComplexError.OsError) as cm:
+            coveralls.maybe_throw_complex(1)
+        self.assertEqual(cm.exception.code, 10)
+        self.assertEqual(cm.exception.extended_code, 20)
+
+        with self.assertRaises(ComplexError.PermissionDenied) as cm:
+            coveralls.maybe_throw_complex(2)
+        self.assertEqual(cm.exception.reason, "Forbidden")
+
+        # Test panics, which should cause InternalError to be raised
+        with self.assertRaises(InternalError) as cm:
+            coveralls.maybe_throw_complex(3)
 
     def test_self_by_arc(self):
         coveralls = Coveralls("test_self_by_arc")
@@ -116,7 +136,7 @@ class TestCoverall(unittest.TestCase):
         with self.assertRaises(CoverallError.TooManyHoles):
             coveralls.take_other_fallible()
 
-        with self.assertRaisesRegex(InternalError, "expected panic: with an arc!"):
+        with self.assertRaisesRegex(InternalError, "Rust panic"):
             coveralls.take_other_panic("expected panic: with an arc!")
 
         coveralls.take_other(None)
--- a/fixtures/coverall/tests/bindings/test_coverall.rb
+++ b/fixtures/coverall/tests/bindings/test_coverall.rb
@@ -80,10 +80,6 @@ class TestCoverall < Test::Unit::TestCase
       Coverall::Coveralls.panicing_new('expected panic: woe is me')
     end
 
-    assert_raise_message /expected panic: woe is me/ do
-      Coverall::Coveralls.panicing_new('expected panic: woe is me')
-    end
-
     begin
       objects = 10.times.map { Coverall::Coveralls.new 'c1' }
       assert_equal 12, Coverall.get_num_alive
@@ -94,9 +90,9 @@ class TestCoverall < Test::Unit::TestCase
     assert_equal 2, Coverall.get_num_alive
   end
 
-  def test_errors
-    coveralls = Coverall::Coveralls.new 'test_errors'
-    assert_equal coveralls.get_name, 'test_errors'
+  def test_simple_errors
+    coveralls = Coverall::Coveralls.new 'test_simple_errors'
+    assert_equal coveralls.get_name, 'test_simple_errors'
 
     assert_raise Coverall::CoverallError::TooManyHoles do
       coveralls.maybe_throw true
@@ -105,9 +101,31 @@ class TestCoverall < Test::Unit::TestCase
     assert_raise Coverall::InternalError, 'expected panic: oh no' do
       coveralls.panic 'expected panic: oh no'
     end
+  end
 
-    assert_raise_message /expected panic: oh no/ do
-      coveralls.panic 'expected panic: oh no'
+  def test_complex_errors
+    coveralls = Coverall::Coveralls.new 'test_complex_errors'
+    assert_equal coveralls.maybe_throw_complex(0), true
+
+    begin
+      coveralls.maybe_throw_complex(1)
+    rescue Coverall::ComplexError::OsError => err
+      assert_equal err.code, 10
+      assert_equal err.extended_code, 20
+    else
+      raise 'should have thrown'
+    end
+
+    begin
+      coveralls.maybe_throw_complex(2)
+    rescue Coverall::ComplexError::PermissionDenied => err
+      assert_equal err.reason, "Forbidden"
+    else
+      raise 'should have thrown'
+    end
+
+    assert_raise Coverall::InternalError do
+      coveralls.maybe_throw_complex(3)
     end
   end
 
@@ -147,7 +165,7 @@ class TestCoverall < Test::Unit::TestCase
     begin
       coveralls.take_other_panic "expected panic: with an arc!"
     rescue Coverall::InternalError => err
-      assert_match /expected panic: with an arc!/, err.message
+      # OK
     else
       raise 'should have thrown'
     end
--- a/uniffi/src/ffi/rustcalls.rs
+++ b/uniffi/src/ffi/rustcalls.rs
@@ -1,0 +1,189 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! # Low-level support for calling rust functions
+//!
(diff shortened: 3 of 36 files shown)


uniffi/src/ffi/rustcalls.rs
```diff
@@ -0,0 +1,218 @@
...
+use std::panic;
+use std::mem::MaybeUninit;
+
+/// Represents the success/error of a rust call
+///
+/// ## Usage
+///
+/// - The consumer code allocates a `RustCallStatus`
+///    - The error_buf field can either be uninitialized or initialized to zero-capacity and a null
+///      data buffer.  The data here is going to be overwritten so any allocations will leak memory.
```
bendk, line 25: So basically the way we're using `RustBuffer` is:
  - The consumer passes in a zeroed out rust buffer struct
  - In the case of an Error, we overwrite that data with a `RustBuffer` that contains the error.

Is this usage supported or should it be?  If so I think we should update the `RustBuffer` documentation to mention that.  If not, is there a better way to achieve this?

uniffi/src/ffi/rustcalls.rs
```diff
@@ -0,0 +1,218 @@
...
+const CALL_SUCCESS: i8 = 1;
+const CALL_ERROR: i8 = 2;
+const CALL_PANIC: i8 = 3;
+
+
+// Generalized rust call handling function
+fn make_call<F, R>(out_status: &mut RustCallStatus, callback: F) -> R::Value
+where
+    F: panic::UnwindSafe + FnOnce() -> Result<R, RustBuffer>,
+    R: IntoFfi,
```
bendk, line 76: I would really love to get rid of the dependency on `ffi_support::InfoFfi` here.  The only thing we're using it for is to define a value that we can return for errors/panics, which the consumer code will just throw away.  I can think of a couple ways, but I'm not sure if they make sense and are in scope for this project:

  - Return a `MaybeUninit<>` value.  This feels like the cleanest solution, but then it would require quite a few changes to the scaffolding templates.
  - Add a `uniffi` trait that defines a error return value for each `ViaFFI::FfiType`.  One issue with this is it would be a breaking change for any consumers that defined their own `ViaFfi` impls with other `FfiType` types.

uniffi_bindgen/src/bindings/ruby/templates/ErrorTemplate.rb
```diff
@@ -1,50 +1,89 @@
...
-    {% endfor %}
-    raise 'Unknown error code'
+# Map error modules to the RustBuffer method name that reads them
+@error_module_to_reader_method = {
+{%- for e in ci.iter_error_definitions() %}
+{%- let typ=ci.get_type(e.name()).unwrap() %}
+{%- let canonical_type_name = typ.canonical_name()|class_name_rb %}
+  {{ e.name()|class_name_rb }} => :read{{ canonical_type_name }},
+{%- endfor %}
+}
```
bendk, line 48: Would it be a good idea to make `error_module_to_reader_method` private somehow or just flagging it internal like in Python?  I don't know nearly enough about Ruby here.

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -1,149 +1,95 @@
...
-    {% endfor %}
-}
-
 {% endfor %}
 
-// Helpers for calling Rust with errors:
+// Helpers for calling Rust
 // In practice we usually need to be synchronized to call this safely, so it doesn't
 // synchronize itself
-private inline fun <U, E: RustErrorReference> nullableRustCall(callback: (E) -> U?, err: E): U? {
```
bendk, line 133: As far as I could tell, we never called `nullableRustCall` so I removed support for it.  Tell me if that's a bad idea.

bendk:
[coverall-generated-code.zip](https://github.com/mozilla/uniffi-rs/files/6811735/coverall-generated-code.zip)

Here's a copy of the generated code for the coverall fixture.  I found it much easier to reason about the error handling templates when I could see the output.

uniffi/src/ffi/rustcalls.rs
rfk: Hrm, interesting. I think you're right the `MaybeUninit` feels like the right thing, but I don't have a good sense of how disruptive that would be.

I suppose an alternative would be to do what Rust does under the hood, and actually return a `union` that can contain the error buffer *or* the success value. But that feels likely to be clunky for common cases where we're returning primitive values.

rfk (commented):
Thanks @bendk, some initial comments below.

uniffi/src/ffi/rustcalls.rs
bendk: That makes a lot of sense.  For some reason I was thinking that `RustBuffer`s were intended to be immutable, but after re-reading the docs I realize that's not the case.  Is it generally true that passing in a `RustBuffer` to a function means transferring ownership and therefore the calling code shouldn't try to access it once they pass it in?  If so, maybe I'll add a bullet point to the `RustBuffer` doc comment about that.

fixtures/coverall/tests/bindings/test_coverall.py
```diff
@@ -75,7 +75,7 @@ def test_constructors(self):
         with self.assertRaises(CoverallError.TooManyHoles):
             Coveralls.fallible_new("", True)
 
-        with self.assertRaisesRegex(InternalError, "expected panic: woe is me"):
+        with self.assertRaisesRegex(InternalError, "Rust panic"):
```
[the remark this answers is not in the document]
bendk: Oops, meant to flag this but I forgot.  I was worried about what happens if the code to extract the string itself panics.  For example, it seems like the current code could panic: https://github.com/mozilla/ffi-support/blob/31ffbbd6eec90038d825dab145c5382ca248b2f4/src/error.rs#L257.  I think calling `to_string()` on a `str` shouldn't panic.  But couldn't calling `clone()` on a `String` panic when trying to allocate the new string?  There's also the issue of converting the `String` to a `RustBuffer`, which I think should be panic-free, but I'm not sure about that.

I was hoping that the panic hook could log enough info to debug things, maybe as a Sentry breadcrumb.  I'm not sure if this will work though.  (BTW, what happens if the panic hook panics?  I couldn't find any documentation on this).

If that's not possible, then I think we should put the code to extract the panic string in a second `catch_unwind` block.

uniffi_bindgen/src/bindings/python/templates/RustBufferStream.py
```diff
@@ -163,6 +163,34 @@ def read{{ canonical_type_name }}(self):
...
+    {% when Type::Error with (error_name) -%}
+    {%- let e = ci.get_error_definition(error_name).unwrap().wrapped_enum() %}
+
+    # The Error type {{ error_name }}
+
+    # Top-level read method
+    def read{{ canonical_type_name }}(self):
+        variant = self._unpack_from(4, ">i")
+        try:
+            read_variant_method = getattr(self, 'read{{canonical_type_name}}{}'.format(variant))
```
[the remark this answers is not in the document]
bendk: I forgot about this one, but since `variant` is an integer here,  maybe it's okay.  I don't think `MyEnumeration::1` is a valid name.

uniffi_bindgen/src/bindings/ruby/templates/ErrorTemplate.rb
bendk: My second attempt is to make the variable name upper case (`ERROR_MODULE_TO_READER_METHOD`).  That makes the code work, although I'm still not sure if it's idomatic Ruby.

bendk:
@rfk Thanks for all the review!  I just pushed some changes based on your comments:

 - Changed `STATUS_SUCCESS` to be 0
 - Specified that the caller should initialize `RustCallStatus` to STATUS_SUCCESS and an empty rust buffer
 - Fixed a bunch of spelling issues
 - Updated the changelog
 - Deleted `ErrorTemplate.rs`
 
TODO:
  - Implement swift
  - Figure out how to handle panics messages
  - Figure out the error values

uniffi/src/ffi/rustcalls.rs
bendk: I thought about a union and it could be very nice.  I think the fact that it's going to be sized larger than most primitives should be fine since we get to avoid creating the `RustCallStatus` out param.  The issue I couldn't figure it out was how to deal with all the different return types:
  - You could make a separate union for each return type `union U8Return { value: u8, error: RustBuffer }`, but this is going to be a lot of unions both on the scaffolding and bindings side.
  - You could make a single union that covers all return types, but then if 1 return type is very large then the entire union is going to be inefficient.

One question I had is if we could make any assumptions on which `FfiType` instances can be returned.  I think uniffi only uses integers, `void *`, and `RustBuffer`.  Can we assume that the consumer code isn't going to use those too, or could they return some exotic type?

uniffi_bindgen/src/bindings/ruby/templates/ErrorTemplate.rb
saks: I had a quick look. Having this mapping defined as a constant `ERROR_MODULE_TO_READER_METHOD` is definitely a good idea and quite idiomatic for ruby as for me. There are ways to mark constants/classes/modules as private but language still allows these to be accessed with simple meta-programming techniques.

I'll spend more time on this later today.
Thanks!

uniffi/src/ffi/rustcalls.rs
rfk: > One question I had is if we could make any assumptions on which FfiType instances can be returned. I think uniffi
> only uses integers, void *, and RustBuffer. Can we assume that the consumer code is going to use those too, or could
> they return some exotic type?

This is probably true in practice right now, but architecturally it feels like we'll make a better longer-term design if we do not assume this.

I also think it would be totally fine to punt this to followup work and continue to use `IntoFfi` for the first round.

fixtures/coverall/tests/bindings/test_coverall.rb
```diff
@@ -80,10 +80,6 @@ def test_constructors
       Coverall::Coveralls.panicing_new('expected panic: woe is me')
     end
 
-    assert_raise_message /expected panic: woe is me/ do
-      Coverall::Coveralls.panicing_new('expected panic: woe is me')
-    end
```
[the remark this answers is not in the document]
saks: FYI: this test passes for me on this branch if I change it to:
```ruby
    assert_raise_message 'Rust panic' do
      Coverall::Coveralls.panicing_new('Rust panic')
    end
```

uniffi_bindgen/src/bindings/ruby/templates/ErrorTemplate.rb
saks: I marked few constants as private @bendk, please have a look https://github.com/bendk/uniffi-rs/pull/1

fixtures/coverall/tests/bindings/test_coverall.py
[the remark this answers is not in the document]
rfk: Yes, you're right, I believe at least the allocation underlying `String::clone` could fail in this case and trigger a panic.

> I was hoping that the panic hook could log enough info to debug things, maybe as a Sentry breadcrumb.

We haven't had good luck getting rust-level debugging output into Sentry, although it's probably time we tried again. 

> I think we should put the code to extract the panic string in a second catch_unwind block.

Making a best-effort attempt to get the string out along these lines seems valuable to me 👍🏻

fixtures/coverall/tests/bindings/test_coverall.rb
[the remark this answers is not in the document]
bendk: The original reasoning was that I couldn't match the message anymore and I thought there was another test case for simply checking if the exception was raised.  I'm going to try to re-implement the panic message functionality then add this test back.

uniffi_bindgen/src/bindings/ruby/templates/ErrorTemplate.rb
bendk: @saks that's wonderful, thanks so much!  I just merged them into the branch I'm working on, they'll be in the final product.

bendk:
Thanks for all the help with this one.  I think the branch is getting close to ready.  I'm going to close this one and open a new PR for it.

PR #981, opened 2021-07-17
Prevent generation of unused swift bindings

### LINKED ISSUE #362: Swift bindings should not generate helper methods for unused types
The Kotlin and Python bindings inspect the set of types used by the UDL and only generate helper methods for the types that are in use. The Swift bindings unconditionally generate helpers for all primitive types.

For consistency (and potentially for a small win in the size of the final resulting artifacts) we should update the Swift bindings to avoid generating helper methods for unused types.  (And we'll need to watch out for https://github.com/mozilla/uniffi-rs/pull/357 when we do so).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-46)
┆Epic: UniFFI
┆Issue Number: UNIFFI-46
┆Sprint End Date: 2021-08-06

lougeniaC64:
Fixes #362 by conditionally providing swift binding depending on the types included in the component interface.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
@@ -1,3 +1,7 @@
+// For every type used in the interface, we provide helper methods for conveniently
+// lifting and lowering that type from C-compatible data, and for reading and writing
+// values of that type in a buffer.
+
 // Helper classes/extensions that don't change.
 // Someday, this will be in a libray of its own.
 
@@ -161,7 +165,77 @@ extension ViaFfiUsingByteBuffer {
 }
 
 // Implement our protocols for the built-in types that we use.
+{% if ci.contains_optional_types() %}
+extension Optional: ViaFfiUsingByteBuffer, ViaFfi, Serializable where Wrapped: Serializable {
+    fileprivate static func read(from buf: Reader) throws -> Self {
+        switch try buf.readInt() as Int8 {
+        case 0: return nil
+        case 1: return try Wrapped.read(from: buf)
+        default: throw UniffiInternalError.unexpectedOptionalTag
+        }
+    }
 
+    fileprivate func write(into buf: Writer) {
+        guard let value = self else {
+            buf.writeInt(Int8(0))
+            return
+        }
+        buf.writeInt(Int8(1))
+        value.write(into: buf)
+    }
+}
+{% endif %}
+
+{% if ci.contains_sequence_types() %}
+extension Array: ViaFfiUsingByteBuffer, ViaFfi, Serializable where Element: Serializable {
+    fileprivate static func read(from buf: Reader) throws -> Self {
+        let len: Int32 = try buf.readInt()
+        var seq = [Element]()
+        seq.reserveCapacity(Int(len))
+        for _ in 0..<len {
+            seq.append(try Element.read(from: buf))
+        }
+        return seq
+    }
+
+    fileprivate func write(into buf: Writer) {
+        let len = Int32(self.count)
+        buf.writeInt(len)
+        for item in self {
+            item.write(into: buf)
+        }
+    }
+}
+{% endif %}
+
+{% if ci.contains_map_types() %}
+extension Dictionary: ViaFfiUsingByteBuffer, ViaFfi, Serializable where Key == String, Value: Serializable {
+    fileprivate static func read(from buf: Reader) throws -> Self {
+        let len: Int32 = try buf.readInt()
+        var dict = [String: Value]()
+        dict.reserveCapacity(Int(len))
+        for _ in 0..<len {
+            dict[try String.read(from: buf)] = try Value.read(from: buf)
+        }
+        return dict
+    }
+
+    fileprivate func write(into buf: Writer) {
+        let len = Int32(self.count)
+        buf.writeInt(len)
+        for (key, value) in self {
+            key.write(into: buf)
+            value.write(into: buf)
+        }
+    }
+}
+{% endif %}
+
+{% for typ in ci.iter_types() %}
+{% let canonical_type_name = typ.canonical_name()|class_name_swift %}
+{%- match typ -%}
+
+{% when Type::String -%}
 extension String: ViaFfi {
     fileprivate typealias FfiType = RustBuffer
 
@@ -204,7 +278,7 @@ extension String: ViaFfi {
     }
 }
 
-
+{% when Type::Boolean -%}
 extension Bool: ViaFfi {
     fileprivate typealias FfiType = Int8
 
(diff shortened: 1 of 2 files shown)


uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -456,3 +485,18 @@ extension Dictionary: ViaFfiUsingByteBuffer, ViaFfi, Serializable where Key == S
         }
     }
 }
+
+{% when Type::Enum with (enum_name) -%}
```
lougeniaC64, line 489: @rfk I grabbed the applicable when clauses from the kotlin helper file. Let me know if this looks right.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -397,16 +418,20 @@ extension Double: Primitive, ViaFfi {
     }
 }
 
-extension Optional: ViaFfiUsingByteBuffer, ViaFfi, Serializable where Wrapped: Serializable {
-    fileprivate static func read(from buf: Reader) throws -> Self {
+{% when Type::Optional  with (inner_type) -%}
+{% let inner_type_name = inner_type|type_swift %}
+
+extension Optional {
+// extension Optional: ViaFfiUsingByteBuffer, ViaFfi, Serializable where Wrapped: Serializable {
```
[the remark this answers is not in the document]
rfk: The downside of the approach you're taking here, is that it complicates the expressions for lifting and lowering values of these types. For example, take a look at the `lower_swift` filter, which we use for generating the code that takes a Swift object and lowers it into something to pass across the FFI:

* https://github.com/mozilla/uniffi-rs/blob/d44b96c9de105a6fafc4ff1e1f66e3248631a129/uniffi_bindgen/src/bindings/swift/gen_swift.rs#L226

For almost all values `v` of type `T` this just generates `v.lower()`, so when we want to lower an `Optional<T>` we generate: `v.lower()` and let the Swift compiler figure out how to dispatch it to the right type logic. With the change here that will no longer work, because there is no `lower()` method on `Optional<T>`; you'll instead have to generate `v.lowerUint64` or similar, along the lines of what the case for `Type::Duration` is doing.

Having typed that out...perhaps trying to generate the same generic code as before, but only generate it once, would be a smaller change overall.

rfk (commented):
Thanks @lougeniaC64, this seems headed in the right direction! There's some complexity around the handling of `Optional`/`Sequence`/`Map` types which I didn't anticipate being quite so different here vs in Kotlin, some notes below.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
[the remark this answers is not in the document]
lougeniaC64: Yes, this is a very nice explanation of the crossroad at which I found myself. I saw how verbose the generated code would be if I continued down that path and was growing doubtful that my changes would be beneficial. Since then I took the scenic route to opting for adding higher-level methods to `ComponentInterface`. I did so for the `Optional` type and if things look good I'll wrap this up by making the same changes for `Array` and `Dictionary`.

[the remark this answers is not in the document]
rfk: Heh, I think the scenic route is very much in the spirit of the meta-goal of this PR so even if we don't like it, it will have been worthwhile :-)

I have a good feeling about it at a high level though, will dig into the details here shortly.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -241,6 +241,38 @@ impl<'ci> ComponentInterface {
         }
     }
 
+    pub fn contains_optional_references(&self) -> bool {
+        self.iter_types().iter().any(|t| self.type_contains_optional_references(&t))
+    }
+
+    fn type_contains_optional_references(&self, type_: &Type) -> bool {
```
rfk, line 248: I don't think we need this type-specific version, since IIUC we only want to ask the "contains optional types?" question about the interface as a whole. We already know that `self.iter_types()` will yield every single type in the interface, so we don't need to recurse into each type, we can just do a flat early-return match like:

```
self.iter_types().iter().any(|t| matches!(t, Type::Optional(_))
```

(By the way, I have a PR that makes these various `iter_blah()` methods return actual iterators rather than `Vec`, which would remove the need for the `.iter()` here; it's waiting on a new release of Askama)

rfk (commented):
I think you found the best path forward here, full steam ahead on the `contains_blah_types()` approach 👍🏻

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -241,6 +241,22 @@ impl<'ci> ComponentInterface {
         }
     }
 
+    pub fn contains_optional_types(&self) -> bool {
```
lougeniaC64, line 244: I really don't like the redundancy here and wish I could pass the sought type into a generic `countains _type` function but I have learned that "destructuring assignments are unstable" and explicitly covering all of the enum variants for these types would be silly. I'm not sure if this is macro territory but if you have any suggestion for how to refactor this I am open.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -456,3 +483,18 @@ extension Dictionary: ViaFfiUsingByteBuffer, ViaFfi, Serializable where Key == S
...
+{# Helpers for Enum types are defined inline with the Enum class #}
+
+{% when Type::Record with (record_name) -%}
+{# Helpers for Record types are defined inline with the Record class #}
+
+{% when Type::Object with (object_name) -%}
+{# Helpers for Object types are defined inline with the Record class #}
+
+{% else %}
+{# This type cannot be lifted, lowered or serialized (yet) #}
```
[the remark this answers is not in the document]
rfk: nit: the `This type cannot be lifted, lowered or serialized (yet)` comment here is no longer valid, this should be something like `The methods for lifting/lowering/serializing this type are implemented inline with the type itself`.

uniffi_bindgen/src/interface/mod.rs
rfk: Yeah, I see what you mean. I don't think a macro would help maintainability here so I'm happy to just eat the redundancy.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -241,6 +241,22 @@ impl<'ci> ComponentInterface {
...
+
+    pub fn contains_sequence_types(&self) -> bool {
+        self.iter_types()
+            .iter()
+            .any(|t| matches!(t, Type::Sequence(_)))
+    }
+
+    pub fn contains_map_types(&self) -> bool {
+        self.iter_types().iter().any(|t| matches!(t, Type::Map(_)))
+    }
```
rfk, line 258: For completeness, you could consider adding some basic unittests for these methods in the suite at the bottom of this file. I don't expect them to be particularly high-value tests though, the code here is about as "obviously correct" as code ever gets...

rfk (approved):
Looks great, thank you for diving in!

PR #983, opened 2021-07-22
Gh 460: Support rich structured Enums for reporting errors

bendk:


uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h
```diff
@@ -40,11 +40,10 @@ typedef struct ForeignBytes
...
-    int32_t code;
-    char *_Nullable message;
-} NativeRustError;
+typedef struct RustCallStatus {
+    int8_t code;
+    RustBuffer errorBuf;
+} RustCallStatus;
 
 // ⚠️ Attention: If you change this #else block (ending in `#endif // def UNIFFI_SHARED_H`) you *must* ⚠️
 // ⚠️ increment the version in all instance of UNIFFI_SHARED_HEADER_V1 in this file.                   ⚠️
```
bendk, line 49: I wasn't sure what this meant.  Should I change the name to `UNIFFI_SHARED_HEADER_V2`?

uniffi_bindgen/src/interface/error.rs
```diff
@@ -19,69 +19,146 @@
...
+        self.enum_.variants()
+    }
+
+    pub fn is_flat(&self) -> bool {
+        self.enum_.is_flat()
+    }
+
+    // For compatibility with the Enum interface
+    pub fn contains_object_references(&self, _: &ComponentInterface) -> bool {
+        false
```
bendk, line 128: This one confused me since it seemed like only object reference would be an Interface, which doesn't seem possible to nest in an Error or an Enum.  What's the correct behavior here?

uniffi_bindgen/src/bindings/gecko_js/gen_gecko_js.rs
```diff
@@ -103,13 +103,13 @@ impl<'config, 'ci> Context<'config, 'ci> {
...
-    /// A `RustError` is a Plain Old Data struct that holds an error code and
-    /// a message string. See the docs for `ffi_rustbuffer_type` about why this
-    /// type name must be unique for each component.
-    pub fn ffi_rusterror_type(&self) -> String {
-        format!("{}_RustError", self.ci.ffi_namespace())
+    /// A `RustCallStatus` is a Plain Old Data struct that holds an call status code and
+    /// `RustBuffer`. See the docs for `ffi_rustbuffer_type` about why this type name must be
+    /// unique for each component.
+    pub fn ffi_rustcallstatus_type(&self) -> String {
+        format!("{}_RustCallStatus", self.ci.ffi_namespace())
```
bendk, line 112: I was really just guessing on the gecko JS code.  Is there a good way to test the JS bindings?

uniffi_bindgen/src/scaffolding/templates/ErrorTemplate.rs
```diff
@@ -1,20 +1,48 @@
...
+        };
     }
+
+    {% if e.is_flat() %}
+    // If a variant doesn't have fields defined in the UDL, it's currently still possible that
+    // the Rust enum has fields and they're just not listed.  Let's just punt on implemented
+    // try_read() in that case, which is no issue since passing back Errors into the rust code
+    // isn't supported.
+    fn try_read(_buf: &mut &[u8]) -> uniffi::deps::anyhow::Result<Self> {
+        panic!("try_read not supported for fieldless errors");
```
bendk, line 28: This was kind of unfortunate, but I think it shouldn't cause issues unless someone wants to pass an error from the consumer code into the rust code.  Going forward, maybe we could deprecate error enums with associated data that don't define that data in the UDL file.

CHANGELOG.md
```diff
@@ -12,17 +12,19 @@
 
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.12.0...HEAD).
 
+### ⚠️ Breaking Changes ⚠️
```
rfk, line 15: 👍🏻 thanks for this

uniffi/src/ffi/rustcalls.rs
```diff
@@ -0,0 +1,251 @@
...
+/// Represents the success/error of a rust call
+///
+/// ## Usage
+///
+/// - The consumer code creates a `RustCallStatus` with an empty `RustBuffer` and `CALL_SUCCESS`
+///   (0) as the status code
+/// - A pointer to this object is passed to the rust FFI function.  This is an
+///   "out parameter" which will be updated with any error that occurred during the function's
+///   execution.
+/// - After the call, if code is `CALL_ERROR` then `error_buf` will be updated to contain
```
rfk, line 28: nit: backticks around `code` for the docstring here

uniffi/src/ffi/rustcalls.rs
```diff
@@ -0,0 +1,251 @@
...
+/// ```c,no_run
+/// struct RustCallStatus {
+///     int8_t code;
+///     RustBuffer error_buf;
+/// };
+/// ```
+///
+/// #### The `code` field.
+///
+///  - `CALL_SUCESS` (0) for successful calls
```
rfk, line 45: sp: `SUCESS` -> `SUCCESS`

uniffi/src/ffi/rustcalls.rs
```diff
@@ -0,0 +1,251 @@
...
+///   responsible for freeing this `RustBuffer`.
+#[repr(C)]
+pub struct RustCallStatus {
+    pub code: i8,
+    // code is signed because unsigned types are experimental in Kotlin
+    pub error_buf: MaybeUninit<RustBuffer>,
+    // error_buf is MaybeUninit to avoid dropping the value that the consumer code sends in:
+    //   - Consumers should send in a zeroed out RustBuffer.  In this case dropping is a no-op and
+    //     avoiding the drop is a small optimization.
+    //   - If consumers pass in invalid data, then we should avoid trying to drop in.  In
```
rfk, line 61: sp: "drop in" -> "drop it"

uniffi/src/ffi/rustcalls.rs
```diff
@@ -0,0 +1,251 @@
...
+    F: panic::UnwindSafe + FnOnce() -> Result<R, RustBuffer>,
+    R: IntoFfi,
+{
+    let result = panic::catch_unwind(|| {
+        // We should call:
+        //
+        // init_panic_handling_once();
+        //
+        // This is called in ffi-support::call_with_result_impl().  The current plan is to make it
+        // `pub` in ffi-support and call it from here.
```
rfk, line 88: When we chat later this week, it's work through getting that release cut so we can do it here.

uniffi/src/ffi/rustcalls.rs
```diff
@@ -0,0 +1,251 @@
...
+const CALL_ERROR: i8 = 1;
+const CALL_PANIC: i8 = 2;
+
+// A trait for errors that can be thrown to the FFI code
+//
+// This gets implemented in uniffi_bindgen/src/scaffolding/templates/ErrorTemplate.rs
+pub trait FfiError: RustBufferViaFfi {}
+
+// Generalized rust call handling function
+// callback should map errors to a RustBuffer that contains them
```
rfk, line 76: It took me a little while to understand what you meant about the callback here. It may be worth a few extra words to explain, along the lines of: "The callback is responsible for serializing errors into a `RustBuffer` so they can be returned over the FFI".

(FWIW, I think the name `callback` also didn't help my understanding here, it sounded like a helper that would be called *in the case of error* rather than being the actual wrapped function that we call).

uniffi/src/ffi/rustcalls.rs
```diff
@@ -0,0 +1,251 @@
...
+    });
+    match result {
+        // Happy path.  Note: no need to update out_status in this case because the calling code
+        // initializes it to CALL_SUCCESS
+        Ok(Ok(v)) => v,
+        // Callback returned an Err.
+        Ok(Err(buf)) => {
+            out_status.code = CALL_ERROR;
+            unsafe {
+                out_status.error_buf.as_mut_ptr().write(buf);
```
rfk, line 99: It would be useful to briefly summarize the safety invariants here. IIUC it's specifically the `write` that's unsafe, and we're upholding the invariant that it's a valid and aligned pointer, by assumption that the passed `out_status` pointer is valid.

uniffi/src/ffi/rustcalls.rs
```diff
@@ -0,0 +1,251 @@
...
+                };
+                log::error!("Caught a panic calling rust code: {:?}", message);
+                String::lower(message)
+            }));
+            if let Ok(buf) = message_result {
+                unsafe {
+                    out_status.error_buf.as_mut_ptr().write(buf);
+                }
+            }
+            // Ignore the error case.  We've done all that we can at this point
```
rfk, line 125: In practice, I guess this will cause the foreign-language bindings to just read an empty string from the (presumably) empty buffer that it passed in. May be worth noting that explicitly here.

uniffi_bindgen/src/bindings/gecko_js/gen_gecko_js.rs
rfk: There is not; TBH I think at this point we should just remove them because they're not actually used for anything in production.

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -1,149 +1,102 @@
...
+    } else if (status.isError()) {
+        throw errorHandler.lift(status.error_buf)
+    } else if (status.isPanic()) {
+        // when the rust code sees a panic, it tries to construct a rustbuffer
+        // with the message.  but if that code panics, then it just sends back
+        // an empty buffer.
+        if (status.error_buf.len > 0) {
+            throw InternalException(String.lift(status.error_buf))
+        } else {
+            throw InternalException("Rust panic")
```
rfk, line 84: Do we need to free `error_buf` here?

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -1,49 +1,91 @@
...
+            {%- if variant.has_fields() %}
+            {%- for field in variant.fields() %}
+            self.{{ field.name()|var_name_py }} = {{ field.name()|var_name_py }}
+            {%- endfor %}
+            {% else %}
+            pass
+            {%- endif %}
+
+        def __str__(self):
+            return "{{ e.name()|class_name_py }}.{{ variant.name()|class_name_py }}({% for field in variant.fields() %}{{ field.name() }}={}{% if !loop.last %},{% endif %}{% endfor %})".format({% for field in variant.fields() %}self.{{ field.name() }}{% if !loop.last %},{% endif %}{% endfor %})
```
rfk, line 41: yikes, but also, probably not that much value in trying to split this over multiple lines..,

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -1,49 +1,91 @@
...
+        else:
+            raise consume_buffer_into_error(error_class, call_status.error_buf)
+    elif call_status.code == RustCallStatus.CALL_PANIC:
+        # When the rust code sees a panic, it tries to construct a RustBuffer
+        # with the message.  But if that code panics, then it just sends back
+        # an empty buffer.
+        if call_status.error_buf.len > 0:
+            msg = call_status.error_buf.consumeIntoString()
+        else:
+            msg = "Unkown rust panic"
```
rfk, line 87: sp: "Unkown" -> "Unknown"

uniffi_bindgen/src/bindings/python/templates/RustBufferStream.py
```diff
@@ -163,6 +163,35 @@ def read{{ canonical_type_name }}(self):
...
+        try:
+            read_variant_method = getattr(self, 'read{{canonical_type_name}}{}'.format(variant))
+        except AttributeError:
+            raise InternalError("Unexpected variant value for error {{ canonical_type_name }} ({})".format(variant))
+        return read_variant_method()
+
+    # Read methods for each individual variants
+    {%- for variant in e.variants() %}
+
+    def read{{ canonical_type_name }}{{ loop.index }}(self):
```
rfk, line 183: I remain lightly concerned about name conflicts here in pathological cases, e.g. if I had one error named `MyError` and another error named `MyError1`. What do you think about putting the variant helpers into hidden methods with a different prefix like `_readVariant{{ loop.index }}Of{{ canonical_type_name }}` or similar?

uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h
rfk: Yes, I think we should - otherwise we might end up using the declarations from some other component where `NativeRustError` is defined an `RustCallStatus` is not.

uniffi_bindgen/src/interface/attributes.rs
```diff
@@ -637,15 +642,4 @@ mod test {
             "conflicting attributes on interface definition"
         );
     }
-
-    #[test]
-    fn test_other_attributes_not_supported_for_interfaces() {
-        let (_, node) =
-            weedle::attribute::ExtendedAttributeList::parse("[Threadsafe, Error]").unwrap();
```
rfk, line 644: Could we keep this test, but use some other attribute like `ByRef` or similar to replace `Error`?

uniffi_bindgen/src/interface/error.rs
rfk: I think the right thing to do either way is to delegate to `self.enum_.contains_object_references`, and either let it continue to do its thing or update it to always return `false`. Off the top of my head I don't see anything that would *prevent* declaring an enum containing an interface type, but I haven't thought about it very hard.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -119,6 +119,8 @@ impl<'ci> ComponentInterface {
             println!("{}", remaining);
             bail!("parse error");
         }
+        // Unconditionally add the String type, which is used by the error handling
```
rfk, line 122: Specifically it's used by *panic* handling, so we need it even if the interface doesn't define any user-level errors.

uniffi_bindgen/src/scaffolding/templates/ErrorTemplate.rs
```diff
@@ -1,20 +1,48 @@
 {#
-    // For each error declared in the UDL, using the [Error] attribute, we assume the caller has provided a corresponding
-    // rust Error enum with the same name. We provide the traits for sending it across the FFI, which will fail to
-    // compile if the provided enum has a different shape to the one declared in the UDL. 
-    // Here we define the neccessary converstion to allow the error to propegate through the FFI as an error.
+// For each enum declared in the UDL, we assume the caller has provided a corresponding
```
rfk, line 2: IIUC, I think you still mean "For each error" rather than "For each enum", since the close link between errors and enums is not necessarily that obvious in the UDL.

uniffi_bindgen/src/scaffolding/templates/ErrorTemplate.rs
```diff
@@ -1,20 +1,48 @@
...
+                {%- endfor %}
+            },
             {%- endfor %}
-        }
+        };
     }
+
+    {% if e.is_flat() %}
+    // If a variant doesn't have fields defined in the UDL, it's currently still possible that
+    // the Rust enum has fields and they're just not listed.  Let's just punt on implemented
```
rfk, line 24: sp: "implemented" -> "implementing"?

uniffi_bindgen/src/scaffolding/templates/ErrorTemplate.rs
rfk: Agreed, this seems safe to punt on, and deprecating it seems like a good path forward.

CHANGELOG.md
```diff
@@ -12,17 +12,19 @@
...
 
 ### What's Changed
 
 - Both python and ruby backends now handle U16 correctly.
-- Python timestamps will now be in UTC and timezone-aware rather than naive.
-- Replaced `lower_into_buffer()` and `try_lift_from_buffer()` with the
-  `RustBufferViaFfi` trait.  If you use those functions in your custom ViaFfi
-  implementation then you'll need to update the code.  Check out the `Option<>`
-  implementation in uniffi/src/lib.rs for an example.
+- Errors with nested fields are now supported.
```
rfk, line 27: I'm not sure if "nested fields" communicates the change effectively to users who haven't been keeping an eye on this. What about something like: "Error variants can now contain named fields, similar to Enum variants"?

rfk (approved):
@bendk this is great! A bunch of small comments throughout, but nothing worth blocking or worth another round of review.

uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h
rfk: (The failure behaviour if we *don't* do this will be safe, e.g. the user will get an error like "`RustCallStatus` not defined" or "`NativeRustError` not defined". But still, we should provide a better error in this case, since they may not realize it's due to mixing versions of UniFFI).

uniffi/src/ffi/rustcalls.rs
bendk: I think the main reason it's unsafe is it's `MaybeUninit`, so we won't be dropping any previously initialized data.  I'll add few comments about that one, it's definitely not clear from the code.

I didn't think about alignment issues, I guess we just have to assume that all the bindings code gives us a properly aligned struct.

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
bendk: I think we can avoid freeing since it's the same `RustBuffer` that was initially passed in with `data=null`.

uniffi_bindgen/src/scaffolding/templates/ErrorTemplate.rs
bendk: Your comments reminded me that we only support Error enums with named fields, which makes me hesitant to deprecate this and force consumers to change how they structure their errors.  I'm thinking we should just not worry about this corner case for now.

rfk: Ah yes, fair enough. I agree it's seems very unlikely to cause us problems in practice, happy to just punt on it for now.

bendk:
@rfk I think this one is finally ready, but can you just double check that I did the right thing when updating the ffi-support version?

rfk:
> can you just double check that I did the right thing when updating the ffi-support version?

LGTM 👍🏻

mhammond:
Sorry I'm late to the party here and didn't help review, but this is great work and I'm really excited to see it land!

PR #984, opened 2021-07-26
Restructure Swift bindings for easier megazording.

rfk:
Fixes https://mozilla-hub.atlassian.net/browse/SYNC-2679.
Needs a lot more docs, but I wanted to see how it'd work out
in practice. I think it's going to work!


uniffi_bindgen/src/bindings/swift/gen_swift.rs
```diff
@@ -18,36 +16,64 @@ use crate::MergeWith;
...
+
+    /// The name of the `.modulemap` file for the lower-level C module with FFI declarations.
     pub fn modulemap_filename(&self) -> String {
-        format!("{}.modulemap", self.module_name())
+        format!("{}.modulemap", self.ffi_module_name())
     }
+
+    /// The name of the `.h` file for the lower-level C module with FFI declarations.
     pub fn header_filename(&self) -> String {
-        format!("{}-Bridging-Header.h", self.module_name())
```
rfk, line 35: I've disliked this `-Bridging-Header.h` suffix for a long time. IIUC it's like this because that's the default that Xcode generates when you ask it to make a bridging header for you, but IMHO it's more confusing than helpful in our case, leading the consumer to ask questions like "what is a bridging header?" and "do I need to do something special with bridging headers?" when really they just need to treat it like an ordinary header.

uniffi_bindgen/src/bindings/mod.rs
```diff
@@ -112,27 +112,24 @@ pub fn write_bindings<P>(
     out_dir: P,
     language: TargetLanguage,
     try_format_code: bool,
-    is_testing: bool,
```
rfk, line 115: Pretty happy to get rid of this hackery :-)

rfk:
The broad idea here is:

* For users building a standalone module, the defaults will continue to just work for them. They'll get a new `.modulemap` file that previously wasn't generated by default, but can disable it using `generate_module_map = false` in config.
* For the existing application-services Xcode build, we want to compile multiple modules and their FFI layers into a single artifact, so we would:
    * Disable generation of the modulemap file, we don't need it.
    * Either update the names of our `.h` files in the Xcode config to use the new `BlahFFI` name, or use the `header_filename` config option to continue calling them `blah-Bridging-Header.h`.
    * Build in Xcode just like before.
* For the new XCFramework build being proposed for application-services, we want to compile all the Rust code for multiple modules together into a single lib, so we would:
    * Disable generation of the modulemap file, because we provide our own hand-written one.
    * Either update the names of our `.h` files in the build script to use the new `BlahFFI` name, or use the `header_filename` config option to continue calling them `blah-Bridging-Header.h`.
    * Use config to set the name of the ffi module to `MozillaRustComponents` for all modules, so they all load the same underlying framework when looking for the Rust code.
    * Build all the Rust code together into a single `.xcframework`, including all the generated header files along with a custom modulemap and umbrella header.

I'll aim to write some docs on how each of those three options work as part of this PR.

rfk:
OK, I think this is ready to go! Since this is in service of the XCFramework stuff, @tarikeshaq pinging you for review.

CHANGELOG.md
```diff
@@ -13,6 +13,17 @@
...
 
 ### ⚠️ Breaking Changes ⚠️
+- The Swift bindings now explicitly generate two separate Swift modules, one for
+  the high-level Swift code and one for the low-level C FFI. This change is intended
+  to simplify distribution of the bindings via Swift packages, but brings with it
+  some changes to the generated file layout.
+  - For an interface namespace "example", we now generate:
+    - A bridged C module named "exampleFFI" containing the low-level C FFI,
+      consisting of an `exampleFFI.h` file and matching `exampleFFI.modulemap`
+      file. The name can be customized using the `ffi_module_name` config option.
```
rfk, line 23: FWIW, I'm open to the suggestion that we should leave the default names of these files unchanged in order to avoid disruption to consumers, I just think the new names may help form a more clear mental model of what's happening.

tarikeshaq (approved):
A couple of super nits, but otherwise looks good to me!!

a quick clarification for myself 
> Disable generation of the modulemap file, because we provide our own hand-written one

we'll still have exactly one module map for all the modules after https://github.com/mozilla/application-services/pull/4304 correct? This is why we don't want individual ones for each component, even though each component is its own module

rfk:
> we'll still have exactly one module map for all the modules after mozilla/application-services#4304 correct?
> This is why we don't want individual ones for each component, even though each component is its own module

Correct 👍🏻

uniffi_bindgen/src/bindings/swift/templates/wrapper.swift
```diff
@@ -2,9 +2,13 @@
 // Trust me, you don't want to mess with it!
 
 import Foundation
-{% if is_testing -%}
-import {{ config.module_name() }}
-{% endif -%}
+
+// Depending on the consumer's build setup, the low-level FFI code
```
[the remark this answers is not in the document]
rfk: Yep, it's deliberately in the generated bindings to help our future selves with debugging the generated code, when we've forgotten the nitty-gritty details of the import setup here.

docs/manual/src/swift/overview.md
```diff
@@ -0,0 +1,45 @@
...
+* Errors are represented as Swift enums that conform to the `Error` protocol.
+* Function calls that have an associated error type are marked with `throws`,
+  and hence must be called using one of Swift's `try` syntax variants.
+* Failing assertions, Rust panics, and other unexpected errors in the generated code
+  are translated into a special `UniffiInternalError` enum that conforms to the
+  `Error` protocol.
+    * If this happens inside a throwing Swift function, it can be caught and handled
+      like any other error (but do so at your own risk, because it indicates that
+      something has gone seriously wrong).
+    * If this happens inside a non-throwing Swift function, it will be converted
```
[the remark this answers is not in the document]
rfk: Actually this is out-of-date as of https://github.com/mozilla/uniffi-rs/pull/989, I will re-word to clarify.

Issue #985, opened 2021-07-26
uniffi_foreign_language_testcase_test_chronological_kts is intermittently failing for me


bendk:
When I run the full test suite, this test fails on a somewhat regular basis for me.  The tests usually work when I re-run them.

┆Issue is synchronized with this [Jiraserver Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-72)
┆Issue Number: UNIFFI-72
┆Sprint End Date: 2021-08-06

bendk:
I think this is similar to the issue with timestamp resolution that @rfk fixed before.  I added some debugging printouts here's a case that failed:
```
kotlinBefore: 2021-07-28T19:15:57.621Z
rustNow:      2021-07-28T19:15:57.622096929Z
kotlinAfter:  2021-07-28T19:15:57.622Z
```

PR #986, opened 2021-07-28
Another fix for the Kotlin chrono tests (#985)

bendk:


rfk (approved):
👍🏻 makes sense; I don't like sleeping in unit tests but sometimes it really is the right thing on balance :-)

Issue #987, opened 2021-07-30
iOS sample app fails to build with error: "Invalid redeclaration of 'UniffiInternalError'"


rfk:
I think this is some unexpected fallout from https://github.com/mozilla/uniffi-rs/pull/983, specifically the switch from `UniffiInternalError` being `fileprivate` to `public`. /cc @bendk.

STR:

* `cargo build -p uniffi_bindgen` to make sure you have an up-to-date `uniffi-bindgen` executable.
* In Xcode, open `./examples/app/ios/IOSApp.xcodeproj`.
* In Xcode, select "Product" -> "Test"

I get a build failure about an invalid redeclaration of the `UniffiInternalError` struct.

IIUC what's happening here is:

* The same app depends on two other example crates, `arithmetic` and `todolist`.
* The generated `arithmetic.swift` bindings contain a `public UniffiInternalError` and so do the generated `todolist.swift` bindings.
* Because the app builds all the generated Swift code together as a single module, Swift sees these as conflicting declarations and throws an error.

We've had this problem in the past with other shared definitions like the `RustBuffer` struct, but since those were at the C level, we were able to work around them using [standard `#ifdef`-style hacker](https://github.com/mozilla/uniffi-rs/blob/28e13d80f61441b601f3613b5bc93fadb30928dd/uniffi_bindgen/src/bindings/swift/templates/Template-Bridging-Header.h#L12). But this definition is in Swift, and while Swift does have some limited `#ifdef`-like functionality for conditional compilation, I don't believe it supports the same pattern as the C header does.

Note that our application-services build handles Swift files in a similar way, so I expect it would have a similar problem.

I'm not entirely sure how to fix this. We may be able to revert `UniffiInternalError` to being `fileprivate` for now, but I don't recall the motivation for making it public. @bendk what do you think?

(The longer-term solution here is better hygiene between Swift modules, which I'm hoping we'll get out of the Swift Package work. But we need to continue supporting these all-in-one builds in the meantime).

┆Issue is synchronized with this [Jiraserver Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-73)
┆Issue Number: UNIFFI-73

bendk:
The main reason was just to be able to test it.  I just made a PR that avoids that and makes `UniffiInternalError` `fileprivate` again.

rfk:
Fixed by https://github.com/mozilla/uniffi-rs/pull/989

PR #989, opened 2021-07-30
Make UniffiInternalError fileprivate again (#987)

bendk:


tarikeshaq (approved):
LGTM!

This means that swift consumers can never catch our internal errors right? which is exactly how it used to be I guess...  I think it's OK since if a function is non-throwable it'll fatal error anyways. In A-C they're catching the InternalException from the tabs component, is our stance that in swift we don't want to allow consumers to do that? (FWIW I'm okay with this stance, for the same reason panics usually go uncaught if they don't cross an FFI)

bendk:
> This means that swift consumers can never catch our internal errors right? which is exactly how it used to be I guess... I think it's OK since if a function is non-throwable it'll fatal error anyways. In A-C they're catching the InternalException from the tabs component, is our stance that in swift we don't want to allow consumers to do that? (FWIW I'm okay with this stance, for the same reason panics usually go uncaught if they don't cross an FFI)

I'm pretty sure this is correct and I also feel okay about this stance.

PR #990, opened 2021-08-02
Remove the unused Gecko-JS bindings.

rfk:
A lot of excellent work went into making working Gecko-JS bindings,
but the fact is that we're not using them and we expect to have to
make non-trivial changes to the way they're architected before we
can use this in Desktop Firefox for real.

Given that they're becoming a maintenance burding during unrelated
refactors of UniFFI, I think we're better off removing them until
we take another concerted run at integrating into Desktop Firefox.
We can get them out of the git history if and when we need them.


rfk:
Dual to this PR, I filed a new issue https://github.com/mozilla/uniffi-rs/issues/991 to consolidate the current context on Gecko bindings in a single place, so that when we're ready to take another run at it, there's a single obvious starting point.

mhammond (approved):

PR #992, opened 2021-08-02
Document some high-level design principles of the tool.

rfk:
**[Rendered View](https://github.com/mozilla/uniffi-rs/blob/c6745aa0f4a9e80f25901a4874ef465b07fe1a23/docs/manual/src/internals/design_principles.md)**.

These are some of the things that I've been implicitly considering when building UniFFI, and I figured it would probably be useful to have them more explicit. It's also a great opportunity for us to decide that we don't agree with some of the things I've said here, and work on a replacement!

### DIFF
--- a/docs/manual/src/internals/design_principles.md
+++ b/docs/manual/src/internals/design_principles.md
@@ -1,0 +1,91 @@
+# Design Principles
+
+These are some high-level points to consider when making changes to UniFFI (or when wondering why past changes were made in a particular way).
+
+
+### Prioritize Mozilla's short-term needs
+
+The initial consumers of this tool are teams working on features for Mozilla's mobile browsers.
+While we try to make the tool generally useful, we'll invest first in things that are the most valuable
+to those teams, which are reflected in the points below.
+
+
+### Safety First
+
+The generated bindings need to be safe by default. It should be impossible for foreign-language code
+to trigger undefined behaviour in Rust by calling the public API of the generated bindings, even if it
+is called in egregiously wrong or malicious ways. We will accept reduced performance in the interests
+of ensuring this safety.
+
+(The meaning of "impossible" and "public API" will of course depend on the target language. For example,
+code in Python might mutate internal attributes of an object that are marked as private with a leading
+underscore, and there's not much we can do to guard against that.)
+
+Where possible, we use Rust's typesystem to encode safety guarantees. If that's not possible then the
+generated Rust code may use `unsafe` and assume that the generated foreign-language code will uphold
+safety guarantees at runtime.
+
+**Example:** We insist that all object instances exposed to foreign-language code be `Sync` and `Send`,
+so that they're safe to access regardless of the threading model of the calling code. We do not allow
+thread-safety guarantees to be deferred to assumptions about how the code is called.
+
+**Example:** We do not allow returning any borrowed data from function calls, because we can't make
+any guarantees about when or how the foreign-language could would access it.
+
+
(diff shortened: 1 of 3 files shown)


docs/manual/src/internals/design_principles.md
```diff
@@ -0,0 +1,95 @@
...
+
+**Example:** Initial versions of the tool used opaque integer handles and explicit mutexes to manage
+object references, favouring simplicity (in the "we're confident this works as intended" sense) over
+performance. As we got more experience and confidence with the approach tool we replaced handles with
+raw `Arc` pointers, which both simplified the code and removed some runtime overheads.
+
+**Violation:** Initial versions of the tool pass structured data over the FFI by serializing it to
+a byte buffer, favouring ease of implementation and understanding over performance. However, we have
+not done any work to measure the performace impact or iterate towards something with lower overhead
+(such as using `repr(C)` structs).
```
rfk, line 53: I think it's also interesting to understand where we have parts of the tool that don't fit our aspirational goals and to call them out explicitly. For example, I think we made the right call initially with serializing data to cross the FFI, but as time goes on and we don't do the work to understand the perf/complexity tradeoff there, it's probably becoming wronger and wronger.

docs/manual/src/internals/design_principles.md
```diff
@@ -0,0 +1,95 @@
...
+
+However, we know that performance can degrade through thousands of tiny cuts, so we'll keep iterating
+towards the winning combination of "simple *and* fast" over time.
+
+**Example:** Initial versions of the tool used opaque integer handles and explicit mutexes to manage
+object references, favouring simplicity (in the "we're confident this works as intended" sense) over
+performance. As we got more experience and confidence with the approach tool we replaced handles with
+raw `Arc` pointers, which both simplified the code and removed some runtime overheads.
+
+**Violation:** Initial versions of the tool pass structured data over the FFI by serializing it to
```
mhammond, line 50: The use on `Initial versions` here makes it sound like "in the past", where this is still true today - whereas in the example above, it truly is now in the past. I think it's probably OK to say something like "at time of writing ..." to clear that up.

docs/manual/src/internals/design_principles.md
```diff
@@ -0,0 +1,95 @@
...
+The people using the tool need to be empowered to debug, maintain and develop it.
+
+If you're using UniFFI-generated bindings and something doesn't work quite right, it should be possible
+for you to dig in to the generated foreign-language code, follow it through to the underlying Rust code,
+and work out what's going wrong without being an expert in Rust or UniFFI.
+
+**Example:** We try to include comments in the generated code to help guide users who may be reading
+through it to debug some issue.
+
+**Example:** We use a separate interface description file rather than annotating the Rust code directly
```
mhammond, line 86: I don't really get this point, not how the UDL helps "trace the flow of data" and I think the UDL actually works against that - understanding weedle to maintain the tool is a liability rather than a feature. The UDL is also a liability in other ways, such as limiting ways we can express concepts uniffi should have.  This point would make sense if there was a section titled something like "move fast, make pragmatic decisions" or similar, but there's not (and I don't really think it needs one).

mhammond (approved):
This is great, thanks! I think one of the points should be deleted, but I'll leave that to you :)

docs/manual/src/internals/design_principles.md
rfk: Thanks, I will just delete this one.

PR #993, opened 2021-08-02
Unit of Code: Refactor kotlin backend to use the language oracle pattern

jhugman:
Refactor the Kotlin backend so that each type is in its own file.

Every type that is declared has its own `CodeDeclaration`. Every type that is used has its own `CodeType`. The code for each type (the `CodeDeclaration` and `CodeType`) lives in its own file.

`CodeType`, is a trait which knows how to render foreign language code:

 * its helper code
 * imports
 * expressions to call into that code. (lift, lower, etc).

These are used for referring to types in the member definitions.

`CodeDeclaration` is a trait which knows how to render foreign language code:

 * its definition
 * imports
 * initialization code

These are used for member types defined in the UDL.

For each member type (e.g. an `interface` or `enum` defined in the UDL) there will be: a `CodeType` (so other things can refer to it) and a `CodeDeclaration`. Optionally, there may be a `CodeDecalaration` setting up the runtime, (e.g. `object.rs` and `callback_interface.rs` both add runtime types not needed unless it is needed.

Types which are not defined by the UDL, but travel across the FFI (primitives, the mythical external types) etc, need only a `CodeType`.

Types which are defined by the UDL, but do not travel across the FFI (e.g. @travis79 's delegate types) need a `CodeDeclaration`, but not a `CodeType`.

The `CodeOracle` knows how to look up a given `CodeType` from a `Type`.

Thus, every type knows how to render itself, and call into other types (in order to render itself).

This PR very deliberately edits the templates as little as possible, aside from moving them around, and making them work with the new machinery. There are some "witness marks" left in to facilitate a larger refactoring of these templates:

 * `CodeDeclaration`s accept an oracle, though this is almost never needed. Perhaps `CodeDeclaration`s should own their own oracles?
 * a new oracle is made by each call to a filter. Mechanical changes to the filter call sites are needed so the oracle is passed from the `CodeDeclaration` into the templates, through the filters and then in to the code types themselves.

These are left in as a transitionary point: will we keep Askama? This is a large enough decision that it should not be done in this already huge PR.

There is more to do, but enough progress is made here that **we should probably land this** and then start iterating. 

Some additional work around `Type` is needed which touches other backends.

 * let `CodeType`s decide which `FFIType` should be used.
 * implement an `ExternalType` or a `TransformType`
 * Move the FFI functions into the appropriate `CodeDeclaration`s. By adding a function `Vec<FFIFunction>` to `CodeDeclaration` (née `MemberDeclaration`) `NamespaceLibraryTemplate.kt` could iterate through in the same way `wrapper.kt` does with imports.
 * [X] Refactor conditionally run once only code (e.g. only if there is at least one object or one callback interface) in `Helpers.kt`.
 * There is a big refactor to be had around the templates, so I'm awaiting how this lands before doing anything with the filters.
 * Identify what is common to all backends…
 * …and how to specialize `CodeType` to allow other tasks needed by other backends (e.g. python's `coerce`).
 * Find the minimal sized code to make this copy/paste/adaptable for the other backends.
 * Move documentation around to as to make it easier to follow, or duplicate.

A [dreamcode version of this is in this gist](https://gist.github.com/jhugman/9d7a969ca345420797f5f1819c8a77e1).


jhugman:
@rfk I'm in a fairly good place if you want to start looking at templates. I've left some notes in `gen_kotlin.rs`.

uniffi_bindgen/src/bindings/backend.rs
```diff
@@ -0,0 +1,80 @@
...
+    ///
+    /// This replaces "Error" at the end of the name with "Exception".  Rust code typically uses
+    /// "Error" for any type of error but in the Java world, "Error" means a non-recoverable error
+    /// and is distinguished from an "Exception".
+    fn exception_name(&self, nm: &dyn fmt::Display) -> String;
+
+    fn ffi_type_label(&self, ffi_type: &FFIType) -> String;
+}
+
+pub trait CodeType {
```
rfk, line 39: From its presence in a `backend.rs` I infer that this trait may be meant to be shared between the different language backends. That seems valuable, but also it's not immediately obvious to me that it will work out well - for example the Python backend has a "coerce" operation that isn't present in the Kotlin backend. Will we end up with something here that's the union of all possible backend features? Will we end up with a "parent" trait that various backends can specialize?

(That's not intended as stop-energy on this sharing approach, just reacting out loud on initial read-through)

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -63,236 +71,179 @@ pub struct KotlinWrapper<'a> {
 }
 impl<'a> KotlinWrapper<'a> {
     pub fn new(config: Config, ci: &'a ComponentInterface) -> Self {
+        // I _think_ I want to make this the holder of the LanguageOracle, which holds the CI.
+        // I haven't quite worked out if each type's CodeType should be one type (with defintion_code and lift/lower calling)
+        // or two (e.g. enums_::CodeType, enums_::CodeDeclType). a CodeDecl type would be an ideal place to put an askama Template definition,
```
rfk, line 76: I also thought something along these lines when reading your dreamcode - the `CodeType` seemed to be doing double duty:

* The `declaration_code` method is intended to be called exactly once to render the code for implementing the actual type.
* The other methods are intended to be called multiple times at various other points in the rendering, to emit helper code that calls into parts of the declaration code.

Also, there are things for which we'll need declaration code which are not in any meaningful way a "type", such a function definitions. Splitting the declaration code from the type-helpers code seems like a promising approach to me.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -63,236 +71,179 @@ pub struct KotlinWrapper<'a> {
...
-            FFIType::RustArcPtr => "Pointer".to_string(),
-            FFIType::RustBuffer => "RustBuffer.ByValue".to_string(),
-            FFIType::ForeignBytes => "ForeignBytes.ByValue".to_string(),
-            FFIType::ForeignCallback => "ForeignCallback".to_string(),
-        })
-    }
+impl KotlinLanguageOracle {
+    fn create_code_type(&self, type_: TypeIdentifier) -> Box<dyn CodeType> {
+        // I really want access to the ComponentInterface here so I can look up the interface::{Enum, Record, Error, Object, etc}
+        // However, there's some violence and gore I need to do to (temporarily) make the oracle usable from filters.
```
rfk, line 88: Lol, I know what you mean - it would be nice to be able to have backward references from the leaf nodes here (`Enum`, `Record`, etc) to the containing `ComponentInterface`.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -63,236 +71,179 @@ pub struct KotlinWrapper<'a> {
...
-        })
+        let oracle = oracle();
+        Ok(oracle.find(type_)?.read(&oracle, nm))
     }
-}
+
+    /// Get the Kotlin syntax for representing a given low-level `FFIType`.
+    pub fn type_ffi(type_: &FFIType) -> Result<String, askama::Error> {
+        Ok(oracle().ffi_type_label(type_))
+    }
```
rfk, line 219: To check my expectations - the idea is that these filters will eventually go away in favour of directly calling methods on the `CodeType`, right?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
rfk: Also, if we were able to do this in a pre-processing step, it might avoid some of this awkward passing around of global context :-)

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/enum_.rs
```diff
@@ -0,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk, line 3: One thing I like about this re-arrangement of the code, is that the docstring of this module would be a great place to put the narrative docs about how enums work in the Kotlin. I like having an obvious place for such docs.

rfk (commented):
Some unordered thoughts from a first look at this, I figure it's better to optimize for over-sharing of the things that pop into my head given the async nature here.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
jhugman: I think so. If there was an easier way to get things into a template, without making it part of a struct this would be my goal.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
jhugman: I can see having a pre-processing step would be nice.

I've recently come a-cropper with `ci.type_contains_object_references` and `member.contains_unsigned_types(ci)`. It would be nice if I didn't need to calculate this all the time.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
jhugman: This approach is beginning to bare fruit.

uniffi_bindgen/src/bindings/backend.rs
jhugman: Yeah, it's not clear to me yet.

uniffi_bindgen/src/bindings/backend.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn type_label(&self, oracle: &dyn LanguageOracle) -> StringReturn;
+
+    fn canonical_name(&self, oracle: &dyn LanguageOracle) -> StringReturn;
+
+    fn literal(&self, oracle: &dyn LanguageOracle, literal: &Literal) -> StringReturn;
+    /// Get a Kotlin expression for lowering a value into something we can pass over the FFI.
+    ///
+    /// Where possible, this delegates to a `lower()` method on the type itself, but special
+    /// handling is required for some compound data types.
+    fn lower(&self, oracle: &dyn LanguageOracle, nm: &dyn fmt::Display) -> StringReturn;
```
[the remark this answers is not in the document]
jhugman: This is more like `lift_expression`, rather than `lift_code`.

examples/rondpoint/tests/test_generated_bindings.rs
```diff
@@ -2,8 +2,8 @@ uniffi_macros::build_foreign_language_testcases!(
     "src/rondpoint.udl",
     [
         "tests/bindings/test_rondpoint.kts",
-        "tests/bindings/test_rondpoint.swift",
-        "tests/bindings/test_rondpoint.py",
-        "tests/bindings/test_rondpoint.rb",
+        // "tests/bindings/test_rondpoint.swift",
+        // "tests/bindings/test_rondpoint.py",
+        // "tests/bindings/test_rondpoint.rb",
```
rfk, line 7: (Just noting in case you weren't aware, it's possible to run just the kotlin tests by using `cargo test kt` without commenting out the ones for the other languages)

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
rfk: This is one of those cases where having parent pointers in the member structs would be valuable as well.

uniffi_bindgen/src/bindings/backend.rs
tarikeshaq: Yeah... 
I think if we go with this `CodeType` approach,
>  Will we end up with a "parent" trait that various backends can specialize

or a variation of it resonates more with me more than the alternative. So we end up having specific operations, like `coerce`, not to use this pattern or depend on the `oracle`. Otherwise, if we keep the union of all operations, when the backends are separated, backend authors may want to keep adding to the `CodeType`... and that creates work for **all** the other backends

Could be worth it for us to say "those are the core operations, everything else is backend-specific, and backend authors should write their own logic" 😞

tarikeshaq (commented):
Did a go through of this and overall think this can definitely help us standardize how backends are built - dropped a couple of thoughts as I was going through 😄

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
jhugman: I'm not convinced that parent pointers would be worth the hassle. 

I'm able to make `MemberDeclaration`s query the `ci` at their construction. 

What I _am_ finding problematic is "Does this `Type::` need an `@ExperimentalUnsignedTypes` annotation, i.e. going from `Type` to  CI member.

examples/rondpoint/tests/test_generated_bindings.rs
jhugman: This has helped immensely.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -60,239 +71,199 @@ impl MergeWith for Config {
...
-                type_,
-                match radix {
-                    Radix::Octal => format!("{:#x}", i),
-                    Radix::Decimal => format!("{}", i),
-                    Radix::Hexadecimal => format!("{:#x}", i),
-                },
-            )?,
-            Literal::Float(string, type_) => typed_number(type_, string.clone())?,
-        })
+impl LanguageOracle for KotlinLanguageOracle {
```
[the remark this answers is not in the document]
jhugman: This seems reasonable. If we introduce a `Type::External` which looks up type transforms from the `Config`…

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -60,239 +72,216 @@ impl MergeWith for Config {
...
+            FFIType::Float64 => "Double".to_string(),
+            FFIType::RustArcPtr => "Pointer".to_string(),
+            FFIType::RustBuffer => "RustBuffer.ByValue".to_string(),
+            FFIType::ForeignBytes => "ForeignBytes.ByValue".to_string(),
+            FFIType::ForeignCallback => "ForeignCallback".to_string(),
+        }
+    }
+}
+
+pub mod filters {
```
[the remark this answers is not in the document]
jhugman: OOh. That's really nice. I have been thinking that the `MemberDeclaration`s may  end up owning a `LanguageOracle`, so the filters can accept an oracle as an argument so we can have a common filters module.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use std::fmt;
+
+use crate::bindings::backend::{
+    CodeType, LanguageOracle, Literal, MemberDeclaration, StringReturn, TypeIdentifier,
+};
+use crate::interface::{ComponentInterface, Object};
+use askama::Template;
+
+use super::filters;
```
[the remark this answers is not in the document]
jhugman: Yeah, a comment is needed here.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
jhugman: The filters are all now incredibly regular. I'm resisting the urge to do away with filters yet, or pass the `oracle` into the filters from the `MemberDeclaration`.

bendk:
I think the general direction of this code is great and have a lot of things to say on a conceptual level.  I haven't looked through all the details, so I can't comment on if everything is correct.

It's been really interesting for me to contrast the work you did this week with what I was up to.  In a way they represent opposite approaches to the same problem, but both solutions make sense when you consider which parts of the code we were working on.

The problem as I see it is that it was hard to understand how types are implemented because the code to implement them were scattered between different parts of the template code and the filter functions.  There wasn't one obvious place to look to see how types were implemented and if you wanted to add a type you would need to fiddle with a lot of little bits (look at the hoops Mark had to go through to support imported types for an example).

The way I approach the scaffolding code was to refactor the trait so that things followed a consistent interface and types didn't need much special case code (for example, `FfiConverter` makes it so you don't need `lift_rs`).  At the end, there were only a couple type-specific functions left, and I moved them together into the `type_logic` module in #1010.

The way you approached it was to keep the complexity, but manage it using the Oracle -> CodeType pattern.  I still have a dream to refactor the bindings code so that we don't need that complexity, but that would be a lot more work on Kotlin than it was with the scaffolding.  I think your approach is great for dealing what we have now.

I think there's one last step we could both do to bring the approaches a bit closer together and make things even more obvious to a new user.  What if both scaffolding and bindings/kotlin got a `type_impl` mod, which is where type-specific implementation code lives.  In scaffolding we could just rename the `type_logic.rs` file from #1010.  In Kotlin, it could be split up over multiple files (maybe oracle.rs, plus a file for each codetype).

I also like the concept of MemberDefinition.  I think your concept of an IR finally clicked for me.  I just posted a link to #uniffi about that.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/miscellany.rs
```diff
@@ -0,0 +1,164 @@
...
+
+use crate::bindings::backend::{CodeOracle, CodeType, Literal};
+use askama::Template;
+use paste::paste;
+use std::fmt;
+
+#[allow(unused_imports)]
+use super::filters;
+
+macro_rules! impl_code_type_for_miscellany {
```
jhugman, line 13: This PR takes no view on whether or not this is a sensible resting place for these types. If it is, then this would also be a good place to put JSON and GUID.

If it isn't, then this would be a good place to put helpers to make adapter/transform `CodeType`s.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs
```diff
@@ -0,0 +1,126 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
jhugman, line 1: This file is another good example of the pattern.

We have a `CodeType` to allow things to move across the FFI, a `CodeDeclaration` to declare the foreign language version of the object, and a `CodeDeclaration` that declares the runtime for objects.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/record.rs
```diff
@@ -0,0 +1,95 @@
...
+    inner: Record,
+    contains_unsigned_types: bool,
+    contains_object_references: bool,
+}
+
+impl KotlinRecord {
+    pub fn new(inner: Record, ci: &ComponentInterface) -> Self {
+        Self {
+            contains_unsigned_types: inner.contains_unsigned_types(ci),
+            contains_object_references: inner.contains_object_references(ci),
```
jhugman, line 76: We derive these from the CI and the member itself, so we don't need to do it in the filters or the askama macros.

uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceRuntime.kt
```diff
@@ -0,0 +1,76 @@
...
+internal const val IDX_CALLBACK_FREE = 0
+
+internal abstract class CallbackInternals<CallbackInterface>(
+    protected val foreignCallback: ForeignCallback
+) {
+    val handleMap = ConcurrentHandleMap<CallbackInterface>()
+
+    // Registers the foreign callback with the Rust side.
+    // This method is generated for each callback interface.
+    abstract fun register(lib: _UniFFILib)
```
jhugman, line 59: Now, this method, like every other call into this file, is done from `callback_interface.rs`.

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -1,33 +1,9 @@
...
-        return code == 2
-    }
-}
-
-class InternalException(message: String) : Exception(message)
-
-// Each top-level error class has a companion object that can lift the error from the call status's rust buffer
-interface CallStatusErrorHandler<E> {
-    fun lift(error_buf: RustBuffer.ByValue): E;
-}
```
jhugman, line 24: This all disappears into `Helpers.kt`. It could go into a `ErrorRuntime.kt`, but it doesn't change and is always needed.

uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
```diff
@@ -18,13 +18,13 @@ inline fun <reified Lib : Library> loadIndirect(
...
             loadIndirect<_UniFFILib>(componentName = "{{ ci.namespace() }}")
-            {% let callback_interfaces = ci.iter_callback_interface_definitions() %}
-            {%- if !callback_interfaces.is_empty() -%}
+            {% let initialization_code = self.initialization_code() %}
+            {%- if !initialization_code.is_empty() -%}
             .also { lib: _UniFFILib ->
-                {% for cb in callback_interfaces -%}
-                CallbackInterface{{ cb.name()|class_name_kt }}Internals.register(lib)
+                {% for code in initialization_code -%}
+                {{ code }}
```
jhugman, line 27: Initialization is now distributed to the edges (i.e. the members `CodeDeclaration`s). Currently, only `callback_interface.rs` uses it.

uniffi_bindgen/src/bindings/kotlin/templates/ObjectRuntime.kt
```diff
@@ -0,0 +1,161 @@
...
+//
+// The easiest way to ensure this method is called is to use the `.use`
+// helper method to execute a block and destroy the object at the end.
+interface Disposable {
+    fun destroy()
+    companion object {
+        fun destroy(vararg args: Any?) {
+            args.filterIsInstance<Disposable>()
+                .forEach(Disposable::destroy)
+        }
```
jhugman, line 15: This method is new, and comes from the reason that when destructors are rendered we no longer have access to the `ComponentInterface`.

Some lifetime-fu was needed, or we could skip it and ask Kotlin to do the little bit of extra work.

uniffi_bindgen/src/bindings/kotlin/templates/wrapper.kt
```diff
@@ -23,47 +23,23 @@ import com.sun.jna.Pointer
...
 
+{%- for imported_class in self.imports() %}
+import {{ imported_class }}
+{%- endfor %}
+
+// The Rust Buffer and 3 templated methods (alloc, free, reserve).
 {% include "RustBufferTemplate.kt" %}
 
-{% include "RustBufferHelpers.kt" %}
+{% include "Helpers.kt" %}
```
jhugman, line 34: I'd like to get this to a point where `Helpers.kt` and `RustBufferTemplate.kt` had no additional templated code left (it's very close).

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -404,34 +413,32 @@ impl<'ci> ComponentInterface {
...
-        self.objects
-            .iter()
-            .map(|obj| {
-                vec![obj.ffi_object_free().clone()]
-                    .into_iter()
-                    .chain(obj.constructors.iter().map(|f| f.ffi_func.clone()))
-                    .chain(obj.methods.iter().map(|f| f.ffi_func.clone()))
-            })
-            .flatten()
+        vec![]
```
jhugman, line 416: How `ffi_function_definitions` are rendered hasn't change yet, but this is the beginning of that idea.

uniffi_bindgen/src/interface/object.rs
```diff
@@ -128,6 +128,14 @@ impl Object {
         self.uses_deprecated_threadsafe_attribute
     }
 
+    pub fn iter_ffi_function_definitions(&self) -> Vec<FFIFunction> {
+        vec![self.ffi_object_free().clone()]
+            .into_iter()
+            .chain(self.constructors.iter().map(|f| f.ffi_func.clone()))
+            .chain(self.methods.iter().map(|f| f.ffi_func.clone()))
+            .collect()
+    }
```
jhugman, line 137: We should like to be able to get `CodeDeclaration` to dispense their own `FFIFunction`s, but this isn't done yet.

jhugman (commented):
Final self review.

tarikeshaq (commented):
Dropping a few comments before I move on to something else, will come back and do a complete read-through

I'll take it on myself to fix up any small nits once I get time since James is on PTO. Will get time to tackle some of those as they pop up tomorrow

rfk:
(I'm sorry I haven't looked at this more thoroughly since last week, but I've been deliberately prioritizing other things for my final few days at Mozilla. This one, I figure I'll be quite happy to dig in to even after I'm no longer being paid to do so 😁)

uniffi_bindgen/src/bindings/backend.rs
```diff
@@ -0,0 +1,131 @@
...
+    ///
+    /// This is especially useful when creating specialized objects or methods to deal
+    /// with this type only.
+    fn canonical_name(&self, oracle: &dyn CodeOracle) -> String {
+        self.type_label(oracle)
+    }
+
+    /// A representation of the given literal for this type.
+    /// N.B. `Literal` is aliased from `interface::Literal`, so may not be whole suited to this task.
+    fn literal(&self, oracle: &dyn CodeOracle, literal: &Literal) -> String;
```
bendk, line 64: Should literal be a `CodeType` method?  This implies that we would handle literals differently depending on which type they're being assigned to.  It seems simpler to just have 1 literal function and have the target language figure out if the assignment is correct or not.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/compounds.rs
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    {% call kt::unsigned_types_annotation(self) %}
+    internal fun read{{ canonical_type_name }}(buf: ByteBuffer): {{ inner_type_name }}? {
+        if (buf.get().toInt() == 0) {
+            return null
+        }
+        return {{ "buf"|read_kt(inner_type) }}
+    }
+ "#
```
bendk, line 133: I mentioned this in the other thread, but to me it feels bad to read/edit code in a string literal.  I would rather these are files in the templates folder and we set the template `path` rather than `source`.

bendk (approved):
I left 2 slightly negative comments, but I like the overall approach.  I'd love to see this merged soon, then we can continue to iterate on it.

Not sure how you feel about the template source vs path issue I've been mentioning.  Like you said in the other issue, it's not a hill I'd choose to die.  If you prefer the code inside the rust files, then maybe we could take a poll or something?   I think it would bother me less if I knew it was decided.

uniffi_bindgen/src/bindings/backend.rs
jhugman: IIRC `Literal`s are already typed, either implicitly or as part of the enum.

The agreement between the argument type and the literal is type is checked in `uniffi_bindgen/src/interface/literal.rs`.

In general, I think the backends should need to do as validity checking/error detection on the `ComponentInterface` as possible.

jhugman:
>  If you prefer the code inside the rust files, then maybe we could take a poll or something? I think it would bother me less if I knew it was decided.

Hey thanks for your review! 

The reasoning where to putting the templates for the helpers in the Rust files is fairly weakly held.

 * keeping the helper code and the expressions needed to call them together.
 * laziness. It meant I had to think about where to put the many small files, and I didn't want to do that.

In the interests of expediency, what say we land this, and then file a good-first-bug issue to move these and the Swift templates into their own files?

uniffi_bindgen/src/bindings/backend.rs
jhugman: But your question: should it be a method in `CodeType`? 

We do get some additional benefits around checking that the type and literals match up, but I don't think it's the point.

One of the motivations for this refactor was to reduce the need for huge exhaustive `match` expressions on `Type`. 

The analogous GOF pattern this would be refactoring to would be the State pattern, which has one `match` expression to decide what state you're in and then the state be able to answer all the specific questions you were matching on before.

The motivation for that is two fold: 

1. make it easier to add and remove states.
2. make it easier to reason about states in isolation.

By leaving the literal function as a mega-`match` makes it slightly harder to do each of these.

tarikeshaq (approved):
+1 on landing this and iterating

Do we need to add anything for the changelog for this at all, or should we not let consumers know of the refactor?

jhugman:
@rfk Not sure if you have any thing to add before merging?

rfk:
> @rfk Not sure if you have any thing to add before merging?

I'll try to take a final look at this this afternoon, but please don't feel like you need to block on me.

rfk (approved):
I really like the direction this is headed, so +1 to landing and iterating. I expect we need to "live in" the new approach for a while before it completely gels and we'll learn more about the right shape for the abstraction boundaries as we work with other backends. Landing and iterating seems like exactly the right way to achieve that :-)

uniffi_bindgen/src/bindings/backend.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn enum_variant_name(&self, nm: &dyn fmt::Display) -> String;
+
+    /// Get the idiomatic rendering of an exception name
+    ///
+    /// This replaces "Error" at the end of the name with "Exception".  Rust code typically uses
+    /// "Error" for any type of error but in the Java world, "Error" means a non-recoverable error
+    /// and is distinguished from an "Exception".
+    fn error_name(&self, nm: &dyn fmt::Display) -> String;
+
+    fn ffi_type_label(&self, ffi_type: &FFIType) -> String;
```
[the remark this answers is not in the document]
jhugman: I adopted `type_label` here and elsewhere as "the foreign language string used to specify a type". For simple types, this _is_ the name; for structural types, the punctuation would be included in the label e.g.`String?`. (I'm not sure I'm convincing myself here).

For the `_name` functions, they all take a String, so they're lexical transformations.

For `canonical_name`, I'm somewhat on the fence. It's not a type label, but an identifier part.

FWIW: Until I did this refactor, I hadn't realised how small a part the `FFIType` played in the bindings.

uniffi_bindgen/src/bindings/backend.rs
jhugman: Gosh darn it. "as _little_ validity checking". Yes.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -60,241 +76,294 @@ impl MergeWith for Config {
...
+            )
+            .collect()
+    }
+
+    pub fn imports(&self) -> Vec<String> {
+        let oracle = &self.oracle;
+        let mut imports: Vec<String> = self
+            .members()
+            .into_iter()
+            .filter_map(|member| member.import_code(oracle))
```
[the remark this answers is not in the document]
jhugman: I've renamed the `CodeType` and `CodeDeclaration` methods as `imports`, since the templates are rendering the import _statements_.

I was thinking about import aliases when writing this as `import_code`, but I don't think that will pan out:

e.g. `import JSONObject as _ImportedJSONObject`.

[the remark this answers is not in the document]
bendk: That actually works great,[ I used it to implement `External`.](https://github.com/mozilla/uniffi-rs/pull/1046/files/1553062784d85a5b5bf7267a7f3ade4ea70a5ae4#diff-7db1b800084cb40cbbd4343d949272a8e3b8972d6c8f4a607fd7c73f8f58897aR42-R48)

PR #994, opened 2021-08-03
WIP: Shared uniffi types

### LINKED ISSUE #478: Improve situation around sharing types between crates.

mhammond:
This is a WIP to fix #478. It has a number of hacky things which need cleaning up, but it works for Python. To understand what I'm trying to do here, see the [new fixture](https://github.com/mhammond/uniffi-rs/tree/shared-uniffi-types/fixtures/imported-types), and in particular, [this UDL](https://github.com/mhammond/uniffi-rs/blob/shared-uniffi-types/fixtures/imported-types/lib/src/imported-types-lib.udl)


uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -62,6 +62,9 @@ pub enum Type {
     Optional(Box<Type>),
     Sequence(Box<Type>),
     Map(/* String, */ Box<Type>),
+    // Types generated by uniffi but defined in other crates.
+    // Associated: (type_name, crate_name)
+    Imported(String, String),
```
rfk, line 67: I wonder if there's value in either making these named fields, or having a helper strict `ImportedType` with named fields and having this be `Imported(ImportedType)`.

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -89,12 +92,14 @@ impl Type {
             // API defined types.
             // Note that these all get unique names, and the parser ensures that the names do not
             // conflict with a builtin type. We add a prefix to the name to guard against pathological
-            // cases like a record named `SequenceRecord` interfering with `sequence<Record>`
-            Type::Object(nm) => format!("Object{}", nm),
+            // cases like a record named `SequenceRecord` interfering with `sequence<Record>`.
+            // However, types that support importing all end up with the same prefix of "Type", so
+            // that the import handling code knows how to find the remote reference.
```
rfk, line 97: It seem worth adding more to the note here to say that it's impossible for such types to have conflicting names, which is what makes it safe for them to share the same prefix. Also I think it'd be fine to include any primitive user-defined type here even if it can't (yet) be imported, such as `Error`.

mhammond:
This has been subsumed by other work that's now largely complete.

PR #995, opened 2021-08-03
Remove workaround for JNA bug when passing small structs by value.

### LINKED ISSUE #334: Revert struct-padding workaround once upstream bug is fixed in JNA

badboy:
This (partially) reverts commit 937f9b76f56e52247bb9a53d281554bacba2553e

The upstream JNA bug[1] has been resolved.
JNA 5.7 includes the fix.

We're upgrading our Docker test image to 5.8 already.

Fixes #334

[1]: https://github.com/java-native-access/jna/issues/1259

### DIFF
--- a/docs/manual/src/kotlin/gradle.md
+++ b/docs/manual/src/kotlin/gradle.md
@@ -42,3 +42,18 @@ allprojects {
     }
 }
 ```
+
+## JNA dependency
+
+UniFFI relies on [JNA] for the ability to call native methods.
+JNA 5.7 or greater is required.
+
+Set the dependency in your `build.gradle`:
+
+```groovy
+dependencies {
+    implementation "net.java.dev.jna:jna:5.7.0@aar"
+}
+```
+
+[JNA]: https://github.com/java-native-access/jna
--- a/uniffi_bindgen/src/bindings/swift/templates/BridgingHeaderTemplate.h
+++ b/uniffi_bindgen/src/bindings/swift/templates/BridgingHeaderTemplate.h
@@ -11,15 +11,15 @@
 // We ensure they are declared exactly once, with a header guard, UNIFFI_SHARED_H.
 #ifdef UNIFFI_SHARED_H
     // We also try to prevent mixing versions of shared uniffi header structs.
-    // If you add anything to the #else block, you must increment the version suffix in UNIFFI_SHARED_HEADER_V2
-    #ifndef UNIFFI_SHARED_HEADER_V2
+    // If you add anything to the #else block, you must increment the version suffix in UNIFFI_SHARED_HEADER_V3
+    #ifndef UNIFFI_SHARED_HEADER_V3
         #error Combining helper code from multiple versions of uniffi is not supported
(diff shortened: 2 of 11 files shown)


docs/manual/src/kotlin/gradle.md
```diff
@@ -42,3 +42,18 @@ allprojects {
     }
 }
 ```
+
+## JNA dependency
+
+UniFFI relies on [JNA] for the ability to call native methods.
+JNA 5.7 or greater is required.
```
rfk, line 49: Could you please also add an entry to the changelog to advertise this requirement? It feels reasonable to me to call this out as a breaking change, since if a consumer updates UniFFI without updating a pinned version of JNA, they might start getting crashes.

uniffi_bindgen/src/bindings/swift/templates/BridgingHeaderTemplate.h
```diff
@@ -48,7 +43,7 @@ typedef struct RustCallStatus {
 // ⚠️ Attention: If you change this #else block (ending in `#endif // def UNIFFI_SHARED_H`) you *must* ⚠️
 // ⚠️ increment the version suffix in all instances of UNIFFI_SHARED_HEADER_V2 in this file.           ⚠️
 #endif // def UNIFFI_SHARED_H
```
rfk, line 45: Hrm...I wonder if technically we should be revving to `UNIFFI_SHARED_HEADER_V2` here. Otherwise, if someone accidentally compiles together two UniFFIed crates that use different versions of UniFFI, one of them will end up using a `RustBuffer` struct definition with missing (or extra) fields. @badboy what do you think?

(I'm also curious if the warnings in the code here could have been better, to help better understand whether we need to do this)

rfk (changes requested):
Thanks! The "request changes" here is specifically about the versioning in the Swift header file, everything else about this looks great 👍🏻

uniffi_bindgen/src/bindings/swift/templates/BridgingHeaderTemplate.h
badboy: I just didn't _read_ that file, I did a simple search-and-replace for the padding fields.
Now re-reading it I agree, a version increment is necessary now

rfk (approved):
🎉

Issue #439, opened 2021-05-07
Move some test code from the examples into coveralls.


mhammond:
Some of the examples have code which is really better described as "test" code, and so tends to detract from the example itself - I'm thinking "Callbacks" in particular, and possibly rondpoint (but that's already more a fixture than an example anyway). #431 added a new "coverall" fixture which probably is a better home for such code.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-55)
┆Issue Number: UNIFFI-55

Issue #333, opened 2020-10-21
Improve linker error reporting during ComponentInterface checksum mismatch to avoid time wastage


dmose:
If an interface has changed, and one side of the bindings (eg gecko_js) has been regenerated, but the other (eg .rs) hasn't, linking fails by design, but in an impressively inscrutable way:

```
 0:11.34 toolkit/library/build/XUL
 0:11.36 warning: /Users/dmose/r/hg-mc/toolkit/library/rust/shared/Cargo.toml: dependency (uniffi_build) specified without providing a local path, Git repository, or version to use. This will be considered an error in future versions
 0:14.18     Finished dev [optimized + debuginfo] target(s) in 2.84s
 0:19.79 Undefined symbols for architecture x86_64:
 0:19.79   "_nimbus_3b6_NimbusClient_opt_out", referenced from:
 0:19.82       __ZN7mozilla3dom12NimbusClient6OptOutERK12nsTSubstringIDsERNS_11ErrorResultE in Unified_cpp_components_nimbus0.o
 0:20.04   "_ffi_nimbus_3b6_rustbuffer_reserve", referenced from:
 0:20.07       __ZN7mozilla3dom13nimbus_detail6Writer9WriteInt8ERKa in Unified_cpp_components_nimbus0.o
 0:20.07       __ZN7mozilla3dom13nimbus_detail6Writer11WriteStringERK12nsTSubstringIDsE in Unified_cpp_components_nimbus0.o
 0:20.29   "_nimbus_3b6_NimbusClient_opt_in_with_branch", referenced from:
 0:20.32       __ZN7mozilla3dom12NimbusClient15OptInWithBranchERK12nsTSubstringIDsES5_RNS_11ErrorResultE in Unified_cpp_components_nimbus0.o
 0:20.53   "_nimbus_3b6_NimbusClient_reset_enrollment", referenced from:
 0:20.56       __ZN7mozilla3dom12NimbusClient15ResetEnrollmentERK12nsTSubstringIDsERNS_11ErrorResultE in Unified_cpp_components_nimbus0.o
 0:20.78   "_nimbus_3b6_NimbusClient_get_active_experiments", referenced from:
 0:20.81       __ZN7mozilla3dom12NimbusClient20GetActiveExperimentsER8nsTArrayINS0_18EnrolledExperimentEERNS_11ErrorResultE in Unified_cpp_components_nimbus0.o
[...]

PR #996, opened 2021-08-03
Move callbacks testing code into its own fixture crate for testing 

badboy:
Part of #439


mhammond (approved):
This is a great improvement, thanks

mhammond:
except for the test failures 😬

rfk:
I wonder if the failures here are the same underlying cause as https://github.com/mozilla/uniffi-rs/issues/333

badboy:
fun fact: I had the same locally until I renamed stuff. Then they went away, and that's why I pushed it. I'm looking into why that's happening now on CI

Issue #299, opened 2020-09-28
[Discuss] Could bindings generators live in separate crates?


rfk:
For our current stage of development, I like the way that we currently bundle the bindings generators for all languages as part of the main `uniffi_bindgen` crate. It simplifies testing and it (mostly) ensures that we don't leave any languages behind. But I wonder if it's the right thing long term:

* As we've seen with the Gecko backend, some integrations are very hard to test in CI. If the Gecko backend was able to live in mozilla-central and be tested as part of Firefox CI then I expect that would be simpler.
* We currently include a Python backend, but it's falling behind, and I [kind of get the impression](https://github.com/mozilla/uniffi-rs/pull/214) it's hard to prioritize reviews for that code because it's not on the critical path for our consumers. We might be better off if I moved that into a separate repo and maintained it as a "passion project".
* If we get more contributors, they might be coming from a different language and have their own set of needs, and they might find it hard to add an additional backend and keep it up to date, similar to the Python one above.

Other projects in the Rust ecosystem appear to have had some success in externalizing specific backends into separate crates, e.g. serde. What would it look like for us to take a similar approach, and when (if ever) would it be worth the costs of doing so?

jhugman:
I agree with all of your above; especially around Gecko, and backends from contributors. 

It might be worth writing down the necessary pre-conditions before we'd consider doing this:

 * finalize the intermediate representation between parser and bindings that we can support in a backwards compatible manner, i.e. `ComponentInterface` and friends.
 * finalize the shape of the generated rust scaffolding, which new languages need will be calling, e.g. errors, strings, destructors, handles etc
 * document the binary protocol for packing and unpacking of records, sequences, optionals, etc.

FWIW: I think as we get closer to maturity for the shape of how backends should be written, the python binding will be easier to review, test and keep up. I'm optimistic now `test_rondpoint.py` is running, adding to the third backend will be considerably faster than the first and second.

rfk:
I've been thinking a bit more about this in the context of #435 and the resulting Ruby backend. I think I was actually over-complicating it in my head a bit, and we more-or-less have all the pieces that we'd need to start restructuring things in this way.

Currently the `uniffi-bindgen` executable has three main jobs:

* When run as `uniffi-bindgen scaffolding`, it reads the `.udl` file and generates the corresponding Rust scaffolding. This currently does not depend on any details of the intended foreign-language consumer(s).
[...]

PR #997, opened 2021-08-03
[WIP] split backends into their own crate

tarikeshaq:
for #299 

An extremely early draft just for visibility

The change count is very deceptive, because right now I didn't remove anything, only added an extra crate that acts as the backend. Note that 99.9% of the code added here is just straight copied from `uniffi_bindgen/kotlin` to be its own crate. only the `ForiegnLanguageBackend` trait and its implementation are the additions

The process was very straight forward and I can right now generate the bindings using a separate crate without invoking the kotlin code in `uniffi_bindgen` at all. For today, I'll create a concrete list of tasks to make this reviewable and define what a demo might look like (to magnify the impact here, I might have a separate repo setup)


rfk:
A random thought...one thing that I don't have a good idea about for this multi-crate setup is, how do we maintain the level of test coverage? It starts to make less sense to have example crates with foreign-language tests for a bunch of different languages, when those languages may be implemented in an entirely separate repo. I'm sure it's tractable, just, something to think about :-)

tarikeshaq:
Dropping a couple of thoughts before I forget, I'm trying to figure out what the best testing story here looks like...

- I'll look at how easy it is to get the `udl` file from a published crate, alternatively, I'll see about having our `coverall` crate dynamically returning it's IDL to backend crates... I may be stretching our "separate the tests" story a little too far

- I'm currently doing something very hacky (import the `coverall` crate as a dev dependency, then point the tests at `target/debug/deps` to find the `cdylib`) - I'll try to find a cleaner alternative to this, at worst have this hacky-ness abstracted away from the backend developer and into our test definition macros

rfk:
> I'll look at how easy it is to get the udl file from a published crate

Hrm. Yes, I bet this is quite hard right now, and it seems like the sort of thing that's worth us investing in making easier.

tarikeshaq:
Spent a bit of time today on figuring out how to get the udl file from a crate published on `crates.io`... possible, but it's not pretty, here's what I have working locally:
- As a part of the uniffi_macros::build_tests macro, we send a network request to `https://crates.io/api/v1/crates/coverall/0.12.0/download`, this downloads a tarball that includes the source code...
- We then use a couple of crates to decompress and unpack the tarball, then get to the `.udl` inside `src` directory

It's very hacky, but possible! and if we abstract it away inside our macros it might even be neat! To make sure I don't sink myself into this, I'm pausing on the testing work here for a second and will get back to this tomorrow when I have more time to work with this

tarikeshaq:
preceded by #1157

PR #998, opened 2021-08-04
Build and deploy the Docker image on `docker-vX` tag pushes

### LINKED ISSUE #443: Move circleci docker image to something with shared ownership
Our CircleCI tests currently use a custom docker image [`rfkelly/uniffi-ci`](https://hub.docker.com/repository/docker/rfkelly/uniffi-ci) in order to execute with the appropriate Kotlin environment, Swift tooling, etc. This image is currently owned by me and I'm the only one that can push to it.

Let's remove this single point of access failure.

Short-term, it seems I can add individual collaborators to the image, so @jhugman @mhammond @lougeniaC64 @dmose and anyone else who'd like the ability to update that image, please let me know your dockerhub username and I can add you.

Longer term, it would be good to have this image be owned by the mozilla dockerhub organization, or some similar arrangement with more scalable access control management. Perhaps we could even build it in CI and arrange for it to be automatically published in a similar way to how e.g. FxA releases are automatically pushed to dockerhub?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-57)
┆Issue Number: UNIFFI-57

badboy:
Fixes #443

To do:

* [ ] Docs on how to release this
* [x] Talk to ops to set the environment variables in our CI env

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,6 +122,28 @@ jobs:
           build-dir: docs/manual/book
           ssh-fingerprints: "ac:68:a3:78:ea:ee:00:05:30:e1:dc:1f:2e:2f:7c:81"
 
+  Deploy Docker:
+    docker:
+      - image: cimg/python
+    steps:
+      - checkout
+      - setup_remote_docker
+
+      - run:
+          name: Build container image
+          command: |
+            docker build -t uniffi-ci:build -f docker/Dockerfile-build .
+      - run:
+          name: Push to Dockerhub
+          command: |
+            cd docker
+            if [ -n "${CIRCLE_TAG}" ]; then
+              # Strip the `docker-` prefix of the tag
+              TAG="${CIRCLE_TAG#docker-}"
+              ./deploy-dockerhub.sh "$TAG"
+              ./deploy-dockerhub.sh "latest"
+            fi
+
 workflows:
   version: 2
   check-formatting:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -112,6 +112,28 @@ jobs:
...
+            docker build -t uniffi-ci:build -f docker/Dockerfile-build .
+      - run:
+          name: Push to Dockerhub
+          command: |
+            cd docker
+            if [ -n "${CIRCLE_TAG}" ]; then
+              # Strip the `docker-` prefix of the tag
+              TAG="${CIRCLE_TAG#docker-}"
+              ./deploy-dockerhub.sh "$TAG"
+              ./deploy-dockerhub.sh "latest"
```
rfk, line 134: I wonder if there'd ever be a case where the new tag is *not* actually the new "latest"

badboy: Honestly for that Docker image I would keep it simple (not even use `X.Y.Z` versioning, just strictly incrementing the version `X -> X+1`.
Rollbacks should be easy enough.

badboy:
issue filed: https://mozilla-hub.atlassian.net/browse/SVCSE-22

badboy:
The credentials are now set up.

badboy:
I tested this on my own image now here: https://hub.docker.com/r/janerik/uniffi-ci-test/tags -- seems to work as expected!

bendk (approved):
Looks good to me, thanks for taking care of this.

badboy:
I'll put up docs once I have this deployed correctly.

PR #1000, opened 2021-08-04
Make `ViaFfi.write()` input an owned value instead of a reference

bendk:
This makes sense since the idea is that code in FFI-land is taking
ownership of the value.  I think that some of my next work is going to
be easier if write() and lower() both have the same self type.

### DIFF
--- a/uniffi/src/lib.rs
+++ b/uniffi/src/lib.rs
@@ -179,7 +179,10 @@ pub unsafe trait ViaFfi: Sized {
     /// This trait method can be used for sending data from rust to the foreign language code,
     /// in cases where we're not able to use a special-purpose FFI type and must fall back to
     /// sending serialized bytes.
-    fn write(&self, buf: &mut Vec<u8>);
+    ///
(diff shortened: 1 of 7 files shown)


uniffi/src/lib.rs
```diff
@@ -179,7 +179,7 @@ pub unsafe trait ViaFfi: Sized {
     /// This trait method can be used for sending data from rust to the foreign language code,
     /// in cases where we're not able to use a special-purpose FFI type and must fall back to
     /// sending serialized bytes.
-    fn write(&self, buf: &mut Vec<u8>);
+    fn write(self, buf: &mut Vec<u8>);
```
rfk, line 182: The docs for the `lower` method have an explicit note to explain why it takes an owned self, could you please add similar docs to the `write`method here for completeness?

rfk (approved):
Great! I'm a little surprised there was no fallout from this change, but I guess that's good news, it means the original choice of `&self` probably wasn't motivated by any particularly deep reason.

PR #1003, opened 2021-08-05
WIP give `ComponentInterface` members a pointer to their parent.

rfk:
Here's something I wanted to toy with given recent work on refactoring our code-generation templates, and the desire for each interface member to be able to render itself. The issue is that sometimes we need to know information about the surrounding interface or its types in order to do the code generation, which shows up in awkward calls like:

```
{% if e.contains_object_references(ci) %}
```

What would it take to make this be just a plain method call without extra global context? Like so:

```
{% if e.contains_object_references() %}
```

In a garbage-collected language, we might conveniently give the `Record`, `Enum` etc objects a reference to their containing `ComponentInterface` and be done with it. That's hard to do in the obvious way in Rust, and the usual workarounds like handing out an `Arc<RefCell<ComponentInterface>>` seem extremely verbose/complex for our needs.

Here's a sketch of something that might work, although I'm not convinced of the value/complexity tradeoff.

Taking the `Enum` struct as an example, I've split it up into two parts:

* `EnumDescr` is the data that describes an enum. Instances of this are owned by a `ComponentInterface` in a normal tree structure without backwards references.
* `Enum` is a kind of aggregate transient pointer, combining a reference to an `EnumDescr` with a reference to its containing `ComponentInterface`. Instances of this are synthesized on demand when inspecting a `ComponentInterface`.

From the consumer's perspective, it calls `ci.get_enum_definition("MyEnum")` and receives an `Enum` struct, and it can call methods on the `Enum` struct to inspect its definition. But these methods all delegate to the underlying `EnumDescr` for the actual data.

It gets a little complicated in the details of nested strcuts like `Variant` and `Field`, and I'll leave some musings inline.

Thoughts?

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
@@ -36,7 +36,7 @@ enum class {{ e.name()|class_name_kt }} {
 {% else %}
 
 {% call kt::unsigned_types_annotation(e) %}
-sealed class {{ e.name()|class_name_kt }}{% if e.contains_object_references(ci) %}: Disposable {% endif %} {
+sealed class {{ e.name()|class_name_kt }}{% if e.contains_object_references() %}: Disposable {% endif %} {
     {% for variant in e.variants() -%}
     {% if !variant.has_fields() -%}
     object {{ variant.name()|class_name_kt }} : {{ e.name()|class_name_kt }}()
@@ -85,14 +85,14 @@ sealed class {{ e.name()|class_name_kt }}{% if e.contains_object_references(ci) 
         }.let { /* this makes the `when` an expression, which ensures it is exhaustive */ }
     }
 
-    {% if e.contains_object_references(ci) %}
+    {% if e.contains_object_references() %}
     @Suppress("UNNECESSARY_SAFE_CALL") // codegen is much simpler if we unconditionally emit safe calls here
     override fun destroy() {
         when(this) {
             {%- for variant in e.variants() %}
             is {{ e.name()|class_name_kt }}.{{ variant.name()|class_name_kt }} -> {
                 {% for field in variant.fields() -%}
-                    {%- if ci.type_contains_object_references(field.type_()) -%}
+                    {%- if field.contains_object_references() -%}
                     this.{{ field.name() }}?.destroy()
                     {% endif -%}
                 {%- endfor %}
--- a/uniffi_bindgen/src/interface/enum_.rs
+++ b/uniffi_bindgen/src/interface/enum_.rs
@@ -78,57 +78,76 @@
 
 use anyhow::{bail, Result};
 
-use super::record::Field;
+use super::record::{Field, FieldDescr};
 use super::types::Type;
-use super::{APIConverter, ComponentInterface};
+use super::{APIConverter, ComponentInterface, CINode};
 
 /// Represents an enum with named variants, each of which may have named
 /// and typed fields.
 ///
 /// Enums are passed across the FFI by serializing to a bytebuffer, with a
 /// i32 indicating the variant followed by the serialization of each field.
+#[derive(Debug)]
+pub struct Enum<'a> {
+    pub(super) parent: &'a ComponentInterface,
+    pub(super) descr: &'a EnumDescr,
+}
+
 #[derive(Debug, Clone, Hash)]
-pub struct Enum {
+pub struct EnumDescr {
     pub(super) name: String,
-    pub(super) variants: Vec<Variant>,
+    pub(super) variants: Vec<VariantDescr>,
     // "Flat" enums do not have, and will never have, variants with associated data.
     pub(super) flat: bool,
 }
 
-impl Enum {
+impl<'a> Enum<'a> {
     pub fn name(&self) -> &str {
-        &self.name
+        &self.descr.name
     }
 
-    pub fn variants(&self) -> Vec<&Variant> {
-        self.variants.iter().collect()
+    pub fn variants(&self) -> Vec<Variant<'_, Self>> {
+        self.descr
+            .variants
+            .iter()
+            .map(|v| Variant {
+                parent: self,
+                descr: v,
+            })
+            .collect()
     }
 
     pub fn is_flat(&self) -> bool {
-        self.flat
+        self.descr.flat
     }
 
-    pub fn contains_object_references(&self, ci: &ComponentInterface) -> bool {
-        // *sigh* at the clone here, the relationship between a ComponentInterace
-        // and its contained types could use a bit of a cleanup.
-        ci.type_contains_object_references(&Type::Enum(self.name.clone()))
-    }
-
-    pub fn contains_unsigned_types(&self, ci: &ComponentInterface) -> bool {
+    pub fn contains_object_references(&self) -> bool {
         self.variants().iter().any(|v| {
-            v.fields()
-                .iter()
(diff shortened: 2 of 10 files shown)


uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -85,14 +85,14 @@ sealed class {{ e.name()|class_name_kt }}{% if e.contains_object_references(ci)
...
-    {% if e.contains_object_references(ci) %}
+    {% if e.contains_object_references() %}
     @Suppress("UNNECESSARY_SAFE_CALL") // codegen is much simpler if we unconditionally emit safe calls here
     override fun destroy() {
         when(this) {
             {%- for variant in e.variants() %}
             is {{ e.name()|class_name_kt }}.{{ variant.name()|class_name_kt }} -> {
                 {% for field in variant.fields() -%}
-                    {%- if ci.type_contains_object_references(field.type_()) -%}
+                    {%- if field.contains_object_references() -%}
```
rfk, line 95: This seems like a nice improvement in usability for the template logic.

uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -78,65 +78,84 @@
...
 /// Represents an enum with named variants, each of which may have named
 /// and typed fields.
 ///
 /// Enums are passed across the FFI by serializing to a bytebuffer, with a
 /// i32 indicating the variant followed by the serialization of each field.
+#[derive(Debug)]
+pub struct Enum<'a> {
+    pub(super) parent: &'a ComponentInterface,
+    pub(super) descr: &'a EnumDescr,
+}
```
rfk, line 94: So the `Enum` struct is now a kind of "fat pointer" thing that contains both a pointer to the data and parent ref to the containing object.

uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -78,65 +78,84 @@
...
-impl Enum {
+impl<'a> Enum<'a> {
     pub fn name(&self) -> &str {
-        &self.name
+        &self.descr.name
     }
 
-    pub fn variants(&self) -> Vec<&Variant> {
-        self.variants.iter().collect()
+    pub fn variants(&self) -> Vec<Variant<'_, Self>> {
```
rfk, line 109: Here, when consuming code lists the variants of the enum, so see that we create a bunch of `Variant` instances on demand with the appropriate pointers.

uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -78,65 +78,84 @@
...
     }
 
     pub fn is_flat(&self) -> bool {
-        self.flat
+        self.descr.flat
     }
 
-    pub fn contains_object_references(&self, ci: &ComponentInterface) -> bool {
-        // *sigh* at the clone here, the relationship between a ComponentInterace
-        // and its contained types could use a bit of a cleanup.
```
rfk, line 113: Hey look, no more clone!

uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -176,27 +195,42 @@ impl APIConverter<Enum> for weedle::InterfaceDefinition<'_> {
 /// Represents an individual variant in an Enum.
 ///
 /// Each variant has a name and zero or more fields.
+
+#[derive(Debug)]
+pub struct Variant<'a, Parent: CINode> {
+    pub(super) parent: &'a Parent,
```
rfk, line 201: So this starts to get a bit complicated...the `Variant` struct may be used as part of an `Enum` or as part of an `Error`, so its parent pointer type needs to be generic. The main property that we want from the parent pointer is the ability to get all the way up to the containing `ComponentInterface`, so I made a trait for that.

An alternative could be to have it point directly to the containing `ComponentInterface` rather than its containing object, which would make this more of an "owner" pointer than a "parent" pointer. But, I can imagine reasons why a variant might need to be able to inspect the properties of its containing Enum/Error.

uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -176,27 +195,42 @@ impl APIConverter<Enum> for weedle::InterfaceDefinition<'_> {
...
     pub fn has_fields(&self) -> bool {
-        !self.fields.is_empty()
+        !self.descr.fields.is_empty()
+    }
+    pub fn contains_object_references(&self) -> bool {
+        self.fields().iter().any(|f| f.contains_object_references())
+    }
+}
+
+impl<'a, Parent: CINode + 'a> CINode for Variant<'a, Parent> {
```
rfk, line 226: I could imagine generating these impls automatically from some sort of macro.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -141,25 +144,34 @@ impl<'ci> ComponentInterface {
     }
 
     /// List the definitions for every Enum type in the interface.
-    pub fn iter_enum_definitions(&self) -> Vec<Enum> {
-        self.enums.to_vec()
+    pub fn iter_enum_definitions(&self) -> Vec<Enum<'_>> {
+        self.enums
+            .iter()
+            .map(|e| Enum {
```
rfk, line 150: Here again we see that we're synthesizing the aggregate pointer thingy on demand.

uniffi_bindgen/src/interface/record.rs
```diff
@@ -48,79 +48,107 @@ use anyhow::{bail, Result};
...
+        Ok(RecordDescr {
             name: self.identifier.0.to_string(),
             fields: self.members.body.convert(ci)?,
         })
     }
 }
 
 // Represents an individual field on a Record.
+#[derive(Debug)]
+pub struct Field<'a, Parent: CINode> {
```
rfk, line 113: We re-use the `Field` struct in two places, for enum variants fields and record fields, hence the generic.

Which makes me wonder...I wonder if enum variants could actually contain a wrapped `Record` in the same way that errors contain a wrapped `Enum`...

uniffi_bindgen/src/interface/record.rs
```diff
@@ -172,7 +200,7 @@ mod test {
         assert_eq!(record.fields().len(), 1);
         assert_eq!(record.fields()[0].name(), "field");
         assert_eq!(record.fields()[0].type_().canonical_name(), "u32");
-        assert!(!record.fields()[0].required);
+        assert!(!record.fields()[0].required());
```
rfk, line 203: We can no longer access the fields directly because they're hidden behind a pointer.

uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -78,65 +78,84 @@
...
+    pub fn contains_object_references(&self) -> bool {
+        self.variants().iter().any(|v| {
+            v.contains_object_references()
+        })
     }
 
-    pub fn contains_unsigned_types(&self, ci: &ComponentInterface) -> bool {
-        self.variants().iter().any(|v| {
-            v.fields()
+    pub fn contains_unsigned_types(&self, _ci: &ComponentInterface) -> bool {
```
badboy, line 130: What's the reason to keep `_ci` here? Is it necessary or just work-in-progress for now?

badboy (commented):
This actually looks simpler than I thought it would be.
It certainly does make the templating a bit easier, so I'm leaning on the side of "worth the complexity cost"

bendk:
This is the first time I'm seeing this one and it's pretty interesting, but very bitrotted and I think we're moving away from the ComponentInterface anyways.  I'm going to close it to clean up our PR queue.

PR #1005, opened 2021-08-06
Add a re-usable abstraction for iterating over contained types.

rfk:
We currently have several methods that want to be able to recurse
through all the types contained within a particular item, such as
`type_contains_object_references` and `type_contains_unsigned_types`.
These methods duplicate the logic for recursing into our various
types of struct and are a bit weird to use.

This commit proposes a cleanup to make these methods appear more
consistent to the caller, and reduce code duplication in their
implementation.

The core idea is a new trait `IterTypes`. It can be implemented by
anything that contains instances of the `Type` enum, and provides a
single method `iter_types` to iterate over the the contained types.
It's designed to encapsulate the recursion logic, e.g. the `IterTypes`
impl for `Object` knows that it needs to recurse into each of the
methods, the `IterTypes` impl for a `Method` knows that it needs
to recurse into each argument and return type, etc.

On top of this, the `CompontentInterface` can provide a generic
implementation of methods like `contains_object_references` and
`contains_unsigned_types` that will work for any `IterTypes`. I
think this provides a nice little increase in consistency with how
those methods are implemented and called.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
@@ -36,7 +36,7 @@ enum class {{ e.name()|class_name_kt }} {
 {% else %}
 
 {% call kt::unsigned_types_annotation(e) %}
-sealed class {{ e.name()|class_name_kt }}{% if e.contains_object_references(ci) %}: Disposable {% endif %} {
+sealed class {{ e.name()|class_name_kt }}{% if ci.item_contains_object_references(e) %}: Disposable {% endif %} {
     {% for variant in e.variants() -%}
     {% if !variant.has_fields() -%}
     object {{ variant.name()|class_name_kt }} : {{ e.name()|class_name_kt }}()
@@ -85,14 +85,14 @@ sealed class {{ e.name()|class_name_kt }}{% if e.contains_object_references(ci) 
         }.let { /* this makes the `when` an expression, which ensures it is exhaustive */ }
     }
 
-    {% if e.contains_object_references(ci) %}
+    {% if ci.item_contains_object_references(e) %}
     @Suppress("UNNECESSARY_SAFE_CALL") // codegen is much simpler if we unconditionally emit safe calls here
     override fun destroy() {
         when(this) {
             {%- for variant in e.variants() %}
             is {{ e.name()|class_name_kt }}.{{ variant.name()|class_name_kt }} -> {
                 {% for field in variant.fields() -%}
-                    {%- if ci.type_contains_object_references(field.type_()) -%}
+                    {%- if ci.item_contains_object_references(field) -%}
                     this.{{ field.name() }}?.destroy()
                     {% endif -%}
                 {%- endfor %}
--- a/uniffi_bindgen/src/bindings/kotlin/templates/RecordTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/RecordTemplate.kt
@@ -8,7 +8,7 @@ data class {{ rec.name()|class_name_kt }} (
     {%- endmatch -%}
     {% if !loop.last %}, {% endif %}
     {%- endfor %}
-) {% if rec.contains_object_references(ci) %}: Disposable {% endif %}{
+) {% if ci.item_contains_object_references(rec) %}: Disposable {% endif %}{
     companion object {
         internal fun lift(rbuf: RustBuffer.ByValue): {{ rec.name()|class_name_kt }} {
             return liftFromRustBuffer(rbuf) { buf -> {{ rec.name()|class_name_kt }}.read(buf) }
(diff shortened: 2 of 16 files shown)


uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -85,14 +85,14 @@ sealed class {{ e.name()|class_name_kt }}{% if e.contains_object_references(ci)
         }.let { /* this makes the `when` an expression, which ensures it is exhaustive */ }
     }
 
-    {% if e.contains_object_references(ci) %}
+    {% if ci.contains_object_references(e) %}
```
rfk, line 88: I'm interested in feedback on this change from a consumer perspective. Rather than asking the enum whether it contains object references and giving it the `ci` for context, we know ask the `ci` whether the given item contains object references. I think it makes things a bit more consistent, especially when you look at the thing that's handling fields below, which was already doing the "ask the `ci`" variant.

uniffi_bindgen/src/bindings/kotlin/templates/RecordTemplate.kt
```diff
@@ -33,11 +33,11 @@ data class {{ rec.name()|class_name_kt }} (
         {% endfor %}
     }
 
-    {% if rec.contains_object_references(ci) %}
+    {% if ci.contains_object_references(rec) %}
     @Suppress("UNNECESSARY_SAFE_CALL") // codegen is much simpler if we unconditionally emit safe calls here
     override fun destroy() {
         {% for field in rec.fields() %}
-            {%- if ci.type_contains_object_references(field.type_()) -%}
+            {%- if ci.contains_object_references(field) -%}
```
rfk, line 40: OTOH, maybe the name here needs some sort of subject indicator as the first word...`ci.item_contains_object_references` or similar?

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -216,75 +216,63 @@ impl<'ci> ComponentInterface {
         self.types.get_type_definition(name)
     }
 
-    /// Check whether the given type contains any (possibly nested) Type::Object references.
+    /// Iterate over all the types found within in the given item.
+    /// Unlike the `IterTypes::iter_types` method, this implementation is able to
+    /// recurse into the definitions of user-defined types like records and enums.
+    fn iter_types_in<'a, T: IterTypes>(&'a self, item: &'a T) -> TypeIterator<'a> {
```
rfk, line 222: This is now the single generic place where we do the "recurse through all the types in a given item", and the various other methods and thin wrappers on top of it.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
```diff
@@ -165,7 +165,7 @@ extension ViaFfiUsingByteBuffer {
 }
 
 // Implement our protocols for the built-in types that we use.
-{% if ci.contains_optional_types() %}
+{% if ci.contains_optional_types(ci) %}
```
rfk, line 168: This is a little bit of a usability regression, but OTOH it's nice for it to be consistent with the other methods. Here, we can ask the `ci` whether a thing contains optional types, and we ask it about the `ci` itself. Maybe we should keep the existing `ci.contains_optional_types()` method and call the other ones something like `ci.item_contains_blah_blah(item)`?

rfk (commented):
@lougeniaC64, what do you think of this based on your recent work in and around these methods?

@jhugman does any of this make your work on refactoring the codegen harder (or easier)?

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -216,75 +216,63 @@ impl<'ci> ComponentInterface {
...
+        Box::new(
+            item.iter_types()
+                .map(move |t| {
```
rfk, line 225: Don't ask me why this `move` needs to be here, I wasn't really able to get my head around it to my own satisfaction; all I know is that Rust said to put `move` here and I did and it compiled 😬

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -216,75 +216,63 @@ impl<'ci> ComponentInterface {
...
+                    let nested_types = match t {
+                        Type::Object(nm) => {
+                            self.get_object_definition(nm).map(IterTypes::iter_types)
+                        }
+                        Type::Error(nm) => self.get_error_definition(nm).map(IterTypes::iter_types),
+                        Type::Enum(nm) => self.get_enum_definition(nm).map(IterTypes::iter_types),
+                        Type::Record(nm) => {
```
rfk, line 232: Hrm, actually this probably needs to recurse through `iter_types_in` rather than just `iter_types`, but then it might trigger pathological cases of infinite recursion...

rfk: Indeed, we have this unbounded recursion in the version of this currently on `main`, ref https://github.com/mozilla/uniffi-rs/issues/1006

rfk: I've pushed an update the resolves the infinite-recursion problem.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
bendk: I kind of like `contains_optional_types()`.  What about renaming this method `contains_any_optional_types()`?  I also have a refactor in mind that might remove the need for it altogether.

bendk (approved):
I love this change. I personally had a hard time figuring out a couple of these methods because you had to scan over all the tree walking code before you could see what it's doing.  Now the logic is right in the method, like it should be.

uniffi_bindgen/src/bindings/swift/templates/RustBufferHelper.swift
rfk: Looking at this gain with fresh eyes, I've decided to keep the existing `contains_optional_types()` with no arguments, and made the other ones `item_contains_blah_bah(item)`. I think it's conceptually nicer overall.

rfk:
I'm going to merge this on the strength of Ben's r+, but feedback still welcome of course.

PR #1007, opened 2021-08-06
Implement the core support for defining external types via a typedef

mhammond:
Because @bendk and I are both working towaards what can be broadly described as "external types", I thought I'd put up for review the support for defining these types via `typedef`s in the UDL. Because all bindings bail when this is seen, it doesn't actually enable any new functionality, so there should not be any backwards compatibility issues should we need to tweak things later.

I ended up with a single new type - `Type::External`, but that holds a new enum `ExternalTypeKind`, supporting both `Uniffi` (ie, what I've previously been calling "imported" and `Wrapping` (ie, what I've been previously calling just "external".)

I'm open to all bikeshed, including 2 new types and dropping the new enum, and all the names I've used - but they do seem reasonable to me. I've landed on fairly verbose attribute names in the UDL, but it should avoid confusion if I used plain old "imported" (which could mean anything!)

I see no reason to not land this as soon as both Ben and I need it, but we could also consider just basing our branches on this branch and land as soon as we have actual functionality to land.

WDYT?

### DIFF
--- a/uniffi_bindgen/src/interface/types/finder.rs
+++ b/uniffi_bindgen/src/interface/types/finder.rs
@@ -21,8 +21,8 @@ use std::convert::TryFrom;
 
 use anyhow::{bail, Result};
 
-use super::super::attributes::{EnumAttributes, InterfaceAttributes};
-use super::{Type, TypeUniverse};
+use super::super::attributes::{EnumAttributes, InterfaceAttributes, TypedefAttributes};
+use super::{ExternalTypeKind, Type, TypeUniverse};
 
 /// Trait to help with an early "type discovery" phase when processing the UDL.
 ///
@@ -90,14 +90,48 @@ impl TypeFinder for weedle::EnumDefinition<'_> {
 
 impl TypeFinder for weedle::TypedefDefinition<'_> {
     fn add_type_definitions_to(&self, types: &mut TypeUniverse) -> Result<()> {
-        if self.attributes.is_some() {
-            bail!("no typedef attributes are currently supported");
+        let name = self.identifier.0;
+        let attrs = TypedefAttributes::try_from(self.attributes.as_ref())?;
+        // It is simple to support aliases, but it's not clear they really
+        // add value here and having such different semantics from `[Imported]`
+        // ones might just add confusion.
+        // If we *did*, it would be as easy as:
+        // > let t = types.resolve_type_expression(&self.type_)?;
+        // > types.add_type_definition(name, t)
+        if attrs.is_wrapping() {
+            // external hand-written type.
+            let t = types.resolve_type_expression(&self.type_)?;
+            match t {
+                // slight tension with `Type` being the "primitive" - `FfiType` is
+                // closer, but awkward in other ways. There's no good reason to insist on strings
+                // but our experimental implementations might assume so in the short term.
+                Type::String => types.add_type_definition(
+                    name,
+                    Type::External {
+                        name: name.to_string(),
+                        kind: ExternalTypeKind::Wrapping {
+                            primitive: Box::new(t),
+                        },
+                    },
+                ),
+                _ => bail!("only external aliases to strings are supported"),
+            }
+        } else {
+            match attrs.get_external_uniffi_crate_name() {
+                // Note we don't resolve `self._type` - ideally we'd check it is the string
+                // `extern`, but that seems trickier than it looks!
+                Some(crate_name) => {
+                    let type_ = Type::External {
+                        name: name.to_string(),
+                        kind: ExternalTypeKind::Uniffi {
+                            crate_name: crate_name.to_string(),
+                        },
+                    };
+                    types.add_type_definition(name, type_)
+                }
+                None => bail!("only `[ExternalUniffi=crate_name]` or `[ExternalWrapping]` typedefs are supported"),
+            }
         }
-        // For now, we assume that the typedef must refer to an already-defined type, which means
-        // we can look it up in the TypeUniverse. This should suffice for our needs for
-        // a good long while before we consider implementing a more complex delayed resolution strategy.
-        let t = types.resolve_type_expression(&self.type_)?;
-        types.add_type_definition(self.identifier.0, t)
     }
 }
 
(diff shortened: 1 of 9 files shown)


uniffi_bindgen/src/interface/types/finder.rs
```diff
@@ -90,14 +90,48 @@ impl TypeFinder for weedle::EnumDefinition<'_> {
 
 impl TypeFinder for weedle::TypedefDefinition<'_> {
     fn add_type_definitions_to(&self, types: &mut TypeUniverse) -> Result<()> {
-        if self.attributes.is_some() {
-            bail!("no typedef attributes are currently supported");
+        let name = self.identifier.0;
+        let attrs = TypedefAttributes::try_from(self.attributes.as_ref())?;
+        // It is simple to support aliases, but it's not clear they really
+        // add value here and having such different semantics from `[Imported]`
```
mhammond, line 96: `[Imported]` is out of date

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -62,6 +62,10 @@ pub enum Type {
     Optional(Box<Type>),
     Sequence(Box<Type>),
     Map(/* String, */ Box<Type>),
+    External {
+        name: String,
+        kind: ExternalTypeKind,
+    },
```
bendk, line 68: This comment is going to assume a couple things for simplicity:
  - The code in my branch lands in some form or another
  - We adopt Ryan's proposal to rename `ViaFfi` to `FfiConverter`

Maybe this could be `ExternalFfiConverter` since we can support eternal types with Record/Enum/Object -- what's different here is the `FfiConverter` is located somewhere else.

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -146,10 +152,20 @@ impl From<&Type> for FFIType {
...
     }
 }
 
+// The "kind" of external type.
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub enum ExternalTypeKind {
+    // A type defined by UniFFI from a different crate that this crate `use`s
+    Uniffi { crate_name: String },
+    // A manually hand-written type that wraps the specified primitive.
+    Wrapping { primitive: Box<Type> },
```
bendk, line 166: I have an implementation for this in mind -- it just needs 1 more attribute: `ffi_converter_name: String`.

I also have a dream to extend this to: 
```
  Wrapping {
    ffi_converer_name: String,
    primitive: Box<Type>,
    python_template_string: Option<String>,
    python_create_ffi_converter: Option<String>,
    kotlin_template_string: Option<String>,
    kotlin_ffi_converter_name: Option<String>,
    # repeat for other languages
}
```

bendk (approved):
Overall, I love it.  I'm going to try to do some experiments today in implementing some of it.

uniffi_bindgen/src/interface/types/mod.rs
bendk: Thought about this some more and I think a) we may be able to standardize this some more and b) `ExternalFfiConverter` is an awkward name when you also need to append to it to name the kind

```
enum CustomType {
  # These might be enough to handle all of the scaffolding code
  ffi_converter_name: String,
  primitive: Box<Type>,

  # The bindings code is going to need more specialization
  bindings_kind: CustomTypeBindingsKind,
}

enum CustomTypeBindingsKind {
  StandardBinding(Box<Type>), // Handles the Guid -> String case
  ImportedFromOtherCrate(String /*path to udl file */),
  CustomTemplates(/* some story here about extension types */),
}
```

uniffi_bindgen/src/interface/types/mod.rs
mhammond: That makes sense! Although the language specific stuff looks like maybe it should be inside the various [config structs](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/mod.rs#L68) and maybe specified [in the toml](https://github.com/mhammond/uniffi-rs/commit/40be445c85cbcb31631a7d0f4c539eba63860ee0#diff-eb0847bbe6a9781ee040e3ef30d46ec9fa8bc5a547ff56ca8fcf58f6c365decd)?

bendk: :) I deleted my parent comment because I like my suggestion above much more.  But linking to a config struct is a great suggestion.

uniffi_bindgen/src/interface/types/mod.rs
bendk: I think your comment applies here to.  `ImportedFromOtherCrate` could point to a `Config` object or something similar.

bendk:
I really like this idea of starting with some shared definitions and implementing from there.  I just pushed my `custom-types` branch which is based off the `ffi-converter` branch, takes your changes, and implements the tweaks I suggested above.  What do you think?  I'm definitely open to more iterating/bikeshedding/whatever.

https://github.com/bendk/uniffi-rs/compare/ffi-converter...bendk:custom-types?expand=1

uniffi_bindgen/src/interface/types/mod.rs
bendk: I'm struggling to find a good name of this one.  "CustomType" doesn't really communicate anything about what's going on.  Maybe `PrimitiveWrapper`?

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -146,10 +152,20 @@ impl From<&Type> for FFIType {
...
             | Type::Timestamp
             | Type::Duration => FFIType::RustBuffer,
+            Type::External { .. } => FFIType::RustBuffer,
         }
     }
 }
 
+// The "kind" of external type.
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
+pub enum ExternalTypeKind {
```
rfk, line 162: Reading the code, it's not obvious to me what we want to do differently between the different kinds of external type here. I presume it's something like:

* For `UniFFI` types, we intend to use the generated scaffolding/trait impls from the external crate.
* The `Wrapping` types, we intend to generating our own scaffolding/trait impls.

Is there more to it than that?

It's also not obvious to me how much the two cases will have in common (which, I presume they will have much in common because otherwise we'd make them two different variants of `Type` enum..?).

uniffi_bindgen/src/interface/types/mod.rs
rfk: I don't know what the original comment was, but I'm strongly in favour of keeping language-specific stuff out of the UDL and in the TOML :-)

rfk (commented):
> I see no reason to not land this as soon as both Ben and I need it

FWIW, I support landing it as soon as that becomes the most convenient way to continue iterating on it. (As long as we generate errors when folks try to use the parts that aren't ready yet, as opposed to e.g. successfully generating invalid or unsafe code).

mhammond:
> I don't know what the original comment was, but I'm strongly in favour of keeping language-specific stuff out of the UDL and in the TOML :-)

Further, I don't believe that the .udl or the rust code "owns" the decisions about how the foreign bindings should treat an external type. One Python consumer of a crate that exposes a Guid might be more than happy with the plain string, while a different consumer might want their own custom wrapper that's private to their app.

Re this PR, Ben's revamped it quite a bit, which I'm still getting my head around, so I think this version as it stands probably will not land - so I'm going to close this.

PR #1008, opened 2021-08-06
Ffi converter

bendk:
Okay, I think this code is actually ready to merge.  How are you all feeling about it?

Couple notes here:
  - I wrote up an ADR which reflects my thinking on the main design choice here.  Then I read over jhugman's PR, I realized that we currently use the other option for the Kotlin bindings (I think all the bindings except swift) and it's working okay.  Let's consider the ADR a work in progress.  I don't think that needs to block the merge though.
  - I think there's actually a better way of implementing the pattern here, where we define a struct that can convert between a Rust type and a FFI type.  It basically does the same thing, but without the weirdness of the associated types or needing to generate so many unit structs.  I also don't think that needs to block the merge, this code can be a stepping block to there.


docs/adr/0006-external-types.md
```diff
@@ -0,0 +1,82 @@
+# Replace ViaFfi with FfiConverter to handle external types
+
+* Status: proposed
+* Deciders: bendk, rfkelly, mhammond
+* Consulted: jhugman, jan-erik, travis
+* Date: 2021-08-06
```
bendk, line 6: I'm never sure what to put for "deciders" and "consulted".  I put Ryan and Mark on for deciders since I remember them saying I should make a real PR for the code.  I put James, Jan-Erik, and Travis as consulted since I did some demoing of the code to them yesterday.  Tell me if I should change this and sorry if I forgot anyone's contribution, there's been a lot going on this week!

fixtures/external-types/lib/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+use crate_one::CrateOneType;
+use crate_two::CrateTwoType;
+
+pub struct CombinedType {
+    pub cot: CrateOneType,
+    pub ctt: CrateTwoType,
+}
```
bendk, line 7: mhammond mentioned that these fields don't have the greatest names and that he has a similar fixture which can be merged with this one.  My proposal is that we merge this code to main, then one of us can create a PR that merges/improves the fixture.

fixtures/external-types/lib/src/external-types-lib.udl
```diff
@@ -0,0 +1,16 @@
+namespace external_types_lib {
+    CombinedType get_combined_type(optional CombinedType? cval);
+};
+
+dictionary CrateOneType {
```
skhamis, line 5: Hmm if we stick with the UDL approach I wonder if we'll need something to denote that this is indeed external from the current crate (even though it's awesome we can do this now) even if functionally it does nothing. Something like `[ExternalType]` or even `[Crate=<ExternalCrateName>]` feels like it'd help the consumer/viewer quickly understand (imagine if we're using types from 5+ crates the mental juggling would become unwieldy). 

However this just might be me -- open to what other people think

uniffi/src/lib.rs
```diff
@@ -297,19 +307,20 @@ unsafe impl ViaFfi for bool {
...
+unsafe impl FfiConverter for String {
+    type RustType = Self;
     type FfiType = RustBuffer;
 
     // This returns a struct with a raw pointer to the underlying bytes, so it's very
     // important that it consume ownership of the String, which is relinquished to the
     // foreign language code (and can be restored by it passing the pointer back).
-    fn lower(self) -> Self::FfiType {
-        RustBuffer::from_vec(self.into_bytes())
+    fn lower(obj: Self::RustType) -> Self::FfiType {
```
skhamis, line 317: `Self::RustType`initially took my head for a spin but after seeing the implementation and understanding all the context behind it I have to say it's a really interesting approach!

skhamis (commented):
I'll let someone else closer to this do the final approval but after rereading it a couple of times I definitely think this is a great approach to what is essentially sidestepping the orphan rule (but it seems, without creating a whole nest of other problems).

My only issue is (if we're still using udl in the near future) that I personally think there might need to be some way denote that this type is coming from an external crate but otherwise looks awesome! Really nice job with this!

fixtures/external-types/lib/src/external-types-lib.udl
mhammond: This is interesting because Rust doesn't actually need to be told where it is, because it's just a "plain old type" - so it's actually the including Rust code that must have the `use crate_one::CrateOneType` - I was actually surprised this didn't work on main today, but forgot that uniffi in the `lib` crate will try to generate `impl uniffi::RustBufferViaFfi for CrateOneType {` - which fails for the reasons Ben mentions in the adr.

In practice, I doubt that this will end up being a common use-case (ie, that the .udl for a type will be in a different crate than where the type is defined) - I think we'll end up with `crate-one` having its own .udl, and `lib` will want to refer to that via the `typedef` approach we are experimenting with elsewhere. But this is still a good step on the path to that.

uniffi_bindgen/src/scaffolding/templates/EnumTemplate.rs
```diff
@@ -2,31 +2,38 @@
 // For each enum declared in the UDL, we assume the caller has provided a corresponding
 // rust `enum`. We provide the traits for sending it across the FFI, which will fail to
 // compile if the provided struct has a different shape to the one declared in the UDL.
+//
+// We define a unit-struct to implement the trait to sidestep Rust's orphan rule (ADR-0006).
 #}
+
+struct {{ e.type_()|ffi_converter_name }};
```
mhammond, line 9: it's wild to me how this all hangs together without a `self`!

mhammond (approved):
I love this! Let's wait for Ryan, but take my upvote ;)

fixtures/callbacks/src/lib.rs
```diff
@@ -33,6 +33,15 @@ impl RustGetters {
     fn get_list(&self, callback: Box<dyn ForeignGetters>, v: Vec<i32>, arg2: bool) -> Vec<i32> {
         callback.get_list(v, arg2)
     }
+
+    fn get_string_optional_callback<'a>(
```
rfk, line 37: FWIW, this seemed to work correctly for me without the explicit `'a` lifetime annotation, is there a reason it is required?

uniffi_bindgen/src/scaffolding/mod.rs
```diff
@@ -70,67 +68,55 @@ mod filters {
         })
     }
 
-    pub fn lower_rs(nm: &dyn fmt::Display, type_: &Type) -> Result<String, askama::Error> {
-        // By explicitly naming the type here, we help the rust compiler to type-check the user-provided
-        // implementations of the functions that we're wrapping (and also to type-check our generated code).
+    /// Get the name of the FfiConverter implementation for this type
+    ///
+    /// - For primitives / standard types this is the type itself.
+    /// - For user-defined types, this is a unique generated name.  We then generate a unit-struct
```
rfk, line 74: nit: This "We then generate a unit-struct" feels like the first half of some larger thought, is it something that needs to be fleshed out more, or a leftover to be removed?

uniffi_bindgen/src/scaffolding/templates/CallbackInterfaceTemplate.rs
```diff
@@ -73,44 +73,43 @@ impl {{ trait_name }} for {{ trait_impl }} {
...
+unsafe impl uniffi::FfiConverter for {{ trait_impl }} {
+    // This RustType allows for rust code that inputs this type as a Box<dyn CallbackInterfaceTrait> param
+    type RustType = Box<dyn {{ trait_name }}>;
     type FfiType = u64;
     
-    // Lower and write are trivially implemented, but carry lots of thread safety risks, down to
-    // impedence mismatches between Rust and foreign languages, and our uncertainty around implementations
-    // of concurrent handlemaps.
+    // Lower and write are tricky to implement because we have a dyn trait as our type.  There's
+    // probably a way to, but but this carries lots of thread safety risks, down to impedence
```
rfk, line 90: nit: "but but" -> "but"

uniffi_bindgen/src/scaffolding/templates/EnumTemplate.rs
```diff
@@ -2,31 +2,38 @@
 // For each enum declared in the UDL, we assume the caller has provided a corresponding
 // rust `enum`. We provide the traits for sending it across the FFI, which will fail to
 // compile if the provided struct has a different shape to the one declared in the UDL.
+//
+// We define a unit-struct to implement the trait to sidestep Rust's orphan rule (ADR-0006).
```
rfk, line 6: Please consider adding a little more to this comment to help future intrepid debuggers who may be working through the generated Rust code. Suggestion:

"""
We define a unit-struct to implement the trait rather than implementing it directly on the enum.
This sidesteps Rust's orphan rules in the case where the underlying type has been imported from
a separate crate (ADR-0006).
"""

uniffi_bindgen/src/scaffolding/mod.rs
```diff
@@ -70,67 +68,55 @@ mod filters {
...
+            // CallbackInterface is handled by special case code on both the scaffolding and
+            // bindings side.  It's not a unit-struct, but the same name generation code works.
+            Type::Enum(_) | Type::Record(_) | Type::Error(_) | Type::CallbackInterface(_) => {
+                format!("FfiConverter{}", type_.canonical_name())
+            }
+            // Wrapper types are implemented by generics that wrap the FfiConverter implementation of the
+            // inner type.
+            Type::Optional(inner) => format!("Option<{}>", ffi_converter_name(inner)?),
+            Type::Sequence(inner) => format!("Vec<{}>", ffi_converter_name(inner)?),
+            Type::Map(inner) => format!("HashMap<String, {}>", ffi_converter_name(inner)?),
```
rfk, line 96: Aha! Coming back to this work, this is the bit that made things click for me. We're using the generated unit structs to construct an expression *in the typesystem* that knows how to convert nested structures containing imported types, without having to actually materialize any values of that type. So we might end up with `Vec<Option<FfiConverterForImportedType>>` which is not a value at runtime, but which knows how to convert a corresponding `Vec<Option<ImportedType>>` value back-and-forth across the FFI. Neat.

rfk (approved):
This is a very nice refactor, rock on! 👍🏻

CHANGELOG.md
```diff
@@ -25,10 +25,6 @@
       which imports and uses the low-level C FFI. The name can be customized using
       the `module_name` config option.
 - Python timestamps will now be in UTC and timezone-aware rather than naive.
-- Replaced `lower_into_buffer()` and `try_lift_from_buffer()` with the
-  `RustBufferViaFfi` trait.  If you use those functions in your custom ViaFfi
-  implementation then you'll need to update the code.  Check out the `Option<>`
-  implementation in uniffi/src/lib.rs for an example.
```
rfk, line 31: Please consider adding a changelog entry about the new trait here, I think it'll be to our longer-term benefit if we keep ourselves accountable for communicating these changes outward.

uniffi/src/ffi/rustcalls.rs
```diff
@@ -137,6 +135,9 @@ where
 
 /// Wrap a rust function call and return the result directly
 ///
+/// callback is responsible for making the call to the rust function.  It must convert any return
```
badboy, line 138:
```suggestion
/// `callback` is responsible for making the call to the Rust function.  It must convert any return
```

uniffi/src/ffi/rustcalls.rs
```diff
@@ -137,6 +135,9 @@ where
...
+/// value into a type that implements IntoFfi (typically handled with FfiConverter::lower()).
```
badboy, line 139:
```suggestion
/// value into a type that implements `IntoFfi` (typically handled with `FfiConverter::lower()`).
```

uniffi/src/ffi/rustcalls.rs
```diff
@@ -149,7 +150,12 @@ where
     make_call(out_status, || Ok(callback()))
 }
 
-/// Wrap a rust function call that returns a `Result<>`
+/// Wrap a rust function call that returns a `Result<_, RustBuffer>`
+///
+/// callback is responsible for making the call to the rust function.
+///   - callback must convert any return value into a type that implements IntoFfi
+///   - callback must convert any Error the into a `RustBuffer` to be returned over the FFI
+///   - (Both of these are typically handled with FfiConverter::lower())
```
badboy, line 158:
```suggestion
/// `callback` is responsible for making the call to the Rust function.
///   - `callback` must convert any return value into a type that implements `IntoFfi`
///   - `callback` must convert any `Error` the into a `RustBuffer` to be returned over the FFI
///   - (Both of these are typically handled with `FfiConverter::lower()`)
```

uniffi/src/lib.rs
```diff
@@ -8,23 +8,24 @@
 //! component scaffolding in order to transfer data back and forth across the C-style FFI layer,
 //! as well as some utilities for testing the generated bindings.
 //!
-//! The key concept here is the [`ViaFfi`] trait, which must be implemented for any type that can
-//! be passed across the FFI, and which determines:
+//! The key concept here is the [`FfiConverter`] trait, which is responsible for converting between
+//! a Rust type and a low-level C-style type that can be passed across the FFI:
 //!
-//!  * How to [represent](ViaFfi::FfiType) values of that type in the low-level C-style type
+//!  * How to [represent](FfiConverter::FfiType) values of the rust type in the low-level C-style type
```
badboy, line 14:
```suggestion
//!  * How to [represent](FfiConverter::FfiType) values of the Rust type in the low-level C-style type
```

(I know I'm pedantic here)

uniffi/src/lib.rs
```diff
@@ -8,23 +8,24 @@
...
 //!    system of the FFI layer.
-//!  * How to ["lower"](ViaFfi::lower) rust values of that type into an appropriate low-level
+//!  * How to ["lower"](FfiConverter::lower) values of the rust type into an appropriate low-level
```
badboy, line 16:
```suggestion
//!  * How to ["lower"](FfiConverter::lower) values of the Rust type into an appropriate low-level
```

uniffi/src/lib.rs
```diff
@@ -8,23 +8,24 @@
...
 //!    FFI value.
-//!  * How to ["lift"](ViaFfi::try_lift) low-level FFI values back into rust values of that type.
-//!  * How to [write](ViaFfi::write) rust values of that type into a buffer, for cases
+//!  * How to ["lift"](FfiConverter::try_lift) low-level FFI values back into values of the rust
+//!    type.
+//!  * How to [write](FfiConverter::write) values of the rust type into a buffer, for cases
 //!    where they are part of a compound data structure that is serialized for transfer.
-//!  * How to [read](ViaFfi::try_read) rust values of that type from buffer, for cases
+//!  * How to [read](FfiConverter::try_read) values of the rust type from buffer, for cases
```
badboy, line 22:
```suggestion
//!  * How to [read](FfiConverter::try_read) values of the Rust type from buffer, for cases
```

uniffi/src/lib.rs
```diff
@@ -8,23 +8,24 @@
...
 //!
-//!  * How to [represent](ViaFfi::FfiType) values of that type in the low-level C-style type
+//!  * How to [represent](FfiConverter::FfiType) values of the rust type in the low-level C-style type
 //!    system of the FFI layer.
-//!  * How to ["lower"](ViaFfi::lower) rust values of that type into an appropriate low-level
+//!  * How to ["lower"](FfiConverter::lower) values of the rust type into an appropriate low-level
 //!    FFI value.
-//!  * How to ["lift"](ViaFfi::try_lift) low-level FFI values back into rust values of that type.
-//!  * How to [write](ViaFfi::write) rust values of that type into a buffer, for cases
+//!  * How to ["lift"](FfiConverter::try_lift) low-level FFI values back into values of the rust
```
badboy, line 18:
```suggestion
//!  * How to ["lift"](FfiConverter::try_lift) low-level FFI values back into values of the Rust
```

uniffi/src/lib.rs
```diff
@@ -8,23 +8,24 @@
...
+//!    type.
+//!  * How to [write](FfiConverter::write) values of the rust type into a buffer, for cases
```
badboy, line 20:
```suggestion
//!  * How to [write](FfiConverter::write) values of the Rust type into a buffer, for cases
```

uniffi/src/lib.rs
```diff
@@ -8,23 +8,24 @@
...
 //!    where they are part of a compound data structure that is serialized for transfer.
-//!  * How to [read](ViaFfi::try_read) rust values of that type from buffer, for cases
+//!  * How to [read](FfiConverter::try_read) values of the rust type from buffer, for cases
 //!    where they are received as part of a compound data structure that was serialized for transfer.
 //!
 //! This logic encapsulates the rust-side handling of data transfer. Each foreign-language binding
```
badboy, line 25:
```suggestion
//! This logic encapsulates the Rust-side handling of data transfer. Each foreign-language binding
```

uniffi/src/lib.rs
```diff
@@ -8,23 +8,24 @@
...
 //! must also implement a matching set of data-handling rules for each data type.
 //!
-//! In addition to the core` ViaFfi` trait, we provide a handful of struct definitions useful
+//! In addition to the core` FfiConverter` trait, we provide a handful of struct definitions useful
```
badboy, line 28:
```suggestion
//! In addition to the core `FfiConverter` trait, we provide a handful of struct definitions useful
```

uniffi/src/lib.rs
```diff
@@ -142,7 +143,14 @@ macro_rules! assert_compatible_version {
 /// In general, you should not need to implement this trait by hand, and should instead rely on
 /// implementations generated from your component UDL via the `uniffi-bindgen scaffolding` command.
 
-pub unsafe trait ViaFfi: Sized {
+pub unsafe trait FfiConverter: Sized {
+    /// The type used in rust code.
```
badboy, line 147:
```suggestion
    /// The type used in Rust code.
```

uniffi/src/lib.rs
```diff
@@ -142,7 +143,14 @@ macro_rules! assert_compatible_version {
...
+    ///
+    /// For primitive / standard types, we implement FfiConverter on the type itself and RustType=Self.
```
badboy, line 149:
```suggestion
    /// For primitive / standard types, we implement `FfiConverter` on the type itself with `RustType=Self`.
```

uniffi/src/lib.rs
```diff
@@ -162,7 +170,7 @@ pub unsafe trait ViaFfi: Sized {
     ///
     /// Note that this method takes an owned `self`; this allows it to transfer ownership
```
badboy, line 171:
```suggestion
    /// Note that this method takes an owned `Self::RustType`; this allows it to transfer ownership
```

uniffi/src/lib.rs
```diff
@@ -182,7 +190,7 @@ pub unsafe trait ViaFfi: Sized {
     ///
     /// Note that this method takes an owned `self` because it's transfering ownership
```
badboy, line 191:
```suggestion
    /// Note that this method takes an owned `Self::RustType` because it's transfering ownership
```

uniffi/src/lib.rs
```diff
@@ -215,7 +223,7 @@ pub fn check_remaining(buf: &[u8], num_bytes: usize) -> Result<()> {
     Ok(())
 }
 
-/// Blanket implementation of ViaFfi for numeric primitives.
+/// Blanket implementation of FfiConverter for numeric primitives.
```
badboy, line 226:
```suggestion
/// Blanket implementation of `FfiConverter` for numeric primitives.
```

badboy (approved):
Couple of nits.

I like this approach.

fixtures/external-types/lib/src/external-types-lib.udl
bendk: I'm 50/50 on if it should be tagged in the UDL.  On one hand, it's nice to make it clear where the thing is coming from.  On the other hand, it's kind of redundant.  I went this way because it was just easier to implement, but either is fine with me.

docs/adr/0006-external-types.md
```diff
@@ -0,0 +1,82 @@
...
+  * We define `FfiConverter` which works almost the same as `ViaFfi` except
+    instead of always converting between `Self` and `FfiType`, we define a
+    second associated type `RustType` and convert between `RustType` and
+    `FfiType`.
+
+  * For each user-defined type (Record, Error, Object, and Enum), we create a
+    new unit struct and set `RustType` to the type.  This handles external types
+    without issues since we're implementing `FfiConverter` on our on struct.
+    The orphan rule doesn't apply to associated types.
+
```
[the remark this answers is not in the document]
bendk: This ADR needs some work I think.  I actually have 1 more idea that comes from your suggestion of trying to leverage generics that I think could be really neat.

I think I should change the focus away from the specifics of FfiConverter and focus on the decision to do the wrapping in the target language rather than in the template code.

PR #1009, opened 2021-08-06
Drop ffi-support dependency

bendk:
This seems like a nice little cleanup.  The `rustcalls` mod was importing
`IntoFfi`, but only using a trivial part of the interface.  Now we just
define our own trait to do that.

I was a slightly worried about this when I initially implemented the
`rustcalls` module, because I was concenced about uniffi consumers defining their own
`ViaFfi` implementation with some other `FfiType`.  However I now realize
this is impossible, we have a `FfiType` enum in the scaffolding code that
spells out the possible types.

The last bit of functionality was the panic handling hook, I just copied
the code for that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.12.0...HEAD).
 
 ### ⚠️ Breaking Changes ⚠️
+- UniFFI no longer has ffi-support as a dependency.  This means it handles
+  panic logging on its own.  If you previously enabled the `log_panics` feature
+  for `ffi-support`, now you should enable it for `uniffi`.
 - The Swift bindings now explicitly generate two separate Swift modules, one for
   the high-level Swift code and one for the low-level C FFI. This change is intended
   to simplify distribution of the bindings via Swift packages, but brings with it
--- a/uniffi/src/ffi/ffidefault.rs
+++ b/uniffi/src/ffi/ffidefault.rs
@@ -1,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! FfiDefault trait
+//!
+//! When we make a FFI call into Rust we always need to return a value, even if that value will be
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.12.0...HEAD).
 
 ### ⚠️ Breaking Changes ⚠️
+- uniffi no longer has ffi-support as a dependency.  This means it handles
+  panic logging on its own.  If you previously enabled the `log_panics` feature
+  for `ffi-support`, now you should enable it for `uniffi`.
```
bendk, line 18: I think this is correct, but can someone double check that I'm saying the right thing here?

mhammond: That sounds correct to me. This patch is great - it took me some time to see how it worked before it clicked just how little `IntoFfi` was used.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.12.0...HEAD).
 
 ### ⚠️ Breaking Changes ⚠️
+- uniffi no longer has ffi-support as a dependency.  This means it handles
```
rfk, line 16: nit: *UniFFI* 😁

uniffi/src/ffi/ffidefault.rs
```diff
@@ -0,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! FfiDefault trait
+//!
+//! When we make a FFI call into Rust we always need to return a value, even that value will be
```
rfk, line 7: nit: "even that" -> "even if that"

uniffi/src/ffi/ffidefault.rs
```diff
@@ -0,0 +1,52 @@
...
+//! ignored because we're flagging an exception.  This trait defines what that value is for our
+//! supported ffi types.
```
rfk, line 9: nit: capitalization of FFI (which, I don't feel strongly about the right capitalization, but there are two different ones used in the same paragraph here).

rfk (approved):
Makes sense to me!

PR #1010, opened 2021-08-06
[WIP] Refactor to make it easier to implement new types

bendk:
What do y'all think about this refactor of the scaffolding code?  It's not a big change, but I think it might help people understand the current type system better and also be able to create new types easier.  It's based against my ffi-converter branch -- I'm hoping we can land that one.  The main changes are:

- Added a step list in the Type doc comment
- Put all the scaffolding/generation logic in 1 module.  As a side
  benefit, we don't have to make those functions return
  `askama::Result` when they always succeed.
- Document where FFIConverters are defined and a broad overview
  of how they work in the ffi_converter_name() method.


bendk:
I'm not sure I love `type_logic` as the name of the module.  Maybe `type_impl` would be better.  I want to communicate to a new developer, "this is where you look to see where the types are implement in".

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -35,6 +35,18 @@ pub(super) use resolver::{resolve_builtin_type, TypeResolver};
 /// Represents all the different high-level types that can be used in a component interface.
 /// At this level we identify user-defined types by name, without knowing any details
 /// of their internal structure apart from what type of thing they are (record, enum, etc).
+//
+// How to implement a new type:
+//
+// - Add a new Type variant
```
rfk, line 41: I believe we eventually want to get out of the business of adding more variants to this Enum in order to support more types, and it may be worth mentioning that here. That said, it's useful to have the current state of the world documented, even if we're still trying to figure out what the new state of the world can look like, so 👍🏻 to add this.

uniffi_bindgen/src/scaffolding/type_logic.rs
```diff
@@ -0,0 +1,68 @@
+// All type-specific logic for the scaffolding templates lives here.
+//
```
rfk, line 2: May as well make these `//!` comments to turn them into a docstring

rfk (approved):
Makes sense to me, and I think it'll help us iterate towards something even better. I agree with your suggestion of `type_impl` as a better name FWIW, but I don't feel too strongly about it.

mhammond:
FTR, I'm slightly "meh" about this - I agree entirely with Ryan's comment about us wanting to get out of the "implement a new type" business entirely, so I'm "meh" about making it easier :) But "meh" isn't a 👎 , so 👍 :)

uniffi_bindgen/src/interface/types/mod.rs
bendk: Yeah, I agree.  The way I see it, we want to add some sort of wrapper/adapter type.  However, it's hard to add new types right now.  So I'm hoping to improve that situation then implement the wrapper type.  Then we're at a point where it's now easy to add a new type, but we probably don't need to because we can just wrap an existing type.  It's silly in a way, but I still think it's useful because the work to make it easier to add a new type will also make it easier to understand how the current types work.

PR #1011, opened 2021-08-06
[WIP] Type wrappers

bendk:
I wish I had another half hour to polish this some more, but it seems like this approach would be doable to me.  The python tests are currently passing.  I haven't tested GuidHelper yet, but I don't see why it wouldn't work.

I feel like the scaffolding side of things is makes it much easier to implement new types, I would love to refactor the bindings templates to put the FfiConverter code for a type all in 1 place.  Right now it feels like to add a type you need to make changes in a bunch of random files.  On that topic, I saw an crash in the `RustBuffer` code and fixed that to make the tests pass, but I suspect that different primitives would cause a failure (like a `Timestamp` type that wrapped `u64`).

### DIFF
--- a/fixtures/ext-types/guid/src/lib.rs
+++ b/fixtures/ext-types/guid/src/lib.rs
@@ -1,0 +1,68 @@
+use uniffi::FfiConverter;
+
+// A trivial guid.
+pub struct Guid(pub String);
+
+fn get_guid(guid: Option<Guid>) -> Guid {
+    match guid {
+        Some(guid) => guid,
+        None => Guid("NewGuid".to_string()),
+    }
+}
+
+pub struct GuidHelper {
+    pub guid: Guid,
+    pub guids: Vec<Guid>,
+}
+
+fn get_guid_helper(vals: Option<GuidHelper>) -> GuidHelper {
+    match vals {
+        None => GuidHelper {
+            guid: Guid("first-guid".to_string()),
+            guids: vec![
+                Guid("second-guid".to_string()),
+                Guid("third-guid".to_string()),
+            ],
+        },
+        Some(vals) => vals,
+    }
+}
+
+// The name is hard-coded based on the 2 types involved.  Definitely open to a suggestions for how
+// we generate the name.
+struct FFIConverterWrappingGuidString;
+
+impl FFIConverterWrappingGuidString {
+    pub fn to_wrapper(guid: Guid) -> String {
+        guid.0
+    }
+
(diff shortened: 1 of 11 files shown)


rfk (commented):
I don't feel like I have the context for any detailed review here and will leave it to @mhammond, but I'll watch with interest!

uniffi_bindgen/src/bindings/python/gen_python.rs
```diff
@@ -156,7 +156,7 @@ mod filters {
                 coerce_py(&"v", t)?,
                 nm
             ),
-            Type::Custom { .. } => panic!("No support for coercing external types yet"),
+            Type::Custom { primitive, .. } => coerce_py(nm, primitive.as_ref())?,
```
mhammond, line 159: I ended up with 2 new types - `External` and `Wrapped` and dropped the new "Kind" enum, which helped keep the templates fairly simple.

fixtures/ext-types/guid/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+                Guid("third-guid".to_string()),
+            ],
+        },
+        Some(vals) => vals,
+    }
+}
+
+// The name is hard-coded based on the 2 types involved.  Definitely open to a suggestions for how
+// we generate the name.
+struct FFIConverterWrappingGuidString;
```
mhammond, line 33: Based on b5e7836dcc429a43a5a7a72879a985c6bddfa2f9, the canonical name for (almost) all types can be deduced knowing just its name and without knowing its type - this means we can now just `use` an external type without knowing what is in that remote idl.

uniffi_bindgen/src/interface/types/finder.rs
```diff
@@ -106,7 +107,7 @@ impl TypeFinder for weedle::TypedefDefinition<'_> {
                 Type::Custom {
                     name: name.to_string(),
                     primitive: wrapped_type.clone().into(),
-                    ffi_converter_name: "TODO".into(),
+                    ffi_converter_name: format!("FFIConverterWrapping{}{}", name, wrapped_type.canonical_name().to_camel_case()),
```
mhammond, line 110: Given the predictable name, we no longer need to carry this around.

mhammond (commented):
Thanks heaps for this Ben. Over the last week I've pushed further on this and the results of this are in #1030.

bendk:
I'm going to close this one since #1030 actually implements the feature.

PR #1012, opened 2021-08-09
[WIP] Multi-pass code generation prototype

bendk:
This is an attempt at a POC for generating code using a multi-pass approach.  I tried to implement something that followed the concept that I layed out in that google doc and I think it has a ton of promise.

The main idea is to render code with a series of passes that take a vector of stuff and transforms it into a vector of other stuff.  The initial input is a vector of top-level UDL items.  The final output is a vector of strings.  In the middle is the IR `Vec<CodeBlock>`.

The best part of this all is you can log what's going on in each pass, which makes the logic easy to follow.  Take a look at the [example logs](https://github.com/mozilla/uniffi-rs/pull/1012/commits/dec803ff155f6da681b90da5207690351e008d10).  You can also set the input `Vec<UdlItem>` in `main.rs`, do `cargo run`, and get your own set of log files for the process.  The log files have the format:
  - `< [Item]` is an input item
  - `> [Item]` is an output item
  - `X [Item]` is an output item that was pruned because it was a duplicate

The passes are:
  - 00 - Start with a Vec of type-level UDL items
  - 01 - Add the nested UDL items from inside those and remove dupes
  - 02 - Map the UDL items to CodeBlocks and remove dupes
  - 03 - Order the code blocks the right way and also throw in the `Common` code block
  - 04 - Map each code block to a template name and set of JSON data to render it with
  - 05 - Render the templates into string
  - 06 - Join it all together.

This was not all that hard to throw together today.  I still have some TODOs left in the templates logic, but hopefully the overall idea is clear enough.  On the topics of templates, I switched from askama to tera, to support a nice hack enable step 04.

I'm personally think that something like this would be a great improvement.  It makes the logic of the whole system very clear, which is great for newcomers who want to understand what's happening and I think it would make it very easy to implement the new types we want.  It's definitely a decent chunk of work, but it's not so bad.  If nothing else, coding these transform passes was more fun than going through the current template code.


multi-pass-generation-prototype/record-example/06-output.py
```diff
@@ -0,0 +1,148 @@
...
+    pack_fmt = NotImplemented
+
+
+    @staticmethod
+    def lower(builder, value):
+        builder._pack_into(self.size, self.pack_fmt, value)
+
+    @staticmethod
+    def lift(stream, value):
+        return stream._unpack_from(self.size, self.pack_fmt, value)
```
bendk, line 129: This code is totally wrong, but hopefully you can imagine it calling `RustBuffer.consumerWithStream()` then unpacking from there.

bendk:
I should probably clarify exactly what benefit we get from this:
  - More understandable code generation process.  Hopefully this will allow newcomers to get started faster and make UniFFI easier to extend.
     - The logging I mentioned above seems very useful for a newcomer
     - It's straightforward to understand how a Type/function is implemented.  Look how it's mapped in `udl_item_into_code_blocks()`, then check out the template code that corresponds to the output(s).
     - This provides a really nice way to implement wrapper types.  We could have a system where we let the external crate generate the code blocks for its  type.
   - Simplified interface code.  All it needs to do is generate a list of top-level items.
      - This seems like a better separation of responsibilities.  It feels wrong to have to write a method like `iter_rust_wrapped_types()` -- the interface code shouldn't be so coupled to the templates.
      - If we ever want to write our own parser this will be very nice.
  - Automatic dependency management.  We don't generate code blocks for types that aren't used.  If multiple types want to import some library, we will only do it once (which is very nice for external types)

bendk:
I'm gonna close this one.  This would be a fun idea, but I don't know how much benefit it brings over the unit-of-code work, especially if something like  #1052 could be added.

PR #1013, opened 2021-08-09
(cargo-release) version v0.13.0

rfk:


tarikeshaq (approved):
LGTM! Not sure what the breakages are on Application Services, but I imagine we still need to keep this change, but cut another release on top once all is fixed?

rfk:
> I imagine we still need to keep this change, but cut another release on top once all is fixed?

Yep, exactly.

PR #1014, opened 2021-08-10
Restore implicit `Into` conversion when throwing errors.

rfk:
This fixes a subtle regression from #983. Prior to that change, we
were returning errors using Rust's builtin `?` operator and thus
performing an implicit `Into::into` conversion on the error value.
This meant that the underlying implementation of a `[Throws=MyErr]`
function could return anything that was `Into<MyErr>` and our scaffolding
would do the right thing.

In #983 we switched to doing an explicit `.map_err` rather than using
the `?` operator, losing the implicit `Into::into` and causing a
subtle breaking change for any consumers that were relying on the
old behaviour. It turns out our Logins component was relying on it.

This commit restores the previous behaviour by adding an explicit
`Into::into`, and a coveralls testcase to prevent regressing it
in future. I think it's worth commiting to this being explicitly
intended semantics given that it matches the behaviour of the `?`
operator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,15 @@
 
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.13.0...HEAD).
 
+
+### What's Changed
+
+- Fixed an accidental regression in v0.13.0 where errors were no longer being coerced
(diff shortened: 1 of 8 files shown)


tarikeshaq (approved):
Makes sense to me!

PR #1016, opened 2021-08-10
Use fully-qualified type names in generated scaffolding code.

### LINKED ISSUE #1015: Use fully-qualified type names in scaffolding
In #1008 we introduced some new generated scaffolding code that uses the `HashMap` type. This type is not in scope by default and we shouldn't assume that the containing crate will correctly bring it into scope for us. (In practice, the Nimbus crate uses the UDL `Map` type but doesn't bring `HashMap` into scope at the top level).

Let's change the codegen to use fully-qualified paths for types so that we don't depend on details of the surrounding code.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-78)
┆Issue Number: UNIFFI-78

rfk:
This fixes #1015, which was a regression in the v0.13.0 release.

Where possible we should be robust against changes in the details of
the Rust code that implements a UniFFI component, but the generated
Rust scaffolding accidentally started assuming that the `HashMap` type
would be in scope for any interface using a map. This turned out not
to be true in practice.

To improve things, this commit makes the generated scaffolding use
fully-qualified type names for `HashMap`, `Vec` and `Option`. There
is probably more we could do on this front (e.g. the `String` type)
but these were low-hanging fruit that fixed a concrete regression,
so that's where I started.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,14 +12,15 @@
 
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.13.0...HEAD).
 
-
 ### What's Changed
 
 - Fixed an accidental regression in v0.13.0 where errors were no longer being coerced
   to the correct type via `Into`. If the UDL declares a `[Throws=ExampleError]` function
   or method, the underlying implementation can now return anything that is `Into<ExampleError>`,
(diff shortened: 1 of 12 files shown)


tarikeshaq (approved):
💯

PR #1021, opened 2021-08-11
Refactor the tests for type finding in valid UDL

mhammond:
This particular test has been a PIMA; the bigger it gets, the further the test code gets from the UDL under test, and merge conflicts always bite here. I think this is short term pain (ie, conflicts) for longer term gain. Full disclosure - it has lost the check for the total count of the "known types", but I think that's OK (and avoids me ~explaining~ working out how `String` ended up there once `typedef` aliases land 😅


rfk (approved):
👍🏻

PR #1022, opened 2021-08-11
Use a predictable canonical_name for types which support being imported

mhammond:
This seems required in every iteration of the "external types" effort and can't cause harm, so here I am. Flagging Ryan as he suggested adding `Error` etc in another WIP patch, so is somewhat familiar.


rfk (approved):
Sounds good 👍🏻

PR #1023, opened 2021-08-12
Add `[Export]` attribute, which makes the generated FfiConverter for some types `pub`

mhammond:
This is on the path to the external types work and seems clear and stand-alone enough that we can consider merging it before the external types work is actually ready to land. Landing this now avoids rebase pain for the other attributes external-types would like, but they are still in more of a state of flux.

WDYT about landing this now, or would you prefer to wait for the external types work to solidify?

### DIFF
--- a/uniffi_bindgen/src/interface/types/finder.rs
+++ b/uniffi_bindgen/src/interface/types/finder.rs
@@ -21,7 +21,7 @@ use std::convert::TryFrom;
 
 use anyhow::{bail, Result};
 
-use super::super::attributes::{EnumAttributes, InterfaceAttributes};
+use super::super::attributes::{DictionaryAttributes, EnumAttributes, InterfaceAttributes};
 use super::{Type, TypeUniverse};
 
 /// Trait to help with an early "type discovery" phase when processing the UDL.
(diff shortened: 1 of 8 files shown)


mhammond:
Oh, and FTR, we can't unconditionally make those structs `pub`, because if the wrapped type isn't pub, rust will give a compiler error about leaking crate private types. This will force the type with the new attribute to be `pub` for this reason, but we currently support exposing types not marked as pub. At some point in the future we could consider insisting on exposed types being `pub`, because they logically are, but that's a larger discussion about breaking b/w compat we can defer for now.

rfk:
🤔 hrm...

> we currently support exposing types not marked as pub

That's unfortunate. Do we actually do this in practice in any of our production components? (I see that our examples do it, which is unfortunately, and likely entirely accidental).

My concern is that this attribute will be very short-lived, because it's semantically kind of silly - *everything* in the interface file should be exported by default, otherwise what's it doing there?

On the other hand, I guess there's no reason for anyone to actually add this annotation unless they want the specific functionality that you're working on here for cross-crate UniFFI linkage, so probably it won't be that much of a problem if we add it, use it for a release or two, then deprecate it when we insist things be public by default (kind of like the `[Threadsafe]` annotation in that regard).

> At some point in the future we could consider insisting on exposed types being pub,
> because they logically are, but that's a larger discussion about breaking b/w compat we can defer for now.

FWIW, I think there's a compelling argument that we should only expose public types, and I'd support adding some static assertions or other light hackery to enforce this even for things that don't get a `pub impl FFIConverter` generated for them.

uniffi_bindgen/src/interface/types/finder.rs
```diff
@@ -59,11 +59,25 @@ impl TypeFinder for weedle::InterfaceDefinition<'_> {
...
+        } else if ia.contains_error_attr() {
+            if ia.is_exported() {
+                types.mark_as_exported(&name)
+            }
             types.add_type_definition(self.identifier.0, Type::Error(name))
         } else {
+            // `[Exported]` on interfaces is a bit hacky - we support the attribute on an
+            // `Interface` because sometimes they are actually enums or errors, as handled above.
+            // But we don't allow "real" interfaces to be exported because it's tricky to generate
+            // the code and there's no good reason - the original can just be used directly.
```
rfk, line 77: Because they directly impl `FFIConverter` rather than using separate unit struct?

mhammond:
> That's unfortunate. Do we actually do this in practice in any of our production components?

I don't think so, and if we did, it would be entirely accidental too.

> FWIW, I think there's a compelling argument that we should only expose public types, and I'd support adding some static assertions or other light hackery to enforce this even for things that don't get a `pub impl FFIConverter` generated for them.

The patch I'm working would indeed end up enforcing this - in the absence of this attribute, we'd always generate that `pub` struct under the assumption it might be used elsewhere, because without that attribute there's no way to specify it.

So if you prefer, I'd be happy to assume all types are pub and treat that as a breaking change and avoid this short-lived attribute entirely.

rfk (approved):
I support landing this if it keeps your work unblocked, but I don't feel great about it, mostly because I think it's hard to explain what the attribute does and why it needs to exist at all. Do you see this as something we need to document and commit to as part of the UDL syntax, or as a temporary hack until we grapple with the b/w compat concern of requiring things to be public?

If the latter, maybe we could even call it something like `_Exported` (dunno if that's supported by WebIDL) to subtly indicate that it's special.

If the former, how do we explain what it does? Something like "makes the generated bindings useable by other Rust crates?". Seems weird to me, but there's plenty of weird going around I guess :-)

rfk:
> So if you prefer, I'd be happy to assume all types are pub and treat that as a breaking change and avoid
> this short-lived attribute entirely.

I think I'd prefer that outcome, but it seems worth getting a bit broader input on it, and I wouldn't object to landing this in the meantime.

mhammond:
> I think I'd prefer that outcome, but it seems worth getting a bit broader input on it, and I wouldn't object to landing this in the meantime.

OK - for now I'll just keep it in my patch queue and we can make a decision about it later - it's not really *that* much if a PITA for me.

PR #1025, opened 2021-08-13
Make "flat" error enums carry a stringified error message.

### LINKED ISSUE #1024: Undocumented breaking change for "flat" error enums
When we introduced the new "rich" structured error enums, we also introduced an undocumented (maybe unintended?) change to the behaviour of "flat" error enums. Consider an error declared like:

```
[Error]
enum LoginsStorageError {
    "SyncAuthInvalid",
    "MismatchedLock",
    ... others omitted ...
};
```

Prior to v0.13, the generated bindings for this error would be an enum-shaped thing where each variant has a `message` field, and we'd populate the `message` field using the stringification of the Rust-side error. Taking Swift as an example, calling code would match on it like this:

```
switch err{
    case .SyncAuthInvalid(message)
        log.error(message)
    ... others omitted ...
}
```

With v0.13, the generated bindings for this error are now an enum-shaped thing with no fields, and no way to access more info about the error. The Swift calling code now needs to look like this:

```
switch err{
    case .SyncAuthInvalid
        log.error("invalid auth")
    ... others omitted ...
}
```

It's not clear from the changelog that this has occurred, why, or what to do about it.

I figure we have two options, depending on what we think about the new behaviour.

#### 1) Document how to migrate and restore the previous behaviour.

Here, we decide that the new behaviour is a keeper and we update the changelog with a note about the breaking change and some details on how to best migrate to the new richer enums. In theory, consumers can declare an explicit `message` field on enums and populate it on the Rust side to get back to the old behaviour. We should test this out to see if it actually works before committing to it.

#### 2) Revert flat error enums to their previous behaviour.

Here, we decide that this change isn't good for consumers on balance, because they *probably* want to know the underlying error message. We figure out how to revert flat enums to the previous behaviour while keeping the new rich structured enums as they are. Two downsides to watch out for:

* This is probably non-trivial in codegen, we'd have to keep an eye on the complexity.
* This is technically semver-breaking for anyone who already migrated to the v0.13 behaviour.

Any strong opinions on the right way to proceed?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-81)
┆Issue Number: UNIFFI-81
┆Sprint End Date: 2021-08-20

rfk:
Sam Edit: Taken over by @skhamis (but don't want to make a new PR for now). Continuing the work Ryan did on this after having a sync before he left. The current best plan is to leverage the `is_flat()` to determine if it's a simple enum vs the new complex error handling. Issue linked below has more details on the matter:

Fixes #1024 by restoring the behaviour of "flat" error enums declared
with the `[Error] enum` syntax. The generated bindings of these errors
will now carry a single `message` field giving the stringified Rust
error.

### DIFF
--- a/fixtures/coverall/tests/bindings/test_coverall.rb
+++ b/fixtures/coverall/tests/bindings/test_coverall.rb
@@ -98,17 +98,19 @@ class TestCoverall < Test::Unit::TestCase
     coveralls = Coverall::Coveralls.new 'test_simple_errors'
     assert_equal coveralls.get_name, 'test_simple_errors'
 
-    assert_raise Coverall::CoverallError::TooManyHoles do
+    err = assert_raise Coverall::CoverallError::TooManyHoles do
       coveralls.maybe_throw true
     end
+    assert_equal err.message, 'The coverall has too many holes'
 
     assert_raise Coverall::CoverallError::TooManyHoles do
       coveralls.maybe_throw_into true
     end
 
-    assert_raise Coverall::InternalError, 'expected panic: oh no' do
+    err = assert_raise Coverall::InternalError do
       coveralls.panic 'expected panic: oh no'
     end
+    assert_equal err.message, 'expected panic: oh no'
 
     assert_raise_message /expected panic: oh no/ do
       coveralls.panic 'expected panic: oh no'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@
     `private type <type-name> in public interface`
   or similar, please declare the types as `pub` in your Rust code.
 
+- Errors declared using the `[Error] enum` syntax will now expose the error string from
+  Rust to the foreign language bindings. This reverts an unintended change in behaviour
+  from the v0.13 release which made the error message inaccessible.
+
 ### What's Changed
 
 - You can now use external types of various flavours - see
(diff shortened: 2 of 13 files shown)


fixtures/coverall/tests/bindings/test_coverall.rb
```diff
@@ -98,17 +98,19 @@ def test_simple_errors
     coveralls = Coverall::Coveralls.new 'test_simple_errors'
     assert_equal coveralls.get_name, 'test_simple_errors'
 
-    assert_raise Coverall::CoverallError::TooManyHoles do
+    err = assert_raise Coverall::CoverallError::TooManyHoles do
       coveralls.maybe_throw true
     end
+    assert_equal err.message, 'The coverall has too many holes'
```
rfk, line 104: My initial attempt at making a failing testcase here didn't fail, which lead me to discover that we're using the wrong syntax for asserting the error message in Ruby. I tested this new syntax on the existing assertion below to confirm that it works.

skhamis:
I will be continuing the work from @rfk on this. I have gotten the ruby and swift portions down following the pattern he started in python. I almost have kotlin but that one needs a little more nuance to get right so i'll leave it in draft for now!

mhammond (approved):
This LGTM, but sorry about the conflicts - they are my fault but should be easy to resolve.

PR #1026, opened 2021-08-13
Split buffer from the type reader/writer for python.

mhammond:
This came out of a discussion Ryan and I had last week for external types. The idea is that by splitting the "type reader/writer" from the buffer ownership we can support external types in our buffers. For example, external types will end up generating code that looks like:

```python
    @staticmethod
    def readSomeExternalType(stream):
        from some_module import RustBufferTypeReader;
        return RustBufferTypeReader.readSomeExternalType(stream)
```
ie, uses the reader from a different module with a stream that we own.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/templates/RustBufferBuilder.py
+++ b/uniffi_bindgen/src/bindings/python/templates/RustBufferBuilder.py
@@ -1,6 +1,6 @@
 
 class RustBufferBuilder(object):
-    """Helper for structured writing of values into a RustBuffer."""
+    # Helper for structured writing of bytes into a RustBuffer.
 
     def __init__(self):
         self.rbuf = RustBuffer.alloc(16)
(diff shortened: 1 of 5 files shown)


uniffi_bindgen/src/bindings/python/templates/RustBufferBuilder.py
```diff
@@ -1,6 +1,6 @@
 
 class RustBufferBuilder(object):
-    """Helper for structured writing of values into a RustBuffer."""
+    # Helper for structured writing of bytes into a RustBuffer.
```
travis79, line 3: Any reason why you are changing this from a docstring to a standard comment?

uniffi_bindgen/src/bindings/python/templates/RustBufferStream.py
```diff
@@ -1,6 +1,6 @@
 
 class RustBufferStream(object):
-    """Helper for structured reading of values from a RustBuffer."""
+    # Helper for structured reading of bytes from a RustBuffer
```
travis79, line 3: Same question about the reasoning behind demoting this from a docstring to a comment

travis79 (approved):
No need to block on my questions about docstrings, this LGTM

uniffi_bindgen/src/bindings/python/templates/RustBufferBuilder.py
mhammond: docstrings have a (trivial) runtime cost that I don't think is worth it here - this is never going to be rendered for a human - so while I was touching it.. :)

mhammond:
Thanks!

PR #1029, opened 2021-08-15
The Rust impls of all `dictionary`, `enum` or `error` types must be pub

mhammond:
The "external types" effort will cause us to generate code that will fail
to build if they aren't, and over #1023 @rfk and I agreed this was a
reasonable requirement.

Note this patch doesn't actually enforce the new rule - it was
generated by sprinkling

```
pub struct {{ e.name() }}Wrapper(pub {{ e.name() }});
```

around scaffolding_template.rs and fixing the build errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 - The Rust implementations of all `dictionary`, `enum` or `error` types defined in UDL must be
   public. If you see errors like:
     `private type <type-name> in public interface`
-  or similar, please declare the types as `pub` in your Rust.
+  or similar, please declare the types as `pub` in your Rust code.
 
 ## v0.13.1 (_2021-08-09_)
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,12 @@
 
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.13.1...HEAD).
 
+### ⚠️ Breaking Changes ⚠️
+- The Rust implementations of all `dictionary`, `enum` or `error` types defined in UDL must be
+  public. If you see errors like:
+    `private type <type-name> in public interface`
+  or similar, please declare the types as `pub` in your Rust.
```
badboy, line 19:
```suggestion
  or similar, please declare the types as `pub` in your Rust code.
```

badboy (approved):

PR #1030, opened 2021-08-16
More complete "External types"

mhammond:
This builds on Ben's #1011 and show-cases external types that completely work! Only Python is currently supported, but I don't think it's necessary to insist on functioning bindings for other languages before landing this. It has thorough examples, fixtures and documentation - so I think this is actually ready to land!

(I also need to work out how we can arrange to request review from @rfk)


docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -0,0 +1,76 @@
...
+
+    fn wrap(val: Self::Wrapped) -> uniffi::Result<Self> {
+        Ok(Handle(val))
+    }
+
+    fn unwrap(obj: Self) -> Self::Wrapped {
+        obj.0
+    }
+}
+```
```
bendk, line 54: I like this trait.  It makes the wrapping really simple for the library author.

uniffi_bindgen/src/scaffolding/templates/ExternalTypesTemplate.rs
```diff
@@ -0,0 +1,46 @@
...
+    type Wrapped;
+
+    fn wrap(val: Self::Wrapped) -> uniffi::Result<Self> where Self: Sized;
+    fn unwrap(obj: Self) -> Self::Wrapped;
+}
+
+{%- endif -%}
+
+// Type `{{ name }}` wraps a `{{ prim.canonical_name() }}`
+pub struct FfiConverterType{{ name }};
```
bendk, line 24: Now that we're defining a couple things in the scaffolding code, I wonder if we should have a common prefix to namespace things.  Like maybe this should be named `UniffiFfiConverterType{{name}}`

bendk (approved):
This looks great to me. I don't really have many comments because once you add the  `UniffiCustomTypeWrapper` trait, everything is very straightforward.

bendk:
I think this changeset should fix the tests: https://github.com/bendk/uniffi-rs/commit/70a90b440703aec4d3d2680b482c0a21526067d8

uniffi_bindgen/src/scaffolding/mod.rs
```diff
@@ -46,6 +48,8 @@ mod filters {
             Type::Optional(t) => format!("std::option::Option<{}>", type_rs(t)?),
             Type::Sequence(t) => format!("std::vec::Vec<{}>", type_rs(t)?),
             Type::Map(t) => format!("std::collections::HashMap<String, {}>", type_rs(t)?),
+            Type::External { .. } => panic!("External types coming to a uniffi near you soon!"),
+            Type::Wrapped { .. } => panic!("Wrapped types coming to a uniffi near you soon!"),
```
bendk, line 52: Should we add `rust_type` to the enums?  I think we need to implement this in order to return the types directly from a function.

uniffi_bindgen/src/bindings/python/gen_python.rs
```diff
@@ -217,6 +221,8 @@ mod filters {
                 nm,
                 class_name_py(&type_.canonical_name())?
             ),
+            Type::Wrapped { prim, .. } => lift_py(nm, prim.as_ref())?,
+            Type::External { .. } => panic!("should not be necessary to lift External types"),
```
bendk, line 225: Maybe these should be implemented for `Type::External`?  It seems like you would need that if you had a function that returned and/or inputted an external type directly (not as a field in a record like the current code).

bendk (approved):
Took a closer looked and realized there's 1 limitation with the current code.  I still think it should be okay to merge, but it could be good to fix this one.

uniffi_bindgen/src/scaffolding/mod.rs
mhammond: Ooof, that's a good point - and one that I'm not quite sure how to resolve - we don't know the rust type for `External` types because we don't need to look in the "source" udl. Then I saw #1034 which is going to conflict with any changes here anyway, so I think I'm just going to land this with that limitation, just in the interests of allowing us to keep moving forward.

Issue #466, opened 2021-05-26
Skip tests when missing language interpreters, rather than failing them.


rfk:
As suggested in https://github.com/mozilla/uniffi-rs/pull/464#pullrequestreview-668533632, it's a little hostile to expect that everyone working on UniFFI will have *all* of the supporting software for *all* supported languages installed on their development machine in order to run `cargo test`. That's what we have CI for!

Instead of throwing an error when e.g. the Ruby interpreter or the Swift compiler is not available, can we do something similar to "skip test" so that it doesn't fail local builds? (And then in CI, require that no tests be skipped?).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-61)
┆Issue Number: UNIFFI-61

rfk:
It doesn't seem to be possible to dynamically decide to skip a test in Rust (e.g. you can't do a `raise SkipTest` like you could in Python) but Rust does have support for [marking tests as ignored](https://doc.rust-lang.org/book/ch11-02-running-tests.html):

```
#[test]
#[ignore]
fn this_test_wont_run_by_default() {
    assert!(false, "should not be run by default");
}
```

Luckily for us, all the foreign-language tests are auto-generated via the [`build_foreign_language_testcases` macro](https://github.com/mozilla/uniffi-rs/blob/d58e92f58a48e65b0c6ec10c0cb22d103a266e50/uniffi_macros/src/lib.rs#L30) macro. What we could do, is have this macro inspect the system to see whether it's *possible* to run each of the target tests, and include the `#[ignore]` annotation automatically if it is not.  It's kind of poor form for macros to inspect the system like that, but I think it would be fine in this case.

One nice thing about this approach is that in CI, we could run the tests with `cargo test -- --ignored` to force it to run any ignored tests, helping us make sure that the tests actually work in CI even if they don't run on individual dev machines.

rfk:
> What we could do, is have this macro inspect the system to see whether
> it's possible to run each of the target tests, and include the #[ignore] annotation automatically if it is not.

Concretely, we'd have to expose a new function for each language backend like `can_run_tests` or similar, that inspects the system and returns a boolean whether we can run the tests.

[...]

PR #1031, opened 2021-08-16
Specifying $UNIFFI_TESTS_DISABLE_EXTENSIONS can skip foreign tests.

mhammond:
In my local environment I can execute:

```shell
UNIFFI_TESTS_DISABLE_EXTENSIONS=swift:rb cargo test
```

output is nicely:
```
test uniffi_foreign_language_testcase_test_chronological_swift ... ignored
   Compiling uniffi-fixture-time v0.13.1 (/mnt/c/src/moz/uniffi-rs/fixtures/uniffi-fixture-time)
    Finished dev [unoptimized + debuginfo] target(s) in 3.62s
test uniffi_foreign_language_testcase_test_chronological_kts ... ok
test uniffi_foreign_language_testcase_test_chronological_py ... ok

test result: ok. 2 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 12.16s
```

(as opposed to now, when it's lots of noise about failing to execute ruby and swift)

I think we can declare this as fixing #466 too - that suggests a more automated approach, but I'm not sure that will be necessary given @tarikeshaq's work splitting the bindings; this seems fine to me.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -60,6 +60,10 @@ test suite you will need:
   * A `ruby` interpreter.
   * The [`FFI`](https://github.com/ffi/ffi) Ruby gem, installable via `gem install FFI`.
 
+We also support an environment variable `UNIFFI_TESTS_DISABLE_EXTENSIONS`;
+It is a set of file extensions, without a leading period and separated by commas.
+Eg, `UNIFFI_TESTS_DISABLE_EXTENSIONS=swift,rb cargo test` will skip test filenames ending in
+`.swift` or `.rb`
 
 ## Navigating the code
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@
     `private type <type-name> in public interface`
   or similar, please declare the types as `pub` in your Rust code.
 
+### What's Changed
+
+- An environment variable `UNIFFI_TESTS_DISABLE_EXTENSIONS` can disable foreign language bindings
+  when running tests. See [the contributing guide](./contributing.md) for more.
+
 ## v0.13.1 (_2021-08-09_)
 
 [All changes in v0.13.1](https://github.com/mozilla/uniffi-rs/compare/v0.13.0...v0.13.1).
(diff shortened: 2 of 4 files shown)


docs/contributing.md
```diff
@@ -60,6 +60,10 @@ test suite you will need:
   * A `ruby` interpreter.
   * The [`FFI`](https://github.com/ffi/ffi) Ruby gem, installable via `gem install FFI`.
 
+We also support an environment variable UNIFFI_TESTS_DISABLE_EXTENSIONS; It is a set of file
```
badboy, line 63:
```suggestion
We also support an environment variable `UNIFFI_TESTS_DISABLE_EXTENSIONS`; It is a set of file
```

docs/contributing.md
```diff
@@ -60,6 +60,10 @@ test suite you will need:
...
+extensions in the format of your platform's PATH variable. Eg,
```
badboy, line 64:
```suggestion
extensions in the format of your platform's `PATH` variable. E.g.,
```

docs/contributing.md
```diff
@@ -60,6 +60,10 @@ test suite you will need:
...
+`UNIFFI_TESTS_DISABLE_EXTENSIONS=swift:rb cargo test` will skip test filenames ending in `.swift`
+or `.rb`. (On Windows, use `swift;rb`)
```
badboy, line 66: (In general I'm a [fan of semantic linebreaks](https://sembr.org/), that makes future changes much less noisy in diffs)

badboy: More on context: huh, TIL Rust ships that as a function (`split_paths`), but I wonder if we should simply standardize on something here? Not sure how helpful the difference is per OS.
Then again it's a small feature so maybe not even worth spending time over.

badboy (approved):

docs/contributing.md
mhammond: Yeah, I went around in circles here too, and just because I didn't like explaining it was different per-platform I just went with
"comma separated".

mhammond:
FTR, I also had to add `println!("cargo:rerun-if-env-changed=UNIFFI_TESTS_DISABLE_EXTENSIONS");` to our build macro as this var is considered at *build* time. It would be possible to have this var evaluated at runtime, but then we lose the nice "ignore" handling and it would look like the skipped tests ran and passed.

PR #1034, opened 2021-08-16
[Don't Merge] More wrapping

bendk:
This one depends on Mark's `wrapper-types` branch, which I think could be merged right now.

The basic idea is to define a FfiConverter class for each type we handle.  These replace various template filter functions:
  - `coerce_py`/`lower_py` -> `[FfiConverterForType].lower()`
  - `lift_py` -> `[FfiConverterForType].lift()`
  - `read_py` -> `[FfiConverterForType].read()`
  - `write_py` -> `[FfiConverterForType].write()`

I think this system is a easier to understand because most of the type-specific code is now in one place.  It could also simplifies the template generation and makes the wrapping easier.  For example we can now handle a function that directly returns an external type.

I'm hoping that this could get us to the final phase of all this type wrapping work: an extension crate that can handle the wrapping on both the scaffolding and bindings side (i.e. something that could handle `serde_json::Value`).  I'm hoping to get that working tomorrow.  Right now it's just hand-waving.


uniffi_bindgen/src/bindings/python/templates/FfiConverters.py
```diff
@@ -0,0 +1,607 @@
...
+    lift = staticmethod(float)
+    lower = staticmethod(float)
+
+    @staticmethod
+    def read(stream):
+        return stream._unpack_from(8, ">d")
+
+    @staticmethod
+    def write(builder, v):
+        builder._pack_into(8, ">d", v)
```
bendk, line 168: I considered creating a base class that handled the common parts of these, but I went with a flat hierarchy because I was a paranoid about performance.  We're going to be using these everytime we want to do anything with an argument/return value, so I figured we should keep the number of lookups to a minimum.

On the topic of performance.  One negative about this system is that we always have a `FfiConverter` call.  Before, for simple types like this we would just call `int(x)`.  We could consider adding back the lift_py filter function, but have it only handle optimizations like that.

uniffi_bindgen/src/bindings/python/templates/FfiConverters.py
```diff
@@ -0,0 +1,607 @@
...
+    @staticmethod
+    def read(stream):
+        from {{ crate_name|mod_name_py }} import {{ typ|ffi_converter_name }} as ExternalFfiConverter
+        return ExternalFfiConverter.read(stream)
+
+    @staticmethod
+    def write(builder, value):
+        from {{ crate_name|mod_name_py }} import {{ typ|ffi_converter_name }} as ExternalFfiConverter
+        ExternalFfiConverter.write(builder, value)
+
```
bendk, line 586: The final step for the JSON work is to do the conversions on the bindings side.  My idea for that is we create the concept of an "extension crate" and define a `Type::Extension` to handle that.  The scaffolding side would be pretty much the same as the current `guid` crate.  But the extension crate could also point to a python file with an `FfiConverter` class that gets inserted here.

uniffi_bindgen/src/bindings/python/templates/FfiConverters.py
mhammond: I don't really have a strong opinion here, and given we don't really rely on Python bindings, I'd be fine with a comment about the optimization. What I still can't quite picture here is what this means for the strongly-typed bindings I we should probably start focusing on (say) kotlin to make sure we aren't painting ourselves into a dynamically-typed corner.

mhammond (commented):
This LGTM - is it your intention that `External` types could just do the "from other_mod import blah" to sidestep the consuming crate knowing the foreign type?

bendk:
> This LGTM - is it your intention that `External` types could just do the "from other_mod import blah" to sidestep the consuming crate knowing the foreign type?

Exactly and there's code to do it.  It also supports lowering/lifting them directly.

PR #1035, opened 2021-08-17
WIP: Enhance external/wrapper types example and fixtures to show lack of l…

mhammond:
…ift/lower support

Just incase it helps.  As somewhat expected, tests fail with:

> thread 'test_exttypes_python' panicked at 'should not be necessary to coerce External types', uniffi_bindgen/src/bindings/python/gen_python.rs:164:38


rfk:
Is this still relevant after #1107?

mhammond:
No, I think this no longer has any value!

PR #1039, opened 2021-08-18
ADR-0006

bendk:
I rewrote this one to focus more on the general question of handling
types using template filters vs code in the target language.  I think
that's more useful going forward than the specifics of how
`FfiConverter` was implemented.


docs/adr/0006-external-types.md
```diff
@@ -1,82 +0,0 @@
-# Replace ViaFfi with FfiConverter to handle external types
```
mhammond, line 1: I'm fine with this PR, however, I do think there is space *somewhere* for a discussion about "External" types - in app-services it's actually "External" types we care about more than "Wrapped" because of [code like this](https://github.com/mozilla/application-services/blob/94baebdc8ad1295aa823c8770214e487e783092c/components/tabs/src/storage.rs#L13-L40). Not sure External types do need an ADR though, so I'm fine with this, but worth thinking about where we might find that space.

mhammond (approved):
👍

PR #1040, opened 2021-08-19
uniffi is all grown up so gets its own matrix room

mhammond:
(considered renaming "matrix" to "element" because that's what my tab title seems to be, and I'm still confused what the difference is, but wiki.m.o still says "matrix", so 🤷 )

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ If this tool sounds interesting to you, please help us develop it! You can:
 
 * View the [contributor guidelines](./docs/contributing.md).
 * File or work on [issues](https://github.com/mozilla/uniffi-rs/issues) here in GitHub.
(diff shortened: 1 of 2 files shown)


badboy:
Matrix is the network, protocol & server.
Element is the (web) client. Formerly known as Vector (briefly) and then Riot.im (longer).

badboy (approved):

mhammond:
Thanks!

> Matrix is the ..., Element is the...

Yeah - but in this context I need to choose one word ;)

mhammond:
(I guess I was trying to make a theatrical point about needing to choose one work :)

badboy:
[What is the Matrix?](https://www.youtube.com/watch?v=O5b0ZxUWNf0)

Issue #469, opened 2021-05-27
Implement Timestamp and Duration types in Ruby backend


rfk:
These new types were added in https://github.com/mozilla/uniffi-rs/pull/434, concurrently with development on the new Ruby backend. I ended up merging the Ruby backend with support for these types stubbed out, but it would be good to have them added in for completeness.

/cc @saks; no urgency here, but I wanted to file it for our future reference and for your information.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-62)
┆Issue Number: UNIFFI-62

PR #1041, opened 2021-08-24
Upgrade to a newer base docker image

saks:
This newer base image is positioned by CircleCI team as:
> This image is designed to supercede the legacy CircleCI Rust image, circleci/rust.

My original motivation was to bring a newer version of ruby into base image. I was working on #469 and found that base docker images includes ruby 2.5 which is [no longer maintated](https://www.ruby-lang.org/en/downloads/branches/).

AFAIK [all CI checks pass](https://app.circleci.com/pipelines/github/saks/uniffi-rs/57/workflows/2065fda0-5fc9-418a-b3de-13606e82ea98) when using updated docker image. And just in case you're wondering: new image is slightly smaller (4.65GB VS 4.78GB).

Please let me know if we need a changelog item for this PR.

### DIFF
--- a/docker/Dockerfile-build
+++ b/docker/Dockerfile-build
@@ -5,7 +5,7 @@
 # This builds a docker image containing all the tools we need to run our
 # test suite in CI, including rust, kotlin, and swift.
 
-FROM circleci/rust:buster
+FROM cimg/rust:1.52.1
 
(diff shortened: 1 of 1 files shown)


rfk (commented):
Thanks! This looks fine to me, although I wasn't quite sure about the management of Rust versions, hence my question below.

rfk (approved):
Thanks!

PR #1045, opened 2021-08-28
Callbacks in swift

pagr:
Add support for callbacks in swift. Basically a copy of how it is done for Kotlin, except that reference counting is only implemented for callbacks and not for classes. Im not sure if it is needed in swift.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift
@@ -66,8 +66,48 @@ let {{ callback_interface_impl }} : ForeignCallback =
         }
     }
 
-let {{ callback_internals }} = _{{ callback_internals }}()
-class _{{ callback_internals }}<T: {{ type_name }}>: CallbackInternals<T> {
+// This erasure is probably not needed and can be replaced by 
+// a better implementation of _{{ callback_internals }}
+private class {{ type_name }}Erased: {{ type_name }} {
+    
+    let parent: {{ type_name }}
+
+    init(_ parent: {{ type_name }}) {
+        self.parent = parent
+    }
+
+    {% for meth in cbi.methods() -%}
+    func {{ meth.name()|fn_name_swift }}({% call swift::arg_list_protocol(meth) %}) {% call swift::throws(meth) -%}
+    {%- match meth.return_type() -%}
+    {%- when Some with (return_type) %} -> {{ return_type|type_swift -}}
+    {%- else -%}
+    {%- endmatch %}
+    {
+        {%- if meth.arguments().len() != 0 -%}
+            {#- Calling the concrete callback object #}
+            
+            
+            parent.{{ meth.name()|fn_name_swift }}(
+                    {% for arg in meth.arguments() -%}
+                    {{ arg.name() }}: {{ arg.name() }}
+                    {%- if !loop.last %}, {% endif %}
+                    {% endfor -%}
+                )
+        {% else %}
+            parent.{{ meth.name()|fn_name_swift }}()
+        {% endif -%}
+    }
+    {% endfor %}
+}
+
+extension _{{ callback_internals }} where T == {{ type_name }}Erased {
+    func lower(_ v: {{ type_name }}) -> Handle {
+        super.lower({{ type_name }}Erased(v))
+    }
+}
+
+private let {{ callback_internals }} = _{{ callback_internals }}<{{ type_name }}Erased>()
+private class _{{ callback_internals }}<T: {{ type_name }}>: CallbackInternals<T> {
 
     init() {
         super.init(foreignCallback: {{ callback_interface_impl }})
--- a/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift
@@ -1,0 +1,81 @@
+{% let type_name = cbi.name()|class_name_swift %}
+public protocol {{ type_name }} : AnyObject {
+    {% for meth in cbi.methods() -%}
+    func {{ meth.name()|fn_name_swift }}({% call swift::arg_list_protocol(meth) %}) {% call swift::throws(meth) -%}
+    {%- match meth.return_type() -%}
+    {%- when Some with (return_type) %} -> {{ return_type|type_swift -}}
+    {%- else -%}
+    {%- endmatch %}
+    {% endfor %}
+}
+
+{% let canonical_type_name = cbi.type_().canonical_name()|class_name_swift %}
+{% let callback_internals = format!("{}Internals", canonical_type_name) -%}
+{% let callback_interface_impl = format!("{}FFI", canonical_type_name) -%}
+
+let {{ callback_interface_impl }} : ForeignCallback = 
+    { (handle: UInt64, method: Int32, args: RustBuffer) -> RustBuffer in
+
+        {% for meth in cbi.methods() -%}
+    {% let method_name = format!("invoke_{}", meth.name())|fn_name_swift %}
+
+    func {{ method_name }}(_ swiftCallbackInterface: {{ type_name }}, _ args: RustBuffer) throws -> RustBuffer {
+        defer { args.deallocate() }
+        {#- Unpacking args from the RustBuffer #}
+            {%- if meth.arguments().len() != 0 -%}
+            {#- Calling the concrete callback object #}
+            
+            let reader = Reader(data: Data(rustBuffer: args))
+            let result = swiftCallbackInterface.{{ meth.name()|fn_name_swift }}(
+                    {% for arg in meth.arguments() -%}
+                    {{ arg.name() }}: try {{ "reader"|read_swift(arg.type_()) }}
+                    {%- if !loop.last %}, {% endif %}
+                    {% endfor -%}
+                )
+            {% else %}
+            let result = swiftCallbackInterface.{{ meth.name()|fn_name_swift }}()
+            {% endif -%}
+            
+        {#- Packing up the return value into a RustBuffer #}
+                {%- match meth.return_type() -%}
+                {%- when Some with (return_type) -%}
+                let writer = Writer()
+                result.write(into: writer)
+                return RustBuffer(bytes: writer.bytes)
+                {%- else -%}
+                return RustBuffer()
+                {% endmatch -%}
+                // TODO catch errors and report them back to Rust. 
+                // https://github.com/mozilla/uniffi-rs/issues/351
+        
+    }
+    {% endfor %}
+
+        return {{ callback_internals }}.handleMap.callWithResult(handle: handle) { cb -> RustBuffer in 
+            switch method {
+                case IDX_CALLBACK_FREE: return {{ callback_internals }}.drop(handle: handle)
+                {% for meth in cbi.methods() -%}
+                {% let method_name = format!("invoke_{}", meth.name())|fn_name_swift -%}
+                case {{ loop.index }}: return try! {{ method_name }}(cb, args)
+                {% endfor %}
+                // This should never happen, because an out of bounds method index won't
+                // ever be used. Once we can catch errors, we should return an InternalError.
+                // https://github.com/mozilla/uniffi-rs/issues/351
+                default: return RustBuffer()
+            }
+        }
+    }
+
+let {{ callback_internals }} = _{{ callback_internals }}()
+class _{{ callback_internals }}<T: {{ type_name }}>: CallbackInternals<T> {
+
+    init() {
+        super.init(foreignCallback: {{ callback_interface_impl }})
+        register()
+    }
+    func register() {
+        try? rustCall { (err: UnsafeMutablePointer<RustCallStatus>) in
+            {{ cbi.ffi_init_callback().name() }}(self.foreignCallback, err)
+        }
+    }
+}
--- a/uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
@@ -14,6 +14,7 @@ class Lock {
 typealias Handle = UInt64
 class ConcurrentHandleMap<T: AnyObject> {
     private var leftMap: [Handle: T] = [:]
+    private var counter: [Handle: UInt64] = [:]
     private var rightMap: [ObjectIdentifier: Handle] = [:]
 
     private let lock = Lock()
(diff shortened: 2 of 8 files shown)


uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
```diff
@@ -0,0 +1,108 @@
...
+        objc_sync_enter(self)
+        defer { objc_sync_exit(self) }
+        return try f()
+    }
+}
+
+typealias Handle = UInt64
+class ConcurrentHandleMap<T: AnyObject> {
+    private var leftMap: [Handle: T] = [:]
+    private var counter: [Handle: UInt64] = [:]
```
pagr, line 17: Reference counting is done here for callbacks. In rust it is done in FFIObject and for all objects.

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift
```diff
@@ -0,0 +1,121 @@
...
+                // https://github.com/mozilla/uniffi-rs/issues/351
+                default: return RustBuffer()
+            }
+        }
+    }
+
+// This erasure is probably not needed and can be replaced by 
+// a better implementation of _{{ callback_internals }}
+private class {{ type_name }}Erased: {{ type_name }} {
+
```
pagr, line 72: A type erased version of callback_internals, this is needed since functions don't accept protocols with associated types or generics. There is probably a nicer way of doing this.

jhugman:
Oh this is _exciting_. Thanks for diving in.

I'm going to be looking at this either tomorrow or Tuesday.

Are there any tests to go with this? The kotlin callbacks are tested (IIRC) in `fixtures/callbacks/tests/bindings/test_callbacks.kts`, so adding a corresponding `test_callbacks.swift` and an entry in the list in `fixtures/callbacks/tests/test_generated_bindings.rs` would be the next thing if you haven't done it already.

Thank you!

pagr:
I added tests that are a direct copy of the Kotlin tests. Doing that I noticed that optional callbacks where not supported, so I added that too.

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift
```diff
@@ -0,0 +1,134 @@
...
+            {{ callback_internals }}.write(v: {{ type_name }}Erased(callback), into: writer)
+        case .none:
+            writer.writeInt(Int8(0))
+        }
+        return RustBuffer(bytes: writer.bytes)
+    }
+}
+
+private let {{ callback_internals }} = _{{ callback_internals }}<{{ type_name }}Erased>()
+private class _{{ callback_internals }}<T: {{ type_name }}>: CallbackInternals<T> {
```
jhugman, line 123: Nit: make 

```
{% let callback_internals = format!("{}Internals", canonical_type_name)|var_name_swift -%}
```

so you can do:

```suggestion
private let {{ callback_internals }} = {{ callback_internals|class_name_swift }}<{{ type_name }}Erased>()
private class {{ callback_internals|class_name_swift }}<T: {{ type_name }}>: CallbackInternals<T> {
```

In general, I'd prefer to see the variable name defined once, rather than cobble it together each time.

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift
jhugman: I wonder if this should be the class which implements `ViaFfi`, in the same way `Bool` does.

This would make passing around callback interface objects as collections, maps or options Just Work™.

I tried to get `{{ type_name }}` to provide an implementation of the protocol, but Swift stopped me in the two places I tried:

 * you can't add an extension to a protocol, because it looks like inheritance.
 * you can't add refine a private protocol.

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift
```diff
@@ -0,0 +1,134 @@
...
+    {%- match meth.return_type() -%}
+    {%- when Some with (return_type) %} -> {{ return_type|type_swift -}}
+    {%- else -%}
+    {%- endmatch %}
+    {% endfor %}
+}
+
+{% let canonical_type_name = cbi.type_().canonical_name()|class_name_swift %}
+{% let callback_internals = format!("{}Internals", canonical_type_name) -%}
+{% let callback_interface_impl = format!("{}FFI", canonical_type_name) -%}
```
jhugman, line 14: The `callback_interface_impl` is now a variable identifier rather than another class/protocol:

```suggestion
{% let callback_interface_impl = format!("{}", canonical_type_name)|var_name_swift -%}
```

uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    func delete(handle: Handle) {
+        remove(handle: handle)
+    }
+
+    @discardableResult
+    func remove(handle: Handle) -> T? {
+        lock.withLock {
+            defer { counter[handle] = (counter[handle] ?? 1) - 1 }
```
jhugman, line 58: I was confused about mention of ref counting here. 

I'm going to think outloud, without the refcounting you're doing:

1. If I pass the same callback in to Rust twice, and Rust creates a "shadow" implementations for each
2. Rust `drop`s one of the shadows, causing Swift to remove the actual callback interface
3. Rust makes a call on the other shadow, and Swift `fatalError`s?.

Would this be ameliorated if you were passing around/storing in the handlemap different instances of `{{ type_name }}Erased` holding the same callback interface ?

jhugman (commented):
This is really exciting stuff.

I've left some nits and suggestions and asked some questions in the review comments. I'd be really happy to work with you to get this landed.

Full disclosure: there are [some fairly large changes][1] landing fairly soon which will intersect with this code. This is mostly a re-arrangement of the code to make it easier for exactly these sorts of projects. Again, I'm very happy to work with you to make those changes.

Thanks so much for digging in.

[1]: https://github.com/mozilla/uniffi-rs/pull/1042

fixtures/callbacks/tests/bindings/test_callbacks.swift
```diff
@@ -0,0 +1,77 @@
...
+    import callbacks
+#endif
+
+// A bit more systematic in testing, but this time in English.
+//
+// 1. Pass in the callback as arguments.
+// Make the callback methods use multiple aruments, with a variety of types, and
+// with a variety of return types.
+let rustGetters = RustGetters()
+class KotlinGetters: ForeignGetters {
```
patrick-fitzgerald, line 15:
```suggestion
class SwiftGetters: ForeignGetters {
```

fixtures/callbacks/tests/bindings/test_callbacks.swift
```diff
@@ -0,0 +1,77 @@
...
+    assert(rustGetters.getStringOptionalCallback(callback: callback, v: "TestString", arg2: false) == "TestString")
+    assert(rustGetters.getStringOptionalCallback(callback: nil, v: "TestString", arg2: false) == nil)
+
+    // rustGetters.destroy()
+
+    // 2. Pass the callback in as a constructor argument, to be stored on the Object struct.
+    // This is crucial if we want to configure a system at startup,
+    // then use it without passing callbacks all the time.
+
+    class StoredKotlinStringifier: StoredForeignStringifier {
```
patrick-fitzgerald, line 61:
```suggestion
    class StoredSwiftStringifier: StoredForeignStringifier {
```

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift
pagr: I tried this, but didn't manage to get it working, could you post the full change required?

fixtures/callbacks/tests/bindings/test_callbacks.swift
```diff
@@ -0,0 +1,77 @@
...
+        func fromSimpleType(value: Int32) -> String { "kotlin: \(value)" }
```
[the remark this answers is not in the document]
pagr: Fixed!

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift
```diff
@@ -0,0 +1,134 @@
...
+        return RustBuffer(bytes: writer.bytes)
+    }
+}
+
+private let {{ callback_internals }} = _{{ callback_internals }}<{{ type_name }}Erased>()
+private class _{{ callback_internals }}<T: {{ type_name }}>: CallbackInternals<T> {
+
+    init() {
+        super.init(foreignCallback: {{ callback_interface_impl }})
+        register()
```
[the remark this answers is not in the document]
pagr: Thanks 😄

uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
```diff
@@ -0,0 +1,109 @@
...
+    private var counter: [Handle: UInt64] = [:]
+    private var rightMap: [ObjectIdentifier: Handle] = [:]
+
+    private let lock = Lock()
+    private var currentHandle: Handle = 0
+    private let stride: Handle = 1
+
+    func insert(obj: T) -> Handle {
+        lock.withLock {
+            let handle = rightMap[ObjectIdentifier(obj)] ?? {
```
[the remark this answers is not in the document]
pagr: Very useful in some very special cases :)

uniffi_bindgen/src/bindings/swift/templates/RustBufferTemplate.swift
```diff
@@ -14,6 +14,17 @@ fileprivate extension RustBuffer {
...
+public extension RustBuffer {
+    // Allocate a new buffer, copying the contents of a `UInt8` array.
+    static var empty: RustBuffer {
+        return RustBuffer.init(
+                capacity: 0,
+                len: 0,
+                data: nil
+            )
+    }
+}
```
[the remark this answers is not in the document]
pagr: Seems not, let's remove it. I could have sworn I used it many times...

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift
pagr: I tried this, but didn't manage to get it working, could you post the full change required?

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift
pagr: 👍

uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+}
+
+
+// Magic number for the Rust proxy to call using the same mechanism as every other method,
+// to free the callback once it's dropped by Rust.
+private let IDX_CALLBACK_FREE: Int32 = 0
+
+class CallbackInternals<CallbackInterface: AnyObject> {
```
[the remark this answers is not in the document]
pagr: Easy fix!

uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
pagr: That's really smart, let swift do the counting for us! That means we can remove the guard in insert
let handle = rightMap[ObjectIdentifier(obj)] ?? {
and that means there are no reads to rightMap so that can be removed too!
I tested a little and it seems to work!
If you think this is right I'll get rid of the counting and the right map.

pagr: Replying to the part about implementing ViaFfi on the generated callback types it seems like it might not be possible. I agree that it would be really nice to conform to the protocol, but since we generate the code its not really needed. I find solving these types of problems in swift are really hard since they explore parts of the language that are not normally used. Its like solving a puzzle that might not have a solution.

pagr:
Thanks for the speedy review and having such a through look through the code! Looking forward to discussing the reference counting and implementing ViaFfi.

uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
jhugman: > I find solving these types of problems in swift are really hard since they explore parts of the language that are not normally used. Its like solving a puzzle that might not have a solution.

This is very nicely put. And we're trying to solve parallel interlocking puzzles in multiple languages. That may not have solutions. :)

jhugman (commented):
I'm coming to the conclusion that we can get quite far with something like:

```swift
private class {{ callback_holder_type }} {
    let parent: {{ type_name }}

    init(_ parent: {{ type_name }}) {
        self.parent = parent
    }
}

extension {{ callback_holder_type }}: ViaFfiUsingByteBuffer {
    fileprivate static func read(from buf: Reader) throws -> Self {
        return try {{ callback_internals }}.read(from: buf) as! Self
    }

    fileprivate func write(into buf: Writer) {
        {{ callback_internals }}.write(v: self, into: buf)
    }
}

// we can implement some structural types here, but not recursively.
extension Optional where Wrapped == {{ type_name }} {
    func lower() -> RustBuffer {
        switch self {
        case .some(let callback):
            return {{ callback_holder_type }}(callback).lower()
        case .none:
            return (nil as {{ callback_holder_type }}?).lower()
        }
    }
```

You have to tweak how `lower` is called (I hacked up `lower(cb: …)` to be different to `lower(…)`.

```swift
private let {{ callback_internals }} = _{{ callback_internals|class_name_swift }}()
private class _{{ callback_internals|class_name_swift }}: CallbackInterfaceInternals<{{ callback_holder_type }}> {

    init() {
        super.init(foreignCallback: {{ callback_interface_impl }})
        register()
    }

    func lower(cb callback: {{ type_name }}) -> UInt64 {
        return self.lower({{ callback_holder_type }}(callback))
    }

    func register() {
        try? rustCall { (err: UnsafeMutablePointer<RustCallStatus>) in
            {{ cbi.ffi_init_callback().name() }}(self.foreignCallback, err)
        }
    }
}
```

This isn't the right thing long term: structural types don't work as they might expect to; we  special cased `T?`, and could even do `List<T>` and `Dictionary<String, T>`, but we couldn't nest them arbitrarily.

To make this work long-term, I think some refactor around `write` is needed:

 * the assumption that a `Serializable` or `ViaFfi` extension can be added to a type doesn't hold in this case.
 * we need a filter methods for `write_swift`, and get all the templates using them.
 * to do this refactor we need to make the backend more like the Kotlin backend: by iterating through known types and generating internals-like objects for every concrete type used.

What do you say to getting this in a shape to land (rebase against `main`, code cleanup/tidy, perhaps some more tests, some docs, release notes etc), then afterwards we can think about this extra `write_swift` filter method?

PR #1046, opened 2021-08-31
Wrapped/External support on Kotlin

bendk:
This is based off the `unit-of-code` branch.  The changes should be a bit clearer once that's merged.

The basic idea here is the same on python:
  - Define a `FFIConverter` interface that knows how to lift/lower/read/write Kotlin types
  - Have a well-defined system to map `Type` -> `FFIConverter`
  - Handle wrapping by defining a new `FFIConverter` that wraps an existing `FFIConverter`
  - Handle external types by importing the `FFIConverter` from the external package.

One big piece that's still TODO is figure out how this will all be built together in the real-world.  Right now the code supports testing, but I think it probably has to change to support the actual use-cases.


uniffi_bindgen/src/bindings/backend.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// A representation of the given literal for this type.
+    /// N.B. `Literal` is aliased from `interface::Literal`, so may not be whole suited to this task.
+    fn literal(&self, oracle: &dyn CodeOracle, literal: &Literal) -> String;
+
+    /// An expression for lowering a value into something we can pass over the FFI.
+    ///
+    /// N.B. This should align with the `helper_code` generated by this `CodeType`.
+    fn lower(&self, _oracle: &dyn CodeOracle, _nm: &dyn fmt::Display) -> String {
+        panic!("lower not implemented")
+    }
```
bendk, line 74: I added default implementations, because we don't need to implement these for Kotlin anymore.  This makes me want to split up `CodeType` into a language-agnostic version and language-specific version.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/external.rs
```diff
@@ -0,0 +1,59 @@
...
+        ])
+    }
+
+    fn definition_code(&self, oracle: &dyn CodeOracle) -> Option<String> {
+        // Generate a type alias so that consumers can import the external type from this package
+        let crate_name = oracle.mod_name(&self.crate_name);
+        Some(
+            format!("@kotlin.ExperimentalUnsignedTypes\ntypealias {} = uniffi.{}.{}\n",
+                self.name, crate_name, self.name))
+    }
```
bendk, line 58: Part of the motivation for all of this was to test how well the `unit-of-code` and `FFIConverter` patterns would to implement these things.  This makes me feel like both were successful.  I like how you can just define a bunch of imports and the generic code takes care of de-duping them.

bendk:
It's amazing how we can use external/imported types without accessing the external UDL file.  We don't know the kind of type, it could be `Object`, `Record`, `Enum`, or even `CallbackInterface`, or potentially `Error`.  We don't know if it's `Disposable` or not.  But it still more-or-less works.

That said, I think we will want to access that UDL file eventually to support some more cases.  For example right now I don't think you can have combine `External` and `Wrapped`, since `External` assumes the FFI type is `RustBuffer`.  I think rfk mentioned moving the UDL file to the crate root.  That's seeming like a good idea to me right now.

fixtures/ext-types/guid/src/lib.rs
```diff
@@ -24,15 +24,15 @@ fn get_guid_helper(vals: Option<GuidHelper>) -> GuidHelper {
...
+        Some(values) => values,
     }
 }
 
 impl UniffiCustomTypeWrapper for Guid {
     type Wrapped = String;
 
-    fn wrap(val: Self::Wrapped) -> uniffi::Result<Self> {
-        Ok(Guid(val))
+    fn wrap(values: Self::Wrapped) -> uniffi::Result<Self> {
```
mhammond, line 34: nit, but seeing you changed the name, the singular `value` seems more appropriate here?

fixtures/ext-types/lib/tests/test_generated_bindings.rs
```diff
@@ -27,3 +27,21 @@ fn test_exttypes_python() -> uniffi::deps::anyhow::Result<()> {
 
     uniffi::testing::run_foreign_language_testcase(&pkg_dir, &udl_files, &test_file_path)
 }
+
+#[test]
```
mhammond, line 31: ouch - I forgot about this re 4a5e7716deba31ac80823e8c3fc9d9a8dbf8ff58 - we should be very close to allowing something like `uniffi_macros::build_foreign_language_testcases!` here (either new macro for the array of udls, or maybe use arrays everywhere?) Need not block on that though

mhammond (commented):
Thanks heaps Ben! I love where this is heading, but I'm struggling to get my head around the entirety of it. As I mention below, I think I'd find it very helpful if we can work out how to land c39cc5ae66fb03310f0effddef7444ec097c0330 and related commits in a different PR - in short, I'm struggling to work out which parts of this PR I can rubber-stamp/defer-to-james/etc, vs which parts I want to understand more deeply. If that's more work than I think, then feel free to push back, but otherwise I think it will take me some time to get my head around all the parts here.

In particular, and although I've spent very little time trying to unpack it, I don't yet understand the fixture's uniffi.toml changes. I think it would be extremely helpful for examples/wrapper-types to grow a trivial kotlin test - like the Python one, just using the raw types.

I'll look more over the next couple of days, but I do love where this is going.

bendk:
> Thanks heaps Ben! I love where this is heading, but I'm struggling to get my head around the entirety of it. As I mention below, I think I'd find it very helpful if we can work out how to land [c39cc5a](https://github.com/mozilla/uniffi-rs/commit/c39cc5ae66fb03310f0effddef7444ec097c0330) and related commits in a different PR - in short, I'm struggling to work out which parts of this PR I can rubber-stamp/defer-to-james/etc, vs which parts I want to understand more deeply. If that's more work than I think, then feel free to push back, but otherwise I think it will take me some time to get my head around all the parts here.
> 
> In particular, and although I've spent very little time trying to unpack it, I don't yet understand the fixture's uniffi.toml changes. I think it would be extremely helpful for examples/wrapper-types to grow a trivial kotlin test - like the Python one, just using the raw types.
> 
> I'll look more over the next couple of days, but I do love where this is going.

Yeah, the branches are getting confusing -- I thought the logins rework was going to be the hard one.  Here's my take:

- I think this all depends #993 landing in main.  I was hoping that would be done by the time someone reviewed it, but maybe we should wait to review this one until that lands.
- I could definitely put  c39cc5a in a separate PR.  I would probably refactor so that the remaining changes are just about the wrapper/external types.  I put these together, because  I was thinking that the wrapper/external types provide evidence that the refactor in c39cc5a is worth the effort.  But maybe we can just agree that it's good to organize the code around the `FFIConverter` pattern.  I think that ADR provides some motivation at least.

jhugman:
I've now landed #993 . Hopefully this should make this PR a good deal easier to review.

FWIW: I agree with both you and @mhammond : let's land https://github.com/mozilla/uniffi-rs/commit/c39cc5ae66fb03310f0effddef7444ec097c0330 as a different PR, probably after this one lands.

BTW: love the branch name, [wrapper-delight](https://github.com/bendk/uniffi-rs/tree/wrappers-delight); @linacambridge would be proud of you.

bendk:
Splitting up sounds good.  I just made #1053 for the first part.  I think this one can be closed now.

PR #1049, opened 2021-09-03
Improve error reporting when an object impl isn't Sync+Send.

### LINKED ISSUE #1048: Question regarding the usage of Arc and compiler errors
Hey! I'm trying to expose existing Rust code to Swift using uniffi. The Rust code uses `rusqlite` to access a SQLite database. The connection to the database is kept in a [`Connection` struct](https://docs.rs/rusqlite/0.25.3/rusqlite/struct.Connection.html#impl-Sync), which is `Send` but not `Sync`. I therefore constructed the following interface (simplified names):

``` Rust
struct Database {
  connection: Connection
}

struct Engine {
  database: Database
}

struct SwiftInterface {
  engine: Arc<RwLock<Engine>>
}

impl SwiftInterface {
  ...
  fn query(&self, query: String) -> Result<Arc<Item>> {
    ..
  }
  ...
}
```

UDL:
``` udl
interface SwiftInterface {
  [Throws=SomeError]
  Item query(string query)
}
```

All access to the database always goes through the `Arc<RwLock<Engine>>` in the `SwiftInterface`. However, compiling fails with:

```
the trait bound `Arc<SwiftInterface>: FfiConverter` is not satisfied
```

When I remove the `connection` it compiles fine (e.g.):

```
struct Database {
  // connection: Connection
}
```

On the other hand, wrapping the `connection` into an `Arc<RwLock<Connection>>` doesn't work either. Am I doing something wrong? I thought by explicitly gatekeeping all access to my Rust entities through a single `Arc<RwLock<Engine>>` the safety guarantees should be held up?

Thanks in advance for any help on this matter.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-88)
┆Issue Number: UNIFFI-88

rfk:
We require the implementation of all object structs to be Sync+Send,
and the generated Rust scaffolding will fail to compile if this
requirement is not met. Unfortunately, the failure is currently
triggered via some trait bounds on `FfiConverter` for `Arc<T>`,
so the error message it produces is quite inscrutable and doesn't
help the user debug the underlying issue.

This commit adds an explicit static assertion about the Sync+Send
requirement, which helps the Rust compiler to emit the proper diagnostic
about why the underlying struct is not `Sync` or `Send`. It also has
the added benefit of guarding against any future refactor of the
code that might accidentally factor away the Sync+Send requirement.

Fixes #1048. (Well, not really *fixes* I guess because the code in
experience of debugging such cases much better).

### DIFF
--- a/examples/sprites/src/lib.rs
+++ b/examples/sprites/src/lib.rs
@@ -31,8 +31,7 @@ pub fn translate(p: &Point, v: Vector) -> Point {
 // and which can move about over time.
 #[derive(Debug)]
 pub struct Sprite {
-    // All interfaces must be `Send+Sync`, and our naive `Point` struct
-    // isn't, so we wrap it in a `RwLock` to make it so.
+    // We must use interior mutability for managing mutable state, hence the `RwLock`.
     current_position: RwLock<Point>,
 }
 
--- a/fixtures/uitests/tests/ui/interface_not_sync_and_send.stderr
+++ b/fixtures/uitests/tests/ui/interface_not_sync_and_send.stderr
@@ -1,0 +1,32 @@
+error[E0277]: `Cell<u32>` cannot be shared between threads safely
+   --> $DIR/counter.uniffi.rs:119:1
+    |
+119 | uniffi::deps::static_assertions::assert_impl_all!(Counter: Sync, Send);
+    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    | |
+    | `Cell<u32>` cannot be shared between threads safely
+    | required by this bound in `assert_impl_all`
+    |
+    = help: within `Counter`, the trait `Sync` is not implemented for `Cell<u32>`
+    = note: required because it appears within the type `Counter`
+    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
+
+error[E0277]: the trait bound `Arc<Counter>: FfiConverter` is not satisfied
+   --> $DIR/counter.uniffi.rs:145:9
+    |
+145 |         <std::sync::Arc<Counter> as uniffi::FfiConverter>::lower(_arc)
+    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `FfiConverter` is not implemented for `Arc<Counter>`
+    |
+    = help: the following implementations were found:
+              <Arc<T> as FfiConverter>
+    = note: required by `lower`
+
+error[E0277]: the trait bound `Arc<Counter>: FfiConverter` is not satisfied
+   --> $DIR/counter.uniffi.rs:160:14
+    |
+160 |             &<std::sync::Arc<Counter> as uniffi::FfiConverter>::try_lift(ptr).unwrap(),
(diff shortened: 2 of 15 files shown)


examples/sprites/src/lib.rs
```diff
@@ -31,8 +31,7 @@ pub fn translate(p: &Point, v: Vector) -> Point {
 // and which can move about over time.
 #[derive(Debug)]
 pub struct Sprite {
-    // All interfaces must be `Send+Sync`, and our naive `Point` struct
-    // isn't, so we wrap it in a `RwLock` to make it so.
+    // We must use interior mutability for managing mutable state, hence the `RwLock`.
```
rfk, line 34: Some of our comments here were misleading - you can't actually use an `RwLock` to make something `Sync` if it isn't already `Sync`, as the user in #1048 found out. The thing that the `RwLock` is actually giving us in this case is *mutability* and I've changed the comments to reflect that.

fixtures/uitests/tests/ui/interface_not_sync_and_send.stderr
```diff
@@ -0,0 +1,32 @@
...
+    |
+119 | uniffi::deps::static_assertions::assert_impl_all!(Counter: Sync, Send);
+    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    | |
+    | `Cell<u32>` cannot be shared between threads safely
+    | required by this bound in `assert_impl_all`
+    |
+    = help: within `Counter`, the trait `Sync` is not implemented for `Cell<u32>`
+    = note: required because it appears within the type `Counter`
+    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```
rfk, line 12: This is the error produced by the new assertion, which I think does a good job of helping the user find the underlying problem.

fixtures/uitests/tests/ui/interface_not_sync_and_send.stderr
```diff
@@ -0,0 +1,32 @@
...
+
+error[E0277]: the trait bound `Arc<Counter>: FfiConverter` is not satisfied
+   --> $DIR/counter.uniffi.rs:145:9
+    |
+145 |         <std::sync::Arc<Counter> as uniffi::FfiConverter>::lower(_arc)
+    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `FfiConverter` is not implemented for `Arc<Counter>`
```
rfk, line 18: This is the existing error, which is pretty confusing if you don't know what it's about.

In my ideal world, `trybuild` would allow us to assert that the error output matches the expected error while ignoring the additional ones, but it's not really built for that.

rfk (commented):
Most of the code here is about testing that we generate a useful diagnostic, and I've added some comments on it inline.

I'm lightly concerned that these tests will be fragile to changes in the output from the Rust compiler, but if we intend to aim for improved error output over time, it seems worth trying to have some test coverage for it.

examples/sprites/src/lib.rs
mhammond: yeah, that's very interesting and took me by surprise!

mhammond (approved):
This is great, and typically thorough! Thanks!

PR #1050, opened 2021-09-03
Have build_foreign_language_testcases take an array of UDL filenames

mhammond:
In 4a5e771 we added support for the test infrastructure to take an array of UDL names, but #1046 reminded me that there's no macro support for that.

I made a fairly large assumption that this macro isn't widely used outside of the crate, so thought I'd just modify the single macro and change all consumers - but this is still technically a breaking change. I'd be fine rethinking this and having 2 versions of the macro, but it's not clear to me how I would avoid cloning a fairly large chunk of code (although maybe not - I admit I didn't even try).


jhugman (approved):
This looks pretty straightforward to me. ✅

Issue #1047, opened 2021-09-01
Move templates from multi-line strings to their own files


jhugman:
#993 and #1042 moved swift and kotlin types from one giant template file to many different multi-line strings in Rust.

This issue is about moving these multi-line strings in lots of small template files, so that syntax highlighting can work again for the foreign languages.

The types with the strings are in `compounds.rs`, `primitives.rs` and `miscellany.rs`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-87)
┆Issue Number: UNIFFI-87

bendk:
[This PR](https://github.com/mozilla/uniffi-rs/pull/1052) covers Kotlin, so just focus on Swift

PR #1051, opened 2021-09-03
Decorator objects for decorating foreign language method calls in to Rust

jhugman:
This PR introduces the concept of Delegate objects, for decorating method calls into Rust.

We add three annotations:

1. `[Delegate]`, to mark an `interface` as a Delegate object.
2. `[Delegate=]`, to specify a delegate object for the annotated `interface`. Such objects have a delegate argument added to each of its constructors.
3. `[CallsWith=]`, on methods of objects with delegate objects.

Delegate objects in Kotlin are `interface`s. Methods accept the delegated object and a closure representing the Rust call.

Examples are given in the [`docs/manual/src/udl/delegates.md` file][1]: error catching, launching on background threads, notifying a UI after a call.

This PR only includes a Kotlin implementation, so is *not* ready for prime time, but otherwise ready for review.

I'd like to get more feedback before adding a Swift implementation.

Feedback sought:

* is this a concept worth pursuing?
* are the names right; do the annotations do what you'd think, do they require much explanation?
* should delegates do more? or less? e.g. do they need to be more configurable with more annotations?

[1]: https://github.com/mozilla/uniffi-rs/blob/4ed0a731acf7411bbed260cffad768f3c47b7f8f/docs/manual/src/udl/delegates.md

### DIFF
--- a/docs/manual/src/udl/decorators.md
+++ b/docs/manual/src/udl/decorators.md
@@ -1,0 +1,240 @@
+# Decorator objects
+
+We'll be considering the UDL for an object called `DemoRustObject`:
+
+```webidl
+interface DemoRustObject {
+    void do_expensive_thing();
+    [Throws=RustyError]
+    void do_crashy_thing();
+    void do_consequential_thing();
+}
+```
+
+`DemoRustObject` is a `struct` written in Rust, and uniffi generates a class of that name in the foreign languages which forward method calls to the Rust implementation.
+
+## Problem
+
+Just by the names of these methods, we can see that the application might not want to deal with the `DemoRustObject` by itself, without some proper care taken while calling its methods:
+
+* it may want to run `do_expensive_thing()` off the main thread
+* it may want to catch and report errors from `do_crashy_thing()`.
+* it may want to inform the rest of the application each time `do_consequential_thing()` is called.
+
+A common pattern when using uniffied code is to run some common, but app-specific, code before or after calling the Rust code.
+
+```kotlin
+class DemoLib(
+    val backgroundScope: CoroutineContext,
+    val errorReporter: (e: Exception) -> Unit,
+    val listener: () -> Void
+) {
+    val demoRustObject = new DemoRustObject()
+
+    fun doExpensiveThing() {
+        backgroundScope.launch {
+            demoRustObject.doExpensiveThing()
+        }
+    }
+
+    fun doCrashyThing() {
+        try {
+            demoRustObject.doCrashyThing()
+        } catch (e: Exception) {
+            errorReporter(e)
+        }
+    }
+
+    fun doConsequentialThing() {
+        demoRustObject.doConsequentialThing()
+        listener()
+    }
+}
+```
+
+This causes a proliferation of boiler plate code. Worse, everytime we add a new method to the `DemoRustObject`, handwritten foreign language code needs to be written to expose it safely.
+
+The more methods that want to do similar things, the more repetitive this gets.
+
+We could isolate the repeated code into a decorator class with `onBackgroundThread`, `withErrorReporter` and `thenNotifyListeners` methods.
+
+We'll handwrite a little decorator interface, then write an implementation that we pass the thread pool/coroutine context/whatever, and the error reporter.
+
+```kotlin
+interface DemoDecorator {
+    fun onBackgroundThread(rustCall: () -> Unit)
+    fun <T> withErrorReporter(rustCall: () -> T): T
+    fun thenNotifyListeners(obj: DemoLib, rustCall: () -> T)
+}
+
+class MyDemoDecorator(
+    val backgroundScope: CoroutineContext,
+    val errorReporter: (e: Exception) -> Unit,
+    val listener: () -> Void
+) : DemoDecorator {
+    fun onBackgroundThread(rustCall: () -> Unit) {
+        backgroundScope.launch {
+            rustCall()
+        }
+    }
+
+    fun <T> withErrorReporter(rustCall: () -> T) =
+        try {
+            rustCall()
+        } catch (e: Exception) {
+            errorReporter(e)
+        }
+
+    fun thenNotifyListeners(obj: DemoLib, rustCall: () -> T) {
+        try {
+            rustCall()
+        } finally {
+            listener()
+        }
+    }
+}
+```
+
+Then, we can re-write the `DemoLib` as:
+
+```kotlin
+class DemoLib(
+    val decorator: DemoDecorator = MyDemoDecorator()
+) {
+    val demoRustObject = new DemoRustObject()
+
+    fun doExpensiveThing() = decorator.onBackgroundThread {
+        demoRustObject.doExpensiveThing()
+    }
+
+    fun doCrashyThing() = decorator.withErrorReporter {
+        demoRustObject.doCrashyThing()
+    }
+
+    fun doConsequentialThing() = decorator.thenNotifyListeners(this) {
+        demoRustObject.doConsequentialThing()
+    }
+}
+```
+
+This is much better, but it's still looking a bit cut and pasty. Enter decorator objects.
+
+## Solution
+
+```webidl
+namespace demo {}
+
+[Decorator]
+interface DemoDecorator {
+    Any? with_error_reporter();
+    void on_background_thread();
+    Any  then_notify_listeners();
+}
+
+[Decorator=DemoDecorator]
+interface DemoRustObject {
+    [CallsWith=on_background_thread]
+    void do_expensive_thing();
+
+    [Throws=RustyError, CallsWith=with_error_reporter]
+    void do_crashy_thing();
+
+    [CallsWith=then_notify_listeners]
+    void do_consequential_thing();
+}
+```
+
+With this UDL: the `[Decorator]` annotation declares an `DemoDecorator` as a decorator object. Decorator objects never cross the FFI.
+
+> Decorator objects take their name from [Python's decorator methods][py-decorators].
+>
+> In Pythonic terms, Uniffi's decorator objects are a collection of decorator functions.
+>
+> Swift and Kotlin don't provide functionality to capture arbitrary `*args` and call a function with those same `*args`, so
+> at this time, decorator objects aren't as powerful as Python decorators. Nevertheless, they can be still quite useful.
+
+[py-decorators]: https://www.python.org/dev/peps/pep-0318/#on-the-name-decorator
+
+They are implemented as a `protocol` in Swift, and `interface` in Kotlin.
+
+They declare zero argument methods, but can return an arbitrary concrete type, `void` or the generic `Any` type. The `DemoDecorator` interface is now generated,
+and each method accepts the decorated object and a generic closure which will call the Rust code.
+
+```kotlin
+// generated by uniffi
+interface DemoDecorator<ObjectType> {
+    fun <ReturnType> onBackgroundThread(obj: ObjectType, rustCall: () -> ReturnType)
+    fun <ReturnType> withErrorReporter(obj: ObjectType, rustCall: () -> ReturnType): ReturnType?
+    fun <ReturnType> thenNotifyListeners(obj: ObjectType, rustCall: () -> ReturnType)
+}
+```
+
+The `[Decorator=DemoDecorator]` annotation above the `DemoRustObject` `interface` declaration ties the decorator object to our original `DemoRustObject` class.
+
+This changes the generated API of `DemoRustObject`:
+
+* it adds a `DemoDecorator` argument to every constructor.
+* it changes the return types and throws types for each method called with a decorator method to match the decorator method. Where the decorator method returns `Any`, the original return type remains.
+
+Now the generated Rust calls go through the app-specific decorator methods, we could more safely hand the rust object to the application to use directly.
+
+```kotlin
+val decorator = MyDemoDecorator(backgroundScope, errorReporter, listener)
+val demoRustObject = DemoRustObject(decorator)
+```
+
+This is a considerable improvement! Now the decorator methods can be specified by the app, and the UDL uses them. Each time the UDL changes, the foreign language bindings keeps up.
+
+### Re-writing the wrapper
+
+In our example above, we had a handwritten `DemoLib` which had all the methods of the `DemoRustObject` but did all error catching and launching on a background thread. It was also a good place to put additional handwritten code. With decorator objects, the need for `DemoLib` all but disappears.
+
+However, there may already be a considerable amount of application code pointing to it, which we don't want to change.
+
+Uniffi generates a Kotlin `interface` or Swift `protocol` for each `Object` in the UDL. In this case, `DemoRustObject` is the class, and `DemoRustObjectInterface` is the interface.
+
+#### Kotlin
+
+We can use this and [Kotlin's Interface delegation features][1] to re-write `DemoLib` so that it too keeps up and in sync with the UDL.
+
+[1]: https://kotlinlang.org/docs/delegation.html
+
+```kotlin
+class DemoLib private constructor(
+    demoRustObject: DemoRustObject
+) : DemoRustObjectInterface by demoRustObject {
+
+    constructor(decorator: DemoDecorator) = this(DemoRustObject(decorator))
+
+    constructor(
+        backgroundScope: CoroutineContext,
+        errorReporter: (e: Exception) -> Unit,
+        listener: () -> Unit
+    ) = this(MyDemoDecorator(backgroundScope, errorReporter, listener))
+}
+```
+
+We can then add hand-written code to this class.
+
+#### Swift
+
+In Swift, we can't get `DemoLib` to implement `DemoRustObjectProtocol`, and automatically implement the methods with `DemoRustObject`. Instead we literally replace it with a `typealias` and add a convenience constructor to take the original arguments. Handwritten code can then be added to this extension.
+
+```swift
+typealias DemoLib = DemoRustObject
+extension DemoLib {
+    convenience init(
+        backgroundQueue: OperationQueue,
+        errorReporter: @escaping (Error) -> Void,
+        listener: @escaping () -> Void
+    ) {
+        self.init(
+            MyDemoDecorator(
+                backgroundQueue: backgroundQueue,
+                errorReporter: errorReporter,
+                listener: listener
+            )
+        )
+    }
+}
+```
--- a/uniffi_bindgen/src/interface/types/mod.rs
+++ b/uniffi_bindgen/src/interface/types/mod.rs
@@ -37,6 +37,7 @@ pub(super) use resolver::{resolve_builtin_type, TypeResolver};
 /// of their internal structure apart from what type of thing they are (record, enum, etc).
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
 pub enum Type {
+    Generic,
     // Primitive types.
     UInt8,
     Int8,
@@ -58,6 +59,7 @@ pub enum Type {
     Enum(String),
     Error(String),
     CallbackInterface(String),
+    DecoratorObject(String),
     // Structurally recursive types.
     Optional(Box<Type>),
     Sequence(Box<Type>),
@@ -113,6 +115,16 @@ impl Type {
             Type::Map(t) => format!("Map{}", t.canonical_name()),
             // A type that exists externally.
             Type::External { name, .. } | Type::Wrapped { name, .. } => format!("Type{}", name),
+            Type::DecoratorObject(nm) => format!("Decorator{}", nm),
+            Type::Generic => "GenericAny".into(),
+        }
+    }
+
+    pub fn is_generic(&self) -> bool {
+        match self {
+            Type::Generic => true,
+            Type::Optional(t) | Type::Sequence(t) | Type::Map(t) => t.is_generic(),
+            _ => false,
         }
     }
 }
@@ -156,7 +168,31 @@ impl From<&Type> for FFIType {
             | Type::Duration
             | Type::External { .. } => FFIType::RustBuffer,
             Type::Wrapped { prim, .. } => FFIType::from(prim.as_ref()),
+            Type::DecoratorObject(_) => {
+                unreachable!("Decorator objects should never cross the FFI")
+            }
+            Type::Generic => unreachable!("Generic types should never cross the FFI"),
         }
+    }
+}
+
+pub(crate) fn rewrite_generic(l: &Type, r: &Type) -> Type {
+    match l {
+        // Generic gets replaced with the right hand side
+        Type::Generic => r.to_owned(),
+
+        // Special case T? when T is itself optional, to avoid String??
+        Type::Optional(t) if matches!(**t, Type::Generic) && matches!(r, Type::Optional { .. }) => {
+            rewrite_generic(t, r)
+        }
+
+        // Structural types
+        Type::Optional(t) => Type::Optional(Box::new(rewrite_generic(t, r))),
+        Type::Sequence(t) => Type::Sequence(Box::new(rewrite_generic(t, r))),
+        Type::Map(t) => Type::Map(Box::new(rewrite_generic(t, r))),
+
+        // Nothing else is re-written.
+        _ => l.to_owned(),
     }
 }
 
@@ -226,7 +262,7 @@ impl TypeUniverse {
     /// methods during the type resolution process.
     pub fn add_known_type(&mut self, type_: Type) -> Result<Type> {
         // Types are more likely to already be known than not, so avoid unnecessary cloning.
-        if !self.all_known_types.contains(&type_) {
+        if !self.all_known_types.contains(&type_) && !type_.is_generic() {
             self.all_known_types.insert(type_.clone());
         }
         Ok(type_)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,13 @@
 
 [All changes in v0.14.1](https://github.com/mozilla/uniffi-rs/compare/v0.14.0...v0.14.1).
 
+### What's Changed
+
+- [Decorator objects](docs/manual/src/udl/decorators.md) reduce boiler plate in Foreign Language bindings.
+  - `[Decorator]` declares an `interface` of decorator functions
+  - `[Decorator=]` adds a decorator to an `interface`
+  - `[CallsWith=]` directs method calls through a decorator function.
+
 ## v0.14.0 (_2021-08-17_)
 
 [All changes in v0.14.0](https://github.com/mozilla/uniffi-rs/compare/v0.13.1...v0.14.0).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,15 @@
   constructor, you must add one explicitly.
 
 ### What's Changed
+
 - Kotlin and Swift, like Python, now support [simple "wrapped" types](https://mozilla.github.io/uniffi-rs/udl/ext_types_wrapped.html).
+
+### What's New
+
+- [Decorator objects](docs/manual/src/udl/decorators.md) reduce boiler plate in Foreign Language bindings.
+  - `[Decorator]` declares an `interface` of decorator functions
+  - `[Decorator=]` adds a decorator to an `interface`
+  - `[CallsWith=]` directs method calls through a decorator function.
 
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.14.1...main).
 
@@ -33,13 +41,6 @@
 - Kotlin: Added some defensive programming around `RustBufferBuilder.discard()`
 
 [All changes in v0.14.1](https://github.com/mozilla/uniffi-rs/compare/v0.14.0...v0.14.1).
-
-### What's Changed
-
-- [Decorator objects](docs/manual/src/udl/decorators.md) reduce boiler plate in Foreign Language bindings.
-  - `[Decorator]` declares an `interface` of decorator functions
-  - `[Decorator=]` adds a decorator to an `interface`
-  - `[CallsWith=]` directs method calls through a decorator function.
 
 ## v0.14.0 (_2021-08-17_)
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
 
   "fixtures/coverall",
   "fixtures/callbacks",
+  "fixtures/decorators",
 
   "fixtures/ext-types/guid",
   "fixtures/ext-types/uniffi-one",
--- a/examples/todolist/src/lib.rs
+++ b/examples/todolist/src/lib.rs
@@ -27,7 +27,7 @@ pub enum TodoError {
     DuplicateTodo,
     #[error("Empty String error!: {0}")]
     EmptyString(String),
-    #[error("I am a delegated Error: {0}")]
+    #[error("I am a decorated Error: {0}")]
     DeligatedError(#[from] std::io::Error),
 }
 
--- a/fixtures/decorators/Cargo.toml
+++ b/fixtures/decorators/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "decorators"
+version = "0.14.0"
+authors = ["Firefox Sync Team <sync-team@mozilla.com>"]
+edition = "2018"
+publish = false
+
+[lib]
+crate-type = ["staticlib", "cdylib"]
+name = "uniffi_decorators"
+
+[dependencies]
+uniffi_macros = {path = "../../uniffi_macros"}
+uniffi = {path = "../../uniffi", features=["builtin-bindgen"]}
+thiserror = "1.0"
+lazy_static = "1.4"
+
+[build-dependencies]
+uniffi_build = {path = "../../uniffi_build", features=["builtin-bindgen"]}
--- a/fixtures/decorators/README.md
+++ b/fixtures/decorators/README.md
@@ -1,0 +1,11 @@
+# A "Decorators" test for uniffi components
+
+This is similar to the `decorators` example, but it's intended to be contrived and to
+ensure we get good test coverage of all possible options.
+
+It's here so it doesn't even need to make a cursory effort to be a "good"
+example; it intentionally panics, asserts params are certain values, has
+no-op methods etc. If you're trying to get your head around uniffi then the
+"examples" directory will be a much better bet.
+
+This is its own crate, because the decorator mechanism is not implemented for all backends yet.
--- a/fixtures/decorators/build.rs
+++ b/fixtures/decorators/build.rs
@@ -1,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+fn main() {
+    uniffi_build::generate_scaffolding("./src/decorators.udl").unwrap();
+}
--- a/fixtures/decorators/src/decorators.udl
+++ b/fixtures/decorators/src/decorators.udl
@@ -1,0 +1,25 @@
+namespace decorators {};
+
+[Decorator]
+interface MyDecorator {
+  Any? with_return();
+  void string_saver();
+  i32  with_counter();
+};
+
+[Decorator=MyDecorator]
+interface RustObject {
+  constructor();
+
+  [Name=from_string]
+  constructor(string string);
+
+  [CallsWith=string_saver]
+  string identity_string(string s);
+
+  [CallsWith=with_counter]
+  string get_string();
+
+  [CallsWith=with_return]
+  i32 length();
+};
--- a/fixtures/decorators/src/lib.rs
+++ b/fixtures/decorators/src/lib.rs
@@ -1,0 +1,34 @@
+use std::convert::TryInto;
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#[derive(Debug, Clone)]
+pub struct RustObject {
+    string: String,
+}
+
+impl RustObject {
+    fn new() -> Self {
+        Self { string: "".into() }
+    }
+
+    fn from_string(string: String) -> Self {
+        Self { string }
+    }
+
+    fn identity_string(&self, s: String) -> String {
+        s
+    }
+
+    fn get_string(&self) -> String {
+        self.string.clone()
+    }
+
+    fn length(&self) -> i32 {
+        self.string.len().try_into().unwrap()
+    }
+}
+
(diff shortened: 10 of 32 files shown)


uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -299,6 +304,25 @@ impl IterTypes for TypeUniverse {
     }
 }
 
+/// A type enum to encapsulate a Generic any type.
+#[derive(Clone, Debug, Hash)]
+pub enum ReturnType {
+    Concrete(Type),
+    Generic,
+    Void,
+}
```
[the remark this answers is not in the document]
jhugman: Whelp. It turns out that (I couldn't get) weedle to parse `any?`, though `Any?` and `T?` were available.

jhugman (commented):
Left a question for @rfk and @mhammond .

docs/manual/src/udl/delegates.md
```diff
@@ -0,0 +1,242 @@
...
+    fun <ReturnType> thenNotifyListeners(obj: ObjectType, rustCall: () -> ReturnType): ReturnType
+}
+```
+
+The `[Delegate=DemoDelegate]` annotation above the `DemoRustObject` `interface` declaration ties the delegate object to our original `DemoRustObject` class.
+
+This changes the generated API of `DemoRustObject`:
+
+* it adds a `DemoDelegate` argument to every constructor.
+* it changes the return types and throws types for each method called with a delegate method to match the delegate method. Where the delegate method returns `any`, the original return type remains.
```
rfk, line 182: My initial reaction to these changes is that they seem too magical, especially the extra argument to each constructor. You can't see this extra argument by looking at the Rust code, or by looking at the UDL for the construtors, you just have to know that this is what happens when you use the special `[Delegate]` annotation.

What would it look like to specify the delegate argument explicitly in each constructor definition in the UDL? Would it be worth it, to have the UDL more directly match the shape of the generated code?

docs/manual/src/udl/delegates.md
```diff
@@ -0,0 +1,242 @@
...
+
+class MyDemoDelegate(
+    val backgroundScope: CoroutineContext,
+    val errorReporter: (e: Exception) -> Unit
+) : DemoDelegate {
+    fun onBackgroundThread(rustCall: () -> Unit) {
+        backgroundScope.launch {
+            rustCall()
+        }
+    }
```
rfk, line 86: I think it would be useful to work an example of asyncifying a method call here - not just having it run on a background thread, but having it asynchronously return a value to the caller via whatever mechanism is appropriate for the foreign language in question. I suspect that handling the return types in a generic way may be challenging, but it'll definitely be interesting.

(I am assuming that we'd like to use this facility to replace [this kind of asyncification boilerplate](https://github.com/mozilla-mobile/android-components/blob/71e124a5dbafffb800168ea615fdbc54d7398f50/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt#L27) as well).

uniffi_bindgen/src/interface/types/mod.rs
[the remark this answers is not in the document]
rfk: (Just flagging that I saw this, but I haven't been able to think it through in a useful way just yet)

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -156,6 +160,7 @@ impl From<&Type> for FFIType {
             | Type::Duration
             | Type::External { .. } => FFIType::RustBuffer,
             Type::Wrapped { prim, .. } => FFIType::from(prim.as_ref()),
+            Type::DelegateObject(_) => unreachable!("Delegate objects should never cross the FFI"),
```
rfk, line 163: I was surprised that this gained a variant in the `Type` enum. I guess that's to support the synthesized delegate argument in constructors?

rfk (commented):
I've been "snacking" on a review of this over the last couple of days, I don't think I fully have my head around it yet but wanted to leave some initial feedback. I hope it doesn't come across as too negative, as I very much support your high-level goals here, but I think it's also useful for me to work through some of my initial reactions out loud.

I've mostly been looking at the docs to start with rather than the concrete implementation, comments inline below.

mhammond (commented):
Heh - so Ryan and I were looking at this at the same time! I agree with everything he said, and in particular, the high-level goals. I've also spent very little time on it so far, but thought I'd throw my initial comments up.

One thing I'm struggling with here is how much of the "delegate" (I quite liked Ryans 'decorator' suggestion, but that might be showing a Python bias) is truly part of the object being described, vs for a single consumer of the foreign side. IOW, it's not clear every single consumer of a uniffi-wrapped rust library will have the same requirements, nor that every language will even inside the same org.

This is similar to my comments on the wrapping stuff @bendk has been poking at - the .udl seems like the wrong place to specify this policy - but unlike the wrapping work, this isn't really suited to expressing in a .toml.

As a thought experiment, imagine uniffi takes over the world, and someone builds a component that's used by browsers from different companies. How could each of them have a different delegate, or even no delegate? Or even with Mozilla glasses on, how could iOS choose to use the interface directly without a delegate?

That said though, let me reiterate that I wholeheartedly support the problem you are solving here, and I haven't thought about it, or looked at this enough, to have concrete helpful suggestions, particularly for strongly-types languages. Maybe we can reduce the scope here to just handling threading concerns? Eg, my comment above about the method returning a string but also catching exceptions doesn't fit in my head, but limiting the scope to just handing async considerations might be more tractable?

Thanks for starting this, and I'll keep thinking about it.

uniffi_bindgen/src/interface/types/mod.rs
jhugman: Gaining a `Type` variant surprised me too! Once I started on error checking, it became a necessity.

e.g. this would be a bit harder to detect as an error:

```
[Decorator]
interface Object1 {}

interface Object1 {}

namespace {
  void set_not_a_decorator(Object1 obj);
}
```

docs/manual/src/udl/delegates.md
```diff
@@ -0,0 +1,242 @@
...
+
+## Problem
+
+Just by the names of these methods, we can see that the application might not want to deal with the `DemoRustObject` by itself, without some proper care taken while calling its methods.
+
+A common pattern when using uniffied code is to run some common, but app-specific, code before or after calling the Rust code.
+
+```kotlin
+class DemoLib(
+    val backgroundScope: CoroutineContext,
```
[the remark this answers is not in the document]
jhugman: > Would the other use-cases show here still be sufficient to justify this "delegate" feature?

Yes, I believe so.

This happens to be good for one particular set of async call patterns, but I think async stuff a much bigger change which will not solve the use-cases shown here.

docs/manual/src/udl/delegates.md
```diff
@@ -0,0 +1,242 @@
...
+[Delegate]
+interface DemoDelegate {
+    any  with_error_reporter();
+    void on_background_thread();
+    any then_notify_listeners();
+}
+
+[Delegate=DemoDelegate]
+interface DemoRustObject {
+    [CallWith=on_backghround_thread]
```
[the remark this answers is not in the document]
jhugman: I found I preferred the descriptive `CallsWith`, over the imperative `CallWith`.

> To zoom right out...what _is_ the UDL file, at a high level? Is it a description of the underlying Rust API that happens to be easier for us to parse? Is it a description of the API that should be exposed to foreign-language code? Is it a kind of pseudocode for how the wrapper layer is to be implemented? Is it all of these things?

Aside: I like these questions, they give me permission to communicate with more florid metaphors. I started to think about reinforced windows, stained glass and lenses which distort the world they let through, and then plastic packaging which shows off a toy but it also provides a good place to write product messages and conveniences for the shop to hang it up on a rail and get it through its inventory system, but then I realised that these possibly weren't that helpful. 

~Have you ever looked at your hands, man? I mean— really— looked at your hands?~

It's a _description_ of an underlying Rust API.
It's a _specification_ of the FFI wrapper layer to access that API.
It's a _convenience_ that allows you to forget about lots of boiler plate you would have had to written.

The foreign language code is written _against_ the FFI wrapper layer, not against the Rust API directly, or the UDL text file. Almost by definition, a Kotlin enum isn't the same as a Rust enum; but the FFI has provided the illusion that it is.

The FFI wrapper layer is itself a translation layer, and some of that translation needs help from the consuming library; which itself might need help from the embedding app. The extension mechanisms we're so far talking about is `[Wrapped]` for wrapping data-types. If we want to specialize the method calls, we could only wrap the whole uniffi'd object in a handwritten object that we presented to the app.

docs/manual/src/udl/delegates.md
jhugman: Yes, it does feel magical.

Other ways we could do: pass in one or more decorators into each constructor. Some amount of error checking would be needed to ensure that every constructor had the same decorators. This would allow the user to specify where in the arg list the decorator objects came.

jhugman:
Thanks both. 

### Naming

I can definitely get behind `Decorator`, which I'd have got from a Patterns book. The full pattern give the decorator access to the call arguments so you can transform the args before calling. This was why I shied away from that language, but happy that you're ok with it.

The `delegate` terminology features heavily in iOS and macOS development, as a route to customising or listening to events of a more general component.

The goal with this was to generate the code we're already writing by hand— [boring][1] [repetitive][2] [code][3] that we're already (well [I am][4]) getting wrong; useful work without getting into off-main-thread work.

### Async

For the async use-case, I agree with you both: I don't think I'm presenting a general solution for async uses. This isn't a long term fix for asynchronous calls.

However, I'm not sure that the various patterns (executors, reactive extensions, Promises/Futures, async/await, coroutines, actors, handlers etc etc etc) have stabilized on any of the platforms we operate, let alone Rust, for us to pick winners. For context, async/await was in 2021's WWDC, Coroutines only stabilized in Kotlin in 2018; Rx is popular on both iOS and Android, though being supplanted by Combine. This also doesn't account for community choice and other library preference which can lag for years.

As some (but not all) of these have language support, we can't even provide a `Wrapped`-style of extension that could support async/await/suspend, which spread through the codebase like `@ExperimentalUnsignedType` annotations.

### Big picture

You've both encouraged me to think about the bigger picture: 

 * when we have multiple consumers the same library— I don't see the FFI as the API to the library. Uniffi makes the FFI (and the Rust code) an implementation detail of the foreign language libraries we're delivering.
 * what is the UDL file— I think it's a specification of the FFI translation layer. UDL itself has— or should have— extension mechanisms to let either side of the layer help with that translation.
 

[1]: https://github.com/mozilla/application-services/blob/6c839a3adeaaa58b2655217aa53def14f1ca906d/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/PersistedFirefoxAccount.kt#L246-L250
[2]: https://github.com/mozilla-mobile/android-components/blob/71e124a5dbafffb800168ea615fdbc54d7398f50/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt#L112-L116
[3]: https://github.com/mozilla/application-services/blob/793e49c9418731ece36432a1fe20d47e575411d5/components/nimbus/android/src/main/java/org/mozilla/experiments/nimbus/Nimbus.kt#L457-L464
[4]: https://github.com/mozilla/application-services/pull/4217

docs/manual/src/udl/delegates.md
```diff
@@ -0,0 +1,242 @@
...
+
+This is much better, but it's still looking a bit cut and pasty. Enter delegate objects.
+
+## Solution
+
+```webidl
+namespace demo {}
+
+[Delegate]
+interface DemoDelegate {
```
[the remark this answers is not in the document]
jhugman: I don't think I do have a strong opinion either way.

docs/manual/src/udl/delegates.md
```diff
@@ -0,0 +1,242 @@
...
+```
+
+This is much better, but it's still looking a bit cut and pasty. Enter delegate objects.
+
+## Solution
+
+```webidl
+namespace demo {}
+
+[Delegate]
```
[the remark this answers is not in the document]
jhugman: I actually like `Decorator`.

bendk:
Agreed that the high-level goal of reducing boring/boiler plate code is good and worth the effort.  The discussion is making me wonder if there's another way to achieve it though.  I don't know if the `.udl` file is the right or wrong place for specifying decorators to run, but there issue with generics means that there's a practical issue at least.

Also, thanks for making the `delegates.md` file.  It's great to see the docs written with the code and even better to have examples to discuss.  On the topic of the example, it's not obvious that [this](https://github.com/mozilla/uniffi-rs/pull/1051/files#diff-26385b1c6c30d9da68b6e718db82fa2e1ad0fb0ab1a781ca4e6337feefee72daR108-R127) is more cut-and-pasty than [this](https://github.com/mozilla/uniffi-rs/pull/1051/files#diff-26385b1c6c30d9da68b6e718db82fa2e1ad0fb0ab1a781ca4e6337feefee72daR145-R158).  In the first example, you need to decorate all the function calls in kotlin, in the second example you need to decorate them all in the `.udl` file.

I do think there's a reason to prefer the second and that's functions that input or return  `DemoRustObject`.  If you decorate by hand, then you also need to unwrap each parameter and wrap each return value.  However,  @mhammond's and @jhugman's comments made me realize that this is just the wrapper pattern.  `DemoRustObject` is being wrapped by `DemoLib`.

@jhugman mentioned that, but then said "we could only wrap the whole uniffi'd object in a handwritten object that we presented to the app."  That actually seems okay to me -- it would be more-or-less [that first code snippet](https://github.com/mozilla/uniffi-rs/pull/1051/files#diff-26385b1c6c30d9da68b6e718db82fa2e1ad0fb0ab1a781ca4e6337feefee72daR108-R127).  One issue might be if you had many methods that weren't wrapped.  Writing `fun doNormalThing() { return demoRustObject.doNormalThing() }` dosen't seem fun.  However, there are ways around that:

  - On Kotlin `DemoApp` would delegate to `DemoRustObject` and only override the methods that needed decorating.
  - On Swift `DemoApp` would inherit from `DemoRustObject` and only override the methods that needed decorating.

If we implemented wrapping on the bindings side, does that solve this issue?

uniffi_bindgen/src/interface/types/mod.rs
jhugman: We also need to refer to it by type name in the foreign language code (not just lifting and lowering).

jhugman:
> Also, thanks for making the delegates.md file. It's great to see the docs written with the code and even better to have examples to discuss. 

Once again, I see the wisdom of doing this first, and get feedback _before_ writing any code.

> Writing` fun doNormalThing() { return demoRustObject.doNormalThing() }` dosen't seem fun

Yes, you're right.

I'd go as far as saying that wrting`fun doOtherThing() { return decorator.decoratingMethod { demoRustObject.doOtherThing() } }` each time I add a new method doesn't seem fun either.

bendk:
> I'd go as far as saying that wrting`fun doOtherThing() { return decorator.decoratingMethod { demoRustObject.doOtherThing() } }` each time I add a new method doesn't seem fun either.

What if we extended the concept of wrapping objects to decorating methods methods?  I'm picturing a .toml file like this:

```
[decorator.run-in-background.kotlin]
definition = """\
fun <T> runInBackground(rustCall: () -> T): Future<T> {
    ...
}
"""
return_type = "Future<{T}>"
```

This would add the definition code to the kotlin bindings code and allow methods in the UDL to be annotated with `[decorator=run-in-background]`

jhugman:
Raised #1054 to stop this PR being derailed by trying to solve the async case. While this PR may be useful for some async usages, it's not its primary goal.

Analogy: delegates is to async, as `for` loops are to recursion.

mhammond:
> However, @mhammond's and @jhugman's comments made me realize that this is just the wrapper pattern. `DemoRustObject` is being wrapped by `DemoLib`.

It took me some time to get my head around this, but it's an interesting concept. Up until now, I've really just been thinking of this "wrapping" concept applying to primitive types.

But on the *foreign* side, there's no good reason for that limitation. Why not wrap a `namespace`, `dictionary` or `interface`? At first glance, this doesn't make as much sense for the rust side, but maybe it does? Particularly for the async discussions - maybe not quite as compelling as the foreign side, but there might still be value.

(No real point to this comment other than to motivate me to step back and squint a little to try and find that uniffi-ing (hah!) concept that seems to be eluding me but, if it exists, deftly solves all these discussions)

docs/manual/src/udl/delegates.md
jhugman: I'm leaning towards removing the example around launching on a background thread, and leaving that to #1054 .

While this delegate/decorator pattern fits very well what we're doing with off the main thread work in some projects, its focus is to clean up boiler plate, not to solve the async problem. As such, talking about threading in this issue is a distraction, which has derailed the discussion.

bendk:
> The main use-case for `Any` is swallowing exceptions, right? If that wasn't a requirement, you could imagine a world where decorators could be described in toml - the signatures can be inferred.

I think I might be confused on this one.  Here's what I thought, please correct any misunderstandings:
  -  `Any` means the return type is the same as the original function
  - The other option is always changing the return value to a constant type (`void` or `i32`).
  - Other possibilities that depend on the original type, like `Deferred<T>`, are not supported

mhammond:
> I think I might be confused on this one. Here's what I thought, please correct any misunderstandings:
> 
>     * `Any` means the return type is the same as the original function

Yes, I mis-spoke there - I think I meant to say `Any?` but that's not correct either. But I do recall an example where the decorator suppressed the exception and returned null, thereby turning the declared `type` into a `type?` - and I got the impression that the decorator could actually change the result to anything - but even if that's wrong, `type` to `type?` is still a bit of a red flag for me.

But what you described is roughly what I'm after - that a decorator can't change any types, even making them nullable. With this restriction, ISTM that decorators would seem to need far less type information described because there must be an obvious mapping from the underlying interfaces (ie, could possibly be described in toml?) and that `Any` wouldn't be necessary because it can be trivially inferred.

bendk:
To me, changing the signature is a requirement for many use cases.  One of the main ones is turning a `T` into a `Deferred<T>`.  The extra cognitive overhead seems about the same as `JsonObject` which will be a different types for different consumers.  My concern with it all is flexibility.  One consumer, one might want a `Deferred<T>`, one might want a `Job`, and another might want a regular `T`.

I hope I'm not confusing the issue here too much, but I just posted an alternate proposal that moves much of the configuration to `uniffi.toml`.  My goal is to have something concrete to compare this proposal too.  I'm interested to hear people's opinions on that one.

mhammond:
> To me, changing the signature is a requirement for many use cases. One of the main ones is turning a `T` into a `Deferred<T>`.

That does make sense.

> My concern is flexibility. One consumer, one might want a `Deferred<T>`, one might want a `Job`, and another might want a regular `T`.

Yep, I'm 100% behind that - but like JsonObject, I'd much prefer something that uses a (probably different) concrete type per binding than spelling it as "Any".

tarikeshaq (commented):
Just want to drop a comment here since I did a read of the docs and might as well voice out my thoughts. A disclaimer that I don't have tons of experience with this kind of stuff, so take my thoughts with a grain of salt and feel free to ignore them 🙃 

- I do like that we get rid of tons of boilerplate, this seems like a very important issue, especially with Nimbus, where we are the owners of that code and have directly been hit by this in practice, although maybe not urgent to fix right away if this requires more thorough discussion
- The decorator pattern does seem to me like it's not language-agnostic, and my naive view of the `.udl` is that it's most closely coupled with the Rust code, so adding that syntax to it **might** throw readers off, again I could be overthinking here
- I like the pattern though, and I kind of like @bendk idea of having it live in the `uniffi.toml` (or wherever we'll have language-specific features, like we currently have the java package name etc)
- I was talking to @mhammond, and he brought a good point that sometimes our consumers (in practice, this only concerns android components so far) want low-level control over some of the boilerplating - they run a bunch of their co-routines under the same scope. I have a feeling we can still do this work and tackle that, whether the answer to that is "well then they shouldn't use this pattern" or something more practical.
- Related to the above point, we care about this a lot for Nimbus because we own lots of consumer code. But we already have a few other uniffi'ed components, is it worth asking our consumers there (people in Android components I imagine) if this would help their case? Keeping in mind that in that AS-AC relationship, we usually write and manage the `.udl`, and they review and consume it (but maybe that relationship should evolve a little now that it's easier with `uniffi`? I'm digressing)

bendk:
> Related to the above point, we care about this a lot for Nimbus because we own lots of consumer code. But we already have a few other uniffi'ed components, is it worth asking our consumers there (people in Android components I imagine) if this would help their case? Keeping in mind that in that AS-AC relationship, we usually write and manage the `.udl`, and they review and consume it (but maybe that relationship should evolve a little now that it's easier with `uniffi`? I'm digressing)

When I was working on logins, I kept thinking how nice this feature could be:
  - [In Fenix we have code to decorate all our methods with `withContext`](https://github.com/mozilla-mobile/android-components/blob/main/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/SyncableLoginsStorage.kt#L143-L146)
  - [I believe swift does something very similar with `queue.async`](https://github.com/mozilla-mobile/firefox-ios/blob/main/Storage/Rust/RustLogins.swift#L479-L507)

I think this feature would handle both cases very well.

Also, logins is littered with code that transforms our structs/enums into the a-c versions ([`toLoginEntry()` and `toEncryptedLogin()` in this line](https://github.com/mozilla-mobile/android-components/blob/main/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/SyncableLoginsStorage.kt#L231)).  These seem like perfect candidates for the type wrapping feature.

docs/manual/src/udl/decorators.md
```diff
@@ -0,0 +1,240 @@
...
+
+```webidl
+namespace demo {}
+
+[Decorator]
+interface DemoDecorator {
+    Any? with_error_reporter();
+    void on_background_thread();
+    Any  then_notify_listeners();
+}
```
bendk, line 132: I think this should be in `uniffi.toml` and not the UDL.  This gives a couple benefits:

 - It allows consumers to use different return types.  A different consumer might want `with_error_reporter` to log exceptions, but still propagate them up the stack.
 - It allows support for return types that UniFFI doesn't know about and/or types that are spelled differently in different languages.  The main example here are async types.  I completely agree that UniFFI shouldn't force users into any particular async framework, which probably means that UniFFI can't know what the return type should be.  Also, each language has slightly different types names.  It seems like the best way to handle this is to allow consumers to specify the return type in `uniffi.toml`.  One consumer could use `Deferred<T>`, another could use `Promise<T>`, etc.  I see a lot of parallels between this and wrapped types.
 - We don't need to define the `DecoratorObject` and `Generic` types.  I don't like adding 2 new types that work very differently than all the others.  This is more of a minor issue, but something about it feels off to me.

However, I think there's still some disagreement on this one and even if there wasn't we would need to figure out what the syntax would be.  In the meantime, could we defer the decision by just leaving this interface definition out?  It seems like we could implement a first-cut of the feature without this information:
  - The interface name is specified as part of the `[Decorator]` attribute
  - The method names are specified in the `[CallsWith]` attributes
  - Kotlin can infer the return types.  Python and Ruby don't need them.  Swift may or may not be able to do the inference, but this PR only targets Kotlin to start.

bendk (changes requested):
I'm torn because this code as-is would make life better for the Nimbus devs, but I think we need to think through how some of the core concepts are handled.

I think we could strike a balance by leaving out the code that parses the decorator definition from the UDL file and leaning on Kotlin's type inference instead.  I'm pretty sure this this is technically possible.  It addresses my main concerns and I also think the concerns that @mhammond raised.

The other suggestion I had was about how the decorators change the object API and if we could handle that differently.  I would love if we could come to a consensus on this one, but if we can't then I would be okay with merging the current proposal and potentially changing it down the line.  It seems like it would be easier to change how this works than to change how decorators are defined.

docs/manual/src/udl/decorators.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+The `[Decorator=DemoDecorator]` annotation above the `DemoRustObject` `interface` declaration ties the decorator object to our original `DemoRustObject` class.
+
+This changes the generated API of `DemoRustObject`:
+
+* it adds a `DemoDecorator` argument to every constructor.
+* it changes the return types and throws types for each method called with a decorator method to match the decorator method. Where the decorator method returns `Any`, the original return type remains.
+
+Now the generated Rust calls go through the app-specific decorator methods, we could more safely hand the rust object to the application to use directly.
```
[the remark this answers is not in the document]
bendk: Thinking about this more, another benefit is that it works with "getter functions", for example `PlacesApi`

jhugman:
> I'm torn because this code as-is would make life better for the Nimbus devs.

FWIW, as Mark suggested, I [re-factored `Fxaclient` in terms of a decorator][1]; I didn't go as far as I did with the Nimbus refactor but I think it serves as a reasonable second example. (`PersistedFirefoxAccount` is half size, and hand written code doesn't need to change when the UDL is changed).

There's going to be a lot of boring Swift boiler plate to write next year, and this is a way out of a lot of it.

> putting code in `uniffi.toml`…

There is [a different proposal](#1093) which does put the code in the `uniffi.toml` file but doesn't seem to work for the problem at hand— i.e. how does the embedding app customize the thread on which jobs are launched, or the error reporter; how might it help remove background threading in a testing environment? 

Between us on zoom, @bendk and I worked out we could do it, but only by relying of changing a global state. :(. Neither of us liked the sound of that.

I think there is a part to play for the `uniffi.toml` that can toggle the decorator generation on and off, but I think it would be a mistake to put code into a configuration file where it's uncompilable and untestable and where it won't get any kind of syntax highlighting when editing (recall #1047 was an important followup to be done for multiple reviewers).

Additionally I think there _is_ room for the `uniffi.toml` to change the spelling of how promises work or adding `suspend` or `async` to method signatures, but I think [the way we want async to work](#1054) will be solved by bringing together two or three concepts; for example:

* decorators so that the app developer has runtime control of thread creation, 
* `typedef`/wrapped types to change the type of `Promise`/`Deferred` etc, and 
* an `[Async]` attribute to add extra language features to function signatures…. 

But I think these interlocking features won't come together all in one go. To succeed, I think these features need to work well on their own, and unlock interesting things when combined.

As an aside, I'm confused about how Kotlin type inference is going to help here. I'm guessing that's by disallowing decorators to change the return types and error types; but again, that's doesn't appear to solve the problem.

Second aside: I quite like the idea of the `DecoratedClass` attribute, but it has a certain whiff of YAGNI about it. :-)

>  I would love if we could come to a consensus on this one, but if we can't then I would be okay with merging the current proposal and potentially changing it down the line. It seems like it would be easier to change how this works than to change how decorators are defined.

Yes, I think this is a good path forward. My suggestion would be to:

* Land it and see how useful / offensive / elegant / clunky it actually is practice. Mark it in the docs and change log as "Experimental".
* File a followup to add a switch in the `uniffi.toml` file to turn the behaviour off, either for all decorators in the UDL file, or just selected ones.
* File the `DecoratedClass` attribute as an issue, and flesh it out ready for engineering, but wait until a demand for it.
* If another proposal arises which solves the same problems then it needs to be fleshed out ready for engineering and then  implemented.

[1]: https://github.com/mozilla/application-services/pull/4595/commits/6ca5b2642dd1afcd1f6d0544b64ff2e2bcfdd556

mhammond:
> I think there is a part to play for the `uniffi.toml` that can toggle the decorator generation on and off, but I think it would be a mistake to put code into a configuration file where it's uncompilable and untestable and where it won't get any kind of syntax highlighting when editing

My main objection here is that the decorators are described in the .udl. I might still be missing something, but the semantics of what a decorator with args means really isn't clear, not what a decorator that returns anything other than `Any` or `void` means. There's also massive complexity in supporting `Any` and decorators as new types, which seems quite wrong too.

> I would love if we could come to a consensus on this one, but if we can't then I would be okay with merging the current proposal and potentially changing it down the line. 

I'm not going to try and veto that, but I think it would be a mistake - this patch is way too large to back away from. I'd much rather see a consensus and clear, significant benefits before committing to an approach.

mhammond:
Another reason I'm keen to keep it out of the .udl is one I haven't explicitly stated. The tl;dr is that IMO, udl is doing us amazing service, but ultimately is more liability than asset. I was very sad when Ryan's PR to replace udl with macros didn't turn out, and my brief look at Diplomat shows a similar strategy could yet work - so not entrenching it further seems like a good outcome to me.

bendk:
> > I would love if we could come to a consensus on this one, but if we can't then I would be okay with merging the current proposal and potentially changing it down the line.
> 
> I'm not going to try and veto that, but I think it would be a mistake - this patch is way too large to back away from. I'd much rather see a consensus and clear, significant benefits before committing to an approach.

To be clear, I was referring to my suggestion define a separate DecoratedClass rather than alter the constructor of the existing class.  I still like my suggestion, and the more I think about places the more I think we might actually need it, but it seems like that change could be done down the line.

OTOH, I really want to think through the syntax around specifying the decorator class before we merge.  There's a lot of code dedicated to it and it's adding new concepts to the type system.  I also think that it's not needed for a first cut.  I'm going to try to write a PR that takes this as a starting point shows how it would work without that code.

bendk:
I'm going to try to rephrase my concerns and also use some code to make it concrete.

My main concern with the current code is how the [decorator class is specified in the UDL](https://github.com/mozilla/uniffi-rs/pull/1051/files#diff-f89b3c2d1d5e4c052b92d6c4f9aa5f172c77f1e25ede12992f6db3f9a397c29fR127-R132).  My suggestion is that we move that specification to `uniffi.config` for a couple reasons:

- It adds a lot of complexity to the `ComponentInterface`.  It requires the `Generic` and `DecoratorObject` types as well as the `type_t_label` method.  This is quite a few lines of code and more importantly it makes the type system more complicated.  Now each type needs to think about generics and there's 2 new types that work very differently than the rest.  Also, this approach can only handle certain generics and I don't see a path forward for handling others.
- I don't think the UDL should specify the decorator return types.  I'm hoping that from the same UDL, one consumer could have `in_background_thread` return a `Deferred<T>` and another could have it return a `Promise<T>`.  One consumer could have `with_error_reporter` handle exceptions and return a `T?` and another consumer could have it propagate exceptions and return a `T`.  Just like with wrapper types, it seems like the UDL should treat these as somewhat abstract and let the config file specify the concrete type.

My other point is that we don't actually need to specify the decorator interface for a Kotlin-only first-cut.  The only information it's giving us is the return types of the decorator functions, but we can live without that.
 - In [the class definition](https://github.com/mozilla/uniffi-rs/pull/1051/files#diff-525584388b71eb2f5f328ae05791b81b99a6e01a60dc625eec335b2250f9a7b7R54-R57) we don't actually specify the return types for the function.  We just do `fun foo(...) = decoratorMethod { foo_ffi(...) }` and let Kotlin figure out the type.  I think this might not work for Swift, but in that case we could just have `uniffi.config` specify the return types.
 - We use it to generate the decorator interface, but if we tweak the system slightly, we can use the decorator class itself instead.
 - We also use it to generate the interface for the object itself, but that doesn't seem necessary to me.  It seems like the only reason is to support delegation and the only reason we need delegation is to have a constructor with a different signature than the auto-generated one.  What about making the class open and having a subclass?  If that doesn't work, what about a factory function?  It's very possible I'm missing something here though.

I agree that we should avoid code in the config file.  The only thing I want the config file to do is to point to a decorator class that the consumer defines.  If Swift needs it, that config would also specify the return types of the methods.  That's it, no need for class definitions in the config.

[Here's my current vision of how decorators should work](https://github.com/mozilla/uniffi-rs/compare/abf01fc46296174a6dfb6089b218764bbaf5c219...bendk:decorators).  This branch makes it so the decorator class is specified in the config.  It also makes the other change I recommended: rather than changing the existing class, we define a new decorated class.  It has a working Kotlin implementation and some hand-wavey descriptions of how this would work on Python/Ruby/Swift.

If there's a need to get something implemented while we discuss more, [here's my proposal for a minimal first cut](https://github.com/mozilla/uniffi-rs/compare/abf01fc46296174a6dfb6089b218764bbaf5c219...bendk:decorators-mvp).  This keeps things closer to this PR, but removes the code for `Generic`, `DecoratorObject` and `type_t_label`.  Instead, the consumer defines the decorator class in the same package as the bindings and we just use that class.

Both of these have an ugly hack to the testing code in order to support adding another source file to the Kotlin package.  This should be easy to clean up once I understand how that testing macro is working.

jhugman:
> The tl;dr is that IMO, udl is doing us amazing service, but ultimately is more liability than asset  […] - so not entrenching it further seems like a good outcome to me.

Thanks, this is useful perspective. I'm not sure if that ship has sailed or not; we've had several cracks at it so far. 

I think if we do re-write the front end as a series of proc-macros, it will be a sufficiently large change that we'd be able drop features either permanently or temporarily until we worked out to support them in the new scheme.

jhugman:
@bendk thanks for trying to square the circle here.

I'm supportive of efforts to keep this all in the code generating backends of uniffi, though I'm not optimistic for its success. 

This example is a super paired down version of what we're doing in Nimbus, but hopefully it illustrates the general case of the decorator being able to change the return type and error types.

```kotlin
// generated, these are stand-ins for the Rust calls
fun stringMaker(): String = "string"
fun boolMaker(): Boolean = true

// handwritten re-usable decorator code.
fun <T> catchAll(method: () -> T): T? {
    return try {
        method()
    } catch(e: Throwable) {
        null
    }
}

// generated decorated method. The return type is inferred at kotlinc compile time.
fun decoratedStringMaker() = catchAll(::stringMaker)
fun decoratedBoolMaker() = catchAll(::boolMaker)
```

It *will* work in Kotlin because you can omit the return type and use the `=` form, and Kotlin doesn't have checked exceptions.

```kotlin
fun decoratedStringMaker() = catchAll(::stringMaker)
```

It *will* work in Python because Python doesn't need you to specify the return type.

It *won't* work in Swift unless the the code knows about how the decorator is going to change the return types or error types.

```swift
func stringMaker() throws -> String  { "string" }
func boolMaker() throws -> Bool { true }

// handwritten decorator
func catchAll<T>(_ thunk: () throws -> T) -> T? {
    do {
        return try thunk()
    } catch {
        return nil
    }
}

// generated decorated functions must know the return types at generation time.
func decoratedStringMaker() -> String? {
    catchAll(stringMaker)
}

func decoratedBoolMaker() -> Bool? {
    catchAll(boolMaker)
}
```

For the case of Java (which we support via Kotlin), I'm unsure about checked exceptions. I think if the decorator method throws or rethrows an exception (which #1099 makes a checked exception) then it's going to need to know the exception type.

FWIW spending some time with a Swift playground, or a Kotlin scratch file is good way of prototyping solutions to these code generation problems.

jhugman:
So my proposal remains the same:

* Land it and see how useful / offensive / elegant / clunky it actually in practice. Mark it in the docs and change log as "Experimental".
* File a followup to add a switch in the `uniffi.toml` file to turn the behaviour off, either for all decorators in the UDL file, or just selected ones.
* File the `DecoratedClass` attribute as an issue, and flesh it out ready for engineering, but wait until a demand for it.
* If another proposal arises which solves the same problems then it needs to be fleshed out ready for engineering and then  implemented.

I'd add: if we replace the UDL with proc-macros, then we should not block that re-write while we work out how to support decorators.

Happy to receive actionable feedback on a path forward. I'd be super happy if a *better* solution arises in the future, but I don't think it should stop a *working* solution now.

We have a rough consensus that decorators as a concept are useful, and working code. Forward momentum and course correction.

bendk:
> Regrettably Uniffi has to know more about the decorator that just the method names, and codegen backends don't have enough logic in them to find that out themselves.

I think I'm not communicating my thoughts properly here.  I'm not focused on *what* needs to be specified, but *where* it should be specified.  I like specifying them in the config file rather than the UDL for the [reasons listed at the top of this comment](https://github.com/mozilla/uniffi-rs/pull/1051#issuecomment-969308351).  The fact that you can get away with not specifying the return types for some languages is just a side benefit.

I didn't think too hard about checked exceptions, but I think the same reasoning applies to them as to the return type.  I want 2 consumers to be able to use the same UDL file and handle the exceptions differently in their decorators.  One can propagate the checked exceptions and one can catch and handle them.

What do you think about specifying things in the config instead of the UDL?

If you have concerns with specifying things in the config, then would [my proposal where the return types are inferred](https://github.com/mozilla/uniffi-rs/compare/abf01fc46296174a6dfb6089b218764bbaf5c219...bendk:decorators-mvp) work to get started?  The goal is not to never specify the return types, it's just to get something working and merged as we discuss where that specification should live.  I think that would let you get started using the feature on Kotlin.  Python and Ruby would be fairly easy to add.  I agree that Swift wouldn't be possible though.  Is Swift a requirement to get started?

jhugman:
>  Is Swift a requirement to get started?

Yes, Swift and Kotlin is the minimum set of languages that any feature of uniffi must support.

> I think I'm not communicating my thoughts properly here. I'm not focused on what needs to be specified, but where it should be specified.  I like specifying them in the config file rather than the UDL for the reasons listed at the top of this comment.

I understand this, and I think it's a good place to start. But at some point you have to look at what it looks like to specify everything that needs to be specified in the place you want to specify it.

We seem to be at deadlock here in a cycle of:

 * Let's do A!
 * Hmm. That does look useful, but I don't like A. We should do B.
 * Hmm. I like B, but B doesn't work. A works already, we should do A.
 * Well, as I've said, I don't like A and we should look at B.
 * I don't think B will work, and you don't like A.

Let's try and break this cycle.

Before writing any more Rust on a competing proposal, I think it would a useful exercise for you to prototype a tiny sketch: a UDL file, a `uniffi.toml` for Swift and a **compilable** sketch of the Swift that the toml points to, the generated Swift and what it's like to use it. 

I think [`stringMaker`/`boolMaker`][1] example above might be a little too toy (it only has one decorator function, and it's not collected into a `protocol`) but it's about the right order of complexity: it catches exceptions, and changes the return type.

The Swift must compile, the TOML must be valid, and the UDL must be valid IDL.

That exercise will give us enough of an idea of the user experience for specifying decorators in the toml in a required language we're know we're going to find hard, test some assumptions, and show us what Rust we need to write.

Let's review it in our next 1-1, and out of this PR.

Hopefully, this will get us to a point where we can both see that B is viable or not, and if it's not, then an actionable glide path to landing A.

See you after Thanksgiving.

[1]: https://github.com/mozilla/uniffi-rs/pull/1051#issuecomment-978191931

bendk:
What I'm not understanding is why there would be extra difficulty with the TOML file compared to the UDL file.  It seems like at the very least you could use the system proposed here with `Any` and `Any?` to handle generics, but just move that to `uniffi.toml`.  What issues do you foresee happening there?

That said, I think that we could improve on `Any` and instead use something similar to the type wrapping proposal.  The config file could specify return values with `{}` replaced with the original method's return value.

Here's a stab at the `stringMaker`/`boolMaker` example.  I moved both of those functions into an interface and wrapped it with a decorator class that has `catchAll` as its only function.

- [UDL file](https://gist.github.com/bendk/9afdb7932013618ed37054b43952af3c)
- [uniffi.toml](https://gist.github.com/bendk/ce8c907118df51ac790e38322abfd397)
- [Swift code](https://swiftfiddle.com/vam4y4r2rnb2bhus3cttn5h7je)

jhugman:
Closed, for lack of a good way forward.

Issue #400, opened 2021-02-23
How to handle acronyms when normalizing names?


rfk:
Over in https://github.com/mozilla/application-services/pull/3876 I'm working on UniFFIing the fxa-client crate, and one nit that I noticed is how it handles method names like `begin_oauth_flow`. The proper render in camel-case should IMHO be `beginOAuthFlow` because `OAuth` is the correct capitalization of OAuth, but uniffi renders it as `beginOauthFlow` which I find incredibly jarring. Any ideas how we might provide more nuanced handling for cases such as this?

mhammond:
At a high-level, a new set of optional attributes would make sense - but I guess you already thought of that. Is there some reason that will not work, or are you after bike-shedding for specific names etc?

rfk:
TBH I really haven't thought about it in any detail, just wanted to get this on file to get it out of my head for now :-)

An attribute could work, specifying the preferred capitalization under various schemes. You could also add entries in the config file if you wanted to control the rendering for a specific backend.

jhugman:
Sketching to think: providing an override attribute.

```
[CamelCase=setURL]
void set_url(string url_string);

[CamelCase=getURL]
string get_url();
```

Would we need any other override attributes? Probably not, but I STR we did talk about swift formatting of method names at some point.

The other way feels like a foot gun right out the gate but I'll add it to the row of strawfoxes here: putting identifier fragments in `uniffi.toml`.

```toml
[overrides.naming]
url = URL
oauth = OAuth
```

which should act upon everything with `(^|_)oauth(_|$)` and (^|_)url(_|$): e.g. 
`begin_oauth_flow`, `get_oauth_url`.

rfk:
> identifier fragments in uniffi.toml.

Would this feel less foot-gunny if they were full names instead of fragments? For example:

```
[overrides.naming]
set_url = setURL
get_url = getURL
```

(Not saying I particularly like this, just thinking out loud)

jhugman:
Yes, that seems the obvious halfway point.

(I'm not attached to the TOML table naming btw; it was the first thing that came to mind).

PR #1052, opened 2021-09-03
UOC: macro dispatch

bendk:
This is my attempt to iterate on the UOC work. The main idea is to implement the dispatch pattern using macros.  I think this brings a few advantages:

  - No need to convert `Type` to `Box<dyn trait>`, the traits get  implemented on `Type` itself.
  - Only need to define the matching code once.  The language-specific`CodeOracle` is replaced by a macro in `interface::types`.  Also, we don't need to keep a `CodeOracle` around and pass it in to all the methods.
  - We can split up the traits into a general and languages-specific version.  I think this is hard with `dyn traits` although I could be wrong about that.
  - Extends the dispatching to types that correspond to a `Type` variants.  `Record`, `Enum`, and friends get `canonical_name()` and `ffi_type()` methods.  I'm not sure how much it matters, but it seems nice and isn't a lot of work).
- Less lines of code.  These changes result in a trivial drop, but moving swift over should result in fairly significant one.

The other main change is the name functions.  I wrote out some comments in the second changeset, but basically I think naming gets simpler if we attach a `name()` method to everything using traits.


bendk:
This one also implements #1047 for Kotlin.

mhammond:
This looks good to me, but I don't really understand the UOC work well enough to make meaningful comments, so I'm going to explicitly flag James for review, but happy to rubber stamp it if it helps anyone out!

rfk:
(I also hope to take a look at this in the next day or two, but as usual, don't feel like you need to block on me)

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/error.rs
```diff
@@ -2,89 +2,48 @@
...
-        format!("{}.write({})", oracle.var_name(nm), target)
-    }
-
-    fn lift(&self, oracle: &dyn CodeOracle, nm: &dyn fmt::Display) -> String {
-        format!("{}.lift({})", self.type_label(oracle), nm)
+    fn declare_code(&self, code_builder: CodeBuilder, ci: &ComponentInterface) -> CodeBuilder {
+        code_builder.code_block(KotlinError::new(
+            ci.get_error_definition(self.name)
+                .expect("Error definition not found")
+                .clone(),
```
rfk, line 19: How hard would it be to store a reference in `KotlinError` rather than doing a `clone` here? Naively it seems tractable, because the reference `ci` should comfortably outlive the entire rendering process.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/names.rs
```diff
@@ -0,0 +1,92 @@
+/* This Source Code Form is subject to the terms of the Mozilla Publie
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk, line 3: I have a vague hand-wavy concern about factoring this away from the `CodeOracle` and into static helper functions - we lose the ability to do any config-based adjustments to the generated names. I know we currently do not support config-based adjustments to names, but it's been in the back of my mind as something that will come up at some point (see e.g. some of the proposals in https://github.com/mozilla/uniffi-rs/issues/400).

On the other hand, it does simplify things *a lot*, so I'm open to calling YAGNI on any clever name-generating shenanigans until it does actually come up in a concrete use-case.

@jhugman what do you think?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/names.rs
```diff
@@ -0,0 +1,92 @@
...
+        None => name.to_string(),
+        Some(stripped) => {
+            let mut kt_exc_name = stripped.to_owned();
+            kt_exc_name.push_str("Exception");
+            kt_exc_name
+        }
+    }
+}
+
+// `KotlinName` defines the Kotlin name for various parts of the interface.
```
rfk, line 44: Might as well make this a three-slash doc comment :-)

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/primitives.rs
```diff
@@ -2,373 +2,158 @@
...
+    lower_expr: String,
+    read_expr: String,
+    write_expr: String,
+}
 
-    internal fun Double.Companion.read(buf: ByteBuffer): Double {
-        val v = buf.getDouble()
-        return v
+impl PrimitiveTemplate {
+    fn int(handler: &SimpleTypeHandler, get_func: &str) -> Self {
```
rfk, line 125: It's a bit jarring to see this switching back and forth between "simple type" and "primitive template", what do you think about trying to consolidate on "primitive type" as the name for these things throughout the codebase?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin.rs
```diff
@@ -71,299 +72,62 @@ impl MergeWith for Config {
...
-                let inner = *inner.to_owned();
-                Box::new(compounds::MapCodeType::new(inner, outer))
-            }
-            Type::External { .. } => panic!("no support for external types yet"),
-            Type::Wrapped { .. } => panic!("no support for wrapped types yet"),
-        }
-    }
+#[derive(Template)]
+#[template(syntax = "kt", escape = "none", path = "wrapper.kt")]
+pub struct KotlinWrapper<'a> {
```
rfk, line 107: Given the size of the change you're making here, I suggest taking the opportunity to call this something other than "wrapper" and avoid any future confusion with "wrapper types". `KotlinBindings` maybe?

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -57,12 +55,12 @@ sealed class {{ toplevel_name }}: Exception(){% if self.contains_object_referenc
...
     }
 
-    {% if self.contains_object_references() %}
+    {% if contains_object_references %}
     @Suppress("UNNECESSARY_SAFE_CALL") // codegen is much simpler if we unconditionally emit safe calls here
     override fun destroy() {
         when(this) {
             {%- for variant in e.variants() %}
-            is {{ e.name()|class_name_kt }}.{{ variant.name()|class_name_kt }} -> {
+            is {{ e.nm() }}.{{ e.variant_name(variant) }} -> {
```
rfk, line 63: Avoiding the use of filters here is a massive improvement to readability of the templates.

uniffi_bindgen/src/interface/types/codetype.rs
```diff
@@ -0,0 +1,210 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
rfk, line 3: How would you feel about moving this into the `::bindings::backend` module (or submodules thereof) rather than having it live here under `::interface`? It's code for the "output" phase rather than the "input" phase of the tool, and I feel like folks are more likely to look for such code over in `::bindings` somewhere.

Organizing the code so that all the generic helpers for writing a backend are under `::bindings::backend` also gives us a single obvious place to write high-level docs about how to build a backend - they can be in the module-level doc comments for that generic helper code.

OTOH, maybe that would make it a bit weird for the scaffolding code to import and use it. A shared `::codegen` module maybe? Somehow, it feels important enough that it shouldn't be hidden away inside `::interface` with a bunch of internal implementation details of the parsing code.

uniffi_bindgen/src/interface/types/dispatch.rs
```diff
@@ -0,0 +1,196 @@
...
+//!      their `Type` variant.
+//!
+//! This module helps solve both issues using the dispatch pattern and macros:
+//!
+//!  - Define a set of handler structs.  Each one will handle some subset of `Type` variants
+//!    (usually 1).
+//!  - Define a macro that can dispatch function calls for `Type` to one of the handler structs.
+//!  - Wrap traits with the `type_dispatch!` macro, defined here.  `type_dispatch!` derives a trait
+//!    impl for `Type`, `Record`, `Object`, `CallbackInterface`, etc. by dispatching the calls to
+//!    the appropriate handler type.  See `CodeType` and `KotlinCodeType` for an example.
```
rfk, line 22: Consider using the ``[`CodeType`]`` syntax here, so that when this doc comment gets rendered into the online docs, the reader can click a link to jump directly to the examples in question.

uniffi_bindgen/src/interface/types/dispatch.rs
```diff
@@ -0,0 +1,196 @@
...
+
+use super::Type;
+
+// Dispatch handler for simple types that correspond to standard types on the target language.
+// It's easier to implement traits for all of these together than with separate structs.
```
rfk, line 27: As mentioned above, I think we should try to pick a single name for this type of type and use it consistently, with "primitive type" being my current favourite.

uniffi_bindgen/src/interface/types/dispatch.rs
```diff
@@ -0,0 +1,196 @@
...
+    Float64,
+    Boolean,
+    String,
+}
+
+// Other variants are mapped 1-1 to structs.
+//
+// When handlers need to reference names and other data from the `Type` instance, we use the `'d`
+// lifetime.  This means that these references are only valid for the dispatched call.
+pub struct RecordTypeHandler<'d> {
```
rfk, line 47: FWIW, I didn't find the word "handler" very meaningful on my first read-through. I don't object to it, it just didn't really do anything to help me understand what was happening. I wonder if something like `RecordTypeDispatcher` work here, leaning in a little bit more to the explicit use of dispatch pattern and the name of this module.

(I'm not willing to put up even a token resistance to the name "handler" if you're happy with it, just sharing my initial impressions)

uniffi_bindgen/src/interface/types/dispatch.rs
```diff
@@ -0,0 +1,196 @@
...
+            }.$fn_name($($param),*),
+            Type::Wrapped { name, prim } => WrapperTypeHandler {
+                name,
+                wrapped: prim.as_ref(),
+            }.$fn_name($($param),*),
+        }
+    }
+);
+
+// The `type_dispatch!` trait wrapper.  F
```
rfk, line 122: Dangling `F`?

uniffi_bindgen/src/interface/types/dispatch.rs
```diff
@@ -0,0 +1,196 @@
...
+        type_dispatch_impl_for_type!($tt, $name);
+        type_dispatch_impl_for_related!(crate::interface::Record, $tt, $name);
+        type_dispatch_impl_for_related!(crate::interface::Enum, $tt, $name);
+        type_dispatch_impl_for_related!(crate::interface::CallbackInterface, $tt, $name);
+        type_dispatch_impl_for_related!(crate::interface::Error, $tt, $name);
+        type_dispatch_impl_for_related!(crate::interface::Object, $tt, $name);
+    };
+    (
+        $(#[$meta:meta])*
+        $vis:vis trait $name:ident : $super:ident $tt:tt
```
rfk, line 141: It took me a little bit of staring to understand that the only difference between these two rules was the `super` trait. Please consider adding a comment to that effect :-)

Alternately, would something like `$( : $super:ident)?` work here to handle both in a single rule? I don't have much experience with optional matching in macro rules so I'm not sure what's possible.

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -66,55 +71,12 @@ pub enum Type {
...
-            // A type that exists externally.
-            Type::External { name, .. } | Type::Wrapped { name, .. } => format!("Type{}", name),
-        }
-    }
+    // To add a new type:
+    //   - Add a variant to this enum
+    //   - In `dispatch.rs` add a type handler struct and a case in `dispatch_type_function!`
+    //   - Implement `NewCodeType` for the type handler
+    //   - For each language, implement `[Language]CodeType` for the type handler (Note: currently
+    //     only supported on Kotlin)
```
rfk, line 79: What's the failure mode if someone doesn't do this last step? Will the compile give them enough clues to figure out what they missed?

uniffi_bindgen/src/interface/types/codetype.rs
```diff
@@ -0,0 +1,210 @@
...
+
+use crate::interface::types::dispatch::*;
+use crate::interface::FFIType;
+
+type_dispatch! {
+    /// Type behavior for generated code
+    ///
+    /// This trait handles the language-agnostic behavior.  We also create a language-specific
+    /// version in the bindings/scaffolding code.
+    pub trait NewCodeType {
```
rfk, line 13: What does the "new" in `NewCodeType` signify? I wasn't able to parse the name into a sensible mental model of what it's for in a way that makes use of "new".

rfk (commented):
I've done a pass of this, reviewing first for high-level "shape" and the core ideas rather than looking too much into the specific details of each time. I'm assuming that a green CI run means nothing has gone too far off the rails in the details there :-)

Overall, I like it! It simplifies away a fair bit of boilerplate and replaces it with not *too* much magic, although I do think it would be worth trying to add some brief module-level narrative docs in order to explain the core idea. I also left some thoughts about organizing the code to try to help make the key idea here easier to find (although as always, I'm just one guy, and would welcome differing opinions on that front).

I'd like to hear James's thoughts on this before I dive in for a deeper pass, as it seems likely he'll highlight additional perspectives that I've missed.

> the traits get implemented on Type itself

I recall @jhugman having some reasons why he wanted to have at least the *option* of implementing against a language-specific wrapper type rather than directly on the `Type` itself, but I don't recall what they were (and I might have imagined or misunderstood them). James?

> Also, we don't need to keep a CodeOracle around and pass it in to all the methods.

I like how this has simplified things, but I'll admit I'm a little concerned that we'll want to pass at least the `Config` around to various rendering methods in the future. On balance, probably not concerned enough to block landing this, but we should give it some thought for future customization/extensibility.

> Extends the dispatching to types that correspond to a Type variants. Record, Enum, and friends get canonical_name()
> and ffi_type() methods. I'm not sure how much it matters, but it seems nice and isn't a lot of work).

This is pretty neat :-)

uniffi_bindgen/src/bindings/backend.rs
```diff
@@ -163,3 +168,43 @@ pub trait CodeDeclaration {
...
+            self.items.push(template.render()?);
+        }
+        Ok(())
+    }
+
+    fn calc_hash<T: 'static + Hash>(&self, template: &T) -> u64 {
+        let mut s = DefaultHasher::new();
+        // Make sure to include the type id to make things unique
+        template.hash(&mut s);
+        std::any::TypeId::of::<T>().hash(&mut s);
```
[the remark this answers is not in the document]
bendk: The main issue I was struggling with is that the templates are different structs so you can't  put them in the same `BTreeSet`.  We could use a `BTreeSet` to store the hash values though, I just figured a hash collision was not something to worry about.

This issue, and a few others, makes me wonder about an alternative template system though.  `terra` basically uses `serde_json::Value` as their context struct.  In that case we could make a struct that contained the template name + the context and directly store that in a `BTreeSet`

uniffi_bindgen/src/interface/types/dispatch.rs
bendk: I'm happy to change the name, but I'd prefer something other than dispatcher.  For me "dispatcher" is the thing that routes the method call , the macro in this case, not the thing that the call is routed to.

uniffi_bindgen/src/interface/types/dispatch.rs
bendk: I think I tried that but couldn't get it to work.  I think the issue there is that an `ident` is also a  `tt`, so it makes the rules ambiguous.

uniffi_bindgen/src/interface/types/codetype.rs
bendk: "New" just means "not the current one".  If we decide to go with this approach then the plan would be to update the swift code to use it as well, then rename `NewCodeType` -> `CodeType`.

uniffi_bindgen/src/interface/types/mod.rs
bendk: It's not the best, but not that mysterious.  The compiler tells you that you're missing a method and maybe you need to implement a trait.  `KotlinCodeType` is listed, but so are other traits.  I think it would be better if the `dispatch_type_function` macro was updated to use `<HandlerObj as $trait_name>`, but that makes the macro more complicated.  I think each match needs to become a block rather than a single expression.  Maybe it's worth it though.  What do you think?

bendk:
@rfk thanks for the review!  I agree with the documentation/code organization comments and I'll try to update the branch to reflect that.

You also bring up a bigger point about accessing `Config` in the language-specific code.  When I was writing this I went back and forth between different ways of implementing things and was never quite sure which way was the best:

- Implementing traits on `Type`. Like what's in this PR
- Convert `Type` instances to a language-specific struct before rendering.  The nice thing about this is the conversion process can get access to `Config`, `ComponentInterface`, or whatever additional context it needs.  However, I couldn't figure out a way to do this that was as simple as this branch.
- The last idea I had was leaning on filters more rather than less.  You could define a macro that created a filters module using the dispatch pattern.  I couldn't figure out a real benefit to doing this though, just mentioning it in case it sparks someone else's imagination.

So I guess it comes down how likely it for `Config` to change the codegen behavior and also how much  would change.  If `Config` only changes the names of things, then maybe we could use a hybrid solution:
  - `KotlinCodeType` loses the `nm()` method, but stays the same otherwise
  - `Type` and `FFIType` get a `name` field/method, which is initially set to the name used in the UDL file for the type
  - At the start of rendering, we do a name conversion step, where we convert the name for all the various structs in `ComponentInterface` from generic UDL names to language-specific names.   The code here would look a lot like the current `gen_kotlin/names.rs` module, but it could have access to `Config`.

rfk:
> Convert Type instances to a language-specific struct before rendering. The nice thing about this is the conversion
> process can get access to Config, ComponentInterface, or whatever additional context it needs. However, all the ways
> I could figure out to implement this were more complicated to this branch.

I hear you. Conceptually this is the approach I like the best, a kind of multi-stage compilation pipeline - take the input `ComponentInterface` and convert it into some sort of `KotlinComponentInterface` using the `Config`, then statelessly render the `KotlinComponentInterface` out to disk.

In practice, having to generate a suite of mirror structs like `KotlinComponentInterface`, `KotlinRecord`, `KotlinField`, etc etc quickly becomes more boilerplate than it's worth.

I have a vague idea about making the structs from `::interface` templatable in some way in order to reduce the amount of boilerplate, but haven't pursued it in any detail.

I don't think we should block on this, because we don't have a concrete use for it right now; rather I think we should keep an eye on it as an eventual outcome of the ongoing refactoring workstream here.

bendk:
100% agree. The approach feels right, but wouldn't provide all that much practical benefit right now.  I tried to think it through as concretely as possible and this how I think it could be done:

- Parametrize `Field`, `Argument`, `Record`, `Object`, ..., all the way up to `ComponentInterface` on `<T=Type, F=FFIType>`.
- Define a `ComponentInterfaceConverter<T, F>` trait, that each language would implement.  This trait would handle the concrete parts of the conversion process.  This trait would have a set of converter methods: 
  - `convert_type(Type) -> T` (for Kotlin `T=KotlinType`)
  - `convert_ffi_type(FFIType) -> F` (for Kotlin you could just have `F=String`)
  - `convert_field_name(String) -> String`
  - ... more name conversion functions
  - I also would want `convert_literal(Literal) -> String`, but that might add more complexity than it's worth
 - Then define a generic converter function that converts the entire CI by calling the methods of `CompontentInterfaceConverter`
- You still need to handle `KotlinType`.  We probably don't want to go back to the mega-match style.  Maybe `KotlinType` has a field like this: `impl: Box<dyn KotlinTypeImpl>`

I think that could work and it wouldn't be a huge amount of code.  But it would be a decent amount and I can't really see how that pushes any features forward.

bendk:
Finally got around to integrating @rfk's comments into this branch.  Is this mergable?

PR #1059, opened 2021-09-23
Allow dictionaries to have fields which are interfaces

### LINKED ISSUE #1058: Allowing a dictionary to have fields of an interface type in UDL
Would it be possible to enable dictionaries to have fields that are of an interface type in UDL? For example:

interface Foo {
   constructor();
   string to_string();
};

dictionary Bar {
   Foo field1;
   Foo field2;
};

The [documentation](https://mozilla.github.io/uniffi-rs/udl/structs.html) actually does have an example of this but currently trying to implement such a UDL gives the following error: `Failed to parse UDL: Objects cannot currently appear in record fields` which is thrown [here](https://github.com/mozilla/uniffi-rs/blob/b5e7836dcc429a43a5a7a72879a985c6bddfa2f9/uniffi_bindgen/src/interface/record.rs#L135).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-92)
┆Issue Number: UNIFFI-92

mhammond:
As noted in #1058, our [documentation](https://mozilla.github.io/uniffi-rs/udl/structs.html) promises dictionaries can have fields which are interfaces, so long as they are behind an `Arc`. It turns out this works fine in the scaffolding and even in all the bindings [edit: removed references to "bindings generate invalid code" - turns out I used a kotlin keyword - that's now supported!]

Fixes #1058

### DIFF
--- a/uniffi_bindgen/src/interface/record.rs
+++ b/uniffi_bindgen/src/interface/record.rs
@@ -129,9 +129,6 @@ impl APIConverter<Field> for weedle::dictionary::DictionaryMember<'_> {
             bail!("dictionary member attributes are not supported yet");
         }
         let type_ = ci.resolve_type_expression(&self.type_)?;
-        if let Type::Object(_) = type_ {
-            bail!("Objects cannot currently appear in record fields");
-        }
         let default = match self.default {
             None => None,
             Some(v) => Some(convert_default_value(&v.value, &type_)?),
--- a/fixtures/coverall/src/coverall.udl
+++ b/fixtures/coverall/src/coverall.udl
@@ -115,6 +115,21 @@ interface Coveralls {
 
     /// integer->integer dictionary
     record<u32, u64> get_dict3(u32 key, u64 value);
+
+    /// Adds a new repair at the current time.
+    void add_patch(Patch patch);
+
+    /// Adds a new repair at the specified time.
+    void add_repair(Repair repair);
+
+    /// Returns all repairs made.
+    sequence<Repair> get_repairs();
+};
+
+// coveralls keep track of their repairs (an interface in a dict)
+dictionary Repair {
+    timestamp when;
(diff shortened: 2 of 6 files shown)


uniffi_bindgen/src/interface/record.rs
```diff
@@ -131,9 +131,6 @@ impl APIConverter<Field> for weedle::dictionary::DictionaryMember<'_> {
             bail!("dictionary member attributes are not supported yet");
         }
         let type_ = ci.resolve_type_expression(&self.type_)?;
-        if let Type::Object(_) = type_ {
-            bail!("Objects cannot currently appear in record fields");
-        }
```
jhugman, line 136: Does this mean you just need to write tests in Kotlin Swift and Ruby to prove to yourself that it works as documented?

mhammond: > Does this mean you just need to write tests in Kotlin Swift and Ruby to prove to yourself that it works as documented?

I suspect the bindings themselves might also need work - eg, CI is showing that the kotlin code being generated here has a syntax error.

mhammond:
On top of  #1237, but is looking good. Needs kotlin support/tests.

uniffi_bindgen/src/interface/record.rs
mhammond: I haven't done ruby - I think python, swift and kotlin not needing any changes demonstrates it works ok.

bendk (approved):
This seems good to me.  At some point, I think that we should try to think more about the type system overall (#1254), but I don't see any issue with including an interface in a record.  We'd need to free the object when the record is destroyed, but strings have the same issue.

mhammond:
@jhugman last call for your feedback :)

PR #1061, opened 2021-09-28
Re-write read and write Swift template calls to go through filters

jhugman:
Re-write read and write Swift template calls to go through filters:

* needed to make wrapped types possible in Swift
* needed to land the Callback Interfaces (#1045)

This necessitated moving the structural types to where they are unrolled
(as seen in Kotlin). This, in turn, meant making an `FfiConverter` enum for
each concrete implementation of the structural type declared in the UDL.

This is could also be seen as part of #1047, but not quite.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/RecordTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/RecordTemplate.kt
@@ -31,7 +31,7 @@ data class {{ rec.name()|class_name_kt }} (
 
     internal fun write(buf: RustBufferBuilder) {
         {%- for field in rec.fields() %}
-            {{ "(this.{})"|format(field.name())|write_kt("buf", field.type_()) }}
+            {{ "this.{}"|format(field.name())|write_kt("buf", field.type_()) }}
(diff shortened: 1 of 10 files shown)


badboy (approved):
So code-wise this looks fine.
But this moves us into a direction where the Kotlin and Swift implementaions differ even further.
#1053 adds a similar pattern in Kotlin, but is not merged yet and is based on some other work.

Do we apply this pattern to the Kotlin codegen next or do we get the other PR merged and then see that we get all codegens more unified again?

jhugman:
> But this moves us into a direction where the Kotlin and Swift implementaions differ even further.

I'm not sure I understand / agree with your analysis.

https://github.com/mozilla/uniffi-rs/pull/1042 and https://github.com/mozilla/uniffi-rs/pull/993 made the code-generators almost identical; #1042 started as a copy paste of the new Kotlin backend.

The changes here are to move structural types to be implemented as concrete converters (away from extension methods on the generic type), to match how Kotlin does it. Kotlin can't do it in the way Swift was doing it, and so this PR moves the Swift implementation closer to the Kotlin.

PR #1063, opened 2021-10-02
Allow constructors to have a `[Self=ByArc]` attribute

mhammond:
We are finding ourselves wanting to expose a [constructor that returns an Arc](https://github.com/mozilla/application-services/blob/6632fe69f1c6c14c9a82ea5195d689e006ae3ca4/components/places/src/api/places_api.rs#L90) and it seems reasonable that UniFFI allow this. Reusing `[Self=ByArc]` isn't a *perfect* fit, but it seems close enough.

WDYT?


fixtures/coverall/src/coverall.udl
```diff
@@ -47,6 +50,10 @@ interface ComplexError {
 interface Coveralls {
     constructor(string name);
 
+    // There's one "global" coverall - this creates or returns it via an Arc<>
+    [Name="global_new", Self=ByArc]
+    constructor();
```
jhugman, line 55: If I'm reading this correctly:

```swift
let c1 = Coveralls.globalNew()
let c2 = Coveralls.globalNew()

assert c1 != c2 // fail
```

If we named this `get_instance`, we can see that this isn't a constructor at all.

Have I got that right? Was this your intent?

docs/manual/src/udl/interfaces.md
```diff
@@ -189,6 +189,26 @@ impl TodoList {
...
+    constructor();
+    ...
+};
+```
+
+```rust
+impl TodoList {
+    fn new() -> Arc<Self> { ... }
+}
+```
```
jhugman, line 210: From this file, I'm not sure why this is a good thing.

jhugman (commented):
I like what this PR enables, however I don't like that it changes the semantics of what a constructor is.

If I understand it properly, I should be able to implement a singleton with a constructor; which doesn't seem right.

bendk:
> I like what this PR enables, however I don't like that it changes the semantics of what a constructor is.

I think this gets to the heart of the matter.  Is `PlacesAPI::new` a constructor or is it a static method?  If we added a `[Staticmethod]` annotation, and put that on `new`, would that just work?  If so, maybe that's a better path -- it simplifies describing how constructors work at least.  It also combines nicely with your other change to support interfaces without constructors.

fixtures/coverall/src/coverall.udl
mhammond: The intention is that `assert c1 == c2` would pass. As part of that process, a `Coveralls` object was constructed.

I'm trying to wrap the pattern at https://github.com/mozilla/application-services/blob/a3d01428113d4a237c4a76e7095efad9c13fe958/components/places/src/api/places_api.rs#L90 and exposed to Kotlin at https://github.com/mozilla/application-services/blob/a3d01428113d4a237c4a76e7095efad9c13fe958/components/places/android/src/main/java/mozilla/appservices/places/PlacesConnection.kt#L37. I think it's a perfectly reasonable pattern - you might have chosen to use a different pattern, but I think it would be very brave to say that the places_api pattern is objectively wrong. I don't think UniFFI should be so opinionated that it refuses to support that pattern.

mhammond:
> I think this gets to the heart of the matter. Is `PlacesAPI::new` a constructor or is it a static method?

At a high level, Rust itself makes no distinction. Eg, https://rust-unofficial.github.io/patterns/idioms/ctor.html says:

> Rust does not have constructors as a language construct. Instead, the convention is to use a static new method to create an object.

So IMO it's all about intent.

PlacesApi has:

```
impl PlacesApi {
    /// Create a new, or fetch an already open, PlacesApi backed by a file on disk.
    pub fn new(db_name: impl AsRef<Path>) -> Result<Arc<Self>> { ... }
```

I certainly view that as a constructor. It the only way possible to create an instance of a `PlacesApi`.

> If we added a [Staticmethod] annotation, and put that on new, would that just work?

It probably would, yeah. But as above, if I have to choose between: "is this (a) a struct with no constructors but with a static method named new  (b) is this a struct with a constructor that returns an arc ", I'm inclined to say (b)

I guess I don't really object to `[Staticmethod]` but it seems odd to me that we would be adding support for it not because we actually have static methods, but because we have what's logically a constructor that returns an arc.

fixtures/coverall/src/coverall.udl
jhugman: Gosh.

jhugman:
>> Rust does not have constructors as a language construct. 

>> If we added a `[Staticmethod]` annotation, and put that on new, would that just work?

> It probably would, yeah. 

> I guess I don't really object to `[Staticmethod]`

I support this plan. :)

bendk:
I see the method as 50/50 between a static method and a constructor since it sometimes (mostly?) creates a new instance and sometimes returns an existing one.  I would say the same thing for method that creates a singleton on the first call, then returns the same instance afterwards.

I could go either way.  One nice thing about calling it a static method is that you can keep the rule that constructors return owned objects and methods return Arcs.   There's definitively a logic to that.  OTOH, it does play the role of a constructor and if we think users will feel like it's surprising or wrong to not call it a constructor, then I can think we should call it one.

mhammond:
> Looking [at the function body](https://github.com/mozilla/application-services/blob/a3d01428113d4a237c4a76e7095efad9c13fe958/components/places/src/api/places_api.rs#L89-L93), I see that though it's named `new`, it implements a cache of `PathBuf` to objects. The actual "construction" of a new struct happens after a [cache miss](https://github.com/mozilla/application-services/blob/a3d01428113d4a237c4a76e7095efad9c13fe958/components/places/src/api/places_api.rs#L119), which then goes on to have side-effects on global state.

Thanks for that, but I'm sure you know that I already knew all of that. We also don't *need* static methods - any global method would work too - I'm a little surprised you didn't "well, actually" that argument too. I can't bring myself to make a PR that includes the change entry "Added support for static methods so constructors can return an Arc", so I'm abandoning this entirely.

PR #1064, opened 2021-10-02
Allow interfaces to have no constructors

mhammond:
It seems reasonable that many interfaces will not want to expose a constructor, so I was surprised when one was created for an interface that didn't specify one. I even checked out the docs etc to make sure I hadn't missed anything, and the docs don't even hint at this behaviour.

It would be possible to avoid having this be a breaking change by, say, introducing a new attribute, but that doesn't really sit quite right with me - I think this new behaviour is intuitive and matches what the docs actually say. The cost of this breaking change is very small, so I'm leaning towards doing it this way - but if someone feels strongly I'll implement something else to avoid the breaking change.


jhugman:
The code looks good, but I don't understand why this might be useful.

An end-to-end example and a few paras in the docs about how to use constructor-less objects would be instructive.

mhammond:
> The code looks good, but I don't understand why this might be useful.

There are plenty of Rust structs that don't declare constructors - think factory structs etc.

```
struct Parent {
  pub fn get_child(&self) -> Child { ... }
}

struct Child {
  // no `pub` ctor - must get it from the parent
}
```

and this is exactly what we currently have for "places"

> a few paras in the docs about how to use constructor-less objects

TBH I don't think that makes sense - it's difficult to document a negative, which is why there are no docs that needed to change for this PR. eg, I can't find docs in rust itself for the concept of "no public constructor". Can you please suggest specifically where and what I would say?

AlexDBlack:
> The code looks good, but I don't understand why this might be useful.
> 
> An end-to-end example and a few paras in the docs about how to use constructor-less objects would be instructive.

We've run into this a number of times, and were forced to add constructors like this as a workaround:
```rust
    pub fn new() -> Foo {
        // Uniffi expects a constructor even though we don't want/need one :(
        unimplemented!("Not supported")
    }
```

jhugman:
Ok, so it's to prevent direct construction by the foreign language side. I'm still getting used the world of lifting and lowing to `pointer`s instead of `handle`s.

Doc ideas to add to the [`interfaces.md`][1] doc.
---

By convention, the `constructor()` corresponds to a `new()` method on the Rust struct.

Where this method isn't called `new()` or in the case where multiple constructors are needed, a `[Name]` annotation is used to specify the names of static methods.

```webidl
interface FirefoxAccount {
    constructor();
    [Name=from_json]
    constructor(JsonObject obj);
}
```

In the case where no `constructor` is specified, the foreign language side will be unable to instantiate the Rust object, and have to rely on factory methods or functions.

```webidl
namespace blah {
    Singleton get_instance();
}

interface Singleton {
    void method();
}
```

---

[1]: https://github.com/mozilla/uniffi-rs/blob/82b8b1917c0ac383c52b5103dfa1d130e0c05a01/docs/manual/src/udl/interfaces.md

jhugman (approved):
This looks good.

I've left some ideas of documenting this, which you can either copy direct or write your own.

I don't think we need to block for another review cycle.

PR #1075, opened 2021-10-18
Load Python library from absolute path

badboy:
First of the `*.so` file extension is valid on more than just Linux,
e.g. on BSD and Solaris too.
We can just consider it the default case and only change it for Windows
and macOS.

Next, for final deployment the compiled library can be placed next to
the Python code and that way the wrapper code can find it by its
absolute path.

---

this is what we currently [do support Glean Python](https://github.com/mozilla/glean/blob/b2f72f861f17339689f7c6881bd8471bbcd79b88/glean-core/python/glean/_ffi.py#L21-L36).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,6 +81,11 @@ jobs:
       - run:
           name: "Set RUSTFLAGS to fail the build if there are warnings"
           command: echo 'export RUSTFLAGS="-D warnings"' >> $BASH_ENV
+      - run:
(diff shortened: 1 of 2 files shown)


thunderbiscuit:
Hello folks,

Just wanted to say that I was having troubles with packaging as well (see #1112), and tested this PR on macOS and Ubuntu. Solved my problem well. We can now provide the native lib right beside the module and the package finds it on its own.

PR #1085, opened 2021-10-28
Update default Rust version to 1.56.

mhammond:
As per our Rust policy, we track mozilla-central, and that bumped to
1.56 in bug 1736459.

Note that CI is currently broken by 9a1c543b08c874edef09629bf77f528d315a00e6 - rather than fix that I just bumped Rust even further, which also has some clippy related fallout.


bendk (approved):

skhamis:
Before this lands, there is actually a super recent rust version update `1.56.1` that has a security fix. Might be worth it or not but more of a FYI -- https://blog.rust-lang.org/2021/11/01/Rust-1.56.1.html

PR #1087, opened 2021-10-28
Simplify codegen by implementing `CodeType` on more structs

bendk:
I'm trying to take the code in `uoc-macro-dispatch`, break it up into smaller chucks and avoid macros.  Here's the first piece.

- Implement `CodeType` on `Type` as well as type-like structs
  (`Record`, `Object`, `Field`, etc.).
- Update the Kotlin filter functions to input any of those structs, not
  just `Type`
- This allows us to simplfy the templates by not calling `type_()` so
  much.  I think it will help more when I try to implement the
  `FFIConverter` patterns.

If this looks good for Kotlin, then we can do the same thing for
Swift/Python.


uniffi_bindgen/src/bindings/backend.rs
```diff
@@ -163,3 +163,112 @@ pub trait CodeDeclaration {
...
+    fn code_type_impl(&self, oracle: &dyn CodeOracle) -> Box<dyn CodeType> {
+        oracle.find(&self.type_())
+    }
+}
+
+impl CodeTypeDispatch for Argument {
+    fn code_type_impl(&self, oracle: &dyn CodeOracle) -> Box<dyn CodeType> {
+        oracle.find(&self.type_())
+    }
+}
```
jhugman, line 220: I like very much how regular this is, and that it ties together the `ComponentInterface` structures with the code generation via only an oracle.

uniffi_bindgen/src/bindings/backend.rs
```diff
@@ -163,3 +163,112 @@ pub trait CodeDeclaration {
...
+    }
+
+    fn helper_code(&self, oracle: &dyn CodeOracle) -> Option<String> {
+        self.code_type_impl(oracle).helper_code(oracle)
+    }
+
+    fn imports(&self, oracle: &dyn CodeOracle) -> Option<Vec<String>> {
+        self.code_type_impl(oracle).imports(oracle)
+    }
+}
```
jhugman, line 274: It took me a while to get why this was necessary: so now we can treat `ComponentInterface` structures as `CodeTypes`! Is that right?

Could provide a comment on the `impl`?

jhugman (approved):
I'm really happy that you pursued this.

This is really neat.

I'd request a few more comments, but definitely nothing that needs another review cycle.

I'd also suggest that you raise an issue on each of the other backends so we can introduce the same changes there too (excepting the ones in `backend` o/c).

uniffi_bindgen/src/bindings/backend.rs
```diff
@@ -163,3 +163,112 @@ pub trait CodeDeclaration {
...
+// Needed to handle &&Type and &&&Type values, which we sometimes end up with in the template code
+impl<T, C> CodeTypeDispatch for T
+where
+    T: std::ops::Deref<Target = C>,
+    C: CodeTypeDispatch,
+{
+    fn code_type_impl(&self, oracle: &dyn CodeOracle) -> Box<dyn CodeType> {
+        self.deref().code_type_impl(oracle)
+    }
+}
```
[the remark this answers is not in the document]
bendk: This one was a learning process for me as well :)

PR #1091, opened 2021-10-28
Move backend and gen_[language] module paths

bendk:
Moving some modules around, no functional changes.  See the commits for my reasoning.  I'm not overly concerned with the current layout, but I think these are small improvements.  What do you think?


jhugman (approved):
As you say, this seems to be a fairly mechanical change; from what I can see— all for the better. Thank you.

Edit: my only nit: could you change the PR title to something more descriptive: which module paths? What is shuffling mean in this context?

PR #1092, opened 2021-11-01
RFC: bindings-side wrappers

bendk:
This is taking off where my [`extension-types` proposal](https://github.com/mozilla/uniffi-rs/pull/1055) left off.  I'm really liking how much this enables without too many design changes.

### DIFF
--- a/examples/wrapper-types/uniffi.toml
+++ b/examples/wrapper-types/uniffi.toml
@@ -9,3 +9,30 @@ cdylib_name = "wrapper_types"
 
 [bindings.kotlin]
 cdylib_name = "wrapper_types"
+
+# Python wrapper implementation for JsonObject
+[wrappers.JsonObject.python]
+
+code = """
+class JsonWrapper:
+    fn from_wrapped(string):
+        json.loads(string)
+
+    fn into_wrapped(obj):
+        json.dumps(obj)
+}
+"""
+imports = ["json"]
+wrapper_class = "JsonWrapper()"
+
+# Kotlin wrapper implementation for JsonObject
+[wrappers.JsonObject.python]
+
+code = """
+class JsonWrapper:
+    fn from_wrapped(value: string) = JSONObject(value)
+    fn into_wrapped(value: JSONObject) = value.toString()
+}
+"""
+imports = ["org.json.JSONObject"]
+wrapper_class = "JsonWrapper()"


examples/wrapper-types/uniffi.toml
```diff
@@ -9,3 +9,30 @@ cdylib_name = "wrapper_types"
 
 [bindings.kotlin]
 cdylib_name = "wrapper_types"
+
+# Python wrapper implementation for JsonObject
+[wrappers.JsonObject.python]
+
+code = """
```
bendk, line 16: In the last proposal, I separated out the Python/Kotlin code into their own files and I still think that has some merit.  Here's how I'm thinking about that:
 - We can start by just using a string in the `.toml file`
 - If we decide we need something like that, then maybe we could add an way for `uniffi.toml` to include other files.  Maybe instead of including the code inline you could do something like `code = "@include("uniffi/json-object.py")`

examples/wrapper-types/uniffi.toml
```diff
@@ -9,3 +9,30 @@ cdylib_name = "wrapper_types"
...
+class JsonWrapper:
+    fn from_wrapped(string):
```
bendk, line 18: `from_wrapped/into_wrapped` is my latest attempt at making it clear which direction these functions are going.  I think `wrap()` and `unwrap()` would do just fine as well though.

examples/wrapper-types/uniffi.toml
```diff
@@ -9,3 +9,30 @@ cdylib_name = "wrapper_types"
...
+        json.loads(string)
+
+    fn into_wrapped(obj):
+        json.dumps(obj)
+}
+"""
+imports = ["json"]
+wrapper_class = "JsonWrapper()"
```
bendk, line 26: In the last proposal you specified the wrapping/unwrapping functions.  I like specifying an object to handle this for a few reasons:
  - It's how the scaffolding side works.
  - It keeps the concept of an "extension type" alive.  We could share a `jsonwrapper.py` file between projects to avoid rewriting the same code.
  - It would work better if we ever needed to add extra methods or arguments to these methods.

examples/wrapper-types/uniffi.toml
mhammond: huh - I really think we need to come up with something better here, because my brain hurts every time I think about it. It says "a jsonobject is a wrapper around a string", or shorter, "a json object wraps a string" - which is reflected in the `JsonWrapper` name. So in this world, "wrap" would take a string and return a JSONObject (ie, take the thing being wrapped, and return the wrapper). However, 'wrapped' seems more problematic - after you wrap a string, it is wrapped in a `JsonWrapper` - so `from_wrapped` could just as easily mean "from json wrapper" - which is the opposite of what's going on - although your example *also* looks correct if you tilt your head.

`from_wrapper` is more consistent in this example, given we've named the class `JsonWrapper` - but that's the exact opposite of your intent. `from_primitive/into_primitive` seem to avoid that.

examples/wrapper-types/uniffi.toml
mhammond: IMO, one of the most important things here is that each consumer can specify a different uniffi.toml than the component itself - which I think would naturally lead to allowing each language binding to have its own too.

I don't think the use-case for including .py code makes much sense TBH - uniffi doesn't need to get involved here - the consumer could just stick `json-object.py` wherever they like and just import it. It might make sense for other languages though, and maybe I'm missing your point a little.

examples/wrapper-types/uniffi.toml
mhammond: My problem with your proposal above is that it's a class that exists only as a holder for the 2 methods (which should be marked as `@static`? A downside is that it gives the impression that this `JsonWrapper` is actually going to be used by the consumer - ie, that a casual reader would assume that anything declared in the UDL would take/get `JsonObject`, but that's not true. So `JsonObject` should probably be called something like `JsonConverter`?

But it still seems overkill - `json.loads({})` and `json.dumps(})` is significantly less boilerplate and has identical semantics. So I'd be inclined to simplify this out to:

```
setup = "import json"
from_primitive = "json.loads({})
to_primitive = "json.dumps({})"
```

but if someone really did think the wrapper added value, they could still write:
```
setup = """
import json # or keep 'imports', but still not sold on that ;)
class JsonWrapper:
    @staticmethod
    fn from_prim(string):
        json.loads(string)
    @staticmethod
    fn into_prim(obj):
        json.dumps(obj)
}
from_primitive = "JsonWrapper.from_prim({})"
to_primitive = "JsonWrapper.to_prim({})"
```

which IIUC is identical to your example stylistically, and identical to both the above semantically.

(Not too bothered by this though)

mhammond (commented):
I think this is great - I'm into pure bikeshedding which is a good sign IMO :) I'll start thinking about the "poc" or "big real world example" etc that can tie a few of these threads together, but I think we are converging on the shape, just debating some of the spelling!

jhugman:
Mark has just pointed me to this, apologies for not getting to it sooner.

As an RFC, a lone TOML file is a bit… sparse. :( I'm not sure what I'm reviewing here, and what context is assumed.

Would you mind fleshing it out into some end-user facing  documentation? This will help reduce the area of confusion and number of assumptions that we may or may not share.

* What happens if a TOML file is missing the entry for the type?
* What happens if a TOML is there, but the conversion code is not available at runtime/compile time?
* Do I have to wrap string, or can I wrap something else (e.g. a struct or enum)?
* How would I re-implement Duration or Timestamp, all the way from Rust to Swift?
* Could you give some examples of what the generated Kotlin or Swift looks like?
* What are the implications for the compile time? Does this/Should this have any symmetry with the Rust compile time (hint: yes)?
* How does the proposed syntax compliment/compose/intersect with other features?
* Is there a route to be able to support generic types, expressed either as a concrete type in the UDL or as a generic type in UDL (e.g. `typealias FutureString = Promise<String>`

FWIW, I think having a larger, more long running document we can collaborate and comment upon would help us iterate and nail everything down quicker.

From what little is here, 

 * I think we should keep the code out of the TOML as much as possible: I'd prefer having an `imports` array and a `from`/`into` expressions (which is what you have!).
 * I don't think we're going to get away without an `imports` array for Swift.

If this is an RFC as a PR, I think we should have a landable artifacts at the end of the process (e.g. docs and some examples), otherwise, let's move this to a Google Doc first where it's a bit easier to comment.

bendk:
This is definitely sparse. I think we can accomplish a lot with only a few TOML entries.  I just opened #1107, which uses a similar syntax to this and mor-or-less implements the feature.  I just realized I forgot to add docs to that PR, but I'll add them very soon.

I'll try to answer your questions as best as possible.  Then let's take the conversation to the other issue.

-   What happens if a TOML file is missing the entry for the type?
    -  It will just be an alias for the inner/primitive type.
-   What happens if a TOML is there, but the conversion code is not available at runtime/compile time?
    - This would break things, probably you would get a compile error.
-  Do I have to wrap string, or can I wrap something else (e.g. a struct or enum)?
   - You can wrap anything, although there's currently a requirement that there's a typedef for it.
-  How would I re-implement Duration or Timestamp, all the way from Rust to Swift?
   - You could kind of implement this now, but I'm hoping to make it easier by allowing users to wrap any type on the bindings side, regardless of if it's in a Wrapped typedef or not.  This way you don't have to change all your function signatures to input a `wrappedTimestamp` or something like that.
    - This feature would allow you to *wrap* a timestamp, for example maybe converting it to your local time.  A separate feature that we might want to support is to replace the FfiConverter for timestamp.  This would allow completely re-implementating a type.
-  Could you give some examples of what the generated Kotlin or Swift looks like?
   - Check out #1107
 -  What are the implications for the compile time? Does this/Should this have any symmetry with the Rust compile time (hint: yes)?
    - I'm not sure what you're getting at here.  I did try to make the codegen for the scaffolding similar to the codegen for the bindings.
-  How does the proposed syntax compliment/compose/intersect with other features?
   - I think we should have a similar pattern for this feature, the decorators feature, and other related features:
        - The UDL file uses an abstract names for things like `JsonObject`
        - The config file specifies the concrete details of those things (type signatures, wrapper functions, the decorator class, etc).
        - The config file name things rather than provide code.  The code should be provided from a user-supplied source file and the config can import it.
 - Is there a route to be able to support generic types, expressed either as a concrete type in the UDL or as a generic type in UDL (e.g. typealias FutureString = Promise<String>
   - I think this is more related to the decorator feature and wrote some code shat showed how you could have generic types there.  In general, I think you can handle generic types in the config file just fine.

PR #1093, opened 2021-11-01
RFC: example of how decorators could be specified in uniffi.toml

bendk:
We've been discussing the concept of function decorators in mozilla/application-services#4595 and #1051.  One idea that's been thrown around a bit is defining the decorators outside of the UDL file.  Here's a proposal of how they could be defined in `uniffi.toml`.  It's hand-wavey, but hopefully it gives a good picture of the general approach.  I think there are a few nice things about this:
  - Decorators aren't tied to classes.  You can have a decorator that's shared by methods of different classes and top-level functions.
  - Consumer's don't need to implement the decorator if they don't want to.
  - It doesn't change the constructor signature.  This allows the objects to continue to work inside `Record` and other containers.
  - It seems to me that wrapper types and decorator functions should be defined in the same place.  If we're going to define wrapper types in `uniffi.toml` then that seems like a natural place for the decorators.
  - It supports all types of generics (this seems like more of an IDL-parsing issue than an issue of where the config should live though).
 
Side issue: I think we should lean towards names based on the decorated function (`io`) rather than the decorator (`on_db_thread`).  `io` would work with running the function in a coroutine, a thread pool, a separate process, etc.  `on_db_thread` implies that it's going to run on a dedicated thread.

### DIFF
--- a/examples/decorators/src/decorators.udl
+++ b/examples/decorators/src/decorators.udl
@@ -1,0 +1,22 @@
+namespace decorators {
+    // The Decorator attribute says that this function may be wrapped by a
+    // decorator named "io".  We define what that means in uniffi.toml
+    [Decorator=io]
+    void sync_all_users();
+};
+
+record User {
+    id: string,
+    name: string,
+}
+
+interface UserDatabase {
+    // Another use of the decorator from above
+    [Decorator=io]
+    User lookup();
+
+    // A different decorator
+    [Decorator=log-errors]
+    [Throws=RustyError]
+    void import_legacy_data(string: data);
+}
--- a/examples/decorators/uniffi.toml
+++ b/examples/decorators/uniffi.toml
@@ -1,0 +1,43 @@
+[bindings.python]
+cdylib_name = "wrapper_types"
+
+[bindings.ruby]
+cdylib_name = "wrapper_types"
+
+[bindings.swift]
+cdylib_name = "wrapper_types"
+
+[bindings.kotlin]
+cdylib_name = "wrapper_types"
+
+# Defining the log-errors decorator for the Python bindings
+[decorators.log-errors.python]
+imports = ["logging"]
+code = """
+error_logger = logging.getLogger("errors")
+
+def logErrors(func):
+  try:
+    return func()
+  except:
+    error_logger.error("Error calling {}".format(func), exc_info=True)
+    raise
+"""
+decorator = "logErrors"
+
+# Defining the io decorator for the Python bindings
+[decorators.io.kotlin]
+imports = [
+    "kotlinx.coroutines.Dispatchers",
+    "kotlinx.coroutines.CoroutineScope"
+]
+code = """
+ioScope = CoroutineScope(Dispatchers.IO)
+"""
+decorator = "ioScope.launch"
+
+# I think the return type can be inferred and we don't need the next line.  But
+# this is an example of how the return type could be specified.  Swift might
+# need something like this.
+return_type = "Deferred<{}>" # "{}" will be replaced with the decorated function's return type
+


bendk:
I'll just mention up front that the main drawback that I can see is that this requires global decorator functions which would be a pain if you needed complex runtime configuration or different decorators for different objects.  I'm pretty sure this could be reworked to support those things though.

examples/decorators/uniffi.toml
```diff
@@ -0,0 +1,43 @@
...
+cdylib_name = "wrapper_types"
+
+[bindings.swift]
+cdylib_name = "wrapper_types"
+
+[bindings.kotlin]
+cdylib_name = "wrapper_types"
+
+# Defining the log-errors decorator for the Python bindings
+[decorators.log-errors.python]
```
mhammond, line 14: really starting to bikeshed here, but is there a good reason not to have `bindings.python.decorators.log-errors`?

examples/decorators/uniffi.toml
```diff
@@ -0,0 +1,43 @@
...
+]
+code = """
+ioScope = CoroutineScope(Dispatchers.IO)
+"""
+decorator = "ioScope.launch"
+
+# I think the return type can be inferred and we don't need the next line.  But
+# this is an example of how the return type could be specified.  Swift might
+# need something like this.
+return_type = "Deferred<{}>" # "{}" will be replaced with the decorated function's return type
```
mhammond, line 42: Kotlin seems to support type inference of return values in some cases - but let's assume it didn't and/or swift can't make it work. Let's say we had 2 functions described in UDL, one returning a String and one returning an Int - and we want to decorate them both as "io", but want to avoid using `Any` in Kotlin because that would be a step backward. I wonder if there's a way to just assume the "correct" signatures exist? Obviously you'd need to arrange <handwave> for there to be 2 functions with different return values (and therefore, almost certainly different names). If you later go ahead and mark a function returning (say) a float, you can imagine us generating the code that naively assumes a float version of the decorator exists and let the compiler complain if it doesn't.

This isn't that different to the uniffi contract with your Rust code today - it just blindly assumes that what you wrote matches what actually exists, and if not, the compiler complains.

mhammond (commented):
I really like this. The Python example is perfect because that's what Python does, but I think we will want to flesh out the type-safe languages a little more, but I love the idea that while `Any` isn't a "uniffi" thing, I can see a path for the Kotlin bindings to use `Any` if it makes sense (or as I hand-wave below, maybe even avoid to keep type safety by putting the burden on the decorator author.

I'll circle back to the other RFC for wrapping, but at face value, I think this fits with that model reasonably well - including the possibility that the .toml where this is expressed can be next to the component itself, or miles away from it in an external repo where the component is consumed (ie, in the megazord dir, say). We should maybe try and put together a "real toy example" - a bit of a contradiction in terms, but maybe we can come up with something.

examples/decorators/uniffi.toml
```diff
@@ -0,0 +1,43 @@
...
+[bindings.kotlin]
+cdylib_name = "wrapper_types"
+
+# Defining the log-errors decorator for the Python bindings
+[decorators.log-errors.python]
+imports = ["logging"]
+code = """
+error_logger = logging.getLogger("errors")
+
+def logErrors(func):
```
mhammond, line 19: So stretching the other comments further, this should really be `def logErrors(func, *args, **kw): ...` - but how do we spell that in a generic way?

examples/decorators/src/decorators.udl
```diff
@@ -0,0 +1,22 @@
...
+};
+
+record User {
+    id: string,
+    name: string,
+}
+
+interface UserDatabase {
+    // Another use of the decorator from above
+    [Decorator=io]
```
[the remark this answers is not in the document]
bendk: I considered using "tags" when I made this.  I like how it does feel very general and it makes it clear that all you're doing the the UDL is marking the function.  I could definitely get behind it.

examples/decorators/uniffi.toml
```diff
@@ -0,0 +1,43 @@
...
+    return func()
+  except:
+    error_logger.error("Error calling {}".format(func), exc_info=True)
+    raise
+"""
+decorator = "logErrors"
+
+# Defining the io decorator for the Python bindings
+[decorators.io.kotlin]
+imports = [
```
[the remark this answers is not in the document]
bendk: There's a few nice things about separating the imports out.  If multiple decorators needed to import the same thing, then we would only have 1 import statement in the generated file.  You can put the imports at the top of the file.  Neither of these is all that compelling to me though.

I'm actually not sure how scoping would work.   I'm currently just thinking of dropping this code in the middle of all the other generated code.  If the code block lived in it's own module, then I'd be completely on board.  In python you could use the class definition as a scope and put the imports inside there.  But does that work in Kotlin/Swift?

examples/decorators/uniffi.toml
bendk: I was thinking that we would always input a function with no args and UniFFI would create a closure for that.

examples/decorators/uniffi.toml
bendk: As long as the decorator is a generic function, I think it all works out.  The generated code would be something like this:

```
fun foo(a: Int, b: String) = ioScope.launch {
   {{ func|lift }} (
      {{ func|ffi_func_name }}({{ func.args.a|lower }}, {{ func.args.b|lower}})
   )
}
```

bendk:
>  We should maybe try and put together a "real toy example" - a bit of a contradiction in terms, but maybe we can come up with something.

I think that could be great, but the one nagging issue on my mind is the one I mentioned before: the decorator functions need to be global.

In the other PR, I mentioned how these decorators could be great for the [android-components logins store code](https://github.com/mozilla-mobile/android-components/blob/main/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/SyncableLoginsStorage.kt#L171-L173), but that example depends on the `coroutineContext` field of that class.  If the decorarator is a global function, it's not going to have access to that.

In that example, `coroutineContext` is always `Dispatchers.IO`, so it wouldn't be much of a problem.  But what about a class that inputted its `coroutineContext` as a constructor argument?  I've seen that pattern when you want to use a different dispatcher for testing.

Do you think that would be a problem in practice?  @jhugman also interested what you think about this one.  I'm thinking that maybe I could change the system to support this use case better.

examples/decorators/uniffi.toml
mhammond: Right, for some reason I misunderstood that, including in James's patch. I think some of the confusion came from the fact they were specified in the UDL, and I guess because (IIUC), decorators in Python will want the `*args, **kwargs` in most cases. But here, we will *never* have args specified, and it will forever be impossible for a "decorator" to get access to the args? Also still not clear to me quite yet how return types should be managed efficiently, but as you say, maybe that all "just works"

jhugman:
To recap a call @benk and I have had about this proposal:

 * while this and #1051 appear to be approximately the same, I'm not sure that this RFC solves the problem that #1051 sets out to: letting the consuming app (or tests) being able to customize the calls into Rust, e.g. setting what `coroutineContext` and `errorReporter` the decorator uses at runtime.
 * most of the use cases for decorators are going to need the decorator to customize the return type or the exceptions thrown or both. This needs to be expressed in type safe languages.
 * for decorator methods to be re-usable and type safe we're probably not able to support the decorators being able to see the args that the method is  called.
 * I'm pretty unsure about putting code in the TOML. I could be persuaded that it's ok to put a templated expression (e.g. to wrap or unwrap a value) that refer to other code.

jhugman:
Reading [your iteration][1] on #1051, I think it will work in Kotlin because the generated code does not have to annotate the return type, so generate code will look like

```kt
fun myMethod(myString: String, myInt: Int) = myDecorator.method(this) { … } 
```

I love that the decorator can provide any return type it wants (including `Unit` or even `Deferred<T>`) because the generated code doesn't need to generate the type label for the return type.

However, I'm not clear how this would work in languages that require us to generate the return type label: e.g. in Swift

```swift
func myMethod(myString: String, myInt) -> TransformedType {
    return myDecorator.method(this) { … }
}
```




[1]: https://github.com/mozilla/uniffi-rs/commit/381c9e5c708cc01cbb6ffb2f3d199508678f6266

bendk:
> Could you sketch out how we specify/generate that TransformedType? The first concrete example of this would be the error catch, report and swallow case.

I'm thinking this could be in `uniffi.toml`.  Something like this:

```
[bindings.swift.decorators.RustObject]
import = "path/to/class/def"
class_name = "MyDecorator"

[bindings.swift.decorators.RustObject.return_types]
myMethod = "TransformedType"
myOtherMethod = "Unit"
```

What do you think about closing this one and moving the discussion to the other PR?  It seems like we only need to hold one of these open.

jhugman:
> What do you think about closing this one and moving the discussion to the other PR? It seems like we only need to hold one of these open.

Ok, sure, let's go back to #1051 .

PR #1097, opened 2021-11-03
Fix missing fmt argument in error message

dtolnay:
Without this, the error message would literally be "Error while running tests: {}" with curly braces in it instead of the exit status.


rfk (approved):
Whoops, thank you!

Issue #1098, opened 2021-11-04
Kotlin exceptions


l-maia:
The generated kotlin does not include the @Throws annotation which prevents the exceptions from being caught when called from Java.

I think that a small modification could provide better compatibility for applications mixing Java and Kotlin.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-109)
┆Issue Number: UNIFFI-109

PR #1099, opened 2021-11-04
Add throws annotation to kotlin exceptions

l-maia:
A PR adding the throw annotation as identified in issue #1098, to provide compatibility with Java.


bendk:
Seems like a good idea, but does this mean the exceptions must be caught in Java?  I'm wondering if this could be configurable either globally or on a per-function basis.  What do you think?

l-maia:
Yes, kotlin does not have checked exceptions so the Java signatures of Kotlin exceptions, so in Java it won't have an indication than an exception is to be thrown unless the annotation is used.

I think it is safer to have the exceptions and handling them in Java, than not having them at all.

https://kotlinlang.org/docs/java-to-kotlin-interop.html#checked-exceptions

mhammond:
huh - that link implies that the annotation is needed to make it *possible* to catch the exceptions in java - which surprises me, but I should learn to stop being surprised by Kotlin and Java :) @jhugman is our resident Kotlin expert, so hopefully he will weigh in, but at face value this LGTM.

jhugman:
This looks great. Thank you so much for your contribution.

I think the only thing I'd ask before landing this would be to do the same for the `uniffi_bindgen/src/bindings/kotlin/templates/TopLevelFunctionTemplate.kt` file, which would need similar annotations.

jhugman (changes requested):
Just adding this a formal review.

jhugman (approved):
This looks great, thank you so much.

jhugman:
There is one conflict in the `ObjectTemplate.kt` could you resolve this (or I can for you) and then we can hit the Big Green Button!

PR #1103, opened 2021-11-16
Python: Extract type coercing into a `CodeType` method

badboy:
Now every type knows how to coerce itself.
So far only Python needs that (Kotlin/Swift rely on the type system,
Ruby hasn't been migrated to UoC yet)

---

Follow up to #1081


jhugman (approved):
Neat!

PR #1104, opened 2021-11-16
Use correct path to backend module

badboy:
This was done in #1091, but missed in the recent Python UoC work


bendk (approved):
Looks good

Issue #1106, opened 2021-11-18
Refactor Swift FFI type handling


bendk:
The `type_ffi` filter on swift is a bit confusing and should be refactored.  It currently outputs the type that we want to use in the header file (e.g. `int32_t`).  This is different than all the other bindings, which outputs the type that `lower()` returns (`Int32` in this case).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-113)
┆friendlyId: UNIFFI-113

jhugman:
I wonder if this should prompt us to break the file in to a `gen_swift` and `gen_header` files, with two different filters modules?

```rust
use  super::header_filters as filters;
```

bendk:
This was fixed by #1106.

Issue #1089, opened 2021-10-28
Simplify code declaration system


bendk:
This is part of a plan to take pieces from the [uoc-macro-dispatch](https://github.com/mozilla/uniffi-rs/pull/1052) PR and present them as smaller, simpler, macro-less, PRs.

Make a PR that extracts the `CodeBuilder` and related changes from the closed PR.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-106)
┆friendlyId: UNIFFI-106
┆sprintEndDate: 2021-12-10

bendk:
I didn't make this exact change, but I did land a few related PRs.  I think we can close this one.

Issue #411, opened 2021-03-02
Support for arbitrary map type with string keys?


leplatrem:
I would like to generate cross-platform bindings for [our canonical JSON serializer](https://github.com/mozilla-services/canonicaljson-rs/blob/eb4019e2baaf389aaaa83f87e5fa6a8ec265be4a/src/ser.rs#L219-L225).

The input argument is a generic map or object (like `json.dumps()` or `JSON.stringify()`).
I had it working [in Wasm using  `JsValue`](https://github.com/leplatrem/canonicaljson-wasm/blob/fd53f1f42a6b748b1bd9d5723696b65d8bd2cde5/src/lib.rs#L9), and [`PyObject` in Python](https://github.com/mozilla-services/python-canonicaljson-rs/blob/f972cae893d9def0a061f076ec19c3248758825e/src/lib.rs#L87-L167)

Since it onlt has a single function and signature, I'm tempted to go on with my manual bindings instead of investigating how to support this kind of `Any` parameter.
What do you think? Is it something that would be achievable? 
Thanks :pray:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-51)
┆Issue Number: UNIFFI-51

rfk:
👋 hi @leplatrem, thanks for taking a look here!

I agree that this kind of "take any arbitrary JSON object" is probably a bit awkward atop the current set of primitives in UniFFI. It's also not clear to me what an idiomatic consumer API would look like in our strongly-typed target languages like Kotlin or Swift. They do both appear to have an `Any` type, but I don't know how well that maps to working with JSON in practice. A quick look suggests that that might use a typed-enum model similar to serde.

A couple of possible ways to go from here, ordered from what I think would be "best experience for consumers of your API" through to worst:

* Try to make some hand-written bindings to Kotlin and/or Swift to get a feel for how it should work, and then later we can take those learnings and try to expand the automated capabilies of UniFFI.
* Accept a JSON string, parse it on the Rust side, and return a canonicalized string.
* Define a typed-enum-based API [like the one that serde uses](https://docs.serde.rs/serde_json/value/enum.Value.html) and expose that via UniFFI.

/cc @jhugman who I know has been thinking a bit about how to expose JSON objects to Kotlin consumers.

jhugman:
Hi @leplatrem , thanks @rfk 

I've been thinking about this for the last couple of days, so I could say something sensible.

[...]

Issue #396, opened 2021-02-19
Add support for recursive enums


rfk:
In https://github.com/mozilla/uniffi-rs/pull/381 we added support for enums with associated data, but it doesn't properly support *recursive* enums, like the classic "tree" example where the enum variants are `Tree::Leaf` and `Tree::Branch(left: Tree, right: Tree)`. Some notes on what we'd need in order to support such:

* In the rust code, we'll need to be prepared to `Box` the values in the recursive fields.
* In the swift binding, we'll need to use `indirect` keyword to support [recursive enums](https://docs.swift.org/swift-book/LanguageGuide/Enumerations.html#ID536).

Hywan:
Still necessary :-). Any idea if we are close to get something, or should I try?

mhammond:
I've not heard of anyone with plans to do this soon.

PR #1107, opened 2021-11-18
Bindings wrapper types

bendk:
Attempt at implementing foreign side bindings for Python, Swift, and Kotlin.  I think Ruby should be very possible as well, I just didn't get around it it.

Updated user docs: https://github.com/mozilla/uniffi-rs/blob/48fbf17a8145da33659732556f522de3d9b5c5e8/docs/manual/src/udl/ext_types_wrapped.md

The wrapping config is specified in in `uniffi.toml`:
  - You specify a wrap and unwrap function.
  - For typed languages, you specify the type name
  - Optionally you can specify a list of things to import.  This allows you to provide a module for uniffi to get the wrap/unwrap functions from
  - The scaffolding code also uses `uniffi.toml`.  It has a slightly different syntax than the bindings code, but the idea is the same.  I did this because a) it makes the scaffolding system more similar to the bindings and b) it allows you to wrap something on the bindings side only.  I finally thought of a good use case for this one too -- wrapping a record in order to provide a different API for the consumer code -- like how I wrapped our `EncryptedLogin` struct to present a flat API for `android-components`.

Couple of side issues:
  - I took this opportunity to make the `CodeOracle` types stateful so they can access the `Config`.  This leads to some weird code in filters modules, but it's not terrible IMO.  It seems like something we're going to need to do sooner or later.
  - This breaks the python external types test cases, but I think that's okay because the feature is not really ready in general yet.  I also want to work on the python code to make it use the FfiConverter pattern, which makes external types much simpler.
  - I updated the test code system a bit.  My idea here is that we can continue to use the macro for simple cases, but for more complicated cases -- like when we need to supply extra source files to provide the wrapper code -- we can manually create and run a `ForeignLanguageTestCase`.
  - I had to add some ugly FFI-type handling code on Swift because of #1106

I really wanted to show how JSON wrapping could work in Kotlin, but AFAICT there's no builtin support for it and I didn't want the tests to depend on a 3rd party library.  I think I could have made it work on Swift, but I would need to have a better understanding of the lagnuage first.

I think the next step would be to allow the foreign bindings to wrap any type, not just types that are have the Wrapped typedef in the UDL.  This would allow things like letting them override the default timestamp handling and it would make wrapping records/enums a bit nicer.  I think this would be possible, but I'd like to do a few more template code refactoring changes first.


uniffi_bindgen/src/bindings/kotlin/templates/WrappedTemplate.kt
```diff
@@ -0,0 +1,41 @@
+object FfiConverterType{{ name }} {
+    {%- match config %}
+    {%- when None %}
+    {#- No wrapper config, just forward all methods to our inner type #}
```
jhugman, line 4: I think this `None` branch should collapse down to:

```kt
typealias {{ name|type_name }} = {{ inner|type_name }}

```

followed by an FfiConverter object that fowards on to the inner with `inline` functions.

uniffi_bindgen/src/bindings/mod.rs
```diff
@@ -130,21 +130,35 @@ pub fn compile_bindings<P>(
...
     let out_dir = out_dir.as_ref();
     match language {
-        TargetLanguage::Kotlin => kotlin::compile_bindings(&config.kotlin, ci, out_dir)?,
-        TargetLanguage::Swift => swift::compile_bindings(&config.swift, ci, out_dir)?,
-        TargetLanguage::Python => (),
+        TargetLanguage::Kotlin => {
+            kotlin::compile_bindings(&config.kotlin, ci, extra_files, out_dir)?
+        }
+        TargetLanguage::Swift => swift::compile_bindings(&config.swift, ci, extra_files, out_dir)?,
+        TargetLanguage::Python => copy_extra_files(extra_files, out_dir)?,
```
jhugman, line 145: A comment here would be good to explain why Python is `copying`, Ruby is doing nothing, and Swift and Kotlin are doing approximately the same.

It looked odd, until I looked at the context and thought about it. Don't make me think, gosh dern it. :)

jhugman (changes requested):
This is a big PR, so I haven't got to all of it, but there are enough changes required here (and! the! backends! are! all! the! same!) :) to keep you unblocked with some fast feedback.

On the whole, I like the direction of this.

docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -57,13 +83,83 @@ Because `UniffiCustomTypeWrapper` is defined in each crate, this means you can w
...
+This time however, we will update the Kotlin section of the `uniffi.toml` file:
+
+```toml
+[bindings.kotlin.wrappers.LegacyUser]
+# Name of the Kotlin class that will wrap `User`
+type_name = "LegacyUser"
+# Function that converts `User` to `LegacyUser`
+wrap = "wrapUser"
+# Function that converts `LegacyUser` to `User`
+unwrap = "unwrapUser"
```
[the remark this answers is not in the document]
bendk: Swift is really tricky.  I implemented it using unnamed params, but I'm not sure if that's good or bad.  Placeholders would also cut down on some boilerplate, like being able to use `{}.toString()` on Kotlin rather than having to create a separate function to do that.

I removed them just because I wanted to avoid anything code-like in the `.toml` file, but I'll try adding them back.

docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -57,13 +83,83 @@ Because `UniffiCustomTypeWrapper` is defined in each crate, this means you can w
...
+Wrapper types require some configuration in `uniffi.toml`.  Here's how it works:
 
-We intend on making it possible to customize this - for example, `serde_json::Value` could be
-transformed into a suitable JSON object in the bindings.
+* Create a `[scaffolding.wrappers.{WrapperTypeName}]` table to enable wrapping on the scaffolding side.  This only has one key:
+  * `type_name`: Rust type name for the wrapper.
+* Create a `[bindings.{language}.wrappers.{WrapperTypeName}]` table to enable wrapping on a bindings side.  This has several subkeys:
+  * `type_name` (Typed languages only): Type/class name for the wrapped type
+  * `wrap`: Function to wrap the UDL type, converting it to the wrapped type.
+  * `unwrap`: Function to unwrap the wrapped type, converting it to the UDL type.
```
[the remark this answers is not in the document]
bendk: I don't disagree, but I haven't been able to get anything better.  One idea I was thinking about was `from_inner`, `into_inner` instead of `wrap` / `unwrap`.

For `Type::Wrapped`, I wonder if we could eventually making that just `Type::Typedef` and allow any type to be wrapped.  I don't know if this fixes the naming issue though, it's just that a different thing would be labeled "wrapped".

examples/wrapper-types/src/wrapper-types.udl
```diff
@@ -4,9 +4,27 @@ typedef string JsonObject;
...
+dictionary UserInfo {
+    string name;
+};
+
+dictionary AccountInfo {
+    i32 id;
+};
+
+[Wrapped]
+typedef UserAccount WrappedUserAccount;
```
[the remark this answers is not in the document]
bendk: I can't think of a good example that wraps timestamp.  How would you do it?

IMO, one thing that makes this mindbending is everything needs to be wrapped in a separate typedef with a separate name.  When wrapping primitives that seems completely natural, but when wrapping structs/enums you end up with names like `WrappedUserAccount` that don't feel intuitive to me.  I do want to change this, but I think that needs to be down the line.

FWIW, this is very similar to the actual [code we just shipped for logins](https://github.com/mozilla-mobile/android-components/blob/main/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/Types.kt#L16-L28).  On the Rust side `EncryptedLogin` has several nested objects, but on in `android-components` we want it to be flat.  We also need to convert it from the data class that UniFFI defines to the data class defined in the abstract interface (`mozilla.components.concept.storage.EncryptedLogin` )

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/wrapped.rs
```diff
@@ -41,20 +63,52 @@ impl CodeType for WrappedCodeType {
...
+    pub fn new(codetype: &WrappedCodeType) -> Self {
+        Self {
+            name: codetype.name.clone(),
+            inner: codetype.inner.clone(),
+            config: codetype.config.clone(),
+        }
+    }
+
+    fn inner_ffi_type(&self) -> FFIType {
+        (&self.inner).into()
```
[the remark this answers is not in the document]
bendk: > The KotlinWrappedType should be impl CodeDeclaration, and then that can render the declaration_code and imports instead of going through the WrappedCodeType.

Can you explain this one a bit more?

One thing I notice is that in the other modules, we split up the code to handle the `Type` vs the code to handle the definition struct.  `ObjectCodeType` handles `Type::Object` and `KotlinObject` handles `Object`.  This got tricky for wrapper types, because there's no separate definition struct -- `Type::Wrapped` contains all the data.  I'm also just not sure of why we need that split, IMO it would be nicer if `ObjectCodeType` and `KotlinObject` were merged into 1 struct.

> All the languages we support support type aliasing. If the config doesn't specify wrapping expressions, the new name should still be used, as it conveys the intent of the API design, and it will simplify/clarify this CodeType/WrappedType.
> I suspect if you do these things, then the need for a stateful oracle will go away.

I think you're saying that we could have `WrappedType` not store the config, and then make `imports()` and `helper_code()` input it as an argument.  That was my initial plan, but there's one very annoying issue there: Each language has a different `Config`, which makes adding it as an argument to the `CodeDeclaration` trait tricky.  I think there's a way to parameterize the trait, but it's not so simple.  I also thought about just using `Any` as the parameter type, but I didn't like that so much.

examples/wrapper-types/src/wrapper-types.udl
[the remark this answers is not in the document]
jhugman:
```udl
/// We want to interoperate between `java.util.Timestamp` type and a `chrono`
/// Uniffi can transport a dictionary
dictionary InternalTimestamp {
    i64 seconds;
    i32 nanos;
};

/// Then a toml expression can transform into and from the internal struct
typedef InternalTimestamp Timestamp;
```

> everything needs to be wrapped in a separate typedef with a separate name

Yes, I see that will get tiresome fairly quickly. A quick look at the WebIDL spec: no, there isn't an obvious path forward. Agreed this is an ergonomics issue we shouldn't tackle this minute.

> FWIW, this is very similar to the actual code we just shipped for logins. 

Oh, I'd forgotten about that code. This cleans that up nicely. I figured that an example with Timestamp would show mapping to and from types most obviously out of our control.

examples/wrapper-types/uniffi.toml
```diff
@@ -1,11 +1,45 @@
...
+type_name = "FlatUser"
+wrap = "wrapUserAccount"
+unwrap = "unwrapUserAccount"
+
+
 [bindings.kotlin]
 cdylib_name = "wrapper_types"
+
+# It would be nice to wrap the Json type, but that seems to require 3rd party
+# libraries, which we don't want to depend on for the tests.
```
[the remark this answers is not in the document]
jhugman: or `Regexp`?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/wrapped.rs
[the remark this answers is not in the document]
jhugman: > because there's no separate definition struct

Oh. Perhaps it should, it declares another type.

declare your KotlinWrappedType (arguably a `TypeDefCodeDeclaration`) as a `CodeDeclaration`. 

```rust
#[derive(Template)]
#[template(syntax = "kt", escape = "none", path = "WrappedTemplate.kt")]
pub struct KotlinWrappedType {
    name: String,
    inner: Type,
    config: Option<WrapperConfig>,
}
/// snip
impl CodeDeclaration for KotlinWrappedType {
    fn definition_code(&self, _oracle: &dyn CodeOracle) -> Option<String> {
        Some(self.render().unwrap())
    }

    fn imports(&self, _oracle: &dyn CodeOracle) -> Option<Vec<String>> {
        Some(self.config.as_ref()?.imports.as_ref()?.clone())
    }
}
```

add it to the `members()` method: 

```rust
        .chain(
            ci.iter_wrapped_types() // handwave
                .into_iter()
                .map(|inner| {
                      // handwave, get the wrapper config etc, then make
                    Box::new(wrapped::KotlinWrappedType::new(…)))
                        as Box<dyn CodeDeclaration>
                }),
        )
```

and the imports and helper code do their thing without the `CodeType` needing to have the config file.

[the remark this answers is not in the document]
jhugman: The general rule is the `CodeType` is used when other types need to refer to a type.

The `CodeDeclaration` is used for every type that is declared in the UDL.

So, Objects has both a `CodeType` because other types refer to it, and at least one declaration.

`String` isn't declared in the UDL, so doesn't have a `CodeDecalation`, but does have a `CodeType`.

docs/manual/src/udl/ext_types_wrapped.md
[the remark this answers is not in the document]
jhugman: Cool. I like that you have some flexibility (e.g. you've got `{}.toLogin()` `{}.toLogin()` for the Logins example you gave earlier).

docs/manual/src/udl/ext_types_wrapped.md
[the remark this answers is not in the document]
jhugman: I like the `from` and `into` formulation; I'd been think `from_raw` or `from_lower`.

I like the `Type::Typedef`; though it may be a  bit too general — we may find some clever use for `typedef`s in the future.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/wrapped.rs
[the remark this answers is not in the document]
bendk: That makes sense and I think that would work for this change, but I like having `CodeOracle` store the `Config`.  It seems like it enables a lot of things and it's something we want going forward.  For example I want to allow the config to wrap builtin types like `timestamp` or user-defined structs without needing to define the wrapper in the UDL.  That brings up the same issue of the type name depending on the config.  We could make it so every type always uses a type alias, but seems like it makes the generate code needlessly complicated.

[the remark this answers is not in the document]
jhugman: Isn't this the point of so called External types? Isn't an explicit goal to remove `timestamp` from the set of built in types?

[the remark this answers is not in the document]
bendk: I'm not completely sure about External, but I think they're for a type defined in a different UDL file.  I'm thinking of use-cases like the logins one.  I'm not sure what the plan for `timestamp`, I thought it might be to keep the built-in type, but allow users to override it.

uniffi_bindgen/src/backend/config.rs
```diff
@@ -0,0 +1,18 @@
...
+/// These define an expression to do the conversion, with `{}` as a placeholder for the value being
+/// wrapped/unwrapped.
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct WrapExpression(String);
+
+impl WrapExpression {
+    pub fn render(&self, var: &str) -> String {
+        self.0.replace("{}", var)
+    }
+}
```
jhugman, line 18: Do I understand correctly: this is called `WrapExpression` because it's used for wrapping and unwrapping? How about naming it for what it does (e.g. `TemplateExpr`)?

jhugman (commented):
This is coming along nicely. I think I'm persuaded that putting the `Config` in the `Oracle` is a necessary step. I've put in some comments re simplifying the `WrappedCodeType` when no `WrappedConfig` exists.

I think the blockers still to fix: 

- [x] implement the `WrappedType` as `CodeDeclaration`, and include them in the `members()` function. This further simplifies `WrappedCodeType` to just needing a name.
- [x] fixup the language around `wrapping` and `unwrapping`. Even `Custom` types or `User` types are better than wrapped. :(
- [ ] some documentation fixups
- [ ] an Enum example (e.g. a `JSONObject` as a `JSONEnum` which translates to a `serde_json::Value`)
- [ ] an example re-implementing timestamp.
- [ ] changelog entry

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -66,24 +75,31 @@ impl MergeWith for Config {
...
         }
     }
 }
 
+// Generate kotlin bindings for the given ComponentInterface, as a string.
+pub fn generate_bindings(config: &Config, ci: &ComponentInterface) -> Result<String> {
+    let oracle = KotlinCodeOracle::new(config.clone());
+    filters::set_oracle(oracle.clone());
+
+    KotlinWrapper::new(oracle, config.clone(), ci)
```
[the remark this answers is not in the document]
bendk: It's a legacy name that I would love to change.

fixtures/ext-types/lib/tests/test_generated_bindings.rs
```diff
@@ -1,9 +1,9 @@
...
-);
+// uniffi_macros::build_foreign_language_testcases!(
+//     [
+//         "../guid/src/guid.udl",
+//         "../../../examples/wrapper-types/src/wrapper-types.udl",
+//         "../uniffi-one/src/uniffi-one.udl",
+//         "src/ext-types-lib.udl",
+//     ],
+//     ["tests/bindings/test_imported_types.py",]
+// );
```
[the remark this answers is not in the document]
bendk: It's commented out until the external types are working again on python.  I don't think they ever were working great and the uoc refactor plus this change broke the tests.  I really hope to get them working again soon.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -219,10 +241,10 @@ impl KotlinCodeOracle {
                 Box::new(compounds::MapCodeType::new(inner, outer))
             }
             Type::External { .. } => panic!("no support for external types yet"),
-            Type::Wrapped { name, prim } => Box::new(wrapped::WrappedCodeType::new(
-                name,
-                self.create_code_type(prim.as_ref().clone()),
-            )),
+            Type::Wrapped { name, inner } => {
+                let config = self.config.wrappers.get(&name).cloned();
+                Box::new(wrapped::WrappedCodeType::new(name, *inner, config))
```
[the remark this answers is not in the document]
bendk: My goal here is to use the FfiConverter class so that the lift/lower/read/write functions are always in the same place.  This would be one way to implement external types.  But it's kind of weird having only this code follow that pattern and I'm actually not sure how external types are going to shake out.

In other words, yeah this can just be `self.find(inner)` :)

bendk:
Seems like we're most of the way there.  I'll do some work to update the examples/docs.  Let me think on the naming a bit.

I can definitely add a timestamp replacement example.

What's your vision for `serde_json::Value`?  We could implement that as an enum right now if we wanted.  On Python I think we could write some code that converts the enum into a regular python object.  But I'm not sure what we would do with it on Swift/Kotlin.

jhugman:
I'm looking for examples and docs which are comprehensive but feel like they're clear and grounded in reality. I find the [Swift book][1] _excellent_ for this.

[1]: https://docs.swift.org/swift-book/LanguageGuide/Enumerations.html
[2]: https://doc.rust-lang.org/stable/book/ch06-01-defining-an-enum.html

So adding the timestamp example falls into the "feels like [it's] grounded in reality" category, and adding examples for going via an enum (and primitives, objects, dictionaries) fall into the "comprehensive" category.

> I can definitely add a timestamp replacement example.

Yay.

> What's your vision for serde_json::Value? We could implement that as an enum right now if we wanted. On Python I think we could write some code that converts the enum into a regular python object.

In #396 the OP wants a canonical serialization of a JSON object, so needs to pass through a representation of JSON that _isn't_ serialized from strings.

```
[Enum]
interface InternalJSONValue {
    Null();
    Boolean(boolean v);
    String(string v);
    Int(i32 v);
    Array(sequence<InternalJSONValue> v);
    Object(dictionary</* string, */ InternalJSONValue> v);
};

typedef InternalJSONValue JSONObject;

namespace json {
    string to_string(JSONObject obj);
    JSONObject from_string(string str);
}
```

Then add a transformer between `InternalJSONValue` and `JSONObject`.

#396 was an actual user issue, so that's why I went to going in and out of the `serde_json::Value` to a non-generated type. 

If you have another example with a different enum it would work equally well. My mind is going to ASTs and UI View hierarchies now, but that might be unnecessarily complicated.

(I do see that #411 means recursive enums aren't supported yet in Swift, so a simpler example may be needed for now).

docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -1,29 +1,55 @@
...
 
-It's possible to expose types in your Rust crate to UniFFI, so long as the type can
-be expressed as one of UniFFI's [builtin_types](./builtin_types.md).
+`Wrapped` types allow you to extend the UniFFI type system with custom types.
+These types rely on a "normal" UDL type move data across the FFI, then use a
+custom type to wrap those types on the scaffolding and/or bindings side.
 
-For example, consider the following trivial Rust abstraction for a "handle" which wraps an
+Some examples:
+  - You have Rust type named `Handle` which uses the new-type pattern to wrap a
```
mhammond, line 10: nit: Describing a `Handle` as wrapping a string breaks my mental model of what a handle usually represents. Is there a reason why you changed it from wrapping an int? Having something other than a `String` being described seems worthwhile.

(I just noticed the example for handle below did stick with ints)

docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -1,29 +1,55 @@
...
+    string and provides additional functionality.  We can have `Handle` wrap
+    `String` on the scaffolding side so that the Rust code gets to use that
+    functionality, but on the bindings side the type behaves like a normal
+    string.
+  - You've updated a `Record` in your Rust library, but your bindings code is
+    still using the legacy version of it.  You can wrap the `Record` on the
+    bindings side with a type that provides the legacy API.
+  - You want to use `Json` to exchange data between your Rust library and it's
```
mhammond, line 18: "both sides with a `JsonObject`" isn't clear to me - I think I know what you mean, but that's just because we've spent months talking about this :) Specifically, `JsonObject` has no meaning to Python nor to Rust. I think generalizing this might help - something like "The string can be wrapped with whatever type makes sense on both sides -for example, the Rust side might use a `serde_json::Value` enum whereas your [Kotlin code might use a {whatever}|Python code might use objects returned by the `json` module]"

docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -57,13 +83,86 @@ Because `UniffiCustomTypeWrapper` is defined in each crate, this means you can w
...
+# Function that converts `User` to `LegacyUser`
+wrap = "wrapUser({})"
+# Function that converts `LegacyUser` to `User`
+unwrap = "unwrapUser({})"
+# Where we can import those functions from
+imports = [
+  "my.wrappers.package.wrapUser",
+  "my.wrappers.package.unwrapUser",
+]
+```
```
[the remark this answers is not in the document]
mhammond: Yeah, Ben and I did throw a few names around - eg, `Custom` or `CustomType` - which did seem better than wrapping when talking about primitive types, but it's still not quite clear in this example (ie, "User" and "LegacyUser" are both "custom types" by many definitions).

TBH I'm not sure how I feel about this example (although I do understand where it came from :) This kinda sits exactly 1/2 way between the original "wrapped types" idea and the "decorator" idea. I need to get my head around it some more.

(Maybe my only problem is that this example seems too convoluted for the manual - the fact this scheme can support such a use-case might be seen as a strength, but that doesn't mean it's a "good" example. Maybe this is better demonstrated as code in, say, examples/advanced-custom-types, and the docs just say "for more advanced examples, see ...")

docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -57,13 +83,86 @@ Because `UniffiCustomTypeWrapper` is defined in each crate, this means you can w
...
+
+`wrap` and `unwrap` specifies the code used to wrap/unwrap the type.  `{}`
+will be replaced with the value to be wrapped or unwrapped.
+
+`my.wrappers.package` would have code like this:
+
+```kotlin
+fun wrapUser(user: User) = LegacyUser(
```
mhammond, line 136: as Ben and I very briefly discussed, maybe leaning into "lift" and "lower" is actually what we should be doing here - those terms don't (afaik) come with existing baggage, and once explained are quite clear IMO. ie, it's a slightly steeper learning curve for new people, but there's less confusion once you've climbed it.

(OTOH, unless the semantics actually are identical to how uniffi already uses that term, the end result could be more confusion rather than less, which is clearly a terrible outcome - that's not clear to me yet)

docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -57,13 +83,86 @@ Because `UniffiCustomTypeWrapper` is defined in each crate, this means you can w
...
+
+fun unwrapUser(user: LegacyUser) = User(
+   fullName = user.name,
+   age = user.yearsOld,
+)
+```
+
+## Wrapping on both the scaffolding and bindings side
+
+Of course, it's possible to wrap on both the scaffolding and bindings side.
```
mhammond, line 149: nit: drop "Of course" - it's like saying "Obviously", which often isn't :) Also, this just seems to restate what's at the start of this section - eg, the sentence "allows you to define types that feel like new builtin types" is important, but it's buried here. This para and the discussion about the json should either be deleted or moved up where you talk about the handle. (before your patch, the handle and json object discussions were very close - now the whole "user" stuff is between them, which seems wrong)

examples/wrapper-types/src/lib.rs
```diff
@@ -1,8 +1,9 @@
-// We use `JsonObject` in our UDL. It moves to and from Uniffi bindings via a string.
-pub type JsonObject = serde_json::Value;
+// Trivial "Handle" type which wraps an i64.
+pub struct Handle(pub i64);
 
-// We must implement the UniffiCustomTypeWrapper trait.
-impl UniffiCustomTypeWrapper for JsonObject {
+// We must implement the UniffiCustomTypeWrapper trait for each type that we want to wrap on the
+// rust side
+impl UniffiCustomTypeWrapper for serde_json::Value {
```
mhammond, line 6: I must be missing something here - how does `JsonObject` in the UDL work now that you've removed the type alias?

examples/wrapper-types/src/lib.rs
```diff
@@ -29,16 +39,39 @@ impl UniffiCustomTypeWrapper for Handle {
...
+pub struct UserAccount {
+    info: UserInfo,
+    account: AccountInfo,
+}
+
+pub struct UserInfo {
+    name: String,
+}
+
+pub struct AccountInfo {
```
mhammond, line 52: I think this example is over wrought - is `AccountInfo` adding value here?

examples/wrapper-types/src/wrapper-types.udl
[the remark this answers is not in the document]
mhammond: I'm inclined to agree with this - as above, I think the user wrapping example is confusing enough to us that it will be intractable for people new to uniffi.

examples/wrapper-types/src/wrappers.kt
```diff
@@ -0,0 +1,17 @@
...
+
+fun wrapUserAccount(user: UserAccount) = FlatUser(
+    userName = user.info.name,
+    accountId = user.account.id,
+)
+
+fun unwrapUserAccount(user: FlatUser) = UserAccount(
+    info = UserInfo(name = user.userName),
+    account = AccountInfo(id = user.accountId)
+)
```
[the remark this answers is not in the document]
mhammond: agreed - this would be better in a more "advanced" example

examples/wrapper-types/src/wrappers.swift
```diff
@@ -0,0 +1,34 @@
...
+        self.accountId = accountId
+    }
+}
+
+extension FlatUser: Equatable, Hashable {
+    public static func == (lhs: FlatUser, rhs: FlatUser) -> Bool {
+        return lhs.userName == rhs.userName && lhs.accountId == rhs.accountId
+    }
+
+    public func hash(into hasher: inout Hasher) {
```
mhammond, line 16: I don't quite understand how this is relevant to the example? And even the implementation seems suspect - what does the same accountId but different name mean? :)

examples/wrapper-types/uniffi.toml
```diff
@@ -1,11 +1,58 @@
 [bindings.python]
 cdylib_name = "wrapper_types"
 
+[bindings.python.wrappers.JsonObject]
+imports = ["json"]
```
mhammond, line 5: not going to die on this hill, but I still find "imports" as too prescriptive. A common (and often significant) Python optimization here would be moving from `import json`/`json.loads(blah)` to `from json import loads`/`loads(blah)` - but this prevents me from expressing that. That optimization gets even more significant on `deep.packages.with.many.modules`

examples/wrapper-types/uniffi.toml
```diff
@@ -1,11 +1,58 @@
...
+]
+type_name = "FlatUser"
+wrap = "wrapUserAccount({})"
+unwrap = "unwrapUserAccount({})"
+
+[scaffolding.wrappers.Handle]
+type_name = "Handle"
+
+[scaffolding.wrappers.JsonObject]
+type_name = "serde_json::Value"
```
mhammond, line 55: hrm - I think this answers my question above, but I don't think I like it. IIUC, unlike the "foreign" side of the world, it would be literally impossible to change this to anything else, because all the rust side of the world must already use this.  IOW, the reason I like the foreign types being in this .toml is that they *can* change the type, but it's literally impossible for the Rust side, right?

mhammond (commented):
Sorry I didn't get very far here, and sorry most of my comments sound negative - I really appreciate the effort you are putting in to this and I think we are heading in the right direction. The best terminology still escapes me, and I fear some of the examples and documentation are heading off into the weeds and might risk obscuring the best path forward - as the saying goes, I've a sense that I can't see the forest for the trees.

I'll find more time next week to dig into this, but thanks for your perseverance Ben!

docs/manual/src/udl/ext_types_wrapped.md
jhugman: I have similar feels about lift and lower; as in: I like and prefer those terms, but worry about approachability.

> which is clearly a terrible outcome - that's not clear to me yet

😄

examples/wrapper-types/uniffi.toml
jhugman: I like the symmetry here. However, I'm confused where the unwrap/wrap expressions are for the scaffolding.

> but it's literally impossible for the Rust side, right?

I'm 95% in a agreement with this, but in the context of `impl`/`dyn` and `type` aliasing, I'm less sure.

mhammond: Having this in the .toml is like saying "this option exists in the '.toml tree' for maximum configurability. However, it is impossible to configure it." :)

mhammond (commented):
I think this is great!

Sorry I kept repeating myself about the config influencing the scaffolding - but I'm really not a fan - I think it's a solution in search of a problem and it has (IIUC) unintended edge-cases. It will be a lot easier to add than to take away in the future.

If this was removed, we'd really just have 3 main changes here:

* The rust scaffolding changes are actually very minor if we ignore the config. I think these are quite non-controversial and could probably land independently.
* The foreign bindings side, which looks great to me, although I really didn't look in detail at the implementation. There's a lot of boilerplate, which is unfortunate, but not the cause of this patch.
* The docs, which as I mention, think veer a little too far into describing how far you can push this, rather than focusing on explaining the key concepts - IOW, I think the docs are aimed more at the reviewers of this patch than people new to uniffi :)

So what *I'd* love to see happen is something like:

* ideally agree to leave rust config off the table for now, then land all the core uniffi scaffolding changes.
* markh to better understand how "external" types fit in this world - it's still not quite clear to me how the foreign bindings work with external types, nor whether wrapping external types works correctly (it probably does, this is just a reflection on me not understanding that part correctly)
* We all brainstorm on better names! I think everyone can unite in their hatred for "wrap" We then land a "great rename". I'm starting to lean towards `Type::CustomLocal` (currently `Type::Wrapped`) and `Type::CustomExternal` (currently `type::External`) (or maybe `LocalCustom` and `ExternalCustom`?) and `lift/lower` for the actual operations, but this is a very weak preference and I'm bad at names, and primarily responsible for the names we currently have and hate!, so take care :)) We could even make a release here, deprecating the old name and removing it in the next build and before we do the following.
* land the foreign bindings, docs and examples using the new names.
* profit?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -25,13 +25,21 @@ mod primitives;
 mod record;
 mod wrapped;
 
-// Some config options for it the caller wants to customize the generated Kotlin.
-// Note that this can only be used to control details of the Kotlin *that do not affect the underlying component*,
```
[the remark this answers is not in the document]
mhammond: (to be clear, I was more saying "I hope the kotlin config can't change the generated rust" which is what I think remains true in this patch)

examples/wrapper-types/src/wrappers.swift
bendk: I don't love this either, but it's needed because `WrappedTypesDemo` wants to implement `Hashable` by calling `hash()` on each of it's components.  I guess it's just one more reason why this is not a great example.

examples/wrapper-types/uniffi.toml
bendk: I'd like to change `imports` to specify the entire import statement.  This lets us use the `from ... import` version on Python and I think it could be an improvement for other languages as well.

bendk:
I like the idea of limiting this to just the foreign bindings side and also deferring the decision to come up with names.  One reason is because I really want to explore the idea of allowing the foreign bindings to wrap any type, not just `Type::Wrapped` or whatever we rename it to.  If that was possible to do and something we wanted to support, then that affects both the naming decision and the scaffolding wrapper config one.

So what do you think of this plan:
- For this PR:
  - Remove the scaffolding changes
  - Remove the `User/FlatUser` example
  - Implement the WrappedType as CodeDeclaration
  - Don't change any names yet (I'll change `inner` back to `prim`)
  - Add Changelog entry, fixup the docs
- Future Work:
  - Flesh out external types
  - Implement support for wrapping any type on the bindings side.  This is where I would want to put advanced examples for  `User`/`FlatUser` example and reimplementing timestamp.
  - Figure out the names

bendk:
> > What's your vision for serde_json::Value? We could implement that as an enum right now if we wanted. On Python I think we could write some code that converts the enum into a regular python object.
> 
> In #396 the OP wants a canonical serialization of a JSON object, so needs to pass through a representation of JSON that _isn't_ serialized from strings.
> ...
> If you have another example with a different enum it would work equally well. My mind is going to ASTs and UI View hierarchies now, but that might be unnecessarily complicated.
> 
> (I do see that #411 means recursive enums aren't supported yet in Swift, so a simpler example may be needed for now).

My issue is I can't really think of good real-world style examples for this one.  I'm definitely open to suggestions though.

bendk:
Just pushed some updated commits for this.  It's basically the plan from my last comment, with a few differences:
  - After talking with both of you, it seemed like we should actually try to get the names correct before landing this one.  I made an attempt at this, mostly based on the things we agree on.  Really hoping these are okay, but feel free to push back.  I do think getting the names right is important.
     - I went with `into_custom()` and `from_custom()` for the functions, but the lift/lower terminology also sounded good to me and I'm happy to change it to something like that.
  - I tried to use good examples in the example app.  I removed the `JsonObject` example, since it's not totally clear how this should be handled as an example for swift/kotlin.  I think the `Url` one is doing basically the same thing -- a conversion on both the scaffolding and bindings side.
 - We don't even need the `extra_scripts` functionality anymore.  Maybe I'll package that up into a separate PR.


I'm not sure if this should go into the same release as #1127 or not.  Either way, I think we should slightly tweak the CHANGELOG.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/wrapped.rs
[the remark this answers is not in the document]
mhammond: >  Isn't an explicit goal to remove timestamp

FTR, I don't think it's necessarily an explicit goal - it would be nice, but I'm not sure it will be possible to replace them without breaking code. My view of the goal is something like "it would have been possible use this PR to get the same functionality without introducing a new type" or something similar.

examples/custom-types/Cargo.toml
```diff
@@ -8,14 +8,15 @@ publish = false
...
-name = "wrapper_types"
+name = "custom_types"
 
 [dependencies]
 anyhow = "1"
 bytes = "1.0"
 serde_json = "1"
 uniffi_macros = {path = "../../uniffi_macros"}
 uniffi = {path = "../../uniffi", features=["builtin-bindgen"]}
+url = "2.0"
```
mhammond, line 19: see below re this needing to be "2.2"

uniffi_build/src/lib.rs
```diff
@@ -26,6 +26,7 @@ use std::process::Command;
 /// command-line tool. This is mostly useful for developers who are working on uniffi
 /// itself and need to test out their changes to the bindings generator.
 pub fn generate_scaffolding(udl_file: &str) -> Result<()> {
+    println!("cargo:rerun-if-changed=uniffi.toml");
```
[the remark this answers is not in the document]
mhammond: I think this line can die now?

fixtures/ext-types/lib/tests/test_generated_bindings.rs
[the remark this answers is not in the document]
mhammond: They are very close - I had a play, but it wasn't obvious - but I think this can land as-is and we fix it up later.

mhammond (approved):
This is absolutely fantastic Ben! I really like the naming we've managed to land on and I think it is very cohesive and thorough. Please treat all my doc comments as suggestions, and get approval from @jhugman before landing, but I think this is 🚀

fixtures/ext-types/lib/Cargo.toml
```diff
@@ -20,9 +20,9 @@ uniffi-one = {path = "../uniffi-one"}
 ext-types-guid = {path = "../guid"}
 
 # Reuse one of our examples.
-wrapper-types = {path = "../../../examples/wrapper-types"}
+custom-types = {path = "../../../examples/custom-types"}
 
-serde_json = "1"
+url = "2.0"
```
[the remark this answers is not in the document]
mhammond: semver confuses me - maybe the correct fix is just "2"?

docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -125,15 +124,40 @@ The behavior of the generated scaffolding will be:
...
+
+However, custom types can also have be converted on the bindings side.  If you
+have a custom Kotlin Handle class you can configure UniFFI to use that for the
+`Handle` type by adding code like this to `uniffi.toml`:
+
+```toml
+[bindings.kotlin.custom_types.Handle]
+# Kotlin custom type name
+type_name = "Handle"
+# Expression to convert Strings to Handles.  `{}` will be replaced with the string value
```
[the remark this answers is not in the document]
bendk: Oops, I don't know why I keep messing up that type name.  And yes, the `{}` gets replaced with an expression that evaluates to a String / Handle.

docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -125,15 +124,40 @@ The behavior of the generated scaffolding will be:
...
+```
 
-In the example above, the foreign bindings just see the "wrapped" value - eg, the bindings will
-get an integer for the `Handle`.
+Here's how the configuration works in `uniffi.toml`.
 
-We intend on making it possible to customize this - for example, `serde_json::Value` could be
-transformed into a suitable JSON object in the bindings.
+* Create a `[bindings.{language}.custom_types.{CustomTypeName}]` table to enable a custom type on a bindings side.  This has several subkeys:
+  * `type_name` (Typed languages only): Type/class name for the custom type
```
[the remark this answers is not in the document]
bendk: I think that's good as a default, but I think in practice users will often need to set it anyways.  `Url` needed to be changed to `URL` in the example.  I think `JsonObject` will need its type specified too.

So, I made `type_name` optional, but kept the docs the same.  It seems better to risk giving users more info then they need than the other way around.

bendk:
Thanks for the review!  Just pushed some updated commits, I think they address all the comments.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -219,9 +245,10 @@ impl KotlinCodeOracle {
                 Box::new(compounds::MapCodeType::new(inner, outer))
             }
             Type::External { .. } => panic!("no support for external types yet"),
-            Type::Wrapped { name, prim } => Box::new(wrapped::WrappedCodeType::new(
-                name,
-                self.create_code_type(prim.as_ref().clone()),
+            Type::Custom { name, builtin } => Box::new(custom::CustomCodeType::new(
+                name.clone(),
+                builtin.as_ref().clone(),
+                self.config.custom_types.get(&name).cloned(),
```
jhugman, line 251: I wonder if you can simplify things here by checking 

```rust
if Some(config) = self.config.custom_types.get(&name) {
    Box::new(custom::CustomCodeType::new(name.clone(), builtin.as_ref().clone(), config.clone())
} else {
    self.create_code_type(builtin.as_ref().clone())
}
```

and only invoke `CustomType` templates and `CodeType` and `CodeDecalration` when the config exists.

uniffi_bindgen/src/bindings/kotlin/templates/CustomType.kt
```diff
@@ -0,0 +1,43 @@
+object FfiConverterType{{ name }} {
+    {%- match config %}
+    {%- when None %}
+    {#- No custom type config, just forward all methods to our builtin type #}
```
jhugman, line 4: Logic in the template which affects almost the whole of the file indicate we should probably have split this template in two.

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -24,14 +25,20 @@ mod miscellany;
...
     cdylib_name: Option<String>,
+    #[serde(default)]
+    custom_types: HashMap<String, CustomTypeConfig>,
+}
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct CustomTypeConfig {
+    imports: Option<Vec<String>>,
+    into_custom: TemplateExpression,
+    from_custom: TemplateExpression,
```
jhugman, line 41: I _think_ we might want to move this to a more common spot once we get land this.

jhugman (commented):
From the list above:

- [ ] an Enum example (e.g. a `JSONObject` as a `JSONEnum` which translates to a `serde_json::Value`)
- [ ] an example re-implementing timestamp.

It looks like you've punted on "customize any type", which would allow these two items; this is probably sensible.

If there isn't one already, could you file issues for these two items, and a "customize any type" piece of work.

FWIW I'm hoping this mean we can customize each side of the FFI independently of one another: IIUC there still seems to be some unnecessary coupling between the two.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
bendk: This gets into a requirement that I've been half-committed to: having a consistent FFI conversion API for all types.  The end goal there would be that you can support external types by importing a `FFIConverter` subclass and calling it's methods, without knowing much about the underlying type.  I say "half-committed" because I'm not really sure how it's all going to turn out, but I'm going to try at least.

uniffi_bindgen/src/bindings/kotlin/templates/CustomType.kt
bendk: Agreed, although it's not so easy to do this currently.  I'll try to refactor this as part of #1089.

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
bendk: I was wondering about that.  The only reason not to is that the fields vary slightly between different languages (python doesn't have `type_name`, on kotlin/swift it's optional).  That's not all that compelling though, I could go either way on this one.

bendk:
Fixed the typo and made some issues: #1138, #1139, djc/askama#575.  Maybe we can get this merged soon?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -121,6 +137,10 @@ impl<'a> KotlinWrapper<'a> {
                         as Box<dyn CodeDeclaration>
                 }),
         )
+        .chain(ci.iter_custom_types().into_iter().map(|(name, type_)| {
+            let config = self.config.custom_types.get(&name).cloned();
+            Box::new(custom::KotlinCustomType::new(name, type_, config)) as Box<dyn CodeDeclaration>
+        }))
```
jhugman, line 143: 😻

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
jhugman: I had a little go, and it seemed like a string that we could pull on in a followup. 

I think a unified `FFIConverter<T, U>` interface is a good goal too, but I think that's orthogonal to this.

uniffi_bindgen/src/bindings/kotlin/templates/CustomType.kt
```diff
@@ -0,0 +1,43 @@
...
+        val builtinValue = ({{ "value"|lift_var(builtin) }})
+        return {{ config.into_custom.render("builtinValue") }}
+    }
+
+    fun lower(value: {{ self.type_name(config) }}): {{ self.builtin_ffi_type()|ffi_type_name }} {
+        val builtinValue = {{ config.from_custom.render("value") }}
+        return {{ "builtinValue"|lower_var(builtin) }}
+    }
+    {%- endmatch %}
+}
```
jhugman, line 43: FWIW: this file should be called `CustomTypeTemplate.kt`.

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
jhugman: Hmm, yes. Worth a comment perhaps?

docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -125,15 +124,40 @@ The behavior of the generated scaffolding will be:
...
+]
+```
 
-In the example above, the foreign bindings just see the "wrapped" value - eg, the bindings will
-get an integer for the `Handle`.
+Here's how the configuration works in `uniffi.toml`.
 
-We intend on making it possible to customize this - for example, `serde_json::Value` could be
-transformed into a suitable JSON object in the bindings.
+* Create a `[bindings.{language}.custom_types.{CustomTypeName}]` table to enable a custom type on a bindings side.  This has several subkeys:
```
jhugman, line 154: In the context of "any type can be customized", I'm not sure I understand what the point of the`[Custom]` attribute is to the foreign language bindings.

This isn't something for this PR just asking the question early.

docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -125,15 +124,40 @@ The behavior of the generated scaffolding will be:
 * Calling `take_handle_2` with a value of `-1` will always panic.
 * All other values will return `Ok(ExampleHandle)`
 
-## Custom Foreign Wrappers
+## Custom types in the bindings code
+
+*Note: The facility described in this document is not yet available for the Ruby bindings.*
+
+By default, the foreign bindings just see the builtin type - eg, the bindings will get an integer
+for the `Handle`.
```
jhugman, line 132: Perhaps use a different example here, suggest URL or UUID.

I got confused because `Handle` is not customized in the `examples/` but is in this documentation.

examples/custom-types/tests/bindings/test_custom_types.kts
```diff
@@ -0,0 +1,19 @@
...
+import java.net.URL
+
+import customtypes.*
+
+// TODO: use an actual test runner.
+
+// Get the custom types and check their data
+val demo = getCustomTypesDemo(null)
+assert(demo.url == URL("http://example.com/"))
+assert(demo.handle == 123L)
```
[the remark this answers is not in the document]
jhugman: Could definitely use a comment to say this isn't expected to be customized, and falls back to the Long.

jhugman (approved):
I think this is ready to land. Well done, and thank you for your patience and responsiveness.

🎉 🚀 🎸 🎪 🎿

bendk:
Merged with some minor changes.  Thanks :rocket:

PR #1108, opened 2021-11-19
Added a check on Kotlin for RustBuffer.alloc()

bendk:
I want to check out a hunch that this is returning null sometimes, which
causes issues like
https://github.com/mozilla/application-services/issues/4640.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/RustBufferTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/RustBufferTemplate.kt
@@ -13,7 +13,11 @@ open class RustBuffer : Structure() {
 
     companion object {
         internal fun alloc(size: Int = 0) = rustCall() { status ->
-            _UniFFILib.INSTANCE.{{ ci.ffi_rustbuffer_alloc().name() }}(size, status)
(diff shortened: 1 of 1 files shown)


badboy (approved):
Huh, that's weird.
I don't really see a way that `RustBuffer` returns an null pointer. Even an empty vec should use `0x1` as the dangling data pointer.
Further the `NullPointerException` in the linked bug is on the Kotlin side when calling `lower` which makes me believe that whatever `lower` is called on is a null pointer?

Regardless in that context this PR seems harmless, so I'm not opposed

bendk:
> Huh, that's weird. I don't really see a way that `RustBuffer` returns an null pointer. Even an empty vec should use `0x1` as the dangling data pointer. Further the `NullPointerException` in the linked bug is on the Kotlin side when calling `lower` which makes me believe that whatever `lower` is called on is a null pointer?
> 
> Regardless in that context this PR seems harmless, so I'm not opposed

I don't see a way either, but I also don't see a way that lower should fail either.  In a-c code there's a null check just before the call.  It all seems 100% impossible to fail, but somehow it is :)  No clue if this will identify the issue, but I guess there's a chance.

PR #1110, opened 2021-11-22
Update to new version of cargo release which has better workspace support

mhammond:
This seems to work well with a newer `cargo-release` - so well I accidentally published 0.15.0 to crates.io 😅 which I've since yanked.

(I'd like to make a new 0.15.1 for use by app-services, so after approval and landing, feel free to test it out! Otherwise I'll cut a new release tomorrow.)


bendk (approved):
Easier releases sounds great to me.

PR #1114, opened 2021-11-24
Remove periods in the canonical type name for Timestamps and Duration…

mhammond:
… in Kotlin.

Previously, optional timestamps would generate, eg:

`internal fun writeOptionaljava.time.Instant(v: java.time.Instant?, buf: RustBufferBuilder) ...`

and code which calls it:

`writeOptionaljava.time.Instant(v, buf)`

which upset the Kotlin compiler, causing errors like:

`Unresolved reference: writeOptionaljava`

(as found in app-services!)

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/miscellany.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/miscellany.rs
@@ -58,7 +58,7 @@ macro_rules! impl_code_type_for_miscellany {
 
 impl_code_type_for_miscellany!(
     TimestampCodeType,
-    "java.time.Instant",
+    "Instant",
     "Timestamp",
     vec!["java.time.Instant", "java.time.DateTimeException"],
     "TimestampHelper.kt"
@@ -66,7 +66,7 @@ impl_code_type_for_miscellany!(
 
 impl_code_type_for_miscellany!(
     DurationCodeType,
-    "java.time.Duration",
+    "Duration",
     "Duration",
     vec!["java.time.Duration", "java.time.DateTimeException"],
     "DurationHelper.kt"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ## [[UnreleasedVersion]] - (_[[ReleaseDate]]_)
 
+### What's Changed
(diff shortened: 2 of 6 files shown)


bendk (approved):
Looks great!

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/miscellany.rs
```diff
@@ -58,15 +58,15 @@ macro_rules! impl_code_type_for_miscellany {
...
+    "Instant",
     "Timestamp",
     vec!["java.time.Instant", "java.time.DateTimeException"],
     "TimestampHelper.kt"
 );
 
 impl_code_type_for_miscellany!(
     DurationCodeType,
-    "java.time.Duration",
+    "Duration",
```
jhugman, line 69: Couldn't comment where the actual fix needed so commenting here. 

This changes the `$type_label`, instead of using the `$canonical_name`.

That should be used in the `canonical_name` method, but that's missing. 

The correct fix is to add that method (say at [L24][1]).

```rust
                 fn canonical_name(&self, _oracle: &dyn CodeOracle) -> String {
                     $canonical_name.into()
                 }
```

The fully qualified class name should be available as a type label, so as to avoid collisions with UDL declared types. e.g. if a UDL type is called `Instant` or `Duration`.

[1]: https://github.com/mozilla/uniffi-rs/pull/1114/files#diff-326998556286e54771636144110306a7bf185b8d277ff93208998226cb90116cR24

bendk: Adding the canonical_name method makes a lot of sense.  If the type name is going to be fully qualified, then should we remove the imports?

mhammond: I don't quite understand the concrete suggestion here, and if the intent is to allow a `Duration` type to be defined in a UDL and not conflict with this `Duration`, I think things will need to get even more sophisticated - `writeOptionalInstant` will conflict - so this case will want to do something like `writeOptionaljava_time_Instant`.

Given these problems already exist, that this is fixing a regression which is preventing app-services from upgrading, and because this change doesn't prevent fixing this correctly, I've decided to merge this as is, and I'll open an issue for that "correct" fix.

PR #1117, opened 2021-11-25
Clarify some build instructions

mhammond:
Things I learnt doing a point release. The git instructions can probably be simplified, especially in the "normal" case, but we can't assume that. Please let me know if you have suggestions.

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -11,12 +11,26 @@ It's not (yet) quite an ideal fit for our workflow, but it helps! Steps:
    * Anything that could cause a consumer project to behave differently if it upgraded
      to the new version, should be called out as a breaking change and should trigger
      a minor version bump (we're below `v1.0` for semver purposes).
-1. Start a new branch for the release. We typically use a single branch for all point releases,
-   so it should be named something like `release-v0.6.x`:
-    * `git checkout -b release-v{MAJOR}.{MINOR}.x`
-    * `git push -u origin release-v{MAJOR}.{MINOR}.x`
+
+1. Identify the branch for the release. We typically use a single branch for all point releases,
+   so it should be named something like `release-v0.6.x`. This also means that if you are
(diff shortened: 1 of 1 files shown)


docs/release-process.md
```diff
@@ -11,12 +11,25 @@ It's not (yet) quite an ideal fit for our workflow, but it helps! Steps:
...
-    * `git push -u origin release-v{MAJOR}.{MINOR}.x`
+
+1. Identify the branch for the release. We typically use a single branch for all point releases,
+   so it should be named something like `release-v0.6.x`. This also means that if you are
+   making a point release, the branch will already exist.
+
+   * If this is the first minor release, then create the branch:
+      * `git checkout -b release-v{MAJOR}.{MINOR}.x`
+      * `git push -u origin release-v{MAJOR}.{MINOR}.x`
+   * If you are making the branch, merge main into it.
```
skhamis, line 22: I guess the only thing I'm wondering is when someone would be making a release not based on latest main? since it'll most likely have the latest code. More of a question than a blocker though.

skhamis (approved):
Nice to get fresh eyes on the release process! Even though I've done a couple of versions it's always good to take a step back and see where the gaps of documentation are.

docs/release-process.md
mhammond: I had bad wording there - let me know if that updated version helps. What I'm trying to convey here is:

* The branch might exist locally (eg, when you did a 0.x.1 release) and also exist remotely but be different (eg, someone else did a 0.x.2 release) - you want to do 0.x.3, but need to make sure the branch you have locally is up-to-date.
* You then want to merge current main into it.

Suggestions how I can clarify that would be most welcome!

skhamis: That small change actually clarifies quite a bit! I think it is a great addition to the documentation -- thanks!

PR #1118, opened 2021-11-25
Use canonical_name in miscellany.rs instead of type label

### LINKED ISSUE #1115: Improve canonical type name patch from #1114
This changes the `$type_label`, instead of using the `$canonical_name`.

That should be used in the `canonical_name` method, but that's missing. 

The correct fix is to add that method (say at [L24][1]).

```rust
                 fn canonical_name(&self, _oracle: &dyn CodeOracle) -> String {
                     $canonical_name.into()
                 }
```

The fully qualified class name should be available as a type label, so as to avoid collisions with UDL declared types. e.g. if a UDL type is called `Instant` or `Duration`.

[1]: https://github.com/mozilla/uniffi-rs/pull/1114/files#diff-326998556286e54771636144110306a7bf185b8d277ff93208998226cb90116cR24

_Originally posted by @jhugman in https://github.com/mozilla/uniffi-rs/pull/1114#discussion_r756345671_

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-116)
┆friendlyId: UNIFFI-116

jhugman:
Fixes #1115 and #1114.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/miscellany.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/miscellany.rs
@@ -11,7 +11,7 @@ use std::fmt;
 use super::filters;
 
 macro_rules! impl_code_type_for_miscellany {
-     ($T:ty, $class_name:literal, $canonical_name:literal, $imports:expr, $template_file:literal) => {
+     ($T:ty, $class_name:literal, $canonical_name:literal, $template_file:literal) => {
          paste! {
              #[derive(Template)]
              #[template(syntax = "kt", escape = "none", path = $template_file )]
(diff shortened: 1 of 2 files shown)


uniffi_bindgen/src/bindings/swift/gen_swift/miscellany.rs
```diff
@@ -22,6 +22,10 @@ macro_rules! impl_code_type_for_miscellany {
                      $class_name.into()
                  }
 
+                 fn canonical_name(&self, _oracle: &dyn CodeOracle) -> String {
+                    $canonical_name.into()
+                }
+
```
jhugman, line 28: I've also checked Python which has an implementation already and Ruby which hasn't moved to the unit-o'-code pattern.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/miscellany.rs
```diff
@@ -58,16 +62,16 @@ macro_rules! impl_code_type_for_miscellany {
 
 impl_code_type_for_miscellany!(
     TimestampCodeType,
-    "Instant",
+    "java.time.Instant",
     "Timestamp",
-    vec!["java.time.Instant", "java.time.DateTimeException"],
+    vec!["java.time.DateTimeException"],
```
bendk, line 67: It could be nice to remove this import as well, just in case the user has a `DateTimeError` struct.

bendk (approved):
Changes look good.  I made one suggestion, but it's not that important.  Feel free to skip it.

Issue #1076, opened 2021-10-25
Support exception hierarchies


bendk:
In the logins component, we currently have code like [this](https://github.com/mozilla/application-services/blob/1dab6d76e9163a5bb8d6b17859cdf4f88458ea96/components/logins/src/error.rs#L16-L18) and [this](https://github.com/mozilla/application-services/blob/1dab6d76e9163a5bb8d6b17859cdf4f88458ea96/components/logins/src/error.rs#L83-L98).  We create one error Enum, then wrap that as a variant of a top-level error Enum.

I think the most idomatic way to translate this into most languages would be an exception hierarchy like this:

* `LoginsError`
  * `InvalidLogin`
    * `EmptyOrigin`
    * `EmptyPassword`
    * ...

It would be great if UniFFI could support this.  I think the main thing to figure out is how to specify it in the UDL.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-101)
┆Issue Number: UNIFFI-101

PR #1126, opened 2021-11-30
Allow errors converting custom types to propagate

mhammond:
This gets close, but strikes a mismatch between the `Err` types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,13 @@
 
 ## [[UnreleasedVersion]] - (_[[ReleaseDate]]_)
 
-[All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.9.0...HEAD).
+[All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.9.0...HEAD)
+
+### ⚠️ Breaking Changes ⚠️
+
+- Error handling when converting custom types has been updated. If your `wrap()`
+  function returns an `Err`, in some cases it now [may not panic but instead
+  return the error declared by the function](https://mozilla.github.io/uniffi-rs/udl/ext_types_wrapped.html#error-handling-during-conversion).
 
 ## v0.15.2 - (_2021-11-25_)
 
--- a/docs/manual/src/udl/ext_types_wrapped.md
+++ b/docs/manual/src/udl/ext_types_wrapped.md
@@ -57,6 +57,74 @@ Because `UniffiCustomTypeWrapper` is defined in each crate, this means you can w
 if they are not defined in your crate - see the 'wrapper_types' example which demonstrates
 how `serde_json::Value` can be wrapped.
 
+## Error handling during conversion
+
+You might have noticed that the `wrap` function returns a `uniffi::Result<Self>` (which is an
+alias for `anyhow::Result`) and might be wondering what happens if you return an `Err`.
+
+It depends on the context. In short:
+
+* If the value is being used as an argument to a function/constructor that does not return
+  a `Result` (ie, does not have the `throws` attribute in the .udl), then the uniffi generated
+  scaffolding code will `panic!()`
+
+* If the value is being used as an argument to a function/constructor that *does* return a
+  `Result` (ie, does have a `throws` attribute in the .udl), then the uniffi generated
+  scaffolding code will use `anyhow::Error::downcast()` to try and convert the failure into
+  that declared error type and:
+  * If that conversion succeeds, it will be used as the `Err` for the function.
+  * If that conversion fails, it will `panic()`
+
+### Example
+For example, consider the following UDL:
+```idl
+[Wrapped]
+typedef i64 Handle;
+
+[Error]
+enum ExampleErrors {
+    "InvalidHandle"
+};
+
+namespace errors_example {
+    take_handle_1(Handle handle);
+
+    [Throws=ExampleErrors]
+    take_handle_2(Handle handle);
+}
+```
+
+and the following Rust:
+```rust
+#[derive(Debug, thiserror::Error)]
+pub enum ExampleErrors {
+    #[error("The handle is invalid")]
+    InvalidHandle,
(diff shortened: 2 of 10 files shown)


bendk (commented):
The approach seems reasonable.  Updating that FFI calling code is going to be annoying, but it seems doable.

My one concern is that the error type of the wrapper needs to match the error type of the method.  In practice, I think this is going to mean that libraries need to define a global error type.  That's not the worst, especially if we implemented something like #1076, but it does seem like it constrains how the libraries need to be written.

I'd prefer if we could make it so the error types for the wrapper and methods weren't linked together.  It seems like this should be possible with some modification to the `rustcall` module.

mhammond:
This is all green apart from the "uitests", and I'm very confused by them :) See cbd152c35fbf9d1f01be071853bfa37fd7d5ce8b - *some* changes are expected, but the change from `$DIR/counter.uniffi.rs:160:13` to `$WORKSPACE/target/debug/build/uniffi_uitests-6472345683e61af2/out/counter.uniffi.rs` is obviously a problem, but that's what was generated when I took the advice to use `TRYBUILD=overwrite` - anyone have any clue about this?

fixtures/ext-types/guid/src/guid.udl
```diff
@@ -1,13 +1,26 @@
...
     Guid? maybe_guid;
 };
 
 namespace ext_types_guid {
+    // Note this intentionally does not throw an error - uniffi will panic if
+    // a Guid can't be converted.
     Guid get_guid(optional Guid? val);
+
+    // Uniffi will handle failure converting a string to a Guid correctly if
+    // the conversion returns `Err<GuidError>`, or panic otherwise.
```
mhammond, line 21: wrong bracket! Should be `Err(GuidError)`

CHANGELOG.md
```diff
@@ -6,7 +6,13 @@
...
 ## [[UnreleasedVersion]] - (_[[ReleaseDate]]_)
 
-[All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.9.0...HEAD).
+[All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.9.0...HEAD)
+
+### ⚠️ Breaking Changes ⚠️
+
+- Error handling when converting custom types has been updated. If your `wrap()`
+  function returns an `Err`, in some cases it now [may not panic but instead
+  return the error declared by the function](https://mozilla.github.io/uniffi-rs/udl/ext_types_wrapped.html#error-handling-during-conversion).
```
bendk, line 15: Nit: there's a extra `[` before "may"

fixtures/ext-types/guid/src/lib.rs
```diff
@@ -1,13 +1,51 @@
...
+
 fn get_guid(guid: Option<Guid>) -> Guid {
+    // This function doesn't return a Result, so all conversion errors are panics
     match guid {
-        Some(guid) => guid,
+        Some(guid) => {
+            assert!(
+                !guid.0.is_empty(),
+                "our UniffiCustomTypeWrapper already checked!"
+            );
```
bendk, line 26: IIUC if this assertion fails, then that would be a test error.  I think this string would make it hard to figure out what went wrong in that case though.  How about changing the it to "our UniffiCustomTypeWrapper failed to check for an empty GUID" or something like that.

docs/manual/src/udl/ext_types_wrapped.md
```diff
@@ -57,6 +57,74 @@ Because `UniffiCustomTypeWrapper` is defined in each crate, this means you can w
...
+* If the value is being used as an argument to a function/constructor that does not return
+  a `Result` (ie, does not have the `throws` attribute in the .udl), then the uniffi generated
+  scaffolding code will `panic!()`
+
+* If the value is being used as an argument to a function/constructor that *does* return a
+  `Result` (ie, does have a `throws` attribute in the .udl), then the uniffi generated
+  scaffolding code will use `anyhow::Error::downcast()` to try and convert the failure into
+  that declared error type and:
+  * If that conversion succeeds, it will be used as the `Err` for the function.
+  * If that conversion fails, it will `panic()`
```
bendk, line 76: I like how this lets us keep using `anyhow::Result` as the return type in all cases.

uniffi_bindgen/src/scaffolding/templates/macros.rs
```diff
@@ -8,8 +8,29 @@
...
+            Err(err) => {
+                match err.downcast::<{{ e }}>() {
+                    Ok(actual_error) => return Err({{ func.throws_type().unwrap()|ffi_converter }}::lower(actual_error)),
+                    Err(ohno) => panic!("Failed to convert arg '{{ arg.name() }}': {}", ohno),
+                }
+            }
+        {% else %}
+            Err(err) => panic!("Failed to convert arg '{{ arg.name() }}': {}", err),
+        {% endmatch %}
+        }
```
bendk, line 33: This code looks good to me.  Does it get inserted for each function argument though?  Seems like that could make the scaffolding really hard to read.  I wonder if it could be refactored into a function in `uniffi` somehow, but I don't think that should block merging this.

bendk (approved):
The more I think about the options, the more I think the least bad one is to require that the wrapper and function error types match.  Maybe we could relax that in the future, but maybe we won't need to either.

Given that approach, I like this implementation a lot.  Keep everything using `anyhow` and use `downcast_ref` to check the error type.

I think the only issue is the UI tests, no clue what's happening with those though.  I'll make approve, but I do think we should get those green before merging.

tarikeshaq (approved):
LGTM! Glad `anyhow` is making our lives easier 🕺 

I agree with Ben that we might want a long-term discussion/fix to mark the error story better, but this here seems like the best way forward for now!

(I haven't looked at the uitests before so also not sure I can give any guidance there :( )

mhammond:
Thanks for the comments. I've still not tried to tackle records yet, but I think this is OK to land and unstick our "places" work.

PR #1128, opened 2021-12-01
Remove 'fileprivate' modifier inside a fileprivate extension.

badboy:
This causes a whole bunch of Swift warnings during the build:

    'fileprivate' modifier is redundant for static method declared in a fileprivate extension


bendk (approved):
Yes please, I hate those warnings :)

PR #1130, opened 2021-12-01
[Swift] Omit argument labels in functions

badboy:
PR is a strawman. See the details first:

For the default function definition, e.g. `myfun(parameterName: Int)`, Swift requires passing in named parameters: `myfun(parameterName: 3)`.
This can be changed by either adding another label `myfun(newLabel parameterName: Int)` and be called like `myfun(newLabel: 3)`.
Or you can omit the label with `myfun(_ parameterName: Int)` and call it like `myfun(3)`.

Currently UniFFI always generates the first form.
Sometimes that makes the functions read better when called.
But for Glean's API it makes it slightly awkward because now these labels become required:

```swift
GleanMetrics.cat.myCounter.add(amount: 1)
```

It also means that the names we define in the UDL are part of the public API and so changing them will break users.

For Glean I now have 4 options:

1. Live with it and break users and force everyone to pass in values with labels
2. Write wrapper code around every metric type that allows to omit the label
3. Convince UniFFI folks that we don't need labels (basically what this PR starts)
4. Come up with a way to let me specify whether I want a label or not (same `_` syntax in UDL?)

What do the others think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### What's Changed
 
 - Python: Added Callback Interface support
+- Swift bindings can now omit argument labels in generated functions using `omit_argument_labels = true` in the configuration.
 
 ## v0.15.2 - (_2021-11-25_)
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,4 +31,5 @@ members = [
   "fixtures/regressions/cdylib-crate-type-dependency/cdylib-dependency",
   "fixtures/uitests",
   "fixtures/uniffi-fixture-time",
+  "fixtures/swift-omit-labels",
 ]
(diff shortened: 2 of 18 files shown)


jhugman:
I hate to say it, but this needs to be a configuration option, defaulted to off :( not least because this is a big ole breaking change.

badboy:
I changed the PR to use a new option in the config file now

jhugman:
> Convince UniFFI folks that we don't need labels (basically what this PR starts)

I don't know if it's the uniffi folks, or the Swift folks. :D

One of the explicit goals is to generate bindings which are good citizens within each language community.

I'll r+ this now, but it would be a good idea to drop something in the changelog and the big book of config options (I say that as if there is such a document).

jhugman (approved):
This looks fine. Thank you. Could you add a changelog entry and something to document the option.

badboy:
> I'll r+ this now, but it would be a good idea to drop something in the changelog and the big book of config options (I say that as if there is such a document).

I just saw that there's _no_ docs about uniffi.toml at all. Let me get something started for this.

PR #1131, opened 2021-12-02
Update to uitest 1.0.53

mhammond:
Got very confused by uitest, so thought I'd offer clues to the next poor victim.

### DIFF
--- a/fixtures/uitests/Cargo.toml
+++ b/fixtures/uitests/Cargo.toml
@@ -16,4 +16,4 @@ uniffi_macros = {path = "../../uniffi_macros", features=["builtin-bindgen"]}
 uniffi_build = {path = "../../uniffi_build", features=["builtin-bindgen"]}
 
 [dev-dependencies]
-trybuild = "1"
+trybuild = "1.0.53"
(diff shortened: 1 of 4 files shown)


badboy:
So how does one do figure out the right thing?

mhammond:
> So how does one do figure out the right thing?

Sadly, as I mentioned in that ticket, you instrument the tool - but hopefully that ticket will offer a better solution and we can update the docs - or if it doesn't, I can go into more detail about how to instrument it.

I'm happy to wait until there's something more significant to say, but I felt that if someone struck this problem tomorrow, having a link to that ticket is better than not having it :)

mhammond:
That ticket was closed and a fix checked in, so theoretically the next release will deal with this better - I'll update this PR when that happens.

badboy:
> > So how does one do figure out the right thing?
> 
> Sadly, as I mentioned in that ticket, you instrument the tool - but hopefully that ticket will offer a better solution and we can update the docs - or if it doesn't, I can go into more detail about how to instrument it.
> 
> I'm happy to wait until there's something more significant to say, but I felt that if someone struck this problem tomorrow, having a link to that ticket is better than not having it :)

Ah, I misread that ticket. I thought you did find a hacky way to fix it here.
Guess [there was a release](https://github.com/dtolnay/trybuild/commit/feec62dde6114f9fbf9be6cede92596aa6c43c9b) and so we can fix it once and for all here too.

mhammond:
Updated to the new release and updated the test output (which wasn't strictly necessary as that tool does support "old" formats, but doing it now should make it less surprising to the next person who changes it)

badboy (approved):

PR #1132, opened 2021-12-02
Move generated arg lowering/error code into a helper function

mhammond:
As @bendk suggested in #1126, we can move lots of the generated boilerplate into the uniffi crate. This almost works - it's types all the way down and my brain is melting :)

I think the failure to handle `Arc` is tractable, but I guess my biggest concern are the `uitest` changes - it makes user-facing error messages intractable. I'll keep playing, but thoughts welcome about whether this is actually an improvement or not? The generated+formatted code without this patch isn't *that* bad - one `match` per arg still leaves it readable, just verbose. I'm on the fence.

### DIFF
--- a/uniffi/src/lib.rs
+++ b/uniffi/src/lib.rs
@@ -625,6 +625,20 @@ unsafe impl<T: Sync + Send> FfiConverter for std::sync::Arc<T> {
     }
 }
 
+pub fn lower_anyhow_error_or_panic<ErrConverter>(
+    err: anyhow::Error,
+    arg_name: &str,
+) -> ErrConverter::FfiType
+where
+    ErrConverter: FfiConverter,
+    ErrConverter::RustType: 'static + Sync + Send + std::fmt::Debug + std::fmt::Display,
+{
+    match err.downcast::<ErrConverter::RustType>() {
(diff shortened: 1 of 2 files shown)


bendk:
I think you can fix the `Arc` issue by wrapping the call in  `std::ops::Deref::deref()`, but at that point I'm not sure the generated code is any simpler.

What if we tried only removing the inner `match` for the `Err` branch?  https://github.com/bendk/uniffi-rs/commit/90465348b180a0c6c9260715c6ee8b3ee68ee42c

mhammond:
I think that is a small improvement and it exists, so 👍. It doesn't touch the uitest output which is an added bonus - and I just stole you commit directly, so I'll r+ it.

mhammond:
welp, apparently I can't r+ it :)

rfk (approved):
> welp, apparently I can't r+ it :)

Let's see if I can help 😁 

Also, 👋🏻 hi! Hope you're doing well @mhammond!

mhammond:
> Also, 👋🏻 hi! Hope you're doing well @mhammond!

👋 hi - I am doing well, but still trying to fill the hole you left 😅 . I hope you are great too and it's always a joy to see you pop up here, so I hope that continues :)

Issue #1102, opened 2021-11-11
Prepend local variables and parameters with `_` or `__`


jhugman:
Currently uniffi generates code like: 

```kotlin
rustCall() { status -> _UniFFILib.INSTANCE.glean_83a2_process(uuid.lower(), foo.lower(), status) }
```

Here, uniffi needs the variable `status`, and the user  is using `foo` and `uuid`.

If the user wants to use `status` then the bindings and/or scaffoldings will not compile.

To allow this, we should use characters in internal variables/parameters that aren't parsed by WebIDL and/or won't survive the changing by `heck::MixedCase`.

```kotlin
rustCall() { _status -> _UniFFILib.INSTANCE.glean_83a2_process(uuid.lower(), status.lower(), _status) }
```

Potential followups: review our uses of fully qualified class names in Kotlin, additional prefixes in Swift class names.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-112)
┆Issue Number: UNIFFI-112

name1e5s:
I'll try this.

PR #1133, opened 2021-12-05
add `_` for local variables for kotlin template when necessary

name1e5s:
#1102 

...or we'd better pretend **ALL** local variables and paramaters with `_`?


badboy:
Could you add a regression test for this?
Probably easiest to just add a new function to `fixtures/coverall/src/coverall.udl` a la

```
string get_status(string status);
```

Then call that in the Kotlin test.

badboy:
I went ahead and added the test. Waiting for CI, then we can merge it.

PR #1134, opened 2021-12-05
Update FFI Ruby gem install command

name1e5s:
`gem install FFI` doesn't work, the gem should install via `gem install ffi` according to [here](https://github.com/ffi/ffi).

```
➜  uniffi-rs gem install FFI
Ignoring iStats-1.6.1 because its extensions are not built. Try: gem pristine iStats --version 1.6.1
ERROR:  Could not find a valid gem 'FFI' (>= 0) in any repository
```


mhammond (approved):
Thanks!

PR #1143, opened 2021-12-22
[discuss] Decorator methods

bendk:
This is a rebase of the branch I shared in #1051.  @mhammond and I have been talking about how decorators may be a big part in getting more components on desktop, so I thought it would be good to have something for him to look at.  I think @jhugman is working on a second try at these too, so let's not merge until that's out and we can compare and contrast.

TODO:
  - Remove the hack for adding extra scripts.  I have the code for this in another branch, just need to merge it in.
  - Add back the Kotlin interface so that that you can use code completion when defining the decorator.  I think this means the type names need to be specified in the Kotlin config, but that's not too bad.

### DIFF
--- a/docs/manual/src/udl/decorators.md
+++ b/docs/manual/src/udl/decorators.md
@@ -1,0 +1,242 @@
+# Decorator objects
+
+We'll be considering the UDL for an object called `DemoRustObject`:
+
+```webidl
+interface DemoRustObject {
+    void do_expensive_thing();
+    [Throws=RustyError]
+    void do_crashy_thing();
+    void do_consequential_thing();
+}
+```
+
+`DemoRustObject` is a `struct` written in Rust, and uniffi generates a class of that name in the foreign languages which forward method calls to the Rust implementation.
+
+## Problem
+
+Just by the names of these methods, we can see that the application might not want to deal with the `DemoRustObject` by itself, without some proper care taken while calling its methods:
+
+* it may want to run `do_expensive_thing()` off the main thread
+* it may want to catch and report errors from `do_crashy_thing()`.
+* it may want to inform the rest of the application each time `do_consequential_thing()` is called.
+
+A common pattern when using uniffied code is to run some common, but app-specific, code before or after calling the Rust code.
+
+```kotlin
+class DemoLib(
+    val backgroundScope: CoroutineContext,
+    val errorReporter: (e: Exception) -> Unit,
+    val listener: () -> Void
+) {
+    val demoRustObject = new DemoRustObject()
+
+    fun doExpensiveThing() {
+        backgroundScope.launch {
+            demoRustObject.doExpensiveThing()
+        }
+    }
+
+    fun doCrashyThing() {
+        try {
+            demoRustObject.doCrashyThing()
+        } catch (e: Exception) {
+            errorReporter(e)
+        }
+    }
+
+    fun doConsequentialThing() {
+        demoRustObject.doConsequentialThing()
+        listener()
+    }
+}
+```
+
+This causes a proliferation of boiler plate code. Worse, everytime we add a new method to the `DemoRustObject`, handwritten foreign language code needs to be written to expose it safely.
+
+The more methods that want to do similar things, the more repetitive this gets.
+
+We could isolate the repeated code into a decorator class with `onBackgroundThread`, `withErrorReporter` and `thenNotifyListeners` methods.
+
+```kotlin
+class MyDemoDecorator(
+    val backgroundScope: CoroutineContext,
+    val errorReporter: (e: Exception) -> Unit,
+    val listener: () -> Void
+) {
+    fun onBackgroundThread(rustCall: () -> Unit) {
+        backgroundScope.launch {
+            rustCall()
+        }
+    }
+
+    fun <T> withErrorReporter(rustCall: () -> T) =
+        try {
+            rustCall()
+        } catch (e: Exception) {
+            errorReporter(e)
+        }
+
+    fun thenNotifyListeners(rustCall: () -> T) {
+        try {
+            rustCall()
+        } finally {
+            listener()
+        }
+    }
+}
+```
+
+Then, we can re-write the `DemoLib` as:
+
+```kotlin
+class DemoLib(
+    val demoRustObject: DemoRustObject,
+    val decorator: MyDemoDecorator,
+) {
+    fun doExpensiveThing() = decorator.onBackgroundThread {
+        demoRustObject.doExpensiveThing()
+    }
+
+    fun doCrashyThing() = decorator.withErrorReporter {
+        demoRustObject.doCrashyThing()
+    }
+
+    fun doConsequentialThing() = decorator.thenNotifyListeners() {
+        demoRustObject.doConsequentialThing()
+    }
+}
+```
+
+This is much better, but it's still looking a bit cut and pasty. Enter decorator objects.
+
+## Decorators to the rescue
+
+> Decorator objects take their name from [Python's decorator methods][py-decorators].
+>
+> In Pythonic terms, Uniffi's decorator objects are a collection of decorator functions.
+>
+> Swift and Kotlin don't provide functionality to capture arbitrary `*args` and call a function with those same `*args`, so
+> at this time, decorator objects aren't as powerful as Python decorators. Nevertheless, they can be still quite useful.
+
+[py-decorators]: https://www.python.org/dev/peps/pep-0318/#on-the-name-decorator
+
+### UDL
+
+In the UDL we:
+
+* Specify that `DemoLib` should be a decorated version of `DemoRustObject` using the `[Decorated=`] annotation
+* Specify how methods should by decorated using the `[CallsWith=]` annotation
+* Don't specify the class that holds the decorator functions.  This is done on a per-app basis in `uniffi.toml`.
+
+```webidl
+namespace demo {}
+
+[Decorated=DemoLib]
+interface DemoRustObject {
+    [CallsWith=onBackgroundThread]
+    void do_expensive_thing();
+
+    [Throws=RustyError, CallsWith=withErrorReporter]
+    void do_crashy_thing();
+
+    [CallsWith=thenNotifyListeners]
+    void do_consequential_thing();
+}
+```
+
+### Write your decorator class
+
+```kotlin
+package my.uniffi.bindings.package
+
+class MyDemoDecorator(
+    val backgroundScope: CoroutineContext,
+    val errorReporter: (e: Exception) -> Unit,
+    val listener: () -> Void
+) {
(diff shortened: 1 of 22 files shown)


docs/manual/src/udl/decorators.md
```diff
@@ -0,0 +1,242 @@
...
+
+* Specify that `DemoLib` should be a decorated version of `DemoRustObject` using the `[Decorated=`] annotation
+* Specify how methods should by decorated using the `[CallsWith=]` annotation
+* Don't specify the class that holds the decorator functions.  This is done on a per-app basis in `uniffi.toml`.
+
+```webidl
+namespace demo {}
+
+[Decorated=DemoLib]
+interface DemoRustObject {
```
rfk, line 136: The changing of the name here from `DemoRustObject` to `DemoLib` was a bit surprising to me from a whole-of-API design perspective. In practice it seems likely that the underlying Rust object will already have a good name, and decorated version on the other side of the FFI is *conceptually* the same thing so it might want to share the same good name. I wonder what it would look like if you kept the same name for the thing the user is expected to use (which in this case would be something like `DemoObject` and gave the raw Rust object wrapper a generated name like `DemoObjectRawUniFFI`.

From another perspective, the way this is structured in the UDL file makes `DemoRustObject` the most prominent name, but that's not the name we expect callers to use. Would someone scanning the `.udl` file understand that the thing they're supposed to use is `DemoLib`?

docs/manual/src/udl/decorators.md
```diff
@@ -0,0 +1,242 @@
...
+    }
+}
+```
+
+### Specify the decorator class in your uniffi.toml
+
+```toml
+
+[bindings.kotlin.decorators.DemoRustObject]
+class_name = MyDemoDecorator
```
rfk, line 186: I'd be tempted to call this `decorator_class_name` or similar for clarity, since you have three separate class names in play here (the undecorated class, the decorator, and the resulting decorated class).

docs/manual/src/udl/decorators.md
```diff
@@ -0,0 +1,242 @@
...
+constructor will input a `DemoRustObject` and `MyDemoDecorator`.
+
+For example:
+
+```kotlin
+val demoLib = DemoLib(
+   DemoRustObject(),
+   MyDemoDecorator(backgroundScope, errorReporter, listener))
+
+// This will call MyDemoDecorator.stateChange { DemoRustObject.doConsequentialThing() }
```
rfk, line 202:
```suggestion
// This will call MyDemoDecorator.thenNotifyListeners { DemoRustObject.doConsequentialThing() }
```

docs/manual/src/udl/decorators.md
```diff
@@ -0,0 +1,242 @@
...
+
+[Decorated=DemoLib]
+interface DemoRustObject {
+    [CallsWith=onBackgroundThread]
+    void do_expensive_thing();
+
+    [Throws=RustyError, CallsWith=withErrorReporter]
+    void do_crashy_thing();
+
+    [CallsWith=thenNotifyListeners]
```
rfk, line 143: It's kind of a shame that the required structure of the decorator class can't be declared in the UDL here for visibility (and for checking against typos etc) but I guess that's a function of UDL not being expressive enough for the type parameters and what-not.

docs/manual/src/udl/decorators.md
```diff
@@ -0,0 +1,242 @@
...
+
+```webidl
+namespace demo {}
+
+[Decorated=DemoLib]
+interface DemoRustObject {
+    [CallsWith=onBackgroundThread]
+    void do_expensive_thing();
+
+    [Throws=RustyError, CallsWith=withErrorReporter]
```
rfk, line 140: I'm slowly forgetting all the vagaries of WebIDL...would these work if you stacked them for readability? Like:

```
[Throws=RustyError]
[CallsWith=withErrorReporter]
void do_crashy_thing();
```

docs/manual/src/udl/decorators.md
rfk: (I see from further reading that using an auto-generated name isn't going to work, but I'm leaving the comment since it may be some helpful insight into how I built up a mental model of the proposal)

docs/manual/src/udl/decorators.md
```diff
@@ -0,0 +1,242 @@
...
+The above code will make UniFFI define a `DemoLib` class that uses
+`MyDemoDecorator` to decorate calls to `DemoRustObject`.  The `DemoLib`
+constructor will input a `DemoRustObject` and `MyDemoDecorator`.
+
+For example:
+
+```kotlin
+val demoLib = DemoLib(
+   DemoRustObject(),
+   MyDemoDecorator(backgroundScope, errorReporter, listener))
```
rfk, line 200: Ah I see, so actually the caller has to know about *all three* of the classes involved here and use their explicit names, so it wouldn't make sense to try to give the underlying Rust object an auto-generated name. It also means that you can't fully understand the generated API just by looking at the `.udl` file, because the name `MyDemoDecorator` only appears in the `uniffi.toml`.

This seems cognitively very heavy to me. I don't have a good suggestion for how to get out from under it though, assuming that the user needs to be able to customize the construction of a `MyDemoDecorator`.

docs/manual/src/udl/decorators.md
```diff
@@ -0,0 +1,242 @@
...
+        try {
+            rustCall()
+        } finally {
+            listener()
+        }
+    }
+}
+```
+
+Then, we can re-write the `DemoLib` as:
```
rfk, line 90: The `Lib` suffix here confuses me a bit FWIW, in my mental taxonomy a "lib" is a very different sort of thing to an "object". Would this all still work if you just called to `DemoObject` or similar, making it clearer that it's similar in shape and purpose to the underlying `DemoRustObject`?

rfk (commented):
Just so you know, I'm still subscribed to issues/PRs on this repo and while I don't dive in much, I do quite enjoy seeing the discussions go by :-)

I haven't looked at any of the code for this, but I was curious to read the updated docs and left some comments as I went. I hope they're helpful!

docs/manual/src/udl/decorators.md
```diff
@@ -0,0 +1,242 @@
...
+In the UDL we:
+
+* Specify that `DemoLib` should be a decorated version of `DemoRustObject` using the `[Decorated=`] annotation
+* Specify how methods should by decorated using the `[CallsWith=]` annotation
+* Don't specify the class that holds the decorator functions.  This is done on a per-app basis in `uniffi.toml`.
+
+```webidl
+namespace demo {}
+
+[Decorated=DemoLib]
```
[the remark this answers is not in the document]
bendk: This version only allows interfaces.  One of the goals was allow decorators to be specified at runtime.  This allows for example the `onBackgroundThread()` normally schedules things in the IO thread, but for the tests it just in the main thread.  If you take that as a requirement, then I couldn't think of a good way to decorate stand-alone functions.  I think it's worth thinking about this more though.

docs/manual/src/udl/decorators.md
bendk: That's half of it.  The other half is you want flexibility for the consumers.  One consumer might want their `onBackgroundThread` decorator to return a `Promise<T>` and another might want a `Deferred<T>`.  One consumer might want the logger to handle exceptions and change the return type to `Option<T>` another might not handle exceptions and keep the return type the same.  Maybe there's a 3rd consumer that doesn't want to use decorators at all.

There's definitely a trade-off here, since you can't just have a single UDL file that declares everything.  When we were discussing wrapper/custom types it seemed like people were valuing the consumer flexibility, so I figured it would also be good here.

docs/manual/src/udl/decorators.md
bendk: I agree that it seems like there should be a less complicated way, but I can't see a good way to do that without giving up on the goal of having the decorators be runtime customizable.  You're making me wonder if that tradeoff is worth it though.

If we decide to stick with this approach, maybe updating the example, starting with your suggestion about "lib" vs "object" could help.

rfk: > I can't see a good way to do that without giving up on the goal of having the decorators be runtime customizable.

Same :-/

I still like the "decorator" terminology better than other alternatives I've seen, but to be honest, my familiarity with Python decorators is probably what led me to assume some of these names would be hidden from the user. I don't have a concrete change to suggest, just noting it out loud.

docs/manual/src/udl/decorators.md
jhugman: > It's kind of a shame that the required structure of the decorator class can't be declared in the UDL here for visibility (and for checking against typos etc) but I guess that's a function of UDL not being expressive enough for the type parameters and what-not.

For completeness, #1051 is how that works out.

jhugman (commented):
Thanks for diving into this. I haven't looked too hard at the implementation, but it's interesting to see someone else take another run at this.

I think there are a lot of problems to be avoided by generating the foreign language decorator interface/protocol, and it's become confusing to read what the actual API is now.

Please keep going: this will cut out such a lot of boiler plate.

Oh also nice to see @rfk: 👋 hi Ryan!

docs/manual/src/udl/decorators.md
```diff
@@ -0,0 +1,242 @@
...
+* Specify how each method will be decorated in the decorator class using the `[CallsWith={decorator_name}]` attribute
+  * This specifies that this method should be called with function named `decorator_name`
+  * Methods that don't have the `CallsWith` attribute will be called directly
+
+### uniffi.toml
+
+* To enable a decorated class for a specific bindings language, add a table named `[bindings.{language}.decorators.{interface_name}]` with the following keys:
+  * `class_name`: name of the decorator implementation class
+  * `import` (optional): import line to add at the top of the file.  Use this to import your decorator class if needed.
+  * `return_types` (swift only): Table that maps decorator names to the return type for the decorator function.
```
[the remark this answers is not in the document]
bendk: First off, I agree that it would be nice to generate the interface on Kotlin, which means that the return type would need to be specified there as well.  I also think that would help with error messages, since you should get a message saying that you didn't implement the interface correctly.

On throwing: I think you could have a value in `uniffi.toml` that specifies the throwing behavior.  I would start with these options:
  - `rethrows` (the default).  The decorator may re-throw exceptions from the original function
  - `handles`.  The decorator handles exception from the original function and never throw.

On `await`/`async`:   I was thinking that rather than making the decorator `async` itself, it could return something that was compatible with `async` code.  For example, in Kotlin, the decorator can return a `Future<T>`, and a `suspend` function could call it's `.await()` method.  I'm not sure if that works in swift or not.

That said, I think this could be a config value as well.  If you specify `async=true` in `uniffi.toml`, then we could add `async` to the decorator method signature and the call site.

bendk:
I agree with the criticisms that this is getting overly complex.  #1148 is an attempt at a simplified approach.  WDYT?

PR #1144, opened 2021-12-23
Implemented the FfiConverter pattern on Kotlin

bendk:
This makes the lift/lower/read/write functions more uniform, which is
good for a couple reasons:

 - Reduces the code needed in each of the CodeType implementations
 - Opens the door for the last phase of custom types types.  It's
   simpler to add the custom type adapter if all types use the same
   basic ffi conversion system.
 - Support external types, since we can know how to
   lift/lower/read/write a value by the type name alone.  To help
   facilitate this, I made all user-types (record, enum, object, etc)
   use "Type{name}" as their canonical name and "FfiConverterType{name}"
   as their FFIConverter name.


mhammond (approved):
This looks both straightforward and desirable to me 👍

uniffi_bindgen/src/backend/types.rs
```diff
@@ -58,6 +58,18 @@ pub trait CodeType {
...
+    ///
+    /// This is the object that contains the lower, write, lift, and read methods for this type.
+    /// Depending on the binding this will either be a singleton or a class with static methods.
+    ///
+    /// This is the newer way of handling these methods and replaces the lower, write, lift, and
+    /// read CodeType methods.  Currently only used by Kotlin, but the plan is to move other
+    /// backends to using this.
+    fn ffi_converter_name(&self, oracle: &dyn CodeOracle) -> String {
+        oracle.class_name(&format!("FfiConverter{}", self.canonical_name(oracle)))
+    }
```
jhugman, line 71: Do any types _not_ cross the FFI? (objects didn't used to).

uniffi_bindgen/src/backend/types.rs
```diff
@@ -58,6 +58,18 @@ pub trait CodeType {
...
+
     /// An expression for lowering a value into something we can pass over the FFI.
     ///
     /// N.B. This should align with the `helper_code` generated by this `CodeType`.
```
jhugman, line 75: I wonder why you didn't delete L73-L105, or better/simpler: provide replace the `unimplemented!` macros with implementations.

uniffi_bindgen/src/bindings/kotlin/templates/BooleanHelper.kt
```diff
@@ -1,15 +1,17 @@
...
+    }
 
-internal fun Boolean.write(buf: RustBufferBuilder) {
-    buf.putByte(this.lower())
-}
\ No newline at end of file
+    fun write(v: Boolean, buf: RustBufferBuilder) {
+        buf.putByte(lower(v))
+    }
+}
```
jhugman, line 17: I love that this refactor is moving converter methods away from the object itself.

uniffi_bindgen/src/bindings/kotlin/templates/CustomTypeTemplate.kt
```diff
@@ -1,43 +1,29 @@
...
-        return {{ "value"|lift_var(builtin) }}
-    }
-
-    fun lower(value: {{ builtin|type_name }}): {{ self.builtin_ffi_type()|ffi_type_name }} {
-        return {{ "value"|lower_var(builtin) }}
-    }
+{%- match config %}
+{%- when None %}
+{#- No custom type config, just forward all methods to our builtin type #}
+internal typealias {{ outer|ffi_converter_name }} = {{ builtin|ffi_converter_name }}
```
jhugman, line 4: Nice.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/callback_interface.rs
```diff
@@ -17,55 +15,21 @@ impl CallbackInterfaceCodeType {
...
 
 impl CodeType for CallbackInterfaceCodeType {
     fn type_label(&self, oracle: &dyn CodeOracle) -> String {
         oracle.class_name(&self.id)
     }
 
-    fn canonical_name(&self, oracle: &dyn CodeOracle) -> String {
-        format!("CallbackInterface{}", self.type_label(oracle))
+    fn canonical_name(&self, _oracle: &dyn CodeOracle) -> String {
+        format!("Type{}", self.id)
```
jhugman, line 26: You've done this change in multiple places. Should the default implementation go live in the same place as the default implementation of `ffi_converter_name`?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/miscellany.rs
```diff
@@ -5,54 +5,37 @@
...
-             pub struct $T;
-
-             impl CodeType for $T  {
-                 fn type_label(&self, _oracle: &dyn CodeOracle) -> String {
-                     $class_name.into()
-                 }
-
-                 fn canonical_name(&self, _oracle: &dyn CodeOracle) -> String {
-                    $canonical_name.into()
+    ($T:ty, $class_name:literal, $canonical_name:literal, $template_file:literal) => {
```
jhugman, line 13: Do these macros disappear down in to one implementation now? (he asks, with some excitement)

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -346,33 +346,24 @@ pub mod filters {
...
-        nm: &dyn fmt::Display,
-        codetype: &impl CodeType,
-    ) -> Result<String, askama::Error> {
-        Ok(codetype.read(oracle(), nm))
+    pub fn lift_fn(codetype: &impl CodeType) -> Result<String, askama::Error> {
+        Ok(format!("{}.lift", codetype.ffi_converter_name(oracle())))
+    }
+
+    pub fn read_fn(codetype: &impl CodeType) -> Result<String, askama::Error> {
+        Ok(format!("{}.read", codetype.ffi_converter_name(oracle())))
```
jhugman, line 366: I don't especially like this change from `*_fn` to `*_var`: I'd prefer it if these calls always produced a valid expression within the language it's generating.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/custom.rs
```diff
@@ -100,6 +64,10 @@ pub struct KotlinCustomType {
 impl KotlinCustomType {
     pub fn new(name: String, builtin: Type, config: Option<CustomTypeConfig>) -> Self {
         Self {
+            outer: Type::Custom {
+                name: name.clone(),
+                builtin: builtin.clone().into(),
+            },
```
jhugman, line 70: Curious (to have it in this PR).

Could this be passed in from the oracle?

jhugman (commented):
Overall, I like this PR.

My only requested change is to keep the filter methods as generating valid expressions, rather than parts of expressions.

Good job, thanks.

uniffi_bindgen/src/backend/types.rs
jhugman: Coming back to this, I think default implementations should generate the expressions in terms of the `FfiConverter` methods.

If we need to customize/error out for a particular type then we still can at binding generation time, rather than wait for binding compile time or runtime.

jhugman (changes requested):
I've changed this to Request changes, so as not to land accidentally.

uniffi_bindgen/src/backend/types.rs
bendk: I think `Error` only goes one way, it can't be passed back to rust.  Although that might be fixable that now.  In general, this moves the logic into the the FFIConverter classes.  So if a type can't cross the FFI for some reason, then I think we should have the `FffiConverter` with unimplemented methods.

uniffi_bindgen/src/backend/types.rs
bendk: Agreed.  The only reason I didn't change this was that I didn't want to mess with the bindings that don't use FFIConverter yet.  Once they all get switched over, I think we can do something like this.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/callback_interface.rs
bendk: I personally think so, but wouldn't that make it mega-match style?  Let's discuss that one more in #1145.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/miscellany.rs
bendk: I think that might be possible as well.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/custom.rs
bendk: It could be but it's kind of awkward because of how `iter_custom_types()` is defined.  I really feel like there's one last refactor to do that would clean up a bunch of code, but that can be done after we merge this.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
bendk: I don't feel that strongly about this, but one nice thing about the `*_fn` version is that it avoids creating temporary variables in [some cases](https://github.com/mozilla/uniffi-rs/pull/1144/files#diff-f564b605a6ba32d19bcb47128dc9c5c2f01169e2a41fa60ffaaf0b6eea13845c).  I think that was the only time it happened for this refactored, because of the way the Kotlin code uses scope functions.  But there were several instances of that in the scaffolding code that we cleaned up.  I think there will be more instances in the other languages.

That said, I don't think it would be hard to change my mind on this one.  What's the benefits for having the template code generate the full expression?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/callback_interface.rs
jhugman: No, this is still on `CodeType`, but they're all the same implementation. For the types that need to vary, it's trivial for them to vary.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
jhugman: I'm not sure I care about a temporary variable.

Working with expressions provides a more high-level tool to work with the source code than strings, and gives us the flexibility to change the expression for each type.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/miscellany.rs
jhugman: Do it!

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
bendk: Yes, but I'm not sure that kind of flexibility is good.  I want the FFI functions to be standardized as much as possible for reasons listed at the top.  Can you envision any reason where this system wouldn't work and we would have to use a different kind of expression?

bendk: In other words, I want whatever complexity is needed to exist inside a Kotlin function rather than a filter function.

jhugman (approved):
Rubberstamp LGTM

PR #1145, opened 2021-12-23
Template dispatch / codegen

bendk:
After I implemented the `FfiConverter` pattern on Kotlin, it struck me that there's not all that much logic left in the gen_kotlin submodules.  This is an experiment to to simplify things and remove boilerplate by moving that functionality into other places, then removing those modules.

The main functionality those modules were providing was code-generation: `CodeDeclaration` and the `imports()` and `helper_code()` methods of `CodeType`.  That's all now handled by:
   - Having a single main template that defines the declaration/helpers for that type.
   - With the help of the [components.rs](https://github.com/mozilla/uniffi-rs/blob/501a87093239ba19cad214c08317a44c02133229/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/components.rs) module, those templates can do things like add import statements, add initialization code, and include a template once ([CallbackInterfaceTemplate.kt](https://github.com/mozilla/uniffi-rs/blob/501a87093239ba19cad214c08317a44c02133229/uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceTemplate.kt) is a good example)
   - The [`Type.kt` template](https://github.com/mozilla/uniffi-rs/blob/501a87093239ba19cad214c08317a44c02133229/uniffi_bindgen/src/bindings/kotlin/templates/Type.kt) handles dispatching to the specific template for a type variant, similar to `Oracle.find()`.
nent.rs` so that each of those templates can do things like add imports/initialization code or include_once another template.

This gets us back to one of the early goals that we were discussing at the work week:  Almost all of the logic for a type is in one place.  I didn't expect that place to be an askama template, but I think I like the result.

The `type_name()` and `literal()` methods don't fit into that pattern, so they're back to using plain matches.  This seems fine to me, but I also think we could move that functionality into the template code if we wanted to.  The last bit of functionality from `CodeType` was the `canonical_name()` method, this PR changes Kotlin to just use the shared version of that again.

After everything was done, this removed 800 lines of code.   I think we should be able to save about the same amount for other bindings.  Then we could probably save a bit more once everything uses the new system.


jhugman (commented):
I was surprised how much I liked this.

I liked the `CodeBlock` idea, with the `add_import` and `add_declaration` methods.

I liked the simplification that comes from using template includes, and the `Types.kt` was an acquired taste, but was a nice tidy up of the boilerplate in the codetype modules.

However, it also showed the reason to have those codetype modules: which was to tidy up the thicket of routing implemented by the mega-matches in `canonical_name` (which should've/could've been deleted), `type_name` and `render_literal`.

bendk:
> However, it also showed the reason to have those codetype modules: which was to tidy up the thicket of routing implemented by the mega-matches in canonical_name (which should've/could've been deleted), type_name and render_literal.

This is the main question for me as well, but I don't know if we're seeing it the same way.  Maybe we should try to come to an agreement on the reasoning behind all of this.

IMO, the reason to eliminate the mega-matches is because most of the code for each type variant should live together.  The lift/lower code, the class definitions, the initialization code, the imports, etc. are are very related to each other so it's good to have it in one place.

Based on that:
 - `type_name` should probably be together with the other code, rather than in a mega-match.  I thought about a couple ways to do this.  The one I like best is have each type create a typealias to the canonical name at the top of its template.  Then the `type_name` filter can just return `type_.canonical_name()`, no mega-match needed.
 - `canonical_name` doesn't really need to live with the other code since the only requirement generating a unique string.  It doesn't care about the details of the class definitions, lifting/lowering, or anything else.  I prefer the mega-match for this one because it can be defined in the shared code.
 - `render_literal` kind of sits in the middle.  Most of the cases aren't really related to the other type code, but enum literals are.  Since there's not much code here, a mega-match doesn't feel that bad to me.  I'd also be fine with moving it closer to the other type code, for example maybe we make `EnumLiteral.kt`, `SequenceLiteral.kt`, `IntLiteral.kt`, etc.
 - Overall, this change seems like a big win for keeping relating things together.  For example, the functionality from [`gen_kotlin/object.rs`](https://github.com/mozilla/uniffi-rs/blob/f47d2eb5c41215f6bbba3b8df87648f00ed9a491/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs) is now at the top of [`templates/ObjectTemplate.kt`](https://github.com/mozilla/uniffi-rs/blob/501a87093239ba19cad214c08317a44c02133229/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt).  I think that's a great improvement that outweighs any issues with `render_literal` and the slightly ugly template code.

Maybe I'm missing some other reasons to avoid mega-matches though.  I'm very interested in how you see this all.

bendk:
I still like this code, but I don't think it's worth the time to keep working on it at this point.  Maybe we can revisit down the road.

PR #1146, opened 2021-12-31
Add a doc comparing UniFFI with diplomat

mhammond:
I spent some time having a look at diplomat, primarily to see what UniFFI can learn and opportunities to borrow, and felt it is worth sharing more broadly. I'm not sure if this is the best place for such a doc, and whether we should consider adding a link to README.md (I guess we probably should - explicitly noting diplomat exists in our README could well help some users)

I'd love feedback from everyone, but explicitly flagging just a few - in particular, @Manishearth (who I can't actually flag!) to represent the diplomat team, @rfk as the author of the much referenced #416, and @badboy to see if the document accurately describes his experience trying to use glean with the macro approach in #416. There's no hurry whatsoever, so please continue to enjoy your holidays 🎇 

[rendered version here](https://github.com/mhammond/uniffi-rs/blob/diplomat-comparison/docs/diplomat-and-macros.md)


docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,186 @@
...
+the current time.
+
+## Is this a problem for users of diplomat? Will diplomat solve it?
+
+I couldn't find real examples using diplomat, so it's difficult to know if this
+is a problem in practice. UniFFI came from a world where we had Rust crates and
+a hand-written FFI that exposed types from all over the crate. If these tools
+had started with the limitations from the macro approach in mind, it's possible
+a different, acceptable design might have been made to work. Maybe duplicating
+some structs and supplying suitable `Into` implementations might make things workable?
```
Manishearth, line 141: Yeah so we have a plan here: we will likely provide helpers that allow for duplicating structs and C-like enums with Into impls. Just haven't gotten around to it yet.

We do not want to actually allow for straight up using foreign non-opaque types, because it becomes a safety issue if these types ever change (e.g. someone removes the `#[repr(C)]`).

Come to think of it: how does uniffi handle this? Is there any way to expose types that are not defined by uniffi? It seems like y'all would have isomorphic problems.

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,186 @@
...
+
+As much as some of the UniFFI team dislike the external UDL file, there's no clear path to
+moving away from it. The macro approach is too limiting, and no other promising opportunities
+have presented themselves. There's no clear alternative to UDL which allows a complex
+type universe to be described, and at this stage, any replacement would need to be
+compelling enough to make a change worthwhile, which is hard to imagine.
+
+In the short term, the best we can probably do is to enumerate the perceived problems
+with the UDL file and try to make them more ergonomic - for example, avoiding repetition of
+`[Throws=SomeError]` would remove alot of noise, and some strategy for generating
```
Manishearth, line 185: diplomat just treats documentation as "yet another backend", which works reasonably well, since the architecture of a diplomat backend is just "here's the type structure, you know what to expect on the FFI layer, do what you want".

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,186 @@
...
+
+This document attempts to describe these different approaches and discuss the pros and cons of each.
+It's not going to try and declare one better than the other, but instead just note how they differ.
+If you are reading this hoping to find an answer to "what one should I use?", then that's easy -
+each tool currently supports a unique set of foreign language bindings, so the tool you should
+use is the one that supports the languages you care about!
+
+(There may even be a future where these 2 tools converge - that seems like a lot of work, but
+might also provide a large payoff - more on this later)
+
```
Manishearth, line 16: i think a difference not really covered in this document is the style of actual interface. UniFFI basically seems to be serializing types across FFI with this cool FFIConverter scheme, whereas Diplomat uses raw `repr(C)` so backends do not need much ceremony to read things. Both are valid approaches, but it's an interesting difference likely borne out of the choice of languages to prioritize (Diplomat cares heavily about C++)

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    impl MyFFIType {
+        pub fn create() -> MyFFIType { ... }
+        ...
+    }
+}
+```
+
+This works fine, but starts to come unstuck if you want the types defined somewhere else. In this trivial example, something like:
```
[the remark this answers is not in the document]
Manishearth: Note that the _macro_ does not know what's in other modules, however the _macro_ only does code generation for the C API and as such does not need to know about types in other modules. A typical diplomat run will `cargo build` to get the `.a` file, and then `diplomat-tool --cpp` (or whatever) to generate the headers (or js module, or whatever)

`diplomat-tool` is able to perform the holistic collection of all modules and can understand how types interact.

Manishearth (changes requested):
Hmm, this is a good idea, but I think there might be some misunderstandings about how diplomat handles modules.

docs/diplomat-and-macros.md
Manishearth: Ah, it seems like uniffi never directly exposes the rust values across ffi, it encodes them into its own format independent of `repr(C)` and `repr(Rust)`. clever.

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,186 @@
...
+for example, there's a `Type` enum where, for example, `Type::Record(Record)` represents a
+Rust struct. In other words, diplomat's type world can not be divorced from Rust,
+whereas UniFFI's already is.
+
+That said though, there might be a future where merging or otherwise creating some
+interoperability between these type universes might make sense. You could imagine
+a world where you can use diplomat to describe your type universe, but use UniFFI's foreign
+generation code to generate the Kotlin bindings. Similarly, a world where you use UniFFI
+and UDL files to describe your type universe, but then use diplomat to generate
+the NodeJS bindings.
```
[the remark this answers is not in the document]
rfk: Thinking a bit more seriously about this, I don't think that there's realistically much value in having a kind of two-way interoperability between UniFFI bits and Diplomat bits. It would require the projects to converge on a great many details of how the FFI layer works, and I don't think there are sufficiently many quick wins to fuel that kind of work. For example, it would almost certainly be quicker for Diplomat to gain a Kotlin backend by writing one from scratch than by trying to iterate the UniFFI Kotlin backend towards something that works for both.

However, what I could see happening in future is UniFFI becoming a kind of higher-level wrapper around Diplomat. I can imagine a Diplomat *backend* for UniFFI that converts a `.udl` file into a bridge module and then uses the Diplomat toolchain to generate bindings from it, keeping  some of the additional affordances/conveniences we've built for our specific use-cases (e.g. around megazording).

An interesting thought experiment, anyway 😁

rfk (commented):
@mhammond thanks for writing this up, and for looping me in! I'll try to chew on this a bit over the coming days as I have a little downtime, but some initial thoughts below.

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,186 @@
...
+    pub b: bool,
+}
+```
+
+Might be enough for the generation of the Rust scaffolding. However, the problems are in the foreign bindings.
+
+## How the type universe is constructed for the macro approach.
+
+In both diplomat and [#416](https://github.com/mozilla/uniffi-rs/pull/416), the approach taken
+is that the generation process wants a path to the Rust source file that contains the module in
```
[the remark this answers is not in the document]
rfk: > there may be multiple #[diplomat::bridge] tagged modules in the crate

@mhammond @badboy I wonder how much of a difference this would make to a use-case like Glean, if you could e.g. have one `#[diplomat::bridge]` declaration per metric type but the tool knows how to find them all and expose them as a single API surface when generating the bindings.

docs/diplomat-and-macros.md
Manishearth: > . If Diplomat lets you duplicate the struct into the bridge module, and decorate it with a helper macro in order to generate some automatic Into helpers, then that feels pretty close to the same amount of work as what you have to do with UniFFI:

Note that we don't allow this just yet, but ... "close enough"; it's a very easy feature to add and we mostly haven't added it because we haven't needed it that much yet.

docs/diplomat-and-macros.md
mhammond: As Ryan notes later, I think this is a reflection of expedience and initial use-cases, but not necessarily a key difference between the tools at a high level - I think UniFFI *could* support "native" `repr(C)` types in some cases.

If you still think this is important to communicate after seeing my updated version, please give me a rough idea of what it should say and where :)

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,186 @@
...
+all types known by each of the tools. Both of these tools build their own "type universe" then
+use that to generate both Rust code and foreign bindings.
+
+## UniFFI's type universe
+UniFFI's model is to parse an external ffi description from a `.udl` file which describes the
+entire "type universe". This type universe is then used to generate both the Rust scaffolding
+(on disk as a `.rs` file) and the foreign bindings.
+
+**What's good about this** is that the entire type system is known when generating both the rust code
+and the foreign binding.
```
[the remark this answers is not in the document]
mhammond: Actually, the point I had in mind was simply that the foreign bindings need to know the names and types of struct elements so it can recreate the same struct on the other side. But I've tried to capture that in the new version.

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,186 @@
...
+## Diplomat's type universe
+
+Diplomat defines its "type universe" (ie, the external ffi) using macros.
+
+**What's good about this** is that the "ffi module" defines the canonical API and it is defined in
+terms of Rust types - the redundant UDL is removed. The Rust scaffolding can also be generated
+by the macros, meaning there are no generated `.rs` files involved.
+
+Ryan even tried this for UniFFI in [#416](https://github.com/mozilla/uniffi-rs/pull/416) - but we
+struck **what's bad about this**: the context in which the macro runs doesn't know about types defined
```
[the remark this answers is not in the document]
mhammond: Thanks - I've reworked the doc to try and capture this better - lmk if the new version looks like a more accurate representation.

docs/diplomat-and-macros.md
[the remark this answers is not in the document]
mhammond: thanks! I've tried to capture this in the new version.

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,186 @@
...
+        pub fn create() -> MyFFIType { ... }
+        ...
+    }
+}
+```
+
+This works fine, but starts to come unstuck if you want the types defined somewhere else. In this trivial example, something like:
+
+```Rust
+pub struct MyFFIType {
```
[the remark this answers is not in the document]
mhammond: As above, I hope I've captured this

docs/diplomat-and-macros.md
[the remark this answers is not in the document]
mhammond: This is certainly worth experimenting with, especially if we can avoid the "module" limitation - eg, if all types needed `[uniffi::magic]` then maybe there's actually no reason to insist on a wrapping module? Regardless, I've tried to capture this idea in the doc.

docs/diplomat-and-macros.md
mhammond: I've tried to capture this too

docs/diplomat-and-macros.md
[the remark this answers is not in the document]
mhammond: yeah, I did get a bit carried away - I've tried to tone this down a little in the new version.

mhammond (commented):
Thanks to both of you for your considered comments. I've made some significant changes in a version I'm about to push, so I'd love continuing to get feedback on the new version. It would also be great if you could "resolve" any comments which you think are suitably represented in the new version, just so I can keep track of what I should consider in future iterations.

docs/diplomat-and-macros.md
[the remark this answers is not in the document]
badboy: I think in the 416 approach that would still have not worked, because we hash all input to put that into the generated function names as a way to prevent use of the wrong version of a library.
A single `#[diplomat::bridge]` invocation cannot know all input and thus can't generate one hash.
Of course this is a limitation we built ourselves, so maybe could be lifted.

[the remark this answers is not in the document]
badboy: Re-reading quickly how diplomat works that same hashing could be applied to only the limited bridged module while still giving the same benefits.
Given that `diplomat-tool` parses the same code, it would come up with the same hash and know what functions to call in the foreign language code.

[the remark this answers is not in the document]
Manishearth: Yeah the reason we have the module wrapper is mostly for convenience, because we need to look at imports as well sometimes.

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,264 @@
...
+#[uniffi::magic]
+pub struct MyFFIType {
+    pub a: i32,
+    pub b: bool,
+}
+```
+
+Might be enough for the generation of the Rust scaffolding. However, the problems are in the
+foreign bindings, because, eg, those foreign bindings do not know the names and types of the
+struct elements.
```
bendk, line 123: Can you elaborate on this one?  If we can use these macros as the source for scaffolding code, why couldn't we generate the bindings code as well?

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,264 @@
...
+    impl MyFFIType {
+        pub fn create() -> MyFFIType { ... }
+        ...
+    }
+}
+```
+
+So while we haven't exactly reduced the duplication, we have removed the UDL.
+We probably also haven't helped with documentation, because the natural location for
+the documentation of `MyFFIType` is probably at the *actual* implementation.
```
bendk, line 237: This section is some serious food for thought.

For one, as rfk points out, we're currently duplicating the struct/function definitions in the UDL.  The way I see it: both projects have made a similar design decision.  @Manishearth describes it as "we do not want changes far away to change the FFI API".  I would maybe reword that to "the FFI should be fully defined in one place".  For UniFFI, that place is the UDL.  For diplomat, it's the ffi module.  (Multiple ffi modules complicates this picture a bit, but doesn't fundamentally change things).

However, there is one difference: in some cases you only need code inside the ffi module.  This has the potential to reduce duplication.  For example, it would be very natural to move our `Store` definitions inside the ffi module.  Also, since types defined inside the ffi module are visible to the rest of the rust code, so we could move types like `Login`, `EncrytpedLogin`, `CreditCard`, `Address`, etc. into the ffi module.  I think external data types would still need duplicate definitions, but maybe that's it.

One potential issue with this is that it couples the library with the FFI code.  But this doesn't seem to be a problem for our components.  As @markh points out, our Rust APIs exists purely to service the FFIs.

It really makes me wonder about switching from a UDL-based approach to a macro based approach.  At the start, maybe each consumer would simply refactor their current UDL file to a macro.  This shouldn't be much work, maybe we could even automate it.  After that, we have the ability to do to small refactors that eliminate the unneeded duplication.

BTW, this also could solve some documentation issues since a) we can actually see docstrings when using syn and b) if there's only 1 place where a type is defined then it's clear where to put the docstrings.

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,264 @@
...
+While the process for defining the type universe is different, the actual in-memory
+representation of that type universe isn't radically different from UniFFI - for example,
+here's the [definition of a Rust struct](https://github.com/rust-diplomat/diplomat/blob/main/core/src/ast/structs.rs),
+and while it is built from a `syn` struct, the final representation is independent of `syn`
+and its ast representation.
+
+## UniFFI's experience with the macro approach.
+
+Ryan tried this same macro approach for UniFFI in [#416](https://github.com/mozilla/uniffi-rs/pull/416) -
+but we struck **what's bad about this** - at least for UniFFI's use-cases: the context in which the
```
Manishearth, line 68: nit: something's missing here?

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,264 @@
...
+struct elements.
+
+### Why is this considered a limition for UniFFI but not diplomat?
+
+As mentioned above, diplomat considers the limitation described above as an intentional design
+feature. By limiting where FFI types can be described, there's no risk of changes made "far away"
+from the FFI to change the FFI. This was born of experience in tools like `cbindgen`.
+
+For Uniffi, all use-cases needed by Mozilla don't share this design goal, primarily because the
+FFI is the primary consumer of the crate. The Rust API exists purely to service the FFI. It's not
```
Manishearth, line 132: ah, this explains a lot the design decisions :)

Manishearth (approved):
Seems good!

docs/diplomat-and-macros.md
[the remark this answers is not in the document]
mhammond: Yeah, thanks for clarifying - and while that's interesting context for this discussion, I don't think a discussion of the hashing needs to go into that document. LMK if you disagree though.

docs/diplomat-and-macros.md
mhammond: Thanks! But I can't see what's missing in this line - which probably means I should reword it :) Can you please let me know what is unclear?

docs/diplomat-and-macros.md
mhammond: I hope I cleared this up in a86e9496623f9bbee77fb7b3b23e960ee47ebdc2 - lmk if I need more about this.

docs/diplomat-and-macros.md
mhammond: I added a brief note in a86e9496623f9bbee77fb7b3b23e960ee47ebdc2 which doesn't capture all of this comment, but does briefly say why we should consider this more.

docs/diplomat-and-macros.md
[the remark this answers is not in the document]
badboy: Leaving that out sounds right to me.
It's a thing we will need to consider if we try the macro approach again, but not necessary to discuss in comparison/

badboy:
Thanks for writing this up!
This actually makes me want to get back to trying a macro-based approach again.
While I like our UDL (probably more than others here) it's duplication of concerns can be a bit annoying.

Regarding the request for my experience: [In my #416 comment](https://github.com/mozilla/uniffi-rs/pull/416#issuecomment-893327808) I mentioned "In the current implementation a lot of errors are hidden", unfortunately I can't remember further details.
What I _think_ I encountered was bugs in both the macro-annotated module and outside of it, where `rustc` would then only show the errors from the macro. This made iterating on the code base annoying as I couldn't see all errors at once.
But hard to say after the fact whether that's an implementation bug or some more underlying problems.

badboy (approved):

docs/diplomat-and-macros.md
rfk: IIUC you want this sentence to treat the `**what's bad about this**` as a kind of inline link to suggest "but we struck the bad things noted here". It might be clearer as an actual link :-)

docs/diplomat-and-macros.md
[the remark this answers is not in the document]
rfk: > Re-reading quickly how diplomat works that same hashing could be applied to only the limited bridged module
> while still giving the same benefits.

Good point, and in retrospect this thing about including a hash of the component in the name of the generated function probably deserved its own separate design doc. A summary of what's in my head based on the above, in case some of the details aren't obvious:

The main goal of naming our FFI functions like `component_<hash>_do_the_thing` instead of just `component_do_the_thing` is to guard against undefined behaviours if foreign-language bindings for version X of the interface are accidentally used with a `.so` for version Y of the crate. That is: if the details of how to call one of the FFI functions change, then we want the generated name of that function to change.

We don't necessarily have to achieve that by using a single `<hash>` of the whole API surface. If we split the FFI into several independent macro invocations, then as long as they're *self-contained* then they could all safely use their own individual hash.

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,272 @@
...
+Diplomat defines its "type universe" (ie, the external ffi) using macros.
+
+**What's good about this** is that an "ffi module" (and there may be many) defines the canonical API
+and it is defined in terms of Rust types - the redundant UDL is removed.
+The Rust scaffolding can also be generated by the macros, meaning there are no generated `.rs`
+files involved. Types can be shared among any of the ffi modules defined in the project -
+for example, [this diplomat ffi module](https://github.com/unicode-org/icu4x/blob/7d9f89fcd7df4567e17ddd8c46810b0db287436a/ffi/diplomat/src/pluralrules.rs#L50-L51)
+uses types from a [different ffi module](https://github.com/unicode-org/icu4x/blob/7d9f89fcd7df4567e17ddd8c46810b0db287436a/ffi/diplomat/src/locale.rs#L19).
+
+It also offers better control over the stability of the API, because where the FFI is defined
```
rfk, line 56: FWIW I don't really understand what's "better" about using a macro than a separate interface-definition file in this regard. IIUC, in UniFFI the API is stable as long as the `.udl` doesn't change, while in Diplomat the API is stable as long as the contents of hte macro block don't change.

(Making this claim doesn't bother me, I just don't get it and wonder if I'm missing something).

docs/diplomat-and-macros.md
rfk: (That is to say, I think this would read more clearly if you make the `**what's bad about this**` a kind of sub-heading in the same style as the UniFFI section above, and reference it more explicitly in the "UniFFI's experience" part).

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,272 @@
...
+
+As mentioned above, diplomat considers the limitation described above as an intentional design
+feature. By limiting where FFI types can be described, there's no risk of changes made "far away"
+from the FFI to change the FFI. This was born of experience in tools like `cbindgen`.
+
+For Uniffi, all use-cases needed by Mozilla don't share this design goal, primarily because the
+FFI is the primary consumer of the crate. The Rust API exists purely to service the FFI. It's not
+really possible to accidentally change the API, because every API change made will be in service
+of exposing that change over the FFI. The test suites written in the foreign languages are
+considered canonical.
```
rfk, line 140: Ah, I think this helps answer my question about about API stability in UniFFI vs Diplomat. I still feel like the `.udl` serves a similar purpose in the UniFFI world, but because of the way the `.udl` kind of maps directly onto the underlying Rust code, I can see an argument that changes in a UniFFI component are *more likely* to change the FFI surface than in Diplomat.

They won't change to FFI surface by stealth, because you'll have to update the `.udl` file...but changing the `.udl` file is likely to be the simplest way to accommodate a change in the underlying Rust code. In Diplomat you might instead write some adapter code in the FFI module in order to avoid changing the FFI surface, because there's an obvious place to do it.

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,272 @@
...
+is very similar - it takes a path to a the Rust source file/tree, and uses `syn` to locate the special modules (ie, ones annotated with `#[diplomat:bridge]` in the case of diplomat.)
+
+While some details differ, this is just a matter of implementation - #416 isn't quite as agressive
+about consuming the entire crate to find multiple FFI modules (and even then, diplomat doesn't
+actually *process* the entire crate, just modules tagged as a bridge), but could easily be
+extended to do so.
+
+But in both cases, for our problematic example above, this process never sees the layout of the
+`MyFFIType` struct, so that layout can't be communicated to the foreign bindings.
+As noted above, this is considered a feature for diplomat, but a limitation for UniFFI.
```
rfk, line 154: How does this work in Diplomat, does the FFI allow you to pass it around as a pointer but not construct one for yourself on the foreign-language side?

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,272 @@
...
+
+```Rust
+#[uniffi::magic]
+pub struct MyFFIType {
+    pub a: i32,
+    pub b: bool,
+}
+```
+
+maybe can made to work, so long as we are happy to help UniFFI discover where such annotations
```
rfk, line 190:
```suggestion
maybe can be made to work, so long as we are happy to help UniFFI discover where such annotations
```

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    impl MyFFIType {
+        pub fn create() -> MyFFIType { ... }
+        ...
+    }
+}
+```
+
+So while we haven't exactly reduced the duplication, we have removed the UDL.
```
rfk, line 240: FWIW, I think the toy example is the worst-case for highlighting the duplication here because the two duplicate declarations are separated by just 7 lines of text. In a real-world crate I would expect the duplication to not *feel* quite so bad because the source struct and its redeclaration would be further apart.

(That doesn't help with some of the other contra points raised here though)

docs/diplomat-and-macros.md
```diff
@@ -0,0 +1,272 @@
...
+We probably also haven't helped with documentation, because the natural location for
```
rfk, line 241: You could imagine explicitly telling the macro about the path to the redeclared struct, like:

```
    #[uniffi::declare_imported_type(super::MyFFIType)]
    pub struct MyFFIType {
        pub a: i32,
        pub b: bool,
    }
```

Then if it wanted to, the code processing this declaration could go find the corresponding `super::myFFIType` and pull out e.g. its docstring.

I think this would *probably* be more trouble than it's worth (we don't want to re-implement vast swathes of Rust's name lookup machinery, for example) but it's interesting to think about.

rfk (approved):
Lots of great discussion here, thanks for driving it!

docs/diplomat-and-macros.md
Manishearth: ah! i see

docs/diplomat-and-macros.md
mhammond: yeah, I think we agree that UniFFI's use of the .udl file and Diplomat's decision to restrict where types are exposed do serve the same purpose in that regard. The broader point I'm trying to make though is that application-services would prefer to *not* have those guards in place - ie, it would probably prefer a world where the UDL file didn't exist and nor did any limitation about where these types could be defined.

docs/diplomat-and-macros.md
mhammond: No - diplomat doesn't have that problem because the struct definition must appear inside the module, so the foreign bindings, which parse that module, do know the struct elements. This is the same as we discovered in #416 - that forcing all type definitions into the single ffi module would technically work, but the impact that would have on how our code is organized made it less appealing than the status quo.

I added a few words here to try and make that clearer.

docs/diplomat-and-macros.md
mhammond: >  I would expect the duplication to not feel quite so bad

This is a good point, but from my POV, that can almost make the duplication feel worse - eg, you get a rust compiler, and it can be hard to work out whether the UDL needs to change or the rust duplicate of that UDL needs to change - eg, making something optional means adding a `?` to the udl and a corresponding `Option<>` to the Rust - it can be difficult to work out which one you screwed up :)

docs/diplomat-and-macros.md
rfk: Aha, I see, thanks for clarifying 👍🏻

mhammond:
Thanks all - I also added a link to this to the README. I'm going to merge this now - we can always change it later if we think of something else to add or if it becomes outdated in any way.

PR #1148, opened 2022-01-07
[RFC] Simplified decorator approach

bendk:
A big source of issues with #1143 was the complications around having 3 different classes to handle decorators -- the base class, the decorator class, then the decorated class.  This makes the concepts harder to understand than they need to be and also makes the UDL syntax clunky.

This makes me wonder if the approach from #1093 wasn't so bad after all.  I took the ideas from that, modified them based on our experience with custom types, and put it into a docs-first approach.

The good:
  - Much simple because there's a lot less parts.   You're just modifying the UniFFI-generated class, rather than combining it with with a decorator class to create a decorated class.
  - The UDL syntax is more intuitive.
  - You can also decorate top-level functions

The downside is that decorators are fixed, you can't dynamically create objects with a particular set of decorators.  I can think of a few concrete issues because of this:
  - For testing, it's nice to not schedule things on the IO thread.  In 1143 you could handle this by passing in a decorator class with a dummy `onBackgroundThread` method.  It's slightly harder here, but not that much.  Maybe `onBackgroundThread` could get it's `CoroutineContext` from a global variable that's normally `Dispatchers.IO`, but in the tests we change that to `TestCoroutineScope`.
  - You can't have each instance have a different listener object.  Is this important for Nimbus?

The question seems to be if the flexibility is worth the added complexity.  I'm personally liking the simpler option at this point.

I also tried to make this one address some of the thornier questions about putting this in the `.toml` file: how do you configure if a decorator is async or if it rethrows?  I think it reads okay in the .toml file, although it's not the prettiest.

### DIFF
--- a/docs/manual/src/udl/decorators.md
+++ b/docs/manual/src/udl/decorators.md
@@ -1,0 +1,158 @@
+# Decorator objects
+
+We'll be considering the UDL for an object called `DemoRustObject`:
+
+```webidl
+interface DemoRustObject {
+    string do_expensive_thing();
+
+    Sequence<i32> do_expensive_thing2();
+
+    [Throws=RustyError]
+    bool do_crashy_thing();
+
+    [Throws=RustyError]
+    u8 do_crashy_thing2();
+}
+```
+
+`DemoRustObject` is a `struct` written in Rust, and uniffi generates a class of that name in the
+foreign languages which forward method calls to the Rust implementation.
+
+## Problem
+
+Just by the names of these methods, we can see that the application might want to wrap the
+`DemoRustObject` methods with some app-specific code.
+
+* it may want to run `do_expensive_thing()` and `do_crashy_thing2()` in an `async` function that
+  executes off the main thread.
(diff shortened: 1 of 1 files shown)


bendk:
>    the keys in a decorator method's table define a method signature. Perhaps we should use an actual method signature, with a parser to parse the string.

That's very interesting.  If we went that route, maybe we should put the signatures shouldn't be in the `.toml` file, but in some sort of IDL.  We could have a bindings-side UDL which handles custom types and decorators.  There's also revived interested in moving to a macro-based approach, which opens up a new set of possibilities.

I should be clear that I don't think `uniffi.toml` is definitely the best place for specifying decorators (or the custom types).  I just think it's not too bad, it's specific to the bindings, and it's something we already have support for.  Even if we merged some code that implemented this, I'd be open to thinking about moving the specifications to a different file at some point.

>   perhaps a fruitful area of inquiry may be to explore extending your Custom types to include generics.

That's an interesting idea, so would this mean a custom type that could wrap any type in a generic?

bendk:
This one is very old and not going to get merged.  I'm going to close it.

PR #1150, opened 2022-01-17
Allow new bindings to exist in external crates ADR

tarikeshaq:
a part of #299

An ADR/RFC describing a proposal to separate the binding generators into their own crates.

I'm looking for the following feedback:
- Feedback on the overall concept (i.e should we do this, I'm confident we want to, but open for my mind to change!)
- Feedback on the described implementation design and is there another approach I haven't thought of that could be better? Are there any big no-nos that you can catch?
- Feedback on testability for crates outside our repository (the last section at the bottom). Mainly whether this is an urgent problem to solve, or if we can discuss it separately (either now or whenever the first crate that wants to run tests against our fixtures arrives)

Rendered: https://github.com/mozilla/uniffi-rs/blob/binding-generator-seperate/docs/adr/0007-enable-implementing-bindings-seperately.md


docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+Chosen option:
+
+*  **[Option 2] Expose a trait in `uniffi_bindgen` that can be implemented by other crates, to separate the implementation of the bindings**
+
+### Changes Proposed
+#### The exposed trait for binding generators to implement
+`unffi_bindgen` would define a public trait, `BindingGenerator` that includes one main function:
+```rs
+/// Takes in the ComponentInterface and writes the bindings into directory `out_dir`
+fn write_bindings<P: AsRef<Path>>(&self, ci: &CompenentInterface, out_dir: P) -> Result<()>;
```
badboy, line 89:
```suggestion
fn write_bindings<P: AsRef<Path>>(&self, ci: &ComponentInterface, out_dir: P) -> Result<()>;
```

How would we handle configuration (`uniffi.toml`)?
Right now we _first_ need to parse the UDL to then load the configuration and adjust it according to the target binding.
If we just pass the configuration file path (or leave that up to the binding implementation fully) we lose control that it's all neatly inside a single uniffi.toml

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+
+1. A user would invoke `uniffi-bindgen-kotlin bindings -o <OUT_PATH> <UDL_PATH>`
+1. `uniffi-bindgen-kotlin` would call a generic function `generate_bindings` in `uniffi-bindgen` that takes type parameter `B: BindingGenerator`
+1. `uniffi-bindgen` parses the `udl`, then calls the generic type's `write_bindings` function (which lives in `uniffi-bingen-kotlin`)
+
+Version compatibility is a serious consideration with this approach, and we would benefit from exposing a `scaffolding` option to `uniffi-bindgen-kotlin`. This way the same `uniffi-bindgen-kotlin` is used to generate both scaffolding and bindings. We can also add a warning to `uniffi-bindgen`'s scaffolding that we recommend using a specific binary (like `uniffi-bindgen-kotlin`) to generate the scaffolding as well. Note: The scaffolding generating logic would still live in `uniffi-bindgen` we would only expose it to be called. `uniffi-bindgen-kotlin` would not modify the logic at all, this is purely for version compatibility.
```
badboy, line 127: Scaffolding can happen in a crate's `build.rs`. That's foreign-language-independent and it has to stay so.
~~In Glean we generate a single library that could be consumed by different foreign languages, it doesn't make sense to encourage use of a specificy language binding cli for that.~~
Right now we don't need to think about how we build that, a simple `cargo build` will do the right job, as long as we use the same version for the bindings.
Additionally `uniffi` is _also_ a runtime dependency, so another version that must match.
With this we still need to ensure `uniffi-bindgen-{kotlin,swift,python}` are compatible, or make our builds more complex for when they aren't (choosing the right combo of dependency & tool).

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+## **Implementation Option 2: Have `uniffi_bindgen` get dynamically linked with exactly one `uniffi_bindgen_{language}`**
+This approach uses the specific `uniffi_bindgen_{language}` as plugins. It essentially replaces the shelling out in `Implementation Option 1` with calling an exposed unsafe `C` API in a dynamically linked crate. 
+
+I'm not 100% sure how to let `uniffi_bindgen` choose which crate it gets dynamically linked with on run time, but it's possible.
+
+
+## **Implementation Option 3: Have each `uniffi_bindgen_{language}` be run directly**
+This approach has `uniffi_bindgen` only provide scaffolding and traits for each `uniffi_bindgen_{language}` to implement. `uniffi_bindgen` would also include some generic functions that can be called from `uniffi_bindgen_{language}` by treating `uniffi_bindgen` as a library dependency. The flow would look like this:
+
+1. A user would invoke `uniffi-bindgen-kotlin bindings -o <OUT_PATH> <UDL_PATH>`
```
badboy, line 123: Sort of merging Option 1 here: 
We could even provide a simple `uniffi-bindgen` still, which simply shells out to `uniffi-bindgen-$X` based on a `--language` parameter, but no further logic. The rest stays as described.

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+* Bad, because maintainability can grow to be difficult - especially if more bindings are added which the core `uniffi` team is not familiar with
+* Bad, Because testability can also grow to be difficult - as more bindings are added (some type of `gecko-js` is expected in the future) the requirement to test all the bindings together in one repository is difficult to maintain.
+* Bad, Because releases of all the binding generators are tied to the release of `uniffi_bindgen`
+
+### **[Option 2] Expose a trait in `uniffi_bindgen` that can be implemented by other crates, to separate the implementation of the bindings**
+
+* Good, because maintainability will be clear, and members of the community can opt to maintain their own binding generators
+* Good, because testability can improve, where a developer can test a binding generator without building or testing the others
+* Good, because our CI would only need to test the core bindings we maintain, and others can be tested by their own maintainers (for example, a `gecko-js` binding generator should be tested in `mozilla-central` and not here)
+* Good, because a release in one binding wouldn't have an impact on any other ones
```
badboy, line 59: Changes in a binding that also require changes to `uniffi` internals will still impact others and in turn require releases of them.

badboy (commented):
So I generally think exposing the binding API is a neat approach, but it brings up some question about the scaffolding part.

Currently it's easy to support all foreign language bindings we provide within a single code base and no special-casing in builds.
If we move the 3 major bindings out into their own crates we either still need to commit to the same guarantees of keeping them compatible with the scaffolding and the runtime parts.
Or we shift that responsibility to consumers (e.g. Glean, app-services), putting the burden on those projects to figure out the right combination of dependencies & tools.
I don't know of a better way though.

At the very least we need to define the exact guarantees for these bindings.

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+* Deciders: TBD
+* Date: 2022-01-16
+
+Technical Story: [Issue 299](https://github.com/mozilla/uniffi-rs/issues/299)
+
+outdated early prototype: [PR 997](https://github.com/mozilla/uniffi-rs/pull/997)
+
+## Context and Problem Statement
+All the binding generators currently live in the [`uniffi_bindgen`](../../uniffi_bindgen/src/bindings) crate. This creates the following difficulties:
+- All the bindings live in the `uniffi` repository, so the `uniffi` team has to maintain them (or at the very least review changes to them).
```
rfk, line 13: This may be a pro or a con, depending on how much the `uniffi` team likes maintaining the language in question 😆. Regardless, "separate crates" does not necessarily mean "separate repo" - do you intend to move e.g. the Kotlin bindings to their own repo as part of this work, perhaps to simplify CI?

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+- Any changes to a specific binding generator require a new `uniffi_bindgen` release for it to be released. Even if it doesn't impact any of the other bindings.
+- Some bindings require complex build systems to test, and including those build systems in `uniffi` adds its complexity. For example, any type of `gecko-js` bindings would require the mozilla-central build system to build and test.
+- We currently run all the tests for the bindings in our CI and through `cargo test`. This means that if one binding target gets outdated and fails, or if a developer doesn't have the needed libraries installed for one of the targets, the tests would fail.
```
rfk, line 16: Notably missing from the motivators here is the possibility of third-party developers writing bindings for languages they care about but that the core UniFFI devs do not. It may be worth mentioning explicitly whether this is a goal, non-goal or anti-goal of the work.

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+
+This ADR re-evaluates the architecture of how we consume the bindings in hope of improving the developer experience of adding new bindings, testing new and the current bindings and simplifying the ownership of the bindings.
+## Decision Drivers
+
+* Maintainability, we would like to improve the maintainability of our binding generators - this means clear ownership of the bindings.
+* Testability, it should be easy for developers to test the bindings they care about. Without having to navigate and install unfamiliar libraries and build systems.
+* Version compatibility, we need to guarantee that the same version of `uniffi` is used to generate both the scaffolding and the bindings. This is easy to ensure when all the bindings (and the scaffolding) live under the same crate but becomes a concern if they are separated.
+* Developer experience, it shouldn't be harder to write and maintain a new binding generator than it currently is.
```
rfk, line 24: Ideally it should be a little bit easier, because it's simpler to test your work in isolation from the other generators.

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+
+    This means that `uniffi_bindgen` would still handle generic tasks related to binding generation. This includes:
+        1. Parsing the UDL, and passing the `ComponentInterface` to the specific binding generator
+        1. Calling a generic `BindingGenerator::write_bindings` function that is implemented by the specific binding generator
+        1. `uniffi_bindgen` would also expose its `CodeOracle`, `CodeType` and `CodeDeclaration` types to help developers create a standard way to interact with code generation. It shouldn't however, restrict developers to use those concepts if they don't fit a specific binding.
+    But developers would need a separate crate that implements the traits `uniffi_bindgen` exposes.
+
+* **[Option 3] Remove `uniffi_bindgen`, replace it with `uniffi_scaffolding` and have each binding generator implementation fully manage the generation of bindings**
+    This means external crates would have to depend on `uniffi` directly and implement command-line interfaces. This gives full autonomy to the binding generator developer. Developers won't have access to any traits from `uniffi_bindgen`
```
rfk, line 42: Hrm, interesting. It sounds like there's maybe two or three layers of separate decision to be made here:

1. Should we separate bindings generators into different crates at all?
2. How should the separate bindings generators be invoked?
3. What supporting code should we provide to make building bindings generators easier?

The document currently treats (2) as an implementation detail of (3) but I wonder if it may be clearer to think through them the other way around. That is, start from the desired *consumer* experience of how it works to install and invoke a bindings generator, then figure out how to make that a good *developer* experience when building the bindings generators.

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+- We currently run all the tests for the bindings in our CI and through `cargo test`. This means that if one binding target gets outdated and fails, or if a developer doesn't have the needed libraries installed for one of the targets, the tests would fail.
+
+This ADR re-evaluates the architecture of how we consume the bindings in hope of improving the developer experience of adding new bindings, testing new and the current bindings and simplifying the ownership of the bindings.
+## Decision Drivers
+
+* Maintainability, we would like to improve the maintainability of our binding generators - this means clear ownership of the bindings.
+* Testability, it should be easy for developers to test the bindings they care about. Without having to navigate and install unfamiliar libraries and build systems.
+* Version compatibility, we need to guarantee that the same version of `uniffi` is used to generate both the scaffolding and the bindings. This is easy to ensure when all the bindings (and the scaffolding) live under the same crate but becomes a concern if they are separated.
```
rfk, line 23: We may be able to re-frame this constraint a little. IIRC, we don't *guarantee* this even now with everything in the same crate, because you have to invoke `uniffi-bindgen` once for the scaffolding and again for the bindings, and if you really try we probably can't stop you from using a different version of `uniffi-bindgen` for those two calls. I think the questions here are:

* How easy is it to accidentally generate scaffolding and bindings with different versions of UniFFI? (Probably this proposal will make it easier to do that accidentally).
* How much damage can mismatched UniFFI versions cause? (In theory, I think the checksum still prevents memory unsafety here, the failure mode is a hard build- or link-time error and this proposal wouldn't change that).
* How hard is it to debug what's gone wrong if this does happen? (Maybe this proposal is a chance to make it *easier* to debug?)

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+    This means external crates would have to depend on `uniffi` directly and implement command-line interfaces. This gives full autonomy to the binding generator developer. Developers won't have access to any traits from `uniffi_bindgen`
+
+## Pros and Cons of the Options
+
+### **[Option 1] Do nothing**
+
+* Good, because it keeps version compatibility easy to guarantee.
+* Good, it makes it easier to make changes to multiple bindings at a time (in the case of a breaking change in `uniffi`, etc)
+* Bad, because maintainability can grow to be difficult - especially if more bindings are added which the core `uniffi` team is not familiar with
+* Bad, Because testability can also grow to be difficult - as more bindings are added (some type of `gecko-js` is expected in the future) the requirement to test all the bindings together in one repository is difficult to maintain.
```
rfk, line 51:
```suggestion
* Bad, because testability can also grow to be difficult - as more bindings are added (some type of `gecko-js` is expected in the future) the requirement to test all the bindings together in one repository is difficult to maintain.
```

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+* Bad, Because releases of all the binding generators are tied to the release of `uniffi_bindgen`
```
rfk, line 52:
```suggestion
* Bad, because releases of all the binding generators are tied to the release of `uniffi_bindgen`
```

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+
+Overall this option is preferred because the tradeoffs are clear and manageable and because there is a raising need to create bindings for gecko-js, which can't be tested end-to-end without a complex build system.
+
+### **[Option 3] Remove `uniffi_bindgen`, replace it with `uniffi_scaffolding` and have each binding generator implementation fully manage the generation of bindings**
+
+* Good, because maintainability will be clear, and members of the community can opt to maintain their own binding generators
+* Good, because we make it clear that each binding generator is responsible for its own quality
+* Bad, because new binding generators can't take advantage of the prior art of the existing binding generators
+* Bad, because it will be hard to guarantee the quality of binding generators since they won't depend on any traits from `uniffi_bindgen`
+* Bad, because it complicates version compatibility, and we will need more assertions to guarantee it
```
rfk, line 72: It's not clear to me how this "complicates version compatibility" piece would show up in practice, can you give an example of what additional assertions we might need?

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+* Bad, Because releases of all the binding generators are tied to the release of `uniffi_bindgen`
+
+### **[Option 2] Expose a trait in `uniffi_bindgen` that can be implemented by other crates, to separate the implementation of the bindings**
+
+* Good, because maintainability will be clear, and members of the community can opt to maintain their own binding generators
+* Good, because testability can improve, where a developer can test a binding generator without building or testing the others
+* Good, because our CI would only need to test the core bindings we maintain, and others can be tested by their own maintainers (for example, a `gecko-js` binding generator should be tested in `mozilla-central` and not here)
+* Good, because a release in one binding wouldn't have an impact on any other ones
+* Bad, because it complicates version compatibility, and we will need more assertions to guarantee it
+* Bad, because testability for might increase in complexity. Bindings not in this repository will not have access to the extensive fixtures and examples we have. We might need to expose those if take this option.
```
rfk, line 61:
```suggestion
* Bad, because testability might increase in complexity. Bindings not in this repository will not have access to the extensive fixtures and examples we have. We might need to expose those if take this option.
```

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+
+### **[Option 1] Do nothing**
+
+* Good, because it keeps version compatibility easy to guarantee.
+* Good, it makes it easier to make changes to multiple bindings at a time (in the case of a breaking change in `uniffi`, etc)
+* Bad, because maintainability can grow to be difficult - especially if more bindings are added which the core `uniffi` team is not familiar with
+* Bad, Because testability can also grow to be difficult - as more bindings are added (some type of `gecko-js` is expected in the future) the requirement to test all the bindings together in one repository is difficult to maintain.
+* Bad, Because releases of all the binding generators are tied to the release of `uniffi_bindgen`
+
+### **[Option 2] Expose a trait in `uniffi_bindgen` that can be implemented by other crates, to separate the implementation of the bindings**
```
rfk, line 54: FWIW, when reading this I got a bit hung up on the specific "expose a trait" part of this as a concrete implementation detail, but I think many of the pros/cons listed here are not specific to the implementation details and would apply to any "move the generators into separate crates that depend on some helpers from UniFFI"-shaped solution.

docs/adr/0007-seperate-binding-generators.md
rfk: We also have the option of relaxing the "must use the exact some version of UniFFI requirement" and replacing it with something finer-grained, such as "must use the same version of the UniFFI FFI Contract Specification (tm)".

The swift bindings [do a basic version of this](https://github.com/mozilla/uniffi-rs/blob/7dcbe1219eed36545effb29d8f30ca3425154212/uniffi_bindgen/src/bindings/swift/templates/BridgingHeaderTemplate.h#L15) but as you'll see in that file, it requires the presence of a comment block with ⚠️ signs in it warning folks that they must increment the version number when they make a breaking change.

docs/adr/0007-seperate-binding-generators.md
rfk: I like the idea of parsing `uniffi.toml`, pulling out any config in a section named after the target language, and passing that in as another argument here. It would be simpler for implementors and encourage consistency for end-users.

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+1. `uniffi-bindgen-kotlin` asserts that it was compiled using the same `0.16` version, otherwise it panics.
+1. `uniffi-bindgen-kotlin` then calls a `generate_bindings` function in `uniffi-bindgen` (as a library dependency)
+1. `uniffi-bindgen`'s `generate_bindings` takes a generic parameter type that implements a trait `BindingGenerator`.
+1. `uniffi-bindgen` parses the `udl`, then calls the generic type's `write_bindings` function (the implementation lives in `uniffi-bingen-kotlin`)
+
+
+## **Implementation Option 2: Have `uniffi_bindgen` get dynamically linked with exactly one `uniffi_bindgen_{language}`**
+This approach uses the specific `uniffi_bindgen_{language}` as plugins. It essentially replaces the shelling out in `Implementation Option 1` with calling an exposed unsafe `C` API in a dynamically linked crate. 
+
+I'm not 100% sure how to let `uniffi_bindgen` choose which crate it gets dynamically linked with on run time, but it's possible.
```
rfk, line 117: As a consumer I think I would find this kind of direct linking weird. `cargo` and its subcommands have set a pretty solid reference point for how extensible build tools in the Rust ecosystem behave, so we'd want to have a good concrete reason to do something a different way to how cargo does it. (Which absolutely we might have! Just, in the absence of any other reasons, "behave in a way that lets users bring their existing mental models from cargo" seems like a good default).

docs/adr/0007-seperate-binding-generators.md
rfk: FWIW I quite like the way that we've protected the language-independent nature of the scaffolding, so I prefer the option where a top-level `uniffi-bindgen` continues to make that distinction clear.

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+1. A user would invoke `uniffi-bindgen-kotlin bindings -o <OUT_PATH> <UDL_PATH>`
+1. `uniffi-bindgen-kotlin` would call a generic function `generate_bindings` in `uniffi-bindgen` that takes type parameter `B: BindingGenerator`
+1. `uniffi-bindgen` parses the `udl`, then calls the generic type's `write_bindings` function (which lives in `uniffi-bingen-kotlin`)
+
+Version compatibility is a serious consideration with this approach, and we would benefit from exposing a `scaffolding` option to `uniffi-bindgen-kotlin`. This way the same `uniffi-bindgen-kotlin` is used to generate both scaffolding and bindings. We can also add a warning to `uniffi-bindgen`'s scaffolding that we recommend using a specific binary (like `uniffi-bindgen-kotlin`) to generate the scaffolding as well. Note: The scaffolding generating logic would still live in `uniffi-bindgen` we would only expose it to be called. `uniffi-bindgen-kotlin` would not modify the logic at all, this is purely for version compatibility.
+
+<br />
+<br />
+
+The old prototype in https://github.com/mozilla/uniffi-rs/pull/997 did Implementation Option 1, **I suggest we do Implementation Option 3 as its easier to reason about.**
```
rfk, line 132: Can you say more about the "easier to reason about" piece here? I believe you, because you've tried it 😄.  But I don't have a good sense of what extra cognitive load gets added from Option 1. Is it the version-checking in the handoff to the language-specific binary?

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+
+<br />
+<br />
+
+### Testing
+An important consideration for splitting the binding generators is the testing story for crates that live outside of the main `uniffi-rs` repository. We have a high-level decision to make:
+Should we expose our fixtures as testing standards? And if yes, how do we do that.
+I suggest that we leave this as an open question and open another issue to discuss it specifically.
```
rfk, line 140: Yeah, I agree that this isn't worth blocking on if you've got a clear idea of how to move forward. My gut feeling is that publishing a deliberate test-fixtures crate is probably the right thing, but that's a whole lot of speculative work at this point.

rfk (commented):
Thanks for writing this up! I'll be very interested in following along the progress here, and tried to leave a few comments from my admittedly-decaying historical perspective for completeness.

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+There are a few specific ways to implement this
+
+## **Implementation Option 1: Have `uniffi_bindgen` shell out to `uniffi_bindgen_{language}`**
+Using this approach, we would recommend users to install both the `uniffi_bindgen` **binary** and the `uniffi_bindgen_kotlin` (using kotlin as an example) **binary**. Then, the user would run the following:
+1. The user runs `uniffi-bindgen generate -l kotlin -o <OUT_PATH> <UDL_PATH>`
+1. `uniffi-bindgen` gets its own version, let's say `0.16` as an example.
+1. `uniffi-bingen` shells out to `uniffi-bingen-kotlin -o <OUT_PATH> -v 0.16 <UDL_PATH>`
+1. `uniffi-bindgen-kotlin` asserts that it was compiled using the same `0.16` version, otherwise it panics.
```
rfk, line 108: This thing where there's a little handoff where we say "I'm version X, are you also version X" is an example of what I was thinking of when I said that we might be able to make the debugging experience easier. You can print a whole error message here, not just whatever compile-time assertion failure we currently manage to emit in such cases.

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+1. `uniffi-bindgen-kotlin` then calls a `generate_bindings` function in `uniffi-bindgen` (as a library dependency)
+1. `uniffi-bindgen`'s `generate_bindings` takes a generic parameter type that implements a trait `BindingGenerator`.
+1. `uniffi-bindgen` parses the `udl`, then calls the generic type's `write_bindings` function (the implementation lives in `uniffi-bingen-kotlin`)
+
+
+## **Implementation Option 2: Have `uniffi_bindgen` get dynamically linked with exactly one `uniffi_bindgen_{language}`**
```
rfk, line 114: Why "exactly one"?

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,142 @@
...
+- Some bindings require complex build systems to test, and including those build systems in `uniffi` adds its complexity. For example, any type of `gecko-js` bindings would require the mozilla-central build system to build and test.
+- We currently run all the tests for the bindings in our CI and through `cargo test`. This means that if one binding target gets outdated and fails, or if a developer doesn't have the needed libraries installed for one of the targets, the tests would fail.
+
+This ADR re-evaluates the architecture of how we consume the bindings in hope of improving the developer experience of adding new bindings, testing new and the current bindings and simplifying the ownership of the bindings.
+## Decision Drivers
+
+* Maintainability, we would like to improve the maintainability of our binding generators - this means clear ownership of the bindings.
+* Testability, it should be easy for developers to test the bindings they care about. Without having to navigate and install unfamiliar libraries and build systems.
```
bendk, line 22: I think there are two sides to this.  We want to make it easy to ignore languages that we don't care about (Ruby, maybe Python), but we want to make it easy to test all the languages we do (Kotlin and Swift).  I'm hoping that whatever solution we end up with, it's easy to run the test fixtures for all languages I care about with a single command.

docs/adr/0007-seperate-binding-generators.md
bendk: Seems like the right way, but then you run into the issue that parsing `uniffi.toml` means parsing both the general config section and language-specific config section.  I guess `BindingGenerator` would have an associated type for the language-specific config, and we would use that somehow in a generic config parser function.

docs/adr/0007-seperate-binding-generators.md
badboy: I would appreciate if we _do_ care about Python. Glean's gonna be a user of that.

docs/adr/0007-seperate-binding-generators.md
tarikeshaq: Not as a part of this work - but it'll be interesting to have the discussion in the future 😬 The upcoming firefox desktop bindings will probably be the first to live outside the repo 😄

docs/adr/0007-seperate-binding-generators.md
tarikeshaq: Yeah that's a really good point - I'll restructure the options in terms of the consumer experience and have the "defining traits" etc be all apart of the "changes proposed section I have there"

docs/adr/0007-seperate-binding-generators.md
tarikeshaq: Good catch, I should have clarified this and made a little more concrete! It's the same argument in Option 1 where since they scaffolding and generation don't live in the same crate it's easier to have a mismatch.. an example would be if we add a `version` to the CLI of the bindings we own and assert it's the same version they bindings were compiled against... not perfect and mismatch can still happen though 🙃

docs/adr/0007-seperate-binding-generators.md
tarikeshaq: Eeeh I'm a little fuzzy on this whole option but since it's "possible" thought I'd note it down - the idea is `uniffi_bindgen` would call an `extern C` function without knowledge of who actually implements it.. and based on who it got linked with it "just works" - I have little experience and comfort with this type of communication so 😅

docs/adr/0007-seperate-binding-generators.md
tarikeshaq: Yeah it's pretty much the handoff that made me think this, and the hopping across processes makes it a little harder to debug - for me on the debugger - but I do agree with your note that we can take advantage of the handoff to provide more descriptive error messages about version mismatches, etc.

**Seeing the feedback from you and @badboy makes me want to revert back to option 1!** I think I'm now convinced that
1. The benefit from losing that extra hop out of `uniffi_bindgen` isn't worth having our consumers try to enforce compatibility across different `uniffi_bindgen_{lang}`(especially like Glean and App Services where we have multiple bindings at a time)
2. `uniffi_bindgen` staying as the source of truth (with it passing it's version to the other crates) has a good benefit of reducing metal load to keep track of which binary to call

docs/adr/0007-seperate-binding-generators.md
tarikeshaq: > We also have the option of relaxing the "must use the exact some version of UniFFI requirement" and replacing it with something finer-grained, such as "must use the same version of the UniFFI FFI Contract Specification (tm)".

This could be a really good follow-up to this work!! I also really like the three questions you framed so I'll steal those and rephrase them directly into the doc 😝

tarikeshaq:
The promised prototype is in #1157, I still have some TODOs to work out there, any changes will make their way back here. It's still in an early stage but it's up for visibility!

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,136 @@
...
+* Developer experience, it should be easier to write and maintain a new binding generator than it currently is.
+* Releases, cutting releases for changes in one binding generator shouldn't harm another.
+
+## Considered Options
+
+* **[Option 1] Do nothing**
+
+    This means keeping everything as-is, and deciding that all binding generators (at least for now) should live under the `uniffi_bindgen` crate.
+
+* **[Option 2] Publish a separate crate per binding generator and have users use the  `uniffi_bindgen` Command line interface to shell out to the specific binding generators**
```
mhammond, line 38: For this and the next option, how does the `builtin-bindgen` feature work?

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,136 @@
...
+Additionally, we can define optional functions to extend testing functionality
+```rs
+/// Compiles the generated bindings using `write_bindings`
+fn compile_bindings<P: AsRef<Path>>(&self, ci: &ComponentInterface, out_dir: P) -> Result<()>;
+
+/// Runs a script in the language that implements the `BindingGenerator` trait
+fn run_script<P: AsRef<Path>>(&self, out_dir: P, script_file: P) -> Result<()>;
+```
+
+Using this approach, we would recommend users to install both the `uniffi_bindgen` **binary** and the `uniffi_bindgen_kotlin` (using kotlin as an example) **binary**. Then, the user would run the following:
```
mhammond, line 120: Unless I'm misunderstanding this, s/we would recommend/it would be necessary/?

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,136 @@
...
+fn write_bindings<P: AsRef<Path>>(&self, ci: &ComponentInterface, config: Config, out_dir: P) -> Result<()>;
+```
+
+Additionally, we can define optional functions to extend testing functionality
+```rs
+/// Compiles the generated bindings using `write_bindings`
+fn compile_bindings<P: AsRef<Path>>(&self, ci: &ComponentInterface, out_dir: P) -> Result<()>;
+
+/// Runs a script in the language that implements the `BindingGenerator` trait
+fn run_script<P: AsRef<Path>>(&self, out_dir: P, script_file: P) -> Result<()>;
```
mhammond, line 117: I don't quite understand this. It seems you are suggesting this method is declared on the `BindingGenerator` trait, but it also runs a script that implements the `BindingGenerator` trait?

(I'm trying to think a little about how this might work for a `gecko-js` binding - `compile_bindings` might generate webidl and/or some header files? `run_script` would be used for tests? Or for a build? Or maybe I'm missing something :)

mhammond (commented):
Thanks Tarik - I'm a little confused by the section on testing etc, but that's OK - maybe instead of trying to expand that you could make it shorter and do more hand-waving, so the details can be worked out later as I don't think it changes the overall idea.

In general though, I'm 👍 to this.

bendk:
> I don't want to labor this point, but saying "the bindings we/they care about" obviously depends on who "we/they" are. For example, would we accept a good patch from an external contributor that breaks some bindings? Even if that binding was Kotlin or Python? If the answer is "yes", then I think the point as written is fine. If the answer is "no", then there's an implied set of "core bindings" which we should make explicit.

> (Personally, I think I'd be OK with "yes", with semver allowing consumers to avoid updating if the bindings "you" care about are broken, maybe with guidance about how to avoid this happening automatically by recommending you pin to an exact minor version or something?)

I think this is a very important point and the final ADR should have a description of how we handle breaking changes in UniFFI vs an external crate.  I like the semver suggestion.  I also think if we can, we should make things "break gracefully".  For example, if we add a method to `CodeType` or `CodeOracle` to implement a new feature, we should have a default implementation that calls `unimplemented!()`.  That way you can to use the bindings with the newest UniFFI as long as you don't use that feature.

docs/adr/0007-seperate-binding-generators.md
tarikeshaq: Yeah it will be necessary to install the `uniffi_bindgen_kotlin` binary, but not the `uniffi_bindgen`.. but I feel like having `uniffi_bindgen` installed is almost a necessity (users have to use it to generate the scaffolding for example...) I'll change the wording to away from "recommend" to "users would have to install.."

docs/adr/0007-seperate-binding-generators.md
tarikeshaq: Yeah `run_script` is purely for testing purposes - the comment up there is a little misleading, it should say "Runs a script in the target language, useful for running tests"

docs/adr/0007-seperate-binding-generators.md
tarikeshaq: ideally, it shouldn't change! The `builtin-bindgen` feature should keep using a built-in version of `uniffi-bindgen` for generating the scaffolding code. 

The [`embeded-bindgen`](https://github.com/mozilla/application-services/tree/main/tools/embedded-uniffi-bindgen) we use in app-services, can also stay the same - if a version mismatch happens between the built-in `uniffi_bindgen` in our `embeded_bindgen` and the version of `uniffi_bindgen_kotlin`, error handling logic will log the version mismatch clearly and instruct users to use the right version of `uniffi_bindgen_kotlin`... it's not perfect 😞 but maybe a good start?

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,136 @@
...
+This ADR re-evaluates the architecture of how we consume the bindings in hope of improving the developer experience of adding new bindings, testing new and the current bindings and simplifying the ownership of the bindings.
+## Decision Drivers
+
+* Maintainability, we would like to improve the maintainability of our binding generators - this means clear ownership of the bindings.
+* Testability, it should be easy for developers to test the bindings they care about. Without having to navigate and install unfamiliar libraries and build systems.
+* Version compatibility, this refers to the same version of `uniffi` being used to generate both the scaffolding and the bindings. This includes:
+    - How easy is it to accidentally have a version mismatch
+    - How much damage is caused by the the version mismatch
+    - The difficulty of debugging a version mismatch
+* Developer experience, it should be easier to write and maintain a new binding generator than it currently is.
```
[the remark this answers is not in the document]
tarikeshaq: my very naive thought is that it could be easier for third-party developers to write their own bindings without having it live in this repository could make it easier... but I don't have anything concrete to back that claim 🙃

docs/adr/0007-seperate-binding-generators.md
tarikeshaq: Thanks, @mhammond and @bendk!! This brings up something important. I generally agree with accepting changes that break bindings (even if they are the core ones) - I like the idea of using semver to make sure the versions are not coupled... but one thing we'll have to think carefully about while making this decision is whether that forces us to move the bindings we have in this repository away.

Once we establish that it's okay to break bindings, and bindings should depend on uniffi versions for which they are not broken, the bindings that live in this repository should then depend on a concrete version of uniffi - as opposed to a path dependency.

assuming they stay in this repository, I can think of a couple of implications:
- Releases would have to be done separately for the uniffi libraries and the uniffi_bindgen_lang crates, this makes it hard to reason about what github releases on this repository mean (i.e do we even cut github releases for bindings crates? maybe this is OK?)
- CI could start to be complicated, as it would test each binding crate, possibly using a completely different version of uniffi (which could be OK? can't concretely think of a big drawback to this except that CI might have to download different versions of uniffi)
 
I can open another issue to discuss if we should move the current binding generator crates in this repo to another repository (because that has its own implications and tradeoffs). 

But in the meantime, I'm curious what you think about the semver question given the above, are the implications worth it? or should we just say "no, **for now,** you shouldn't break any binding that lives in this repository because they are all tested and released together with `uniffi`"?

mhammond: > just say "no, for now, you shouldn't break any binding that lives in this repository because they are all tested and released together with uniffi"?

Yes, I think that is the most pragmatic thing to do. I was really just thinking out-loud before - I don't think we should try and optimize for the case of people breaking bindings :) It does raise some interesting questions about policies regarding which bindings are special enough to be in the same repo, but I don't think we necessarily need to answer them now.

tarikeshaq:
Thanks all for all the reviews! I think this is in a good place now - I'm happy to take any thoughts!

I'm hoping to get the ADR landed soonish so I can start working on the actual thing, but I want to make sure this is done right so any concerns are very welcome!

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,323 @@
...
+- We currently run all the tests for the bindings in our CI and through `cargo test`. This means that if one binding target gets outdated and fails, or if a developer doesn't have the needed libraries installed for one of the targets, the tests would fail.
+- It's currently difficult to support third-party developers writing bindings for languages the core team does not wish to maintain.
+
+This ADR re-evaluates the architecture of how we consume the bindings in hope of improving the developer experience of adding new bindings, testing new and the current bindings and clarifying the ownership of the bindings.
+## Decision Drivers
+
+* Maintainability, we would like to improve the maintainability of our binding generators - this means clear ownership of the bindings.
+* Testability, it should be easy for developers to test the bindings they care about. Without having to navigate and install unfamiliar libraries and build systems.
```
bendk, line 25: One goal I have with this is to run the test fixtures even if bindings are in different crates.  Is this part of the plan?  If so, then maybe just add a note about it.

docs/adr/0007-seperate-binding-generators.md
```diff
@@ -0,0 +1,323 @@
...
+
+    Users would only need to install `uniffi_bindgen_{lang}` and run it like `uniffi_bindgen_kotlin <ARGS>`, and each `uniffi_bindgen_{lang}` could opt to support generating scaffolding as well.
+## Pros and Cons of the Options
+
+### **[Option 1] Do nothing**
+
+* Good, because it makes it hard to accidentally use different versions of `uniffi` for scaffolding and bindings.
+* Good, it makes it easier to make changes to multiple bindings at a time (in the case of a breaking change in `uniffi`, etc).
+* Bad, because maintainability can grow to be difficult - especially if more bindings are added which the core `uniffi` team is not familiar with.
+* Bad, because testability can also grow to be difficult - as more bindings are added (some type of `gecko-js` is expected in the future) the requirement to test all the bindings together in one repository is difficult to maintain.
```
bendk, line 65: Building is hard too because you need the entire build system in place to build anything.

bendk (approved):
I left a couple of edit suggestions, but overall it looks great.

mhammond (commented):
Thanks for pushing on this Tarik!

Sorry, I didn't get to the "changes required" section, but I assume that's largely just restating what's in that other PR - but I think this is the right direction and most of my comments aren't really significant in terms of that direction. I'd love to see more discussion about who and how the bindings executables get generated!

(And I wonder if there is a meta-option here - the external binding generators being .so files with uniffi itself arranging to hook things up - that would require a "rust" foreign binding, so probably not viable in the short term, but it sounds cool :)

tarikeshaq:
Thanks @mhammond! For broader communication, we discussed this briefly offline with @bendk and a couple of conclusion:
1. The core idea, we all agree on and see the benefits of separating the bindings
1. The details of the *how*, specifically, how users interact with generating the bindings (the shell out dance, using the bindings generator for scaffolding, etc) still needs ironing out and more discussion.

So, what I'll do, is I'll get an MVP pr up, one that does not implement the shelling out, and simply exposes the traits mentioned in the ADR - this should unblock the upcoming gecko bindings work, but shouldn't get us to commit to any specific direction **just yet**. I'll also work on this in parallel

tarikeshaq:
@mhammond @bdk I reduced the scope here to the landed work in #1201 and the #1206 PR (i.e the minimum to allow desktop/new bindings to work) This should be landable (after I address any comments if something is not worded clearly/etc)

docs/adr/0007-allow-implementing-bindings-seperately.md
```diff
@@ -0,0 +1,98 @@
+# Allow implementing bindings in separate crates.
+
+* Status: Accepted
+* Deciders: teshaq, bdk, mhammond
+* Date: 2022-01-16
```
mhammond, line 5: maybe update the date?

docs/adr/0007-allow-implementing-bindings-seperately.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Context and Problem Statement
+All the binding generators currently live in the [`uniffi_bindgen`](../../uniffi_bindgen/src/bindings) crate. This creates the following difficulties:
+
+- All the bindings live in the `uniffi` repository, so the `uniffi` team has to maintain them (or at the very least review changes to them).
+- Any change to a specific binding generator requires a new `uniffi_bindgen` release for it to be accessible by consumers. Even if it doesn't impact any of the other bindings.
+- Some bindings require complex build systems to test. Including those build systems in `uniffi` would require developers to install those build systems, and CI to do the same. For example, any type of `gecko-js` bindings would require the mozilla-central build system to build and test.
+- We currently run all the tests for the bindings in our CI and through `cargo test`. This means that if one binding target gets outdated and fails, or if a developer doesn't have the needed libraries installed for one of the targets, the tests would fail.
+- It's currently difficult to support third-party developers writing bindings for languages the core team does not wish to maintain.
```
mhammond, line 20: for the sake of brevity, how about rolling this into the first point? ie, appending to that first point "this makes it difficult..."

docs/adr/0007-allow-implementing-bindings-seperately.md
```diff
@@ -0,0 +1,98 @@
...
+
+Before [PR 1201](https://github.com/mozilla/uniffi-rs/pull/1201), it was also impossible to write new bindings that did not live in the [`uniffi_bindgen`](../../uniffi_bindgen/src/bindings) crate.
+
+This ADR proposes allowing third-party crates to implement binding generators.
```
mhammond, line 24: maybe append to this  sentence "and describes the necessary uniffi changes to allow this"? (ie, it's not they weren't "allowed" to, it's more that they were unable to. Or maybe just s/allowing/enabling/? Whatever :)

docs/adr/0007-allow-implementing-bindings-seperately.md
```diff
@@ -0,0 +1,98 @@
...
+* Bad, because testability can also grow to be difficult - as more bindings are added the requirement to test all the bindings together in one repository is difficult to maintain.
+* Bad, because releases of all the binding generators are tied to the release of `uniffi_bindgen`.
+
+### **[Option 2] Create a public API for external crates to implement their own bindings.**
+
+* Good, because ownership will be clear, and members of the community can opt to maintain their own binding generators.
+* Good, because our CI would only need to test the core bindings we maintain, and others can be tested by their own maintainers (for example, a `gecko-js` binding generator should be tested in `mozilla-central` and not here).
+* Good, because a release in external bindings wouldn't have an impact on any internal ones unless it changes internal `uniffi` behavior.
+* Bad, because it's easier to accidentally have a version mismatch. (see ADR https://github.com/mozilla/uniffi-rs/pull/1203).
+* Bad, because testability increases in complexity. We are required to publish fixtures and examples we have. (see https://github.com/mozilla/uniffi-rs/pull/1206)
```
mhammond, line 65: maybe linkify this as \[PR 1026\](...) like done above?  (maybe the same above for the ADR as \[this ADR\]?)

mhammond (approved):
Thanks for driving this through Tarik!

docs/adr/0007-allow-implementing-bindings-seperately.md
```diff
@@ -0,0 +1,98 @@
...
+
+    This means keeping everything as-is, and deciding that all binding generators (at least for now) should live under the `uniffi_bindgen` crate.
+
+* **[Option 2] Create a public API for external crates to implement their own bindings.**
+
+    Developers would have traits exposed they can leverage to implement binding generators that do not live in `uniffi_bindgen`. `uniffi_bindgen` would still handle generic tasks related to binding generation. This includes:
+
+    1. Parsing the UDL, and passing the `ComponentInterface` to the specific binding generator.
+    1. Parsing the `uniffi.toml` and passing the language specific configuration down to the binding generator.
+    1. Calling a generic `BindingGenerator::write_bindings` function that is implemented by the specific binding generator.
```
bendk, line 46: This seems like less of a requirement and more an implementation detail.  I think you can remove it, since you describe this at the bottom.

bendk (approved):
Looks great, thanks for finishing this up!  I had one minor suggestion to add to Mark's.

PR #1156, opened 2022-01-25
Allocate rustbuffer once

bendk:
Instead of growing the buffer as we write to it, calculate the size
up-front and allocate it all at once.

The main reason for the change is to get feedback on the general
approach because this is how I want to implement things for the upcoming
the desktop JS bindings.  Does this seem like a reasonable approach to
buffer handling?

For the desktop JS bindings, the plan is to use the `ArrayBuffer` class
to pack data into.  However `ArrayBuffer` doesn't support growing the
capacity.  I think it's simpler overall to allocate everything up-front
rather than implementing a growable `ArrayBuffer`.

There are a some side benefits though:
  - It should be more efficient, since you don't to re-allocate or copy
    the data over.
  - The write() methods input `ByteBuffer` rather than
    `RustBufferBuilder`.  Using a standard type is nice when you start
    dealing with external types.  You don't have to worry about the fact
    the external type will have it's own `RustBufferBuilder` type.
  - The `lowerIntoRustBuffer()` code is simpler than before

I also standardized a few things:
  - Made FfiConverter an interface
  - Made all the lift/lower/read/write argument names the same
  - Organized the classes so the functions come in the same order


mhammond:
I don't understand why this only touches Kotlin - do we need to do the same for Python, Swift and Ruby? Is there any reason we can't leave the existing bindings alone and only do this for the desktop work you mention?

I'm also slightly concerned about this making things a little more fragile - what happens if the allocated size and the written size don't match? (I possibly missed that check, so it might exist - I did see one when we over-counted, but not undercounted?)

This should perform better, but I'm not sure how much. Nor am I sure how much worse the desktop effort would perform by using a growable buffer, even if we have to implement it manually? To put it another way, what would we do if we decided against this?

(I don't really object to this, but am trying to understand the context better - the status quo seems safe and fast enough, so trying to understand the tradeoffs for an as-yet unimplemented binding)

mhammond:
> * I just prefer this system.
> 
> I could definitely be convinced to keep the status quo.

I'm on the fence, so happy to go with a consensus. If @tarikeshaq and/or @jhugman like it, then go for it :)

bendk:
I want to push for this one a bit more.  I just realized that since Kotlin is using the `FfiConverter` code, we're actually very close to getting external types working there.  The last piece of the puzzle is figuring out the `FfiConverter.write()` signature.  Currently it inputs a `RustBufferBuilder`, but that breaks with external types because each module will have it's own `RustBufferBuilder` class.  If we merge this one, then `write()` just inputs a standard `ByteBuffer`so the issue goes away.

The only other solution I can think of is to [modify the `write()` signature to input a callback](https://github.com/mozilla/uniffi-rs/pull/1046/files#diff-a34468ab8f052ef5fe508ac14b0b4cd866194a1adc0368f0cdb32a9189becdf5R57-R59) that handles the buffer write.  I definitely prefer this change to that one.

mhammond (approved):
> Currently it inputs a `RustBufferBuilder`, but that breaks with external types

OK, that sounds like enough justification to me!

PR #1157, opened 2022-01-26
[Don't Merge] Prototype uniffi_bindgen_kotlin

tarikeshaq:
Relates to #1150 and #299 

Early prototype, quite a few quirks to work out and naming/etc isn't great yet but putting it up for visibility so people can see how I envision the implementation of #1150 to look

I'll clean it up a little, biggest question marks/next steps right now:
- [ ] Work out the best way for the `uniffi_bindgen_{lang}` crate to get the explicit version of the `uniffi_bindgen` it was compiled with. 
   - `cargo_metadata` seems to not work really well when it's a path dependency and I would imagine it should look at a specific version (like a `Cargo.lock` almost?)
   - This could be a follow-up, having it a `const` string might be OK for now as long as it's documented well
- [ ] Have the testing flow follow the same pattern as the `generate` flow, where it shells out to the specific crate
- [ ] Create an issue and start thinking of documenting this process. This should probably go in our `uniffi` book, I imagine a tutorial for developers, something along the lines of "how to build a new backend for `uniffi`" 

when all the above is done, I'll try to get #1150 through another pass (or more) of review. Once the ADR lands I'll morph this into a full implementation of it (and clean up all the duplicate code I still have in this PR)


uniffi_bindgen/src/lib.rs
```diff
@@ -115,6 +115,101 @@ use bindings::TargetLanguage;
...
+///         Self {} // This should implement the merging of two Configs, omitted for example
+///     }
+/// }
+/// 
+/// impl From<&ComponentInterface> for KotlinConfig {
+///     fn from(ci: &ComponentInterface) -> Self {
+///         Self {} // This should implement generating a config from a ComponentInterface,
+///                 // omitted for example
+///     }
+/// }
```
bendk, line 142: I don't think this is new, but this From implementation feels confusing to me.  What if this was named `BindingGeneratorConfig::new(ci: &ComponentInterface)` instead?

uniffi_bindgen/src/lib.rs
```diff
@@ -311,6 +457,34 @@ impl From<&ComponentInterface> for Config {
...
+    fn load_config<P: AsRef<Path>>(component: &ComponentInterface, path: P) -> Result<C> {
+        let default_config: Self = component.into();
+        let path = path.as_ref();
+        let contents = slurp_file(&path)
+            .with_context(|| format!("Failed to read config file from {:?}", &path))?;
+        let loaded_config: Self = toml::de::from_str(&contents)
+            .with_context(|| format!("Failed to generate config from file {:?}", &path))?;
+        Ok(loaded_config.merge_with(&default_config).bindings)
+    }
+}
```
bendk, line 478: I'm not understanding the relationship between `BindingGeneratorConfig` and `BackendConfig`.  Is there a way to simplify this?  If not, I think some docstrings would definitely be useful.

uniffi_bindgen_kotlin/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+                .help("Path to the optional uniffi config file. If not provided, uniffi-bindgen will try to guess from the UDL's file location.")
+            )
+        )
+        .get_matches();
+    let uniffi_version = matches
+        .value_of_os("uniffi_version")
+        .expect("`uniffi_version is a required argument"); // required!
+    if uniffi_version != UNIFFI_VERSION {
+        bail!("Invalid uniffi_version, this version of uniffi_bingen_kotlin only works with version {}, found {:?}", UNIFFI_VERSION, uniffi_version)
+    }
```
bendk, line 76: Could some of this be refactored into shared code somehow?

bendk (commented):
I like the direction a lot.  Left some comments about things that I think could be cleaned up some more or that I didn't completely get.  But overall this seems like a very reasonable and straightforward way to split things up.

mhammond (commented):
Thanks heaps for this - it helps give a lot of context to the ADR.

However, it's not quite clear to me how this would end up looking once all bindings have been given the same treatment as kotlin here. ISTM that doing this for just one language and the rest "later" would be a bad outcome - what are your plans/thoughts here?

What's you thinking around that?

uniffi_bindgen/src/lib.rs
```diff
@@ -115,6 +115,101 @@ use bindings::TargetLanguage;
...
+    ) -> anyhow::Result<()>;
+
+    fn run_script<P: AsRef<Path>>(&self, out_dir: P, script_file: P) -> anyhow::Result<()>;
+}
+
+pub fn generate_backend_bindings<B: BindingGenerator, P: AsRef<Path>>(
+    udl_file: P,
+    config_file_override: Option<P>,
+    out_dir_override: Option<P>,
+    _try_format_code: bool, // TODO: approach for formatting? should this be purely backend specific
```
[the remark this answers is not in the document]
tarikeshaq: Hmmm yeah, in #1171 I removed this argument for now. We can re-add it later - this agree that this should be controlled by the binding generator...

uniffi_bindgen/src/lib.rs
tarikeshaq: I dramatically simplified this in #1171, the `BackendConfig` existed only so the configuration can be under the `bindings` key.. but that's unnecessary complexity 😄 I got rid of that type there so now it's only the trait

tarikeshaq:
Closing this one, we should look at #1205 for a more recent example of separating the bindings

PR #1159, opened 2022-01-28
Doc fixes around custom types

bendk:
Fixed the CHANGELOG entry for the feature.  It's not released yet, but
it was listed in the changelog as being part of 0.16.0.  Looks like the
wires got crossed because the feature was merged a couple days after
the 0.16.0 release.

Updated the docs to refer to them as "Custom types" instead of "Wrapped
types"


CHANGELOG.md
```diff
@@ -8,28 +8,34 @@
 
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.9.0...HEAD).
```
jhugman, line 9: Should the `v0.9.0` change here too?

CHANGELOG.md
```diff
@@ -8,28 +8,34 @@
...
+
+### What's Changed
+
+- Custom types (formerly wrapped) now can be configured on the bindings side as
+  well as the scaffolding side.  See the "Custom Types" section of the manual
+  for details.
+
+## v0.16.0 - (_2021-12-15_)
+
+[All changes in v0.16.0](https://github.com/mozilla/uniffi-rs/compare/v0.9.0...v0.16.0)
```
jhugman, line 27: We've done this in a lot of changes in this file, so I don't really know. Is the `v0.9.0` meant to be previous version?

CHANGELOG.md
```diff
@@ -8,28 +8,34 @@
...
    - The trait name has been renamed to `UniffiCustomTypeConverter` from `UniffiCustomTypeWrapper`
    - The method names of that trait have been renamed to `into_custom()` / `from_custom()` instead of `wrap()` and `unwrap()`
    - The associated type has been renamed to `Builtin` instead of `Wrapped`
-- Error handling when converting custom types has been updated. If your `into_custom()`
+
+### What's Changed
+
+- Custom types (formerly wrapped) now can be configured on the bindings side as
+  well as the scaffolding side.  See the "Custom Types" section of the manual
+  for details.
```
jhugman, line 23: Aha, so this hasn't been released yet.

jhugman (approved):
Two questions about version numbers, but this looks fine. Would another release be in order soon?

CHANGELOG.md
bendk: That's a really good question.  I think:
  - It should be the previous version
  - `cargo release` should be updating this but it's not.  We should try to fix the automation or at least update the release docs.

Does that makes sense to you?

bendk: I'm really unsure of what happened, but I just pushed a change that fixes the version numbers here.  Maybe once we merge this PR we could make a new release and test if they get updated correctly or not.

tarikeshaq: I think https://github.com/mozilla/uniffi-rs/blob/main/uniffi/release.toml#L11 is the culprit here, it has a constant v0.9.0 😬

bendk: Nice catch.  I'm not sure exactly how that works, but I pushed a blind update to it.  Does it look good to you?

tarikeshaq: Looks right to me (I'm not exactly sure how it works either, but that seems like it should be good 😅) we can land this and cut a release and catch it in action in the release PR

bendk:
Looks like this has approval, I'm going to merge and cross my fingers that the next release has the right version number.

PR #1162, opened 2022-02-03
Python: Set default values in dict constructor

badboy:


uniffi_bindgen/src/bindings/python/templates/macros.py
```diff
@@ -53,6 +53,21 @@
     {%- endfor %}
 {%- endmacro %}
 
+{#-
+// Field lists as used in Python declarations of Records and Enums.
+// Note the var_name and type_name filters.
```
bendk, line 58: Maybe remove the "and type_name" part.  I don't think you end up using that filter since it's Python.

bendk (approved):
Looks great overall, just had a small doc string suggestion.

uniffi_bindgen/src/bindings/python/templates/macros.py
badboy: classic copy-pasta error.

PR #1163, opened 2022-02-03
(cargo-release) version 0.17.0

bendk:
I just made the 0.17.0 release, now I think we need to merge the changes back into main.

Did I miss something, or should this be included in https://github.com/mozilla/uniffi-rs/blob/main/docs/release-process.md?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@
 
 ## [[UnreleasedVersion]] - (_[[ReleaseDate]]_)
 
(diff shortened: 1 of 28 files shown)


badboy (approved):

badboy:
This should probably be included there.
I merged another PR, which thus messes up the "real" history of the `main` branch a bit when we merge this. That's all not optimal (we do the same on Glean and while it makes patch releases easy, I still think there needs to be better ways)

veehaitch:
Please also add a `v0.17.0` tag 🙂

bendk:
Just pushed.  Sorry about the delay.

PR #1165, opened 2022-02-07
Change scaffold logs to trace level

badboy:
These can be very noisy when relying on other debug messaging from libraries.


jhugman (approved):
Formal LGTM. ✅

mhammond:
app-services actually likes them - trace logs don't end up in logcat by default in Fenix. Eg, when uniffying places, we removed a number of explicit log.debug calls in the hand-written ffi because they were redundant with uniffi. I don't object to revisiting and re-adding log calls in our components, but they are offering actual value to us today

mhammond:
(another potential use-case for decorator-like functionality, particularly if decorators could be "chained" in some way!)

badboy:
Closing then. For now I can filter those (by default) on the Glean end (and unfilter them for debugging)

PR #1166, opened 2022-02-10
Don't run rustfmt on scaffolding when using builtin bindgen

badboy:
We can't guarantee `rustfmt` is available and there's no other way to
skip it.


bendk:
I guess the only question is if the formatting code has enough value that it could be supported as value in `uniffi.toml` or maybe a CLI argument.  Personally I wouldn't miss it, but I wonder what others think.

mhammond (approved):
Does seem like a footgun. Someone might choose to contribute back a way of making this optional, but I see no need to block on that. I suspect formatting foreign bindings might also be a future footgun, but let's cross that bridge when we come to it.

badboy:
> I guess the only question is if the formatting code has enough value that it could be supported as value in `uniffi.toml` or maybe a CLI argument. Personally I wouldn't miss it, but I wonder what others think.

There is a CLI argument for `uniffi-bindgen`. But this is explicitly for the case where there is no `uniffi-bindgen` tool used, but it's run as part of the build.rs.
I argue that there's no real benefit formatting that code, it's somewhere hidden in the `OUT_DIR` anyway.

PR #1167, opened 2022-02-10
Python/Swift FfiConverters

bendk:
This brings these bindings in sync with Kotlin after #1144.

The main benefits of these changes is:
  - The FFI conversion is more standardized, which makes custom and external types easier to implement.  This change allowed me to get external types working on Python again.  I think Swift should be ready to implement both now, but I didn't do that in this PR.
  - We can remove the `(lift|lower|read|write)` `CodeType` functions, which results in a lot less boilerplatey code

This PR contains a ton of changes, but most were very mechanical.  The main exception to this was the Swift `CallbackInterface` changes.  I needed to refactor a bit to make the initialization work with the new system.


uniffi_bindgen/src/bindings/swift/gen_swift/callback_interface.rs
```diff
@@ -2,8 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
```
badboy, line 2: I would _like_ to comment on the commit message .... but GitHub doesn't allow that.

Anyway, typo fixes in the second commit:

```patch
- This types implemen the `FfiConverter` protocol, which replaces
+ These types implement the `FfiConverter` protocol, which replaces
```

and 

```patch
- code, which allows deleting a lot of lines from the gen_python modules.
+ code, which allows deleting a lot of lines from the gen_swift modules.
```

badboy (approved):
I didn't trace all changes exactly, but as you said it's mostly mechanical changes to get to the same consistent pattern.

r+ from me (with minor comments regarding typos in the commit message)

PR #1168, opened 2022-02-11
Added external type support for Kotlin

bendk:
After #1144 and #1156, we're ready to implement external types on Kotlin.

- Updated `kotlin::compile_bindings()` so that uses the same classpath
  as `run_script()`.  This is needed to allow importing the external types.
- Made Kotlin `FfiConverter` objects public so that they can be imported
- Added a public typealias for Kotlin custom types so that they can be
  imported
- Changed ext_types fixture to not use `val` for argument names.

There are still a couple improvements to be done.  I'm thinking we can do a second pass for those.

### DIFF
--- a/fixtures/ext-types/guid/src/guid.udl
+++ b/fixtures/ext-types/guid/src/guid.udl
@@ -15,12 +15,12 @@ dictionary GuidHelper {
 namespace ext_types_guid {
     // Note this intentionally does not throw an error - uniffi will panic if
     // a Guid can't be converted.
-    Guid get_guid(optional Guid? val);
+    Guid get_guid(optional Guid? value);
 
     // Uniffi will handle failure converting a string to a Guid correctly if
     // the conversion returns `Err(GuidError)`, or panic otherwise.
     [Throws=GuidError]
-    Guid try_get_guid(optional Guid? val);
+    Guid try_get_guid(optional Guid? value);
 
-    GuidHelper get_guid_helper(optional GuidHelper? vals);
+    GuidHelper get_guid_helper(optional GuidHelper? values);
 };
--- a/fixtures/ext-types/lib/uniffi.toml
+++ b/fixtures/ext-types/lib/uniffi.toml
@@ -1,2 +1,11 @@
 [bindings.python]
 cdylib_name = "uniffi_ext_types_lib"
+
+[bindings.kotlin]
+cdylib_name = "uniffi_ext_types_lib"
+
+[bindings.kotlin.external_packages]
+# Map the crate names from [External={name}] into Kotlin package names
+custom-types = "customtypes"
+ext-types-guid = "uniffi.ext_types_guid"
+uniffi-one = "uniffi.uniffi_one"
--- a/uniffi_bindgen/src/bindings/kotlin/mod.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/mod.rs
@@ -59,9 +59,8 @@ pub fn compile_bindings(config: &Config, ci: &ComponentInterface, out_dir: &Path
     let status = Command::new("kotlinc")
         // Our generated bindings should not produce any warnings; fail tests if they do.
         .arg("-Werror")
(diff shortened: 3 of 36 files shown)


fixtures/ext-types/lib/uniffi.toml
```diff
@@ -1,2 +1,11 @@
...
 cdylib_name = "uniffi_ext_types_lib"
+
+[bindings.kotlin]
+cdylib_name = "uniffi_ext_types_lib"
+
+[bindings.kotlin.external_packages]
+# Map the crate names from [External={name}] into Kotlin package names
+custom-types = "customtypes"
+ext-types-guid = "uniffi.ext_types_guid"
+uniffi-one = "uniffi.uniffi_one"
```
bendk, line 11: This allows us to decouple the names from the UDL file with the package name that we generate.  Maybe we should do something like this for Python.

uniffi_bindgen/src/bindings/kotlin/mod.rs
```diff
@@ -113,3 +95,29 @@ pub fn run_script(out_dir: &Path, script_file: &Path) -> Result<()> {
     }
     Ok(())
 }
+
+// Calculate the classpath string to use for testing
+pub fn classpath_for_testing(out_dir: &Path) -> Result<OsString> {
```
bendk, line 100: I don't really love this system but it works okay.  I made #1169 to improve on it.

uniffi_bindgen/src/bindings/kotlin/mod.rs
```diff
@@ -59,9 +59,8 @@ pub fn compile_bindings(config: &Config, ci: &ComponentInterface, out_dir: &Path
     let status = Command::new("kotlinc")
         // Our generated bindings should not produce any warnings; fail tests if they do.
         .arg("-Werror")
-        // Reflect $CLASSPATH from the environment, to help find `jna.jar`.
         .arg("-classpath")
-        .arg(env::var("CLASSPATH").unwrap_or_else(|_| "".to_string()))
+        .arg(classpath_for_testing(out_dir)?)
```
mhammond, line 63: huh - I didn't realize `compile_bindings` was only used for tests - do you mind adding a comment in the block at the top of the function noting that?

uniffi_bindgen/src/bindings/kotlin/templates/Float32Helper.kt
```diff
@@ -1,4 +1,4 @@
-internal object FfiConverterFloat: FfiConverter<Float, Float> {
+public object FfiConverterFloat: FfiConverter<Float, Float> {
```
mhammond, line 1: I'm mildly surprised you need to make all `FfiConverter` based objects public, but I assume that's necessary due to Kotlin requirements.

mhammond (approved):
This looks great, thanks!

fixtures/ext-types/guid/src/guid.udl
```diff
@@ -15,12 +15,12 @@ dictionary GuidHelper {
 namespace ext_types_guid {
     // Note this intentionally does not throw an error - uniffi will panic if
     // a Guid can't be converted.
-    Guid get_guid(optional Guid? val);
+    Guid get_guid(optional Guid? value);
```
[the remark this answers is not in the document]
mhammond: (eg, the fact Rust itself chose `fn` and `impl` over the whole words makes me think `val` is a more "rusty" name than the fully spelled out version)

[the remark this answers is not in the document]
bendk: I like `val` too, but it's a keyword in Kotlin, so we can't use that as an argument name.

uniffi_bindgen/src/bindings/kotlin/templates/Float32Helper.kt
bendk: We're generating a different package for each UDL file, but with external types one package needs to import from the other so I think these need to be public.  I'll add a comment about this.

PR #1172, opened 2022-02-15
Annotate generated Python code to avoid type checking

badboy:
The generated code has no type annotations and doesn't conform to what
e.g. mypy expects, thus throwing errors.
Unfortunately there's no way to tell mypy from the outside to ignore
that one included module.


mhammond (approved):
That sounds like a significant limitation in mypy, but I can't see any harm here. Maybe add a link to this issue - mypy isn't particularly mainstream, so the casual reader might be able to learn what the hell this means? As a long time Python user I'd be scratching my head.

badboy:
I actually changed it to be mypy-specific now (turns out even with `type: ignore` you still need more annotations in importing code to not fail...)

mhammond (approved):
lovely, thanks!

PR #1174, opened 2022-02-16
Couple updates to the BindingGenerator trait

bendk:
Use &Path rather than AsRef<Path>.  We can call `as_ref()` before
calling into `BindingGenerator` and keep the implementation code simple.

Added code to extract the binding-specific config from the `uniffi.toml`
config that contains it.

Instead of leveraging `From<&ComponentInterface>` and `MergeWith`, have
some `BindingGeneratorConfig` methods that explicitly do what we need
them to do: update the values read from TOML with defaults from
`ComponentInterface`.

When implementing custom types, I wasn't sure how to properly implement
the `MergeWith` trait.  For example, what if we have 2 configs, each
with a custom type map should we try to take entries from both?  Having
methods specific to the actual process can avoid issues like that.  In
the custom types case, we never get custom types from the component
interface, so we don't need to worry about merging them.

### DIFF
--- a/uniffi_bindgen/src/lib.rs
+++ b/uniffi_bindgen/src/lib.rs
@@ -104,6 +104,7 @@ use std::{
     fs::File,
     path::{Path, PathBuf},
     process::Command,
+    str::FromStr,
 };
 
 pub mod backend;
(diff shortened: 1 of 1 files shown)


bendk:
I tried to get a working example of how this would work, but it was tricky.  I did put together a hand-wavey demo though: https://github.com/bendk/uniffi-rs/commit/19481197887fbddb5c3ba93ac4be1dbc26d24342

tarikeshaq (approved):
This looks good to me! Lets land it on the branch and as we go we can keep updating it

PR #1175, opened 2022-02-18
Fixing unittests

bendk:
I'm not sure why but this stderr output changed between when we created
the PR for it and when we landed it.  Hopefully this fixes the tests.


badboy (approved):
Probably because of #1166

PR #1177, opened 2022-02-22
Seperate bindings mvp

bendk:


uniffi/src/lib.rs
```diff
@@ -639,6 +639,16 @@ where
     }
 }
 
+// This function is not intended to be called, but can be used as for the to workaround
+// https://github.com/rust-lang/rust/issues/50007
+//
+// Basically, if another library adds uniffi as an external crate, then builds into a dylib, that
+// library normally won't have the symbols from uniffi.  However, if that library calls an extern
+// function from function, then it *will* have the symbols.  So this is a dummy function that the
```
tarikeshaq, line 647: nit: `an extern function from function` seems off?

uniffi_bindgen/src/lib.rs
```diff
@@ -242,18 +246,19 @@ pub trait BindingGenerator: Sized {
 /// - `udl_file`: The path to the UDL file
```
tarikeshaq, line 246: nit: we should change the rustdocs to reflect the new arguments

uniffi_bindgen/src/lib.rs
```diff
@@ -242,18 +246,19 @@ pub trait BindingGenerator: Sized {
...
 /// - `config_file_override`: The path to the configuration toml file, most likely called `uniffi.toml`. If [`None`], the function will try to guess based on the crate's root.
 /// - `out_dir_override`: The path to write the bindings to. If [`None`], it will be the path to the parent directory of the `udl_file`
-pub fn generate_external_bindings<B: BindingGenerator, P: AsRef<Path>>(
-    udl_file: P,
-    config_file_override: Option<P>,
-    out_dir_override: Option<P>,
+pub fn generate_external_bindings(
+    binding_generator: impl BindingGenerator,
+    udl_file: impl AsRef<Path>,
```
tarikeshaq, line 251: just out of curiosity, should we change this to be `&Path` like we did in `write_bindings`? (I know it was a generic parameter before this change, so no need for any changes here 😄)

uniffi_bindgen/src/lib.rs
```diff
@@ -151,7 +151,32 @@ fn load_bindings_config<BC: BindingGeneratorConfig>(
...
+    fn get_entry_from_bindings_table(_bindings: &toml::Value) -> Option<toml::Value> {
+        None
+    }
+
+    fn get_config_defaults(_ci: &ComponentInterface) -> Vec<(String, toml::Value)> {
+        Vec::new()
+    }
+}
+
+impl<'de> Deserialize<'de> for EmptyBindingGeneratorConfig {
```
tarikeshaq, line 173: Can we replace this with a derive on the `EmptyBindingGeneratorConfig`? or does serde complain?

uniffi_bindgen/src/lib.rs
```diff
@@ -113,7 +113,7 @@ pub mod interface;
 pub mod scaffolding;
 
 use bindings::TargetLanguage;
-use interface::ComponentInterface;
+pub use interface::ComponentInterface;
```
tarikeshaq, line 116: Do we need this? I have no objections, but it looks like all of `interface` is pub on line 112 - not that it matters much, there is convenience in getting access to the `ComponentInterface` with one less module.. maybe `interface` should be a private module and only expose `ComponentInterface` 🤷

tarikeshaq (approved):
A few nits and questions, but nothing blocking this 😄 LGTM! 

🚢 :shipit:

uniffi_bindgen/src/lib.rs
bendk: We could, but to me this feels like the right function to put `as_ref()` call.  I don't really have any strong reason for it though, so I could be easily convinced otherwise.

uniffi_bindgen/src/lib.rs
bendk: It's a unit struct, but we want to deserialize from an empty map, so the `derive(Deserialize)` doesn't work.  I'll add a comment about this.

uniffi_bindgen/src/lib.rs
bendk: I'm never completely sure how to handle this.  In the bindings code we use a several other components from `interface`.  Maybe we could make that module private and add `pub use interface::*`.  Now that these are all separate crates, it seems more natural to import from the `uniffi_bindgen` rather `uniffi_bindgen::interface`.

Issue #1178, opened 2022-02-23
Custom types panic! in scaffolding.rs


jhugman:
UDL to make it panic:

```
[Custom]
typedef string JsonObject;

callback interface JexlEvaluator {
    boolean jexl_eval(string expression, JsonObject context);
};
```

Line it's crashing on:

https://github.com/mozilla/uniffi-rs/blob/e7ff0dafda31d413ba94065c2f7039166ae03e94/uniffi_bindgen/src/scaffolding/mod.rs#L52

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-136)
┆friendlyId: UNIFFI-136

jhugman:
@bendk if only all the backends looked the same ;o)

bendk:
This one is confusing me.  Is there a way to produce the panic?  I think we have several custom types as examples and they seem to be working.

jhugman:
I edited comment 0 with the minimal UDL to get to the panic.

Stack traces is: 

```
     0: std::panicking::begin_panic
               at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/panicking.rs:543:12
     1: uniffi_bindgen::scaffolding::filters::type_rs
               at /Users/jhugman/.cargo/registry/src/github.com-1ecc6299db9ec823/uniffi_bindgen-0.17.0/src/scaffolding/mod.rs:52:36
     2: <uniffi_bindgen::scaffolding::RustScaffolding as askama::Template>::render_into
               at /Users/jhugman/.cargo/registry/src/github.com-1ecc6299db9ec823/uniffi_bindgen-0.17.0/src/scaffolding/mod.rs:11:10
     3: <uniffi_bindgen::scaffolding::RustScaffolding as core::fmt::Display>::fmt
               at /Users/jhugman/.cargo/registry/src/github.com-1ecc6299db9ec823/uniffi_bindgen-0.17.0/src/scaffolding/mod.rs:11:10
     4: core::fmt::write
               at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/fmt/mod.rs:1163:17
     5: std::io::Write::write_fmt
               at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/io/mod.rs:1696:15
     6: uniffi_bindgen::generate_component_scaffolding
[...]

PR #1179, opened 2022-02-24
Implementing type_rs for custom types (#1178)

bendk:
It's kind of amazing that we didn't need this before, but I guess the scaffolding code doesn't really need to know the type name for custom types, until you start using it in a `CallbackInterface`.


bendk:
Hmm, this is actually going to require some more work.  Let's discuss in the issue.

PR #1180, opened 2022-02-24
Added support for custom types in callback interfaces (#1178)

bendk:
We just needed to implement `type_rs()`.  Amazingly, we were able to get away
with not defining this filter until now.

The other issue was that the Kotlin callback interface code depends on
AtomicLong, but it doesn't import that type.  I guess we were able to
get away with this because callback interfaces were always paired with
objects, which do import AtomicLong.


bendk:
I think the CI errors are just because of the Heroku outage, but I'm not sure how to re-run them.

mhammond (approved):
This makes sense to me!

PR #1185, opened 2022-03-04
Update to Askama 0.11

bendk:
The new Askama version adds refererences in less places.  This required a few changes:
 - `&dyn fmt::Display` args get change to `&str`
 - Enum litererals (`Type::String`) need a `.borrow()` appended to them
 - Method calls that then get filtered, (`something.type_()|type_name`) need a `.borrow()` appended to them.  In a some cases, this can also be fixed by having the template code use a field from the template struct directly rather than via a method call.

Fixed some incorrect whitespace handling code that was only working because of Askama bugs that were fixed in this version.

I wasn't sure if I wanted to push this for review, but @mhammond convinced me to do it.  The extra `.borrow()`s are a negative but the change to `&str` is a positive.  I also have plans to rework some of this code to avoid the `borrow()`s


mhammond (approved):
Thanks! Mildly surprised trybuild didn't abstract some of the uitest changes away, but 🤷

Issue #1188, opened 2022-03-16
Make the bindings generation code thread-safe


bendk:
I was a bit too relaxed about adding [thread-unsafe code like this](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs#L332-L349) to the bindings generator mods.  That code is fine for normal runs, since we don't do any sort of multi-threading in UniFFI.  However, after working on the desktop JS bindings, I realize that it's not safe for testing, which does use multiple threads.

This is working out okay for now because all of our tests are grouped into fixtures and each fixture only generates the bindings for a language once.  I believe each fixture is it's own binary, so there's no threading issues.  However, if we are going to move bindings into separate crates, then it makes sense to have a single test binary that runs all the tests in different threads.  Those tests need to run `cargo metadata` and `cargo build`.  Having a single binary allows us to cache the results, otherwise all the `cargo` executions fight over the lock file.

It might be possible to figure out a different system to cache the results, but that could get quite complex.  In addition, my general feeling is that it was a bad idea to introduce the thread-unsafe code.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-142)
┆epic: Rust Components on Desktop
┆friendlyId: UNIFFI-142
┆sprintEndDate: 2022-04-15

Issue #1183, opened 2022-03-02
External type tests depend on building other packages


bendk:
The Pythonext-types tests only pass if you build the crates for the external types before running the tests. Kotlin works, but only because it lazily loads the library. If we ever needed to call a function from the UDL I think it would fail too. The issue is that we are building the .so files in the target/debug/deps directory in this case, but we're expecting them target/debug/ directory.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-139)
┆friendlyId: UNIFFI-139

bendk:
I think this can be fixed by using the `uniffi_testing` library, but that's currently only used by `uniffi-bindgen-gecko-js`.

PR #1191, opened 2022-03-24
Make binding generation threadsafe (#1188)

bendk:
(This is like a mini version of #1189.  That one contains more changes.  I like the changes, but others may not or just want time to review them.  I don't want to block the desktop JS project, so I'm making a smaller PR that can get approved quicker).

Removed the `Config` field from the `CodeOracle` structs.  This allows the struct to be state-less, which makes the filter code threadsafe (#1188).

The main thing we were using the config for was custom types, since the type name depeneded on a config value.  Changed that so we always use the name from the UDL file as the type name, then create a type alias in the helper code.  PR #1189 takes this idea a lot further, this is like a mini version of that.

The other type that needed changes was external types.  But this was much simpler, we just needed to move some code from the `CodeType` struct to `CodeDeclaration` struct.

This change indirectly makes the python ext-types test fail for me. I'm pretty sure this is because of #1183 I just commented it out for now, I think we are going to have a nice solution for this once the external bindings crate code is ready to merge.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/CustomTypeTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/CustomTypeTemplate.kt
@@ -1,6 +1,9 @@
 {%- match config %}
 {%- when None %}
 {#- Define the type using typealiases to the builtin #}
+// Typealias from the type name used in the UDL file to the builtin type.  This
+// is needed because the UDL type name is used in function/method signatures.
+// It's also what we have an external type that references a custom type.
 public typealias {{ name }} = {{ builtin|type_name }}
 public typealias {{ outer|ffi_converter_name }} = {{ builtin|ffi_converter_name }}
 
@@ -11,6 +14,9 @@ public typealias {{ outer|ffi_converter_name }} = {{ builtin|ffi_converter_name 
 {# When the config specifies a different type name, create a typealias for it #}
 {%- match config.type_name %}
 {%- when Some(concrete_type_name) %}
+// Typealias from the type name used in the UDL file to the custom type.  This
+// is needed because the UDL type name is used in function/method signatures.
+// It's also what we have an external type that references a custom type.
 public typealias {{ name }} = {{ concrete_type_name }}
 {%- else %}
 {%- endmatch %}
--- a/uniffi_bindgen/src/bindings/kotlin/templates/CustomTypeTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/CustomTypeTemplate.kt
@@ -1,9 +1,11 @@
 {%- match config %}
 {%- when None %}
 {#- Define the type using typealiases to the builtin #}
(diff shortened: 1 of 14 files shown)


uniffi_bindgen/src/bindings/kotlin/templates/CustomTypeTemplate.kt
```diff
@@ -1,41 +1,42 @@
...
-{#- Create a typealias so that this type can be imported when used as an external type #}
-public typealias {{ name }} = {{ type_name }}
+{%- let ffi_type_name=builtin.ffi_type().borrow()|ffi_type_name %}
+
+{# When the config specifies a different type name, create a typealias for it #}
+{%- match config.type_name %}
+{%- when Some(concrete_type_name) %}
+public typealias {{ name }} = {{ concrete_type_name }}
+{%- else %}
+{%- endmatch %}
```
jhugman, line 16: … and this is the bit we declare

```kt
public typealias JsonObject = JSONObject
```

I can't tell if this is going to be a breaking change or not in all the languages. But: it looks like APIs that we written `JSONObject` now get the less friendly `JsonObject`.

If so, you should probably update the docs, and possibly add a **⚠️ Breaking Change ⚠️** for all relevant languages.

I'd also add a comment to the `typealias` that IDEs can pick up so foreign language developers understand what and why this is.

Aesthetically, I don't really like this, so am hoping we can use the old behaviour, but I'm willing to have a bit of technical debt to keep your momentum going.

So be aware, all who use it, this is a breaking change.

jhugman (commented):
I've added some comments around comments and docs.

uniffi_bindgen/src/bindings/kotlin/templates/CustomTypeTemplate.kt
bendk: Agreed that this is the biggest downside.  If I understand how typealiases work, the consumer code can stay the same, but this could make the generated docs and code completion worse. 

I like the idea of a comment to try to mitigate this.

If we wanted to keep using `JSONObject` in the Kotlin function signatures, I think there are some options.  Maybe the simplest would be adding a `concrete_type_name` field to `Type::Custom`.  It would start blank then the bindings code could set it.

jhugman (approved):
Some doc changes needed but otherwise this is okay for landing.

I've indicated that you should do some testing our understanding of typealiases in multiple langauges. If there are differences we should document these too.

Good job, thanks, Ben.

bendk:
Just updated the comment formatting.

I think the fact the custom type tests still pass, verifies that the [Kotlin](https://github.com/mozilla/uniffi-rs/blob/main/examples/custom-types/tests/bindings/test_custom_types.kts) and [Swift](https://github.com/mozilla/uniffi-rs/blob/main/examples/custom-types/tests/bindings/test_custom_types.swift) code can continue to use the same type names as before.

bendk:
Just made #1204 for the custom type name issue.  I have some ideas on how to address that, let's talk there.

PR #1193, opened 2022-03-28
Added the `uniffi_reexport_scaffolding` macro

bendk:
This can be used to work around [rust-lang#50007](https://github.com/rust-lang/rust/issues/50007), which is getting to be more and more of an issue on the desktop JS project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ members = [
   "fixtures/external-types/crate-two",
   "fixtures/external-types/lib",
 
+  "fixtures/reexport-scaffolding-macro",
   "fixtures/regressions/enum-without-i32-helpers",
   "fixtures/regressions/fully-qualified-types",
   "fixtures/regressions/kotlin-experimental-unsigned-types",
--- a/docs/manual/src/tutorial/Rust_scaffolding.md
+++ b/docs/manual/src/tutorial/Rust_scaffolding.md
@@ -108,3 +108,21 @@ uniffi_build = { path = "path/to/uniffi-rs/uniffi_build, features=["builtin-bind
 Note that `path/to/uniffi-rs` should be the path to the root of the `uniffi`
 source tree - ie, the 2 path specs above point to different sub-directories
 under the `uniffi` root.
+
+### Libraries that depend on UniFFI components
+
+Suppose you want to create a shared library that includes one or more
+components using UniFFI. The typical way to achieve this is to create a new
+crate that depends on the component crates.  However, this can run into
+[rust-lang#50007](https://github.com/rust-lang/rust/issues/50007).  Under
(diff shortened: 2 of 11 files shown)


uniffi_bindgen/src/scaffolding/templates/TopLevelFunctionTemplate.rs
```diff
@@ -14,3 +15,24 @@ pub extern "C" fn {{ func.ffi_func().name() }}(
...
+
+// Create a macro that can be used by a dependent create to forward all of the scaffolding
+// functions defined here.  This is used in 2 ways:
+//
+//   - Building libraries that contain multiple UniFFI components (megazord, libxul).  The combined
+//     library can depend on each component then call uniffi_forward_scaffolding!() for each one.
+//   - Test fixtures.  Each binding generator needs to test against the test fixtures.  To allow
+//     this, they call uniffi_forward_scaffolding!() for the test fixture in their test module.
+#[macro_export]
+macro_rules! uniffi_forward_scaffolding {
```
tarikeshaq, line 28: Hmm I'm thinking about non-high level function forwarding (like forwarding to a method on an object) - how would that work? or would forwarding just the high-level functions be enough for our fixtures?

tarikeshaq (commented):
Thanks Ben!! Had a question about methods, but otherwise, this looks good!  can we add a test to validate that the macro works properly?

We could have one of our fixtures depend on another or maybe make a new crate that depends on one of the existing ones

uniffi_bindgen/src/scaffolding/templates/TopLevelFunctionTemplate.rs
bendk: Great point.  I thought I was getting those too, but I wasn't.  Just pushed a new commit that I think should do this properly.

uniffi_bindgen/src/scaffolding/templates/ForwardUniFFIScaffolding.rs
```diff
@@ -0,0 +1,19 @@
+// Create a macro that can be used by a dependent create to forward all of the scaffolding
```
mhammond, line 1: We might as well be extra redundant and link to the rust-lang issue here too, so the reason for this file isn't a mystery unless you have the additional context.

mhammond (commented):
This LGTM, but I agree with Tarik that a test for all the cases this covers would be ideal.

bendk:
I tried adding unit tests and got a stack overflow when calling the scaffolding functions.  I'm pretty sure I manually tested it, but this whole issue is very strange.

Instead of trying to define new functions, what if we make with the current hack the official way to do this?  I added a helper macro to assist and some unit tests that hopefully will break if the hack stops working.

PR #1201, opened 2022-03-31
Allow external crates to write their own binding generators

tarikeshaq:
These are the first bits needed to implement an external binding generator that lives in an outside crate.

Most of the changes are in the `uniffi_bindgen/src/lib.rs` to define a trait `BindingGenerator` that an external crate can use to take advantage of the logic in `uniffi` to parse the `udl` into a `ComponentInterface`.

This is not a breaking change, and only exposes the functionality (it doesn't change the existing crates)

cc @bendk @mhammond

### DIFF
--- a/uniffi/src/lib.rs
+++ b/uniffi/src/lib.rs
@@ -639,16 +639,6 @@ where
     }
 }
 
-// This function is not intended to be called, but can be used as for the to workaround
-// https://github.com/rust-lang/rust/issues/50007
-//
-// Basically, if another library adds uniffi as an external crate, then builds into a dylib, that
-// library normally won't have the symbols from uniffi.  However, if that library calls a uniffi
-// function, then it *will* have the symbols.  So as a workaround to 50007, we create a dummy
-// function here, and call it from a dummy function in the library,.
-#[no_mangle]
-pub extern "C" fn extern_symbol_hack() {}
-
 #[cfg(test)]
 mod test {
     use super::*;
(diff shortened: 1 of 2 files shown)


uniffi/src/lib.rs
```diff
@@ -639,6 +639,16 @@ where
     }
 }
 
+// This function is not intended to be called, but can be used as for the to workaround
```
mhammond, line 642: nano-nit: "as for" seems wrong in this comment.

mhammond (approved):
lovely, thanks Tarik!

uniffi/src/lib.rs
```diff
@@ -639,6 +639,16 @@ where
...
+// https://github.com/rust-lang/rust/issues/50007
+//
+// Basically, if another library adds uniffi as an external crate, then builds into a dylib, that
+// library normally won't have the symbols from uniffi.  However, if that library calls a uniffi
+// function, then it *will* have the symbols.  So as a workaround to 50007, we create a dummy
+// function here, and call it from a dummy function in the library,.
+#[no_mangle]
+pub extern "C" fn extern_symbol_hack() {}
```
bendk, line 650: I think we should remove this function since #1193 was just merged.

bendk (changes requested):
Looks good to me and it's been used for our desktop JS code for a while so I think we can have decent confidence in the code.

The one issue is the export hack should be removed since it's been implemented in a better way.

bendk (approved):

PR #1202, opened 2022-04-02
Exposes Enum variants from the enum module

tarikeshaq:
When trying to work with enums and variants in an external crate, realized that the variants aren't exposed

Exposing them here so consumers can work with them - over in https://github.com/bendk/gecko-dev/blob/3bb95300c964533827b1e4629a7f8a3320eac376/toolkit/components/uniffi-bindgen-gecko-js/src/render/js.rs we rely on the types themselves being accessible by an external crate

### DIFF
--- a/uniffi_bindgen/src/interface/mod.rs
+++ b/uniffi_bindgen/src/interface/mod.rs
@@ -61,7 +61,7 @@ mod attributes;
 mod callbacks;
 pub use callbacks::CallbackInterface;
 mod enum_;
-pub use enum_::Enum;
+pub use enum_::{Enum, Variant};
 mod error;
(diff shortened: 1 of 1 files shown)


bendk (approved):
Seems like a reasonable change, although I'm not sure this will allow us to implement a name method quite yet.  At least on Kotlin, enum names change depending on if the enum is flat or not.

What about adding an `is_flat` attribute to `Variant`?

mhammond:
Hey @tarikeshaq, do you still want this?

tarikeshaq:
Nope, we lasted this long without it exposed - if folks working on external binding generators need it we can open a new pr

Issue #1190, opened 2022-03-21
Unable to ensure version match between uniffi core and uniffi-bindgen


thunderbiscuit:
I have read the section [Avoiding version mismatches...](https://mozilla.github.io/uniffi-rs/tutorial/Rust_scaffolding.html#avoiding-version-mismatches-between-uniffi-core-and-uniffi-bindgen) in the readme but the exact way to generate the bindings using this "internal" uniffi-bindgen still evades me.

Our `Cargo.toml` file has the appropriate config:
```toml
[build-dependencies]
uniffi_build = { version = "0.16.0", features = ["builtin-bindgen"] }
```

But the docs don't describe how to generate the bindings any differently than before, which for us is:
```shell
uniffi-bindgen generate src/bdk.udl --no-format --out-dir ../jvm/src/main/kotlin --language kotlin
```

I assume there is something missing in my understanding of how to work this, but I can't quite figure out what... 

The reason why I'm trying to get this work work _without_ the cli tool is that I'm trying to automate the build of the native libs and the bindings file inside a Gradle plugin. We were bug chasing for a little while last week when a newcomer to the project could not get his code to work, and we finally realized he was building using the cli tool on 0.17 whereas we're on 0.16.

Any tips appreciated!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-143)
┆friendlyId: UNIFFI-143

mhammond:
This is all a bit messy, but https://github.com/mozilla/application-services/tree/main/tools/embedded-uniffi-bindgen along with https://github.com/mozilla/application-services/blob/0d4ae3876b1f20da2b40c6fe5701863d0b6696dd/publish.gradle#L206-L226 is how Mozilla's main consumer of uniffi manages this.  We'd welcome your feedback on what improvements we can make here (even if it's just to document out flow a bit better?)

thunderbiscuit:
Thanks for the context and sharing your solution! It's funny because we ended up with a similar (even if less refined) solution, where our [ffi layer](https://github.com/bitcoindevkit/bdk-ffi) which feeds into the three libraries we support (Kotlin, Swift, Python) has a small binary that calls the proper version of uniffi-bindgen, effectively creating a drop-in replacement for the cli. This drop-in replacement is what I then use in the Gradle plugin for building the JVM and Android libraries.

[...]

PR #1203, opened 2022-04-04
[Discuss] ADR about version mismatches

bendk:
The work on desktop JS bindings has brought this issue to the forefront, so I started an ADR for it.  I'm not really sure which option is best, I would love to hear input from others on that.

Note: this is not blocking the desktop JS project.  We have a PR up to add support for external bindings, but doesn't change how the current bindings are implemented.  This should be enough to get the desktop JS project rolled out.  Also, as noted in the PR, desktop JS can somewhat ignore the versioning issue, since the bindgen code is going to included in the moz-central repo.

The main rationale for this ADR is:
  - To prepare for more community supported bindings in addition to Ruby
  -  To prevent the current versioning issues that get reported on `#uniffi`

### DIFF
--- a/docs/adr/0008-preventing-version-mismatches.md
+++ b/docs/adr/0008-preventing-version-mismatches.md
@@ -1,0 +1,233 @@
+# How to prevent version mismatches
+
+* Status: proposed
+* Date: 2021-04-01
+
+## External bindings generators
+
+The ADR from [#1150](https://github.com/mozilla/uniffi-rs/pull/1150) proposes moving bindings generators to a separate
+crate.  This means that instead of having a single `uniffi-bindgen` binary to generate bindings, there will be a set of
+`uniffi-bindgen-[language]` binaries, one for each language needed. This ADR assumes that this plan will eventually be
+adopted.
+
+## Context and Problem Statement
+
+UniFFI is made up of several components whose versions need to be kept in sync to work properly:
+
+  - `uniffi` the runtime dependency used by the scaffolding code
+  - `uniffi-bindgen` used to generate the scaffolding code
+  - `uniffi-bindgen-[language]` used to generate the bindings code
+
+If these get out of sync we could have compatibility issues between the generated bindings and scaffolding code.  For
+example, if we updated `uniffi` and modified the layout of `RustBuffer`, but then used scaffolding or bindings code that
+used the old layout, this would result in undefined behavior.
+
+To prevent the undefined behavior, we use the current version string to generate the `ComponentIterface` hash and
+include that hash in the FFI function names. This means if there is a version mismatch, the user will see a linker error
+or a runtime error when trying to open the dynamic library.  These errors are preferable to undefined behavior, however
+they are not very developer friendly.  Also, this introduces the strict requirement that versions must match exactly
+even though changes that would cause the undefined behavior are relatively rare.  See
+[#1190](https://github.com/mozilla/uniffi-rs/issues/1190) for an example of this causing developer confusion and slowing
+down their workflow.
+
+This issue is compounded by the fact that versions are managed in different ways:
+
+  - The `Cargo.toml` file of the UniFFI component
+  - The `uniffi-bindgen` binary used to generate the scaffolding, which is installed in multiple places:
+     - The `uniffi-bindgen` devs install on their system
+     - The `uniffi-bindgen` that's installed in CI
+     - The `uniffi-bindgen` that downstream consumers install on their system
+  - For each binding language, the `uniffi-bindgen-[language]` used to generate the bindings.  Like `uniffi-bindgen`
+    this is also install in multiple places.
+
+Moving to separate bindings crates means the problem of version compatibility becomes more complex since we can't
+realistically synchronize releases between UniFFI and all external bindings crates:
+
+  - Both UniFFI crates and the external bindings crates should able to make a bugfix release without needing to make a
+    coresponding release for all the other crates.
+  - Even releases with breaking changes don't always require corresponding releases in the other crates.  For example,
+    if we could update how `uniffi-bindgen-kotlin` generates the class hierarchy for Kotlin errors, without needing to
+    change `uniffi` or `uniffi-bindgen`.
+
+This means that our current method of using exact string matching for versions is no longer going to work.
+
+Finally, there can only be one `uniffi-bindgen` and `uniffi-bindgen-[language]`, installed system wide.  This makes it
+difficult for a developer to work on crates that depend on different UniFFI versions.
+
+We've built a decent solution for the scaffolding code: [the builtin-bingen
+feature](https://mozilla.github.io/uniffi-rs/tutorial/Rust_scaffolding.html#avoiding-version-mismatches-between-uniffi-core-and-uniffi-bindgen).
+However, there's no similar solution for the bindings code.
+
+application-services has a custom solution for the bindings generation.  It contains the
+[embedded-uniffi-bindgen](https://github.com/mozilla/application-services/tree/main/tools/embedded-uniffi-bindgen) crate
+which contains very
+[simple binary](https://github.com/mozilla/application-services/blob/main/tools/embedded-uniffi-bindgen/src/main.rs)
+that just runs `uniffi-bindgen`. This allows us to control the `uniffi-bindgen` version using the `Cargo.toml` for this
+crate, rather than relying on the system installed version.
+
+### Breaking changes to the UDL syntax
+
+A related issue is breaking changes to the UDL syntax, for example the change the renamed `[Wrapped]` to `[Custom]`.
+These kind of changes can lead to challenges with scaffolding and bindings generation.  For example, users could get
+into a situtation where `uniffi-bindgen` can parse a UDL file, but `uniffi-bindgen-kotlin` can't.
+
+However, this issue seems less important since a) the errors are easier to diagnose and fix and b) we have historically
+only introduced breaking changes like this for "experimental" features like custom/wrapped types.
+
+This ADR will ignore this issue and only focus on compatibility between the generated scaffolding and bindings code.
+
+## Decision Drivers
+
+- UniFFI should have a friendly developer experience.
+  - If possible, developers shouldn't have to worry or know about this issue.
+  - If that's not possible, then it should be easy for them to identify and fix the issue.
+- We need to support external bindings generators and allow those projects to
+  release new versions without coordinating with the UniFFI team.
+
+## Considered Options
+
+### **[Requirement] More flexible versioning**
+
+As mentioned in the context, once we have external bindings generators we can't rely on simply comparing its version
+string to `uniffi`.  A more flexible versioning policy is a requirement for all other options discussed here.
+
+I think we can use semver for this:
+  - The major version is incremented whenever there's a change that would break compatibility between the bindings and
+    scaffolding generation code. Changing the layout of `RustBuffer` or `RustCallStatus` would be a breaking change. Our
+    custom/external types work would not.
+  - For code generators: the minor version is incremented whenever there's a breaking change between the generated code
+    and the consumer code.  Some examples would be: changing Kotlin classes from open to closed and changing the
+    argument names on Swift.
+  - For `uniffi` itself and other crates, I'm not sure what a minor version change should mean.  Maybe we just don't
+    bump these?
+  - The patch version is incremented for non-breaking, bugfix, releases.
+
+This means that `uniffi-bindgen` and all external bindgen tools are compatible with `uniffi` if their major versions
+match.  We would update `ComponentInterface` so it only uses the major version number when calculating it's hash.
+
+Pros:
+  - Seems like a hard requirement if we want external bindings tools
+  - Would significantly reduce the chance of running into a version mismatch.
+  - Makes it easier to detect version mismatches when parsing `Cargo.toml`. Right now we need to determine the exact
+    version that was resolved and stored in `Cargo.lock`.  If we only needed to compare the major versions, then we
+    could use the specification from `Cargo.toml` directly.
+
+Cons
+  - We would have to be more careful with breaking changes. If we realized a change would break the generated code
+    compatibility, then that change might need to wait before releasing it. If we didn't realize a change would break
+    the generated code compatilibity and released the change anyway, then we've introduced undefined behavior in all
+    consumer code.
+  - Doesn't fully solve the issue of version mismatches. We would need to combine this with at one or more options
+    below.
+
+### **[Option 1] generate bindings in `build.rs`**
+
+As mentioned in the context, we handle this issue for scaffolding by using the `builtin-bindgen` feature of
+`uniffi_build`.  If we also generated the bindings code in `build.rs`, we could use a similar system:
+
+- The Rust crate adds a dependency to each bindings generator they want to use (`uniffi-bindgen-kotlin = 1.10.1`,
+  `uniffi-bindgen-swift = 1.4.0`, etc).
+- Each bindings generator defines a `generate_bindings()` function, which generates the bindings and writes them to the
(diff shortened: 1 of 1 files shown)


docs/adr/0007-preventing-version-mismatches.md
```diff
@@ -0,0 +1,269 @@
...
+
+Pros:
+  - Doesn't rely on a system-wide install of `uniffi-bindgen`.
+  - Bindgen tool versions are specified in the `Cargo.toml`, alongside the
+    `uniffi` version.
+
+Cons:
+  - Forces a specific order to the build process.  If you try building the
+    bindings before the Rust component, it would fail since the source files
+    haven't been generated yet (or worse, succeed with stale code).
```
mhammond, line 186: it's worth thinking about this in a future where .udl files didn't exist and macros are used - that world might make this less of a problem and possibly even a feature? eg, assuming foreign bindings need to parse rust code, it might actually be a build time benefit to generating all bindings at the same time from the same parse tree?

[ok, I've now moved from "thinking out loud" to "making things up" :]

docs/adr/0007-preventing-version-mismatches.md
```diff
@@ -0,0 +1,269 @@
...
+  - Hard to make this work with downstream crates.  For the desktop JS example,
+    the app-services components would need to include a call to
+    `uniffi_bindgen_gecko_js.generate_bindings()` in the `build.rs` file which
+    is not ideal.  Also, it would generate Kotlin/Swift bindings, which aren't
+    needed.  I think the solution would be to create a feature gate for each
+    bindings language that we could enable/disable, but this seems overly
+    complex.  Even this solution wouldn't work for some 3rd-party cases, for
+    example a 3rd-party that wants to use their home-grown bindings generator.
```
mhammond, line 194: I'm not convinced features would be bad fit here TBH. I suspect that we are going to end up feature-gating stuff based on the platform anyway (eg, sync code is likely to be different at the very least)

(to be clear, I'm not advocating for this as the accepted solution, but I think this as worded overstates this con)

docs/adr/0007-preventing-version-mismatches.md
```diff
@@ -0,0 +1,269 @@
...
+Pros:
+  - Doesn't rely on a system-wide install of `uniffi-bindgen`.
+  - Bindgen tool versions are specified in the `Cargo.toml`, alongside the
+    `uniffi` version.
+  - Doesn't require building the main library to generate bindings.
+  - Supports in the downstream crate scenario.  If we take the moz-central
+    example from above, it would create it's own bindgen crate that only
+    generated `gecko-js` bindings, and that crate could generate bindings for
+    the app-services UDL files.
+
```
mhammond, line 228: can we formalize the app-services code inside the uniffi repo in some way? We'd need to get creative, but feature gates might mean uniffi helps support all "known" bindings, even for bindings not in our repo, or something hand-wavey.

(I guess I'm thinking that some of the cons below might not necessarily be as bad as they look on paper - eg, #1190 has a similar work-around. OTOH, maybe the burden is small enough once a consumer becomes "sophisticated" enough, but it might be a large hurdle for new uniffi users who are just investigating it and are you to commit)

docs/adr/0007-preventing-version-mismatches.md
```diff
@@ -0,0 +1,269 @@
...
+Cons:
+  - Extra boilerplate to create the bindgen crate.
+  - Some libraries don't use workspaces, we would be forcing them to switch to
+    using workspaces.  Note though, that users wouldn't have to drastically
+    change their project layout.  They could keep their existing crate in the
+    same place and make it a [root package](https://doc.rust-lang.org/cargo/reference/workspaces.html#root-package).
+
+### **[Option 3] Tool to manage UniFFI installed versions**
+
+We create a tool that managed the bindgen binaries ourselves.
```
mhammond, line 238: nano-nit s/managed/manages/

docs/adr/0007-preventing-version-mismatches.md
```diff
@@ -0,0 +1,269 @@
...
+    - Runs `target/uniffi/uniffi_bindgen_language` and passes it all additional args from the command line.
+
+Pros:
+  - Doesn't rely on a system-wide install of `uniffi-bindgen`.
+  - Doesn't require building the main library to generate bindings.
+  - Doesn't require any additional crates.
+
+Cons:
+  - Requires the most amount of work on our part.
+  - Specifying versions in `uniffi.toml` rather than `Cargo.toml` could be confusing.
```
mhammond, line 257: it would need to be in both, at least for `uniffi` itself?

mhammond (commented):
This is awesome Ben, thanks! All of my comments are just nits, or brain-farts I had on first reading, but I think they are worth sharing. Don't feel like you need to action them.

I also agree that this isn't urgent and we should make sure we have all our ducks in a row before commiting to this (but it's certainly worth having an understanding of the tradeoffs). In particular, I think we should commit to #1150 (which would change some of the words here) and probably even that we should "finish" the gecko bindings *and* the first app-services component that is used on both mobile and desktop, because I suspect there will be some devil in the detail we are yet to discover. A final complication is that I suspect moving away from UDL files would impact this too - and while we are a fair way from making any movement in that area, it would be perfectly fine for this ADR to have a final option of "do nothing until we make some movement there" with the primary con there being, roughly, that #1190 would bite more people until we address this?

But all in all, this is fantastic and a great discussion point.

docs/adr/0007-preventing-version-mismatches.md
bendk: Agreed.  I reworded this section.  It's still listed as a con, but it's emphasized less.

docs/adr/0007-preventing-version-mismatches.md
```diff
@@ -0,0 +1,233 @@
...
+  - If possible, developers shouldn't have to worry or know about this issue.
+  - If that's not possible, then it should be easy for them to identify and fix the issue.
+- We need to support external bindings generators and allow those projects to
+  release new versions without coordinating with the UniFFI team.
+
+## Considered Options
+
+### **[Requirement] More flexible versioning**
+
+As mentioned in the context, once we have external bindings generators we can't rely on a simply compare it's version
```
rfk, line 91:
```suggestion
As mentioned in the context, once we have external bindings generators we can't rely on simply comparing its version
```

docs/adr/0007-preventing-version-mismatches.md
```diff
@@ -0,0 +1,233 @@
...
+string to `uniffi`.  A more flexible versioning policy is a requirement for all other options discussed here.
+
+I think we can use semver for this:
+  - The major version is incremented whenever there's a change that would breaks compatibility between the bindings and
```
rfk, line 95:
```suggestion
  - The major version is incremented whenever there's a change that would break compatibility between the bindings and
```

docs/adr/0007-preventing-version-mismatches.md
```diff
@@ -0,0 +1,233 @@
...
+  release new versions without coordinating with the UniFFI team.
+
+## Considered Options
+
+### **[Requirement] More flexible versioning**
+
+As mentioned in the context, once we have external bindings generators we can't rely on a simply compare it's version
+string to `uniffi`.  A more flexible versioning policy is a requirement for all other options discussed here.
+
+I think we can use semver for this:
```
rfk, line 94: Is this referring specifically to the version number of the `uniffi` crate, the bindings-generator crates, etc? Or, is it a separate version number specifically for the details of the FFI interface, shared between all the different crates in the ecosystem?

If the intent is to use the package version number for this, there may be a nontrivial risk of "false positive" version bumps, where we e.g. rev the major version number of `uniffi` because of some consumer-facing API change in the tooling or integration, causing incompatibility with bindings generators despite no actual change in the FFI layer.

docs/adr/0007-preventing-version-mismatches.md
```diff
@@ -0,0 +1,233 @@
...
+  - Would significantly reduce the chance of running into a version mismatch.
+  - Makes it easier to detect version mismatches when parsing `Cargo.toml`. Right now we need to determine the exact
+    version that was resolved and stored in `Cargo.lock`.  If we only needed to compare the major versions, then we
+    could use the specification from `Cargo.toml` directly.
+
+Cons
+  - We would have to be more careful with breaking changes. If we realized a change would break the generated code
+    compatibility, then that change might need to wait before releasing it. If we didn't realize a change would break
+    the generated code compatilibity and released the change anyway, then we've introduced undefined behavior in all
+    consumer code.
```
rfk, line 119: As part of implementing this change, we would remove the version number from the calculation of the `ComponentInterface` fingerprint, instead depending on orchestration of the build process to detect and prevent version mismatches?

docs/adr/0007-preventing-version-mismatches.md
```diff
@@ -0,0 +1,233 @@
...
+Support for each languages could be behind feature gates, as in `[1]`.  However, libraries could also choose to define a
+preferred set of languages and require that consumers create their own bindgen crates if they want a different set. For
+example, application-services might have a bindgen crate that supports Kotlin and Swift and moz-central could consume
+the application-services components, but use a bindgen crate that supports JS.
+
+Pros:
+  - Doesn't rely on a system-wide install of `uniffi-bindgen`.
+  - Bindgen tool versions are specified in a `Cargo.toml`.
+  - Doesn't require building the main library to generate bindings.
+  - Allows libraries consumers to define their own bindgen crates.
```
rfk, line 189:
```suggestion
  - Allows library consumers to define their own bindgen crates.
```

docs/adr/0007-preventing-version-mismatches.md
```diff
@@ -0,0 +1,233 @@
...
+
+### **[Requirement] More flexible versioning**
+
+As mentioned in the context, once we have external bindings generators we can't rely on a simply compare it's version
+string to `uniffi`.  A more flexible versioning policy is a requirement for all other options discussed here.
+
+I think we can use semver for this:
+  - The major version is incremented whenever there's a change that would breaks compatibility between the bindings and
+    scaffolding generation code. Changing the layout of `RustBuffer` or `RustCallStatus` would be a breaking change. Our
+    custom/external types work would not.
```
rfk, line 97: To make this workable, I think we'd need to generate an explicit list of the things that can cause a breaking change (Probably we'd fail to enumerate all of them on a first attempt, but it's worth attempting anyway).

docs/adr/0007-preventing-version-mismatches.md
```diff
@@ -0,0 +1,233 @@
...
+  - Doesn't rely on a system-wide install of `uniffi-bindgen`.
+  - Doesn't require building the main library to generate bindings.
+  - Doesn't require any additional crates.
+
+Cons:
+  - Requires the most amount of work on our part.
+  - Specifying versions in `uniffi.toml` as well as `Cargo.toml` could be confusing.
+  - If a library consumer wanted to generate bindings for a language that wasn't specified in `uniffi.toml`, then they
+    would need to manage the installed binary themselves.
+
```
rfk, line 223: Trying to think of other ways out of this bind...I wonder what it would look like to declare a uniffi version number in the `.udl` file itself. Syntax would be fiddly as always, but I'm imagining something  like:

```
[UniFFI=1.1]
namespace my_example {
   ...
}
```

This would be clunky and annoying, but it's putting the information in the one place that every component of the system has to look. If you try to generate the scaffolding with an incompatible version of UniFFI, it can detect this and generate a clean helpful error. If you try to generate bindings with an incompatible version of the tool, it can detect this and generate a clean helpful error.

In terms of the goals of the ADR, this would punt entirely on "If possible, developers shouldn't have to worry or know about this issue" because in fact they would have to know about it *more*. But it would let us to a much better job of "If that's not possible, then it should be easy for them to identify and fix the issue". In fact the completely custom error message could direct the user to a page in the manual with tips about how to keep the versions of their tools in sync.

You could treat this version number as the version of UniFFI itself, or as a separate and less-frequently-updated version number that is specifically *the version of UDL* rather than of any particular tool.

rfk (commented):
I'm excited to see a fresh attempt at resolving this pain point! Chipping in with a few notes and suggestions below.

docs/adr/0007-preventing-version-mismatches.md
mhammond: Hey Ryan! I think the intent here was to use the crate versions, with the implication being that uniffi and the bindings would need to stay in a kind of lock-step regarding the major version number. You are correct that "false positives" would be likely in this scenario - I wonder if we can come up with examples of this, because I could imagine that many tooling or integration changes would indeed impact the bindings generators in some way even if they didn't change the ffi layer. Obviously there are cases where this wouldn't be true too.

A separate version for the ffi layer would be good, but then we couldn't rely on cargo - OTOH though, maybe that's not a big problem in practice - the fact that a human might need to determine the exact version of the "foo" bindings which interoperate with version x.y.z of uniffi might be ok.

docs/adr/0007-preventing-version-mismatches.md
mhammond: Assuming the above, where the major version of uniffi and the binding must match, the intent would be to only use that major version for this.

docs/adr/0007-preventing-version-mismatches.md
mhammond: This is an interesting idea, particularly if we can decouple the crate version from the "ffi version". Ben and I were also chatting about how the existing mechanism seems to have saved us approximately zero times, mainly because of the way app-services ensures the versions match via the intermediate tool - so we were trying to think of a way to kill this entirely in a "safe enough" way - but haven't quite come up with anything yet :)

docs/adr/0007-preventing-version-mismatches.md
mhammond: That's a very good point, and I wonder if identifying this might help to get to the crux of what we are trying to solve here.

docs/adr/0007-preventing-version-mismatches.md
bendk: If we decouple the crate version from the FFI version, then something like this seems like a great idea. It simplifies the task of explaining to devs which versions of uniffi are compatible with which versions of the bindgen tools.  I also think the concept of a UDL syntax version would have been nice as we were working through the wrapped/custom types code.  In hindsight it seems like we didn't have a great language to discuss the changes we were making.

Another way to handle this would be to have `uniffi_build` generate some sort of intermediate artifact that includes the version.  In addition to writing `[component].uniffi.rs` to the out dir, it would also serialize the `ComponentInterface` and write it out to the target dir.  The bindgen tool would input that, check the version, then write out its bindings.

I'm not sure if this is better or worse.  Users wouldn't specify the version in the UDL file.  That makes it easier to get started,  but having that version implicit might make compatibility issues harder to understand.

One reason I'm thinking about this is that Mark brought up replacing the UDL file with a macro.  In that world, we might want to do the serialization anyways to avoid having to parse the Rust source tree for each bindings language.

rfk: > Another way to handle this would be to have uniffi_build generate some sort of intermediate artifact that includes
> the version. 

I like to use WebAssembly Interface Types and wasm-bindgen as a touch-point here. In that case, compiling the Rust code produces both the compiled code, and a binary "interface definition" which can then be processed by downstream tools. It seems like a really nice workable model, and you can imagine the UDL as a kinda of temporary hand-written version of that interface definition.

(In the wasm-bindgen case, the definition is actually generated as an additional custom section of the compiled library, not as a separate artifact, which I think is really neat and makes it less likely for build system shenanigans to accidentally use a mis-matched `.so` and interface face...but may be a bit much mucking around for our purposes here).

bendk:
We now have this doc: https://github.com/mozilla/uniffi-rs/blob/main/docs/uniffi-versioning.md and this PR is very old.  Let's close it.

PR #1205, opened 2022-04-04
[Don't merge] Seperate bindings

bendk:
This PR is my attempt to separate the current bindings generators into their own crates (but still in the UniFFI repo).

Let's not merge this one until we figure out the issue of versioning mismatches between the bindings and scaffolding code (#1203).  Having separate bindings crates will probably just make that worse.

The main reason I'm sharing this code is that I want to merge the very first commit which adds a testing framework for this.  That's going to be very useful for the desktop JS bindings.  The other commits can hopefully serve as a example of how that framework would be used.

For each language, I:
- Moved code from `uniffi_bindgen/src/bindings/[lang]/` to `uniffi_bindgen_[lang]/src/`
- Added `uniffi_bindgen_[lang]/src/lib.rs` that creates the CLI by leveraging the `BindingGenerator` trait
- Refactored the Config-related code in `uniffi_bindgen_[lang]/src/gen_[lang]/mod.rs`. Removed the old code and added a  `BindingGeneratorConfig` impl.
- Extracted the test-related code from `uniffi_bindgen_[lang]/src/mod.rs` and put it into `uniffi_bindgen_[lang]/tests/fixtures.rs`, then deleted the mod.rs file


bendk:
This one can't land until we figure out the versioning issue (#1203).  I don't think that's going to be any time soon, so let's close this one.

PR #1206, opened 2022-04-04
Adding support for external fixture/example tests

bendk:
Added code to allow external bindings crates to find the UDL files and cdylibs for examples/fixtures.  Made a new crate for this, since I didn't want to pull in `serde` as a dependency for `uniffi`. Eventually, this can replace `uniffi/testing.rs` and we can remove some other dependencies from `uniffi` like `cargo_metadata`.

This commit adds a testing framework that works with external bindings, but we're not actually using it yet.  The plan is to use is for the desktop JS bindings, whose code will live in moz-central.

For an example of how this code would be use, take a look at #1205.  It moves the existing bindings to separate crates and switches them to using this testing framework.

### DIFF
--- a/fixtures/ext-types/uniffi-one/Cargo.toml
+++ b/fixtures/ext-types/uniffi-one/Cargo.toml
@@ -1,6 +1,5 @@
 [package]
-# TODO: modify the crate name to test non-default names.
-name = "uniffi-one"
+name = "uniffi-fixture-ext-types-lib-one"
 edition = "2018"
 version = "0.17.0"
 authors = ["Firefox Sync Team <sync-team@mozilla.com>"]
@@ -8,7 +7,8 @@ license = "MPL-2.0"
 publish = false
 
 [lib]
-crate-type = ["cdylib", "lib"]
+crate-type = ["lib", "cdylib"]
+name = "uniffi_one"
 
 [dependencies]
 anyhow = "1"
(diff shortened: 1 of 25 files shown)


fixtures/ext-types/uniffi-one/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 # TODO: modify the crate name to test non-default names.
```
mhammond, line 2: I'd be inclined to delete this comment while you are here - I've no idea what it means, and I suspect I added it :)

uniffi_testing/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
```
mhammond, line 1: I think this crate should have a README.md

mhammond (approved):
This LGTM and I think I understand the justification, but it only makes sense in a world where #1150 is approved, right?

bendk:
> This LGTM and I think I understand the justification, but it only makes sense in a world where #1150 is approved, right?

Since #1201 was merged, I figured that we were going forward with this, at least in the limited sense of supporting external bindings.

I'll try to update this based on the other comments.

tarikeshaq:
> but it only makes sense in a world where https://github.com/mozilla/uniffi-rs/pull/1150 is approved, right?

Yeah I think the scope of #1150 should be smaller and landed at this point, I'll cut it down so it only represents the work in #1201 and this PR - keeping the rest of the bindings and details around version management to separate ADRs

PR #1209, opened 2022-04-07
Adds missing licenses and authors

tarikeshaq:
cc @bendk 

Firefox desktop won't let us vendor crates without licenses (I think crates.io will also complain), so I added them wherever I saw them missing


bendk (approved):
:rocket:

PR #1210, opened 2022-04-08
Have the manual link back to the repo

mhammond:
I was looking at some docs from Ben, which had a link to https://mozilla.github.io/uniffi-rs/, and I noticed that page didn't actually link to uniffi itself - this PR fixes that.

### DIFF
--- a/docs/manual/src/Overview.md
+++ b/docs/manual/src/Overview.md
@@ -1,6 +1,7 @@
 # UniFFI
 
-UniFFI is a tool that automatically generates foreign-language bindings targeting Rust libraries.  
+UniFFI is a tool that automatically generates foreign-language bindings targeting Rust libraries.
+The repository can be found on [github](https://github.com/mozilla/uniffi-rs/).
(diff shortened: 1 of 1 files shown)


docs/manual/src/Overview.md
```diff
@@ -1,6 +1,7 @@
 # UniFFI
 
-UniFFI is a tool that automatically generates foreign-language bindings targeting Rust libraries.  
+UniFFI is a tool that automatically generates foreign-language bindings targeting Rust libraries.
+It can be found on [github](https://github.com/mozilla/uniffi-rs/)
```
skhamis, line 4: super nit: Maybe a comma after this statement or have it run with the previous line since In the markdown preview it reads "It can be found on github It fits in..." vs "It can be found on github, It fits..." conversely could say "The repo can be found on github".

skhamis (approved):
Super nit based on what I saw in the preview but otherwise nice add!

docs/manual/src/Overview.md
mhammond: oops, nice catch, thanks.

PR #1211, opened 2022-04-08
Update to clap v3 using derive macros

mhammond:
This upgrades `clap` to version 3.0 via "derive" macros. It seems a bit cleaner to me, and it also seems less error prone, as it highlighted the fact that a `--manifest-path` option was specified for the `generate` command but it was unused.

It enabled more features of `clap` and no longer specifies `default-features = false` - so we get color options etc. I can't see a good reason to avoid that for this use-case, but I might be missing the reason it was done this way in the first place.

I *think* it's complete, but have a local test failure I can't explain, so we'll see what CI says. There's also another PR which sticks with the `builder` interface which I'll put up for comparison, but I'm leaning towards this being the better option.

### DIFF
--- a/uniffi_bindgen/Cargo.toml
+++ b/uniffi_bindgen/Cargo.toml
@@ -20,7 +20,7 @@ weedle = "0.12"
 anyhow = "1"
 askama = { version = "0.11", default-features = false, features = ["config"] }
 heck = "0.3"
-clap = { version = "2", default-features = false }
+clap = { version = "3", features = ["cargo", "std", "derive"] }
 paste = "1.0"
 serde = "1"
 toml = "0.5"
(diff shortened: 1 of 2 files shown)


mhammond:
See #1212 for a WIP using the builder API, but as mentioned, I think this is cleaner. YMMV, so all opinions welcome!

skhamis:
wow! I really like this, I think it makes it much easier to understand/maintain the command options. I think I could also try updating https://github.com/mozilla/uniffi-rs/pull/1002 using this and seeing how that fares but imo it felt faster to parse and understand what each command was trying to do + options.

bendk (approved):
Looks great.  The upgrade to clap v3 will be nice for the desktop work.

PR #1212, opened 2022-04-08
WIP moving to clap version 3 using the builder API.

mhammond:
This is not complete - it will have different semantics and panics in
various cases.

I initially thought that sticking with the builder interface would be
easier to review, but I'm not that happy with how it is working out,
particularly the "magic" `arg!` macro, and I didn't find good docs
on exactly how to specify things.

Then I realized that clap's "derive" mode actually *is* structopt
(see clap's FAQ) and I've been relatively happy with structopt in
app-services - so thought that before I polish this I'd have a go
at using derive mode (see #1211) there and see how that looks, then we can
decide which one to move forward with.


mhammond:
I agree with @skhamis that #1211 is significantly better!

PR #1213, opened 2022-04-12
Do not provide library link search path

badboy:
Somehow this now breaks for me on swiftlang-5.6.0.323.62 clang-1316.0.20.8 for the callback example tests.


travis79 (approved):
Weird...  looks like we don't need this I guess :)

badboy:
So I looked into this once more.
[My own demo project](https://github.com/badboy/rocketscience/pull/2) works fine with and without that fix, running on CI with Xcode 13.2.1

Locally it fails tests when running on `main`, and succeeds with this PR.
Xcode 13.3.1

PR #1215, opened 2022-04-13
Upgrade to Rust 1.60

mhammond:
mozilla-central is moving to this, so I figured it can't hurt to see what the fallout is here - ~~apparently there is none!~~ but of course there is some.

### DIFF
--- a/fixtures/callbacks/src/lib.rs
+++ b/fixtures/callbacks/src/lib.rs
@@ -52,6 +52,7 @@ impl Default for RustGetters {
 
 // Use `Send+Send` because we want to store the callback in an exposed
 // `Send+Sync` object.
+#[allow(clippy::wrong_self_convention)]
 trait StoredForeignStringifier: Send + Sync + std::fmt::Debug {
     fn from_simple_type(&self, value: i32) -> String;
     fn from_complex_type(&self, values: Option<Vec<Option<f64>>>) -> String;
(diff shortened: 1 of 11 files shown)


mhammond:
> it can't hurt to see what the fallout is here - apparently there is none!

doh - I forgot to run clippy :)

uniffi/src/ffi/rustbuffer.rs
```diff
@@ -365,7 +365,11 @@ mod test {
         // We don't want to actually materialize a huge vec, so unsafety it is!
         // This won't cause problems because the contained items are Plain Old Data.
```
jplatte, line 366: Have these tests ever been run through miri? I'm somewhat certain undef is not okay even for Plain Old Data, and you have to go through `MaybeUninit` instead.

bendk: I don't think this should block the PR, but `MaybeUninit` does seem better to me.  I'll make a separate PR for that.

bendk (approved):
LGTM.  I like how each time we upgrade Rust, we get to clean up the code.

uniffi/src/ffi/rustbuffer.rs
mhammond: Ben ended up killing these tests in #1216

jplatte: Yeah I noticed 🙂

PR #1222, opened 2022-04-22
Assert that record keys satisfy HashMap requirements (Eq & Hash)

### LINKED ISSUE #1221: Improve error message when not-supported map key type is used in UDL
Followup to #1129, specifically https://github.com/mozilla/uniffi-rs/pull/1129#discussion_r853735968

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-157)
┆friendlyId: UNIFFI-157

badboy:
This removes the explicit type check in code,
moving the assertion to the generated scaffolding code,
in order to improve the error message.

Fixes #1221

---

I'm not perfectly happy with this solution, but it's the best I could do for now.

### DIFF
--- a/uniffi_bindgen/src/scaffolding/templates/scaffolding_template.rs
+++ b/uniffi_bindgen/src/scaffolding/templates/scaffolding_template.rs
@@ -8,6 +8,15 @@
 // seems to show that single line as context for the user.
 uniffi::assert_compatible_version!("{{ uniffi_version }}"); // Please check that you depend on version {{ uniffi_version }} of the `uniffi` crate.
 
+{% for ty in ci.iter_types() %}
+{%- match ty %}
+{%- when Type::Map with (k, v) -%}
+{# Next comment MUST be after the line to be in the compiler output #}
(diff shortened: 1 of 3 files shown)


uniffi_bindgen/src/scaffolding/templates/scaffolding_template.rs
```diff
@@ -8,6 +8,15 @@
...
 
+{% for ty in ci.iter_types() %}
+{%- match ty %}
+{%- when Type::Map with (k, v) -%}
+{# Next comment MUST be after the line to be in the compiler output #}
+uniffi::deps::static_assertions::assert_impl_all!({{ k|type_rs }}: ::std::cmp::Eq, ::std::hash::Hash); // record<{{ k|type_rs }}, {{ v|type_rs }}>
+{%- else %}
+{%- endmatch %}
+{% endfor %}
+
```
badboy, line 19: We could move this assert (and maybe the one above) to its own template if we wanted.

bendk (approved):
Looks great, thanks for closing the loop here.

PR #1224, opened 2022-04-22
Upgrading heck to v0.4

bendk:
This will help the desktop JS project, since it will reduce the number of duplicate rust crates.


tarikeshaq (approved):
Looks great! :shipit:

PR #1217, opened 2022-04-20
Added some documentation around how we generate foreign bindings

bendk:
As part of the #1189 review, @jhugman asked me to write up a doc on the template code.  Here's my attempt at that.

PR #1226, opened 2022-04-25
Swift python template dispatch

### LINKED ISSUE #1225: Defaults can cause `NameError` in Python
Here's an example from rondpoint:
```
class Optionneur(object):
[...]
    def sinon_enum(self, value = Enumeration.TROIS):
       [...]
```

The use of `Enumeration.TROIS` as a default value is problematic for a couple reasons:
  - It requires `Enumeration.TROIS` to be defined above `Optionneur` in the file.  This means that rondpoint is only currently working because we got lucky in how type definitions are rendered.  We might be able to fix this by doing a dependency analysis based on which types had defaults from other types.  But that still would fail if there were circular dependencies.
  - The default value is always the same object.  If this object was mutable, like if it was `Type::Object`, then you could run into a common python gotcha (https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

Instead, the idomatic python solution is to use a singleton value to indicate a missing argument:

```
DEFAULT = object()
[...]
class Optionneur(object):
[...]
    def sinon_enum(self, value = DEFAULT):
       if value is DEFAULT:
           value = Enumeration.TROIS
       [...]
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-159)
┆friendlyId: UNIFFI-159
┆sprintEndDate: 2022-05-13

bendk:
In #1189 I updated how we handle rendering type definitions and helper code for Kotlin.  This PR:
  - Adds documentation around this (From #1217)
  - Also updates Python and Swift to use the new system
  - Fixes #1225 along the way

Most of this was very mechanical and I just followed the same steps as for Kotlin.  [e5e4a4f](https://github.com/mozilla/uniffi-rs/pull/1226/commits/e5e4a4f4c5fa64200e9bf6d1a8036be8f8801634) was the only functional change here.  I added it because I hit #1225 since the type definitions got reordered.


mhammond (approved):
A bit of a rubber stamp, but it LGTM

jhugman (approved):
I don't think you need my approval on this one.

PR #1228, opened 2022-04-28
Indicate that weedle2 is fully located in the uniffi repo

badboy:


mhammond (commented):
Why the author's name removal?

badboy:
> Why the author's name removal?

Cargo doesn't _require_ it anymore, but you are right that we should keep that.

mhammond (approved):
👍

PR #1229, opened 2022-04-28
Python: Force correct order of evaluation when building the dict

badboy:
In Python 3.7 and before the evaluation order in a dict comprehension
was not according to spec.

The spec says it's left to right: https://docs.python.org/3/reference/expressions.html#expression-lists
But the coveralls test failed with a "read past end of rust buffer"
uniffi exception, because its trying to read the value first.

This seems to have been a known bug back in the day: https://bugs.python.org/issue11205
For that specific instance of a dict literal it was fixed in 3.5
But for a dict comprehension it seems to not have been fixed in 3.7 yet

This can be tested with this short Python snippet:

    def f(i):
        print(i)
        return i

    { f(i):f(10+i) for i in range(2) }

This should print "0 10 1 11" (as it does on Python 3.8.12),
but prints "10 0 11 1" on Python 3.7.12.


bendk (approved):
Good catch on this one.

PR #1235, opened 2022-05-10
Refactor Path(Buf) usage

jplatte:


jplatte:
This simplifies the code by only using `AsRef` generics where functions are actually being called from different places, not converting the `camino::Utf8Path`s received from `cargo_metadata` to Strings (because they're actually very nice to work with directly!) and cleaning up some error handling.

bendk (approved):
Nice set of cleanups/refactors!

jplatte:
So it's approved, what else needs to happen before merge? 😄

bendk:
Oops, meant to merge this yesterday.

PR #1238, opened 2022-05-17
Further improve fs / path code

jplatte:
Basically a follow-up to #1235.

The main commits are the final two: the first one introduces `camino`s `Utf8Path(Buf)` types in more places and the second one replaces `std::fs` by the `fs_err` crate, which improves error messages by attaching the file or directory path upon which an operation was attempted.

The first three commits are just drive-by improvements that should be obvious 🙂


mhammond (approved):
These look like a nice improvement, thanks!

PR #1247, opened 2022-05-23
Do not bundle the WebIDL tests

badboy:
They are large and not needed in consumers, such as mozilla-central,
which bundles all dependencies.


bendk (approved):
Makes sense to me.

PR #1249, opened 2022-05-24
Remove crate-lib = "lib" from some fixtures that don't need it

jplatte:
This might improve test compilation times. At worst, it's only a small simplification of a few manifests.


mhammond (approved):
I think there are more crates we can do this for (particularly some examples), and I suspect one or 2 of the "external types" fixtures might even create cdylib's when they don't need it, but there's no need to let the perfect be the enemy of the good, so 👍

PR #1251, opened 2022-05-25
Simplify ComponentInterface content iteration

jplatte:
Another set of code simplifications, if any of the changes require extra justification, please let me know!

### DIFF
--- a/uniffi_bindgen/src/interface/mod.rs
+++ b/uniffi_bindgen/src/interface/mod.rs
@@ -243,10 +243,7 @@ impl ComponentInterface {
     /// This method uses `IterTypes::iter_types` to iterate over the types contained within the
     /// given item, but additionally recurses into the definition of user-defined types like records
     /// and enums to yield the types that *they* contain.
-    fn iter_types_in_item<'a, T: IterTypes>(
-        &'a self,
-        item: &'a T,
-    ) -> impl Iterator<Item = &'a Type> + 'a {
+    fn iter_types_in_item<'a>(&'a self, item: &'a Type) -> impl Iterator<Item = &'a Type> + 'a {
         RecursiveTypeIterator::new(self, item)
     }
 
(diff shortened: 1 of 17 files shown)


uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -242,73 +258,17 @@ impl TypeUniverse {
...
 }
 
 /// An abstract type for an iterator over &Type references.
 ///
 /// Ideally we would not need to name this type explicitly, and could just
-/// use an `impl Iterator<Item=&Type>` on any method that yields types.
-/// Unfortunately existential types are not currently supported in trait method
-/// signatures, so for now we hide the concrete type behind a box.
+/// use an `impl Iterator<Item = &Type>` on any method that yields types.
 pub type TypeIterator<'a> = Box<dyn Iterator<Item = &'a Type> + 'a>;
```
jplatte, line 270: I kept this type for most places it was previously used because there are a few places where different functions returning it are returned from `match` arms and since the underlying iterators aren't the same, `impl Iterator` wouldn't actually work.

I kept most of the comment on the type though because ideally with something like `enum impl Trait` (or just further refactoring of `uniffi_bindgen`), this type could still be removed eventually.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -140,9 +141,9 @@ impl ComponentInterface {
         self.namespace.as_str()
     }
 
-    /// List the definitions for every Enum type in the interface.
-    pub fn iter_enum_definitions(&self) -> Vec<Enum> {
```
mhammond, line 144: heh - I was going to say I didn't hate the existing `iter_` prefix on these function names - until I got here and noticed they weren't iters at all :)

mhammond (commented):
I like this! I can see how this trait isn't strictly necessary now, but because this trait was added in https://github.com/mozilla/uniffi-rs/pull/1005 and @bendk reviewed that, I think I'll leave the formal review to him.

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -124,6 +124,15 @@ impl Type {
...
     }
+
+    pub fn iter_types(&self) -> TypeIterator<'_> {
+        let nested_types = match self {
+            Type::Optional(t) | Type::Sequence(t) => t.iter_types(),
+            Type::Map(k, v) => Box::new(k.iter_types().chain(v.iter_types())),
+            _ => Box::new(iter::empty()),
+        };
+        Box::new(std::iter::once(self).chain(nested_types))
+    }
```
bendk, line 135: This isn't new, but I'll just note that I find it a bit confusing how this function is split off from the `RecursiveTypeIterator` code and it works differently.  For example, `RecursiveTypeIterator` does work to prevent infinite loops, but this doesn't.  I think the code is correct, because `Optional`, `Sequence`, and `Map` can't have circular references by their nature, but it took me a second to realize that.  This is part of a bigger issue and not related to these changes.  I made #1254 to work on this further.

uniffi_bindgen/src/interface/mod.rs
bendk: Now that we have a new Askama that doesn't block us we could keep the `iter_` part of the method names and return iterators.  But changing these to return slices seems totally reasonable to me as well.

PR #1252, opened 2022-05-25
[Discuss] Prototype of how async could work

bendk:
Here's some code I put together showing how I think async could work for the Nimbus use case: Rust async code that awaits on async callback interfaces methods.  There's a bunch of extra code in this PR, but the idea is we have an example with:

- [A Python callback interface with async methods](https://github.com/bendk/uniffi-rs/blob/96e60ac52bd93a99abed4743e2220efff2fe0dac/examples/async-calc/tests/bindings/test_async_calc.py#L27-L34)
- [A Rust function that awaits those methods](https://github.com/bendk/uniffi-rs/blob/96e60ac52bd93a99abed4743e2220efff2fe0dac/examples/async-calc/src/lib.rs#L61-L65)
- [A Python function that awaits the Rust function](https://github.com/bendk/uniffi-rs/blob/96e60ac52bd93a99abed4743e2220efff2fe0dac/examples/async-calc/tests/bindings/test_async_calc.py#L36-L37)

To make this work, we convert these async functions to a sync function that inputs a Promise as it's first argument then add a bunch of glue/wrapper code.

There's also a [simple async executor](https://github.com/bendk/uniffi-rs/blob/96e60ac52bd93a99abed4743e2220efff2fe0dac/examples/async-calc/src/tasks.rs) that handles this code.  When we a promise is resolved, we call `poll_tasks()` which advances any futures that were awaiting on that promise.  This gives us the nice feature that futures always run in the same thread where the promise was resolved.  This will fail if you try to await on a network request or something, but that's a different use case.

My takeways:
  - This provides motivation for implementing async:
     - It shows it's possible
     - If we made this a UniFFI feature, then the glue code could be auto-generated and it could be a nice developer experience
  - It also provides motivation for not implementing it yet:
     - It shows it's possible to do async with the current UniFFI
     - The amount of glue code scales with the number of async functions and also the number of async return types.  AFAICT, the Nimbus use case is very simple, maybe only a single async function.  It might be reasonable to implement that by hand like I did for this example crate.
   - This requires language support to convert between promise/callback style code and async functions.  Python's Future class works well and I was able to cobble something together on Rust, but I'm not sure all languages have something like this.
   - Async will require some extra dependencies.  I needed to pull in `futures` and `pinky_swear`.  Maybe we could do without the `pinky_swear` dependency, but it would be really hard to do this without `futures`.  If we do async, I think it should be behind a feature gate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
   "uniffi",
   "weedle2",
   "examples/arithmetic",
+  "examples/async-calc",
   "examples/callbacks",
   "examples/geometry",
   "examples/rondpoint",
--- a/examples/async-calc/Cargo.toml
+++ b/examples/async-calc/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "async-calc"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+crate-type = ["lib", "cdylib"]
+name = "uniffi_asynccalc"
+
+[dependencies]
+futures = "0.3.21"
+pinky-swear = "6.1.0"
+uniffi_macros = {path = "../../uniffi_macros"}
+uniffi = {path = "../../uniffi", features=["builtin-bindgen"]}
+
+[build-dependencies]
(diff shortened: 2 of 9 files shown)


bendk:
I think we have an async plan together, let's close this one.

Issue #1242, opened 2022-05-19
MSRV policy


jplatte:
Hi! I'm curious what your thoughts are on what Rust version UniFFI should minimally support. I found that it currently contains a [hack](https://github.com/mozilla/uniffi-rs/blob/52c38d3eecf88a83b46d624066fd6f79288b1375/uniffi_bindgen/src/scaffolding/templates/ReexportUniFFIScaffolding.rs) for a [bug](https://github.com/rust-lang/rust/issues/50007) in stable Rust (include 1.61.0 released today) that should be fixed in 1.62.0 ([PR](https://github.com/rust-lang/rust/pull/95604)).

I would like to remove this hack ahead of posting a PR for a macro frontend (see recent discussion in #385 for plans around that). Is it acceptable to increase the MSRV to 1.62.0 (beta) for that now based on the assumption that there isn't going to be a crates.io release in the next 6 weeks (given the last release is only 2w old)? Or if not now, would it be okay to bump once 1.62.0 is stable?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-164)
┆friendlyId: UNIFFI-164

mhammond:
Our policy is at https://github.com/mozilla/uniffi-rs/blob/main/docs/policies/rust-versions.md, which implies 1.59.0 is the min version we want to support currently.

jplatte:
Given the current `rust-toolchain.toml` contains 1.60 (https://github.com/mozilla/uniffi-rs/blob/main/rust-toolchain.toml), that seems unlikely 😄

Right now it seems to me like it should take at most a few months until MSRV can be 1.62, which is probably fine for upstreaming my changes (at least from my / Element's perspective).

mhammond:
> Given the current `rust-toolchain.toml` contains 1.60 (https://github.com/mozilla/uniffi-rs/blob/main/rust-toolchain.toml), that seems unlikely 😄

Right - that's the "uses" version, which m-c is also currently on. The m-c "requires" version is 1.59 - often there's a larger gap but it's small now. So I'm not quite sure what's "unlikely" - is there something we need to clarify in the docs?

(Regardless, given the policy exists, I'm going to close this, but happy to keep engaging on this)

jplatte:
Yeah I saw this big table with different versions in the moz documentation and understood none of it. What's "requires" vs "uses"?

mhammond:
I'll update the docs to make that clearer, but:
[...]

PR #1253, opened 2022-05-25
Clarify our Rust version policy

mhammond:
As noted in #1242, our Rust policy wasn't as clear as it could be, so I attempted to fix it. I also changed the language about uniffi not currently being in mozilla-central - while that's true today, it look like landing very soon, so I tried to future-proof that. @jplatte, I can't add you as a reviewer, but would like your feedback anyway.

### DIFF
--- a/docs/policies/rust-versions.md
+++ b/docs/policies/rust-versions.md
@@ -7,19 +7,23 @@ which aren't always able to be on that latest version.
 
 Given UniFFI is currently developed and maintained by Mozilla staff, it should
 be no surprise that an important consideration is mozilla-central (aka, the
-main Firefox repository). While at time of writing UniFFI is not used by
-mozilla-central, this policy is unashamedly focused on ensuring it will be
-able to.
+main Firefox repository). This policy exists to ensure UniFFI can always be
+used by mozilla-central.
 
 ## Mozilla-central Rust policies.
 
-It should also come as no surprise that the Rust policy for mozilla-central
-is somewhat flexible. There is an official [Rust Update Policy Document
-](https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html])
-but everything in the future is documented as "estimated".
+The Rust policy for mozilla-central has an official
+[Rust Update Policy Document
+](https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html),
+the tl;dr of which is:
 
-Ultimately though, that page defines 2 Rust versions - "Uses" and "Requires",
-and our policy revolves around these.
+* There's a "Uses" version which all of Mozilla's CI uses for releasing Firefox.
+  Discover it in the most recent updates to [this meta bug on Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1504858)
+
+* There's a "Requires" version, which Mozilla is committed to supporting,
+  primarily for downstream builders such as Linux distributions.
+  [Discover it here](https://searchfox.org/mozilla-central/search?q=MINIMUM_RUST_VERSION&path=python/mozboot/mozboot/util.py). This version is tested in mozilla-central's CI,
+  but nothing is shipped by mozilla with it.
 
 # UniFFI Rust version policy
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ commands:
   # Our minimum supported rust version is specified here.
   prepare-rust-min-version:
     steps:
-      - run: rustup override set 1.57.0
+      - run: rustup override set 1.59.0
       - run: rustup update
   build-api-docs:
     steps:


mhammond:
Fixes #1242

jplatte:
I guess I just find this dual version usage thoroughly confusing in itself. Why would you (only) use a version in CI that is newer than what you must support in the end? That seems like a clear path towards accidentally relying on stuff that's too new.

mhammond:
Me too. The m-c policy describes a rationale, but doesn't really go into detail on what happens when we break downstream. There might even be CI jobs that use that "requires" version - I'm really not sure.

At the end of the day, the decision to align with m-c was simple - but we don't really have the appetite for, nor see any real benefit in, trying to challenge it at this point, particularly when we are actively starting to ship uniffi-backed code into that repo.

jplatte:
Okay, how about spelling out the actual effective MSRV (currently 1.59) then, and updating [CI](https://github.com/mozilla/uniffi-rs/blob/80994f899abca2f1f00cd25a1e536d1795077791/.circleci/config.yml#L25) accordingly? (the latter has to be bumped when MSRV changes anyways, so might as well spell out the exact version in docs as well)

I would also like to advocate for either specifying the true MSRV in `rust-toolchain.toml`, or specifying none at all. Right now, this just leads to people auto-installing older toolchains than `stable` that are still not the minimum version tested in CI.

mhammond:
> Okay, how about spelling out the actual effective MSRV (currently 1.59) then, and updating [CI](https://github.com/mozilla/uniffi-rs/blob/80994f899abca2f1f00cd25a1e536d1795077791/.circleci/config.yml#L25) accordingly?

Oops - yeah, we should.

mhammond:
Hey @jplatte, I can request review from you now :) You had this version bump queued up too, so figured I should just do it at the same time.

docs/policies/rust-versions.md
```diff
@@ -7,19 +7,23 @@ which aren't always able to be on that latest version.
...
-but everything in the future is documented as "estimated".
+The Rust policy for mozilla-central has an official
+[Rust Update Policy Document
+](https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html),
+the tl;dr of which is:
 
-Ultimately though, that page defines 2 Rust versions - "Uses" and "Requires",
-and our policy revolves around these.
+* There's a "Uses" version which all of Mozilla's CI uses for releasing Firefox.
+  Discover it in the most recent updates to [this meta bug on Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1504858)
```
jplatte, line 21: How relevant is this for UniFFI, really?

docs/policies/rust-versions.md
```diff
@@ -7,19 +7,23 @@ which aren't always able to be on that latest version.
...
+
+* There's a "Requires" version, which Mozilla is committed to supporting,
+  primarily for downstream builders such as Linux distributions.
+  [Discover it here](https://searchfox.org/mozilla-central/search?q=MINIMUM_RUST_VERSION&path=python/mozboot/mozboot/util.py). This version is tested in mozilla-central's CI,
+  but nothing is shipped by mozilla with it.
```
jplatte, line 26: Can't the version just be stated here, with an HTML comment to update it in lockstep with `.circleci/config.yml` (and an equivalent comment there)?

docs/policies/rust-versions.md
mhammond: It's relevant in that uniffi backed code is starting to land in m-c. You are advocating for the uniffi policy to just specify a single version, which is the m-c "requires"?

jplatte: Yes. Why would I care to know what Mozilla's CI uses, if I need to constrain myself to the features available in a (slightly) older release anyways?

docs/policies/rust-versions.md
jplatte: The repo has the churn anyways through the CircleCI config, updating two locations isn't really harder than one if you have to do it anyways.

mhammond: The repo does, but this seems redundant - a link to the canonical truth is better than a copy of today's truth? I'm slightly surprised this PR is controversial.

mhammond:
It sounds like you would prefer to see a different policy, which is fine and worth discussion. But is there some way that this PR doesn't clarify the existing policy?

docs/policies/rust-versions.md
jplatte: If you say you have an MSRV of whatever's in Firefox, but then a contributor uses features from that and CI fails because you haven't bumped the value in CircleCI config, that's a bad experience. If you're really against having the version tested by CI in the text here, maybe you can link to the CircleCI config in addition, and leave a note saying that if the two disagree, one should just update the CircleCI config?

bendk (approved):
I think the version policy could definitely use an overhaul, maybe we could stick with the minimum Rust version that Firefox is on.

In the meantime, let's merge this.  I got hit with an issue last week because the minimum version was too low.

mhammond:
> I think the version policy could definitely use an overhaul,

Yeah, agreed.

> maybe we could stick with the minimum Rust version that Firefox is on.

I don't really have a problem with that, but do you mean we'd also release builds using that version? Which would mean uniffi in m-c would be using a different compiler than uniffi itself ships with?

mhammond:
far out, CI was green on this PR but main is now failing.

PR #1258, opened 2022-05-30
Some more proc-macro preparations

jplatte:
A few more refactorings to prepare for the addition of the first proc-macro frontend bits.


bendk (approved):
Looks good to me, another nice set of cleanups.

PR #1263, opened 2022-06-01
Map rustarc to object

### LINKED ISSUE #1262: Expose the name of the object an `RustArcPtr` is assoicated with
In our [`FFIType::RustArcPtr`](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/interface/ffi.rs#L37) we don't know what the name of the object the `RustArcPtr` came from. Having it available could help resolve type is associated with each `RustArcPtr`

tarikeshaq:
Fixes #1262 

This will save us some work in the `uniffi_gecko_js` bindings, where we try to make sure an FFI call is calling the correct pointer type - I feel like there might be a leaky abstraction here, but if we are okay with it, it'll be helpful in the gecko bindings


mhammond (approved):
This looks fine to me, and small enough that it's easy to back away from if necessary (not that I think it will be)

PR #1266, opened 2022-06-03
Make use of implicit format string arguments

jplatte:
Based on #1259

### DIFF
--- a/examples/rondpoint/src/lib.rs
+++ b/examples/rondpoint/src/lib.rs
@@ -179,7 +179,7 @@ impl Stringifier {
         value.to_string()
     }
     fn well_known_string(&self, value: String) -> String {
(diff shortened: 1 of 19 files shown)


mhammond:
I like the string formatting changes (ie, 60d830baa16aa8723184f055c8940951944bdf1f plus the minver bump) - but did you mean to include some of the macro work here too?

mhammond:
(and to be clear, I don't *not* like the macro changes, but they require more than a rubber stamp :)

jplatte:
This is a draft PR because it contains all of the commits from #1259 from the time of creating it. I intend to rebase once that is merged so there will be no more macro changes.

mhammond (approved):
Thanks!

Issue #1260, opened 2022-06-01
Please upgrade weedle2 to nom 7


glandium:
Currently, building uniffi_bindgen requires 2 versions of nom: 5, through weedle2 and 7, through askama. It would be better to only require one.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-169)
┆friendlyId: UNIFFI-169

badboy:
nom 5 -> 6 upgrade for weedle is pretty simple.
but 6 -> 7 is more involved as all macros are dropped, which weedle relies on heavily.

PR #1268, opened 2022-06-07
Upgrade weedle2 to nom 6

glandium:
The big bummer here is that because weedle2 re-exports `nom` types and exposes them in return types, each nom version bump requires a version bump of the weedle2 crate itself.


glandium:
Related: #1260

PR #1273, opened 2022-06-14
Pin to clap 3.1 and modernize some aspects of clap usage.

### LINKED ISSUE #1272: CI fails with clap=3.2
We specify [clap 3.1](https://github.com/mozilla/uniffi-rs/blob/e9d9aae095794d66f257aa11bc28e1b52cb261e8/uniffi_bindgen/Cargo.toml#L22). however, when running against `3.2` some warnings are omitted causing our CI to fail.

It looks like we should either:
* pin exactly 3.1
* upgrade the code to stop warning in 3.2 (but sadly, it appears impossible to have code work in both without warnings in either)
* stop failing on warnings in CI, possibly targeting these specific warnings?

I'm going to put a PR up which does the 2nd option, but I'm not 100% sure we want it yet, as m-c is on 3.1.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-175)
┆friendlyId: UNIFFI-175

mhammond:
Fixes #1272, [edit to remove references to bumping the clap version]

### DIFF
--- a/uniffi_bindgen/Cargo.toml
+++ b/uniffi_bindgen/Cargo.toml
@@ -19,7 +19,7 @@ anyhow = "1"
 askama = { version = "0.11", default-features = false, features = ["config"] }
 cargo_metadata = "0.14"
 camino = "1.0.8"
-clap = { version = "3.1", features = ["cargo", "std", "derive"] }
+clap = { version = "~3.1", features = ["cargo", "std", "derive"] }
 fs-err = "2.7.0"
 heck = "0.4"
 paste = "1.0"
--- a/uniffi_bindgen/src/lib.rs
+++ b/uniffi_bindgen/src/lib.rs
@@ -469,11 +469,12 @@ impl<V: Clone> MergeWith for HashMap<String, V> {
     }
 }
 
-// structs to help our cmdline parsing.
+// structs to help our cmdline parsing. Note that docstrings below form part
+// of the "help" output.
+/// Scaffolding and bindings generator for Rust
 #[derive(Parser)]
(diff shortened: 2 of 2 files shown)


jplatte:
For other projects I've just set the version specification to `~3.1`, which means cargo won't select 3.2 or above. I don't think there's much to gain from `value_parser` right now.

mhammond:
Thanks for that and yeah, I agree that's probably the best course of action. I might try and keep the non-breaking changes in my PR though

jplatte (approved):

badboy (approved):
👍
especially for staying on 3.1
Going to 3.2 will most likely cause pain getting it into m-c because of _other_ things there depending on clap and thus also requiring the update. I'm not keen on doing that anytime soon really (I only just last week did land a 3.1 update to some other part)

uniffi_bindgen/Cargo.toml
```diff
@@ -19,7 +19,7 @@ anyhow = "1"
 askama = { version = "0.11", default-features = false, features = ["config"] }
 cargo_metadata = "0.14"
 camino = "1.0.8"
-clap = { version = "3.1", features = ["cargo", "std", "derive"] }
+clap = { version = "~3.1", features = ["cargo", "std", "derive"] }
```
epage, line 22: This can now be reverted as deprecations are now off by default, see clap-rs/clap#3830 for details.  When you do start working through the deprecations as part of upgrading to 4.0, we've also improved the warnings for derive users.  See clap-rs/clap#3832 for examples.

PR #1278, opened 2022-06-17
Unpin clap

jplatte:
The problematic deprecation warnings are now opt-in.

### DIFF
--- a/uniffi_bindgen/Cargo.toml
+++ b/uniffi_bindgen/Cargo.toml
@@ -19,7 +19,7 @@ anyhow = "1"
 askama = { version = "0.11", default-features = false, features = ["config"] }
 cargo_metadata = "0.14"
 camino = "1.0.8"
-clap = { version = "~3.1", features = ["cargo", "std", "derive"] }
+clap = { version = "3.1", features = ["cargo", "std", "derive"] }
 fs-err = "2.7.0"
 heck = "0.4"
 lazy_static = "1.4"


uniffi_bindgen/Cargo.toml
```diff
@@ -19,7 +19,7 @@ anyhow = "1"
 askama = { version = "0.11", default-features = false, features = ["config"] }
 cargo_metadata = "0.14"
 camino = "1.0.8"
-clap = { version = "~3.1", features = ["cargo", "std", "derive"] }
+clap = { version = "3.2.5", features = ["cargo", "std", "derive"] }
```
badboy, line 22: Please stay at 3.1.
Fixing it to some higher version might be problematic when we integrate that back into mozilla-central, and I'd rather not deal with that at the moment.

badboy (changes requested):

uniffi_bindgen/Cargo.toml
jplatte: Is it okay for the minimum version to be 3.1, but allow higher versions within 3.x? We would ideally want to use 3.2 and also the `builtin-bindgen` feature. (at least we activate it now, I'm not 100% certain we actually need to)

badboy: 3.1 is fine, that what we use currently.
Shouldn't just `version = "3.1."` work now and then pick 3.2 for those that can and 3.1 for when they can't?

jplatte: Yes, that's what I meant with the minimum version. Will adjust.

badboy (approved):

jplatte:
Was this meant to be merged after approval? 😄

badboy:
> Was this meant to be merged after approval? 😄

me: *writes down how to merge*
also me: *forgets his own docs*

yes.

PR #1280, opened 2022-06-20
Add missing metadata information for uniffi_meta

badboy:


uniffi_meta/Cargo.toml
```diff
@@ -2,6 +2,12 @@
 name = "uniffi_meta"
 version = "0.19.1"
 edition = "2021"
+description = "uniffi_meta"
+documentation = "https://mozilla.github.io/uniffi-rs"
```
jplatte, line 6: This overwrites the docs.rs link crates.io would add otherwise, I recommend not including this field.

mhammond (approved):
oops, sorry I missed this!

PR #1282, opened 2022-06-23
Some proc-macro frontend improvements

jplatte:
Actually, there's also a small UDL codegen improvement in here, but just because I was looking what it was doing for exceptions 🙂 

I have support for `Arc<_>` (and exporting `impl` blocks) in a different branch, but differently to these types, I have some soundness concerns with that that I raised on Matrix and would like to discuss before opening the PR.


skhamis (approved):
Nice cleanup/improvements! Yeah the Arc<_> stuff definitely gets less trivial so appreciate putting that in a separate PR. LGTM!

PR #1283, opened 2022-06-23
Replace lazy_static with once_cell

jplatte:
This makes the code a little bit easier to read because it's not a macro, and gets some projects closer to zero dependencies on `lazy_static` (other large projects like tokio have switched quite a while ago).


bendk (approved):
Seems good to me.  I also just checked and moz-central is on the same version, so this shouldn't cause us any dependency headaches.

Issue #1257, opened 2022-05-30
Add a procedural macro frontend


jplatte:
This is a tracking issue for the proc-macro work, initially discussed in [#385](https://github.com/mozilla/uniffi-rs/issues/385#issuecomment-1098194211), on Matrix and recently in a video meeting.

### Outline
* Introduce an attribute `#[uniffi::export]` that can go onto functions and `impl` blocks
  * Attributes like `#[uniffi::default]` will be used to express binding properties that can't be expressed in Rust directly¹
* Introduce a derive attribute `uniffi::FfiConvertible` to allow a custom type to be used in the FFI API²
  * Attributes like `#[uniffi(default)]` will be used to express binding properties that can't be expressed in Rust directly³

### Progress
* [ ] Functions
  * [x] Argument passing
  * [x] Return value passing
  * [x] Fallible (returning `Result<T, E>`)
  * [ ] Argument defaults
* [ ] Types
  * [x] Primitives
  * [x] `Arc<_>`
  * [x] By-value structs and enums
    * [ ] Field defaults
  * [x] Enums
  * [ ] External types
* [ ] Parsing `#[doc = "…"]` attributes
* [x] `impl` blocks
  * [x] Associated functions (`static` methods in some target languages)
  * [x] (True) methods with `self` parameters
  * [x] Recognize `Self` keyword
  * [x] Constructors (`#[uniffi::constructor]`)
* [x] Traits
* [x] Callback interfaces

¹ Using custom syntax like `foo: u32 = 0` the parameter list would be possible, but it's undesireable because (a) it would be confusing and (b) it would degrade IDE functionality
² This could be handled by the same `#[uniffi::export]` attribute but that's undesireable for technical reasons
³ Syntax difference is warranted due to proc-macro peculiarities

I can elaborate on any of these footnotes if somebody wants to know more ^^

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-168)
┆Issue Number: UNIFFI-168

jplatte:
About the current status: Some things have their boxes ticked but don't have PRs yet. The reason for that is that most of the work I've been doing so far has not been under the assumption that we'll want to mix UDL and proc-macro interface definitions. However, recently I realized that this shouldn't be so hard and has a number of big advantages:

* Individual proc-macro bits can be tested and used ahead of achieving feature parity
  * This also means small PRs are possible, which makes collaboration on this much easier
[...]

PR #1292, opened 2022-06-30
Support impl blocks in proc-macro frontend

jplatte:
Part of #1257.

### DIFF
--- a/uniffi_macros/src/export/scaffolding.rs
+++ b/uniffi_macros/src/export/scaffolding.rs
@@ -4,14 +4,14 @@
 
 use proc_macro2::{Ident, Span, TokenStream};
 use quote::{format_ident, quote};
-use syn::{FnArg, ItemFn, Pat, ReturnType};
+use syn::{FnArg, Pat, ReturnType, Signature};
 
 pub(super) fn gen_fn_scaffolding(
-    item: &ItemFn,
+    sig: &Signature,
     mod_path: &[String],
     checksum: u16,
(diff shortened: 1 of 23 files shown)


jplatte:
Ready for review! 🎉

jplatte:
Ah, it looks like we need to generate extra scaffolding code for destructors. I hope the force-push I just did works around that, so this PR can be merged as-is instead of growing more in terms of complexity and code size.

uniffi_macros/src/export/scaffolding.rs
```diff
@@ -3,81 +3,169 @@
...
+            quote! {
+                #ffi_converter::try_lift(this).unwrap_or_else(|err| {
+                    ::std::panic!("Failed to convert arg 'self': {}", err)
+                }).
             }
         }
-    }
+        _ => {
+            params_args.extend(collect_params(&sig.inputs, RECEIVER_ERROR));
+            quote! { #self_ident:: }
```
bendk, line 70: It's very nice how this adds support for static functions on the scaffolding side, but I don't think the support is there on the bindings side yet.  Maybe this should be commented out and replaced with a panic for now.

bendk (commented):
This is great progress.  I just had a couple comments/questions.

uniffi_macros/src/export/scaffolding.rs
jplatte: Right. Adressed in latest commit.

bendk (approved):

PR #1298, opened 2022-07-11
Store macro metadata in the cdylib file

bendk:
The nice thing about this system is that the metadata is always bundled
together with the build output.  This makes it easier to ensure that the
generated scaffolding matches up with the dylib that it's going to be link to.
This avoids the work that `rebuild_metadata()` needed to do.

The downside is that we need to parse a dylib, which feels slightly risky.
However, it seems less risky overall to me, since we don't have to worry about
tracking the JSON files -- especially after fixing the recent the sccache issue.

In order to use the macro metadata, you now need to specify `--cdylib
[path-to-library]` when running `uniffi-bindgen`.  This is annoying, but it
will be simpler once the proc-macros can handle all parts of the interface.
At that point, we can make `uniffi-bindgen` accept either a UDL path or a cdylib
path as it's positional arg.

I didn't add support for external bindings to pass in a cdylib path, since
adding an argument to that function would be a breaking change, then we would
need to do another breaking change to make the param `udl_or_cdylib_file`.  If
external bindings really want to, they can call
`uniffi_bindgen::interface::macro_metadata::add_to_ci` directly.

Added the `uniffi-bindgen dump-json` command that inputs a cdylib path and
prints the matadata as JSON.

I tested that `dump-json` works properly on the following targets:
  - x86_64-unknown-linux-gnu (ELF 64-bit)
  - i686-unknown-linux-gnu (ELF 32-bit)
  - x86_64-apple-darwin (Mach-O 64-bit)
  - x86_64-pc-windows-gnu (DLL 64-bit)
  - i686-pc-windows-gnu (DLL 32-bit)

This seems like good enough coverage to me, although there are a lot of other
systems that would be nice to test on.  The limiting factor was setting up the
cross-compilation toolchains on my machine.  Maybe we should add some more CI
platforms that just run macro-metadata-related tests.

Updated the testing code to pass the cydlib path, rather than the
directory that contains it.


bendk:
@jplatte No rush since you're out on vacation, but I'm definitely interested to hear what you think of this approach.

bendk:
One issue we would eventually need to figure out is the megazording process, where you bundle multiple UniFFI cartes together and therefore all the serialized metadata.  I think we could handle that by:
  - Including the namespace in the symbol name
  - Adding a namespace option to `uniffi-bindgen`
  - If that option is present, we ignore metadata that doesn't match the given namespace

tarikeshaq (commented):
Given the context I have, this looks good! I dropped a few comments as I went through, but nothing was blocking! I'll leave the approval till after Jonas takes a look since it impacts that work 🚀

uniffi_macros/src/export/metadata.rs
```diff
@@ -171,29 +154,13 @@ fn type_not_supported(ty: &impl ToTokens) -> syn::Error {
...
-    // changed (practically this should only happen if it is deleted).
-    //
-    // Ideally this function would return `io::Result<()>` and use a proc_macro function for
-    // "tracking" the file instead of returning this TokenStream. That API is still nightly-only
-    // and subject to change though, see https://github.com/rust-lang/rust/issues/73921
-    let path = path.as_str();
-    Ok(quote! {
-        const _: &str = ::std::include_str!(#path);
-    })
+fn create_metadata_static_var<T: Into<Metadata>>(name: impl IdentFragment, val: T) -> TokenStream {
```
jplatte, line 157: Why is this generic? I think it would be a bit easier to grok if any necessary `.into()` calls happened at the call site.

uniffi_bindgen/src/macro_metadata/extract.rs
```diff
@@ -0,0 +1,105 @@
...
+
+pub fn extract_from_macho(macho: MachO<'_>, file_data: &[u8]) -> anyhow::Result<Vec<Metadata>> {
+    let mut items = vec![];
+    for export in macho.exports()? {
+        // Mach names typically have an extra "_" prepended, skip that.
+        let name = if export.name.starts_with('_') {
+            &export.name[1..]
+        } else {
+            &export.name
+        };
```
jplatte, line 84:
```suggestion
        let name = export.name.strip_prefix('_').unwrap_or(&export.name);
```

jplatte (commented):
This is great! I wouldn't have expected that symbol extraction from binaries is so simple (in terms of lines of code – I don't know anything about these binary formats so the extraction code is pretty opaque to me).

badboy:
Overall this looks fine to me, but I have some questions/requests:

* How is this used? Can we update the docs to make this clear? (I'm fine if that happens in a follow-up)
* If this requires a cdylib, does that mean a user _always_ needs to build a compatible cdylib so it can get parsed? Some targets (iOS) cannot produce cdylibs, they generate static libs.

bendk:
Thanks for all the feedback everyone.  I plan to push changes that address it once I get back to my home tomorrow.

> If this requires a cdylib, does that mean a user _always_ needs to build a compatible cdylib so it can get parsed? Some targets (iOS) cannot produce cdylibs, they generate static libs.

I think we always need a lib, but maybe the static lib could work.  Is it in archive format?

jplatte:
> I think we always need a lib, but maybe the static lib could work. Is it in archive format?

I would expect it to be. The description for "Archive" in goblin just says "a Unix archive", and a `staticlib` file produced by rustc on Linux has its file type identified as "current ar archive". I don't think macOS and iOS are that different when it comes to that kind of low-level thing.

uniffi_bindgen/src/macro_metadata/extract.rs
```diff
@@ -0,0 +1,105 @@
...
+    name.starts_with("UNIFFI_META")
+}
+
+pub fn extract_from_bytes(file_data: &[u8], offset: usize) -> anyhow::Result<Metadata> {
+    // Use the file data starting from offset, without specifying the end position.  We don't
+    // always know the end position, because goblin reports the symbol size as 0 for PE and
+    // MachO files.
+    //
+    // This works fine, because bincode knows when the serialized data is terminated and will
+    // just ignore the trailing data.
```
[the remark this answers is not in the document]
bendk: It's possible, but it seems very unlikely that the data would be still valid bincode.  I was considering adding a checksum and/or a length value before the bincode data.  I decided that it didn't seem worth it, but I'm not completely confident in that.

bendk:
Thanks for all the review!  I updated the PR based on all your comments.


>  How is this used? Can we update the docs to make this clear? (I'm fine if that happens in a follow-up)

I made #1302 as a followup for docs (I think we'll have a lot of documentation to add once this gets closer to being ready to use.

>  If this requires a cdylib, does that mean a user _always_ needs to build a compatible cdylib so it can get parsed? Some targets (iOS) cannot produce cdylibs, they generate static libs.

I added code to parse archive files and this seems to work, when I test it against libraries built with `target=aarch64-apple-ios`.  Inside the archives are a bunch of `.o` files, which contains the symbol data we need.  This now means that we use 2 different strategies to parse mach-o files.  For `.dylib` files, we iterate through the exports.  For those embedded `.o` files, we iterate through the global symbols.  I was a bit worried about this for 2 reasons.  I think it's okay, but I want to double check with you all about it:

 - I was worried that the symbols could be stripped somehow, but I don't think that should happen since they're flagged as globals.  I believe globals are like exports, but I'm confused because goblin doesn't return them in the `.exports()` method.  In any case, I don't think these symbols can be stripped since we generate them just like the scaffolding functions and those definitely need to be linked to somehow.  I tested against a release build and it worked, which gives me some more confidence on this one.
 - I was also worried that using both strategies could lead to duplicate items if somehow the same piece of metadata was listed in `.symbols()` and `.exports()`.  To avoid that, I just stored the extracted symbol names in a hash table and checked that we don't try to extract the same name twice.

bendk:
@badboy @jplatte do you think we can merge this at this point?

jplatte (approved):
Good to go from me, I have one nitpick but no big concerns.

badboy:
> @badboy @jplatte do you think we can merge this at this point?

No blocker from my side!

tarikeshaq (approved):
🚢

PR #1304, opened 2022-08-04
Upgrade weedle2 to nom 7

### LINKED ISSUE #1260: Please upgrade weedle2 to nom 7

flisky:
fix #1260

### DIFF
--- a/uniffi_bindgen/Cargo.toml
+++ b/uniffi_bindgen/Cargo.toml
@@ -29,5 +29,5 @@ paste = "1.0"
 serde = "1"
 serde_json = "1.0.80"
 toml = "0.5"
-weedle2 = { path = "../weedle2" }
+weedle2 = { version = "4.0.0", path = "../weedle2" }
 uniffi_meta = { path = "../uniffi_meta", version = "=0.19.3" }
--- a/uniffi_bindgen/Cargo.toml
+++ b/uniffi_bindgen/Cargo.toml
@@ -29,5 +29,5 @@ paste = "1.0"
 serde = "1"
 serde_json = "1.0.80"
 toml = "0.5"
-weedle2 = { version = "3.0.0", path = "../weedle2" }
+weedle2 = { path = "../weedle2" }
 uniffi_meta = { path = "../uniffi_meta", version = "=0.19.3" }
--- a/weedle2/Cargo.toml
+++ b/weedle2/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "weedle2"
-version = "3.0.0"
+version = "4.0.0"
 authors = ["Sharad Chand <sharad.d.chand@gmail.com>", "Jan-Erik Rediger <jrediger@mozilla.com>"]
 description = "A WebIDL Parser"
 license = "MIT"
@@ -16,5 +16,7 @@ exclude = ["tests"]
 name = "weedle"
 
 [dependencies]
+nom = "7.1.0"
+
+[dev-dependencies]
 fs-err = "2.7.0"
-nom = { version = "6.0.0", default-features = false, features = ["std"] }
--- a/weedle2/src/lib.rs
+++ b/weedle2/src/lib.rs
@@ -21,21 +21,6 @@
 //!
 //! If any flaws found when parsing string with a valid grammar, create an issue.
 
-// need a higher recusion limit for macros
-#![recursion_limit = "128"]
-
-#[macro_use(
-    alt,
-    cond,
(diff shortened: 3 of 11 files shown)


weedle2/src/lib.rs
```diff
@@ -21,21 +21,6 @@
...
-    do_parse,
-    map,
-    many0,
-    opt,
-    recognize,
-    separated_list0,
-    separated_list1,
-    terminated
-)]
 extern crate nom;
```
badboy, line 24: Can probably now drop this `extern crate`.

```suggestion
```

weedle2/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "weedle2"
-version = "3.0.0"
+version = "4.0.0"
```
badboy, line 3: If we don't change the public API nor export any nom types we can make this a minor release.

uniffi_bindgen/Cargo.toml
```diff
@@ -29,5 +29,5 @@ paste = "1.0"
 serde = "1"
 serde_json = "1.0.80"
 toml = "0.5"
-weedle2 = { version = "3.0.0", path = "../weedle2" }
+weedle2 = { path = "../weedle2" }
```
badboy, line 32: Hm, don't we need to specify the version so we can publish this?

weedle2/src/lib.rs
```diff
@@ -47,7 +32,7 @@ use self::literal::StringLit;
 use self::mixin::MixinMembers;
 use self::namespace::NamespaceMembers;
 use self::types::{AttributedType, ReturnType};
-pub use nom::{error::Error, error::ErrorKind, Err, IResult};
+use nom::{error::Error, Err, IResult};
```
badboy, line 35: The `Parse` trait is public and uses the `IResult`, `Err` and `Error` are used in the `parse` function, we should continue exporting them.

badboy (changes requested):
Appreciate the work here!
If the tests pass I'm fairly confident the translation to macro-less nom is correct here.

Minimal nits as comments inline, but it would be great if we get this merged soon then.

weedle2/src/lib.rs
jplatte: That makes this a breaking change though. I still find it unusual to bump the version in this PR rather than at the point of release though.

weedle2/src/literal.rs
```diff
@@ -208,14 +196,14 @@ mod test {
         FloatLit => FloatLit::Value(FloatValueLit("-435.3435"))
     });
 
-    test!(should_parse_float_exp { "5.3434e23" =>
+    test!(should_parse_float_exp { "3e23" =>
```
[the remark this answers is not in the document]
flisky: actually, I just switch the name and position between `should_parse_float_exp` and `should_parse_float_exp_without_decimal`.

weedle2/src/lib.rs
flisky: so, is `3.1.0` a good version for us if I remove the breaking changes? I just follow the #1268, and it said an version bump is needed if `nom` upgrades.

flisky:
I bump the version to `3.1.0`, and hope it's okay for us.

weedle2/src/lib.rs
jplatte: The breaking changes cannot be removed as bumping `nom` is itself a breaking change, since its types are part of the public interface. I would recommend not touching the version of uniffi crates at all and leaving that up to the person who does the release.

badboy: Yeah, doing a breaking change is fine. We basically maintain `weedle2` for use in UniFFI, but we can still be good citizens and do it properly.

badboy:
> I bump the version to `3.1.0`, and hope it's okay for us.

Yeah, as @jplatte correctly mentioned we do need the major version bump, because we do export the types.
You can also fully remove the version bump and I'll take care of that in a release commit.

jplatte:
> You can also fully remove the version bump and I'll take care of that in a release commit.

I think this is what most projects do as doing the bump before the release makes it impossible to `[patch]` in the git dependency to replace the crates.io package.

badboy (approved):

PR #1306, opened 2022-08-04
[Discuss] Async use cases

bendk:
I don't know if this document should be added to the docs section, the main reason I'm opening the PR is so we can discuss the use cases more.

### DIFF
--- a/docs/async-use-cases.md
+++ b/docs/async-use-cases.md
@@ -1,0 +1,95 @@
+We've been talking for awhile about "Async UniFFI", but it's never been clear exactly what that means.  A lot of our async use cases cover very different situations, for example:
+
+  - Creating a thread-pool on the foreign language side that we make blocking Rust calls from.  This is the main async use case in practice today, and there has been discussion about supporting it better by auto-generating decorator classes.
+  - Making a call into Rust that starts a tokio-based network request and getting a Promise/Future back on the foreign side.
+  - Making an async call from Rust into the foreign side, suspending the Rust function until we have a result (JEXL evaluation)
+  - Making a blocking call from a Rust-controlled thread into the foreign side (e.g. a Rust thread calling into viaduct).
+
+There's a lot of possibilities, but maybe we can group all uses cases using 4 dimensions:
+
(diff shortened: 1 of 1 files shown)


docs/async-use-cases.md
```diff
@@ -0,0 +1,95 @@
...
+This is the same as the last one, except the call is async.  The main use case is probably the same
+as the previous section, except when Rust is using an event loop rather than a thread pool.
+
+# New features
+
+Based on the above analysis, I believe there are 2 features that we should consider adding:
+
+  - Rust code making an async call into a CallbackInterface / Foreign code
+    making an async into Rust. I think this could be implemented by having
+    UniFFI generate something like the hand-written demo from #1252.
```
bendk, line 90: This seems like a nice feature, but I'm not sure it's necessary.  I think we could implement the Nimbus work with hand-written code.  Maybe we could just start there and see if there's more demand before generating that code with UniFFI.

docs/async-use-cases.md
```diff
@@ -0,0 +1,95 @@
...
+  - Handling CallbackInterface calls from a Rust-based thread. I'm not exactly sure how this would work, there's at least 2 options here:
+    - Handle everything on the foreign side. In the generate code we currently register a callback to invoke a CallbackInterface call. We might be able to update that code so that it schedules the call to run on the correct thread. But this assumes that it's safe to call that callback on the Rust thread. Is that true for all of our current languages? Are we okay with adding this requirement for future languages?
+    - Use a queue plus a waker. Push the CallbackInterface call to a queue, signal the foreign side (maybe writing a byte to a pipe or socket), then the foreign side would wake up and try to read from the queue. This seems more complicated than the first system, but might let us support more foreign languages.
```
bendk, line 93: This feature seems necessary if we ever want to make a call from a Rust controlled (or C++ controlled) thread into JS.  Maybe we want that for the Nimubs dekstop work, but I'm not sure.

docs/async-use-cases.md
badboy: I agree that hand-writing & using it that way seems sensible. That's how UniFFI was born to begin with.

bendk:
I think we have a good plan here, no need to keep this PR open longer.

PR #1308, opened 2022-08-12
Allow Rust traits to be exposed as an interface

mhammond:
We've a use-case where we'd like to have a .udl describe an interface that's actually implemented on the Rust side of the world as a trait rather than a struct. Hopefully the motivation will be clear when looking at the new fixture.

This actually turned out reasonably well, but there is still quite a bit of cleanup to do, and also other edge-cases to explore. I'd also welcome alternative strategies, and something tells me that there's some synergy now between traits and callback interfaces, but I'm yet to explore that. In the meantime, I thought I'd get this up for early feedback.

### DIFF
--- a/uniffi/src/lib.rs
+++ b/uniffi/src/lib.rs
@@ -629,6 +629,66 @@ unsafe impl<T: Sync + Send> FfiConverter for std::sync::Arc<T> {
     }
 }
 
+/// Support for passing reference-counted trait objects via the FFI.
+///
+/// This is different than the last one, since trait objects are stored as 2 pointers (AKA a "fat
+/// pointer): one for the `Arc<T>` and another for the vtable.  Since the foreign side is expecting
+/// a single pointer, everything gets wrapped in a Box
+unsafe impl<T: Sync + Send + ?Sized> FfiConverter for Box<std::sync::Arc<T>> {
+    type RustType = std::sync::Arc<T>;
+    // Don't use a pointer to <T> as that requires a `pub <T>`
+    type FfiType = *const std::os::raw::c_void;
+
+    /// When lowering, we have an owned `Arc<T>` and we transfer that ownership
+    /// to the foreign-language code, "leaking" it out of Rust's ownership system
+    /// as a raw pointer. This works safely because we have unique ownership of `self`.
+    /// The foreign-language code is responsible for freeing this by calling the
+    /// `ffi_object_free` FFI function provided by the corresponding UniFFI type.
+    ///
+    /// Safety: when freeing the resulting pointer, the foreign-language code must
+    /// call the destructor function specific to the type `T`. Calling the destructor
+    /// function for other types may lead to undefined behaviour.
+    fn lower(obj: Self::RustType) -> Self::FfiType {
+        Box::into_raw(Box::new(obj)) as *const std::os::raw::c_void
+    }
+
+    /// When lifting, we receive a "borrow" of the `Arc<T>` that is owned by
+    /// the foreign-language code, and make a clone of it for our own use.
+    ///
+    /// Safety: the provided value must be a pointer previously obtained by calling
+    /// the `lower()` or `write()` method of this impl.
+    fn try_lift(v: Self::FfiType) -> Result<Self::RustType> {
+        // Use Box::leak to avoid dropping the `Arc<T>` that is owned by the foreign-language code.
+        let foreign_arc = Box::leak(unsafe { Box::from_raw(v as *mut std::sync::Arc<T>) });
+        // Take a clone for our own use.
+        Ok(std::sync::Arc::clone(foreign_arc))
+    }
+
+    /// When writing as a field of a complex structure, make a clone and transfer ownership
+    /// of it to the foreign-language code by writing its pointer into the buffer.
+    /// The foreign-language code is responsible for freeing this by calling the
+    /// `ffi_object_free` FFI function provided by the corresponding UniFFI type.
+    ///
+    /// Safety: when freeing the resulting pointer, the foreign-language code must
+    /// call the destructor function specific to the type `T`. Calling the destructor
+    /// function for other types may lead to undefined behaviour.
+    fn write(obj: Self::RustType, buf: &mut Vec<u8>) {
+        static_assertions::const_assert!(std::mem::size_of::<*const std::ffi::c_void>() <= 8);
+        buf.put_u64(Self::lower(obj) as u64);
+    }
+
+    /// When reading as a field of a complex structure, we receive a "borrow" of the `Arc<T>`
+    /// that is owned by the foreign-language code, and make a clone for our own use.
+    ///
+    /// Safety: the buffer must contain a pointer previously obtained by calling
+    /// the `lower()` or `write()` method of this impl.
+    fn try_read(buf: &mut &[u8]) -> Result<Self::RustType> {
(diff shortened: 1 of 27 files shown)


jplatte:
Uhh... This already works without `[Trait]` though?

https://github.com/matrix-org/matrix-rust-sdk/blob/ae18b01c25a99ec9fe7c11f6524332af67e256ee/bindings/matrix-sdk-ffi/src/messages.rs#L238-L254

https://github.com/matrix-org/matrix-rust-sdk/blob/ae18b01c25a99ec9fe7c11f6524332af67e256ee/bindings/matrix-sdk-ffi/src/api.udl#L163-L165

bendk:
> Uhh... This already works without `[Trait]` though?

The goal would be to support a Trait that's implemented by multiple types and have allow the foreign code to have something that works like `Arc<dyn Trait>`.

bendk:
In the spirit of early feedback, here's a patch that makes the FfiConverter code add the box so user's don't have to: https://github.com/bendk/uniffi-rs/pull/new/traits-unboxed.

I also wonder about avoiding the box altogether.  AFAICT, the reason we need it is because `Arc<dyn Trait>` is 2 pointers (AKA a fat pointer), but the foreign code is only expecting to handle 1 pointer.  If the foreign code stored both pointers we could avoid the heap allocation.  That said, this might add enough complexity that it's worth it to keep the box.

The idea of combining this new thing with `CallbackInterface` is very appealing to me, then you could have Rust code that operated on a set objects where some were implemented in Rust and others were implemented in the foreign language.  It opens up some wild possibilities, like using `SyncManager` on Desktop where some of the `SyncEngines` are implemented in JS:
   - `SyncManager` would get a method like `register_engine(Arc<dyn SyncEngine>)`
   - Our Rust components would have a `create_sync_engine() -> Arc<dyn SyncEngine>` function
   - But you could also implement the `SyncEngine` interface in JS and pass that to `register_engine()`

How could that last part work?  We already have [code](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/scaffolding/templates/CallbackInterfaceTemplate.rs) that inputs a callback interface handle from the foreign side and creates a `Box<dyn CallbackInterfaceTrait>` from it.  We could refactor it to create an `Arc<>` instead, expose that as a scaffolding function, and add some code in the binding templates to glue it all together.  Hand-wavey I know, but I think it's doable.

This makes me think that maybe the `Type` hierarchy should change in a different way.  Instead of grouping this together with `Type::Object`, maybe it should be grouped together with `Type::CallbackInterface` which we then rename to `Type::Interface`.

mhammond:
> How could that last part work? We already have [code](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/scaffolding/templates/CallbackInterfaceTemplate.rs) that inputs a callback interface handle from the foreign side and creates a `Box<dyn CallbackInterfaceTrait>` from it.

Yeah - I think there is real scope here - ultimately, I think the only real difference will be whether the code expects the trait to already exist, or whether uniffi needs to generate it.

Thanks for that tweak - I'll play with it more later next week.

mhammond:
> How could that last part work? We already have [code](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/scaffolding/templates/CallbackInterfaceTemplate.rs) that inputs a callback interface handle from the foreign side and creates a `Box<dyn CallbackInterfaceTrait>` from it.

I tweaked the callbacks example to try and move towards this - there's now a "SimCard" trait (and I think that probably means my new fixture can die). My next step is to work out how to have the foreign code also implement this new trait, which IIUC, is what you described above.

> This makes me think that maybe the `Type` hierarchy should change in a different way. Instead of grouping this together with `Type::Object`, maybe it should be grouped together with `Type::CallbackInterface` which we then rename to `Type::Interface`.

I suspect it might play out slightly different to that and we might even be able to just remove CallbackInterface entirely - but that should become clearer once we flesh out some examples.

uniffi/src/lib.rs
```diff
@@ -629,6 +629,67 @@ unsafe impl<T: Sync + Send> FfiConverter for std::sync::Arc<T> {
...
+    }
+
+    /// When lifting, we receive a "borrow" of the `Arc<T>` that is owned by
+    /// the foreign-language code, and make a clone of it for our own use.
+    ///
+    /// Safety: the provided value must be a pointer previously obtained by calling
+    /// the `lower()` or `write()` method of this impl.
+    fn try_lift(v: Self::FfiType) -> Result<Self::RustType> {
+        // Use Box::leak to avoid dropping the `Arc<T>` that is owned by the foreign-language code.
+        let foreign_arc = Box::leak(unsafe { Box::from_raw(v as *mut std::sync::Arc<T>) });
```
bendk, line 663: And I guess this is where the leak comes from?  This really makes me think `Arc<Box<T>>` is the way to go and maybe it's more precise to call this `Arc<Box<dyn Trait>>`.  I think I can even wrap my head around it because it's closer the the way we both intuitively thought that things worked.  `Box<dyn Trait>` is a fat pointer, where one points to the underlying T and the other points to the vtable.  Then `Arc<Box<dyn trait>>` let's you share that between the foreign and Rust code, with the nice side benefit it's just a regular, skinny, pointer.

In fact, could you then just use the existing `Arc<T>` impl?

mhammond: I'm still a bit confused by this TBH - I think I need to better understand the leak.

My only real problem with `Arc<Box<dyn T>>` is the ergonomics, and I'm not even sure how much of a problem I think it is :) But having an argument be `foo: Arc<Box<impl Foo>>` seems like it's going to wind up with lots of `&**foo` and other gymnastics which makes the code look messy and just surfaces the limitations of UniFFI rather than looking "natural".

So I'm still trying to think about ways we can avoid this and use the slightly more natural `Arc<impl Foo>` in the signatures of functions being exposed. Although we discussed how ugly and hacky it would be, and I expressed hostility to the idea, I am starting to wonder how bad it would be to try and pull some of the tricks mentioned in [this discussion about `std::raw::TraitObject`](https://github.com/rust-lang/rust/issues/27751)? Or whether we can make the current model not leak (which seems like it *should* be possible, although might end up being very heavy on many small, short-lived allocations). Or something :) WDYT?

> I think whatever way we choose to implement this, we need good documentation about how everything is layed out.

Agreed - mermaid FTW I guess :)

mhammond: > I think I need to better understand the leak.

I can't actually reproduce a leak in that lower/lift implementation - there is a leak in my branch, but it doesn't look like it's there - so let me get more info before we start exploring the more radical options.

bendk: > My only real problem with `Arc<Box<dyn T>>` is the ergonomics, and I'm not even sure how much of a problem I think it is :) But having an argument be `foo: Arc<Box<impl Foo>>` seems like it's going to wind up with lots of `&**foo` and other gymnastics which makes the code look messy and just surfaces the limitations of UniFFI rather than looking "natural".

That's a good point.  I wonder if we could write our own wrapper type that uses `Arc<Box<dyn T>>` internally and has `Deref` impl that makes it easy to use.  Or maybe the type should wrap `Box<dyn T>` in a way that makes it easy to use when wrapped in an Arc.

mhammond:
This is now in #1457

PR #1309, opened 2022-08-16
Fix types in the callback interfaces examples

quad:
I got thrown off trying to follow these examples. This PR updates the documentation to match what worked for us!

### DIFF
--- a/docs/manual/src/udl/callback_interfaces.md
+++ b/docs/manual/src/udl/callback_interfaces.md
@@ -47,7 +47,7 @@ which may use it.
 In UDL:
 
 ```webidl
-object Authenticator {
+interface Authenticator {
     constructor(Keychain keychain);
     void login();
 }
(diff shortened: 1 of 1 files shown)


docs/manual/src/udl/callback_interfaces.md
```diff
@@ -47,7 +47,7 @@ which may use it.
 In UDL:
 
 ```webidl
-object Authenticator {
+interface Authenticator {
```
bendk, line 50: I think this should actually be `callback interface Authenticator {`

bendk (changes requested):
Looks good and sorry that the docs caused issues for you.

I had one small suggestion to improve them even more.

docs/manual/src/udl/callback_interfaces.md
quad:
```suggestion
callback interface Authenticator {
```

quad: Are you sure? As I read it, the `Authenticator` is an interface that _uses_ the `callback interface Keychain`.

bendk: Ahh yes, my mistake.  Merging now.

bendk (approved):

PR #1310, opened 2022-08-17
Fixes argument names for callback interfaces in swift to be camelcase

tarikeshaq:
There was a mismatch in swift in the arguments of the callback interface, when calling the callback interface, the swift code generates snake_case arguments instead of camelcase.

I could define a unique test for this, but I changed the existing test instead, happy to go back and make a regression test for it though!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.19.3...HEAD).
 
 - Implement Timestamp and Duration types in Ruby backend.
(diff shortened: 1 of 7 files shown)


mhammond (approved):
LGTM. a CHANGELOG entry wouldn't hurt, but IIUC, callback interface args with an underscore in their name simply didn't work, so this isn't a breaking change, right?

tarikeshaq:
> LGTM. a CHANGELOG entry wouldn't hurt, but IIUC, callback interface args with an underscore in their name simply didn't work, so this isn't a breaking change, right?

Yeah... but I'll still add a changelog! I'll fix the failing tests (forgot to change them for the other foreign languages to include the underscode) and add a changelog

PR #1313, opened 2022-08-23
Adding Todolist `uniffi.toml` with geckojs section

bendk:
This PR adds a `uniffi.toml` section for configuring threading that I hope to use for the Desktop JS project.

The backstory is that desktop privileged JS has a unique threading model: You should never block the main thread and there's no way support for thread pools or other ways to start a new thread (web workers are currently not used and we didn't want to introduce them just for UniFFI support).  However, the C++ code that runs underneath the JS engine has a simple ways to run code in a worker thread.  We decided to make it so each function runs on a worker thread and returns a promise to the JS code.  This seems like the best choice for most functions, but some functions run quick enough that we can just make a blocking call.  It would be nice to allow those functions to run directly.  This becomes a bit more important once we introduce async functions -- either hand-written or with UniFFI support.  It's silly to run an async function in a worker thread and get a Promise of a Promise.

Those details are unique to Desktop JS, but the overall threading model is similar to other languages.  On Swift/Kotlin, we typically create thread pools to wrap blocking functions with an async API.  Because of this, I tried to come up with something fairly general that we could maybe eventually use for these languages.

This PR doesn't change any behavior.  I'm mainly looking to see if others agree with my design decisions:

- Putting the configuration in `uniffi.toml` rather than some config file in the moz-central
- The general system is to map function/method names to tags that indicate the threading model for that call.  I'm hoping that this could maybe specify a thread pool on Swift/Kotlin, but this is just a hope.  I can't completely picture how this would work.
- I used `direct` and `worker` to specify the threading model for JS calls.  I wanted to distinguish this from `sync` / `async`, which seems like a different distinction to me.  `worker` calls are always async, but if we ever added async support in general to UniFFI I could easily see an async `direct` call (i.e. one that returns a Future).
- I used `-default` to specify the default model.  I think it's important to have a safe default in case a dev adds a function call, but forgets to add configuration for it.  The leading dash is there so that it doesn't conflict with an actual function name
- Method calls get put in their own nested table, with `-constructor` specifying the constructor threading.

What do you think about the overall design and also the TOML syntax?  I'm definitely open to bikeshedding suggestions for how this should be specified.

### DIFF
--- a/examples/todolist/uniffi.toml
+++ b/examples/todolist/uniffi.toml
@@ -1,0 +1,7 @@
+[bindings.geckojs.receiver_thread]
+default = "worker"
+main = [
+  "set_default_list",
+  "TodoList",
+  "TodoList.make_default",
+]
--- a/examples/todolist/src/todolist.udl
+++ b/examples/todolist/src/todolist.udl
@@ -1,6 +1,6 @@
 namespace todolist {
     TodoList? get_default_list();
-    undefined set_default_list(TodoList list);
+    void set_default_list(TodoList list);
 
     [Throws=TodoError]
     TodoEntry create_entry_with(string todo);


skhamis:
Since this is more or less discussion-based. I'll be adding back the full devs review so more eyes can see it.

bendk:
I should have mentioned why I think this should go into `uniffi.toml` and not the UDL file.  The reason is that this is configuring something specific to the bindings rather than the Rust code.

examples/todolist/uniffi.toml
```diff
@@ -0,0 +1,5 @@
+[bindings.geckojs.threading]
+-default = "worker"
+set_default_list = "direct"
+TodoList.-constructor = "direct"
+TodoList.make_default = "direct"
```
bendk, line 5: Maybe "main" makes things more clear than "direct".

skhamis: I like that this is in the toml vs udl as you mentioned it keeps the udl clear from any platform-specific markup. 

I'm not too much of a fan of the dashes as it doesn't give a clear indication of why we need the dashes (if someone brand new is looking at this)

This is just my ideas but don't feel super strongly about it, but some bike-shedding ideas: 
- instead of `default` maybe `default_threading_mode` or just `default_thread_mode`?
- I like `main` as this essentially comes down to worker thread vs main thread

Example with suggestions above:
```toml
[bindings.geckojs.threading]
default_threading_mode = "worker"
set_default_list = "main"
TodoList.-constructor = "main"
TodoList.make_default =  "main"
```

A slightly crazier idea could be to just have a list denoting either main thread or worker thread (while obviously only needing whichever is not the default).
```toml
[bindings.geckojs.threading]
default_thread_mode = "worker"

[bindings.geckojs.threading.mainthread]
list_of_calls = [
  "set_default_list",
  "TodoList.constructor", 
  "TodoList.make_default"
]

Obviously these are just ideas so no worries if it's not in the spirit of what we're trying to do here!

Comment (mhammond) [reply examples/todolist/uniffi.toml:5]: I agree this should be in the toml as it has no meaning on the rust side. OTOH though, the ergonomics of this is probably going to suck a little, so there might be a case to have binding-specific attributes on the UDL and/or in macro-based rust bindings, but this makes perfect sense to me to get the ball rolling.

If we can bash toml into playing nice, Sammy's idea of inverting things almost makes sense. If we consider that methods might end up with other attributes in the future, an alternative might also be to end up with something like:

```
...
[binding.geckojs.something.TodoList.set_default_list]
threading = "worker"
exception_behaviour = "something"
```

might also make sense (but might also make the ergonomic problem worse)

Finally, "threading" doesn't quite seem correct because it carries alot of baggage and comes with alot of assumptions. Something like "receiver_thread" or something *might* be more explicit, but I haven't thought that through, especially in the case of say, callbacks.

I suspect we will want a fixture in gecko to exercise some of this, but as mentioned, I think this is perfectly fine for a bit of a prototype and we can see how it shakes out in practice.

bendk:
Thanks for the suggestions.  I updated this one to:
  - Invert the configuration like Sammy suggested.  It reads way nicer this way.
  - Use "reciever_thread" like Mark suggested.  I agree that the more explicit the better.

mhammond (approved):
I wasn't sure if you actually wanted to land this, or just wanted feedback on the spelling. r=me if you do want to land it, but I think it might be slightly better to just land a new fixture in moz-central dedicated to the js bindings so it will be easier to iterate and be simpler to test edge-cases without adding them in this repo where the context will not be obvious.

bendk:
I was actually thinking of merging this one, but you're point about wanting to iterate is well taken.  I'm just going to close this and we can work in the moz-central branch for now.

PR #1314, opened 2022-08-24
Update Rust to 1.63, MSR to 1.61.0

mhammond:
As per our [Rust version policy](https://github.com/mozilla/uniffi-rs/blob/main/docs/policies/rust-versions.md)

Requires some clippy fixes. Also fixed some clippy complaints in nightly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ commands:
   # Our minimum supported rust version is specified here.
   prepare-rust-min-version:
     steps:
-      - run: rustup override set 1.59.0
+      - run: rustup override set 1.61.0
       - run: rustup update
   build-api-docs:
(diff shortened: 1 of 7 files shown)


mhammond:
A new version of `trybuild` seems to be at least one of the problems here - looking into it.

mhammond:
This works, but only with the min tests running with `-j 1` to avoid OOM. I've no idea why, and it seems worse to bump the machine to `xlarge`, so I'm opening this for review, but sadly it's not quite a rubber stamp.

mhammond:
@badboy pointed me at https://github.com/rust-lang/rust/issues/97549, so I think my approach seems sane.

tarikeshaq (approved):

PR #1316, opened 2022-08-26
Apply `omit_argument_labels` option when calling into a callback

### LINKED ISSUE #1312: [Swift] Bad generation of swift callback interface
Hey Uniffi team!

I just encountered an issue with the swift file generation which I suspect might be bug.

Th issue occurs when I set in the `uniffi.toml` the flag `omit_argument_labels = true`, changing it to `false` resolves the issue.

Screen Shot 2022-08-22 at 4 25 17 PM

The protocol has an underscore `didReceiveFriendRequest(_ user: UserModel)` where the `swiftCallbackInterface` has the extra `user:` which it shouldn't.

The UDL file looks like this:
```
callback interface ClientDelegate {
    void did_receive_friend_request(UserModel user);
};
```
And the Rust implementation like this:
```
pub trait ClientDelegate: Sync + Send {
    fn did_receive_friend_request(&self, user: UserModel);
}
```

Thank you!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-188)
┆friendlyId: UNIFFI-188

badboy:
This also adds a test to make sure we're not breaking it again.

Fixes #1312

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ members = [
   "fixtures/regressions/cdylib-crate-type-dependency/ffi-crate",
   "fixtures/regressions/cdylib-crate-type-dependency/cdylib-dependency",
   "fixtures/regressions/missing-newline",
+  "fixtures/regressions/swift-callbacks-omit-labels",
   "fixtures/uitests",
   "fixtures/uniffi-fixture-time",
   "fixtures/simple-fns",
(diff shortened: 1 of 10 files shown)


travis79 (approved):
Nice!

bendk (approved):
Looks good to me too.

PR #1318, opened 2022-08-28
Provide context to rustfmt-related errors

ja-cop:
If rustfmt is not installed or otherwise not found in $PATH,
uniffi-bindgen now explains that it's rustfmt that wasn't found, instead
of a generic "file not found" error message. Context is also added for
when rustfmt fails to format the scaffolding.

I'm a newbie at Rust and UniFFI, so any feedback is much appreciated! :slightly_smiling_face:

### DIFF
--- a/uniffi_bindgen/src/lib.rs
+++ b/uniffi_bindgen/src/lib.rs
@@ -94,12 +94,13 @@
 
 const BINDGEN_VERSION: &str = env!("CARGO_PKG_VERSION");
 
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
 use clap::{Parser, Subcommand};
 use fs_err::{self as fs, File};
 use serde::{Deserialize, Serialize};
 use std::io::prelude::*;
+use std::io::ErrorKind;
 use std::{collections::HashMap, env, process::Command, str::FromStr};
 
 pub mod backend;
@@ -277,11 +278,11 @@ pub fn generate_component_scaffolding(
     );
     let file_stem = udl_file.file_stem().context("not a file")?;
     let filename = format!("{file_stem}.uniffi.rs");
-    let out_dir = get_out_dir(udl_file, out_dir_override)?.join(filename);
-    let mut f = File::create(&out_dir)?;
+    let out_path = get_out_dir(udl_file, out_dir_override)?.join(filename);
+    let mut f = File::create(&out_path)?;
     write!(f, "{}", RustScaffolding::new(&component)).context("Failed to write output file")?;
     if format_code {
-        Command::new("rustfmt").arg(&out_dir).status()?;
+        format_code_with_rustfmt(&out_path)?;
     }
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


jplatte:
(I have some thoughts on this, but currently busy)

uniffi_bindgen/src/lib.rs
```diff
@@ -443,6 +444,26 @@ fn parse_udl(udl_file: &Utf8Path) -> Result<ComponentInterface> {
...
+    match Command::new("rustfmt").arg(path).status() {
+        Err(e) => {
+            let context = if e.kind() == ErrorKind::NotFound {
+                "formatting was requested, but rustfmt was not found"
+            } else {
+                "unknown error when calling rustfmt"
+            };
+            Err(e).context(context)
+        }
+        Ok(status) => {
```
jplatte, line 457: This match arm can instead be `.map_err(|e| { let ctx = match e.kind() { ... }; anyhow!(e).context(ctx) })?`.

uniffi_bindgen/src/lib.rs
```diff
@@ -443,6 +444,26 @@ fn parse_udl(udl_file: &Utf8Path) -> Result<ComponentInterface> {
...
+            if !status.success() {
+                Err(anyhow!("rustmt failed when formatting scaffolding. Note: --no-format can be used to skip formatting"))
```
jplatte, line 459: Returning `Err(anyhow!(...))` has a shorthand macro in anyhow called `bail!`.

jplatte (approved):
Looks good. I guess you posted this because you didn't hqve rustfmt installed in some setup? I wonder whether [prettyplease](https://github.com/dtolnay/prettyplease) could be a better fit for UniFFI's code formatting needs.

uniffi_bindgen/src/lib.rs
```diff
@@ -443,6 +444,20 @@ fn parse_udl(udl_file: &Utf8Path) -> Result<ComponentInterface> {
     ComponentInterface::from_webidl(&udl).context("Failed to parse UDL")
 }
 
+fn format_code_with_rustfmt(path: &Utf8Path) -> Result<()> {
+    let status = Command::new("rustfmt").arg(path).status().map_err(|e| {
+        let ctx = match e.kind() {
+            ErrorKind::NotFound => "formatting was requested, but rustfmt was not found",
+            _ => "unknown error when calling rustfmt",
+        };
+        anyhow!(e).context(ctx)
```
bendk, line 453: Does this context get printed out when `rustfmt` fails?  It seems like you ignore it below, but I could be missing some anyhow magic.

bendk (changes requested):
Seems like a good idea to me, just had a small question about the error context.

bendk:
My feeling is that formatting is a nice to have, but not at all important.  I would consider just printing a warning if `rustfmt` fails.  In fact, I think that's how the bindings generation works.

uniffi_bindgen/src/lib.rs
jplatte: It's not ignored, it gets bubbled out of the function with `?` on the next line.

bendk:
This seems like a step forward, so let's merge.  I'd like at least one other UniFFI dev to chime in before changing the behavior, so I made #1320 for discussion.

bendk (approved):

PR #1321, opened 2022-08-29
Format scaffolding with library `prettyplease`

### LINKED ISSUE #1320: Change `rustfmt` errors to warnings?
We should consider turning errors when running `rustfmt` on the scaffolding code into warnings.  This would remove friction when `rustfmt` is not installed and it's how the we handle errors with bindings code formatters.

ja-cop:
Previously uniffi-bindgen used rustfmt from the command line, which
has more failure modes than using a library (see #1318). In addition,
prettyplease is explicitly geared towards formatting generated code.

Also fixes #1320, as the prettyplease formatter doesn't fail.

---

Note that `syn::parse_file` may error, but AFAICT, that would mean the generated scaffolding is syntactically incorrect. Maybe this should be a panic instead?

### DIFF
--- a/uniffi_bindgen/Cargo.toml
+++ b/uniffi_bindgen/Cargo.toml
@@ -26,8 +26,10 @@ goblin = "0.5"
 heck = "0.4"
 once_cell = "1.12"
 paste = "1.0"
+prettyplease = "0.1"
 serde = "1"
 serde_json = "1.0.80"
+syn = { version = "1", default-features = false, features = ["full", "parsing"] }
 toml = "0.5"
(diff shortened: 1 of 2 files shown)


mhammond:
Another alternative is to have uniffi not support formatting at all. IIUC, it's currently optional and defaults to not formatting - so we could remove that optional support entirely, then consumers of uniffi can make whatever choices make sense to them.

(I believe mozilla never formats scaffolding generated by uniffi, so taking a new unconditional dependency for this doesn't really appeal to me that much)

skhamis:
I'll take myself off review for now and looks like it'll rerequest when this goes back to review!

PR #1323, opened 2022-08-30
Drop the dependency on `enum_variant_macros` used for a single `From` implementation

badboy:
I feel like this dependency doesn't really carry its weight.


tarikeshaq (approved):
Makes sense 🚀

PR #1324, opened 2022-08-30
Cache an extracted item name after parsing it from the library file

badboy:
I had `uniffi-fixture-simple-fns` fail locally (M1 macbook) because it encountered every function twice. It seems the code was set up to avoid that, but never actually stored the name and so ran into this.


bendk (approved):
I thought I was being so safe by adding the HashMap and code to check it :facepalm:

Issue #1325, opened 2022-08-30
Bindings are initializing callback interfaces incorrectly.


bendk:
The `ffi_init_callback()` function only inputs a `ForeignCallback`, not a `RustCallStatus`.  For example, here's the output of `cargo expand` for the callbacks example:

```
#[doc(hidden)]
static FOREIGN_CALLBACK_ONCALLANSWERED_INTERNALS: uniffi::ForeignCallbackInternals =
    uniffi::ForeignCallbackInternals::new();
#[doc(hidden)]
#[no_mangle]
pub extern "C" fn ffi_callbacks_334c_OnCallAnswered_init_callback(
    callback: uniffi::ForeignCallback,
) {
    FOREIGN_CALLBACK_ONCALLANSWERED_INTERNALS.set_callback(callback);
}
```

However, [all](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceTemplate.swift#L90-L92) [our](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceTemplate.kt#L96-L98) [bindings](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/python/templates/CallbackInterfaceTemplate.py#L81) that support callback interfaces are calling it with a `RustCallStatus`.  I guess this works because the C calling convention ignores the extra params, and those `RustCallStatus` instances are initialized with `code=0`.  Still, it seems like a bad idea and we should fix the bindings code.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-192)
┆friendlyId: UNIFFI-192

badboy:
This was tackled by #1326, right?

PR #1326, opened 2022-08-30
Fix callback interface init signature #1325

bendk:
The bindings were sending a `RustCallbackStatus` out param that the
scaffolding didn't have.  This worked out okay in practice, but we
should have the signatures match.


tarikeshaq (approved):
LGTM 🚀

PR #1332, opened 2022-09-01
Clearly separate different parts of the exports macro

jplatte:
With this PR, serializing and emitting the generated metadata is no longer part of collecting the metadata and generating the scaffolding code, respectively.


bendk (approved):
Nice change, I agree this way is cleaner.

PR #1333, opened 2022-09-02
Support Arc<_> in the proc-macro frontend

jplatte:
Depends on #1332. This does not implement the scheme discussed before where we add (compile-time) assertions that the type inside `Arc` is equal to a type of the same name in `crate::uniffi_types`, because when I thought it through again, I noticed a few issues with that:

* We would have to add the same assertion in UDL-generated code, which is a breaking change for UDL users
* With multi-crate setups (that are actually otherwise really straight-forward with the proc-macro frontend), this check would be insufficient and you could still easily mix up different types of the same (local) name

The best solution I've been able to come up with so far is instead doing a runtime assertion, that is keep a global map of the type name as used in source code to the `TypeId`, and panic in case we encounter two types of the same name with different type IDs.

Does that sound good? If yes, I'll go ahead and implement it.


bendk:
Could you expand on the issue with the compile time assertion more?

>    We would have to add the same assertion in UDL-generated code, which is a breaking change for UDL users
Would this be a breaking change?  The UDL-generated code uses whatever types are in the root module.  It seems like it could create a `uniffi_types` module and import the types there.
 
>   With multi-crate setups (that are actually otherwise really straight-forward with the proc-macro frontend), this check would be insufficient and you could still easily mix up different types of the same (local) name

I think this use case is definitely important, but I don't see how things could get mixed up.  If my crate and the remote crate have types with the same names, I assume I would need to rename the remote type when I imported it.  At that point it seems like your scheme should work, but I could easily be missing something here.

bendk (approved):
Wow this was surprisingly easy.   The changes look good to me, do you want to merge now or wait until the issue with type names is figured out?

jplatte:
Here's a basic example of how you could cause unsoundness with `Arc`. As implemented, it doesn't even need multiple crates, but this is a version that isn't fixable at compile time:

```rust
// crate a
mod uniffi_types {
    use u64 as Foo;
}

#[uniffi::expand]
pub fn rs_print(num: Arc<Foo>) {
    println!("{num}");
}
```

```rust
// crate b
mod uniffi_types {
    use &'static str as Foo;
}

#[uniffi::expand]
pub fn rs_value() -> Arc<Foo> {
    Arc::new("Hello, world!")
}
```

```python
# import both functions from a foreign language, then
rs_print(rs_value()) # 💥
```

bendk:
Ahh yes that would be a problem and I'm not sure that just adding a type check on the Rust side would fix it.  The other issue is that when generating the bindings, this would generate 2 classes named `Foo` which would almost certainly be a compile error.  I think the same issue would happen with 2 functions that had the same name.

I thought about this in a slightly different context but forgot to make an issue for it.  I just wrote up #1334 with my thoughts.

Basically I think that when generating a module in a foreign language, we should only pay attention to items from a single Rust crate.  That crate could import items from another crate and expose them to UniFFI, but then it would need to make sure that the names are unique.

jplatte:
Soo, I think this is ready for review now. The failure modes are pretty horrible unfortunately, if the empty interface definition in the UDL is missing the generation of the bindings fails with

```

**Comment by jplatte** [comment]
As a reference, here is what the export macro generates on the last function, as rendered by rust-analyzer's "expand macro recursively" assist:

```rust
fn set_contains(set: Arc<MyHashSet>, value: String) -> bool {
    set.lock().unwrap().contains(&value)
}
#[doc(hidden)]
#[no_mangle]
pub extern "C" fn _uniffi_uniffi_simple_fns_set_contains_e318(
    arg0: <Arc<MyHashSet> as ::uniffi::FfiConverter>::FfiType,
    arg1: <String as ::uniffi::FfiConverter>::FfiType,
    call_status: &mut ::uniffi::RustCallStatus,
) -> <bool as ::uniffi::FfiConverter>::FfiType {
    ::uniffi::deps::log::debug!("set_contains");
    ::uniffi::call_with_output(call_status, || {
        <bool as ::uniffi::FfiConverter>::lower(set_contains(
            <Arc<MyHashSet> as ::uniffi::FfiConverter>::try_lift(arg0)
                .unwrap_or_else(|err| ::std::panic!("Failed to convert arg 'set': {}", err)),
            <String as ::uniffi::FfiConverter>::try_lift(arg1)
                .unwrap_or_else(|err| ::std::panic!("Failed to convert arg 'value': {}", err)),
        ))
    })
}
::uniffi::deps::static_assertions::assert_type_eq_all!(bool, ::std::primitive::bool);
::uniffi::deps::static_assertions::assert_type_eq_all!(String, ::std::string::String);
::uniffi::deps::static_assertions::assert_type_eq_all!(
    Arc<MyHashSet>,
    ::std::sync::Arc<crate::uniffi_types::MyHashSet>
);
#[no_mangle]
pub static UNIFFI_META_set_contains: [u8; 119usize] = [
    0u8, 0u8, 0u8, 0u8, 1u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 17u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8,
    0u8, 117u8, 110u8, 105u8, 102u8, 102u8, 105u8, 95u8, 115u8, 105u8, 109u8, 112u8, 108u8, 101u8,
    95u8, 102u8, 110u8, 115u8, 12u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 115u8, 101u8, 116u8, 95u8,
    99u8, 111u8, 110u8, 116u8, 97u8, 105u8, 110u8, 115u8, 2u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8,
    3u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 115u8, 101u8, 116u8, 15u8, 0u8, 0u8, 0u8, 9u8, 0u8,
    0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 77u8, 121u8, 72u8, 97u8, 115u8, 104u8, 83u8, 101u8, 116u8, 5u8,
    0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 118u8, 97u8, 108u8, 117u8, 101u8, 11u8, 0u8, 0u8, 0u8, 1u8,
    10u8, 0u8, 0u8, 0u8,
];
```

uniffi_bindgen/src/macro_metadata/ci.rs
```diff
@@ -20,7 +20,17 @@ pub fn add_to_ci(
...
+                let cr = meta.module_path.first().unwrap();
+                let ns = iface.namespace();
+                if cr == ns {
+                    iface.add_function_definition(meta.into())?;
+                } else {
+                    eprintln!("Mixing symbols from multiple crates is not supported yet.");
+                    eprintln!(
+                        "Main crate is expected to be named `{ns}` based on the UDL namespace."
+                    );
+                    eprintln!("Ignoring function `{}` from crate `{cr}`.", meta.name);
```
mhammond, line 32: is there a reason to ignore this rather than bailing?

mhammond (approved):
You'll have to excuse my ignorance here because I'm not as familiar with the macro code as I should be.

> Soo, I think this is ready for review now. The failure modes are pretty horrible unfortunately, if the empty interface definition in the UDL is missing the generation of the bindings fails with

This is only an issue for the (hopefully!) transitional period where you must combine macros with a UDL, right? Even when that time passes, I'm guessing there is likely to always be a world where you *can* combine them though, so ultimately we probably want to make that failure mode better? (To be clear, I think that's certainly fine for now, but I'm trying to understand what the future might look like).

> and one can still cause unsoundness with this by having a different type of the same name visible in `uniffi_types` vs. the place that `include!`s the scaffolding code and defining functions operating on both these types in UDL, and proc-macro land. I think maybe that's okay?

I don't fully understand this - but this doesn't even seem solvable by #1334? Do you see a solution?

Ultimately though, this issue seems more about the entire macro model than about this specific patch, and I agree you certainly need `Arc<>` support. But can we maybe open an issue or discussion about this?

uniffi_bindgen/src/macro_metadata/ci.rs
jplatte: Hm, no. I guess we can error out.

jplatte:
> This is only an issue for the (hopefully!) transitional period where you must combine macros with a UDL, right?

Yes. We could probably even have presence of `Arc<Foo>` in the proc-macro input imply existence of an empty Foo interface, then merge that with what we fjnd elsewhere. 

> Even when that time passes, I'm guessing there is likely to always be a world where you *can* combine them though, so ultimately we probably want to make that failure mode better? (To be clear, I think that's certainly fine for now, but I'm trying to understand what the future might look like).

There might be a time when everything UDL will be removed. I don't think it makes sense to remove interoperability before that. 

> > and one can still cause unsoundness with this by having a different type of the same name visible in `uniffi_types` vs. the place that `include!`s the scaffolding code and defining functions operating on both these types in UDL, and proc-macro land. I think maybe that's okay?
> 
> I don't fully understand this - but this doesn't even seem solvable by #1334? Do you see a solution?

We could require users to use some macro for defining the `uniffi_types` module that also takes care of udl scaffolding inclusion, but that's a breaking change. We could also change the UDL scaffolding to use the uniffi_types module, another breaking change. I think once we document proc-macros and encourage users to switch to them, it makes sense to do something like that, but not right now. 

> Ultimately though, this issue seems more about the entire macro model than about this specific patch, and I agree you certainly need `Arc<>` support. But can we maybe open an issue or discussion about this?

Sure, I can open an issue later.

mhammond:
> There might be a time when everything UDL will be removed. I don't think it makes sense to remove interoperability before that.

That's my point - there *might* be a time when UDL will be removed, but that might end up being in the (relatively) far future - so this failure mode might need something better than a panic unwrapping None before that time?

jplatte:
Yes, it certainly does. But maybe not now, while the proc-macro isn't even documented at all? 😅

It's probably not going to be super hard though, so if you want to have it before being this PR, I'll look into it.

mhammond:
No, I don't think that's necessary, as I said, I was really just trying to understand what this future looks like.

PR #1337, opened 2022-09-12
Ignore clippy errors

bendk:
Surround the scaffolding code with `#[allow(clippy::all)]` to avoid clippy errors.  I saw new clippy errors when upgrading Rust to 1.63 for application-services and it seems good to just ignore all of them.


bendk:
Hmm, looks like this approach is not going to work.  I'm going to work on fixing the actual clippy errors instead.

PR #1342, opened 2022-09-13
Refactored the throws code

bendk:
Added a bool function that returns if a function throws or not. Named this one "throws" and renamed the old method to "throws_name". We could almost get rid of that one, but it's still used by the Ruby code

This is just a bit of prep work for getting callback interface errors working.

### DIFF
--- a/uniffi_bindgen/src/interface/function.rs
+++ b/uniffi_bindgen/src/interface/function.rs
@@ -81,7 +81,11 @@ impl Function {
         &self.ffi_func
     }
 
-    pub fn throws(&self) -> Option<&str> {
+    pub fn throws(&self) -> bool {
+        self.attributes.get_throws_err().is_some()
+    }
+
+    pub fn throws_name(&self) -> Option<&str> {
         self.attributes.get_throws_err()
     }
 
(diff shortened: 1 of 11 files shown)


bendk:
> This looks fine to me, but should we also remove the `exception_name` filter everywhere? (Python defines one too but doesn't seem to use it)

Good point, I removed those for Python/Swift.  Kotlin is still using it though.

uniffi_bindgen/src/interface/function.rs
```diff
@@ -81,7 +81,11 @@ impl Function {
         &self.ffi_func
     }
 
-    pub fn throws(&self) -> Option<&str> {
+    pub fn throws(&self) -> bool {
```
tarikeshaq, line 84: is this a breaking change? since `Function` (and I guess `Method` and `Constructor`) are exposed publically from `uniffi_bindgen::interface` and we are changing  the signature of the `throws` function - should we have changelog entry to represent that?

tarikeshaq (changes requested):
Makes sense to me overall, I had a question about breaking changes, but otherwise 💯

uniffi_bindgen/src/interface/function.rs
bendk: Good point, I just updated the changelog.

tarikeshaq (approved):
🚢

PR #1344, opened 2022-09-14
Callback exceptions

bendk:
This is a work-in-progress for handling exceptions in callback interfaces.  It changes a few things:

- The `Throws` attribute is now offecially supported on CallbackInterface methods, allowing callback interfaces to specify which errors they will throw.  If we catch one of those exceptions, we serialize it to the `RustBuffer` that's normally used for the return value and send it back to Rust.
- On the Rust side, this means the callback interface trait methods need to change to return `Result<T>` instead of `T`.
- For now, `Throws` is optional and we fall back on the old behavior if it's not present.  But this is deprecated.
- We also handle unexpected exceptions in callback interface.  If we see one of those, we stringify it and serialize it to the `RustBuffer`.  This gets converted into a `uniffi::UnexpectedCallbackError` in Rust.  Users need to write a `From<>` impl for that struct to their error enum (unless they're on the old/deprecated behavior).
- I updated `callback_interfaces.md` to describe the new system, updated the callbacks example, and implemented it on Python/Kotlin.
- I made a slight change to the FFI return values for callback interface call.  See my [comment below](https://github.com/mozilla/uniffi-rs/pull/1344#discussion_r972681969) for details.

### DIFF
--- a/docs/manual/src/udl/callback_interfaces.md
+++ b/docs/manual/src/udl/callback_interfaces.md
@@ -18,10 +18,12 @@ by the host operating system (e.g. the key chain).
 
 ```rust,no_run
 pub trait Keychain: Send + Sync + Debug {
-  fn get(&self, key: String) -> Option<String>;
-  fn put(&self, key: String, value: String);
+  fn get(&self, key: String) -> Result<Option<String>, KeyChainError>
+  fn put(&self, key: String, value: String) -> Result<(), KeyChainError>
 }
 ```
+
+### Why + Send + Sync?
 
 The concrete types that UniFFI generates for callback interfaces implement `Send`, `Sync`, and `Debug`, so it's safe to
 include these as supertraits of your callback interface trait.  This isn't strictly necessary, but it's often useful.  In
@@ -30,16 +32,44 @@ particular, `Sync + Send` is useful when:
     interface type)
   - Storing `Box<dyn CallbackInterfaceTrait>` inside a global `Mutex` or `RwLock`
 
-## 2. Define a callback interface in the UDL.
+
+## 2. Setup error handling
+
+All methods of the Rust trait should return a Result.  The error half of that result must
+be an [error type defined in the UDL](./errors.md).
+
+It's currently allowed for callback interface methods to return a regular value
+rather than a `Result<>`.  However, this is means that any exception from the
+foreign bindings will lead to a panic.
+
+### Extra requirements for errors used in callback interfaces
+
+In order to support errors in callback interfaces, UniFFI must be able to
+properly [lift the error](../internals/lifting_and_lowering.md).  This means
+that the if the error is described by an `enum` rather than an `interface` in
+the UDL (see (./errors.md)) then all variants of the Rust enum must be unit variants.
+
+In addition to expected errors, a callback interface call can result in all kinds of
+unexpected errors.  Some examples are the foreign code throws an exception that's not part
+of the exception type or there was a problem marshalling the data for the call.  UniFFI
+uses `uniffi::UnexpectedUniFFICallbackError` for these cases.  Your code must include a
+`From<uniffi::UnexpectedUniFFICallbackError>` impl for your error type to handle those or
+the UniFFI scaffolding code will fail to compile.  See the `example/callbacks` for an
+exapmle of how to do this.
+
+## 3. Define a callback interface in the UDL.
 
 ```webidl
 callback interface Keychain {
+    [Throws=KeyChainError]
     string? get(string key);
+
+    [Throws=KeyChainError]
     void put(string key, string data);
 };
 ```
 
-## 3. And allow it to be passed into Rust.
+## 4. And allow it to be passed into Rust.
 
 Here, we define a constructor to pass the keychain to rust, and then another method
 which may use it.
--- a/examples/callbacks/src/lib.rs
+++ b/examples/callbacks/src/lib.rs
@@ -2,25 +2,35 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-pub trait OnCallAnswered {
-    fn hello(&self) -> String;
-    fn busy(&self);
-    fn text_received(&self, text: String);
+#[derive(Debug, thiserror::Error)]
+pub enum TelephoneError {
+    #[error("Busy")]
+    Busy,
+    #[error("InternalTelephoneError")]
+    InternalTelephoneError,
 }
 
-#[derive(Debug, Clone)]
-struct Telephone;
-impl Telephone {
-    fn new() -> Self {
-        Telephone
+// Need to implement this From<> impl in order to handle unexpected callback errors.  See the
+// Callback Interfaces section of the handbook for more info.
+impl From<uniffi::UnexpectedUniFFICallbackError> for TelephoneError {
+    fn from(_: uniffi::UnexpectedUniFFICallbackError) -> Self {
+        Self::InternalTelephoneError
     }
-    fn call(&self, domestic: bool, call_responder: Box<dyn OnCallAnswered>) {
-        if domestic {
-            let _ = call_responder.hello();
-        } else {
-            call_responder.busy();
-            call_responder.text_received("Not now, I'm on another call!".into());
-        }
+}
+
+pub trait CallAnswerer {
+    fn answer(&self) -> Result<String, TelephoneError>;
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct Telephone;
+impl Telephone {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn call(&self, answerer: Box<dyn CallAnswerer>) -> Result<String, TelephoneError> {
+        answerer.answer()
     }
 }
 
--- a/uniffi/src/ffi/foreigncallbacks.rs
+++ b/uniffi/src/ffi/foreigncallbacks.rs
@@ -114,6 +114,7 @@
 //!
 
 use super::RustBuffer;
+use std::fmt;
 use std::os::raw::c_int;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
@@ -131,8 +132,12 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 ///   passing individual arguments to the user's callback.
 /// * `buf_ptr` is a pointer to where the resulting buffer will be written. UniFFI will allocate a
 ///   buffer to write the result into.
-/// * A callback returns `-1` if an unexpected error happens, `0` if the call succeeded without a
-///   return value or `1` if data was written into the output buffer.
+/// * A callback returns:
+///    - `-2` An error occured that was serialized to buf_ptr
+///    - `-1` An unexpected error ocurred
+///    - `0` is a deprecated way to signal that if the call succeeded, but did not modify buf_ptr
+///    - `1` If the call succeeded.  For non-void functions the return value should be serialized
+///      to buf_ptr.
 ///   Note: The output buffer might still contain 0 bytes of data.
 pub type ForeignCallback = unsafe extern "C" fn(
     handle: u64,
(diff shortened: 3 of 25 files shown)


examples/callbacks/src/lib.rs
```diff
@@ -2,25 +2,33 @@
...
-    fn text_received(&self, text: String);
+#[derive(Debug, thiserror::Error)]
+pub enum TelephoneError {
+    #[error("Busy")]
+    Busy,
+    #[error("InternalTelephoneError")]
+    InternalTelephoneError,
+}
+
+impl From<uniffi::UnexpectedUniFFICallbackError> for TelephoneError {
```
mhammond, line 13: I think a comment here - even just to point at the doc changes you made - might help people unfamiliar with uniffi and checking out the examples.

uniffi/src/ffi/foreigncallbacks.rs
```diff
@@ -131,8 +132,12 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 ///   passing individual arguments to the user's callback.
 /// * `buf_ptr` is a pointer to where the resulting buffer will be written. UniFFI will allocate a
 ///   buffer to write the result into.
-/// * A callback returns `-1` if an unexpected error happens, `0` if the call succeeded without a
-///   return value or `1` if data was written into the output buffer.
+/// * A callback returns:
+///    - `-2` An error occured that was serialized to buf_ptr
+///    - `-1` An unexpected error ocurred
+///    - `0` is a deprecated way to signal that if the call succeeded, but did not modify buf_ptr
```
mhammond, line 138: So I *think* this means that even a function returning `Ok(())` will modify `buf_ptr`, so this is deprecated because it can only happen for a function which returns a void not in a Result - which is deprecated. If that's correct it might be worth expanding this (or not - I'm really just trying to understand how this is "deprecated" from the POV of the user)

uniffi/src/ffi/foreigncallbacks.rs
```diff
@@ -131,8 +132,12 @@ use std::sync::atomic::{AtomicUsize, Ordering};
...
+///    - `1` If the call succeeded.  For non-void functions the return value should be serialized
```
mhammond, line 139: but this confuses me a little given the above - so maybe I'm wrong about what `Ok(())` does?

uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceTemplate.kt
```diff
@@ -34,11 +34,32 @@ internal class {{ foreign_callback }} : ForeignCallback {
...
+                    }
+                    {%- else %} 
+                    val buffer = this.{{ method_name }}(cb, args)
+                    // Success
+                    outBuf.setValue(buffer)
+                    1
+                    {%- endmatch %}
+                } catch (e: Throwable) {
+                    // Unexpected error
+                    -1
```
mhammond, line 61: it's a little sad we can't optionally exact some data from the unexpected error if the bindings allow it. Or if we think that might be a possibility in the future, could the `Unexpected` error struct hold an `Option<String>` to make that future more tractable?

(I don't really care about that, just an idle thought)

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -1,19 +1,23 @@
...
+class UniFFIExceptionTmpNamespace:
+    class {{ type_name }}(Exception):
         pass
-    {%- endfor %}
+    {% for variant in e.variants() %}
+    {%- let variant_type_name = variant.name()|class_name %}
 
+    {%- if e.is_flat() %}
+    class {{ variant_type_name }}({{ type_name }}):
+        def __unicode__(self):
```
mhammond, line 17: I thought this was `__str__` in python 3? https://docs.python.org/3/reference/datamodel.html#object.__str__ doesn't mention `__unicode__` and a quick test:

```
class T:
#  def __str__(self):
#    return "str"
  def __unicode__(self):
    return "uni"

t = T()
print(t)
```

As shown, that prints `<__main__.T object at 0x00000156BC9CA140>` but if `__str__` is uncommented, `str` is printed

mhammond (commented):
I like this alot! I skimmed over some of the implementation details but think this is great and definitely what we should pursue.

bendk:
I added Swift support and also support for serializing error strings for unexpected errors.  I'm thinking this is pretty much what we want.

There's one last issue.  This change requires implementing `FfiConverter.read()` in Rust for errors.  However, this is not backwards compatible, since we currently allow errors to always be specified as "flat" even if they have fields.  Todolist is an example of this.  This works fine now, since we never need to construct the error in Rust, but once we try to implement `read()` we get compile errors.  Even worse, Todolist uses this with tuple-style fields which we don't have a syntax to support and I'm guessing it's not the only crate to do this.  I think there are a couple ways forward, which path do others think is best?

- Forcing users to switch to the fielded-style of errors, which probably means we need to add support for tuple-style fields.
- Only try to implement the `read()` method if the error is used in a callback interface and trying to solve the problem later.

bendk:
After thinking about this one more, I definitely think the second bullet is correct for now at least.  It's pretty easy to do and there's no need to tie this PR to deprecating an error style that I believe we currently use in production.  I updated the code, added a note about this in the suggested doc changes here, and made #1348 for the future work.

I removed the WIP label I think this one is ready for final review.

docs/manual/src/udl/callback_interfaces.md
```diff
@@ -18,28 +18,56 @@ by the host operating system (e.g. the key chain).
...
+It's also, currently allowed for callback interface methods to return a regular
+value rather than a `Result<>`.  However, this is deprecated and we plan to
+remove support at some point.
+
+### Extra requirements for errors used in callback interfaces
+
+In order to support errors in callback interfaces, UniFFI must be able to
+properly [lift the error](../internals/lifting_and_lowering.md).  This means
+that the if the error is described by an `enum` in the UDL it must be truely
+fieldless --  all variants must be unit variants.
```
mhammond, line 50: I don't understand this para - (a) you seem to be saying that if a callback interface throws an error, all variants of that error must be fieldless, but that's not correct is it? I'd expect maybe just the "unexpected" variant should be fieldless? (b) but even that's not true - below you have a callback interface throwing `UnexpectedErrorWithReason` and that takes a string.

ISTM the error variant can be anything at all, but the practical limitation is just that it must be createable given just a `uniffi::UnexpectedUniFFICallbackError`?

(Actually, now I get into the impl, I think I better understand - "truely" is doing a lot of heavy lifting here :) So maybe reword that to better convey that the Rust implementation of the error must not have any fields. Looking at the implementation, it's also not clear what happens if that's not true - should we have a uitest for this scenario?

uniffi/src/ffi/foreigncallbacks.rs
```diff
@@ -195,3 +200,26 @@ impl ForeignCallbackInternals {
...
+
+impl UnexpectedUniFFICallbackError {
+    pub fn from_reason(reason: String) -> Self {
+        Self { reason }
+    }
+}
+
+impl fmt::Display for UnexpectedUniFFICallbackError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "UnexpectedUniFFICallbackError(reason: {})", self.reason)
```
mhammond, line 221: should reason be quoted here? Or probably just ":?".

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -76,12 +76,36 @@ public object {{ e|ffi_converter_name }} : FfiConverterRustBuffer<{{ type_name }
 
     @Suppress("UNUSED_PARAMETER")
```
mhammond, line 77: here and below, I think the `@Suppress` annotations can now die?

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -219,6 +219,25 @@ impl ComponentInterface {
         self.errors.iter().find(|e| e.name == name)
     }
 
+    /// Should we generate read/lower functions for errors?
```
mhammond, line 222: This lower/read is from the POV of the foreign bindings, right? Maybe change this to start with "Should foreign bindings generate ..."?

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -219,6 +219,25 @@ impl ComponentInterface {
...
+    ///
+    /// This is a temporary measure since lower/write can't be generated for some errors,
```
mhammond, line 224: This is still slightly confusing to me (which is a comment on me, not on this patch :) But in what way is this temporary?

uniffi_bindgen/src/scaffolding/templates/CallbackInterfaceTemplate.rs
```diff
@@ -69,31 +71,82 @@ impl r#{{ trait_name }} for {{ trait_impl }} {
...
+                    {% when Some(error_type) -%}
+                    let vec = ret_rbuf.destroy_into_vec();
+                    let mut ret_buf = vec.as_slice();
+                    Err({{ error_type|ffi_converter }}::try_read(&mut ret_buf).unwrap())
+                    {%- else -%}
+                    panic!("Callback return -2, but throws_type() is None");
+                    {%- endmatch %}
+                }
+                // 0 is a deprecated method to indicates success for void returns
+                0 => {
```
mhammond, line 115: IIUC, we only expect to see this for bindings outside of this crate (and maybe Ruby 😅 ) - so how about arranging for a deprecated warning here? (Deprecating and finally removing things without generating a warning seems very likely to catch people by surprise later)

mhammond (changes requested):
This is looking good to me. If we really are deprecating non-result functions in callbacks, I'd really like to see a deprecation warning issued (which probably means this should be a breaking change, which I think is fine) and *possibly* a uitest. OTOH, I guess we could soften this down to a recommendation (ie, "if you want to avoid panics, you should..."). TBH I don't care too much as I hope we can deprecate callback interfaces entirely later, but it's too soon for me to be confident that's going to be possible.

(I'll also try and come back to this to better understand the semantics around error fields, but I don't see that blocking anything here)

uniffi_bindgen/src/interface/mod.rs
mhammond: oh, no, it's for the scaffolding - so should "lower" be "lift"?

docs/manual/src/udl/callback_interfaces.md
bendk: I think you might have figured this out, but I wanted to write it out for posterity.  The issue is using the enum-style definition in the UDL, which doesn't list the fields, when in fact the variants do have fields.  It's now an issue because we have to generate read/lift methods for the scaffolding, so UniFFI needs to know how to construct the variants.

I updated the docs, hopefully it's a bit more clear.  A uitest is a great idea, I put one together and tried to make the error messages as obvious as a could -- although they're still not that obvious.  AFAIK, there's no way to do a static assertion for this and have it print out a nice error message.  If anyone knows one, please share.

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
bendk: Yup.  Also, even though the param isn't used for flat enums the `@Supress` is still not needed since it's overriding an interface method.

uniffi_bindgen/src/interface/mod.rs
bendk: I think that was optimistic on my part.  Once we implement `#1348`, then we can tell users that they shouldn't use the `enum` style definitions for errors if their errors actually have fields, and at some point we can drop support for it.  Maybe a simpler road would be to fully implement proc-macros and drop support for UDL.  But both of those things seems far off so I changed the wording.

uniffi_bindgen/src/scaffolding/templates/CallbackInterfaceTemplate.rs
bendk: Good call.  I added an `eprintln!` for this, is that the best way to signal a warning?

mhammond: A mildly better way might be to have a noop function marked `#[deprecated] ` and call it here, but I'm really not too bothered

mhammond (approved):

PR #1346, opened 2022-09-20
Add Object derive macro

jplatte:
… that generates a free function for an object/interface type.

Part of  #1257.

### DIFF
--- a/uniffi_bindgen/src/interface/mod.rs
+++ b/uniffi_bindgen/src/interface/mod.rs
@@ -77,7 +77,7 @@ pub use record::{Field, Record};
 
 pub mod ffi;
 pub use ffi::{FFIArgument, FFIFunction, FFIType};
-use uniffi_meta::MethodMetadata;
+use uniffi_meta::{MethodMetadata, ObjectMetadata};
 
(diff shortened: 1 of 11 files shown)


uniffi_bindgen/src/interface/mod.rs
```diff
@@ -614,6 +613,17 @@ impl Hash for ComponentInterface {
     }
 }
 
+fn get_or_insert_object<'a>(objects: &'a mut Vec<Object>, name: &str) -> &'a mut Object {
+    // The find-based way of writing this currently runs into a borrock error, so we use position
```
bendk, line 617:
```suggestion
    // The find-based way of writing this currently runs into a borrow error, so we use position
```

bendk (approved):
Looks good to me, I just had a minor spelling suggestion.

uniffi_bindgen/src/interface/mod.rs
jplatte: It was supposed to be "borrowck", but I guess I should expand that to "borrow checker" 😄

jplatte:
@bendk Can you re-review / merge? :)

bendk:
No problem, the code looks great.

PR #1347, opened 2022-09-21
proc-macro: Support Self type in methods

jplatte:
Part of #1257.

### DIFF
--- a/uniffi_macros/src/export.rs
+++ b/uniffi_macros/src/export.rs
@@ -117,7 +117,10 @@ fn fn_type_assertions(sig: &syn::Signature) -> TokenStream {
 
     let input_types = sig.inputs.iter().filter_map(|input| match input {
         syn::FnArg::Receiver(_) => None,
-        syn::FnArg::Typed(pat_ty) => Some(&pat_ty.ty),
+        syn::FnArg::Typed(pat_ty) => match &*pat_ty.pat {
+            syn::Pat::Ident(i) if i.ident == "self" => None,
+            _ => Some(&pat_ty.ty),
+        },
     });
     let output_type = match &sig.output {
         syn::ReturnType::Default => None,
--- a/uniffi_macros/src/export.rs
+++ b/uniffi_macros/src/export.rs
@@ -53,28 +53,28 @@ pub fn expand_export(metadata: ExportItem, mod_path: &[String]) -> TokenStream {
             self_ident,
         } => {
             let method_tokens: TokenStream = methods
-            .into_iter()
-            .map(|res| {
-                res.map_or_else(
-                    syn::Error::into_compile_error,
-                    |Method { item, metadata }| {
-                        let checksum = checksum(&metadata);
-                        let scaffolding =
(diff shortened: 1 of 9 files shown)


bendk (changes requested):
This looks good to me, but can you add some doc strings explaining what's going on in the `uniffi_macros` code.  I think a quick note about why each piece is needed will help a lot in the long-term.

jplatte:
Do you also want docs for the last commit? That code will have to change a bunch to not filter out `self` parameters once we support `static` methods. Currently the metadata just can't represent those, all methods just implicitly have a `self` parameter.

uniffi_macros/src/lib.rs
```diff
@@ -26,7 +27,14 @@ pub fn export(_attr: TokenStream, input: TokenStream) -> TokenStream {
 
     let gen_output = || {
         let mod_path = util::mod_path()?;
-        let item = syn::parse(input)?;
+        let mut item = syn::parse(input)?;
+
+        // If the input is an `impl` block, rewrite any uses of the `Self` type
+        // alias to the actual type, so we don't have to special-case it in the
+        // metadata collection or scaffolding code generation (which generates
+        // new functions outside of the `impl`).
```
bendk, line 35: This is great.

uniffi_macros/src/export.rs
```diff
@@ -117,7 +117,10 @@ fn fn_type_assertions(sig: &syn::Signature) -> TokenStream {
 
     let input_types = sig.inputs.iter().filter_map(|input| match input {
         syn::FnArg::Receiver(_) => None,
-        syn::FnArg::Typed(pat_ty) => Some(&pat_ty.ty),
+        syn::FnArg::Typed(pat_ty) => match &*pat_ty.pat {
+            syn::Pat::Ident(i) if i.ident == "self" => None,
```
bendk, line 121: How about a comment here about why we don't need a type assertion for `self` params.  (Or should this be eventually implemented?).

uniffi_macros/src/export/metadata/convert.rs
```diff
@@ -21,6 +21,11 @@ pub(super) fn fn_param_metadata(
                         syn::Pat::Ident(pat_id) => pat_id.ident.to_string(),
                         _ => unimplemented!(),
                     };
+
+                    if name == "self" {
+                        return None;
+                    }
```
bendk, line 27: And a comment here about why we don't output any metadata since there's an implicit `self`.

bendk:
Github is acting pretty weird for me, but hopefully my comments make sense.  I don't think this is all that critical, it just seems like it would help keep track of things.

uniffi_macros/src/export.rs
jplatte: Hm, I think a type assertion might be generated for the whole impl block's self type. I will have to look at the code again.

jplatte: Oh, actually we don't have the type assertions at all in impl blocks / methods 😅

bendk (approved):
The docs look great, I'll merge this one now

PR #1350, opened 2022-09-23
Add Swift bindings generation support for reserved words

notmandatory:
This change is based on #1237 but isn't quite as complete. I wasn't able to get reserved words working as a pub fn in the namespace. I need this fix to to expose an API in a project I'm working on that is using a Swift reserved word (`internal`) as an enum variant name. 

If I can get this PR ready to merge I'm also happy to do a corresponding change for the master branch. If you'd rather I do the master branch change first I can do it that way, I started with 0.19 because that the version of uniffi my project is currently using.

### DIFF
--- a/fixtures/keywords/README.md
+++ b/fixtures/keywords/README.md
@@ -5,10 +5,10 @@ Each directory is a specialized fixture for each language (although the
 `rust` directory does double-duty for Python)
 
 The reason we don't try and combine these into a single fixture is that we'd
-like a highe degree is assurance that each language has a keyword used
+like a high degree of assurance that each language has a keyword used
 in every possible context. By trying to combine them, we would end up needing
 multiple, say, `enum`, `interface`, function arguments, variant discriminators,
-etc. So there's a reasonable change we'd accidently not have an `enum` with a
+etc. So there's a reasonable change we'd accidentally not have an `enum` with a
 (say) kotlin keyword.
 
 Separate fixtures means someone familiar with Kotlin and look at 1 fixture and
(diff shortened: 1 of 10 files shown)


notmandatory:
PS. thanks for all the amazing work the uniffi-rs team is doing. ❤️

fixtures/keywords/README.md
```diff
@@ -5,10 +5,10 @@ Each directory is a specialized fixture for each language (although the
 `rust` directory does double-duty for Python)
 
 The reason we don't try and combine these into a single fixture is that we'd
-like a highe degree is assurance that each language has a keyword used
+like a high degree of assurance that each language has a keyword used
```
bendk, line 8: +1 to making these small improvements as you work on the code.

bendk (changes requested):
This seems right to me, I'll merge it and I think I can merge the `0.19.x` branch back in to main soon, I don't think you need to make a PR for that.  Do you need a new `0.19.x` dot release?

notmandatory:
> This seems right to me, I'll merge it and I think I can merge the `0.19.x` branch back in to main soon, I don't think you need to make a PR for that. Do you need a new `0.19.x` dot release?

Thanks for the merge! our `bdk-ffi` project could use a `0.19.x` dot release, but it's not urgent since I was able to get our last Swift release out using my fork.

thunderbiscuit:
I just opened the PR for the upgrade to `0.20.0` and everything seems to be working out of the box on our side. I think we should be able to just use that for our next release coming up in two weeks, so no need to do a `0.19.7` patch release.

Thanks a bunch!

bitcoindevkit/bdk-ffi#203

notmandatory:
I was able to successfully get our `bdk-ffi` project working with the uniffi version 0.21.0. I also added some integration tests with you very useful `build_foreign_language_testcases` macro, it was very easy to use and will help us catch problems much sooner. Thanks again for the awesome project and the new release!

PR #1353, opened 2022-10-01
[Discuss] IR for bindings generation

bendk:
I've been trying to brew this one up for a while and I now feel like it's in a place that's worth sharing.  This PR contains the outline of what it could look like to use an intermediate representation for bindings generation.  Credit to @jhugman for suggesting this a while ago.

The Big Idea 🌈™ here is to we define something like an [AST for a language specially tailored for bindings generation](https://github.com/mozilla/uniffi-rs/tree/99ad06c227ed62fc012ab0a86b4c5b7421080e39/bindings-ir/src/ir).  `uniffi-bindgen` transforms the `ComponentInterface` into an AST with the bindings then a language-specific bindings-generate transforms that AST into actual code.  As an example, [uniffi-bindgen transforms object definitions from the UDL into class definitions in the IR](https://github.com/mozilla/uniffi-rs/blob/99ad06c227ed62fc012ab0a86b4c5b7421080e39/uniffi_bindgen/src/bindings/ir/objects.rs) then [bindings-ir-kotlin transforms those class definitions into actual Kotlin classes.](https://github.com/mozilla/uniffi-rs/blob/99ad06c227ed62fc012ab0a86b4c5b7421080e39/bindings-ir-kotlin/src/templates/ClassDef.kt).  The IR is all Serde-serializable and the plan would be to serialize it in the library file, like we do with the proc-macro metadata now.

The current state is an IR system with a Kotlin implementation, and a `uniffi-bindgen` implementation that handles functions, objects, records, and errors.  I got the `geometry`, `arithmetic`, and `sprites` example tests to pass.  It definitely needs a lot more work, but hopefully enough is there that we can talk about the idea concretely.

## Benefits

- **Sharing code**.  There's a large overlap between UniFFI and diplomat.  I just learned about [wpgu](https://github.com/gfx-rs/wgpu-native#bindings) which is also written in Rust with it's own bindings generation system.  In the mozilla app-services circle, we also have Glean and Nimbul-FML generating bindings for their own language.  Each of these projects targets different problems and I'm doubtful that we could merge them all into one big bindings generator, but maybe we could use a shared IR.  That would let us pool our langugage-specific bindings generators and all get access to the languages that the other projects support.  It also seems very likely that we will want to work on new projects that want to generate code and they could use this system to give themselves a huge jump-start.
- **Fixes a lot of versioning issues**.  Our versioning story is pretty sad right now, especially with regards to external bindings and Desktop JS.  In order to land a UniFFI upgrade on Desktop we need to first need to update Glean to use the new version, plus all the app-services components, make a release for both, then vendor all of those upgrades in.  This new system would fix a lot of these problems.  We would be compiling the bindings IR in to the library file, so there's no chance of an FFI versioning mismatch.  The language bindings generators would depend on the IR code, not `uniffi-bindgen` directly, which means we updates to `uniffi-bindgen` wouldn't always need to be breaking changes.  In the new world, Glean and app-services, could leave the patch version unspecified in their `Cargo.toml` files, and we could ship an update to UniFFI, even one that changes the FFI contract, without having to touch Glean/app-services.
- **Makes it easier to change UniFFI**.  Right now any change to UniFFI means updating the bindings for the 4.5 languages we officially support.  If we get external bindings contributions, we would need to take them into consideration as well.  In the new world, we could update how UniFFI generates the IR, then we instantly get the new functionality on all supported languages.
- **IR transforms**.  It would be possible to transform the IR that UniFFI outputs before the bindings generator sees it, which could lead to some nice features.  The first one that comes to mind how we generate a lot of duplicated code for each app-services component (FFIConverters, RustBuffer functionality, etc.), then package it all together in the megazord.  This wastes space/memory and also makes things like external bindings trickier to implement.  We could fix this by having a transform step, where we extract that duplicated code, put it in a shared module, then have each component import from that module.  Another possibility would be to implement things features that extend the base UniFFI functionality like decorators.
- **Safety checks**.  One thing that Nika pointed out with the Desktop JS project is that we're kind of loosy-goosy with our pointers.  Pointers aren't typed so you could potentially use the wrong pointer when calling a method and each language has its own ad-hoc methods to prevent use-after-free bugs.  This IR makes each pointer specify a type and each function call specifies if it's borrowing or consuming the pointer.  At this point there's no enforcement mechanism, but we could add static analysis that ensures generated IR code follow the rules.
- **Improves external bindings**.  Instead of having to understand and implement things like how UniFFI lowers and lifts pointers, external bindings authors would only need to translate the IR to their language.  In order to implement Kotlin, I wrote a basic test suite of IR code which I think could be a nice way for bindings authors to implement things step-by-step.  Bindings authors would have to care much less about UniFFI changes.  I'm hoping that most of the time they could ignore them.  If other projects adopt this, then it's also a much better value proposition for external bindings authors.

## Risks

- **Potential time sink**.  It will take a lot of work to get this to the finish line we need to make sure that it's worth it.  @mhammond points out that this would be a good time to take a step back and make sure that the overall strategy is sound.  Maybe one of those other bindings generators actually has nailed it and we should just join them.  Is there prior art with an FFI IR that we could use?  Are we really going to be adding many more features to UniFFI and/or starting new projects that want to generate bindings, or are we close to feature complete?
- **Not clear that we really get the benefits**.  The benefits statements above are dependent on a few things.  First off, is this IR flexible enough to be implemented in other languages?  I'm hopeful but unsure, especially with Desktop JS.  Secondly, can we keep the IR relatively stable?  A lot of the benefits depend on us being able to add features to UniFFI without needing to add a new expression/statement/definition or whatever to the IR.  If that's not true, then versioning story gets worse, teaming up with the other teams is going to require a lot more maintenance, etc..
- **The IR generation code is kind of ugly**.  At a high-level the `uniffi-bindgen` code is much more pleasant to work with, since you can write Rust rather than askama templates.  However, actually writing out the AST generation is a bit of a pain.  There's a lot of nested AST items, and even nesting in a single item, for example function calls are written as `call(fn_name, [arg1, arg2, ..])`.  `arg1` and `arg2` are often nested AST expressions and I loose track at if I need to close the paren or the bracket.  I thought a lot about improving this, like maybe you could define a psuedo-language then parse it with a macro.  The issue there is that a) it's a lot of work and b) you're editing can no longer help you with code-completion.  In the end, it wasn't that bad once I got used to it and maybe I just need to start using auto-close parens/brackets feature on my editor.
- **Something was a bit weird with the Kotlin compilation**.  I'm not sure exactly what's going on, but every once in a while Kotlin would get confused and say methods weren't defined that I could see defined in the source code.  Once I ran `cargo clean` everything would magically work again.  I doubt this is a serious problem, but we should understand it before we commit to this approach.
- **Lots of cloning**.  Since we're serializing everything, there's lots of cloning going on.  I highly doubt this will matter, especially when you compare it to the time that `kotlinc` takes, but I guess I should put it out there.

## Is the IR at the correct level?

We have a ton of flexibility with the IR and it's very possible that it should be structured differently.  There are a couple design decisions that I wonder about.

One is that I tried to make it very general so that we could write bindings generators for Rust or C.  That very well might be a mistake and maybe it would clean up the code if we should assume more things about the target language like OO,  garbage collection, etc.  Also, maybe the IR is not general enough since it basically assumes an imperative language.  Based on the little I know about Haskell, I don't think we could support a bindings generator for that.

A related question is if the IR could be more declarative.  I kept trying to figure out a way where you could describe the structure of an FFI call and the bindings generator could figure out what to do.  All my attempts failed, but maybe one of you smart people can get the idea to succeed.

## Ideas that we could salvage if this fails

There are a couple ideas here that could be adopted even if we don't decide to go with the big change here.  I think I found a decent way to implement destructors on Kotlin using the `Cleaner` object.  This seems much better than requiring users to manually call `destroy`.  Also, the bindings generation templating system is based on `tera` rather than `askama` which lets you do some nice things at runtime.  This lets us implement a very nice (IMO) version of the CodeOracle system.  There's not a lot of mega-matches, templates can be easily split up over multiple modules, templates for one item can be defined based on it's child items and everything is recursively rendered, small templates can be defined inline and big ones can be in their own file,  event whitespace handling is decent.

## The name

I chose a very boring name of `bindings-ir` that I would love to have replaced with something as creative as `UniFFI`.  Any suggestions?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,9 @@
 [workspace]
 members = [
+  "bindings-ir",
+  "bindings-ir-kotlin",
+  "bindings-ir-tests",
+  "render-macro",
   "uniffi_bindgen",
   "uniffi_build",
   "uniffi_macros",
--- a/bindings-ir-kotlin/Cargo.toml
+++ b/bindings-ir-kotlin/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "bindings-ir-kotlin"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+bytes = "1.0"
+camino = "1.0.8"
+heck = "0.4"
+paste = "1.0"
+bindings-ir = { path = "../bindings-ir" }
+render-macro = { path = "../render-macro" }
+tera = { version = "1", default-features = false }
+
+[dev-dependencies]
+bindings-ir-tests = { path = "../bindings-ir-tests" }
+cargo_metadata = "0.14"
+lazy_static = "1.4"
--- a/bindings-ir-kotlin/src/buffer.rs
+++ b/bindings-ir-kotlin/src/buffer.rs
@@ -1,0 +1,85 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use bindings_ir::Renderer;
+use tera::Result;
+
+pub fn setup_renderer(renderer: &mut Renderer) -> Result<()> {
+    renderer.add_ast_templates([
+        (
+            "BufferStreamDef",
+            include_str!("templates/BufferStreamDef.kt"),
+        ),
+        ("BufStreamCreate", "{{ name }}({{ pointer }}, {{ size}})"),
+        ("BufStreamIntoPointer", "{{ buf }}.ptr"),
+        ("BufStreamPos", "{{ buf }}.byteBuf.position()"),
+        ("BufStreamSetPos", "{{ buf }}.byteBuf.position({{ pos }})"),
+        ("BufStreamSize", "{{ buf }}.size"),
+        ("BufStreamReadInt8", "{{ buf }}.byteBuf.get()"),
+        ("BufStreamReadUInt8", "{{ buf }}.byteBuf.get().toUByte()"),
+        ("BufStreamReadInt16", "{{ buf }}.byteBuf.getShort()"),
+        (
+            "BufStreamReadUInt16",
+            "{{ buf }}.byteBuf.getShort().toUShort()",
+        ),
+        ("BufStreamReadInt32", "{{ buf }}.byteBuf.getInt()"),
+        ("BufStreamReadUInt32", "{{ buf }}.byteBuf.getInt().toUInt()"),
+        ("BufStreamReadInt64", "{{ buf }}.byteBuf.getLong()"),
+        (
+            "BufStreamReadUInt64",
+            "{{ buf }}.byteBuf.getLong().toULong()",
+        ),
+        ("BufStreamReadFloat32", "{{ buf }}.byteBuf.getFloat()"),
+        ("BufStreamReadFloat64", "{{ buf }}.byteBuf.getDouble()"),
+        (
+            "BufStreamReadPointer",
+            "com.sun.jna.Pointer({{ buf }}.byteBuf.getLong())",
+        ),
+        ("BufStreamReadString", "{{ buf }}.readString({{ size }})"),
+        ("BufStreamWriteInt8", "{{ buf }}.byteBuf.put({{ value }})"),
+        (
+            "BufStreamWriteUInt8",
+            "{{ buf }}.byteBuf.put({{ value }}.toByte())",
+        ),
+        (
+            "BufStreamWriteInt16",
+            "{{ buf }}.byteBuf.putShort({{ value }})",
+        ),
+        (
+            "BufStreamWriteUInt16",
+            "{{ buf }}.byteBuf.putShort({{ value }}.toShort())",
+        ),
+        (
+            "BufStreamWriteInt32",
+            "{{ buf }}.byteBuf.putInt({{ value }})",
+        ),
+        (
+            "BufStreamWriteUInt32",
+            "{{ buf }}.byteBuf.putInt({{ value }}.toInt())",
+        ),
+        (
+            "BufStreamWriteInt64",
+            "{{ buf }}.byteBuf.putLong({{ value }})",
+        ),
+        (
+            "BufStreamWriteUInt64",
+            "{{ buf }}.byteBuf.putLong({{ value }}.toLong())",
+        ),
+        (
+            "BufStreamWriteFloat32",
+            "{{ buf }}.byteBuf.putFloat({{ value }})",
+        ),
+        (
+            "BufStreamWriteFloat64",
+            "{{ buf }}.byteBuf.putDouble({{ value }})",
+        ),
+        (
+            "BufStreamWritePointer",
+            "{{ buf }}.byteBuf.putLong(com.sun.jna.Pointer.nativeValue({{ value }}))",
+        ),
+        ("BufStreamWriteString", "{{ buf }}.writeString({{ value }})"),
+    ])
+}
--- a/bindings-ir-kotlin/src/class.rs
+++ b/bindings-ir-kotlin/src/class.rs
@@ -1,0 +1,34 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+use bindings_ir::Renderer;
+use tera::Result;
+
+pub fn setup_renderer(renderer: &mut Renderer) -> Result<()> {
+    renderer.add_ast_templates([
+        ("ClassDef", include_str!("templates/ClassDef.kt")),
+        ("DataClassDef", include_str!("templates/DataClassDef.kt")),
+        ("EnumDef", include_str!("templates/EnumDef.kt")),
+        (
+            "ExceptionBaseDef",
+            include_str!("templates/ExceptionBaseDef.kt"),
+        ),
+        ("ExceptionDef", include_str!("templates/ExceptionDef.kt")),
+        ("ClassCreate", include_str!("templates/ClassCreate.kt")),
+        ("DataClassCreate", "{{ name }}({{ values|comma_join }})"),
+        ("EnumCreate", include_str!("templates/EnumCreate.kt")),
+        ("ExceptionCreate", "{{ name }}({{ values|comma_join }})"),
+        (
+            "MethodCall",
+            "{{ obj }}.{{ name }}({{ values|comma_join }})",
+        ),
+        (
+            "StaticMethodCall",
+            "{{ class }}.{{ name }}({{ values|comma_join }})",
+        ),
+        ("IntoRust", "{{ obj }}.intoRust()"),
+    ])
+}
--- a/bindings-ir-kotlin/src/collections.rs
+++ b/bindings-ir-kotlin/src/collections.rs
@@ -1,0 +1,22 @@
+use bindings_ir::Renderer;
+use tera::Result;
+
+pub fn setup_renderer(renderer: &mut Renderer) -> Result<()> {
+    renderer.add_ast_templates([
+        ("ListCreate", "mutableListOf<{{ inner }}>()"),
+        ("ListLen", "{{ list }}.size"),
+        ("ListGet", "{{ list }}[{{ index }}]"),
+        ("ListSet", "{{ list }}[{{ index }}] = {{ value }}"),
+        ("ListPush", "{{ list }}.add({{ value }})"),
+        ("ListPop", "{{ list }}.removeLast()"),
+        ("ListEmpty", "{{ list }}.clear()"),
+        ("ListIterate", include_str!("templates/ListIterate.kt")),
+        ("MapCreate", "mutableMapOf<{{ key }}, {{ value }}>()"),
+        ("MapLen", "{{ map }}.size"),
+        ("MapGet", "{{ map }}.get({{ key }})"),
+        ("MapSet", "{{ map }}.put({{ key }}, {{ value }})"),
+        ("MapRemove", "{{ map }}.remove({{ key }})"),
+        ("MapEmpty", "{{ map }}.clear()"),
+        ("MapIterate", include_str!("templates/MapIterate.kt")),
+    ])
+}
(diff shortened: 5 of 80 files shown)


mhammond:
Thanks Ben! This is a little mind-blowing - I only spent a few minutes looking at the code and I have no idea how it actually works (ie, how those 3 examples you mention actually work!). I've a few other loose ends to catch up on before I dig further, but thought it worth thinking about the high-level concept even before I understand the code.

Re understanding:
It would be ideal if there was some overview of how this hangs together - eg, documenting how a trivial function in UDL was represented in IR and how bindings then transformed that. That might help us understand the "scope" of this IR. There are a few constructs which I think I know conceptually what they mean, but not how they would actually be used in this context.

Re scaffolding: 
I assume that the scaffolding generation doesn't change in this world, but is that correct? Are there any implications for a macro-based world?

Re: sharing code:
You mention this as the first benefit, and I think it would be wonderful if that turned out to be true. But let's say it didn't (ie, we approached diplomat and they didn't like it or didn't have the time to implement it, and nimbus also turned out to be a non-use case, etc) - do you personally think the other benefits would still make it worthwhile? Do you think it's worth trying to find other consumers who are on-board conceptually first?

Re: Fixes a lot of versioning issues/easier to change uniffi:
Do you see any risk re unintended consequences? Eg, if we update the IR generated by UniFFI but bindings aren't aware, is there any risk that it will break bindings that haven't updated? What kind of new uniffi features would be "breaking" in this world, and would they be easy to identify to the person making changes to uniffi itself?

Re: Risks
ISTM there might be a risk that this makes UniFFI less approachable to new contributors, and significantly increases the learning curve for contributors looking at adding their own bindings. Currently things are a little brittle, but askama is already a bit of a learning curve, so adding an AST might push the complexity over the limits some people can tolerate. Is there any risk about debugability? In particular, your mention that "In the end, it wasn't that bad once I got used to it" is a worry - if you were struggling when this idea was fresh in your mind, I wonder if someone more casually involved in the project might never get used to it? And "wasn't that bad" is very faint praise from someone advocating for this model :)

Re: Naming:
bindiFFI is the best I can come up with with 15 seconds thought :)

I do love the fact you are thinking about ways to reduce some of the pain-points though, and I'll dig a little more - but probably not for a week or so.

bendk:
Thanks for starting to dig in to this.  I realize it's a lot and any feedback is appreciated.  I also think you're hitting on the most critical concepts.

> Re understanding:
> It would be ideal if there was some overview of how this hangs together - eg, documenting how a trivial function in UDL was represented in IR and how bindings then transformed that. That might help us understand the "scope" of this IR. There are a few constructs which I think I know conceptually what they mean, but not how they would actually be used in this context.

This is a great idea.  It's not so easy to describe everything clearly, but I'll do my best with the arithmetic example.

- Here's a [JSON dump of the entire IR](https://gist.github.com/bendk/0a6f4d0e2a3172797b6acaece9a361b3)
- There's a lot there, here's my attempt to [highlight particular parts of it in order to make it easier to understand](https://gist.github.com/bendk/b12df6c956b66da127fa2461926a6402): 
- Here's how the [Kotlin rendering](https://gist.github.com/bendk/f95d0b3981d31d725adfc30e2a128ab9) looks.  It's pretty similar to our current bindings code with some slight differences.  In general, it has a more generated feel to it.

If we commit to this, maybe we should invest some time into making a nice way to visualize the IR.

> Re scaffolding:
>  I assume that the scaffolding generation doesn't change in this world, but is that correct? Are there any implications for a macro-based world?

The scaffolding generation stays the same.  The macros don't need to change much, but I think that instead of generating and serializing the metadata, they would call into `uniffi-bindgen` to generate and serialize the IR.

> Re: sharing code:
>  You mention this as the first benefit, and I think it would be wonderful if that turned out to be true. But let's say it didn't (ie, we approached diplomat and they didn't like it or didn't have the time to implement it, and nimbus also turned out to be a non-use case, etc) - do you personally think the other benefits would still make it worthwhile? Do you think it's worth trying to find other consumers who are on-board conceptually first?

I completely agree that we should reach out to other groups.  In this comment, I'll assume that no one responds positively in order to focus on the other benefits.  In fact, since this is a really hard question to answer, let's focus on what I think are the two most important issues: It's hard to make changes to UniFFI because every change needs to be implemented for all our bindings and it's also hard to make changes because every change is a breaking change that requires releases for other projects, vendoring, etc.

This PR represents a technical solution to address those issues.  Let's take this as a baseline.

Another way to address those issues would be some kind of workflow change.  Maybe we could update our versioning system so that we could push features like callback errors and interface types without having to push a new glean/app-services release.  Maybe we could get some sort of workflow where a new feature is implemented in one language, then experts in the other languages would implement it there.  This may be a better way to address those issues, but I don't know because I don't know how it would work.  I think it's worth trying to figure out how we would do this in order to compare and contrast.

The other approach would be to just accept that it's a hard problem that we aren't going to fix.  The less features we expect to develop, the better the do-nothing approach is compared to this PR.

>Re: Fixes a lot of versioning issues/easier to change uniffi:
> Do you see any risk re unintended consequences? Eg, if we update the IR generated by UniFFI but bindings aren't aware, is there any risk that it will break bindings that haven't updated? What kind of new uniffi features would be "breaking" in this world, and would they be easy to identify to the person making changes to uniffi itself?

I think a breaking change would be a change that added a new IR item.  For example, if we lived in a world where the current UniFFI used this system and we wanted to add callback interface error support, then we would need to add a catch statement.

I think it would be pretty hard to have a change that accidentally broke something.  I think that would involve changing the semantics of an existing IR item, which we should probably just agree to either never do.  There's also the chance that a UniFFI change exorcises a different codepath in the bindings generator, which breaks things.  But AFAICT this is always an issue and the IR doesn't fundamentally change anything.

>Re: Risks
> ISTM there might be a risk that this makes UniFFI less approachable to new contributors, and significantly increases the learning curve for contributors looking at adding their own bindings. Currently things are a little brittle, but askama is already a bit of a learning curve, so adding an AST might push the complexity over the limits some people can tolerate. Is there any risk about debugability? In particular, your mention that "In the end, it wasn't that bad once I got used to it" is a worry - if you were struggling when this idea was fresh in your mind, I wonder if someone more casually involved in the project might never get used to it? And "wasn't that bad" is very faint praise from someone advocating for this model :)

Very important question.  I think it splits into several parts.

The first part is writing bindings, which I think gets easier.  From a bindings writer standpoint, I think the AST is fairly straightforward and you need to compare it to rendering the `ComponentInterface` which means you need to know lots of UniFFI-specific things, like the FFI contract, how objects are supposed to work, etc.  Also, I think the test suite improves because you can test smaller chunks at once.   If I wanted to add new bindings, I'd start by trying to get the `geometry` tests to pass.  That's pretty basic, but much more complex than the first IR test, which is just functions and some basic math.
 
The next part is hacking UniFFI, which now means generating these AST items.  One thing that's annoying there is the amount of nesting in the syntax, which made it very confusing to me to track what thing I was currently inside, what delimiter I needed to close, etc.  Maybe I just need to figure out some better editor settings and we need to publish them for users.  Maybe there's some other solutions too, like one of the worst parts for me was that Rust doesn't support vararg functions so you need to mix `()` and `[]` in weird ways.  Maybe something like the [tuple_list](https://docs.rs/tuple_list/latest/tuple_list/) crate could help (although I think that has the same issue as macros, once you get away from regular functions your editor can't do code completion).  However, if you compare this to the pain of having to update Kotlin/Swift/JS/Python/Ruby(?) code then it doesn't seem so bad.

A related, but different, issue is the semantics of the language.  You basically need to learn and understand a new language, which is bad for potential contributors.  But again, compare that to having to understand all the languages we currently support.

One way to potentially improve the system here would be to make the meta-language less general.  If we decide that we're not going to get other projects on board, then we could maybe lean in to that and make the AST simpler by making more assumptions.  For example, maybe there should be an `FFICall` primitive that makes the call, checks the `RustCallStatus` and throws if there's an error -- right now that takes a lot of lines of code.  Similarly, I think that the buffer handling would be similified if `RustBuffer` was our official buffer struct.

If we didn't do that, then I think we should have a layer in UniFFI that tries to do the same thing.  So there would be a `rust_call_with_error` function or something that would generate the AST items needed to make the call.  Messing around with the low-level layer would require a lot of specialized knowledge, but maybe working at the higher level wouldn't.

bendk:
I think the weakest part of this PR is generating the IR from the `ComponentInterface`.  Like I mentioned in the description and in my last comment it doesn't have great ergonomics and it's also not so easy to understand.  I think those are related issues and maybe if we had a different way to represent the IR it would help with both.   I took the `add()` function, and hand-wrote some other ways it could be represented.  I wonder what others think of these paths:
 
[S-Expressions](https://gist.github.com/bendk/be2da5ac813307a5bcc617d92085f260) is probably a better way to represent the interface if you wanted to print it out for the user.  I used WASM's representation as the base here.  It also makes me think that it would be better to use macros for variable-length expressions.  For example `call!(func, arg1, arg2)` instead of `call(func, [arg1, arg2])`.  Like I mentioned before, it's confusing that `call` has adds 2 levels of nesting, one parens and one braces -- especially when `arg1` and `arg2` are complex expressions.
 
A [Rust-like meta language](https://gist.github.com/bendk/b633b2f71936ec53b476b51eab6d2a4a) is a pretty interesting option.  That opens the door to having `uniffi_bindgen` also use Tera templates, except instead of generating an real language it's generating this meta-language.  This is appealing to me because the I always found the IR -> Kotlin part nicer to work with than the ComponentInterface -> IR part.  We would need to parse this into the AST which is extra work.  I also wonder if it could be more confusing to have a language that's almost Rust, but not quite Rust.

jhugman (commented):
Wow. There's a lot to this; you've obviously put in a lot of work to get it this far.

It's obviously a work in progress, so what are the difficult/risky things left to explore? and how would you prioritize the rest of this work?

bendk:
> It's obviously a work in progress, so what are the difficult/risky things left to explore? and how would you prioritize the rest of this work?

I think the main risk is the potential for spending a lot of time on this without getting the benefits.  The first 3 benefits I listed are probably the most important ones and there's open questions for all of them:
  - **Sharing code**.  Does any project want to opt-in to this and/or do we expect to create new projects that would benefit it?  I think a "yes" to either of these could justify the work almost by itself, but so far there hasn't been any takers.
  - **Fixes a lot of versioning issues.**  This has been a real pain in getting new features to desktop and I think we should have some sort of solution.  However, should consider alternative ways to address this.  I'm going to try to put together a PR with my best alternative and we can compare the two.
 - **Makes it easier to change UniFFI.**  How many new features do we think we want?  What kinds of changes are we planning on doing?  I would love to put together a list and we can consider if this new architecture would help.

The other question in my mind is if there's a way to get these same benefits with less work.  I keep coming back to the idea of an IR that sits at a slightly higher level than this, but I'm not exactly sure how.

PR #1356, opened 2022-10-04
Update add_known_type to add inner types of generic builtins

jplatte:
For some reason the fix didn't work when I tried it on matrix-rust-sdk last week, but now I've added a dedicated test to UniFFI directly and that works. So this at least partially fixes the broken bindings I've been having issues with.


bendk (approved):
Seems fine to me.  It seems like `add_known_type()` is doing slightly different things now depending on if the type is coming from the UDL vs the proc-macro, but that seems fine.  Maybe at the end of this all, we can revisit it and clean it up.  Or maybe not, it doesn't seem like much tech-debt in any case.

PR #1357, opened 2022-10-04
Upgrade cargo_metadata

jplatte:


bendk:
How important is the `clap` upgrade?  That upgrade would require some work to vendor in to the mozilla central tree.  I don't think it's a ton of work, but if it's not needed maybe I would wait on the upgrade.

jplatte:
I see this whole PR only as cleanup. I don't really care whether it's merged now or in a month.

skhamis:
I might not understand the full effects of this update mozilla central, so I'll switch this to someone who might have more context

badboy:
I can take this

badboy:
Ok, I looked. The `cargo_metadata` update is easy.
The clap update requires a whole bunch of new audits in m-c, for which I currently won't have the time, so I'd rather hold back on that for the time being.

jplatte:
Should I split the PR, or can you cherry-pick the commit?

badboy:
> Should I split the PR, or can you cherry-pick the commit?

Splitting it would be wonderful

jplatte:
Done, clap upgrade is now https://github.com/mozilla/uniffi-rs/pull/1383.

badboy (approved):
> [@jplatte](https://github.com/jplatte) pushed 0 commits.

:D

jplatte:
It should have said `-1` 😄

PR #1360, opened 2022-10-04
Add support for records (dictionaries / structs) to the proc-macro frontend

jplatte:
Part of #1257.

Adds a `#[derive(uniffi::Record)]` macro and updates `#[uniffi::export]` to allow usage of arbitrary types as long as they don't have generics and are available under `crate::uniffi_types` and have the right trait implementation.

Outside of `uniffi::export`, when generating bindings, the names of such types are resolved against the known type definitions from both UDL and derives like `Record` (`Enum` might come next).

### DIFF
--- a/fixtures/proc-macro/src/lib.rs
+++ b/fixtures/proc-macro/src/lib.rs
@@ -1,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use std::sync::Arc;
+
+#[derive(uniffi::Record)]
+pub struct One {
+    inner: i32,
+}
+
+#[derive(uniffi::Record)]
+pub struct Two {
+    a: String,
+    b: Option<Vec<bool>>,
+}
+
+#[derive(uniffi::Record)]
+pub struct Three {
+    obj: Arc<Object>,
+}
+
+#[derive(uniffi::Object)]
+pub struct Object;
+
+#[uniffi::export]
+fn make_one(inner: i32) -> One {
+    One { inner }
+}
+
+#[uniffi::export]
+fn take_two(two: Two) -> String {
+    two.a
+}
+
+#[uniffi::export]
+fn make_object() -> Arc<Object> {
+    Arc::new(Object)
+}
+
+include!(concat!(env!("OUT_DIR"), "/proc-macro.uniffi.rs"));
+
+mod uniffi_types {
+    pub use crate::{Object, One, Three, Two};
+}
--- a/uniffi_bindgen/src/interface/function.rs
+++ b/uniffi_bindgen/src/interface/function.rs
@@ -199,16 +199,16 @@ impl Argument {
         &self.name
     }
 
-    pub fn type_(&self) -> Type {
-        self.type_.clone()
+    pub fn type_(&self) -> &Type {
+        &self.type_
     }
 
     pub fn by_ref(&self) -> bool {
         self.by_ref
     }
 
-    pub fn default_value(&self) -> Option<Literal> {
-        self.default.clone()
+    pub fn default_value(&self) -> Option<&Literal> {
+        self.default.as_ref()
     }
 
     pub fn iter_types(&self) -> TypeIterator<'_> {
--- a/uniffi_bindgen/src/interface/types/mod.rs
+++ b/uniffi_bindgen/src/interface/types/mod.rs
@@ -67,6 +67,9 @@ pub enum Type {
     External { name: String, crate_name: String },
     // Custom type on the scaffolding side
     Custom { name: String, builtin: Box<Type> },
+    // An unresolved user-defined type inside a proc-macro exported function
+    // signature. Must be replaced by another type before bindings generation.
+    Unresolved { name: String },
 }
 
 impl Type {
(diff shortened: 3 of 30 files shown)


fixtures/proc-macro/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[uniffi::export]
+fn take_two(two: Two) -> String {
+    two.a
+}
+
+#[uniffi::export]
+fn make_object() -> Arc<Object> {
+    Arc::new(Object)
+}
```
bendk, line 39: It's been nice seeing this feature grow.  There's a quite a bit of functionality here and it's expressed in a simple and clean way.

uniffi_bindgen/src/interface/function.rs
```diff
@@ -199,16 +199,16 @@ impl Argument {
...
     }
 
     pub fn by_ref(&self) -> bool {
         self.by_ref
     }
 
-    pub fn default_value(&self) -> Option<Literal> {
-        self.default.clone()
+    pub fn default_value(&self) -> Option<&Literal> {
+        self.default.as_ref()
```
bendk, line 211: Very nitty, but should this be `borrow()`?  I think of `as_ref()` as doing some sort of conversion between references.

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -67,6 +67,9 @@ pub enum Type {
     External { name: String, crate_name: String },
     // Custom type on the scaffolding side
     Custom { name: String, builtin: Box<Type> },
+    // An unresolved type inside a proc-macro exported function signature.
+    // Must be replaced by another type before bindings generation.
```
bendk, line 71: I think it's worth mentioning that this will only be a `Record`/`Object`/`Enum`, etc. and not a primitive type like `u8` or `String`.  I don't know it's an official UniFFI term, but I've been calling these "user types".

uniffi_macros/src/record.rs
```diff
@@ -0,0 +1,96 @@
...
+                let try_read_field = quote! {
+                    #ident: <#ty as ::uniffi::FfiConverter>::try_read(buf)?,
+                };
+
+                (write_field, try_read_field)
+            })
+            .unzip(),
+        None => {
+            let unimplemented = quote! { ::std::unimplemented!() };
+            (unimplemented.clone(), unimplemented)
```
bendk, line 35: I think this could actually be implemented with an empty body or maybe a `()` statement.  I don't think we have much use for unit structs as records, but it would there be a downside to implementing them?

uniffi_macros/src/record.rs
```diff
@@ -0,0 +1,96 @@
...
+            .iter()
+            .map(|f| {
+                let ident = &f.ident;
+                let ty = &f.ty;
+
+                let write_field = quote! {
+                    <#ty as ::uniffi::FfiConverter>::write(obj.#ident, buf);
+                };
+                let try_read_field = quote! {
+                    #ident: <#ty as ::uniffi::FfiConverter>::try_read(buf)?,
```
bendk, line 27: I like this as a first step.  I guess the next step would be getting it to work with user types, where the type doesn't implement `FfiConverter` itself.  I think that goes back to an issue that we discussed very early on.

Do you have a plan for this?  If not, I could spend some time trying to get it work if that would be helpful.

bendk (changes requested):
Looking great, I just had a few comments about documentation and very small issues.  I'm marking it request changes to put the ball back in your court.  However, I'd be happy merging this as-is if you disagree.

uniffi_bindgen/src/interface/function.rs
jplatte: It can't be `borrow` because this is `Option::as_ref`, not `AsRef::as_ref`. `.borrow()` would result in `&Option<Literal>`, not `Option<&Literal>`.

uniffi_macros/src/record.rs
jplatte: I don't know whether there is a downside to allowing this derive on unit structs, but note that this code path is also taken when the input is an `enum` or `union`. If it is, we also emit a compile error (line 71).

uniffi_macros/src/record.rs
jplatte: Yes, this goes back to what we discussed pretty early on. I would still prefer to not support "external" ffi-converter types, and instead encourage the newtype pattern where people were previously implementing `FfiConverter` with a `RustType` other than `Self`.

For the multi-crate setups I'm thinking about, neither would be required though; there's also the option of types having `#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]` so a crate can *optionally* have a `FfiConverter` impl in an upstream crate.

uniffi_bindgen/src/interface/types/mod.rs
jplatte: I've updated it to "user-defined type".

uniffi_macros/src/record.rs
bendk: Ahh, that makes a lot of sense.

bendk (approved):

uniffi_macros/src/record.rs
bendk: I do think that almost all of the time it's going to be one of those multi-crate setups where you control all the crates.

For completely external types, I can be convinced that requiring the newtype pattern is the best way to go.  I'm mostly just concerned that there's some use case out there where this would be lead to very bad ergonomics.  I'm going to put the question out there on `#uniffi` and see if anyone has a comment.

PR #1362, opened 2022-10-06
docs: Fix trait definition in callback interface

cgarrett-square:
Updated the trait interface example to match current practice.


bendk (approved):
Yup.  I think that's how it should have read the whole time.

PR #1363, opened 2022-10-06
Use nested `canonical_name` for Swift dictionaries

Anderas:
This change allows the disambiguation of multiply-nested compound structures, such as `[String: [String: String]]` and `[String: [String: [String]]]`. At the moment both of these types are represented as `FfiConverterDictionaryStringStringString` in the generated swift file, which leads to a compilation issue due to conflicting definitions.

After this change the generated names are as follows:
- `[String: String]` => `...DictionaryStringString`
- `[String: [String]]` => `...DictionaryStringSequenceString`
- `[String: [String: String]]` => `...DictionaryStringDictionaryStringString`
- `[String: [String: [String]]]` => `...DictionaryStringDictionaryStringSequenceString`
- etc

Note that `SequenceCodeType` and `OptionalCodeType` already use nested `cannonical_name`, whereas `MapCodeType` does not, until this change.


bendk (approved):
Nice catch, thanks for the PR.

PR #1364, opened 2022-10-06
proc-macro: unused-qualifications lint cleanup

jplatte:
We have this lint enabled in our project and for some reason while it doesn't trigger for generated code in one crate, it does in another. This should clean that up.


bendk (changes requested):
Seems good, but can you look in to that failing unit test?  I think the trybuild output needs to be updated.

jplatte:
Latest CI run failed because of a network issue.

bendk (approved):

PR #1365, opened 2022-10-06
Add tests for dictionary nesting in Swift

Anderas:
Thanks @bendk for merging my previous [PR](https://github.com/mozilla/uniffi-rs/pull/1363) so quickly, I was figuring out the best way to test the change so raising this as a separate PR


bendk (approved):
Looks good to me.  I wonder if some of these fixtures can be consolidated, but I think that should be a later discussion.

PR #1368, opened 2022-10-13
proc-macro: Resolve unresolved types nested inside builtin generics

jplatte:
Follow-up to #1360, specifically the second commit. I was hitting one of the panics from that when returning `Vec<RecordTypeThatDerivesUniffiRecord>` from a `#[uniffi::export]`ed function.


bendk (approved):
Looks good, let's merge.

PR #1369, opened 2022-10-14
Better stringify support for fielded errors

bendk:
Improved stringifying errors defined interfaces in the UDL.
  - Added message method in Kotlin and a test for it
  - Added to_s method in Ruby and a test for it
  - Swift already supported this, added tests for it
  - Python already supported this and had tests

I'm hoping to switch the application-services errors to using this style, but don't want it to be a downgrade in the quality of our log messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### What's changed
 
 - Added support for exceptions in callback interface methods.
+- Improved error stringifying on Kotlin and Ruby (the `message` and `to_s` methods respectively).
 
 ## v0.20.0 - (_2022-09-13_)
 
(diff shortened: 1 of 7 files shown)


tarikeshaq (approved):
This maks sense to me

This makes thing (again) about the breaking changes model we have - _technically_ this is a breaking change, consumers depending on the stringfied errors being a certain way will break - I do think that's a bit too far though.

Either way lets add something to the changelog (a non-breaking bullet point probably)

tarikeshaq:
I could also be misunderstanding what it means to add a `message` to the errors (i.e maybe they weren't stringfiable before? if so completely ignore my above comment!)

bendk:
> I could also be misunderstanding what it means to add a `message` to the errors (i.e maybe they weren't stringfiable before? if so completely ignore my above comment!)

No, you got it right.  I'm changing the error message.  That's an interesting point about this being potentially breaking, but I'll think calling it non-breaking is closer to correct.

PR #1371, opened 2022-10-14
proc-macro: Add Enum derive macro

jplatte:
This doesn't contain a test yet, because I didn't actually know how to best test it (having little experience in all the binding languages). Are there existing fixtures for both fieldless enums and for enums with variants that have fields? Then I can pull some of that into the proc-macro example.


jplatte:
All of those force pushes were me compile-testing the new macro with matrix-sdk-crypto-ffi 😅 

So now I'm at least confident that the macro generates valid code in the variant-with-fields case, but I still haven't tested the runtime behavior.

uniffi_macros/src/enum_.rs
```diff
@@ -0,0 +1,144 @@
...
+                        .map(|f| {
+                            let name = f
+                                .ident
+                                .as_ref()
+                                .ok_or_else(|| {
+                                    syn::Error::new_spanned(
+                                        v,
+                                        "UniFFI only supports enum variants with named fields \
+                                         (or no fields at all)",
+                                    )
```
bendk, line 84: With UDL we actually somewhat support enum variants with tuple-style fields, if they're defined with the `enum` style rather than the `interface` style.  I'm hoping with the proc-macros we can properly support them.

In terms of the PR, I think erroring out is great for now.

bendk (changes requested):
Looks great overall.  I had 1 style suggestion, if you don't agree feel free to push back.  I'd be fine merging as-is.

uniffi_macros/src/enum_.rs
jplatte: Oh, I didn't know that. Will look into it 🙂

jplatte: [The docs](https://mozilla.github.io/uniffi-rs/udl/enumerations.html) seem to disagree with you, it's either no fields for all variants (`enum` in UDL) or fields with names (`[Enum] interface` in UDL).

bendk: Oops, I was thinking of errors.  Those use the same UDL syntax, but because of historical reasons when a UDL `enum` is used for an error, the Rust `enum` can actually have fields.  It's really messy and one thing I hope the proc-macros can clean up.

bendk (approved):
Thanks for the update, looks great.

Issue #1250, opened 2022-05-25
Move `FfiConverter` definition into the scaffolding code


bendk:
This is something we've discussed a few times over the year I've been working on UniFFI, I think that maybe it's time to do it.  The reason is to better support a macro-based interface definitions.  Here's a rundown of the issue:

### The FfiConverter trait allows us to generate the scaffolding code without knowledge of the entire `ComponentInterface`, but it gets tricky with user types.

We want to allow the exported interface to be split up over multiple Rust modules.  We also want the proc-macros to be able to generate the scaffolding code.  This is tricky because the proc-macro can't "see" the code in other modules when it generates the scaffolding.  However, we can leverage the `FfiConverter` trait to make this work.

For example, suppose the definitions of `Line` and `Point` in the `geometry` crate were split out into the `geometry.types` module.  Then in the main module, we needed to generate scaffolding functions that input/output those types.

Based on this code:
```
use types::Line;

#[uniffi::export]
fn gradient(ln: Line) -> f64 { ... }
```

We can (almost) generate a scaffolding function like this:

```
pub extern "C" fn geometry_eb69_gradient(
  ln: <Line as FfiConverter>::FfiType,
)  -> <f64 as uniffi::FfiConverter>::FfiType {
  <f64 as uniffi::FfiConverter>::lower(
      gradient(<Line as FfiConverter>::try_lift(ln))
}
```

It's pretty cool how we can generate this function without knowing how `Line`s are lowered or lifted.  However, right now this code will fail because we don't implement `FfiConverter` for `Line` or any other user types.

For structs, records, objects we don't implement `FfiConverter` directly on the type itself.  The reason we do this is `FfiConverter` is defined in the `uniffi` crate, the implementation lives in the library crate that we're generating scaffolding for, and some of our types live in a 3rd crate (for example `SqlInterruptHandle`).  In these cases, the Rust orphan rule prevents the library crate from implementing `FfiConverter` on the type since both the type and trait are defined in other crates.

There are a few way to handle this, but my proposal is to change this so `FfiConverter` is defined by the generated scaffolding code, in the library crate.

### How would FfiConverter work if it was defined in the scaffolding code?

If `FfiConverter` was defined in the scaffolding code, then we couldn't reference it from the `uniffi` crate.  I think most things would continue to work fine though.
  - We used to have generic functions that used `FfiConverter` as a trait bound in `rustcalls.rs`, but we don't anymore.  So I don't think that would be an issue.
[...]

PR #1373, opened 2022-10-18
Remove FfiConverter types

jplatte:
For enums, records, errors and callback interface, instead of there being a separate `FooFfiConverter` that implements `FfiConverter<RustType = ActualType>`, the actual type will now always implement `FfiConverter` itself. Within the trait, `Self` now takes the place of `Self::RustType`.

This means users can no longer define `dictionary`'s or `enum`s in UDL when the underlying Rust type is defined in another crate. However, the `typedef extern` feature continues to work, in a slightly different way (see updated docs).


bendk:
This definitely cleans up the scaffolding code by a significant amount.  It means that it's going to be harder for projects to use remote types in their UniFFI interfaces if those types come from third party crates.  I believe the proposal is to wrap those types in a newtype, is that right?

If those remote types are from a different crate in the same workspace or in general a different crate that you control, then it means you need to implement `FfiConverter` in that crate.  I assume that's easy with a macro.  For users still on UDL does this mean they need to manually implement it for now?

Note that there's a lot of discussion of this in #1250.  I proposed an alternative system that wouldn't require a newtype, but would be more complex internally.  I think I've been convinced on this approach though, since AFAICT we don't currently use any remote types from 3rd-party crates and don't have any plans to.

What do others think?

jplatte:
Well, we seem to actually be using the same pattern in at least one instance so maybe I should actually work on allowing cross-crate usage of the proc-macro before we merge this PR..

examples/custom-types/src/lib.rs
```diff
@@ -40,7 +46,7 @@ pub struct CustomTypesDemo {
 
 pub fn get_custom_types_demo(v: Option<CustomTypesDemo>) -> CustomTypesDemo {
     v.unwrap_or_else(|| CustomTypesDemo {
-        url: Url::parse("http://example.com/").unwrap(),
+        url: url::Url::parse("http://example.com/").unwrap().into(),
```
mhammond, line 49: I'm not that keen on this change - it forces us back into using the wrapper explicitly, which the `UniffiCustomTypeConverter` was added explicitly to avoid.

The other direction of this can be demonstrated by adding to the udl, say:
`string rt_url(Url url);`
and the matching impl:
`pub fn rt_url(url: url::Url) -> String { url.into() }`

works on main and fails with this patch. For our use-case, this would mean that every public function we have, which uses `url::Url` can't be exposed directly and will need to change to use the wrapper type or jump through other hoops.

IOW, we previously had a solution that involved wrapper types, but added `UniffiCustomTypeConverter` because the wrapper types are a PITA. I'm 100% up for changing the implementation, but not really up for regressing the capability.

mhammond (changes requested):

examples/custom-types/src/lib.rs
mhammond: eg, [this udl](https://github.com/mozilla/application-services/blob/7155bbb4a9b091a272cae01f98e8ceefe9235160/components/places/src/places.udl#L65) has [this impl](
https://github.com/mozilla/application-services/blob/7155bbb4a9b091a272cae01f98e8ceefe9235160/components/places/src/ffi.rs#L271) - and obviously every other use of `Url` in that udl.

jplatte:
Superseded by #1444.

PR #1374, opened 2022-10-18
[Discuss] Uniffi combined frontend

bendk:
This PR represents an alternative solution to the versioning problem discussed in #1353.

The problem in a nutshell is that any change to `uniffi-bindgen` is a breaking change.  This includes both impactful changes like the [callback errors change](https://github.com/mozilla/uniffi-rs/commit/f6787147dc4fac4adb25422eeaa1bc8bf22a30ae) that changes the FFI contract, as well as minor things like [renaming `uniffi-bindgen::interface` methods](https://github.com/mozilla/uniffi-rs/commit/0bf18394a49856ce0705a7eae3cb1c0127d6ffb9) which is breaking for external bindings generators but inconsequential for "normal" UniFFI consumers.

Breaking changes are quite annoying for desktop work, since it means we need to do a Glean release, an app-services release, and vendor both of those into moz-central in order to start using the new UniFFI code.

To avoid this issue, this PR moves around the code so that the old `uniffi` crate is now called `uniffi_core` and the new `uniffi` crate works like a frontend to all the other UniFFI crates.  Instead of depending on `uniffi`, `uniffi_build`, `uniffi_macros`, `uniffi_bindgen`, consumers only need to depend on `uniffi`.

The advantage of all of this is that `uniffi` can have different versions than the other crates.  It's version number represents the contract between UniFFI and a UniFFI-ed crate.  We basically only need to bump that version if we make a change that's actually breaking for projects like Glean or app-services (e.g. dropping a UDL feature, replacing UDL with proc-macros, changing how the generated code works, etc.).  Other changes will just need a minor or patch release, including changes that change the FFI contract since both the `uniffi` crate handles both bindings and scaffolding generation.

Related to that, I updated the docs to tell users to use the basic `embedded-uniffi-bindgen` that we're using in app-services.  Users should create their own binary that depends on `uniffi` and runs `uniffi-bindgen`.  This part is essential for making sure that `uniffi_core`, the `uniffi_bindgen` used for scaffolding, and the `uniffi_bindgen` used for bindings are all compatible.  This means that there's one more source of breaking changes:  [`uniffi-bindgen` CLI changes](https://github.com/mozilla/uniffi-rs/commit/1972e60324fbe4d8fbda7b121a311c839f03129c).

I picked `1.0.0` for the new `uniffi` version.  That's maybe promising too much, but I think that we might be at the point where we can start really slowing down the number of breaking changes, at least from the typical UniFFI user's POV.

Like I mentioned at the very start, this is something of an alternative to #1353 that only focuses on the versioning issue and requires a lot less work.  I think this PR solves all the main issues as well as #1353, am I missing something?


docs/manual/src/tutorial/foreign_language_bindings.md
```diff
@@ -2,22 +2,55 @@
...
 As stated in the [Overview](../Overview.md), this library and tutorial does not cover *how* to ship a Rust library on mobile, but how to generate bindings for it, so this section will only cover that.
 
-First, make sure you have installed all the [prerequisites](./Prerequisites.md) - particularly,
-installing `uniffi-bindgen` (or alternatively, understanding how to run it from the source tree)
+## Creating the bindgen binary
 
-## Kotlin
+First, make sure you have installed all the [prerequisites](./Prerequisites.md).
+
+Ideally you would then run the `uniffi-bindgen` binary from the `uniffi` carte to generate your bindings.  However, this
```
badboy, line 9:
```suggestion
Ideally you would then run the `uniffi-bindgen` binary from the `uniffi` crate to generate your bindings.  However, this
```

badboy: Honestly, this is docs. Let's just go with the stable way (and maybe only have a hint at the end "this might change once artifact dependencies are available on stable Rust" or similar)

badboy (approved):
> Breaking changes are quite annoying for desktop work, since it means we need to do a Glean release, an app-services release, and vendor both of those into moz-central in order to start using the new UniFFI code.

Having run into this problem now countless times I appreciate any effort to avoid it.
I was already thinking about it a bit, but more towards just taking out the version number out of the hash calculation completely. 
Your uniffi/uniffi_core split is the better way, as it allows us to still communicate changes.

> Other changes will just need a minor or patch release, including changes that change the FFI contract since both the `uniffi` crate handles both bindings and scaffolding generation.

Just to spell it out once again:
If all we change is internals and the generated API on both sides (foreign language and Rust) stays the same, we can freely update `uniffi_*` crates in m-c and it will just work.
Correct?

I think we should document a (non-exhaustive) list of things that **do** require a version bump of `uniffi`. 
I guess "breaking changes" in `uniffi_*` translate to a version bump in `uniffi` (but not minor/patch*)?
Essentially anything that changes `uniffi_bindgen/src/scaffolding/templates` needs careful evaluation.
The callback changes you mentioned are backwards-compatible, and so are ok.
The renaming changes you mentioned are also fine, because the generated code is the same.

(* leaving out that we are currently <v1 and so minor/patch is not saying much).

I'm in favor of this over #1353 because it allows us to do it now, rather than the need to iterate on the much more complex IR way of doing things. It also doesn't stop us from still eventually going the IR route internally if that ends up being a better way.

bendk:
> Just to spell it out once again: If all we change is internals and the generated API on both sides (foreign language and Rust) stays the same, we can freely update `uniffi_*` crates in m-c and it will just work. Correct?

I should have thought this through some more, but I think it would work something like this:
  - m-c, app-services, and Glean all depend on `uniffi` for generating the scaffolding/bindings.  Those dependencies are loosely specified (something like `>= 1.2.0, < 2.0.0`)
  - `uniffi` depends on specific versions of the `uniffi_*` crates.
  - m-c also hosts the JS bindings generator which depends on `uniffi_bindgen`.  We'd ensure that version matched the version that `uniffi` depended on.

So for something like the callback error change, we could:
  - Merge the change into the UniFFI repo
  - Make a UniFFI release.  This would be a breaking change for the `uniffi_*` crates, but a non-breaking change for `uniffi'. 
  - Vendor the new `uniffi` and `uniffi_bindgen` versions into m-c.  No need to update Glean or app-services, since they only depend on `uniffi`.
  - Start using that code in 
 
> I think we should document a (non-exhaustive) list of things that **do** require a version bump of `uniffi`.

Good point.  I can think of the following:
  - Change to the generated bindings API (for example, changing class names)
  - Removing or changing the existing UDL syntax.  Adding new things should be okay though.
  - Changing the `uniffi-bindgen` CLI args
  - Am I missing anything?

I think.  If we agree to this PR, then we're making an implicit commitment to not make those changes so often. 

> I'm in favor of this over #1353 because it allows us to do it now, rather than the need to iterate on the much more complex IR way of doing things. It also doesn't stop us from still eventually going the IR route internally if that ends up being a better way.

Agreed

jplatte:
Two minor comments:

- I think going to 1.0 now is a little bit too optimistic
- If there is a "core" crate, I would expect that to change *less frequently* than the "main" crate based on the name; maybe a different name makes sense

Also I would love to take a closer look but I'm not sure when I'll have time.

bendk:
> If there is a "core" crate, I would expect that to change _less frequently_ than the "main" crate based on the name; maybe a different name makes sense

Maybe it actually could change less frequently.  If we implemented this change, then maybe we don't need to keep all the crates on the same version anymore.  Consumers will only depend on `uniffi` and we'll make sure the transitive dependencies are compatible.  AFAICT, most of our recent releases have been driven by `uniffi_bindgen` changes, but the other crates haven't changed much.

I also think we could reorganize the code to lessen the need for changes to `uniffi_core`.  Moving the UI tests into `uniffi` helps.  Also, we could avoid changes like [this](https://github.com/mozilla/uniffi-rs/commit/5856921b96ee6a111f0c7934fba90ac1169f0436#diff-01094e148fabb76cca59a5117841e92d58eb67e8ccebda1c413dd0e5ca5ce2d8R61) if the `uniffi` crate handled the re-exporting rather than `uniffi_core`.

bendk:
> If there is a "core" crate, I would expect that to change _less frequently_ than the "main" crate based on the name; maybe a different name makes sense

Maybe it actually could change less frequently.  If we implemented this change, then maybe we don't need to keep all the crates on the same version anymore.  Consumers will only depend on `uniffi` and we'll make sure the transitive dependencies are compatible.  AFAICT, most of our recent releases have been driven by `uniffi_bindgen` changes, but the other crates haven't changed much.

I also think we could reorganize the code to lessen the need for changes to `uniffi_core`.  Moving the UI tests into `uniffi` helps.  Also, we could avoid changes like [this](https://github.com/mozilla/uniffi-rs/commit/5856921b96ee6a111f0c7934fba90ac1169f0436#diff-01094e148fabb76cca59a5117841e92d58eb67e8ccebda1c413dd0e5ca5ce2d8R61) if the `uniffi` crate handled the re-exporting rather than `uniffi_core`.

I just opened #1378 which we should merge before this one.  That updates the testing code so `uniffi` no depends on `uniffi_bindgen`, which would allow us to make a breaking change to `uniffi_bindgen` without needing to bump the `uniffi` version.

mhammond:
I'm inclined to agree with Jonas that we probably don't want to declare 1.0 here and that uniffi-core isn't a great name. uniffi-runtime almost, but not quite, works. I'm not that bothered about the name though.

But the discussion above about minor vs patch levels makes me wonder if this entire scheme depends on us moving to 1.0 here?

I'm also still trying to understand how this helps in practice (which isn't to say it doesn't, just that I don't quite get it). Part of our problem with m-c seems to be that glean ends up in m-c via crates.io, whereas app-services comes in via "vendoring" - and we want m-c to end up with exactly 1 version of uniffi. When glean wants to bump the uniffi version, it will be because it *needs* something in that new version - and ditto for app-services.

Is the idea that it's ok to end up with m-c depending on multiple uniffi_bindgen transitive dependencies, so long as the top-level uniffi dependency is semver compatible? If so, I guess this whole scheme does depend on that top-level crate being 1.0 - which, as above, seems wrong - the fact the sub-crates aren't at 1.0 isn't going to register with people - we'd effectively be moving all of uniffi to 1.0.

I wonder if we should experiment more in those other repos - see what the changes there actually look like, then look in detail at the recent uniffi changes that forced them to take new uniffi versions and make sure it will do what we think it will. I'd hate to land something like this just to discover it doesn't really help in practice.

Sorry I didn't get as much time to look at this as I would have liked and I'm now going away for a week. But if @bendk and @badboy are convinced it will work as expected I'm certainly not going to stand in its way.

mhammond:
Alternatively, could we do this in a different repo? something like a moz-uniffi repo, which takes uniffi as a dep, and glean and a-s move to the new repo? It being 1.0 seems far less problematic.

jplatte:
> Breaking changes are quite annoying for desktop work, since it means we need to do a Glean release, an app-services release, and vendor both of those into moz-central in order to start using the new UniFFI code.

Can you expand on this? If UniFFI a public dependency of some library, or why is a semver-compatible version upgrade an improvement over a semver-incompatible one where you don't have to do any changes other than to bump the version?

Re. v1.0, I can't find anything in the comments above on why it would be *necessary*.

bendk:
> Is the idea that it's ok to end up with m-c depending on multiple uniffi_bindgen transitive dependencies, so long as the top-level uniffi dependency is semver compatible?

The idea is to that m-c will only have 1 uniffi_bindgen transitive dependency.  Maybe we would have a situation like this:
  - app-services depends on `uniffi >= 1.1.0, < 2.0.0` which depends on `uniffi_bindgen:0.30`
  - Glean depends on `uniffi:>= 1.2.0, < 2.0.0`, which depends on `uniffi_bindgen:0.32`
  - moz-central uses `uniffi=1.2` and `uniffi_bindgen:0.32`, which satisfies all the constraints.
  - The same logic applies for other `uniffi_*` components`.

I *think* that's correct, but I'm not sure.

> If so, I guess this whole scheme does depend on that top-level crate being 1.0 - which, as above, seems wrong - the fact the sub-crates aren't at 1.0 isn't going to register with people - we'd effectively be moving all of uniffi to 1.0.

The [cargo book](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility) says "Versions are considered compatible if their left-most non-zero major/minor/patch component is the same", so I think this would work without bumping the version to 1.0.  However, if there was a conflict here I would go with an incorrect version number if it eases the development workflow (and I generally agree that declaring UniFFI 1.0 is probably incorrect).

bendk:
> Can you expand on this? Is UniFFI a public dependency of some library, or why is a semver-compatible version upgrade an improvement over a semver-incompatible one where you don't have to do any changes other than to bump the version?

UniFFI is a dependency for several Mozilla libraries which are dependencies of Firefox.  So each time there's a UniFFI change that's semver-incompatible (based on Cargo's understanding of semver), there needs to release for each of those libraries, then a PR to update the libraries.  So it's just version bumping, but it's several of them and it requires cross-team collaboration.

bendk:
I'm hoping to revive this one and get it merged.  I just pushed a bunch of updates to this one to bring it up to date with the current main as well as update a few things:
  - Changed the version number from `1.0.0` to `0.22.0` since everyone agreed calling this `1.0` was too optimistic.
  - Moved the CLI code and optional clap dependency from `uniffi_bindgen` to `uniffi`.   Now consumers that want to generate bindings programmatically ([example](https://github.com/PyO3/maturin/blob/4bceaa57039e00f3063c918a578730dd09e4308e/src/module_writer.rs#L878)) can depend on `uniffi` without the `clap` dependency.
  - Integrated the testing changes, now `uniffi_core` doesn't depend on any other UniFFI crates.

To recap the new system:
  - For most consumers, the `uniffi` crate is the only crate that they will depend on.  It re-exports/re-packages the code from the other UniFFI crates to present consumer-friendly API.  `uniffi` will be on a different versioning scheme than the other components and we only need to make a breaking version bump if the contract to those consumers changes.  This would mean something like dropping support for a UDL item or changing the class names in the bindings code, but not changing the `uniffi_bindgen::interface` types.  Glean, app-services, etc. could specify their dependencies such that we don't need to update them for those types of changes ([details in this comment](https://github.com/mozilla/uniffi-rs/pull/1374#issuecomment-1295356079)).
 - The other `uniffi_*` crates don't need to keep their version numbers in sync anymore.  We could push a new `uniffi_bindgen` version without having to update `uniffi_core` and the rest.
 - External bindings would still depend on `uniffi_bindgen` so breaking changes to the `uniffi_bindgen::interface` types would be breaking changes for them.
  
I'm open to changing the names of the `uniffi` and `uniffi_core` crates to whatever people like best.  I just didn't know what to pick.

uniffi/src/cli.rs
```diff
@@ -0,0 +1,105 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use camino::Utf8PathBuf;
+use clap::{Parser, Subcommand};
+// structs to help our cmdline parsing. Note that docstrings below form part
+
```
badboy, line 8:
```suggestion

// structs to help our cmdline parsing. Note that docstrings below form part
```

badboy: moving the empty newline above the comment instead of in the middle of it.

uniffi/Cargo.toml
```diff
@@ -11,23 +11,21 @@ edition = "2021"
...
-uniffi_bindgen = { path = "../uniffi_bindgen", default-features = false, optional = true, version = "=0.22.0" }
-uniffi_macros = { path = "../uniffi_macros", version = "=0.22.0" }
-static_assertions = "1.1.0"
-
-[features]
-default = []
-# Use the `uniffi_bindgen` from this workspace instead of the one installed on your system.
-# You probably only want to enable this feature if you're working on uniffi itself.
-builtin-bindgen = ["uniffi_bindgen"]
+clap = { version = "3.1", features = ["cargo", "std", "derive"], optional=true }
```
badboy, line 20:
```suggestion
clap = { version = "3.1", features = ["cargo", "std", "derive"], optional = true }
```

badboy:
> > Is the idea that it's ok to end up with m-c depending on multiple uniffi_bindgen transitive dependencies, so long as the top-level uniffi dependency is semver compatible?
> 
> The idea is to that m-c will only have 1 uniffi_bindgen transitive dependency. Maybe we would have a situation like this:
> 
>     * app-services depends on `uniffi >= 1.1.0, < 2.0.0` which depends on `uniffi_bindgen:0.30`
> 
>     * Glean depends on `uniffi:>= 1.2.0, < 2.0.0`, which depends on `uniffi_bindgen:0.32`
> 
>     * moz-central uses `uniffi=1.2` and `uniffi_bindgen:0.32`, which satisfies all the constraints.
> 
>     * The same logic applies for other `uniffi_*` components`.
> 
> 
> I _think_ that's correct, but I'm not sure.
> 
> > If so, I guess this whole scheme does depend on that top-level crate being 1.0 - which, as above, seems wrong - the fact the sub-crates aren't at 1.0 isn't going to register with people - we'd effectively be moving all of uniffi to 1.0.
> 
> The [cargo book](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility) says "Versions are considered compatible if their left-most non-zero major/minor/patch component is the same", so I think this would work without bumping the version to 1.0. However, if there was a conflict here I would go with an incorrect version number if it eases the development workflow (and I generally agree that declaring UniFFI 1.0 is probably incorrect).

This won't actually work though.
When Glean is built with `uniffi_bindgen:0.32` then the latest release of the Glean Kotlin bindings have a checksum embedded in the Kotlin code that's based on that `0.32`.
Freely updating `uniffi_bindgen` to any newer version in m-c will now embed that newer version into the checksum of generate `extern "C"` scaffolding functions.
We would still need to update uniffi in Glean, followed by a release, before we update `uniffi_bindgen`  in m-c for it to not break when it gets to Fenix.
The only way to not have that is change which version number is input to the checksum ([current code](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/interface/mod.rs#L111-L113)).

bendk:
> This won't actually work though.
> When Glean is built with `uniffi_bindgen:0.32` then the latest release of the Glean Kotlin bindings have a checksum embedded in the Kotlin code that's based on that `0.32`.
> Freely updating `uniffi_bindgen` to any newer version in m-c will now embed that newer version into the checksum of generate `extern "C"` scaffolding functions.
> We would still need to update uniffi in Glean, followed by a release, before we update `uniffi_bindgen` in m-c for it to not break when it gets to Fenix.
> The only way to not have that is change which version number is input to the checksum ([current code](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/interface/mod.rs#L111-L113)).

I didn't think about bindings that are checked in, but of course that's going to be an issue.  Sadly, for checked in code, any change to the FFI contract is a breaking change, including the [callback interface changes](https://github.com/mozilla/uniffi-rs/pull/1344) and the recent [checksum change](https://github.com/mozilla/uniffi-rs/pull/1427).

I think this means that any change to the FFI contract should be a breaking change to `uniffi` and the version should be bumped in a breaking fashion.  We should take some time to consider the current FFI contract, consider if there are changes we want to make, and if so make them all at once.

I think there are some use cases where this doesn't need to be a breaking change but I can't come up with a compelling one.

bendk:
Pushed a change to address the checksum versioning issue and also `docs/uniffi-versioning.md` to describe how we should handle version bumps.

docs/uniffi-versioning.md
```diff
@@ -0,0 +1,40 @@
...
+To expand on the previous point, here are the scenarios where `uniffi` should get a breaking version bump:
+
+    * Backward incompatible changes to the UDL/proc-macro parsing:
+      * Removing a feature.
+      * Changing how existing UDL/proc-macro code is handled -- for example if we changed UniFFI functions to return a `Result<>` enum rather than throwing exceptions.
+      * Note: Adding new UDL or proc-macro features is not a breaking change.
+    * Backward incompatible changes to the FFI contract between the scaffolding and bindings code:
+       * Changing how FFI functions are named.
+       * Changing how FFI functions are called
+       * Changing how types are represented.
```
badboy, line 34: Formatting nit. Indenting it by 4 spaces turns this into a code block.
Not indenting means it's a proper list.

```suggestion
* Backward incompatible changes to the UDL/proc-macro parsing:
  * Removing a feature.
  * Changing how existing UDL/proc-macro code is handled -- for example if we changed UniFFI functions to return a `Result<>` enum rather than throwing exceptions.
  * Note: Adding new UDL or proc-macro features is not a breaking change.
* Backward incompatible changes to the FFI contract between the scaffolding and bindings code:
  * Changing how FFI functions are named.
  * Changing how FFI functions are called
  * Changing how types are represented.
```

docs/uniffi-versioning.md
```diff
@@ -0,0 +1,40 @@
...
+
+## How do we handle breaking changes
+
+  * Increment the minor version of `uniffi`
+    * Once we get to `1.0` then this will change to be a major version bump.
+  * Update the `uniffi_bindgen::UNIFFI_CONTRACT_VERSION` string
```
badboy, line 40: Formatting nit. This was indented by 2 spaces, which doesn't render as code at least, but is inconsistent with other formatting. 

```suggestion
* Increment the minor version of `uniffi`
  * Once we get to `1.0` then this will change to be a major version bump.
* Update the `uniffi_bindgen::UNIFFI_CONTRACT_VERSION` string
```

docs/uniffi-versioning.md
```diff
@@ -0,0 +1,40 @@
...
+    * Backward incompatible changes to the UDL/proc-macro parsing:
+      * Removing a feature.
+      * Changing how existing UDL/proc-macro code is handled -- for example if we changed UniFFI functions to return a `Result<>` enum rather than throwing exceptions.
+      * Note: Adding new UDL or proc-macro features is not a breaking change.
+    * Backward incompatible changes to the FFI contract between the scaffolding and bindings code:
+       * Changing how FFI functions are named.
+       * Changing how FFI functions are called
+       * Changing how types are represented.
+
+## How do we handle breaking changes
```
badboy, line 36: Instead of a question (and then it's missing the question mark) we should make this an action

```suggestion
## How to handle breaking changes
```

badboy (approved):

bendk:
I tested using this to build application-services and it mostly worked, with a couple hitches:
  - We were using the `builtin-bindgen` feature, which I removed.  I added back the feature for backwards compatibility, but now it doesn't do anything.
  - We were calling the `uniffi_bindgen::run_main` function, which is now moved to `uniffi`.  I didn't want to add back the `clap` dependency, so I left the backwards-incompatibility and added a CHANGELOG note for it.

Looks like there's no concerns about the code.  I'm going to merge it.

I think the last step is going to be to write a nice CHANGELOG entry and some more documentation.  I'll do that in a separate PR though, this one seems full to me.

PR #1375, opened 2022-10-19
Use the log facade instead of directly error printing

badboy:
We can't control where `eprintln` goes (heck, I saw it fail and blow up things). We already have logging, so let's use that.


jhugman (approved):
This looks good to me!

bendk (approved):
Agreed, that's a better way to do it than my initial implementation.

PR #1376, opened 2022-10-19
Adding a warning about circular references with callback interfaces

bendk:
This was brought up as an issue from the DOM team when discussing the JS implementation, but it's really a general issue for all bindings.  It seems like a good idea to add this warning to the docs.


tarikeshaq (approved):
This looks good to me!

This may not be feasibly easy with how our templates are structured, but can we add doc comments to our bindings generators? Something that possibly can popout if someone is registering a callback in an IDE

bendk:
> This may not be feasibly easy with how our templates are structured, but can we add doc comments to our bindings generators? Something that possibly can popout if someone is registering a callback in an IDE

That's a really good question.  We generate the interface/protocol that the callback implements, maybe it would be possible to add a doc string there.  But would that show up in an IDE?  I don't think so, but I'm not sure.

PR #1377, opened 2022-10-20
Permit redundant definition of records and enums

jplatte:
… as long as all definitions are exactly the same. Required to use an enum or record in the signatures of both proc-macro-exported functions as well as UDL functions within one API.

Related to #1257.


skhamis (approved):
LGTM! 🚀

Issue #1194, opened 2022-03-29
Create better framework for Swift dylibs when testing


bendk:
The way we build Swift dylibs for testing is pretty weird.  There's 2 libraries involved: the library build from the Rust crate, and the library built from the autogenerated bindings t code.

In `compile_bindings()`, [we hardcode the bindings library extension as `.dylib`](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/swift/mod.rs#L133-L134).  This seems very odd, since the `file` command reports this as an ELF library, not a mach one which the extension implies.

Eventually, we should use one of the [`DLL_SUFFIX` value from `std::env::const`](https://doc.rust-lang.org/std/env/consts/index.html), however that `.dylib` extension is actually doing some work for us, because otherwise some of our test fixtures would use the same filename for both the rust crate and the bindings code (`libcallbacks.so` for the callbacks example).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-145)
┆friendlyId: UNIFFI-145
┆sprintEndDate: 2022-10-28

bendk:
This is fixed in #1378.

PR #1378, opened 2022-10-20
Use `uniffi_testing` for our tests

bendk:
Use `uniffi_testing` for our tests

The new system has several advantages:
  - It has handling for external crates, which makes the python
    ext-types fixtures work again.
  - It uses the correct suffix for Swift modules (#1194)
  - It doesn't require the `uniffi::testing` module, which I think will
    help #1374

This means that the example/fixture crates need to take a dev-dependency on `uniffi_bindgen`.  I also updated several to take a dev-dependency on `uniffi_macros` instead of a regular dependency, since they only needed it for the tests. The required change was making sure all the crates had a `lib` crate-type.

While updating the bindings code to use `uniffi_testing`, I also moved it into its own module.


mhammond:
I'm not a huge fan of this - I think it does add a fair bit of complexity and makes the repo more difficult to understand. For my workflows, I tend to make a change to one example or fixture, then `cargo test -p uniffi-fixture-something` and everything I care about runs. In that scenario I explicitly do not want to run the other tests (and I use `UNIFFI_TESTS_DISABLE_EXTENSIONS` if I want to restrict the foreign bindings I run) - does this allow me to do the same in any sane way? IOW, in this world, how do I run all foreign binding tests for exactly 1 fixture and nothing else?

The fact I don't like it doesn't mean we shouldn't do it though, if it enables #1374 - but it's not quite clear to me yet how it impacts that in practice - you mention it will reduce the number of changes made to the uniffi-core crate, but it's not obvious to me how it does that. That's probably because I'm still trying to get my head around #1374 itself.

bendk:
I'm not sure that this adds complexity.  The new code is definitely complex but so was the code it's replacing.  Can you elaborate on this one?

You can run tests for a specific fixture using an argument like `cargo test geometry` or `cargo test test_kotlin_geometry`.  But it's definitely not as simple as before.  I can re-arrange the code so the tests are back in the fixtures, which would make things closer to before.  The main reason I didn't was I wasn't sure where to put the language specific testing code.  If it stays in `uniffi_bindgen`, then either we ship it with the binary or hide it behind feature flag.  It could also go in `uniffi_testing`, but it seemed more natural to put it in `uniffi_bindgen`.  Any opinions on which one of these options is best?

bendk:
Pushed an update here to move the tests back into the example/fixture crates, which seems better.  This makes running the tests work exactly like before and simplifies the changes.

One issue for the future is what to do if we want to move the bindings generators into external crates, but we can figure that out later.

examples/arithmetic/tests/test_generated_bindings.rs
```diff
@@ -1,5 +1,5 @@
 uniffi_macros::build_foreign_language_testcases!(
-    ["src/arithmetic.udl",],
+    "uniffi-example-arithmetic",
```
mhammond, line 2: The vast majority of these tests now duplicate the crate name the tests are in - how do you feel about `build_foreign_language_testcases` just taking the list of test files and using `$CARGO_PKG_NAME` as the crate name, and if necessary, a slightly different name for the macro taking an explicit crate name if there are any that are not default?

mhammond (commented):
> I'm not sure that this adds complexity.
> it's definitely not as simple as before.

> Can you elaborate on this one?

"definitely not as simple as before" == "adds complexity" to me :)

> Pushed an update here to move the tests back into the example/fixture crates, which seems better.

I agree this is much better, thanks. LGTM, but I think it would be even better to not have the redundant crate name when possible, but let me know if there's a good reason to not do that.

examples/arithmetic/tests/test_generated_bindings.rs
bendk: That's a great suggestion and it makes all the code much simpler.

mhammond (approved):
Thanks!

PR #1387, opened 2022-10-27
Add extra definition checks to ComponentInterface::check_consistency

jplatte:
With the proc-macro frontend, a lot of preconditions that were previously always true can no longer be relied on, sometimes leading to panics later in the code with bad error messages.

This should catch a few precondition violations earlier, and produce better error messages.


bendk (approved):
Looks good.

Issue #1218, opened 2022-04-21
Support external types in swift


mhammond:
We have "Custom" support in swift, but not "External" - witness [the lack of swift tests in our only fixture](https://github.com/mozilla/uniffi-rs/tree/main/fixtures/ext-types/lib/tests/bindings) with an `[External=...]` attribute.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-155)
┆friendlyId: UNIFFI-155
┆sprintEndDate: 2022-10-28

PR #1389, opened 2022-10-27
Swift external types (#1218)

bendk:
Swift has a different module system than most other languages, so it needed a slightly different implementation.  There's basically 2 cases:
  - All generated `.swift` files are compiled together in a single     module.  This is what we currently do for the application-services
    megazord.  In this situation, we can use the external types and FfiConverters directly, as long as they're not fileprivate.
  - Each generated `.swift` file is complied into a separate module.  In this case we need to import the module where the external type is defined.

I'm no Swift expert, so please tell me if this doesn't make sense.

This change required a few related changes:
  - #1378 to be able to test it
  - Renaming some callback interface types to avoid a name collision on `Handle`


fixtures/ext-types/lib-with-seperate-swift-modules/uniffi.toml
```diff
@@ -0,0 +1,17 @@
...
+# Map the crate names from [External={name}] into Kotlin package names
+custom-types = "customtypes"
+ext-types-guid = "uniffi.ext_types_guid"
+uniffi-one = "uniffi.uniffi_one"
+
+[bindings.swift.external_modules]
+# Map the crate names from [External={name}] into Swift module names
+custom-types = "custom_types"
+ext-types-guid = "ext_types_guid"
+uniffi-one = "uniffi_one"
```
mhammond, line 17: I don't quite understand these - not only do they look like they are what the default values would be, there's no mention of the new entry documented above. The test does fail if they are removed, but I really don't yet understand why :)

fixtures/ext-types/lib-with-seperate-swift-modules/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+license = "MPL-2.0"
+publish = false
+
+[package.metadata.uniffi.testing]
+external-crates = [
+    "uniffi-fixture-ext-types-guid",
+    "uniffi-fixture-ext-types-lib-one",
+    "uniffi-example-custom-types",
+]
+swift-use-seperate-modules = true
```
mhammond, line 15: Is this really a testing attribute? I overlooked this expecting to find it in uniffi.toml. How does a "real" consumer specify this in a non-testing context? (This again just shows I'm missing something obvious about this use-case!)

mhammond (commented):
Thanks for taking this on. I ran out of time, but I hope you can help me understand some of the details before I dive back in. I'm quite sure I'm missing something that will be obvious so apologies for some of the ignorant questions.

* You mention there are 2 cases - the first case is what we do now in app-services, but the second case isn't clear to me - what is that second case supporting exactly and why? ie, why do we have 2 scenarios for swift but not for others (eg, I could imagine Python theoretically wanting those 2 scenarios too). There is a new fixture for the other scenario, but I can't see much different between that and the existing fixture except for a uniffi-testing attribute, so I'm slightly confused here.
* That second case, where "Each generated `.swift` file is complied into a separate module." seems to be very similar to what Python does now - the existing `uniffi-fixture-ext-types` fixture generates multiple .so and multiple .py files, and does so by consuming the `crate_name` element of `Type::External` - but best I can tell, this implementation ignores that and I don't quite understand why. (that crate_name impl might need tweaking given the user might want the generated module name to differ from the crate name, but that seems true for everyone and not just swift/kotlin).
* The .md changes document `rust-crate-name` for Kotlin and Swift, but the string `rust-crate-name` appears only in that .md - none of the example seem to use it - so I'm not sure that actually works. uniffi.toml in the new fixture doesn't specify it.
* Assuming there's something simple I'm missing, is `rust-crate-name` the correct name? In `[External="custom-types"]`, the string `custom_types` is the crate name, and what ends up in `Type::External` - so it might be the case that we want to customize the name of the generated module - but doesn't Python also have that problem? But as above, I also don't understand why the crate_name part of `Types::External` appears unused here.

fixtures/ext-types/lib-with-seperate-swift-modules/Cargo.toml
bendk: Yes this is testing only.  In the real consumer situation, UniFFI will generate a set of `.swift` files and it's up to you to decide how to compile them.  You can either compile them as 1 module or as separate modules.  I wonder if UniFFI should be less hands-off about this, but don't know what changes to suggest.

bendk:
> You mention there are 2 cases - the first case is what we do now in app-services, but the second case isn't clear to me - what is that second case supporting exactly and why? ie, why do we have 2 scenarios for swift but not for others (eg, I could imagine Python theoretically wanting those 2 scenarios too). There is a new fixture for the other scenario, but I can't see much different between that and the existing fixture except for a uniffi-testing attribute, so I'm slightly confused here.

The second case is what we used to do for app-services and Sammy/Tarik mentioned maybe wanting to go back to that model in the future.  I think the current model has potential for name collisions.  Based on that I figured the UniFFI code should be flexible enough to handle them both.

Scenario 1 seems very Swift specific since multiple `.swift` files all share the same namespace.

> That second case, where "Each generated .swift file is complied into a separate module." seems to be very similar to what Python does now - the existing uniffi-fixture-ext-types fixture generates multiple .so and multiple .py files, and does so by consuming the crate_name element of Type::External - but best I can tell, this implementation ignores that and I don't quite understand why. (that crate_name impl might need tweaking given the user might want the generated module name to differ from the crate name, but that seems true for everyone and not just swift/kotlin).

Swift imports are weird.  The default import works like `from [module] import *`, which is what I ended up using.  The first draft of this did it in a weird way, now it's handled in `ExternalType.swift`.  There's also a way to import a single name, but then you need to specify if the item kind (class, enum, etc.) and we don't have that information about external types.

> Assuming there's something simple I'm missing, is rust-crate-name the correct name? In [External="custom-types"], the string custom_types is the crate name, and what ends up in Type::External - so it might be the case that we want to customize the name of the generated module - but doesn't Python also have that problem? But as above, I also don't understand why the crate_name part of Types::External appears unused here.

Yup, Python has the same problem but I guess it hasn't been an issue in practice.  Should we fix that in this PR?  I don't think it will be too hard.

mhammond:
Thanks for the explanation, that makes sense.

> Yup, Python has the same problem but I guess it hasn't been an issue in practice. Should we fix that in this PR? I don't think it will be too hard.

I don't think we should TBH - we might not get it quite correct, so when a real consumer has this issue we might find we didn't quite nail it.

I vaguely feel the same about Swift - if we don't have an actual consumer for this, are we better off waiting to support this use-case until we do?

bendk:
> I vaguely feel the same about Swift - if we don't have an actual consumer for this, are we better off waiting to support this use-case until we do?

I thought that app-services was waiting for this to implement some more types as external.  Maybe we could use that as a test case for this code?  But I don't think this is very urgent, I'm happy to have it sit for a while too.

mhammond:
IOW, this patch adds 2 new swift tests - IIUC, app-services is using, roughly, what `fixtures/ext-types/lib/tests/bindings/test_imported_types.swift` is testing, so can't it keep doing that, or must it move to the model used by `fixtures/ext-types/lib-with-separate-swift-modules/tests/bindings/test_imported_types.swift`?

bendk:
> IOW, this patch adds 2 new swift tests - IIUC, app-services is using, roughly, what `fixtures/ext-types/lib/tests/bindings/test_imported_types.swift` is testing, so can't it keep doing that, or must it move to the model used by `fixtures/ext-types/lib-with-separate-swift-modules/tests/bindings/test_imported_types.swift`?

My goal is to support both models with config options.  That way we can keep using the current one, but also could switch if we want.  I think part of the confusion was how the configuration worked.  I updated the code and the docs, hopefully it's clearer now.

badboy:
> IOW, this patch adds 2 new swift tests - IIUC, app-services is using, roughly, what `fixtures/ext-types/lib/tests/bindings/test_imported_types.swift` is testing, so can't it keep doing that, or must it move to the model used by `fixtures/ext-types/lib-with-separate-swift-modules/tests/bindings/test_imported_types.swift`?

Can you point me to where app-services currently uses that?

badboy (commented):
👍 for the support.
The configuration to swap between one/multiple modules is not so complex, so I won't block on it, but I'm with Mark in that I would like to understand how/why/how app-services will end up using it.

bendk:
Thanks for the cleanup!

mhammond:
> > IOW, this patch adds 2 new swift tests - IIUC, app-services is using, roughly, what `fixtures/ext-types/lib/tests/bindings/test_imported_types.swift` is testing, so can't it keep doing that, or must it move to the model used by `fixtures/ext-types/lib-with-separate-swift-modules/tests/bindings/test_imported_types.swift`?
> 
> Can you point me to where app-services currently uses that?

Sorry I was a bit vague - to restate my comment:

* app-services definitely wants external types and is generally blocked on being able to make some nice cleanups due to the lack of swift support (eg, there are copy/pasted enums and structs we could fix with external types)

* IIUC, this PR *also* introduces an alternative way of building swift packages with uniffi which is different than the way app-services currently does. IMO, we don't want or need this alternative in this PR - I just want app-services as it exists today to get external types and I'd rather leave the ability to change how swift is packaged to an issue/PR which is dedicated to that instead of conflated with the concrete requirement we have today.

bendk:
> > IOW, this patch adds 2 new swift tests - IIUC, app-services is using, roughly, what `fixtures/ext-types/lib/tests/bindings/test_imported_types.swift` is testing, so can't it keep doing that, or must it move to the model used by `fixtures/ext-types/lib-with-separate-swift-modules/tests/bindings/test_imported_types.swift`?
> 
> Can you point me to where app-services currently uses that?

I think it happens in https://github.com/mozilla/rust-components-swift

bendk:
Splitting up the PR makes sense to me.  I updated this one to assume that the source files will be compiled together into a single module, which simplifies it quite a bit.  I think I addressed the other comments as well, but tell me if I missed anything.

badboy (approved):

mhammond (approved):
Thanks!

PR #1392, opened 2022-11-01
Add basic documentation for the proc-macro frontend

jplatte:


docs/manual/src/proc_macro/index.md
```diff
@@ -0,0 +1,121 @@
+# Experimental: Attributes and Derives
+
+UniFFI is in the process of having its interface definition mechanism rewritten to avoid the
+duplication of function signatures and type definitions between Rust code and the UDL file.
```
Hywan, line 4: duplication _and_ out-of-sync

docs/manual/src/proc_macro/index.md
```diff
@@ -0,0 +1,121 @@
...
+## The `uniffi::Record` derive
+
+The `Record` derive macro exposes a `struct` with named fields over FFI. All types that are
+supported as parameter and return types by `#[uniffi::export]` are also supported as field types
+here.
+
+It is permitted to use this macro on a type that is also defined in the UDL file, as long as all
+field types are UniFFI builtin types; user-defined types might be allowed in the future. You also
+have to maintain a consistent field order between the Rust and UDL files (otherwise compilation
+will fail).
```
Hywan, line 64: Can you add an example here please?

docs/manual/src/proc_macro/index.md
```diff
@@ -0,0 +1,121 @@
...
+
+## The `uniffi::Enum` derive
+
+The `Enum` derive macro works much like the `Record` derive macro. Any fields inside variants must
+be named. All types that are supported as parameter and return types by `#[uniffi::export]` are
+also supported as field types.
+
+It is permitted to use this macro on a type that is also defined in the UDL file as long as the
+two definitions are equal in the names and ordering of variants and variant fields, and any field
+types inside variants are UniFFI builtin types; user-defined types might be allowed in the future.
```
Hywan, line 74: Can you add an example here please?

docs/manual/src/proc_macro/index.md
```diff
@@ -0,0 +1,121 @@
...
+    fn method_b(&self) {
+        // ...
+    }
+}
+```
+
+## Other limitations
+
+In addition to the per-item limitations of the macros presented above, there is also currently a
+global restriction: You can only use the proc-macros inside of a crate whose name is the same as
```
Hywan, line 120:
```suggestion
global restriction: You can only use the proc-macros inside a crate whose name is the same as
```

Hywan (commented):
Thanks for writing this documentation.

I think we should make it clear that a UDL file is not necessary. It can be an addition to proc-macro, but it's not mandatory.

jplatte:
Well, currently the UDL file is actually still mandatory. It shouldn't be hard to make it optional though.

bendk (commented):
This seems like a good attempt at documentation, but the current state of affairs where we're mixing UDL and proc-macros seems fundamentally confusing and hard to explain.

I think I'd prefer this to be present in the documentation, but not linked in the TOC yet.  This way, we have the docs checked in and we can continue to iterate on them, but we're not suggesting that new users switch to using the proc-macros yet.

Or maybe we should be suggesting that newer users use the proc-macros?  What do you think @jplatte?

jplatte:
I don't think we should be recommending new people to use the proc-macros yet, since you still need the UDL file and a few important things are missing (at least callback interfaces and errors). I think it is useful to have this linked, but clearly marked as experimental and incomplete. In addition to the title I guess I could put a warning on the page?

bendk:
That'd be okay for me.  What do other uniffi devs think?

@mhammond @jhugman @badboy @tarikeshaq @skhamis

mhammond (approved):
I agree - I think the tone of the new docs (thanks!) are exactly correct.

PR #1395, opened 2022-11-02
proc-macro: Fix nested user-defined types in record and variant fields

jplatte:


skhamis (approved):
Excited to start seeing some of the proc-macro stuff come to fruition! LGTM!

PR #1396, opened 2022-11-03
Fix ruby time support

saks:
When using a ruby adapter we noticed an issue with 1 second difference. I was able to reproduce the issue with tests but had to add more methods into `chronological.udl`. Please let me know if you have any questions.

### DIFF
--- a/uniffi_bindgen/src/bindings/ruby/templates/RustBufferBuilder.rb
+++ b/uniffi_bindgen/src/bindings/ruby/templates/RustBufferBuilder.rb
@@ -113,7 +113,12 @@ class RustBufferBuilder
     seconds = v.tv_sec
     nanoseconds = v.tv_nsec
 
-    if seconds < 0
+    # UniFFi conventions assume that nanoseconds part has to represent nanoseconds portion of
+    # duration between epoch and the timestamp moment. Ruby `Time#tv_nsec` returns the number of
+    # nanoseconds for the subsecond part, which is sort of opposite to "duration" meaning.
+    # Hence we need to convert value returned by `Time#tv_nsec` back and forth with the following
+    # logic:
+    if seconds < 0 && nanoseconds != 0
       # In order to get duration nsec we shift by 1 second:
       nanoseconds = ONE_SECOND_IN_NANOSECONDS - nanoseconds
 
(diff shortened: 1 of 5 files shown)


uniffi_bindgen/src/bindings/ruby/templates/RustBufferBuilder.rb
```diff
@@ -114,7 +114,11 @@ def write_{{ canonical_type_name }}(v)
     nanoseconds = v.tv_nsec
 
     if seconds < 0
+      # In order to get duration nsec we shift by 1 second:
```
bendk, line 117: I think this comment can be improved to describe this shift a bit more precisely.  IIUC, the issue is that Ruby and UniFFI represent times before the epoch differently.  For Ruby, the nanoseconds value is it's own positive value that should always be added to the time.  For UniFFI, the nanoseconds value is part of the number, so it should be subtracted when the seconds are negative.   BTW, I'd guess that more thought was put in to the Ruby representation and wonder if it has advantages over ours.

Also, I think this shift should be skipped when `nanoseconds=0`.  Since in that case you're going to end up with `nanoseconds=10^9`, which overflows it's bounds.

bendk (changes requested):
The changes look good, but I'm hoping you can update the comments a bit and also fix a corner case.

uniffi_bindgen/src/bindings/ruby/templates/RustBufferBuilder.rb
saks: Thanks for looking into this @bendk. Indeed ruby implementation is weird and the main reason is the lack of duration type/class in the standard library. One may think that `Time` can be used as a replacement, but unfortunately ruby doesn't support math operations between two time objects:
```ruby
[5] pry(main)> Time.at(0) + Time.at(1)
TypeError: time + time?
from (pry):5:in `+'
```
Hence I had to figure out a way to convert properties of time object into values that ffi type conversion layer can interpret correctly. I've added comments clarifying this appraoch. Please let me know if these are helpful.

I appreciate your point about the corner case with `nanoseconds=0`. No shift is necessary for that scenario, I've updated the code. One issue though is that rust code seems handles overflow gracefully so I can't find a way to reproduce it in tests.

bendk (approved):
The comments look great, thanks for updating them!

PR #1397, opened 2022-11-07
Add proposal to generate submodules.

gogo2464:


gogo2464:
Fell free to assign me to my own PR please.

Best regards.

docs/adr/0008-split-bindings-to-modules-automatically.md
```diff
@@ -0,0 +1,64 @@
+# [short title of solved problem and solution]
+
+* Status: [proposed superseded by [gogo](0008-split-bindings-to-modules-automatically.md)]
+* Deciders: []
+* Date: [2022-07-11 when the decision was last updated]
+
+## Context and Problem Statement
+
+In his current state, uniffi only allow to import any generated code to a single file. This is embarassing because in languages used to make bindings, the user has to import submodules.
```
bendk, line 9: Please add more about your use case here.  How is you hand-written code structured and how do you want the generated module(s) to fit in to that?

Also can you add give describe the drawbacks in a more precise way than "embarrassing".  Is it a time sink in some way?  Does it limit what UniFFI can do?

docs/adr/0008-split-bindings-to-modules-automatically.md
```diff
@@ -0,0 +1,64 @@
...
+        interface SmallHouse {
+            constructor();
+            [Name=build_house]
+            constructor();
+            string get_house_street();
+        };
+    }
+}
+
+ Recommended.
```
bendk, line 27: There's a 3rd option here: Describe the module structure in `uniffi.toml`, which would be my first instinct to use.  The UDL file is shared between all bindings languages, so it's best used for things that can be generalized across all languages.  The `uniffi.toml` file has a section for each bindings languages, so it's best used for things specific to a language.

I'm pretty sure packages/modules are going to need to be described per-language.  The example you provided seems very python-specific to me.

docs/adr/0008-split-bindings-to-modules-automatically.md
```diff
@@ -0,0 +1,64 @@
+# [short title of solved problem and solution]
+
+* Status: [proposed superseded by [gogo](0008-split-bindings-to-modules-automatically.md)]
+* Deciders: []
+* Date: [2022-07-11 when the decision was last updated]
+
+## Context and Problem Statement
+
+In his current state, uniffi only allow to import any generated code to a single file. This is embarassing because in languages used to make bindings, the user has to import submodules.
+
```
bendk, line 10: I have a related question here, are you working with 1 component/UDL file or multiple at once?  I've thought about having UniFFI generate a multi-module directory structure, but that's because I'm working on a multi component project that has multiple UDL files.  Whenever I've thought about doing this, it's been linked to having `uniffi-bindgen` work on multiple UDL files at once which has several potential benefits.

bendk (changes requested):
I think this is has potential, but it needs to be fleshed out.  I think the most important thing it needs is a problem statement that describes specific use cases and explains why the current UniFFI has problems dealing with them.

docs/adr/0008-split-bindings-to-modules-automatically.md
gogo2464: I pushed once again. Fell free to get a look.

docs/adr/0008-split-bindings-to-modules-automatically.md
gogo2464: Good point. I am going to mention the way with several UDL files in order to have several python/kotlin modules. It is less pythonic.

docs/adr/0008-split-bindings-to-modules-automatically.md
gogo2464: I doubt. I mentioned two ways in the file in order to argue on.

docs/adr/0008-split-bindings-to-modules-automatically.md
```diff
@@ -0,0 +1,77 @@
...
+In his current state, uniffi only allows to import any generated code to a single file. This is hard to read and to maintain for a binding library user because in languages used to make bindings, the user has to import submodules.
+
+Currently, I have to write:
+
+from cryptatools_core.python3_bindings import CaesarNumberAlgorithm, Encoding, Alphabet
+
+instead of:
+
+from cryptatools_core.cryptography.encryption.monoalphabetic_ciphers import CaesarNumberAlgorithm
+from cryptatools_core.utils.alphabets import Encoding, Alphabet
```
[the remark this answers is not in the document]
gogo2464: I think. uniffi-bindgen should be run only one time but should take all udl files as input together.

gogo2464:
Does it sounds ready to merge?

mhammond:
I'm mildly against this. `from generated_module import *` seems reasonable to me, and adding the capability for Python to have nested modules is against [the zen of python](https://peps.python.org/pep-0020/) which explicitly says "flat is better than nested" ;) - but more importantly, I don't think we should look at just Python in isolation. As mentioned, why not just use multiple UDL files - they already have a "namespace", so introducing another namespace-like concept just for Python seems short sited and adds additional complexity for what I see, frankly, as a non-problem.

IOW, I think any such proposal should explicitly address all languages and be very clear about what other alternatives exist (such as multiple UDL files)

gogo2464:
When I code in python, I alway use libraries split in several files and folders. I never met `from module import *` except in pwntools but in pwntools, we have the ability to switch to `from module.submodule import *`.

I mentioned the possibility to use multiple UDL files. I am confused. Does it already exists? I made a test with:

`
fn main() {
    uniffi_build::generate_scaffolding("./src/cryptatools.udl").unwrap();
    uniffi_build::generate_scaffolding("./src/caesar_number.udl").unwrap();
}
`

Currently it seems the user has the error `"uniffi_reexport_scaffolding"redefined here`. It could be a good improvement.

docs/adr/0008-split-bindings-to-modules-automatically.md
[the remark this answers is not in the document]
bendk: Having uniffi-bindgen process multiple files has a lot of benefits, but also would require quite a bit of work to implement.  Is it worth it for your use case?

Right now you could use a two-liner like
```
uniffi_bindgen generate --language python --out-dir cryptatools_core/cryptography/encryption monoalphabetic_ciphers.udl
uniffi_bindgen generate --language python --out-dir cryptatools_core/utils alphabets.udl
```

We could make a system where one invocation of `uniffi_bindgen` processed both UDL files, but then how would you specify which UDL file mapped to which `out-dir`?  I think you would have to configure it in the `uniffi.toml` file, but it's not clear to me that that's simpler than using the two-liner.

bendk (changes requested):
The ADR review process is where we discuss the proposal, weigh the pros and cons, and also try to make it as clear as possible.  We can't merge this until we've agreed on the direction and we believe the ADR clearly communicates the points.

I'm still struggling to understand how bad the current state of affairs is for you.  Option 4 seems like a good choice and I believe it's possible now, it just requires running `uniffi-bindgen` multiple times.  Checking in a script file that does that seems like a reasonable way to achieve this without any changes to UniFFI.

gogo2464:
Yes I am ready to make the efforts to code this implementation!

At this moment, Even your first command will fail because we have to put the code:


````
fn main() {
    uniffi_build::generate_scaffolding("./src/cryptatools.udl").unwrap();
    uniffi_build::generate_scaffolding("./src/caesar_number.udl").unwrap();
}```

And there could be only a single uniffi_build::generate_scaffolding to evoid redefinition of some elements.

Of course, I can fix it. ;)

File: docs/adr/0008-split-bindings-to-modules-automatically.md (hunk 4 above)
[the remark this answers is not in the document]
Comment (gogo2464) [reply docs/adr/0008-split-bindings-to-modules-automatically.md:18]: I can only agree with you. I planned to simply copy the current rust project folders tree to parse all the UDL file.  I love rust philosophy. And I think a rewrite of a Makefile like file to parse the project is just a very bad idea and must be avoided.

**Comment by bendk** [comment]
> When I code in python, I alway use libraries split in several files and folders. I never met `from module import *` except in pwntools but in pwntools, we have the ability to switch to `from module.submodule import *`.
> 
> I mentioned the possibility to use multiple UDL files. I am confused. Does it already exists? I made a test with:
> 
> ```
> fn main() {
>     uniffi_build::generate_scaffolding("./src/cryptatools.udl").unwrap();
>     uniffi_build::generate_scaffolding("./src/caesar_number.udl").unwrap();
> }
> ```
> 
> Currently it seems the user has the error `"uniffi_reexport_scaffolding"redefined here`. It could be a good improvement.

Ahh yes, this isn't possible currently because of the scaffolding issue you mention.  One approach would be to try to allow `uniffi_build::generate_scaffolding` to input multiple UDL files, but I wouldn't go that way since that solution might be obsolete once the proc-macro work is finished.

My other suggestion to make this work with the current UniFFI, is to auto-generate a single module, but re-export the symbols in a different module, similar to the Glean example.  That would give you lots of control on where the items ended up.  For example:

- Generate the code to `cryptatools_core/uniffi.py`
- `cryptatools_core/cryptography/encryption/monoalphabetic_ciphers.py`
  ```
  from crypatools_core.uniffi import CaesarNumberAlgorithm
  ```
- `cryptatools_core/utils/alphabets.py`:
  ```
  from cryptatools_core.uniffi import Encoding, Alphabet
  ```

How would this work for you?

**Comment by gogo2464** [comment]
It is basically already the first implementation idea that I mention. It sounds to be the best idea under one condition. With the difference that the different python imports will be self-generated by uniffi. I do not wish that the user need to write one thousands of imports as mentionned in my very first proposal.


Sounds to be a perfect compromise. Just waiting for your approval in order to start coding using this proposal.

**Comment by mhammond** [comment]
> Just waiting for your approval in order to start coding using this proposal.

Are you referring to Ben's idea that starts with "My other suggestion to make this work with the current UniFFI ..."? If so, then I don't think you need to wait for anything as it should be doable with the current UniFFI. If not, can you please clarify?

**Comment by gogo2464** [comment]
In order to clarify, I think I will make a PR on: [./uniffi_bindgen/src/bindings/](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings) to generate Python import into different folders.

Of course I will do same for each language.

**Comment by gogo2464** [comment]
The issue is:
-Is what I want just a tempoarly fix before the new macro system will be released?
-I also need to parse file directories in order to get modules file tree. I might parse the rust itself to get same imports. But is it a real good idea? Sounds very headache. May be a better way.

**Comment by mhammond** [comment]
I think it depends on your timing - any such proposal will probably require quite a bit of discussion, including how it impacts other languages, and will need to take into account existing consumers. If you are looking for something soon, I suggest you take Ben's option and hand-code the "sub modules" yourself. If that really sounds like too much work, then in the short term I suggest you just live with the flat namespace.

**Comment by gogo2464** [comment]
I am fully conscious on how the free software works and on the consumer you could have.

From my part, I suggest to first, use the bdk method with no uniffi PR at this moment. And then when the release with proc-macro will be finished, I will be able to start merging one of the three other ways.

What do you think of the second point?

**Comment by gogo2464** [comment]
> > And then when the release with proc-macro will be finished, I will be able to start merging one of the three other ways.
> 
> It's not clear when the proc-macro work will be "finished". Mozilla hasn't started to think about moving to it yet, so I suspect it will be a long time before it's the only option.
> 
> Re "one of the other three ways", the devil will be in the detail - I don't think there's a clear proposal yet, even with this PR.

I will clarify how. For the moment... Well I think I have the time.

**Comment by stan-irl** [comment]
I think theres a lot of value in this. Im working on a shared library for a startup with a lot of different features. I want to write and maintain different APIs for different features but only have a single binary to reduce the overall app size. Right now I just have to dump everything into the same namespace and same UDL file if I want it all in the same binary (if I understand correctly). 

It would be nice to allow everything to be placed inside a namespace block, not just top level functions, as well as allowing multiple namespace blocks per UDL file. Then you could have a file like:

shared-lib.udl
```
namespace feature1 {
    interface Feature1Manager {
        void do_feature_1_thing();
    }
}j

namespace feature2 {
    interface Feature2Manager {
        void do_feature_2_thing();
    }
}j
```

then if people really want multiple UDL files they can just concat them into a single file in the crate root as a preprocessing step. Although it would be convenient to have that as well.

mhammond:
> Im working on a shared library for a startup with a lot of different features. I want to write and maintain different APIs for different features but only have a single binary to reduce the overall app size. Right now I just have to dump everything into the same namespace and same UDL file if I want it all in the same binary (if I understand correctly).

You should check out the [mozilla application-services repo](https://github.com/mozilla/application-services/) - we do exactly as you describe - maintain close to 10 .udl files, each in different crates, and release them all in a single .so. Eg, https://github.com/mozilla/application-services/tree/main/megazords/full is how we build that .so for Android.

stan-irl:
@mhammond this is awesome. thankyou!

mhammond:
Your welcome - I hope uniffi works out for you. I'll close this for now as it seems to have answered your question.

mhammond:
eek, I think I closed this prematurely given the earlier comments, so I'll reopen it in the hope a specific proposal is still in progress.

gogo2464:
> > Im working on a shared library for a startup with a lot of different features. I want to write and maintain different APIs for different features but only have a single binary to reduce the overall app size. Right now I just have to dump everything into the same namespace and same UDL file if I want it all in the same binary (if I understand correctly).
> 
> You should check out the [mozilla application-services repo](https://github.com/mozilla/application-services/) - we do exactly as you describe - maintain close to 10 .udl files, each in different crates, and release them all in a single .so. Eg, https://github.com/mozilla/application-services/tree/main/megazords/full is how we build that .so for Android.

wait... If I do not like to split my python files as here: https://github.com/gogo2464/cryptatools-rs/blob/master/cryptatools-core/bindings/python3/cryptatools-core/cryptography/encryption/monoalphabetic_cipher/caesar_number.py should I migrate to application service repo? It has even worst doc than uniffi.

mhammond:
application-services is just an example of a consumer which uses multiple UDL files in a single .so.

gogo2464:
> 

Yes But I want it for desktop application. Do you see what I would like ?

gogo2464:
@mhammond I just finished another project. I can provide more time now.

gogo2464:
@bendk I thank a lot of about this issue. 

About this proposal: https://github.com/mozilla/uniffi-rs/pull/1397#discussion_r1016690716 what do you think of add a new subcommand such as: `uniffi_bindgen generate-recursive --language python` ?

I am not convinced that I need the other repo mentionned by @mhammond . In case of doubt you could contact me on slack.

mhammond:
I don't think anyone really understands what is being proposed here. In the attached file, the problem is:

gogo2464:
> IBut if you do want a change to uniffi, it's quite important we understand exactly what you are proposing.


I can only agree with you. I definitely need to learn to be more clear.

gogo2464:
@mhammond thank you very much for according me some time.

the generate-recursive subcommand should look for any .udl files and then run uniffi_bindgen generate for each one.

Should I provide more information?

mhammond:
I guess I wouldn't necessarily object to this if the change was simple, but it doesn't seem particularly useful to me to be honest. How many udl files do you have? Listing explicitly which UDL files should be processed sounds like better practice to me, but if you really wanted to auto-find any UDL file under some directory can't you just use shell tools (eg, find)?

gogo2464:
Absolutely. I was focusing on the option 2. I think it is the best. not the 4.

mhammond:
This is quite stale now - feel free to re-open if you update this based on the comments above.

PR #1399, opened 2022-11-10
Remove trailing spaces that break rustfmt

bancek:
Callback interfaces with long names break formatting with the latest version of `uniffi_bindgen` and `rustfmt`.

Example `.udl` file:

```
callback interface CallbackWithALongName {
  void test();
};

namespace test {
  void test(CallbackWithALongName callback);
};
```

Screenshot 2022-11-10 at 12 34 22

```sh
$ uniffi-bindgen scaffolding src/test.udl
error[internal]: left behind trailing whitespace
  --> .../src/test.uniffi.rs:71:71:1
   |
71 |         
   | ^^^^^^^^
   |

error[internal]: left behind trailing whitespace
  --> .../src/test.uniffi.rs:73:73:1
   |
73 |         
   | ^^^^^^^^
   |

warning: rustfmt has failed to format. See previous 2 errors.

Error: rustmt failed when formatting scaffolding. Note: --no-format can be used to skip formatting

$ uniffi-bindgen --version
uniffi-bindgen 0.21.0

$ rustfmt --version
rustfmt 1.5.1-stable (a55dd71d 2022-09-19)
```

This is a bug in `rustfmt` but it will be faster to fix it here.

https://github.com/rust-lang/rustfmt/issues/2896
https://github.com/rust-lang/rustfmt/issues/5391

PR #1401, opened 2022-11-10
Fix another case where trailing space that breaks rustfmt

bancek:
One more fix like #1399. It breaks with multiple arguments that are too long.


mhammond:
Not sure what's going on with CI, so I'll try closing and reopening this.

bancek:
Can you try again? I've authorized CircleCI now.

mhammond (approved):
(I don't think PR authors each need to auth circleci, but also not sure what is going on)

Regardless, I agree with @badboy that fixing all these is whack-a-mole, but this is simple and fine)

PR #1404, opened 2022-11-15
Swift external type module config

bendk:
This is based off of https://github.com/mozilla/uniffi-rs/pull/1389.  Let's merge that first.  [Here are the changes compared to that one](https://github.com/bendk/uniffi-rs/compare/swift-external-types...bendk:uniffi-rs:swift-external-type-module-config): 

Added support for build processes where each generated .swift file is compiled into its own module rather than compiling all files together. This means that we need imports to make external types work.

-    Added test fixture and updated the testing code to test this
-    Added config code to allow the user to choose their mode
-    Updated the docs to describe the config options


stan-irl:
Are there any plans to merge this?

I think it would really help to make it a lot simpler to namespace and organise generated swift code. it seems very primitive to have to put everything in a single namespace

bendk:
I'll be honest, I don't remember what the point of this one is and at this point it's very bit-rotted.  If there's still a problem to be address, then please comment and/or open an issue for it.  However, I don't think this PR is useful anymore.

PR #1405, opened 2022-11-17
Upgrade goblin to 0.6

messense:
Changelog: https://github.com/m4b/goblin/blob/master/CHANGELOG.md#060---2022-10-23


mhammond (approved):
Thanks!

PR #1406, opened 2022-11-17
Change FFI to Ffi in type names

jplatte:
In accordance with the Rust API Guidelines: https://rust-lang.github.io/api-guidelines/naming.html

If you agree with this, I'm happy to update this PR to also change `FFIArgument`, `FFIFunction` and any others.


jhugman (approved):
This looks like a straightforward and uncontroversial. Thank you!

✅

jplatte:
So there's a few `UniFFI`s in symbol names that I'm not sure what to do with, but everything else is updated.

PR #1408, opened 2022-11-22
proc-macro: Add support for fallible functions

jplatte:
Part of #1257.

Remaining tasks:

- [x] `Error` derive macro
- [x] Documentation
- [x] Testing


uniffi_bindgen/src/interface/error.rs
```diff
@@ -135,6 +135,20 @@ impl Error {
     }
 }
 
+impl From<uniffi_meta::ErrorMetadata> for Error {
+    fn from(meta: uniffi_meta::ErrorMetadata) -> Self {
+        let flat = meta.variants.iter().all(|v| v.fields.is_empty());
```
jplatte, line 140: I found this is actually wrong, and am working on a fix (I'm doing a separate commit though, so review-ability should not be harmed).

jplatte:
I was able to rerun from the CircleCI website but it doesn't seem to change anything 🤷🏼

uniffi_macros/src/error.rs
```diff
@@ -0,0 +1,170 @@
...
+            .collect(),
+        Err(e) => e.into_compile_error(),
+    };
+    let attr_error = attr.err().map(syn::Error::into_compile_error);
+
+    quote! {
+        #ffi_converter_impl
+
+        #[automatically_derived]
+        impl ::uniffi::FfiError for #ident {}
```
badboy, line 65: I have been trying to understand why we have this trait.
Is this just a marker?
It doesn't seem to be used anywhere or am I missing something?

docs/manual/src/proc_macro/index.md
```diff
@@ -154,6 +154,56 @@ impl Foo {
 }
 ```
 
+## The `uniffi::Error` derive
+
+The `Error` derive registers a type as an error that be the `E` in a `Result<T, E>` return type of
+a `#[uniffi::export]`ed function or method. Used on its own, it can be used on any enum that the
+`Enum` derive also accepts, and exposes any variant fields to the foreign code. The generated
+foreign function will have the result's `T` as its return type, and throw the error in case the Rust
+call returns `Err(e)`.
```
badboy, line 163: This reads a bit weird.

Maybe we can rephrase that.

```suggestion
The `Error` derive registers a type as an error and can be used on any enum that the `Enum` derive also accepts.
It exposes any variant fields to the foreign code. 
This type can then be used as the `E` in a `Result<T, E>` return type of an exported function or method.
The generated foreign function for an exported function with a `Result<T, E>` return type
will have the result's `T` as its return type and throw the error in case the Rust call returns `Err(e)`.
```

I tried to make it more concise and re-order the bits a bit, but feel free to adjust that again.

(fwiw I generally recommend [semantic linebreaks](https://sembr.org/), makes further editing and reviewing easier most of the time.)

docs/manual/src/proc_macro/index.md
```diff
@@ -154,6 +154,56 @@ impl Foo {
...
+#[uniffi::export]
+fn do_thing() -> Result<(), MyError> {
+    // ...
+}
+```
+
+You can also use the helper attribute `#[uniffi(flat_error)]` to expose just the variants but none
+of the fields, in this case the error will be serialized using Rust's `ToString` trait and will be
+accessible as the only field on each of the variants. In this case, your variants can have unnamed
+fields, and the types of the fields don't need to implement any special traits.
```
badboy, line 187:
```suggestion
You can also use the helper attribute `#[uniffi(flat_error)]` to expose just the variants but none
of the fields.
In this case the error will be serialized using Rust's `ToString` trait 
and will be accessible as the only field on each of the variants. 
For flat errors your variants can have unnamed fields, and the types of the fields don't need to implement any special traits.
```

docs/manual/src/proc_macro/index.md
```diff
@@ -154,6 +154,56 @@ impl Foo {
...
+    }
+}
+
+#[uniffi::export]
+fn do_thing() -> Result<(), MyError> {
+    // ...
+}
+```
+
+You can also use the helper attribute `#[uniffi(flat_error)]` to expose just the variants but none
```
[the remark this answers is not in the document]
jplatte: Oops 😅

Issue #1416, opened 2022-11-27
Split cli out of uniffi_bindgen


messense:
`uniffi_bindgen` as a library is quite bloated at the moment, for example

```toml
[build-dependencies]
uniffi_build = {version = "0.8", features = [ "builtin-bindgen" ]}
```

`builtin-bindgen` would depends on `clap`. 

I wonder if it's acceptable to split the cli out to a separate crate say `uniffi_bindgen_cli` like [what bindgen does](https://github.com/rust-lang/rust-bindgen/tree/master/bindgen-cli)? 

I'm using `uniffi_bindgen::generate_bindings` function in [maturin](https://github.com/PyO3/maturin/blob/4bceaa57039e00f3063c918a578730dd09e4308e/src/module_writer.rs#L878) and would like to avoid pulling in clap 3.x.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-215)
┆Issue Number: UNIFFI-215

badboy:
With 0.23 we essentially have this back if one uses only the `uniffi` dependency with feature `cli` as documented.
It's possible to avoid that by directly depending on `uniffi_bindgen` right now. 
Maybe we should make it possible to avoid clap there too?

PR #1417, opened 2022-11-28
Make clap optional in uniffi_bindgen

messense:
A less disruptive approach compared to #1416, same as https://github.com/eqrion/cbindgen/pull/363


jplatte:
I like this as a super small & simply stop-gap, it could land as-is regardless of what the next steps are :)

bendk:
Good point about the changes being small.  I don't want to block this one or anything.

badboy (approved):
Let's land this then, @bendk's work can still go on and we can figure out how to not regress this there.

PR #1427, opened 2022-12-16
Make binding checksums more deterministic across platforms and versions of rust

glandium:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1805809.

Notes:
- I made the result of the checksum as close as std::hash::Hash on 64-bits little-endian platforms such that bindings generated before this change on such a platform (quite likely) doesn't change.
- I also have a checksum-0.21 branch with the same changes against what Firefox has in tree in 108/109, which would be great to release as a 0.21.1 for an uplift.


uniffi_checksum_derive/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "uniffi_checksum_derive"
```
jplatte, line 2: Why a new crate instead of putting the derive macro into `uniffi_macros`?

glandium: Because it's not a macro meant to be used as part of the public API, but more importantly, because uniffi_macros depends on uniffi_meta, which would make it circular, although technically speaking, it's also possible to not make uniffi_meta reexport the derive, but that's less convenient.

uniffi_checksum_derive/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+//! Custom derive for uniffi_meta::Checksum
+
+use proc_macro::TokenStream;
+use quote::{format_ident, quote};
+use syn::{parse_macro_input, Attribute, Data, DeriveInput, Expr, ExprLit, Fields, Index, Lit};
+
+fn has_ignore_attribute(attrs: &[Attribute]) -> bool {
+    attrs.iter().any(|attr| {
+        let name = attr.path.get_ident().map(|i| i.to_string());
+        if name.as_deref() == Some("checksum_ignore") {
```
jplatte, line 15:
```suggestion
        if attr.path.is_ident("checksum_ignore") {
```

uniffi_checksum_derive/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+                match &variant.fields {
+                    Fields::Unnamed(fields) => {
+                        let field_idents = fields
+                            .unnamed
+                            .iter()
+                            .enumerate()
+                            .map(|(num, _)| format_ident!("__self_{}", num))
+                            .collect::<Vec<_>>();
+                        let field_stmts = field_idents
+                            .iter()
```
jplatte, line 71: Should work too, I think:

```suggestion
                            .map(|(num, _)| format_ident!("__self_{}", num));
                        let field_stmts = field_idents
                            .clone()
```

uniffi_checksum_derive/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+                    }
+                    Fields::Named(fields) => {
+                        let field_idents = fields
+                            .named
+                            .iter()
+                            .map(|field| field.ident.as_ref().unwrap())
+                            .collect::<Vec<_>>();
+                        let field_stmts = field_idents
+                            .iter()
+                            .map(|ident| quote! { Checksum::checksum(#ident, state); });
```
jplatte, line 88: Same here.

uniffi_checksum_derive/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+                #(#stmts)*
+            }
+        }
+        Data::Union(_) => {
+            panic!("#[derive(Checksum)] is not supported for unions");
+        }
+    };
+
+    quote! {
+        impl #impl_generics Checksum for #name #ty_generics #where_clause {
```
jplatte, line 129:
```suggestion
        #[automatically_derived]
        impl #impl_generics Checksum for #name #ty_generics #where_clause {
```

badboy (approved):

jplatte (approved):

glandium:
I updated both the checksum and checksum-0.21 branches. How do we proceed for mozilla-central and mozilla-beta? I think central is still on 0.21. If that's the case, we may want to pull in the checksum-0.21 changes alone so that they can be uplifted to beta.

badboy:
> I updated both the checksum and checksum-0.21 branches on my repo. How do we proceed for mozilla-central and mozilla-beta? I think central is still on 0.21. If that's the case, we may want to pull in the checksum-0.21 changes alone so that they can be uplifted to beta.

I'm gonna land this here, then pull in checksum-0.21 separately to do a patch release, followed by a minor release.

0.21.1 can then go into m-c immediately (and be uplifted)
0.22 can follow on its own time.

PR #1436, opened 2022-12-28
Fix XCode exmaple not working.

Amosel:
This resolves 3 things that were not working on my fresh clone of the repo:

1. [As specified here](https://stackoverflow.com/questions/63607158/xcode-building-for-ios-simulator-but-linking-in-an-object-file-built-for-ios-f) arm64 macOS machines required (still) that the you add `"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = arm64;` that's done in Xcode, under the main target's **Build settings** in **Excluded Architecture**. 
2. Since the *Generated* folder in the Xcode is not tracked, and for my own lack of understanding of Xcode's build rules, I added a build scripts that to run `uniffi-bindgen` for both the `todolist` and `arithmetics` targets.
3. Both `todolist` and `arithmetics` are missing "staticlib" in their cargo .toml file.

Note:

You must run `cargo build` from the root directory in order to use the local `uniffi-bingen` that's in the `target/debug` folder

### DIFF
--- a/examples/app/ios/IOSApp.xcodeproj/project.pbxproj
+++ b/examples/app/ios/IOSApp.xcodeproj/project.pbxproj
@@ -208,6 +208,7 @@
 			isa = PBXNativeTarget;
 			buildConfigurationList = 395A8838251A062D00562F33 /* Build configuration list for PBXNativeTarget "IOSApp" */;
 			buildPhases = (
+				6C1D133F295C9B5600275D4D /* Generate Binding (incase build rule never ran) */,
 				CEEB59EB25263ECE003C87D1 /* Build Universal Binary for todolist */,
 				39ADC3D125E980F1000EE485 /* Build Universal Binary for arithmetic */,
 				395A880F251A062C00562F33 /* Sources */,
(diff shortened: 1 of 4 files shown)


badboy:
Thanks for the effort.
We should probably have UniFFI be a good example here and use a setup that others can replicate in Xcode.
Unfortunately while your changes likely fix the issues at hand, they are still hacky and better ways exist (which we use in other Mozilla projects).
See below for details.

> [As specified here](https://stackoverflow.com/questions/63607158/xcode-building-for-ios-simulator-but-linking-in-an-object-file-built-for-ios-f) arm64 macOS machines required (still) that the you add `"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = arm64;` that's done in Xcode, under the main target's **Build settings** in **Excluded Architecture**.

The better way is to actually specify the path to the right directory based on the target. That way builds will work, regardless of the platform you're on. I can see that I add that this week.

> Since the _Generated_ folder in the Xcode is not tracked, and for my own lack of understanding of Xcode's build rules, I added a build scripts that to run `uniffi-bindgen` for both the `todolist` and `arithmetics` targets.

I think the build rule should just take care this happens all the time? I'll check again.

> Both `todolist` and `arithmetics` are missing "staticlib" in their cargo .toml file.

Good point, that will be needed. Unfortunately this has big chances to break things, because the iOS targets now can build both cdylibs and staticlibs and then Xcode will bundle up the cdylib (with local paths) if you ever try to build the app for hardware devices.
(recent cargo now supports `--crate-type staticlib` so we can use that).

PR #1437, opened 2023-01-02
iOS Example: Fix the build to make it work again

badboy:
This includes:

* Correctly calling `uniffi-bindgen` through cargo.
* Ditching lipo for individual static libraries per target, that way it easily works on all Mac hardware (x86_64 or arm64) as well as for iOS hardware targets.
* Build a `staticlib` using cargo's new `--crate-type` option

---

Replaces #1436.
This wills locally on my M1 macbook using Xcode 14.2 and rustc 1.65.0


mhammond (approved):
This looks fine to me, but it is still "broken" due to https://github.com/mozilla/uniffi-rs/blob/main/rust-toolchain.toml forcing me on 1.63 - should we bump that first? (Or soon, I don't really care)

(I was going to suggest a change there might be to note "rustup override ..."  in README.md for that example, but that looks quite stale or confused, so 🤷 )

badboy:
> This looks fine to me, but it is still "broken" due to [`main`/rust-toolchain.toml](https://github.com/mozilla/uniffi-rs/blob/main/rust-toolchain.toml?rgh-link-date=2023-01-03T05%3A02%3A02Z) forcing me on 1.63 - should we bump that first? (Or soon, I don't really care)

D'oh! yeah, i've been testing with stable locally here of course. [`--crate-type` was stabilized in 1.64.0](https://github.com/rust-lang/cargo/pull/10838). We should just upgrade. I'll do that.

> (I was going to suggest a change there might be to note "rustup override ..." in README.md for that example, but that looks quite stale or confused, so 🤷 )

Of course I forgot to even look at the README. I'll also update that (as well as the book).

Issue #1439, opened 2023-01-03
Don't log CallbackInterface method calls


bendk:
We currently [log every time a callback interface method is called](https://github.com/mozilla/uniffi-rs/blob/ff195ec15c782a308ffe534150311cbd038ac8f9/uniffi_bindgen/src/scaffolding/templates/CallbackInterfaceTemplate.rs#L58).

This is an issue because one use-case for callback interfaces is to forward the Rust logs to a foreign application.  However, if we try to do that now we run into infinite recursion, since forwarding the log message generates a new log message.  Or worse, it may look like logging is working until you set the level to `DEBUG` and then everything totally breaks (and nothing useful is in the logs).

At some point we may want to figure out a better system in general for logging, but for now I think we should just remove that line.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-224)
┆Issue Number: UNIFFI-224

bendk:
Blocked on https://github.com/mozilla/uniffi-rs/issues/1439

PR #1440, opened 2023-01-03
Don't log callback interface method calls (#1439)

bendk:


fixtures/regressions/logging-callback-interface/tests/bindings/test.py
```diff
@@ -0,0 +1,24 @@
+# This is just a basic "it loaded and ran" test.
+# What we're really guarding against is failure to
```
mhammond, line 2: this comment seems like it's not quite accurate for this test?

mhammond (approved):
👍

PR #1444, opened 2023-01-13
Reworking FfiConverter type to work better with proc-macros

bendk:
Right now the UDL and proc-macro code generation are at odds with each other in how they handle `FfiConverter`. The UDL-based code generates a separate FfiConverter unit struct for each type we want to move across the FFI.  This allows us to generate FfiConverters for remote types without breaking the orphan rules.

However, it requires the code that generates scaffolding functions to know the full path to the FfiConverter struct.  This wasn't an issue with the UDL-based code, but it doesn't work with the proc-macro code. With proc-macros, you can have types with `#[uniffi::export]` defined outside of the crate root, and this breaks the system since we no longer know the path to the FfiConverter.

We've had some attemps at fixing this (#1250, #1373).  I hope this one sticks.

This basically splits the difference between the UDL system and the proc-macro system:
  - We define a single `UniFFIConverter` struct in the crate root.
  - That struct implements `FfiConverter<T>` for each type we want to move across the FFI.
  - Since we're implementing `FfiConverter<T>` on a local type, it gets around the orphan rules.
  - The scaffolding code can get the correct FfiConverter impl using `<crate::UniFFIConverter as ::uniffi::FfiConverter<T>>` and this works regerdless of if `T` was defined in the crate root or a submodule.


fixtures/proc-macro/src/proc-macro.udl
```diff
@@ -1,7 +1,6 @@
-namespace uniffi_proc_macro {};
+namespace uniffi_proc_macro { };
 
-// Test that redundant definition of a dictionary works
-dictionary Two {
-    string a;
-    sequence<boolean>? b;
+// Use this to test that types defined in the UDL can be used in the proc-macros
+dictionary Zero {
+    string inner;
```
bendk, line 5: With this change, you can no longer provide a redundant/duplicate definition between the UDL and proc-macros, but hopefully the story is still good enough:
  - You can define a type in the UDL and use it in the proc-macros, like in this test.
  - You can't `[uniffi:export]` a type in the proc-macro and use it in the UDL.

I think this should be okay for the transition period, especially since proc-macros support a lot of features at this point.  If there are issues, I can add code to handle the second case, but I'm hoping to avoid that.

uniffi/src/ffi_converter_impls.rs
```diff
@@ -0,0 +1,409 @@
...
+
+impl_via_ffi_for_num_primitive! {
+    i8, u8, i16, u16, i32, u32, i64, u64, f32, f64
+}
+
+/// Support for passing boolean values via the FFI.
+///
+/// Booleans are passed as an `i8` in order to avoid problems with handling
+/// C-compatible boolean values on JVM-based languages.
+unsafe impl<FC> FfiConverter<bool> for FC {
```
bendk, line 79: It's slightly weird that this is defining `FfiConverter<bool>` for all types, when we only need to target the `UniFFIConverter` unit structs defined in the scaffolding code.

To avoid that, we could define a trait named something like `FfiConverterDefault` and add `FC: FfiConverterDefault` bounds in these impls.  I wasn't sure if that simplified things or added confusion though.

uniffi/src/lib.rs
```diff
@@ -258,415 +225,83 @@ pub fn check_remaining(buf: &[u8], num_bytes: usize) -> Result<()> {
...
+/// Macro to implement lowering/lifting using a `RustBuffer`
 ///
 /// For complex types where it's too fiddly or too unsafe to convert them into a special-purpose
 /// C-compatible value, you can use this trait to implement `lower()` in terms of `write()` and
 /// `lift` in terms of `read()`.
-pub trait RustBufferFfiConverter: Sized {
-    type RustType;
-    fn write(obj: Self::RustType, buf: &mut Vec<u8>);
-    fn try_read(buf: &mut &[u8]) -> Result<Self::RustType>;
-}
```
bendk, line 395: This trait and the blanket impl below was annoying me, because small changes could lead to overlapping definitions.  I changed it to use a macro istead.

uniffi_bindgen/src/scaffolding/templates/scaffolding_template.rs
```diff
@@ -2,6 +2,13 @@
 // Trust me, you don't want to mess with it!
 {% import "macros.rs" as rs %}
 
+// Unit struct that handles lowering/lifting/serialization.  We define `uniffi::FfiConverter<T>` on
+// it for each T that we want to pass across the FFI
+//
+// This is pub, since we need to access it to support external types
+#[doc(hidden)]
+pub struct UniFFIConverter;
```
bendk, line 10: If we accept this PR, then we're committing to always generating a struct like this.  This means that once we support crates using proc-macros only and not UDL files, users will need to call a macro in the crate root to generate this.

I think this is fine and there's a good chance we will want to use it for other things.  It could also generate the `uniffi_types` module for example.

jplatte:
This is very much at odds with the kind of multi-crate setup we would like to have ultimately: An external crate could not just have compile-time conditional `uniffi` derives on its types that are then automatically picked up by downstream crates. I was actually pretty close to posting a PR that simply allows that and removes the `uniffi_types` module thing as well (I'm no longer certain it's actually needed for soundness).

bendk:
> This seems very much at odds with the kind of multi-crate setup we would like to have ultimately. An external crate could not just have compile-time conditional `uniffi` derives on its types that are then automatically picked up by downstream crates.

Very good point.  I originally had the types used in `FfiConverter` swapped around and I should have kept it that way.  So instead of `impl FfiConverter<MyType> for UniFFIConverter` you would define `impl FfiConverter<UniFFIConverter> for MyType`.  The advantage of this is that it allows the external crate setup you described, since they can write `impl<T> FfiConverter<T> for MyType` and have the implementation available downstream.

I switched the types because it seemed slightly more natural to me and also because I got confused about the external crate setup issue.  I'll update this PR to switch them back.

One issue with that is how to name the unit struct.  Calling it `UniFFIConverter` doesn't feel right to me.  In C++ I would call this a tag type, so maybe I'll name it `UniFFITag` or `LocalUniFFITag`.  I don't love either of those though.  I guess even the Rust orphan rules are easier than naming.

jplatte:
That's an interesting idea, I guess it could work for the transition period. I would really prefer for the proc-macro scaffolding to emit `<T as FfiConverter<()>>` by default though, so it doesn't require there to be UDL scaffolding that declares the `FfiConverterTag` or whatever we call it at the crate root. We could still have something like `#[uniffi::export(tag = crate::FfiConverterTag)]` to opt into impls for foreign types generated from UDL or even written by hand. WDYT?

bendk:
I think I could get behind a system that, but how would proc-macro code that generates scaffolding functions know which case to use for each argument?  I wouldn't want to have to add an attribute each time one of those foreign types was used in a function.

jplatte:
You wouldn't have to use different tags for different arguments, I think. `crate::FfiConverterTag` will work for both:

- First-party `impl<T> FfiConverter<T> for TheThing`
- `impl FfiConverter<FfiConverterTag> for TheThing` blocks in the current crate

If this works (and I think it does), it basically abuses coherence to declare new first-party implementations as breaking changes, because adding a "blanket" implementation is breaking as far as coherence is concerned.

jplatte:
Yes, so I'm thinking that you have to opt into it at the `uniffi::export` level. I.e. not per argument, but per impl block / function that you want to export. It shouldn't be too bad since you only have to write `#[uniffi::export(tag = crate::FfiConverterTag)]` instead of `#[uniffi::export]`, and only on those fns / impls that actually use types that don't have first-party `FfiConverter` impls.

Regarding declaration, I think UDL could declare this when you use it (currently that's the only safe-ish way to write `FfiConverter` impls for third-party types anyways, right?), and for post-UDL we could just require people to declare the type "by hand" (like the `uniffi_types` module right now) if they want to pass third-party types over FFI.

bendk:
I think that could work okay, but I don't see why that's better than always using `FfiConverterTag`.  The main tradeoff I see is the macro ergonomics:
  - If you always you `FfiConverterTag` then we would always need to define it, so you need to invoke a uniffi macro in tho crate root.
  - If you conditionally use `FfiConverterTag` , then if you don't use remote types you can skip the macro.  But if you do use remote types, then you do need the macro and also add `(tag = crate::FfiConverterTag)` to some number of `uniffi::export` attributes.  It might be a small number, but it also could be quite big if you have a lot of standalone functions.

Based on those criteria, I definitely prefer the first option.  Is there a difference beyond the ergonomics?

jplatte:
Yes, making it explicit also makes it more flexible, but I'm not sure how important that is. I don't see why you would pull `FfiConverter` impls from another crate that doesn't define the type the impl is on. I don't hate the idea of always using `crate::FfiConverterTag`.

bendk:
Sounds like you're okay with always defining a tag type in the crate root, so I went ahead and updated this PR to do that.  I'd love review on this updated PR.

jplatte:
Will review next week

docs/manual/src/internals/lifting_and_lowering.md
```diff
@@ -94,3 +94,22 @@ Note that length fields in this format are serialized as *signed* integers
 despite the fact that they will always be non-negative. This is to help
 ease compatibility with JVM-based languages since the JVM uses signed 32-bit
 integers for its size fields internally.
+
+## Code Generation and the FfiConverter trait
+
+UniFFI needs to generate Rust code to lift/lower types.  To help with this, we define the `FfiConverter` trait which contains the code to lift/lower/serialize a particular type.
+
+The most straightforward approach would be to define `FfiConverter` on the type being lifted/lowered/serialized.  However, this wouldn't work for remote types defined in 3rd-part crates because of the Rust orphan rules.  For example, our crates can't implement `FfiConverter` on `serde_json::Value` because both the trait and the type are remote.
```
mhammond, line 102: nano-nit typo - "3rd-part" needs a "y"

docs/manual/src/internals/lifting_and_lowering.md
```diff
@@ -94,3 +94,22 @@ Note that length fields in this format are serialized as *signed* integers
...
+
+To work around this we do several things:
+
+ - `FFIConverter` gets a generic type parameter.  This type is basically arbitrary and doesn't affect the lowering/lifting/serialization process.
```
mhammond, line 106: here and below, should be `FfiConverter` - we *used* to have `FFIConverter` but renamed it, right? If so, do we actually want `UniffiTag`?

docs/manual/src/internals/lifting_and_lowering.md
```diff
@@ -94,3 +94,22 @@ Note that length fields in this format are serialized as *signed* integers
...
+ - We generate a unit struct named `UniFFITag` in the root of each UniFFIed crate.
+ - Each crate uses the `FFIConverter<crate::UniFFITag>` trait to lower/lift/serialize values for its scaffolding functions.
+
+This allows us to work around the orphan rules when defining `FFIConverter` impls.
+ - UniFFI consumer crates can implement lifting/lowering/serializing types for their own scaffolding functions, for example `impl FFIConverter<crate::UniFFITag> for serde_json::Value`.  This is allowed since `UniFFITag` is a local type.
+ - The `uniffi` crate can implement lifting/lowering/serializing types for all scaffolding functions using a generic impl, for example `impl<UT> FFIConverter<UT> for u8`.  "UT" is short for "UniFFI Tag"
+ - We don't currently use this, but crates can also implement lifting/lowering/serializing their local types for all scaffolding functions using a similar generic impl (`impl<UT> FFIConverter<UT> for MyLocalType`).
+
+For more details on the specifics of the "orphan rule" and why these are legal impls, see the [Rust Chalk Book](https://rust-lang.github.io/chalk/book/clauses/coherence.html#the-orphan-rules-in-rustc)
```
mhammond, line 115: nit - spell "implementations" out

mhammond (approved):
This makes sense to me - thanks for the docs! Not sure about the case of the names though, but I'll leave that up to you.

uniffi_macros/src/enum_.rs
```diff
@@ -50,7 +50,7 @@ pub(crate) fn enum_ffi_converter_impl(
                 let write_fields = v.fields.iter().map(write_field);
 
                 quote! {
-                    Self::#v_ident { #(#fields),* } => {
+                    #ident::#v_ident { #(#fields),* } => {
```
jplatte, line 53: Why change this?

uniffi_macros/src/enum_.rs
```diff
@@ -89,14 +89,14 @@ pub(crate) fn enum_ffi_converter_impl(
 
     quote! {
         #[automatically_derived]
-        impl ::uniffi::RustBufferFfiConverter for #ident {
-            type RustType = Self;
+        unsafe impl ::uniffi::FfiConverter<crate::UniFFITag> for #ident {
```
jplatte, line 92: The derive always has its code expanded in the same scope as the type defonition, so it can and should be a blanket impl I think:


```suggestion
        unsafe impl<T> ::uniffi::FfiConverter<T> for #ident {
```

jplatte:
Also, can we please name it `UniFfiTag` or `UniffiTag`? Multiple consecutive uppercase letters in type names are practically not a thing bevause the Rust API guidelines say you should lowercase all but the first letter of all-uppercase acronyms in type names, so I find `UniFFITag` looks really out of place.

uniffi_macros/src/enum_.rs
bendk: No good reason, I'll change it back.

bendk:
Pushed some changes, mostly to rename things and fix typos.  I agree that we should avoid the triple-caps in `FFI`.

uniffi_macros/src/error.rs
```diff
@@ -60,10 +60,6 @@ pub fn expand_error(input: DeriveInput, module_path: Vec<String>) -> TokenStream
 
     quote! {
         #ffi_converter_impl
-
-        #[automatically_derived]
-        impl ::uniffi::FfiError for #ident {}
```
jplatte, line 65: I didn't see the trait being removed, why is it no longer being implemented?

jplatte:
One thing I think I should write down now that the traits are being refactored: I've had the idea that we should maybe split `FfiConverter` into two, one for transferring types Rust > ForeignLangs and one for transferring types in the opposite direction. This would make some currently-impossible things possible, like returning an external `#[non_exhaustive]` struct from a uniffied function, or having a uniffied function accept an external `#[non_exhaustive]` enum as a parameter.

bendk:
> One thing I think I should write down now that the traits are being refactored: I've had the idea that we should maybe split `FfiConverter` into two, one for transferring types Rust > ForeignLangs and one for transferring types in the opposite direction. This would make some currently-impossible things possible, like returning an external `#[non_exhaustive]` struct from a uniffied function, or having a uniffied function accept an external `#[non_exhaustive]` enum as a parameter.

I like this idea conceptually, but I'd rather keep a single trait and feel better about using `unimplemented!()` for methods you can't implement -- or don't want to.

The main reason I'm leaning this way is that if there are any dependencies between the different traits, then it leads to writing blanket impls that often run into Rusts conflicting implementation rules.  I hope to have a PR soon that's an example of this.

This isn't a strong opinion, in fact I think it could be nice to split up the traits.  It's just something I've personally given up on for now.

bendk:
Gonna merge this one now that the 0.23 release is done.

PR #1445, opened 2023-01-13
CHANGELOG/documentation updates for 0.23

bendk:
- Added CHANGELOG sections to describe the new crate structure and breaking changes from PR #1374
- Updated the manual sections describing how to set up `uniffi-bindgen` binaries.  In particular, we're now suggesting naming the binaries `uniffi-bindgen` instead of `run-uniffi-bindgen`.  The "run" part seemed unnecessary to me, especially since `uniffi-bindgen` no longer provides a standalone binary.

My plan is to make a new UniFFI release once this is approved and merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,16 @@
+### Migrating to UniFFI 0.23
+
+- Update your `Cargo.toml` file to only depend on the `uniffi` crate.  Follow the directions from the [Prerequisites section of the manual](https://mozilla.github.io/uniffi-rs/tutorial/Prerequisites.html)
+- Create a `uniffi-bindgen` binary for your project.  Follow the directions from the [Foreign language bindings section of the manual](https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html).
+- Uninstall the system-wide `uniffi_bindgen`: `cargo uninstall uniffi_bindgen`.  (Not strictly necessary, but you won't be using it anymore).
+
 <!-- The sections in this file are managed automatically by `cargo release` -->
 <!-- See our [internal release process docs](docs/release-process.md) and for more general -->
 <!-- guidance, see https://github.com/sunng87/cargo-release/blob/master/docs/faq.md#maintaining-changelog -->
 
 <!-- next-header -->
 
-## [[UnreleasedVersion]] - (_[[ReleaseDate]]_)
+## [[UnreleasedUniFFIVersion]] (backend crates: [[UnreleasedBackendVersion]] - (_[[ReleaseDate]]_)
 
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.22.0...HEAD).
 
@@ -19,12 +25,6 @@
   - Projects that use UniFFI for binding/scaffolding generation now only need to depend on the `uniffi` crate and no longer need to depend on `uniffi_bindgen`, `uniffi_build`, etc.
   - The version numbers for each crate will no longer by kept in sync after this release.  In particular `uniffi` will have breaking changes less often than `uniffi_bindgen` and other crates.  This means that UniFFI consumers that specify their versions like `uniffi = "0.23"` will not need to bump their `uniffi` version as often as before.
 - Callback interface method calls are no longer logged (#1439)
-
-### Migrating to UniFFI 0.23
-
-- Update your `Cargo.toml` file to only depend on the `uniffi` crate.  Follow the directions from the [Prerequisites section of the manual](https://mozilla.github.io/uniffi-rs/tutorial/Prerequisites.html)
-- Create a `uniffi-bindgen` binary for your project.  Follow the directions from the [Foreign language bindings section of the manual](https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html).
-- Uninstall the system-wide `uniffi_bindgen`: `cargo uninstall uniffi_bindgen`.  (Not strictly necessary, but you won't be using it anymore).
 
 ## v0.22.0 - (_2022-12-16_)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,21 @@
 
 ### ⚠️ Breaking Changes ⚠️
 
-- `uniffi_bindgen`: Removed the `run_main` function.  It's moved to `uniffi::uniffi_bindgen_main` and now unconditionally succeeds rather than return a `Result<()>`
+- `uniffi_bindgen` no longer provides a standalone binary.  Having a standalone binary resulted in version mismatches when the `uniffi` version specified in `Cargo.toml` didn't match the `uniffi_bindgen` version installed on the system.  Read [The foreign language bindings](https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html) section of the manual for how to set up a `uniffi-bindgen` binary that's local to your workspace.
+- `uniffi_bindgen`: Removed the `run_main` function.  It's moved to `uniffi::uniffi_bindgen_main` and now unconditionally succeeds rather than return a `Result<()>`.
 
 ### What's changed
 
+- The UniFFI crate organization has been significantly reworked:
+  - Projects that use UniFFI for binding/scaffolding generation now only need to depend on the `uniffi` crate and no longer need to depend on `uniffi_bindgen`, `uniffi_build`, etc.
+  - The version numbers for each crate will no longer by kept in sync after this release.  In particular `uniffi` will have breaking changes less often than `uniffi_bindgen` and other crates.  This means that UniFFI consumers that specify their versions like `uniffi = "0.23"` will not need to bump their `uniffi` version as often as before.
 - Callback interface method calls are no longer logged (#1439)
+
+### Migrating to UniFFI 0.23
+
+- Update your `Cargo.toml` file to only depend on the `uniffi` crate.  Follow the directions from the [Prerequisites section of the manual](https://mozilla.github.io/uniffi-rs/tutorial/Prerequisites.html)
+- Create a `uniffi-bindgen` binary for your project.  Follow the directions from the [Foreign language bindings section of the manual](https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html).
+- Uninstall the system-wide `uniffi_bindgen`: `cargo uninstall uniffi_bindgen`.  (Not strictly necessary, but you won't be using it anymore).
 
 ## v0.22.0 - (_2022-12-16_)
 
--- a/docs/manual/src/tutorial/foreign_language_bindings.md
+++ b/docs/manual/src/tutorial/foreign_language_bindings.md
@@ -10,33 +10,32 @@ Ideally you would then run the `uniffi-bindgen` binary from the `uniffi` crate t
 is only available with [Cargo nightly](https://doc.rust-lang.org/cargo/reference/unstable.html#artifact-dependencies).
 To work around this, you need to create a binary in your project that does the same thing.
 
-
 Add the following to your `Cargo.toml`:
 
 ```toml
 [[bin]]
-name = "run-uniffi-bindgen"
-path = "run-uniffi-bindgen.rs"
+name = "uniffi-bindgen"
+path = "uniffi-bindgen.rs"
 ```
 
-Create `run-uniffi-bindgen.rs`:
+Create `uniffi-bindgen.rs`:
 ```rust
 fn main() {
     uniffi::uniffi_bindgen_main()
 }
 ```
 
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -10,10 +10,14 @@
 
 ### ⚠️ Breaking Changes ⚠️
 
-- `uniffi_bindgen`: Removed the `run_main` function.  It's moved to `uniffi::uniffi_bindgen_main` and now unconditionally succeeds rather than return a `Result<()>`
+- `uniffi_bindgen` no longer provides standalone binary.  Having a standalone binary resulted in version mismatches when the `uniffi` version specified in `Cargo.toml` didn't match the `uniffi_bindgen` version installed on the system.  Read [The foreign language bindings](https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html) section of the manual for how to set up a `uniffi-bindgen` binary that's local to your workspace.
```
badboy, line 13:
```suggestion
- `uniffi_bindgen` no longer provides a standalone binary.  Having a standalone binary resulted in version mismatches when the `uniffi` version specified in `Cargo.toml` didn't match the `uniffi_bindgen` version installed on the system.  Read [The foreign language bindings](https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html) section of the manual for how to set up a `uniffi-bindgen` binary that's local to your workspace.
```

CHANGELOG.md
```diff
@@ -10,10 +10,14 @@
...
+- `uniffi_bindgen`: Removed the `run_main` function.  It's moved to `uniffi::uniffi_bindgen_main` and now unconditionally succeeds rather than return a `Result<()>`.
 
 ### What's changed
 
+- The UniFFI crate organization has been significantly reworked:
+  - Projects that use UniFFI for binding/scaffolding generation now only need to depend on the `uniffi` crate and no longer need to depend on `uniffi_bindgen`, `uniffi_build`, etc.
+  - The version numbers for each crate will no longer by kept in sync after this release.  In particular `uniffi` will have breaking changes less often than `uniffi_bindgen` and other crates.  This means that UniFFI consumers that specify their versions like `uniffi = "0.23"` will not need to bump their `uniffi` version as often as before.
```
badboy, line 20: Should we update our own docs at `docs/release-process.md` on how to release individual crates?

badboy (changes requested):

CHANGELOG.md
bendk: Yeah, but I actually wasn't sure how the new release process would work now.  I was thinking we could run through a few releases and update the docs as we went.  I'm not really sure though, WDYT?

mhammond: We might as well make a best-effort attempt - I assume you will be making the release, so maybe note what you expect to do, then we can tweak as we go?  (ie, if I had to make it, I wouldn't really know where to start)

bendk: That makes sense to me, just pushed a commit with my attempt at this.

docs/release-process.md
```diff
@@ -4,17 +4,39 @@
...
+   the release.
+   * Anything that affects any UniFFI consumers should be listed, this includes consumers that
+     use UniFFI to generate their scaffolding/bindings, external bindings generators, etc.
+
+1. Take a look over `uniffi/CHANGELOG.md` and make sure the "unreleased" section lists changes for
+   the release that affect users `uniffi` of the UniFFI crate.
+   * This should be a copy of the items from the top-level `CHANGELOG.md` that affects
+     scaffolding/bindings generation.
+   * Note that breaking changes for a particular `uniffi_*` crate are not necessarily breaking for
+     the `uniffi` crate.  See `./uniffi-versioning.md` for a discussion of this.
```
bendk, line 29: Having to maintain two separate changelogs doesn't seem great, but I can't think of a better way to handle this.

AFAICT, the changelogs are not published anywhere they're only available by browsing github.  If this is true, it's nice because I don't need to update any automation.  But maybe we should consider including the changelog in the docs.

bendk:
Pushed some updates based on the feature gate changes.  I think this is basically ready, except for one thing:  I added a "how to migrate" section in the changelog, but I wonder if there's a better way to get this info out to UniFFI users since the changelog isn't very visible at this point.  I'm not sure what's a better place though?  The manual?  The README.md file?

docs/manual/src/tutorial/foreign_language_bindings.md
```diff
@@ -10,49 +10,48 @@ Ideally you would then run the `uniffi-bindgen` binary from the `uniffi` crate t
...
 To work around this, you need to create a binary in your project that does the same thing.
 
-
 Add the following to your `Cargo.toml`:
 
 ```toml
 [[bin]]
-name = "run-uniffi-bindgen"
-path = "run-uniffi-bindgen.rs"
+name = "uniffi-bindgen"
```
mhammond, line 17: The name isn't important, right? Maybe a comment saying something like "# whatever name makes sense for your project can be used, but that the rest of this tutorial assumes `uniffi-bindgen`"?

docs/manual/src/tutorial/foreign_language_bindings.md
```diff
@@ -10,49 +10,48 @@ Ideally you would then run the `uniffi-bindgen` binary from the `uniffi` crate t
...
 
 ### Multi-crate workspaces
 
 If your project consists of multiple crates in a Cargo workspace, then the process outlined above would require you
 creating a binary for each crate that uses UniFFI.  You can avoid this by creating a separate crate for running `uniffi-bindgen`:
-  - Name the crate `run-uniffi-bindgen`
+  - Name the crate `uniffi-bindgen`
   - Add this dependency to `Cargo.toml`: `uniffi = {version = "0.XX.0", features = ["cli"] }`
-  - Follow the steps from the previous seciton to add the `run-uniffi-bindgen` binary target
+  - Follow the steps from the previous seciton to add the `uniffi-bindgen` binary target
```
mhammond, line 36: typo - /seciton/

docs/release-process.md
```diff
@@ -4,17 +4,39 @@
...
+    for details).
+
+Steps:
+
+1. Take a look over `CHANGELOG.md` and make sure the "unreleased" section lists all changes for
+   the release.
+   * Anything that affects any UniFFI consumers should be listed, this includes consumers that
+     use UniFFI to generate their scaffolding/bindings, external bindings generators, etc.
+
+1. Take a look over `uniffi/CHANGELOG.md` and make sure the "unreleased" section lists changes for
```
mhammond, line 24: this sentence reads a little odd - "users `uniffi` of the UniFFI crate." - do you mean something like "users of the top-level `uniffi` crate"?

docs/release-process.md
mhammond: I don't understand why we need a new CHANGELOG, and don't understand how I'd actually use both (ie, *every* change to the sub-crates potentially impacts users of the top-level crate, as many features will be released without a top-level version bump. I assume this has something to do with communicating how breaking a breaking change is?

docs/release-process.md
```diff
@@ -4,17 +4,39 @@
...
+   the release that affect users `uniffi` of the UniFFI crate.
+   * This should be a copy of the items from the top-level `CHANGELOG.md` that affects
+     scaffolding/bindings generation.
+   * Note that breaking changes for a particular `uniffi_*` crate are not necessarily breaking for
+     the `uniffi` crate.  See `./uniffi-versioning.md` for a discussion of this.
+
+1. Decide on a new version number for `uniffi` crate.  Since we are pre-`1.0`, if there are breaking
+   changes then this should be a minor version bump, otherwise a patch version bump.
+
+1. Decide on a new version number for the other `uniffi_*` crates.
```
mhammond, line 34: Ah, for some reason I was assuming we'd only bump version numbers for sub-crates which changed, so those numbers would start diverging - but given the above, that's not correct, right? I wonder if that's worth spelling out somewhere?

docs/release-process.md
```diff
@@ -26,14 +48,47 @@ It's not (yet) quite an ideal fit for our workflow, but it helps! Steps:
...
+     worthwhile anyway!.
+
+1. Perform a real run to pump the version numbers, *publish the release to crates.io* and *create
+   the git tag*.
+   * Take care because this will publish the new releases on crates.io.
+   * This will create a local git tag, but does not push it to github.
+   * `cargo release-backend-crates -x {MAJOR}.{MINOR}.{PATCH}` to release the `uniffi_*` crates
+   * `cargo release-uniffi -x {MAJOR}.{MINOR}.{PATCH}` to release the `uniffi` crates
+
+1. Push your branch: `git push origin --tags`
```
mhammond, line 67: Is there a step missing here to update `uniffi/Cargo.toml` with the new version numbers of the sub-crates - there's also a bit of a chicken-and-egg situation here - don't the sub-crates need to be on crates.io before we can release the top-level crate?

mhammond (commented):
Thanks for the update! It did raise some more questions for me (which is good I suppose :)

docs/release-process.md
bendk: Some changes will only affect the sub crates.  Here's some examples from the current CHANGELOG:
 -  The `guess_crate_root` function is now public
- `uniffi_bindgen`: Renamed `FFIArgument`, `FFIFunction` and `FFIType` to `FfiArgument`, `FfiFunction` and `FfiType` (this was a breaking change).

The issue you mentioned was also on my mind: a change could be breaking for sub-crate users, but non-breaking for top-level users.  However, when I looked through the changelog, I didn't see any examples of those.

The last issue was version numbers.  The version numbers for `uniffi` and the sub-crates are going to start to diverge, which makes the changelog tricky.  I don't know how we would setup the `cargo release` template to move changes from the unreleased section, if there are 2 different version numbers involved.

docs/release-process.md
bendk: I think that `cargo release-backend-crates` should do that, although I'm not sure.  I'll verify that when I do the 0.23 release.  I added a warning to that section.

CHANGELOG.md
```diff
@@ -10,12 +10,22 @@
...
 - Callback interface method calls are no longer logged (#1439)
 
+### Migrating to UniFFI 0.23
```
mhammond, line 23: I was wondering whether we could move this section to the top of the file above `<!-- next-header -->` to make it somewhat "sticky", changing the title to something like "Migrating from UniFFI 0.22"? It would probably need an extra introductory para to say something like "UniFFI 0.23 had significant changes to the versioning scheme, here's the info you need to upgrade" too.

My vague concern is that we start leveraging the new versioning scheme to make new releases far more often just to get new features someone wants to leverage. So a UniFFI user might not actually attempt to upgrade until we've made 5 new releases (with `uniffi` still on 0.23.x but the sub-crates up to 0.27.0), so this info might get lost in the noise.

bendk:
Pushed some changes after talking CHANGELOGs with @mhammond:
  - Moved the `Upgrading to 0.23` section to the top of the changelog.  I agree it's good to keep that visible for a while
  - Removed `uniffi/CHANGELOG.md`.  Mark convinced me that it's better to just keep one changelog and label what affects "regular" users vs what affects external bindings authors.  How do we do that?  I'm not totally sure, I think we should just wing it and discuss the larger issue of changelogs in #1458
  - Updated the `release.toml` scripts to hopefully update that changelog correctly when we do the release.  I'll definitely monitor this when it's release time though.

WDYT, can we merge this one and start the `0.23` release?

mhammond (approved):
Thanks for working through this Ben! Probably worth waiting for @badboy's approval too though.

badboy:
> Thanks for working through this Ben! Probably worth waiting for @badboy's approval too though.

on it today

badboy (approved):
minor nits, but otherwise good to go.

CHANGELOG.md
```diff
@@ -1,19 +1,29 @@
...
+
+- Update your `Cargo.toml` file to only depend on the `uniffi` crate.  Follow the directions from the [Prerequisites section of the manual](https://mozilla.github.io/uniffi-rs/tutorial/Prerequisites.html)
+- Create a `uniffi-bindgen` binary for your project.  Follow the directions from the [Foreign language bindings section of the manual](https://mozilla.github.io/uniffi-rs/tutorial/foreign_language_bindings.html).
+- Uninstall the system-wide `uniffi_bindgen`: `cargo uninstall uniffi_bindgen`.  (Not strictly necessary, but you won't be using it anymore).
+
```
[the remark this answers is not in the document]
bendk: @mhammond suggested making it sticky for a while, in case a user misses several releases and then tries to upgrade.  This makes sense for about 6-12 months, then we can move it back down.  That said, I don't have strong feelings about this one.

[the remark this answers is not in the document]
badboy: Ah, didn't see that. Yeah, I'm fine with that.

PR #1447, opened 2023-01-17
Undo workaround of '-j 2' in tests for Rust 1.61

mhammond:


mhammond:
Depends on #1446 but I can't see why we can't get review now.

mhammond:
(only saves ~90s on the main test run, but still probably worthwhile)

badboy (approved):

PR #1451, opened 2023-01-18
Move most `uniffi` dependencies behind feature gates

bendk:
I was thinking about making a UniFFI release with the changes from #1374, but one issue was making me hesitate and I wanted to address it.

One disadvantage of making `uniffi` the frontend crate is that any crate that depended on it would pull in all its dependencies, which could lead to binary bloat.

This change makes many dependencies optional and defines feature gates to enable them.

Updated the tutorial section based on the changes.

### DIFF
--- a/docs/manual/src/tutorial/Rust_scaffolding.md
+++ b/docs/manual/src/tutorial/Rust_scaffolding.md
@@ -4,14 +4,14 @@
 
 Now we generate some Rust helper code to make the `add` method available to foreign-language bindings.  
 
-First, add `uniffi` to your crate as both a dependency and build-dependency.  This adds the runtime support code that powers UniFFI and build-time support for generating the Rust scaffolding code.
+First, add `uniffi` to your crate as both a dependency and build-dependency.  Enable the `build` feature for the build-dependencies.  This adds the runtime support code that powers UniFFI and build-time support for generating the Rust scaffolding code.
 
 ```toml
 [dependencies]
-uniffi = "1.0.0"
+uniffi = "0.XX.0"
 
 [build-dependencies]
-uniffi = "1.0.0"
+uniffi = { version = "0.XX.0", features = ["build"] }
 ```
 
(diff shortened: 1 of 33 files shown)


uniffi/src/lib.rs
```diff
@@ -2,19 +2,25 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-#[cfg(feature = "clap")]
-mod cli;
 /// Reexport items from other uniffi creates
+pub use uniffi_core::*;
+pub use uniffi_macros::{export, include_scaffolding, Enum, Error, Object, Record};
```
bendk, line 7: This could also be behind a feature gate, but I skipped that since it's a proc-macro crate.  I'm not really sure how this works, but I believe proc-macro crates don't contribute to the final binary size.

docs/manual/src/tutorial/Rust_scaffolding.md
```diff
@@ -4,14 +4,14 @@
 
 Now we generate some Rust helper code to make the `add` method available to foreign-language bindings.  
 
-First, add `uniffi` to your crate as both a dependency and build-dependency.  This adds the runtime support code that powers UniFFI and build-time support for generating the Rust scaffolding code.
+First, add `uniffi` to your crate as both a dependency and build-dependency.  Enable the `build` feature for the build-depndencies.  This adds the runtime support code that powers UniFFI and build-time support for generating the Rust scaffolding code.
```
mhammond, line 7: typo: depndencies

mhammond (approved):
This makes sense to me!

badboy (approved):
Also tested this against Glean.
With the `build` feature enabled on the `build-dependency` everything looks good.
No clap in our build \o/ (fwiw: with `default-features = false` on current uniffi main I get the same).

PR #1454, opened 2023-01-19
Address nits from #1442

mhammond:


skhamis (approved):
Thanks for the follow-up. LGTM!

PR #1455, opened 2023-01-20
Add a small note to the manual about Kotlin lifetimes

mhammond:
As noted in https://github.com/mozilla/application-services/pull/5326, it probably makes sense for us to document how our Kotlin wrappers use `AutoClosable`. This is very short and sweet, but that seems OK to me.

### DIFF
--- a/docs/manual/src/kotlin/lifetimes.md
+++ b/docs/manual/src/kotlin/lifetimes.md
@@ -1,0 +1,27 @@
+# Kotlin Lifetimes
+
+All interfaces exposed via Kotlin expose a public API for freeing
+the Kotlin wrapper object in lieu of reliable finalizers. This is done
+by making the "base class" for all such generated objects implement the
+`Disposable` and `AutoCloseable` interfaces.
+
+As such, these wrappers all implement a `close()` method, which must be
+explicitly called to ensure the associated Rust resources are reclaimed.
+
+The best way to arrange for this to be called at the right time is beyond
+the scope of this document; you should consult the official documentation for
+`AutoClosable`, but one common pattern is the Kotlin
+[use function](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/use.html).
+
+## Nested objects
+
+We also need to consider what happens when objects are contained in other objects.
+The current situation is:
+
+* Dictionaries that contain interfaces implement `AutoClosable` with their close() method closing
+  any contained interfaces.
+
+* Enums can't currently contain interfaces.
+
+* Lists/Maps don't implement `AutoClosable`; if you have a list/map of interfaces
+  you need to close each one individually.
--- a/docs/manual/src/SUMMARY.md
+++ b/docs/manual/src/SUMMARY.md
@@ -24,6 +24,7 @@
 # Kotlin
 
 - [Integrating with Gradle](./kotlin/gradle.md)
+- [Kotlin Lifetimes](./kotlin/lifetimes.md)
 
 # Swift
 


docs/manual/src/kotlin/lifetimes.md
```diff
@@ -0,0 +1,14 @@
...
+by making the "base class" for all such generated objects implement the
+`Disposable` and `AutoCloseable` interfaces.
+
+As such, these wrappers all implement a `close()` method, which must be
+explicitly called to ensure the associated Rust resources are reclaimed.
+
+The best way to arrange for this to be called at the right time is beyond
+the scope of this document; you should consult the official documentation for
+`AutoClosable`, but one common pattern is the Kotlin
+[use](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/use.html) function.
```
badboy, line 14: I would probably recommend:

```suggestion
[`use` function](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/use.html).
```

just the small word `use` is a bad click target.

badboy (approved):

docs/manual/src/kotlin/lifetimes.md
```diff
@@ -0,0 +1,14 @@
+# Kotlin Lifetimes
+
+All interfaces exposed via Kotlin expose a public API for freeing
+the Kotlin wrapper object in lieu of reliable finalizers. This is done
+by making the "base class" for all such generated objects implement the
+`Disposable` and `AutoCloseable` interfaces.
+
+As such, these wrappers all implement a `close()` method, which must be
+explicitly called to ensure the associated Rust resources are reclaimed.
```
bendk, line 9: I'd like a paragraph describing what happens when interfaces are contained in other objects.  I think that's:
 - Dictionaries that contain interfaces implement `Autoclosable` and their `close()` method closes the contained interface.
 - Enums can't currently contain interfaces
 - Lists/Maps don't implement `Autoclosable`, if you have a list of interfaces you need to close each one individually.

bendk (approved):
Looks good to me. I think short and sweet is totally fine, but I did have 1 suggested addition.

docs/manual/src/kotlin/lifetimes.md
mhammond: > Lists/Maps don't implement Autoclosable, if you have a list of interfaces you need to close each one individually.

Ouch. A quick google shows that's difficult to solve :(

But thanks, added both suggestions.

jplatte: This worries me, I really hope we're not affected but I'm not very optimistic.

I guess this problem also applies to lists or maps that contain a dictionary with a field of an interface type?

mhammond: Yeah :( What do you suggest?

jplatte: I have no suggestions but I've brought it up for discussion in our team.

jplatte: Just to be sure I understand right: In general interface types have non-deterministic cleanup because that's just how GC's work. But for the case of interfaces in lists / maps, no automatic GC is happening *at all*. And that only affects Kotlin, non of the other backends. Is that right?

bendk: In general, interfaces on Kotlin don't rely on the GC but require you to ensure they are destroyed/closed -- usually this means calling `destroy()` or by using the `use` function.  This makes the cleanup deterministic.

For lists/maps of interfaces, you need to call `destroy()` or `use()` on all the individual items.

For records that contain interfaces, you don't need to `destroy()`/`use()` each field, you can call `destroy()` / `use()` on the record itself.

For Python/Swift, we destroy the interface in the destructor if it's still live.  If you want deterministic destruction, you should follow the above advice.  But if you don't follow it, then your interface will still get freed at some point.

mhammond: > But for the case of interfaces in lists / maps, no automatic GC is happening _at all_. And that only affects Kotlin, non of the other backends. Is that right?

This might be obvious and exactly what you meant, but my understanding is that the GC is happening for these objects, it's just that the `close()` method on our object will not be called. All items in the enum, *including* the generated wrapper are cleaned up, but `close()` isn't called for our interfaces, so the rust `Arc<>` leaks forever.

The [comments here](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/kotlin/templates/ObjectRuntime.kt#L102-L110) imply there might be improvements a Kotlin expert could make - eg, maybe we could also generate a `finalize()` method  - there are a number of restrictions around this which someone would carefully need to look at (eg, that file also says "it's apparently also possible for the JVM to finalize an object while there is an in-flight call to one of its methods" - but I don't think that would bother our Rust scaffolding?)

IOW, it seems like there's still scope here to be able to soften these notes to talk just about how *deterministic* cleanup is, rather than how the current scenario can involve actual leaks if the consumer is not careful IIUC.

jplatte: Yes, I meant the Rust object being leaked / the refcount not being decremented correctly. Thanks for the elaboration!

Issue #1435, opened 2022-12-22
Objects can't be external types


bendk:
Currently, objects can't be supported by external types because we generate [this scaffolding code](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/scaffolding/templates/ExternalTypesTemplate.rs#L6)

This works for enums, errors, and records, where we generate an `FfiConverterType[name]` struct for them.  But it doesn't work for objects, where we use `Arc<T>` as the `FfiConverter`, which is implemented in the `uniffi` crate itself.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-223)
┆Issue Number: UNIFFI-223

stan-irl:
@bendk do you have any tips for how to work around this limitation?

I think this blocks users of this framework from splitting up a workspace into separate crates (like in application-services) for a lot of use cases - my specific usecase is that I want to define an auth delegate to pass from the platform into a number of different rust crates:
```
callback interface AuthDelegate {
  string get_token();
};

```

Right now my backup plan is to have a single crate with many modules, and concatenate all the udl files into a single file in the root. 

Keen to hear what you think

bendk:
I can't think of any good workarounds for now, but once we figure out #1444, I think this might get fixed automatically.  I'm hoping that happens in the next couple weeks

stan-irl:
that would be awesome 🤞

stan-irl:
ok i have a workaround:
I'm building my consumer UDL with the following script
```
use std::error::Error;
use std::fs;
use std::{fs::File, io};

const TMPFILE: &str = "./consumer.udl";

fn main() -> Result<(), Box<dyn Error>> {
    {
        let mut output = File::create(TMPFILE)?;
        let inputs = vec!["./consumer.tmpl.udl", "../auth/auth.udl", "../log/log.udl"];
        for i in inputs {
            let mut input = File::open(i)?;
            io::copy(&mut input, &mut output)?;
        }

        uniffi_build::generate_scaffolding(TMPFILE).unwrap();
    }
    fs::remove_file(TMPFILE)?;
    return Ok(());
}
```

Im concatenating the dependency UDLs (which have the namespace omitted) into the consumer UDL before generating the scaffolding. The I can just re-export the types from the consumer crate.

[...]

PR #1456, opened 2023-01-20
Make external types work with interfaces (#1435)

bendk:
I though #1444 might fix this, since it improves some of the `FfiConverter` name handling, but there was a bigger issue:  interfaces have fundamental differences compared to enums/records when it comes to moving them across the FFI since. Interfaces use `Arc<>` on the Rust side and `void *` on the FFI side.

Added way to flag that external types are interfaces, updated the code to handle these correctly.

This feels a little hacky to me, but I think this won't be needed once we move to proc-macros:
  -   The proc-macros see the type names used in the function, so they can always figure out the correct `FfiConverter`
  - Using the `FfiConverter`, they can always know the correct argument and return types.  ([example](https://github.com/mozilla/uniffi-rs/blob/670eb225ab1742af44ab93c3e378ecd2c642432c/uniffi_macros/src/export/scaffolding.rs#L133))

### DIFF
--- a/docs/manual/src/udl/ext_types_external.md
+++ b/docs/manual/src/udl/ext_types_external.md
@@ -50,6 +50,15 @@ Your `Cargo.toml` must reference the external crate as normal.
 
 The `External` attribute can be specified on dictionaries, enums and errors.
 
+## External interface types
+
+If the external type is an [Interface](./interfaces.md), then use the `[ExternalInterface]` attribute instead of `[External]`:
+
+```idl
+[ExternalInterface="demo-crate"]
+typedef extern DemoInterface;
+```
+
 ## Foreign bindings
 
 The foreign bindings will also need to know how to access the external type,
--- a/uniffi_bindgen/src/interface/attributes.rs
+++ b/uniffi_bindgen/src/interface/attributes.rs
@@ -14,6 +14,7 @@
 //! all handled by a single abstraction. This might need to be refactored in future
 //! if we grow significantly more complicated attribute handling.
 
+use crate::interface::types::ExternalKind;
 use anyhow::{bail, Result};
 use uniffi_meta::Checksum;
 
@@ -32,7 +33,10 @@ pub(super) enum Attribute {
     Threadsafe, // N.B. the `[Threadsafe]` attribute is deprecated and will be removed
     Throws(String),
     // `[External="crate_name"]` - We can `use crate_name::...` for the type.
-    External(String),
+    External {
+        crate_name: String,
+        kind: ExternalKind,
+    },
     // Custom type on the scaffolding side
     Custom,
 }
@@ -69,7 +73,14 @@ impl TryFrom<&weedle::attribute::ExtendedAttribute<'_>> for Attribute {
                     "Name" => Ok(Attribute::Name(name_from_id_or_string(&identity.rhs))),
                     "Throws" => Ok(Attribute::Throws(name_from_id_or_string(&identity.rhs))),
                     "Self" => Ok(Attribute::SelfType(SelfType::try_from(&identity.rhs)?)),
-                    "External" => Ok(Attribute::External(name_from_id_or_string(&identity.rhs))),
+                    "External" => Ok(Attribute::External {
+                        crate_name: name_from_id_or_string(&identity.rhs),
+                        kind: ExternalKind::DataClass,
+                    }),
+                    "ExternalInterface" => Ok(Attribute::External {
+                        crate_name: name_from_id_or_string(&identity.rhs),
+                        kind: ExternalKind::Interface,
+                    }),
                     _ => anyhow::bail!(
                         "Attribute identity Identifier not supported: {:?}",
                         identity.lhs_identifier.0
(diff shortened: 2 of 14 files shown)


stan-irl:
thankyou so much!!!!

tarikeshaq (commented):
Dropped a few comments as I was going through the PR. I'll forward the review to @mhammond though since he's been a bit more involved in external types 😄

Overall seems sensible to me!

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -65,12 +65,21 @@ pub enum Type {
     Sequence(Box<Type>),
     Map(Box<Type>, Box<Type>),
     // An FfiConverter we `use` from an external crate
-    External { name: String, crate_name: String },
+    External {
+        name: String,
+        crate_name: String,
+        is_interface: bool,
```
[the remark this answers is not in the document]
mhammond: heh - I just put up a PR which uses an enum instead of a boolean in an almost identical context :)

uniffi_bindgen/src/interface/attributes.rs
```diff
@@ -27,6 +27,7 @@ pub(super) enum Attribute {
     ByRef,
     Enum,
     Error,
+    Interface,
```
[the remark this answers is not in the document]
mhammond: I don't feel too strongly about this, but another option might be to have the UDL instead do, say ```[ExternalInterface="uniffi-one"]``` - this would mean `External` and `ExternalInterface` end up being mutually exclusive but that should be an easy check to make

This would also tie in with comments below re `bool` vs a new enum - eg, you could have
```
enum ExternalImpl {
  Interface, // better names though? Best to avoid 'Object' though :)
  Other, // "TheRest"? :)
}
```

And `Attribute::External` could just get one of these (along with the existing String). When parsing the attributes, both the literals `"External"` and `"ExternalInterface"` create an `Attribute::External`, just with a different discriminator.

You could then also use this new enum in `Type::External`

That would avoid a new attribute completely and avoid the new `is_interface()` function (but you probably would need something like a new `get_external_imp(&self) -> Option<ExternalImpl>` or something for finder.rs to know how to create the `Type::External`.

If we thought this would ever be needed for other types, another possibility might be, say, ```[External="uniffi-one", ExternalType=interface]``` - you'd still need the new Attribute, but it could probably avoid leaking much from that module (ie, the "public" API would still be to just get the enum without knowing exactly what attributes were used to specify it - but that seems a little like yagni.

mhammond (commented):
This looks great and I'm embarrassed my most recent ext-types PR didn't hit this! All looks great but a little bikeshedding the attributes seems worthwhile.

mhammond:
Oh, and whatever we land on re attributes, the manual will need updating.

uniffi_bindgen/src/interface/types/mod.rs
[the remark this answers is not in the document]
bendk: I like the enum strategy.  I updated the code to use that.

mhammond (approved):
This looks good, thanks!

mhammond (commented):
still needs docs though :)

docs/manual/src/udl/ext_types_external.md
```diff
@@ -29,9 +29,6 @@ dictionary ConsumingDict {
 
 ```
 
-(Note the special type `extern` used on the `typedef`. It is not currently enforced that the
-literal `extern` is used, but we hope to enforce this later, so please use it!)
-
```
bendk, line 34: It seems like this wasn't referring to anything anymore, but please correct me if I missed the point.

uniffi_bindgen/src/scaffolding/mod.rs
```diff
@@ -79,12 +79,23 @@ mod filters {
...
-            "<{} as uniffi::FfiConverter<crate::UniFfiTag>>",
-            type_rs(type_)?
-        ))
+        Ok(match type_ {
+            Type::External {
+                name,
+                kind: ExternalKind::Interface,
+                ..
+            } => {
+                format!("<::std::sync::Arc<r#{name}> as uniffi::FfiConverter<crate::UniFfiTag>>")
```
bendk, line 92: This is a slight step backwards from the changes I made in #1444 to make this more uniform.  However, I don't think it will get in the way of the goal of simplifying things for the proc-macro code.  In any case, proc-macros don't support external types yet and we can figure this out later.

bendk:
Added docs and rebased after the #1444 changes.

@mhammond can you double check that the text I removed in the docs wasn't useful.

docs/manual/src/udl/ext_types_external.md
mhammond: I think it is still relevant - you could change the example in the manual to `typedef foobar DemoDict;` and it would still work - that note is to suggest that we might in the future change things so it would not work without the explicit `extern` literal there.

(IIRC, the intent was that there might be a future where we that literal might mean something in some future, so we'd reserver `extern` for this behavior, but that seems unlikely)

bendk: That makes sense, adding this back in.  I'm not sure where my brain was, I search and search and didn't see the `typedef` there.

PR #1457, opened 2023-01-23
Allow Rust traits to be exposed as an interface.

mhammond:
This PR allows Rust traits to be exposed as interfaces, including when they have associated types. It supersedes #1308 (the memory leaks discussed there were easy to resolve once the other shoe dropped for me)

Instead of repeating what I hope are comprehensive docs and tests, please see the [docs for this](https://github.com/mhammond/uniffi-rs/blob/traits-types/docs/manual/src/udl/interfaces.md#exposing-traits-as-interfaces), see the [new example](https://github.com/mhammond/uniffi-rs/tree/traits-types/examples/traits) (and the coveralls fixture exercises them quite thoroughly).

It has required zero changes to the generated bindings and it all fell out quite cleanly IMO. I've also got a POC in application-services for my specific use-case there, and it works perfectly!

### DIFF
--- a/uniffi_bindgen/src/interface/object.rs
+++ b/uniffi_bindgen/src/interface/object.rs
@@ -66,12 +66,12 @@ use uniffi_meta::Checksum;
 use super::attributes::{Attribute, ConstructorAttributes, InterfaceAttributes, MethodAttributes};
 use super::ffi::{FfiArgument, FfiFunction, FfiType};
 use super::function::{Argument, Callable};
-use super::types::{Type, TypeIterator};
+use super::types::{ObjectImpl, Type, TypeIterator};
 use super::{convert_type, APIConverter, ComponentInterface};
 
-/// An "object" is an opaque type that can be instantiated and passed around by reference,
+/// An "object" is an opaque type that is passed around by reference, can
 /// have methods called on it, and so on - basically your classic Object Oriented Programming
-/// type of deal, except without elaborate inheritance hierarchies.
+/// type of deal, except without elaborate inheritance hierarchies. Some can be instantiated.
 ///
 /// In UDL these correspond to the `interface` keyword.
 ///
@@ -86,6 +86,8 @@ use super::{convert_type, APIConverter, ComponentInterface};
 #[derive(Debug, Clone, Checksum)]
 pub struct Object {
     pub(super) name: String,
+    /// How this object is implemented in Rust
+    pub(super) imp: ObjectImpl,
     pub(super) constructors: Vec<Constructor>,
     pub(super) methods: Vec<Method>,
     // We don't include the FfiFunc in the hash calculation, because:
@@ -96,18 +98,16 @@ pub struct Object {
     //    avoids a weird circular dependency in the calculation.
     #[checksum_ignore]
     pub(super) ffi_func_free: FfiFunction,
-    #[checksum_ignore]
-    pub(super) uses_deprecated_threadsafe_attribute: bool,
 }
 
 impl Object {
-    pub(super) fn new(name: String) -> Object {
-        Object {
+    pub(super) fn new(name: String, imp: ObjectImpl) -> Self {
+        Self {
             name,
+            imp,
             constructors: Default::default(),
             methods: Default::default(),
             ffi_func_free: Default::default(),
-            uses_deprecated_threadsafe_attribute: false,
         }
     }
 
@@ -115,8 +115,22 @@ impl Object {
         &self.name
     }
 
+    /// Returns the fully qualified name that should be used by Rust code for this object.
+    /// Includes `r#`, traits get a leading `dyn`. If we ever supported associated types, then
+    /// this would also include them.
+    pub fn rust_name(&self) -> String {
+        self.imp.rust_name_for(&self.name)
+    }
+
+    pub fn imp(&self) -> &ObjectImpl {
+        &self.imp
+    }
+
     pub fn type_(&self) -> Type {
-        Type::Object(self.name.clone())
+        Type::Object {
+            name: self.name.clone(),
+            imp: self.imp,
+        }
     }
 
     pub fn constructors(&self) -> Vec<&Constructor> {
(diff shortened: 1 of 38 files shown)


uniffi_core/src/lib.rs
```diff
@@ -653,6 +653,68 @@ unsafe impl<T: Sync + Send> FfiConverter for std::sync::Arc<T> {
...
 }
 
+/// Support for passing reference-counted trait objects via the FFI.
+///
+/// This is different than the last one, since trait objects are stored as 2 pointers (AKA a "fat"
+/// pointer): one for `T` and another for the vtable. Both `Arc<dyn T>` and `Box<dyn T>` are fat,
+/// but `Arc<Box<dyn T>>` and `Box<Arc<dyn T>>` are not.
+/// Since the foreign side is expecting a single pointer, we need one of these. We use an inner
+/// `Arc<dyn T>` so we can support that directly as a receiver.
+unsafe impl<T: Sync + Send + ?Sized> FfiConverter for Box<std::sync::Arc<T>> {
```
bendk, line 663: This is going to be an issue for #1444, since that removes `RustType`  and always implements `FfiConverter<UT>` on the type itself.  And Rust doesn't `unsafe impl<T: Sync + Send + !Sized> FfiConverter for std::sync::Arc<T>`

One option would be to add back `RustType`  make this the one impl where `RustType != Self`.  In general, that's not going to work with proc-macros, since they can't know the right `FfiConverter` based on the type name alone.  However in this case, the proc-macros can check for the `dyn` keyword and use that in their logic.  Even then though, the user might alias the `dyn` away, for example:

```
type ButtonInterface = dyn Button;
fn get_buttons() -> Vec<Arc<ButtonInterface>> { ... }
```

Another option would be to always Box the Arc before passing it to the foreign code, but that doesn't seem great.

I think my preferred option would be to make the `FfiConverter` impl specific to the trait.  For each trait interface we see in the UDL (or export with a proc-macro eventually), we define `impl<UC> FfiConverter<UC> for dyn SpecificTrait` in the scaffolding code.  Maybe it could forward the method calls to functions in `uniffi_core` to avoid duplicating the code.

bendk (changes requested):
Very cool and I'm excited to start using it.

I think there's one `FfiConverter` issue to figure out before we can merge this, but everything else looks good to me.

mhammond:
This seems so so close - `cargo test -p uniffi-example-traits` works, but somehow it has broken most other tests - eg, `cargo test -p uniffi-fixture-coverall` is failing with errors for plain old structs, eg:

```
error[E0277]: the trait bound `Patch: Interface<UniFfiTag>` is not satisfied
   --> /mnt/c/src/moz/uniffi-rs/target/debug/build/uniffi-fixture-coverall-82c9c761f5ae2c4d/out/coverall.uniffi.rs:166:1
    |
166 | #[::uniffi::ffi_converter_record(tag = crate::UniFfiTag)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Interface<UniFfiTag>` is not implemented for `Patch`
    |
```

Which confuses me. Out of time for today, but if anyone has a clue please share it with me :)

mhammond:
So this all works and now uses a macro to implement FfiConverter - however, it's a bit different from how objects are done and I haven't even tried to support traits with proc-macros yet as I still need to understand how that works for structs! But it does seem to be in a fairly good place.

uniffi_bindgen/src/interface/object.rs
```diff
@@ -96,27 +101,44 @@ pub struct Object {
...
-            uses_deprecated_threadsafe_attribute: false,
         }
     }
 
     pub fn name(&self) -> &str {
         &self.name
     }
 
+    pub fn rust_name(&self) -> &str {
+        &self.rust_name
```
bendk, line 129: Could `rust_name` be derived from `name` and `imp`?

uniffi_bindgen/src/scaffolding/templates/ObjectTemplate.rs
```diff
@@ -10,35 +10,26 @@
...
-        {#- turn it into an Arc and explicitly drop it. #}
-        drop(unsafe { std::sync::Arc::from_raw(ptr as *const r#{{ obj.name() }}) })
+        {#- turn it into our FfiConverteran Arc and explicitly drop it. #}
+        {%- match obj.imp() -%}
+        {%- when ObjectImpl::Trait %}
+        {#- turn it into a Box<Arc<T> and explicitly drop it. #}
+        drop(unsafe { Box::from_raw(ptr as *mut std::sync::Arc<{{ obj.rust_name() }}>) })
+        {%- when ObjectImpl::Struct %}
+        drop(unsafe { std::sync::Arc::from_raw(ptr as *const {{ obj.rust_name() }}) })
+        {%- endmatch -%}
```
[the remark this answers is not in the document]
bendk: Reading this comment again, I still want to try to leverage the `FfiConverter` trait to handle this difference as was as the difference in dereferencing.  But I think that should be follow-up work, no need to change the code now.

bendk (approved):
This looks great to me.  I left a couple of small bikeshedding comments, but nothing big.  I'd be happy to merge this as-is.

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -34,6 +34,26 @@ pub(super) use finder::TypeFinder;
...
+    Struct,
+    Trait,
+}
+
+impl ObjectImpl {
+    /// Return the fully qualified name which should be used by Rust code for
+    /// an object with the given name.
+    /// Includes `r#`, traits get a leading `dyn`. If we ever supported associated types, then
+    /// this would also include them.
+    pub fn rust_name_for(&self, name: &str) -> String {
```
mhammond, line 48: This seems oddly named and vaguely mis-placed, but all other names and places seem worse :)

mhammond (commented):
I think this is ready to roll! Ben already approved it, but I think it's worth taking another look given the number of changes which happened underneath me since the last version.

PR #1463, opened 2023-01-30
Split out bindgen feature from full CLI

### LINKED ISSUE #1416: Split cli out of uniffi_bindgen

badboy:
This allows consumers to avoid the clap dependency if they don't require the full CLI (with argument parsing).

Fixes #1416

### DIFF
--- a/uniffi/Cargo.toml
+++ b/uniffi/Cargo.toml
@@ -30,9 +30,12 @@ default = []
 # Support for features needed by the `build.rs` script. Enable this in your
 # `build-dependencies`.
 build = [ "dep:uniffi_build" ]
+# Support for `uniffi_bindgen::{generate_bindings, generate_component_scaffolding}`.
+# Enable this feature for your `uniffi-bindgen` binaries if you don't need the full CLI.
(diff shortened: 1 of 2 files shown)


bendk (approved):
Looks good, thanks for catching this.

PR #1464, opened 2023-01-31
Make the proc-macro FfiConverter impls canonical

bendk:
* Split out the proc-macro code to generate `FfiConverter` implementations and added standalone macros for that.
* Removed the `FfiConverter` implementations from the Askama template code.  Instead, the templates now invoke the macros.
* Removed the blanket `Arc<T>` impl from `uniffi_core` and made it so we invoke a macro for each interface.

The reason for this is:

* I want to add some more functionality to `FfiConverter`, but don't want to implement it twice and test it two different ways.
* Removing the blanket `Arc<T>` implementation allows us to customize the `FfiConverter` code for each interface.  I think this makes #1457 easier.
* Everything goes through one code path, which should give us some more confidence as we migrate to proc-macros.  This change caused me to fix some bugs and add some features to them.

### DIFF
--- a/fixtures/uitests/tests/ui/fieldless_errors_used_in_callbacks_cant_have_fields.stderr
+++ b/fixtures/uitests/tests/ui/fieldless_errors_used_in_callbacks_cant_have_fields.stderr
@@ -1,11 +1,7 @@
-error[E0063]: missing field `numerator` in initializer of `ArithmeticError`
-  --> $OUT_DIR[uniffi_uitests]/errors.uniffi.rs
-   |
-   |             2 => r#ArithmeticError::r#DivisionByZero{ },
-   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `numerator`
-
-error[E0063]: missing field `0` in initializer of `ArithmeticError`
-  --> $OUT_DIR[uniffi_uitests]/errors.uniffi.rs
-   |
-   |             3 => r#ArithmeticError::r#UnexpectedError{ },
-   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `0`
+error[E0533]: expected unit struct, unit variant or constant, found struct variant `Self::DivisionByZero`
+ --> $OUT_DIR[uniffi_uitests]/errors.uniffi.rs
+  |
+  | #[::uniffi::ffi_converter_error(crate::UniFfiTag)]
+  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+  |
+  = note: this error originates in the attribute macro `::uniffi::ffi_converter_error` (in Nightly builds, run with -Z macro-backtrace for more info)
--- a/uniffi/tests/ui/proc_macro_arc.stderr
+++ b/uniffi/tests/ui/proc_macro_arc.stderr
@@ -1,31 +1,19 @@
-error[E0271]: type mismatch resolving `<Arc<child::Foo> as child::_::_::{closure#0}::TypeEq>::This == Arc<Foo>`
-  --> tests/ui/proc_macro_arc.rs:21:22
+error[E0277]: the trait bound `Foo: Interface<UniFfiTag>` is not satisfied
+  --> tests/ui/proc_macro_arc.rs:10:1
    |
-21 |     fn take_foo(foo: Arc<Foo>) {
-   |                      ^^^^^^^^ type mismatch resolving `<Arc<child::Foo> as child::_::_::{closure#0}::TypeEq>::This == Arc<Foo>`
+10 | #[uniffi::export]
+   | ^^^^^^^^^^^^^^^^^ the trait `Interface<UniFfiTag>` is not implemented for `Foo`
    |
-note: expected this to be `Arc<Foo>`
-  --> tests/ui/proc_macro_arc.rs:21:22
+   = help: the trait `FfiConverter<UT>` is implemented for `Arc<T>`
+   = note: required for `Arc<Foo>` to implement `FfiConverter<UniFfiTag>`
+   = note: this error originates in the attribute macro `uniffi::export` (in Nightly builds, run with -Z macro-backtrace for more info)
+
+error[E0277]: the trait bound `child::Foo: Interface<UniFfiTag>` is not satisfied
+  --> tests/ui/proc_macro_arc.rs:20:5
    |
-21 |     fn take_foo(foo: Arc<Foo>) {
-   |                      ^^^
-   = note: enum `child::Foo` and struct `Foo` have similar names, but are actually distinct types
-note: enum `child::Foo` is defined in module `crate::child` of the current crate
-  --> tests/ui/proc_macro_arc.rs:18:5
+20 |     #[uniffi::export]
+   |     ^^^^^^^^^^^^^^^^^ the trait `Interface<UniFfiTag>` is not implemented for `child::Foo`
    |
-18 |     enum Foo {}
-   |     ^^^^^^^^
-note: struct `Foo` is defined in module `crate` of the current crate
-  --> tests/ui/proc_macro_arc.rs:8:1
-   |
-8  | pub struct Foo;
-   | ^^^^^^^^^^^^^^
-note: required by a bound in `child::_::_::{closure#0}::assert_type_eq_all`
-  --> tests/ui/proc_macro_arc.rs:21:22
-   |
-21 |     fn take_foo(foo: Arc<Foo>) {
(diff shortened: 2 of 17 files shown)


fixtures/uitests/tests/ui/fieldless_errors_used_in_callbacks_cant_have_fields.stderr
```diff
@@ -1,11 +1,13 @@
...
+  | / ::uniffi::ffi_converter_flat_error_with_read!(
+  | |     r#ArithmeticError
+  | |     {
+  | |         r#IntegerOverflow {
+... |
+  | |     }
+  | | );
+  | |_^
+  |
+  = note: this error originates in the macro `::uniffi::ffi_converter_flat_error_with_read` (in Nightly builds, run with -Z macro-backtrace for more info)
```
bendk, line 13: These error messages aren't very good, but they don't seem any worse that the previous ones.

uniffi_macros/src/lib.rs
```diff
@@ -144,6 +144,90 @@ pub fn include_scaffolding(component_name: TokenStream) -> TokenStream {
...
+    variants: syn::punctuated::Punctuated<syn::Variant, syn::Token![,]>,
+}
+
+impl syn::parse::Parse for FfiConverterEnumInput {
+    fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
+        let ident = input.parse()?;
+        let content;
+        syn::braced!(content in input);
+        let variants =
+            <syn::punctuated::Punctuated<syn::Variant, syn::Token![,]>>::parse_terminated(
```
jplatte, line 181: The exact type should be inferrable from the return expression.

```suggestion
            syn::punctuated::Punctuated::parse_terminated(
```

uniffi_macros/src/lib.rs
```diff
@@ -144,6 +144,90 @@ pub fn include_scaffolding(component_name: TokenStream) -> TokenStream {
...
+#[proc_macro]
+pub fn ffi_converter_record(input: TokenStream) -> TokenStream {
+    let input = syn::parse_macro_input!(input as FfiConverterRecordInput);
+    record::record_ffi_converter_impl(&input.fields, &input.ident).into()
+}
+
+struct FfiConverterEnumInput {
+    ident: syn::Ident,
+    variants: syn::punctuated::Punctuated<syn::Variant, syn::Token![,]>,
+}
```
jplatte, line 173: I think it would make sense to move this into the `enum_` module.

uniffi_macros/src/lib.rs
```diff
@@ -144,6 +144,90 @@ pub fn include_scaffolding(component_name: TokenStream) -> TokenStream {
     }.into()
 }
 
+struct FfiConverterRecordInput {
```
jplatte, line 147: Same here.

jplatte (commented):
Took a quick peek at the macro code, this is not a comprehensive review.

uniffi_macros/src/error.rs
```diff
@@ -84,27 +87,49 @@ pub(crate) fn flat_error_ffi_converter_impl(
...
+    } else {
+        quote! { ::std::panic!("try_read not supported for flat errors") }
+    };
+
     quote! {
         #[automatically_derived]
-        unsafe impl<T> ::uniffi::FfiConverter<T> for #ident {
-            ::uniffi::ffi_converter_rust_buffer_lift_and_lower!(T);
+        unsafe impl ::uniffi::FfiConverter<crate::UniFfiTag> for #ident {
+            ::uniffi::ffi_converter_rust_buffer_lift_and_lower!(crate::UniFfiTag);
```
bendk, line 125: This changes the code to only implement `FfiConverter<crate::UniFffiTag>`, meaning other crates don't get an automatic implementation.  I made the change because:

- I want to move the interface `FfiConverter` impls to the generate code, but that can't be a blanket impl since `Arc<T>` is not a local type
- The blanket impl doesn't work for remote types and this is currently a requirement for the UDL-based code
- I think we can still make sharing types between crates ergonomic.  What if the requirement was that users needed to write `uniffi_use!(external_crate::ExternalType)`?  That seems reasonable to me.  I hope to have another PR, based on top of this, that implements this.

jplatte: I'm still against changing the default. How about having UDL codegen communicate to the proc-macro that it wants this sort of impl? Something like

```rust
#[derive(uniffi::Enum)]
#[uniffi(local_impl)]
enum Foo { ... }
```

bendk: Yes, we can do that.

uniffi_macros/src/error.rs
```diff
@@ -143,24 +197,39 @@ fn error_metadata(
...
+                ..Self::default()
+            })
+        } else if lookahead.peek(kw::with_try_read) {
+            Ok(Self {
+                with_try_read: input.parse()?,
+                ..Self::default()
+            })
+        } else {
+            Err(lookahead.error())
+        }
```
bendk, line 224: This is my best guess at how to handle more attributes here, please tell me if there's a better way to do this.

bendk:
Pushed an update based on the comments from @jplatte:
 - Switched the macros to use `DeriveInput`, which removes the need for parsing code
 - Added code to control when the macros generate `FfiConverter` implementations for all `<T>` vs a specific `UniFfiTag`.

uniffi_macros/src/error.rs
jplatte: Exactly what I would have written 👍🏼

uniffi_macros/src/object.rs
```diff
@@ -29,7 +30,24 @@ pub fn expand_object(input: DeriveInput, module_path: Vec<String>) -> TokenStrea
...
+        Err(e) => return e.into_compile_error(),
+    };
+    interface_impl(&input.ident, tag_handler)
+}
+
+pub(crate) fn interface_impl(ident: &Ident, tag_handler: FfiConverterTagHandler) -> TokenStream {
+    let (impl_spec, _) = tag_handler.into_impl_and_tag_path("Interface", ident);
+    quote! {
+        #[doc(hidden)]
+        #impl_spec { }
```
jplatte, line 51:
```suggestion
        #[automatically_derived]
        #impl_spec { }
```

uniffi_macros/src/util.rs
```diff
@@ -221,3 +222,72 @@ pub fn either_attribute_arg<T: ToTokens>(a: Option<T>, b: Option<T>) -> syn::Res
...
+            Some(tag) => (quote! { impl ::uniffi::#trait_name<#tag> for #ident }, tag),
+            None => (
+                quote! { impl<T> ::uniffi::#trait_name<T> for #ident },
+                Ident::new("T", Span::call_site()).into(),
+            ),
+        }
+    }
+}
+
+impl TryFrom<AttributeArgs> for FfiConverterTagHandler {
```
jplatte, line 251: Why use syn's `Meta` types instead of parsing the attributes using `syn::parse::Parse` like elsewhere? I already added a bunch of helpers to make that easier.

jplatte: I guess this is also the reason you didn't do `tag = crate::Tag` syntax, because `syn::Meta` doesn't support that?

bendk: I tried, but couldn't figure out a good way to get from a `TokenStream` to an `[Attribute]` slicte to use with `parse_uniffi_attributes()`.  Maybe we could handle this as a follow-up.

jplatte: Yeah, I think I can take care of that next week.

jplatte (approved):

mhammond (approved):
This is a great improvement, thanks.

uniffi/tests/ui/proc_macro_arc.stderr
```diff
@@ -1,31 +1,19 @@
...
+10 | #[uniffi::export]
+   | ^^^^^^^^^^^^^^^^^ the trait `Interface<UniFfiTag>` is not implemented for `Foo`
    |
-note: expected this to be `Arc<Foo>`
-  --> tests/ui/proc_macro_arc.rs:21:22
+   = help: the trait `FfiConverter<UT>` is implemented for `Arc<T>`
+   = note: required for `Arc<Foo>` to implement `FfiConverter<UniFfiTag>`
+   = note: this error originates in the attribute macro `uniffi::export` (in Nightly builds, run with -Z macro-backtrace for more info)
+
+error[E0277]: the trait bound `child::Foo: Interface<UniFfiTag>` is not satisfied
```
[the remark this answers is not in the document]
bendk: I think I'm missing something because both error messages don't seem very clear to me.  I'm happy to work on improving these as a followup though.

PR #1465, opened 2023-02-01
Fix clippy 0.1.67

yukibtc:


rust-toolchain.toml
```diff
@@ -5,7 +5,7 @@
 # * ./.circleci/config.yml which also specifies the rust versions used in CI.
 
 [toolchain]
-channel = "1.66.0"
+channel = "1.67.0"
```
bendk, line 8: As per our policy, I don't think we should bump this yet.

bendk (changes requested):
The code changes look good, but I don't think we should be bumping the Rust version.  We have a [policy](https://github.com/mozilla/uniffi-rs/blob/main/docs/policies/rust-versions.md) for that, based on the Firefox Rust version.  The min version is currently `1.65` and the max version is `1.66`.

bendk (approved):
The changes are great, thanks for the contribution!

PR #1467, opened 2023-02-02
Add more error context for uniffi_bindgen

jplatte:

PR #1468, opened 2023-02-03
Add more error context for proc-macro metadata extraction

jplatte:
Follow-up to #1467.


bendk (approved):
Looks good, thanks for making these small but important improvements.

PR #1469, opened 2023-02-03
Use `FfiConverter` for metadata generation/storage

bendk:
This PR makes it so proc-macros aren't responsible for parsing type names to figure out information about the underlying type.  Parsing type names is complicated and impossible to get 100% correct since users can use type aliases.

The main motivation for this is to try out the macro code in application-services.  Currently, we can't use them because we type alias our `Result<>`  types.

The proc-macros needed to parse type names was to generate the interface metadata and store it in the binary.  Now the proc-macros generate `const` functions/expressions to create the static variables.  A key part here is that `FfiConverter` now stores a const `META_TYPE_ID` value, which identifies the type.

The other reason the proc-macros parsed the type names was to handle function returns.  We had two different code paths for this, depending on if the return type was a `Result<>` or not.  This changes things to only have 1 code path, which means the proc-macros can ignore this issue.

This is a WIP because there's still a couple of todo items:
 - [ ] Update the checksum code.  One side-effect of the changes is that we no longer can use the checksum in the FFI function name.  If we decide to go down this route then we need a different system for checksums.  I think what we can do is to store the checksum in a separate symbol, then have the bindings check that value against their expected checksum when loading the library.
 - [ ] Make sure that the uitests [error messages aren't going to change each release](https://github.com/mozilla/uniffi-rs/pull/1469/files#r1098236174)


jplatte:
Based on the PR description, this sounds really great! Going to review next week.

uniffi_core/src/lib.rs
```diff
@@ -124,13 +129,34 @@ macro_rules! assert_compatible_version {
...
+/// means fighting with `rustc` over what counts as a conflicting implemmentation.  In fact, as of
+/// Rust 1.66, that previous example conflicts with `impl<UT> FfiReturn<UT> for ()`, since other
+/// crates can implement `impl FfiReturn<MyLocalType> for ()`. In general, this path gets
+/// complicated very quickly and that distracts from the logic that we want to define, which is
+/// fairly simple.
+///
+/// The main downside of having a single `FfiConverter` trait is that we need to implement it for
+/// types that only support some of the functionality.  For example, `Result<>` supports returning
+/// values, but not lifting/lowering/serializing them.  This is a bit weird, but works okay --
+/// especially since `FfiConverter` is rarely used outside of generated code.
```
bendk, line 158: This is what I'm currently thinking, but I'm not 100% against splitting them up.  We would just need to split them up in a way where we don't have trait implementations dependent on other ones.  Maybe we could follow @jplatte's suggestion and split it up into `Lift`, `Lower`, and `Metadata`.  However, I just don't see a need to make the API super clean since the main consumer is our generate code.

docs/manual/src/internals/object_references.md
```diff
@@ -56,91 +56,30 @@ interface TodoList {
...
-code remains alive.
+On the Rust side of the generated bindings:
+ - The instance constructor will create an instance of the corresponding `TodoList` Rust struct
+ - The owned value is wrapped it in an `Arc<>`
+ - The `Arc<>` is lowered into the foreign code using `Arc::into_raw` and returned as an object pointer.
+
+This is the "arc to pointer" dance. Note that this has "leaked" the `Arc<>`
+reference out of Rusts ownership system and given it to the foreign-language
+code. The foreign-language code must pass that pointer back into Rust in order
+to free it, or our instance will leak.
```
mhammond, line 67: (On one hand I'm sad to see this example generated code be removed, but OTOH it's going to get even more out of date than it already is. I wonder if it might make sense to add a brief and vague note somewhere that people who want to deeply understand the generated code should look at using the `cargo expand` utility?)

docs/manual/src/internals/object_references.md
```diff
@@ -56,91 +56,30 @@ interface TodoList {
...
-        uniffi::deps::log::error!("ffi_todolist_12ba_TodoList_object_free panicked: {:?}", e);
-    }
-}
-```
+the Rust object will be dropped.).  This simply calls `Arc::from_raw`, then
+lets the value drop.
+
 
 Passing instances as arguments and returning them as values works similarly, except that
 UniFFI does not automatically wrap/unwrap the containing `Arc`.
```
mhammond, line 85: On one hand I'm sad to see the example generated code be removed, but OTOH it's going to get even more out of date than it already is. I wonder if it might make sense to add a brief and vague note here (or somewhere else) that notes people who want to inspect or deeply understand the generated code should look at using the cargo expand utility?)

fixtures/keywords/rust/src/lib.rs
```diff
@@ -24,7 +24,7 @@ impl r#break {
 }
 
 #[allow(non_camel_case_types)]
-trait r#continue {
+pub trait r#continue {
```
mhammond, line 27: huh - if this is now necessary, is this technically a breaking change? (TBH I'm fine if it is but should maybe be documented)

fixtures/metadata/src/tests.rs
```diff
@@ -0,0 +1,440 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/// This entire crate is just a set of tests for metadata handling.  We use a separate crate
+/// for testing because the metadata handling is split up between several crates, and no crate
+/// on all the functionality.
```
mhammond, line 7: s/on/owns/ here?

fixtures/uitests/tests/ui/non_hashable_record_key.stderr
```diff
@@ -1,60 +1,14 @@
...
-  |
-  |     <std::collections::HashMap<f32, u64> as uniffi::FfiConverter<crate::UniFfiTag>>::lower(r#get_dict())
-  |                                                                                            ^^^^^^^^^^ not found in this scope
-
 error[E0277]: the trait bound `f32: Hash` is not satisfied
  --> $OUT_DIR[uniffi_uitests]/records.uniffi.rs
   |
-  |     <std::collections::HashMap<f32, u64> as uniffi::FfiConverter<crate::UniFfiTag>>::lower(r#get_dict())
-  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Hash` is not implemented for `f32`
+  | / pub extern "C" fn r#uitests_82f4_get_dict(
```
mhammond, line 4: isn't this hash here and below going to be a problem?

uniffi_core/src/lib.rs
```diff
@@ -124,13 +129,34 @@ macro_rules! assert_compatible_version {
...
 ///
+/// ## Scope
+///
+/// It could be argued that FfiConverter handles too many concerns and should be split into
+/// separate traits (like `FfiLiftAndLower`, `FfiSerialize`, `FfiReturn`).  However, there are good
+/// reasons to keep all the functionality in one trait.
+///
+/// The main reason is that splitting the traits requires fairly complex blanket implementations,
+/// for example `impl<UT, T> FfiReturn<UT> for T: FfiLiftAndLower<UT>`.  Writing these impls often
+/// means fighting with `rustc` over what counts as a conflicting implemmentation.  In fact, as of
```
mhammond, line 149: typo: implementation

uniffi_core/src/lib.rs
```diff
@@ -153,6 +179,11 @@ pub unsafe trait FfiConverter<UT>: Sized {
     /// serialization is simpler and safer as a starting point.
     type FfiType;
 
+    /// Type that the should be returned by scaffolding functions for this type.
```
mhammond, line 182: typo: The type that ...

mhammond (commented):
Far out, this is a big patch and I didn't get through it - but there's alot to like here. I'll get back to it soon, but @jplatte's review is probably the one you should care about :)

fixtures/keywords/rust/src/lib.rs
bendk: I don't think this breaks anything, the issue was at the rust level:
```
warning: private trait `continue` in public interface (error E0445)
   --> /home/bdk/uniffi-rs/ffi-converter-metadata/target/debug/build/uniffi-fixture-keywords-rust-ae8b10cd8ad52109/out/keywords.uniffi.rs:225:5
    |
225 | /     pub extern "C" fn r#keywords_rust_2de3_break_continue(
226 | |         r#ptr: *const std::os::raw::c_void,
227 | |         r#continue: uniffi::RustBuffer,
228 | |     call_status: &mut uniffi::RustCallStatus
229 | |     )  -> <std::option::Option<std::vec::Vec<Box<dyn r#continue>>> as ::uniffi::FfiConverter<crate::UniFfiTag>>::ReturnType {
    | |___________________________________________________________________________________________________________________________^
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #34537 <https://github.com/rust-lang/rust/issues/34537>
    = note: `#[warn(private_in_public)]` on by default

warning: `uniffi-fixture-keywords-rust` (lib) generated 1 warning
```

fixtures/uitests/tests/ui/non_hashable_record_key.stderr
bendk: I could be missing it, but I don't see why.  The error message seems okay to me, what else do we need to watch out for?

docs/manual/src/internals/object_references.md
bendk: I felt pretty mixed about this as well.  I like the idea of pushing people towards using `cargo expand` though, that seems like the best way to handle it.

fixtures/uitests/tests/ui/non_hashable_record_key.stderr
mhammond: I assumed it meant that these `.stderr` files would now require changing each release?

bendk: Ahhh, I see your point now and that would be very annoying.  Maybe it's not going to be an issue, since I need to change the checksum code anyways.  Now that the proc-macros don't know the metadata, they can't include the checksum in the function name.  Instead, they'll have to export that value as a separate symbol that the bindings code checks.  I think the askama templates should do the same, which hopefully will fix this.

jplatte:
Still on my todo, but I didn't manage to get to reviewing this week.

bendk:
Rebased this based on the async changes as well as the macro changes.  @jplatte do you think you can review this soon?  I'm really hoping to get it merged then test out macros with the application-services code.

jplatte:
I'll try.. 🥲
Will start on mobile now 😅

uniffi_meta/src/lib.rs
```diff
@@ -95,41 +98,50 @@ impl Checksum for &str {
     }
 }
 
-#[derive(Clone, Debug, Checksum, Deserialize, Serialize)]
+// The namespace of a Component interface
+//
+// This is used to match up the macro metadata with the UDL items
```
jplatte, line 103:
```suggestion
// The namespace of a Component interface.
//
// This is used to match up the macro metadata with the UDL items.
```

uniffi_meta/src/reader.rs
```diff
@@ -0,0 +1,321 @@
...
+        if !self.is_empty() {
+            let value = self[0];
+            *self = &self[1..];
+            Ok(value)
+        } else {
+            bail!("Buffer is empty")
+        }
+    }
+
+    fn peak_u8(&mut self) -> Result<u8> {
```
jplatte, line 84:
```suggestion
    fn peek_u8(&mut self) -> Result<u8> {
```

jplatte (approved):
Looks really really good! Left Some nitpicks.

jplatte:
With metadata being properly attributed to a namespaced type, is there any reason we should keep the one-crate restriction when generating bindings by the way?

fixtures/metadata/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "uniffi_fixture_metadata"
+version = "0.1.0"
+authors = ["Firefox Sync Team <sync-team@mozilla.com>"]
```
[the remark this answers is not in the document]
bendk: That's great news, I always find it annoying to fill out and rarely useful.

bendk:
> With metadata being properly attributed to a namespaced type, is there any reason we should keep the one-crate restriction when generating bindings by the way?

I hope not, in fact that's one of the first things I want to test when this gets merged.

uniffi_core/src/lib.rs
```diff
@@ -152,6 +178,11 @@ pub unsafe trait FfiConverter<UT>: Sized {
     /// serialization is simpler and safer as a starting point.
     type FfiType;
 
+    /// The type that the should be returned by scaffolding functions for this type.
```
badboy, line 181:
```suggestion
    /// The type that should be returned by scaffolding functions for this type.
```

uniffi_core/src/metadata.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    pub const fn from_code(value: u8) -> Self {
+        Self::new().concat_value(value)
+    }
+
+    // Concatenate another buffer to this one.
+    //
+    // This consumes self, which is convenient for the proc-macro code and also allows us to avoid
+    // allocated an extra buffer
```
badboy, line 109:
```suggestion
    // allocating an extra buffer.
```

uniffi_core/src/metadata.rs
```diff
@@ -0,0 +1,182 @@
...
+            self.size += 1;
+            i += 1;
+        }
+        self
+    }
+
+    // Concatenate a `u8` value to this buffer
+    //
+    // This consumes self, which is convenient for the proc-macro code and also allows us to avoid
+    // allocated an extra buffer
```
badboy, line 126:
```suggestion
    // allocating an extra buffer.
```

uniffi_core/src/metadata.rs
```diff
@@ -0,0 +1,182 @@
...
+        assert!(self.size < BUF_SIZE);
+        self.bytes[self.size] = value;
+        self.size += 1;
+        self
+    }
+
+    // Concatenate a `bool` value to this buffer
+    //
+    // This consumes self, which is convenient for the proc-macro code and also allows us to avoid
+    // allocated an extra buffer
```
badboy, line 137:
```suggestion
    // allocating an extra buffer.
```

uniffi_core/src/metadata.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub const fn concat_bool(self, value: bool) -> Self {
+        self.concat_value(value as u8)
+    }
+
+    // Concatenate a string to this buffer.
+    //
+    // Strings are encoded as a `u8` length, followed by the utf8 data.
+    //
+    // This consumes self, which is convenient for the proc-macro code and also allows us to avoid
+    // allocated an extra buffer
```
badboy, line 147:
```suggestion
    // allocating an extra buffer.
```

badboy (commented):
Didn't get through the important bits yet (but as Mark said: Jonas' review is the important one).
Minor nitpicks in wording, I'll get back to this tonight or tomorrow.

jplatte (approved):
:ship: it

jplatte:
> I think the last step for this is figuring out the checksum code, I hope to have that out soon.

So something is still off then? CI is green 😄
Let me know if I can help :)

uniffi_bindgen/src/scaffolding/templates/namespace_metadata.rs
```diff
@@ -0,0 +1,12 @@
+
+/// Export namespace metadata.
+///
+/// See `uniffi_bindgen::macro_metadata` for how this is used.
+{%- let const_var = "UNIFFI_META_CONST_NAMESPACE_{}"|format(ci.namespace().to_shouty_snake_case()) %}
+{%- let static_var = "UNIFFI_META_NAMESPACE_{}"|format(ci.namespace().to_shouty_snake_case()) %}
+
+const {{ const_var }}: ::uniffi::MetadataBuffer = ::uniffi::MetadataBuffer::from_code(::uniffi::metadata::codes::NAMESPACE)
+    .concat_str(module_path!()) // This is the crate name, since we're in the crate root
```
jplatte, line 9: Could use `env!("CARGO_CRATE_NAME")` here. ([docs](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates))

bendk:
I just pushed what's hopefully the last bit of functionality here and removed the `[WIP]` label, I think this one is maybe ready to go now.

The last commit implemented a new checksum system, which the new metadata system required.   I'd love feedback from on this one.  It also updated the proc-macro code slightly, @jplatte do the changes look okay to you?

There's LOC count is very high on this one, but there are some reasons to be okay with that:
  - A big portion is the proc-macro system which is somewhat experimental and @jplatte has approved those
  - The other main change is the checksum code, but it's accompanied by scripts that can be used to verify what happens when the checksum is off.  (Also those test scripts make up a large number of lines).

FWIW, here's some sample output of the checksum scripts.  This shows an example of what a checksum error will look like with these changes:
```
+ python3 python_test.py
  File "/home/bdk/uniffi-rs/ffi-converter-metadata/target/version-mismatch-workdir/python_test.py", line 1, in <module>
    from fixture_version_mismatch import *
  File "/home/bdk/uniffi-rs/ffi-converter-metadata/target/version-mismatch-workdir/fixture_version_mismatch.py", line 381, in <module>
    _UniFFILib = loadIndirect()
  File "/home/bdk/uniffi-rs/ffi-converter-metadata/target/version-mismatch-workdir/fixture_version_mismatch.py", line 361, in loadIndirect
    uniffi_check_api_checksums(lib)
  File "/home/bdk/uniffi-rs/ffi-converter-metadata/target/version-mismatch-workdir/fixture_version_mismatch.py", line 376, in uniffi_check_api_checksums
    raise InternalError("UniFFI API checksum mismatch: try cleaning and rebuilding your project")
fixture_version_mismatch.InternalError: UniFFI API checksum mismatch: try cleaning and rebuilding your project
```

mhammond (approved):
Wow, yeah, this is a big patch :) It's a bit of a shame the checksum code has grown a fair bit of complexity, but I won't pretend to understand everything well enough to try and suggest it's not necessary. So I'm still +1 on getting this landed sooner rather than later. Note there remain a number of unresolved comments which should probably get addressed before landing too.

jplatte:
I don't fully understand why `module_path!()` didn't work out, it's a bad sad that we're back to the other solution. But it's probably fine. There is just thing I'm curious about, if you declare two identical types with the same name in different modules of one crate, what will happen? They should get exactly identical metadata now, I think (but that might not be a problem).

PR #1471, opened 2023-02-09
Added changelog entry for a potential gotcha with the `UniFfiTag` changes

bendk:


CHANGELOG.md
```diff
@@ -14,6 +14,10 @@
 
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.23.0...HEAD).
 
+### What's changed
```
bendk, line 17: Should this actually be listed as a breaking change?  Code that worked before will no longer work.  That said, [we specifically say to call the macro in `lib.rs`](https://mozilla.github.io/uniffi-rs/tutorial/Rust_scaffolding.html#rust-scaffolding-code), so that code was only accidentally working.

badboy:
Oh. 
Glean is using it in a submodule in order to `#[allow(missing_docs)]` (we build with `deny` otherwise).

Under which circumstances would we hit this problem?
It doesn't seem to happen in my quick test against the `main` branch of uniffi.

bendk:
If you use a submodule, then `use submod::*` from the crate root, I think it should be fine.  It's only if you're doing something like we are in `places` where we didn't import it into the crate root.

badboy:
> If you use a submodule, then `use submod::*` from the crate root, I think it should be fine. It's only if you're doing something like we are in `places` where we didn't import it into the crate root.

Got it. Then Glean is fine.

mhammond (approved):
> Should this actually be listed as a breaking change?

Wouldn't quibble either way

Issue #1301, opened 2022-07-27
Reduce redundency in the bindings code


bendk:
The bindings code generates `FFIConverter` instances for each type it uses.  We often bundle the bindings for several UDL files together, which results in redundant `FFIConverter` definitions (for example if 2 UDL files both use `Sequence<>`, `Option<>`, `Map<>` with the same generics).  The worst one of these is the `RustBuffer` code, which is always duplicated and this causes issues with external types.

It would be nice if there was a way for UniFFI to factor out the common code, put it in a shared module, and have each bindings module import from that module.

bendk:
One of the worst parts for this is that each foreign module opens its own copy of the Rust dylib.  Not only does this waste memory, but it makes initialization tricky (See #1472)

PR #1472, opened 2023-02-09
Add support for external callback interfaces

stan-irl:
## Background
- I just tried to ingest the changes from https://github.com/mozilla/uniffi-rs/pull/1456 but my callback interface still failed.
- I originally made this PR to add some fixtures that would reproduce the issue
- I tried to implement the CallbackInterface while I was messing around in here but I cant get all the tests to pass. I think there is a problem with the platform binding generation. 

## Changes
- Add support for `ExternalKind::CallbackInterface`

## Testing
- [x] Added test cases
- [ ] Tests pass
- [x] Added external callback fixture
- [x] Manually built the fixtures lib which happily compiled

### DIFF
--- a/uniffi_bindgen/src/interface/types/mod.rs
+++ b/uniffi_bindgen/src/interface/types/mod.rs
@@ -85,6 +85,7 @@ pub enum Type {
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Checksum, Ord, PartialOrd)]
 pub enum ExternalKind {
     Interface,
+    CallbackInterface,
     // Either a record or enum
     DataClass,
 }
@@ -200,6 +201,10 @@ impl From<&Type> for FfiType {
                 kind: ExternalKind::Interface,
                 ..
             } => FfiType::RustArcPtr(name.clone()),
+            Type::External {
+                kind: ExternalKind::CallbackInterface,
+                ..
+            } => FfiType::UInt64,
             Type::External {
                 name,
                 kind: ExternalKind::DataClass,
--- a/uniffi_bindgen/src/scaffolding/mod.rs
+++ b/uniffi_bindgen/src/scaffolding/mod.rs
@@ -91,6 +91,13 @@ mod filters {
             } => {
                 format!("<::std::sync::Arc<r#{name}> as uniffi::FfiConverter<crate::UniFfiTag>>")
             }
+            Type::External {
+                name,
+                kind: ExternalKind::CallbackInterface,
+                ..
+            } => {
+                format!("<Box<dyn r#{name}> as uniffi::FfiConverter<crate::UniFfiTag>>")
+            }
             _ => format!(
                 "<{} as uniffi::FfiConverter<crate::UniFfiTag>>",
                 type_rs(type_)?
--- a/uniffi_bindgen/src/scaffolding/templates/ExternalTypesTemplate.rs
+++ b/uniffi_bindgen/src/scaffolding/templates/ExternalTypesTemplate.rs
@@ -8,6 +8,8 @@
 ::uniffi::ffi_converter_forward!(r#{{ name }}, ::{{ crate_name|crate_name_rs }}::UniFfiTag, crate::UniFfiTag);
 {%- when ExternalKind::Interface %}
 ::uniffi::ffi_converter_forward!(::std::sync::Arc<r#{{ name }}>, ::{{ crate_name|crate_name_rs }}::UniFfiTag, crate::UniFfiTag);
+{%- when ExternalKind::CallbackInterface %}
+::uniffi::ffi_converter_forward!(Box<dyn r#{{ name }}>, ::{{ crate_name|crate_name_rs }}::UniFfiTag, crate::UniFfiTag);
 {%- endmatch %}
 {%- endfor %}
 
--- a/fixtures/ext-types/lib/src/ext-types-lib.udl
+++ b/fixtures/ext-types/lib/src/ext-types-lib.udl
@@ -31,7 +31,7 @@ typedef extern UniffiOneEnum;
 [ExternalInterface="uniffi-one"]
 typedef extern UniffiOneInterface;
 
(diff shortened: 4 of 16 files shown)


stan-irl:
@bendk I'm not sure if this is a bug or you dont plan to support this usecase. Feel free to close

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -200,6 +201,10 @@ impl From<&Type> for FfiType {
                 kind: ExternalKind::Interface,
                 ..
             } => FfiType::RustArcPtr(name.clone()),
+            Type::External {
+                kind: ExternalKind::CallbackInterface,
+                ..
+            } => FfiType::UInt64,
```
stan-irl, line 207: not super sure about this - I saw that the regular callback interface is a UInt64 so I just did the same thing here

uniffi_bindgen/src/scaffolding/mod.rs
```diff
@@ -91,6 +91,13 @@ mod filters {
             } => {
                 format!("<::std::sync::Arc<r#{name}> as uniffi::FfiConverter<crate::UniFfiTag>>")
             }
+            Type::External {
+                name,
+                kind: ExternalKind::CallbackInterface,
+                ..
+            } => {
+                format!("<Box<dyn r#{name}> as uniffi::FfiConverter<crate::UniFfiTag>>")
```
stan-irl, line 99: Again - just trying to handle this the same as a regular callback interface

uniffi_bindgen/src/scaffolding/templates/ExternalTypesTemplate.rs
```diff
@@ -8,6 +8,8 @@
 ::uniffi::ffi_converter_forward!(r#{{ name }}, ::{{ crate_name|crate_name_rs }}::UniFfiTag, crate::UniFfiTag);
 {%- when ExternalKind::Interface %}
 ::uniffi::ffi_converter_forward!(::std::sync::Arc<r#{{ name }}>, ::{{ crate_name|crate_name_rs }}::UniFfiTag, crate::UniFfiTag);
+{%- when ExternalKind::CallbackInterface %}
+::uniffi::ffi_converter_forward!(Box<dyn r#{{ name }}>, ::{{ crate_name|crate_name_rs }}::UniFfiTag, crate::UniFfiTag);
```
stan-irl, line 12: I dont quite understand the difference between this and uniffi_bindgen/src/scaffolding/mod.rs but I've just tried to do the same thing

stan-irl:
I'm not sure how to finish this off. I dont think the `external` bindings for the kotlin/python/swift will work here

mhammond:
I don't really see a reason to not support this - I'd actually like to deprecate callbacks entirely in favor of better support for traits, but that's not going to happen in the short term and these changes do seem sane. Don't the foreign tests just need the relevant imports added in the same way the other external types are referenced? You might find visibility issues (eg, callbacks being private where external ones need to be public or something) though.

stan-irl:
It's failing because no handles to lift/lower from other crates are generated for callback interfaces when they are for custom types and externals. I will try and add this this afternoon.

bendk (changes requested):
These changes look good to me, we just need Python and Swift support.

I pushed a commit to my branch that does that: https://github.com/bendk/uniffi-rs/tree/stan_repro_extern_interface_bug

stan-irl:
@bendk I pulled your changes and the existing tests passed but I added some new ones that fail. Im getting panics in the generated rust scaffolding:
```
impl Drop for UniFFICallbackHandlerUniffiOneCallbackInterface {
    fn drop(&mut self) {
        let callback = FOREIGN_CALLBACK_UNIFFIONECALLBACKINTERFACE_INTERNALS.get_callback().unwrap();
        let mut rbuf = uniffi::RustBuffer::new();
        unsafe { callback(self.handle, uniffi::IDX_CALLBACK_FREE, Default::default(), &mut rbuf) };
    }
}
```

because `FOREIGN_CALLBACK_UNIFFIONECALLBACKINTERFACE_INTERNALS.get_callback()` is returning `None`.

do you know how I can fix this? we're getting pretty deep into the lib. the changes I pulled from you were way over my head

bendk:
We are definitely getting deep in the lib, thanks for working this out for me.  And unfortunately, we are kind of running into a wall.

Here's the issue:
 - Callback interfaces require the foreign bindings module to register a handler function with Rust.  Rust calls this function when it wants to invoke a callback.
 - With external types, each bindings module opens its own copy of the Rust dynamic library.
 - Therefore if you want to use an external callback interface, you need to register the handler function with your copy of the dylib.


I tried to fix things with [this commit](https://github.com/bendk/uniffi-rs/commit/2a1a9790802d20f7e535b603d2f7c4f88479c2b1).  But then I ran into the second issue.  On Kotlin, each module also defines foreign library as a JNA class and the registration function expects to input that type.  However, since this prevents us from registering the callback with the library from another module.  It still works on Python, because Python uses duck-typing.  It also works on Swift without needing to re-register, I think because Swift directly links to the library file.

So, I think we could land support for Python and Swift right now, but Kotlin isn't ready.  Maybe there's some way to hack around the type issue I mentioned, but I can't think of one.

The long-term fix is to create some system where we can have a shared module and therefore shared types (#1301).

stan-irl:
ok I understand. that a bummer but if its not currently possible then I guess theres not much we can do.

I'll continue with my workaround of manually merging UDL for now.

thanks for looking into this. I appreciate that this isn't a use case that you need and you still put a lot of time into it 🙂

stan-irl:
should i close this PR?

mhammond:
Sorry it didn't work out, but yeah, we might as well close this - thanks for trying though!

bendk:
Yeah, seems like closing is the best option for now since you have a workaround.  I really appreciate the work you put in on this one.  I'm happy to look into it, since it helped me understand the corner case better.

PR #1473, opened 2023-02-12
Update module.md

heckj:
I thought about suggesting that the code be changed to always name the module `module.modulemap`, but including some documentation for folks who are attempting to create an XCFramework solves the same issue. (It took me an embarrassingly long time to find out that _not_ having this file named `module.modulemap` was the cause for an otherwise silent failure, reporting the module was unable to be imported in Swift.

### DIFF
--- a/docs/manual/src/swift/module.md
+++ b/docs/manual/src/swift/module.md
@@ -23,3 +23,7 @@ swiftc
 
 This will produce an `example.swiftmodule` file that can be loaded by
 other Swift code or used from the Swift command-line REPL.
+
+If you are creating an XCFramework with this code, make sure to rename the modulemap file
+to `module.modulemap`, the default value expected by Clang and XCFrameworks for exposing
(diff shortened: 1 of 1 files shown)


badboy:
huh, interesting. From everything I could find `module.modulemap` _is_ the right choice like pretty much everytime.
Yet dependencies that Glean uses have a `$name.modulemap`, not sure if that's what they generate or is somehow added by other tooling.

I'm fine with merging the doc changes here as is, but I wouldn't be opposed to changing uniffi to produce `module.modulemap` all the time.

heckj:
It’s quite possible to use a named module map by providing the extra option directly to the swift compiler, as the example currently shows, which is why i led with a doc comment about use within an XCFramework, where you can’t provide such an option

badboy:
Let's go with this then for now.

badboy (approved):

Issue #1476, opened 2023-02-17
Is it possible to disable clippy rules for included scaffolding?


stan-irl:
I'm trying to setup some fairly strict rules on my codebase:
```
cargo clippy -- --all-targets --all-features -- \
	-D warnings \
	-D clippy::all \
	-D missing_docs \
	-D clippy::missing_docs_in_private_items \
	-D clippy::pedantic \
	-W clippy::nursery \
	-W clippy::cargo
```

The problem is that wherever I include scaffolding will fail because the generated code doesnt satisfy these rules. For example my lib.rs looks like this:
```
//! This crate contains a test component to use as a placeholder
use log::debug;

/// A basic no-op function that is
/// exposed through the interface
pub fn do_something() {
    debug!("Something");
}

/// A function that will always panic
///
/// # Panics
///
/// Will always panic
pub fn crash_now() {
    panic!("down i go...")
}

uniffi::include_scaffolding!("foo");
```

I can fix this by disabling these lints at the module level but that kinda defeats the purpose of setting them in the first place. the implementations need to be in the same place as `uniffi::include_scaffolding!()` so theres no way I can enforce stricter linting on my own code then you do in the scaffolding unless I use another layer of indirection which is also undesirable.

I've been trying stuff like:

```
#[allow(clippy::pedantic, clippy::missing_const_for_fn)]
uniffi::include_scaffolding!("foo");
``` 

but that doesnt work either. Do you have any recommendations for how to deal with this?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-239)
┆Issue Number: UNIFFI-239

stan-irl:
My main concern right now is that I cant run with `-D missing_docs`  which means that I have no way to enforce comments on the APIs I'm  exposing

mhammond:
I don't think we have a good story for this, but I think we'd be fine with unconditionally adding "reasonable" clippy exclusions, of which `missing-docs` would seem to qualify,

stan-irl:
No worries. I'll add it on monday. it should just be a one-liner right?

mhammond:
You might need a few one them - eg, one for function, enums, structs, etc?

mhammond:
(OTOH though, I wonder why anyone would *ever* want to see clippy errors in uniffi-generated code? Should we see if we can kill them all?)

stan-irl:
[...]

PR #1477, opened 2023-02-20
[CLIPPY] Update templates to allow generated scaffolding to pass pedantic clippy

stan-irl:
## Background
- https://github.com/mozilla/uniffi-rs/issues/1476
- The problem here is that rust scaffolding has to be directly included into the module where the implementation is defined or else is it wont compile. This means that a consumer's code cant have stricter clippy rules than the generated scaffolding.
- Theres not really a way to deal with this except to make the generated scaffolding pass all possible clippy rules
- I think this is fair because consumers of uniffi - I dont have control over the generated code so it should have no impact over their code formatting

## Changes
- Add exceptions of fix all clippy errors that fired with 
```
cargo clippy -- --all-targets --all-features -- \
	-D warnings \
	-D clippy::all \
	-D missing_docs \
	-D clippy::missing_docs_in_private_items \
	-D clippy::pedantic \
	-W clippy::nursery \
	-W clippy::cargo
```
- Add some exceptions so that uniffi scaffolding isnt included in generated cargo docs for consumers

## Tests
- [x] Ingested into my own project and verified by hand
- [x] All uniffi tests pass

### DIFF
--- a/uniffi_bindgen/src/scaffolding/templates/ReexportUniFFIScaffolding.rs
+++ b/uniffi_bindgen/src/scaffolding/templates/ReexportUniFFIScaffolding.rs
@@ -11,10 +11,11 @@
 // megazord).  The combined library has a cargo dependency for each component and calls
 // uniffi_reexport_scaffolding!() for each one.
 
+#[allow(clippy::missing_docs_in_private_items)]
 #[doc(hidden)]
-pub fn uniffi_reexport_hack() {
-}
+pub const fn uniffi_reexport_hack() {}
 
+#[doc(hidden)]
 #[macro_export]
 macro_rules! uniffi_reexport_scaffolding {
     () => {
--- a/uniffi_bindgen/src/scaffolding/templates/ReexportUniFFIScaffolding.rs
+++ b/uniffi_bindgen/src/scaffolding/templates/ReexportUniFFIScaffolding.rs
@@ -11,7 +11,7 @@
 // megazord).  The combined library has a cargo dependency for each component and calls
 // uniffi_reexport_scaffolding!() for each one.
 
-#[allow(clippy::missing_docs_in_private_items)]
+#[allow(missing_docs)]
 #[doc(hidden)]
 pub const fn uniffi_reexport_hack() {}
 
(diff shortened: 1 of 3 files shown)


stan-irl:
@mhammond FYI. I think we've just gotta fix these on a case-by-case basis

uniffi_bindgen/src/scaffolding/templates/macros.rs
```diff
@@ -112,7 +112,8 @@ uniffi::call_with_output(call_status, || {
     {% else -%}
     {% if func.full_arguments().is_empty() %}#[allow(clippy::redundant_closure)]{% endif %}
     {% call to_rs_call(func) %}
+    {%- if func.return_type().is_none() %};{% endif %}
     {% endmatch -%}
-})
+}){%- if func.return_type().is_none() -%};{% endif %}
```
stan-irl, line 117: add semicolons for funcs that dont return anything

uniffi_bindgen/src/scaffolding/templates/RustBuffer.rs
```diff
@@ -2,25 +2,29 @@
...
+#[doc(hidden)]
 #[no_mangle]
 pub unsafe extern "C" fn {{ ci.ffi_rustbuffer_free().name() }}(buf: uniffi::RustBuffer, call_status: &mut uniffi::RustCallStatus) {
-    uniffi::ffi::uniffi_rustbuffer_free(buf, call_status)
+    uniffi::ffi::uniffi_rustbuffer_free(buf, call_status);
 }
 
-#[allow(clippy::missing_safety_doc)]
+#[allow(clippy::missing_safety_doc, missing_docs)]
+#[doc(hidden)]
```
stan-irl, line 27: lets keep this out of consumer cargo docs and also allow them to be undocumented

uniffi_bindgen/src/scaffolding/templates/ReexportUniFFIScaffolding.rs
```diff
@@ -11,10 +11,11 @@
 // megazord).  The combined library has a cargo dependency for each component and calls
 // uniffi_reexport_scaffolding!() for each one.
 
+#[allow(missing_docs)]
 #[doc(hidden)]
-pub fn uniffi_reexport_hack() {
-}
+pub const fn uniffi_reexport_hack() {}
```
stan-irl, line 16: this should be `const` and  we also want to allow it to be undocumented

mhammond (approved):
seems fine to me, thanks!

PR #1479, opened 2023-02-28
Improve UniFFI tag handling in macros

jplatte:
About the first commit: I think down the line it makes sense to support `#[uniffi(tag = Tag)]` on fields to support pulling in an `FfiConverter` impl from another crate, but I don't think this is currently required anywhere, right?


bendk:
> About the first commit: I think down the line it makes sense to support `#[uniffi(tag = Tag)]` on fields to support pulling in an `FfiConverter` impl from another crate, but I don't think this is currently required anywhere, right?

It's not currently required.  Maybe we could support it down the line, I'm not sure.

bendk (approved):
The changes look good to me, let's merge.

PR #1481, opened 2023-03-04
Callback interface speedup

bendk:
This branch improves callback interface performance, which we realized could be an issue when using callbacks for logging.  The performance is okay for small numbers of calls, but it starts to matter once you start making 1000s of calls per second -- which we thought could maybe happen with logging.

This PR adds some very basic benchmarking that I used to test the performance improvements.  The benchmarks are quite noisy but any kind of numbers are nice when working on this.  I also started benchmarks for regular calls, but that's more of a proof-of-concept than anything, that code should definitely be double-checked before making code changes based on it.

The other thing I'm trying is to update the ABI, without creating a breaking change.  https://github.com/mozilla/uniffi-rs/commit/05115dfd7c6585c759471c30082836b94151785b describes how I think we can do that.  Please double check that my assumptions are correct here though.

### DIFF
--- a/uniffi_bindgen/src/bindings/mod.rs
+++ b/uniffi_bindgen/src/bindings/mod.rs
@@ -33,6 +33,13 @@ pub enum TargetLanguage {
     Ruby,
 }
 
+/// Mode for the `run_script` function defined for each language
+#[derive(Copy, Clone, Debug)]
+pub enum RunScriptMode {
+    Test,
+    PerformanceTest,
+}
+
 impl TryFrom<&str> for TargetLanguage {
     type Error = anyhow::Error;
     fn try_from(value: &str) -> Result<Self> {
--- a/uniffi_bindgen/src/bindings/mod.rs
+++ b/uniffi_bindgen/src/bindings/mod.rs
@@ -19,6 +19,13 @@ pub mod python;
 pub mod ruby;
 pub mod swift;
 
+// Express this feature as a static value so that we can use it in the template code
+#[cfg(feature = "new_callback_interface_abi")]
+static NEW_CALLBACK_INTERFACE_ABI: bool = true;
+
+#[cfg(not(feature = "new_callback_interface_abi"))]
+static NEW_CALLBACK_INTERFACE_ABI: bool = false;
+
 /// Enumeration of all foreign language targets currently supported by this crate.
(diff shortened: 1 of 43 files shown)


jhugman (approved):
I like this. I have looked at the Kotlin and Swift bindings, and it matching up with the Rust scaffolding.

I'll leave others to comment on the python and ruby implementations.

The ABI compat makes this a little confusing, but I see the necessity. ✅ from me.

uniffi_bindgen/src/bindings/mod.rs
```diff
@@ -19,6 +19,13 @@ pub mod python;
 pub mod ruby;
 pub mod swift;
 
+// Express this feature as a static value so that we can use it in the template code
+#[cfg(feature = "new_callback_interface_abi")]
```
mhammond, line 23: So yeah, this really is abusing features

mhammond (commented):
I'm really not sure how I feel about the feature abuse and it's not a tiny bit of code to wind back later. I think I'd be ok if it really did solve a pressing problem, but it also seems likely that @badboy will be "meh, updating uniffi in glean is nbd". It looks like he is back next week, so how about we wait for his opinion?

badboy:
> The other thing I'm trying is to update the ABI, without creating a breaking change. https://github.com/mozilla/uniffi-rs/commit/05115dfd7c6585c759471c30082836b94151785b describes how I think we can do that. Please double check that my assumptions are correct here though.

I haven't yet read the full change, but I agree with @mhammond here: the use of a cargo feature here is not optimal.
Major updates are a bit of hassle because of the required coordination, but it's doable. We need to coordinate one sometime soon anyway, why not on this change?

I will review the PR in full tomorrow.

bendk:
> I haven't yet read the full change, but I agree with @mhammond here: the use of a cargo feature here is not optimal.
> Major updates are a bit of hassle because of the required coordination, but it's doable. We need to coordinate one sometime soon anyway, why not on this change?

This makes sense to me.  I guess it was worth a try, but it doesn't seem like it's worth the extra work needed to avoid the breaking change.

And actually... don't review this one.  I've been thinking about this some more and I think we can actually get a much better performance improvement if we are willing to lean into a breaking ABI changes.

PR #1488, opened 2023-03-10
Implement RustArcPtrUnsafe for object FFI free function

arg0d:
This allows FFI free function to have a separate type from regular methods. This is important for C# bindings generator, because regular `RustArcPtr` is represented as native C# `SafeHandle`, while `RustArcPtrUnsafe` is represented as a `void*`. This distinction is required to be able to call free function from inside `SafeHandle::ReleaseHandle`.

https://github.com/NordSecurity/uniffi-bindgen-cs/blob/main/templates/ObjectTemplate.cs#L23


arg0d:
`OwnedRustArcPtr` does sound better in the context of `uniffi-rs`. `Unsafe` was just the name I came up on the spot, didn't have much inspiration to generalize it.

I'm not so sure about the idea to add a flag to `FfiArgument`. Having an additional flag besides the type would imply that any type could be "owned". Also, code gen would become really weird, because if the "owned" flag is true, that means a hardcoded `void*` should replace whatever the type would render. For example, if "owned" is set to true with `Float64`, does that mean that the float type should also be replaced with `void*`?

Maybe the additional flag could be added to `RustArcPtr` itself? That would avoid having to deal with an additional `FfiType` in generators that don't care about it.

bendk:
> Maybe the additional flag could be added to `RustArcPtr` itself? That would avoid having to deal with an additional `FfiType` in generators that don't care about it.

I like this method a lot.  You get the good ergonomics you mention and if we can add a `owned` flag to other FFITypes if we ever wanted to in the future.

uniffi_bindgen/src/interface/object.rs
```diff
@@ -289,7 +292,10 @@ impl Constructor {
     fn derive_ffi_func(&mut self, ci_prefix: &str, obj_name: &str) {
         self.ffi_func.name = format!("{ci_prefix}_{obj_name}_{}", self.name);
         self.ffi_func.arguments = self.arguments.iter().map(Into::into).collect();
-        self.ffi_func.return_type = Some(FfiType::RustArcPtr(obj_name.to_string()));
+        self.ffi_func.return_type = Some(FfiType::RustArcPtr {
+            inner: obj_name.to_string(),
+            owned: true,
```
arg0d, line 297: Constructor returns an owned ptr

uniffi_bindgen/src/interface/object.rs
```diff
@@ -170,7 +170,10 @@ impl Object {
         }
         self.ffi_func_free.arguments = vec![FfiArgument {
             name: "ptr".to_string(),
-            type_: FfiType::RustArcPtr(self.name().to_string()),
+            type_: FfiType::RustArcPtr {
+                inner: self.name().to_string(),
+                owned: true,
```
arg0d, line 175: Free function has an owned ptr as the argument

uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -199,7 +202,10 @@ impl From<&Type> for FfiType {
                 name,
                 kind: ExternalKind::Interface,
                 ..
-            } => FfiType::RustArcPtr(name.clone()),
+            } => FfiType::RustArcPtr {
+                inner: name.clone(),
+                owned: false,
```
arg0d, line 207: I'm hoping that that this never needs to be owned. Please correct me if I'm wrong, because I'm not very familiar with the use cases of these type conversions.

bendk:
I was hoping to try to generalize this a bit more, but that `impl From<&Type> for FfiType` block makes me realize that it's going to be a lot of work.  As it stands, I don't think you can implement that properly because:
  - `Type::Object` doesn't track if it's owned or not.  I don't think it's correct to assume it's always borrowed.
  - `Type::External` can wrap an object, so we would somehow need to track that as well.

To properly solve this I think we would need to overhaul our type system, I think that could be nice eventually since it would allow things like passing in owned objects to functions.  But for now, let's just try to solve your problem.  I think the requirement is to distinguish when we're passing a `RustArcPtr` to the free function.  So what if we go back to your original code, but call the variant `RustArcPtrForFree` or something similar?

arg0d:
Yes, I was thinking about this yesterday as well. `Type::Object` may have these use cases:
- inside a struct (dictionary). In this case, ptr should be owned, right? However, this would make this un-usable for my use case, because that would mean that structs in generated C# would contain `IntPtr` instead of `SafeHandle`.
- inside function/method calls. In this case, its non-owned just like it was made now.

I'm thinking that its possible to implement this in another way. I could add some metadata to `FfiFunction` to make it possible to distinguish it from other functions. Something like `is_object_free_function`. With this metadata, it would be possible to simply generate a separate variant of object free function without messing with the type system.

Please let me know what you think about this kind of approach.

bendk:
I haven't thought through all the details, but something like that makes a lot of sense to me because it's targetted at your specific issue.

uniffi_bindgen/src/interface/ffi.rs
```diff
@@ -64,7 +64,10 @@ pub struct FfiFunction {
     pub(super) is_async: bool,
     pub(super) arguments: Vec<FfiArgument>,
     pub(super) return_type: Option<FfiType>,
-    pub(super) object_free_function: bool,
+    pub(super) has_rust_call_status_arg: bool,
+    /// Used by C# generator to differentiate the free function and call call it with void*
```
jplatte:
```suggestion
    /// Used by C# generator to differentiate the free function and call it with void*
```

bendk (approved):
LGTM other than the typo.  I'll merge once that's updated.

uniffi_bindgen/src/interface/ffi.rs
arg0d: Fixed

PR #1489, opened 2023-03-10
Add additional functions to uniffi-fixture-time

arg0d:
`to_string_timestamp` and `get_pre_epoch_timestamp` separate 2-way back and forth conversion process, making it possible to test 1-way conversion.

`to_string_timestamp` ensures that foreign language bindings lower the timestamp correctly, and that foreign language bindings interpret the timestamp the same way that Rust does.

`get_pre_epoch_timestamp` ensures that foreign language bindings lift the timestamp correctly, and that Rust interprets the timestamp the same way that foreign language bindings do.


arg0d:
I wasn't planning on doing that. These are most useful when implementing bindings for a new language, while still debugging the initial implementation.

PR #1490, opened 2023-03-10
Add "return_type()" to Constructor

arg0d:
This makes Constructor more nominal with Function, i.e. the constructor can be used as a function in macros:
```
{% macro(func) %}
    {{ func.return_type() }}
{% endmacro %}

{% call macro(constructor) %}
```

This is very useful for generating Go bindings, because in Go errors are handled via multiple return values, instead of exceptions. That means that codegen for function return signatures get quite complicated, and being able to re-use the same code for constructors as for regular functions makes it much easier to generate code.

### DIFF
--- a/uniffi_bindgen/src/interface/function.rs
+++ b/uniffi_bindgen/src/interface/function.rs
@@ -249,6 +249,42 @@ impl APIConverter<Argument> for weedle::argument::SingleArgument<'_> {
     }
 }
 
+/// Implemented by function-like types (Function, Method, Constructor)
+pub trait Callable {
+    fn arguments(&self) -> Vec<&Argument>;
+    fn return_type(&self) -> Option<Type>;
+    fn throws_type(&self) -> Option<Type>;
+}
+
+impl Callable for Function {
+    fn arguments(&self) -> Vec<&Argument> {
+        self.arguments()
+    }
+
+    fn return_type(&self) -> Option<Type> {
+        self.return_type().cloned()
+    }
+
+    fn throws_type(&self) -> Option<Type> {
+        self.throws_type()
+    }
+}
+
+// Needed because Askama likes to add extra refs to variables
+impl<T: Callable> Callable for &T {
+    fn arguments(&self) -> Vec<&Argument> {
+        (*self).arguments()
+    }
+
+    fn return_type(&self) -> Option<Type> {
+        (*self).return_type()
+    }
+
+    fn throws_type(&self) -> Option<Type> {
+        (*self).throws_type()
+    }
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
--- a/uniffi_bindgen/src/interface/object.rs
+++ b/uniffi_bindgen/src/interface/object.rs
@@ -65,7 +65,7 @@ use uniffi_meta::Checksum;
 
 use super::attributes::{Attribute, ConstructorAttributes, InterfaceAttributes, MethodAttributes};
 use super::ffi::{FfiArgument, FfiFunction, FfiType};
-use super::function::Argument;
+use super::function::{Argument, Callable};
 use super::types::{Type, TypeIterator};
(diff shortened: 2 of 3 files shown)


bendk:
I was going to mention this when I looked at the C# bindings, but what do you think about adding a `Callable` trait to try to unify functions, methods, and constructors?  I have some [code to do that](https://github.com/mozilla/uniffi-rs/blob/b48013bfce63fdec313d168238178fcaa4c2fd2c/uniffi_bindgen/src/interface/function.rs#L258-L277) in an in-progress PR.  A trait is nicer to use in Rust code, like the [filter functions](https://github.com/mozilla/uniffi-rs/blob/b48013bfce63fdec313d168238178fcaa4c2fd2c/uniffi_bindgen/src/scaffolding/mod.rs#L104).

bendk:
Looks good to me, but I don't want to merge it myself, since that's sort-of like reviewing my own code.  @tarikeshaq what do you think of this approach?

uniffi_bindgen/src/interface/object.rs
```diff
@@ -290,6 +295,9 @@ impl Constructor {
         self.ffi_func.name = format!("{ci_prefix}_{obj_name}_{}", self.name);
         self.ffi_func.arguments = self.arguments.iter().map(Into::into).collect();
         self.ffi_func.return_type = Some(FfiType::RustArcPtr(obj_name.to_string()));
+
+        // this is a bit of a dirty place to put this, but there isn't another "general" pass
```
tarikeshaq, line 299: Can we expand why this is not an ideal place? Inline comment is great, but if we have a more actionable way of making it more ideal then a bug filed with more context would be phenomenal!

uniffi_bindgen/src/interface/function.rs
```diff
@@ -249,6 +249,42 @@ impl APIConverter<Argument> for weedle::argument::SingleArgument<'_> {
     }
 }
 
+/// Implemented by function-like types (Function, Method, Constructor)
+pub trait Callable {
```
tarikeshaq, line 253: out of curiosity, why is the `arguments` API return a vec of references where the other two return owned `Type`?

Is it because it was hard to have then all return references without overcomplicating the code?

Just curious!

tarikeshaq: Additionally, would we like to add a `name` function too?

tarikeshaq (approved):
Thank's a ton for contributing back, I'm super excited to see more backends (and a go backends sounds phenomenal!) 

Had a few nit but this Makes sense to me as long as it's useful for external backends!

uniffi_bindgen/src/interface/object.rs
bendk: [#1469 adds an `object_name` field to `Constructor`](https://github.com/mozilla/uniffi-rs/pull/1469/files#diff-db37fade9d78e8f842a50e60a8be13b5eaf0749ee6d0c448179549e5dc6695dc), which would avoid this.  Does that make sense for the long-term plan?

bendk:
#1469 ended up being merged before this one, should we just close this PR?

uniffi_bindgen/src/interface/function.rs
bendk: > out of curiosity, why is the `arguments` API return a vec of references where the other two return owned `Type`?

I think that was copied from the #1469 implementation.   I didn't put a lot of thought into it, but that seemed like the path of least resistance to me.  In general, it seems like we tend to clone Type rather than try to manage a reference to it.  I'm not sure if that's the best policy going forward or not.

> Additionally, would we like to add a name function too?

Maybe?  I'm not totally sure though, there could be callables that don't have a name.  For example a destructor.  Also, the primary constructor currently has a name field, but it's just hard coded to "new".  That said, I'm not really against it I just didn't put the time into defining it.

PR #1491, opened 2023-03-16
Docker fixes (size reduction, deps)

badboy:
This makes the Dockerfile actually work and reduces its size a bit (down to "just" 6 GB). I'm working on getting us to use that again in CI properly (migrating away from rfkelly's personal docker image on docker hub)


mhammond (approved):
rs=me

PR #1494, opened 2023-03-17
[Discuss] New docs/examples in preparation for async callbacks

bendk:
This PR is how I think async callbacks could work and why they would be useful.  Any feedback would be greatly appreciated.

The main point is that if we have async callbacks, then foreign application that uses the UniFFIed library can drive things.  We don't need to start up new threads or run a tokio event loop.   This opens up some very interesting possibilities.  Check out the `user-store` example for details here.

I also want to suggest a couple ergonomics changes for callbacks, check out `callback_interfaces.md` for details.

### DIFF
--- a/docs/manual/src/udl/callback_interfaces.md
+++ b/docs/manual/src/udl/callback_interfaces.md
@@ -1,107 +1,55 @@
 # Callback interfaces
 
-Callback interfaces are traits specified in UDL which can be implemented by foreign languages.
+Callback interfaces are traits specified in UDL which can be implemented by
+foreign languages.  Callbacks interfaces allow the library consumer to supply
+the Rust library with additional capabilities implemented in the foreign
+language, for example:
 
-They can provide Rust code access available to the host language, but not easily replicated
-in Rust.
+ * accessing app preferences and assets
+ * network requests
+ * logging
+ * task scheduling
 
- * accessing device APIs.
- * provide glue to clip together Rust components at runtime.
- * access shared resources and assets bundled with the app.
+Both sync and async functions are supported.
+
+# Defining callback interfaces
+
+```rust,no_run
+#[uniffi::callback_interface]
+pub trait Keychain {
+  // Access a keychain exposed by the host operating system
+  fn get(&self, key: String) -> Result<Option<String>, KeyChainError>
+  fn put(&self, key: String, value: String) -> Result<(), KeyChainError>
+
+  // Async API
+  async fn get_async(&self, key: String) -> Result<Option<String>, KeyChainError>
+  async fn put_async(&self, key: String, value: String) -> Result<(), KeyChainError>
+}
+```
+
+Note: Rust doesn't support async functions in traits yet.  The proc-macro
+converts async function signatures into a sync function that returns
+`uniffi::ForeignFuture<T>`, similar to the `async-trait` crate.
+`uniffi::ForeignFuture<T>` implements `Future<Output = T>` and `Send`.
+
+## Setup error handling
+
+All methods of the Rust trait should return a Result.  This is needed because
+any callback method call can fail unexpected ways that shouldn't result in a
+panic, for example when the foreign code throws an exception we weren't
+expecting.
+
+If your method can only fail in these unexpected ways, use
+`uniffi::CallbackResult<T>` as your result type.  This is a type alias for
+`Result<T, uniffi::UnexpectedCallbackError>`.
+
+For methods that are expected to throw sometimes, use [error type defined in the UDL](./errors.md)
+and implement `From<uniffi::UnexpectedCallbackError>` for that type. See
+`example/callbacks` for an example of how to do this.
 
 # Using callback interfaces
 
-## 1. Define a Rust trait
-
-This toy example defines a way of Rust accessing a key-value store exposed
-by the host operating system (e.g. the key chain).
-
-```rust,no_run
-pub trait Keychain: Send + Sync + Debug {
-  fn get(&self, key: String) -> Result<Option<String>, KeyChainError>
-  fn put(&self, key: String, value: String) -> Result<(), KeyChainError>
-}
-```
-
-### Why Send + Sync?
-
-The concrete types that UniFFI generates for callback interfaces implement `Send`, `Sync`, and `Debug`, so it's safe to
-include these as supertraits of your callback interface trait.  This isn't strictly necessary, but it's often useful.  In
-particular, `Send + Sync` is useful when:
-  - Storing `Box<dyn CallbackInterfaceTrait>` types inside a type that needs to be `Send + Sync` (for example a UniFFI
-    interface type)
-  - Storing `Box<dyn CallbackInterfaceTrait>` inside a global `Mutex` or `RwLock`
-
-
-## 2. Setup error handling
-
-All methods of the Rust trait should return a Result.  The error half of that result must
-be an [error type defined in the UDL](./errors.md).
-
-It's currently allowed for callback interface methods to return a regular value
-rather than a `Result<>`.  However, this is means that any exception from the
-foreign bindings will lead to a panic.
-
-### Extra requirements for errors used in callback interfaces
-
-In order to support errors in callback interfaces, UniFFI must be able to
-properly [lift the error](../internals/lifting_and_lowering.md).  This means
-that the if the error is described by an `enum` rather than an `interface` in
-the UDL (see [Errors](./errors.md)) then all variants of the Rust enum must be unit variants.
-
-In addition to expected errors, a callback interface call can result in all kinds of
-unexpected errors.  Some examples are the foreign code throws an exception that's not part
-of the exception type or there was a problem marshalling the data for the call.  UniFFI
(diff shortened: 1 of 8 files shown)


docs/manual/src/udl/callback_interfaces.md
```diff
@@ -1,107 +1,55 @@
...
-```
+All methods of the Rust trait should return a Result.  This is needed because
+any callback method call can fail unexpected ways that shouldn't result in a
+panic, for example when the foreign code throws an exception we weren't
+expecting.
 
-In Rust:
+If your method can only fail in these unexpected ways, use
+`uniffi::CallbackResult<T>` as your result type.  This is a type alias for
+`Result<T, uniffi::UnexpectedCallbackError>`.
```
bendk, line 44: This is a small change that I think could improve ergonomics.  If a callback interface method isn't expected to raise an error, then we shouldn't need to define an error type for it.

docs/manual/src/udl/callback_interfaces.md
```diff
@@ -1,107 +1,55 @@
...
-
-## 1. Define a Rust trait
-
-This toy example defines a way of Rust accessing a key-value store exposed
-by the host operating system (e.g. the key chain).
+# Defining callback interfaces
 
 ```rust,no_run
-pub trait Keychain: Send + Sync + Debug {
+#[uniffi::callback_interface]
```
bendk, line 18: I think that if we want async we should definitely lean into the macros.  Rust currently doesn't allow async functions in traits, but a macro can convert those to a concrete type to work around it.  We could also support this in UDL, but then you'd need to write some slightly ugly Rust code -- async functions would need to be re-written as sync functions that returned `uniffi::ForeignFuture<T>`.

docs/manual/src/udl/callback_interfaces.md
```diff
@@ -168,3 +153,71 @@ Take care to avoid this by following guidelines:
    references through intermediate objects.
 3. If you need to break the first 2 guidelines, then take steps to manually break the cycles to avoid memory leaks.
    Alternatively, ensure that the number of objects that can ever be created is bounded below some acceptable level.
+
+## Scheduling tasks using the `TaskQueue` callback interface
```
bendk, line 157: This is orthogonal to user-defined callback interfaces, but I think it could be really useful.  I'm thinking we should implement them at more-or-less the same time.

examples/user-data-store/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    // Wrapper for a database operation that awaits on DB initialization and loading the crypto
+    // key.  f should be a synchronous function.
+    //
+    // This is the main point of this example.  This wrapper uses async callback interfaces to:
+    //   - Initialize the database and load the crypto key asynchronously
+    //   - Schedule the database operation in a background thread
+    //   - Log any errors
+    //   - Notify the application that the operation completed
```
bendk, line 98: This feels like a nice version of [decorators](https://github.com/mozilla/uniffi-rs/pull/1051) to me.

examples/user-data-store/src/lib.rs
```diff
@@ -0,0 +1,132 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! An example of a user data store that with an async interface that's driven by the executor on
+//! the foreign side of the FFI.  This is achieved by awaiting async callback interface methods.
+//! See the callbacks section of the manual for the technical details.
```
bendk, line 7: I expect a bunch of compiler errors if you tried to build this.  The code will definitely need to change, but I think the general spirit is correct.

docs/manual/src/udl/callback_interfaces.md
```diff
@@ -125,36 +81,65 @@ class SwiftKeychain: Keychain {
...
+  pub fn get_password(&self) -> String {
+    UserInfo {
+        username: self.keychain.get("username".into()),
+        password: self.keychain.get("password".into()),
+    }
+  }
 
-Also note, that storing the `Box<dyn Keychain>` in the `Authenticator` required that all implementations
-*must* implement `Send`.
+  pub async fn get_password_async(&self) -> String {
```
badboy, line 131:
```suggestion
  pub async fn get_password_async(&self) -> UserInfo {
```

docs/manual/src/udl/callback_interfaces.md
badboy: Not to say we shouldn't, but from what I read here it might be possible to define and implement this outside of UniFFI.
If so, what benefits does it have to make that part of uniffi directly?

badboy (commented):
(note: I already did some typo/other nitpicks review, feel free to ignore this in this early stage of the discussion)

I think the async interfaces overall look good and possible to support.
If we all agree that proc-macros are the future, then I agree that only supporting async trait functions there is fine.
I'm less enthusiastic about the task queue, but not outright blocking. See inline. (fwiw if that would have existed in the past Glean would have probably ended up using it)

docs/manual/src/udl/callback_interfaces.md
bendk: The issue is that it needs a couple of things not supported in UniFFI:
  - Storing/calling an arbitrary Rust closure
  - Generics (it returns the closure return type)

Both of those could be nice features to support in general and once we do, then we don't need to special case this.  However,  I don't see that happening anytime soon.

docs/manual/src/udl/callback_interfaces.md
```diff
@@ -111,9 +59,17 @@ class KotlinKeychain: Keychain {
...
-    override fun put(key: String) {
+    override fun put(key: String, value: String) {
         // … elide the implementation.
     }
+
+    override async fun getAsync(key: String) = withContext(coroutineContext) {
+        get(key)
+    }
+
+    override async fun putAsync(key: String, value: String) = withContext(coroutineContext) {
```
jhugman, line 70: Nit:
```suggestion
    override suspend fun putAsync(key: String, value: String) = withContext(coroutineContext) {
```

Thinking about it however, doesn't this dance with the `coroutineContext` and `suspend` simply disappear now? (Unless you're intent was to show the kotlin running on a different queue to the `TaskQueue` that was passed in at construction?

jhugman (approved):
This looks very promising/exciting.

Abstracting the various coroutines, dispatch queues, main loops etc into `ThreadQueue` looks a very elegant way to solve the majority of the async use cases we have.

docs/manual/src/udl/callback_interfaces.md
bendk: Good point.  I think you're right, although I'd need to test it to be sure.

docs/manual/src/udl/callback_interfaces.md
mhammond: don't mean to bikeshed too much, but I wonder if something like `TaskExecutor` is a better name here?

docs/manual/src/udl/callback_interfaces.md
```diff
@@ -168,3 +153,71 @@ Take care to avoid this by following guidelines:
...
+val userStore = UserStore(TaskQueue(Dispatchers.IO))
+```
+
+## Async callbacks are driven by foreign executors
+
+Code that awaits an async callback interface method will be woken up by an FFI call from the UniFFI
+foreign bindings code. This allows your Rust code to be driven by the foreign async executor,
+avoiding the need to spawn threads from Rust or run a `tokio` loop.
+
+See the `user-data-store` example for how this could work in action.
```
mhammond, line 212: You intend fleshing that example out with foreign bindings?

docs/manual/src/udl/callback_interfaces.md
```diff
@@ -168,3 +153,71 @@ Take care to avoid this by following guidelines:
...
+
+## Fire and forget callbacks
+
+As mentioned above, UniFFI uses the `ForeignFuture<T>` type to handle async callback interface
+returns.  `ForeignFuture` also implements the `fire()` method, which schedules the future to be run
+by the foreign executor, ignoring the result.  This is good for fire-and-forget style callbacks,
+like logging.
```
mhammond, line 219: ideally that new example would demonstrate that too, so another "See the example" note seems worthwhile

examples/callbacks/README.md
```diff
@@ -0,0 +1,3 @@
+This example is actually several mini-examples that should how callbacks can be used.  The
+emphasis on how they can be used as a port/adapter style system with the embedding application
```
mhammond, line 2: IMO you should replace "port/adaptor style" with a few more words that describes what you mean, as those terms don't appear anywhere else in the repo best I can tell and different people often have different pictures for these generic terms for patterns.

mhammond (approved):
Wow, this is magic and offers some great capabilities! I'm approving this even though I hope you don't intend landing this PR but instead will roll it up into the implementation?

bendk:
Closing this one, thanks for all the feedback.

Hopefully the next PR I open will contain these docs, with the typos fixed, plus a working implementation.

PR #1495, opened 2023-03-20
Upgrade syn to 2.0

jplatte:


badboy:
> I think having both 1.0 and 2.0 in the tree will soon be unavoidable. dtolnay already upgraded his own (big) crates like thiserror and serde_derive.
> 
> But it's not critical to have this merged soon.

yeah, we will have to figure this out over on the m-c side eventually.

jplatte:
Has this been figured out on the m-c side? Attribute parsing is a bit more convenient in syn 2.0, so it would be nice to have that new API when working on the proc-macro frontend.

badboy:
> Has this been figured out on the m-c side? Attribute parsing is a bit more convenient in syn 2.0, so it would be nice to have that new API when working on the proc-macro frontend.
> 
> edit: Judging by [this](https://hg.mozilla.org/mozilla-central/file/tip/Cargo.lock#l5088) being the only syn package in m-c's `Cargo.lock`, nothing has changed.

I'll make myself a todo to find some people to figure this out next week.

badboy:
@jplatte looks like this has some merge conflicts (but GitHub is acting up and not even letting me do the usual resolving in the UI?). Can you rebase this PR?

PR #1497, opened 2023-03-22
Callback interface speedup

bendk:
Attempt number 2 on this one.  It's pretty much the same as the first attempt, but without the feature flag which simplifies things quite a bit.

---

Previous attempt: #1481


bendk:
I had a few other ideas that I wanted to try, but didn't end up going with them.

The first one was removing the `handle` and `method`  arguments and instead packing those into the `RustBuffer`.  The hypothesis was that JNA has poor performance, so the less arguments we use the better.  However, this didn't turn out to be true.  My guess here is that primitive arguments have good performance, while structs and other compound arguments are slow.  The same was true of python.

The second speedup would be to keep around a `RustBuffer` for the purpose of argument packing and never deallocate it.  I still think this could be a nice improvement, but it would be a significant amount of work.

The last speedup would be to buffer the calls in a threadlocal buffer and flush the buffer at the end of the scaffolding function.  I think this would be fairly simple and probably would be a huge speedup.  However, there's no guarantee that the code is running inside a scaffolding function.  It could be running in a spawned thread or a tokio loop or something.  In that case we would never end up actually making the scaffolding calls.  There are ways to work around this, but again it started getting to be too complex.  Maybe this could be interesting for a future project.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@
 
 [All changes in [[UnreleasedVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.23.0...HEAD).
 
+### ⚠️ Breaking Changes ⚠️
+- Implemented a new callback-interface ABI that signifacantly improves performance on Python and Kotlin.
```
mhammond, line 18: typo: signifacantly->significantly

Also wonder if we should call out that it's not breaking in a "break your code" sense, but most certainly breaking if you happen to maintain external bindings. On that note, there doesn't seem to be much guidance or even comments which might help these people update to use this scheme. I don't really want to block you on that though, so maybe you could open an issue which asks for such guidance to be given "soon", and link to that here?

fixtures/benchmarks/src/benchmarks.udl
```diff
@@ -0,0 +1,40 @@
+namespace benchmarks {
+    // Run all benchmarks and print the results to stdout
+    void run_benchmarks(string language_name, TestCallbackInterface cb);
```
mhammond, line 3: nit: very odd indentation all over this file (which I think is true in most of our udl files, but the comments in particular leapt out at me)

uniffi_core/src/ffi/foreigncallbacks.rs
```diff
@@ -128,8 +128,8 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 /// * The `method` selector specifies the method that will be called on the object, by looking it up in a list of methods from
 ///   the IDL. The index is 1 indexed. Note that the list of methods is generated by at uniffi from the IDL and used in all
 ///   bindings: so we can rely on the method list being stable within the same run of uniffi.
-/// * `args` is a serialized buffer of arguments to the function. UniFFI will deserialize it before
-///   passing individual arguments to the user's callback.
+/// * `args_data` and `args_len` represents a serialized buffer of arguments to the function.
```
mhammond, line 131: I think it might be nice to expand these comments a little for the sake of external bindings maintainers. In the old versuin the language was fairly vague, but at least the arg as a concrete `RustBuffer`. Here it's just a pointer, so the slightly vague comments about it being a "serialized buffer" need a little more context.

mhammond (approved):
This is very nice! Please wait for @badboy though :)

CHANGELOG.md
bendk: I expanded this comment with more details.  I don't think users need to do anything other than upgrade to the new version for this one.

uniffi_core/src/ffi/foreigncallbacks.rs
```diff
@@ -128,8 +128,9 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 /// * The `method` selector specifies the method that will be called on the object, by looking it up in a list of methods from
 ///   the IDL. The index is 1 indexed. Note that the list of methods is generated by at uniffi from the IDL and used in all
 ///   bindings: so we can rely on the method list being stable within the same run of uniffi.
```
badboy, line 130: While we're here:
```suggestion
///   the IDL. The list is 1 indexed. Note that the list of methods is generated by UniFFI from the IDL and used in all
///   bindings, so we can rely on the method list being stable within the same run of UniFFI.
```

uniffi_core/src/ffi/foreigncallbacks.rs
```diff
@@ -195,9 +197,160 @@ impl ForeignCallbackInternals {
...
+    fn call_callback(
+        &self,
+        handle: u64,
+        method: u32,
+        args: RustBuffer,
+        ret_rbuf: &mut RustBuffer,
+    ) -> c_int {
         let ptr_value = self.callback_ptr.load(Ordering::SeqCst);
-        unsafe { std::mem::transmute::<usize, Option<ForeignCallback>>(ptr_value) }
+        unsafe {
```
badboy, line 208: We should have a `// SAFETY:` comment here documenting why what we do is in fact safe
(We previously got `ptr_value` from an `Option<ForeignCallback>` and stored that)

This was missing in the old code as well.

uniffi_core/src/ffi/foreigncallbacks.rs
```diff
@@ -195,9 +197,160 @@ impl ForeignCallbackInternals {
...
+    pub fn invoke_callback<R, UniFfiTag>(&self, handle: u64, method: u32, args: RustBuffer) -> R
+    where
+        R: FfiConverter<UniFfiTag>,
+    {
+        let mut ret_rbuf = RustBuffer::new();
+        let callback_result = self.call_callback(handle, method, args, &mut ret_rbuf);
+        match callback_result {
+            1 => {
+                // 1 indicates success with the return value written to the RustBuffer for
+                //   non-void calls.
```
badboy, line 232:
```suggestion
                // non-void calls.
```

badboy (approved):
This is looking good! Couple of smaller things inline, but the overall implementation seems sound.

Do you happen to have the benchmark numbers of the before and after? Would be nice to have them posted here as a reference.

fixtures/benchmarks/README.md
```diff
@@ -0,0 +1,17 @@
+This fixture runs a set of benchmark tests, using criterion to test the performance.
```
badboy, line 1: Can you include the command line to actually run a benchmark?

fixtures/benchmarks/benches/bindings/run_benchmarks.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    func testVoidReturn(a: Int32, b: Int32, data: TestData) {
+    }
+
+    func testNoArgsVoidReturn() {
+    }
+
+    func runTest(testCase: TestCase, count: UInt64) -> UInt64 {
+        let data = TestData(foo: "StringOne", bar: "StringTwo")
+        let start = clock()
```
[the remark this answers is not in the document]
bendk: The reason was that I couldn't figure out how to make the variable declaration work in swift, but I think I have it now.

uniffi_core/src/ffi/foreigncallbacks.rs
```diff
@@ -195,9 +197,160 @@ impl ForeignCallbackInternals {
...
+                let vec = ret_rbuf.destroy_into_vec();
+                let mut ret_buf = vec.as_slice();
+                R::try_read(&mut ret_buf).unwrap()
+            }
+            -2 => {
+                panic!("Callback return -2, but no Error was expected")
+            }
+            // 0 is a deprecated method to indicates success for void returns
+            0 => {
```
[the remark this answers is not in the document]
bendk: That makes a ton of sense.  In fact, since we're breaking the ABI,  I think we should change the codes to make a bit more sense (success = 0, error = 1, unexpected error = 2).

bendk:
Thanks for the review @badboy I think I address all the things you brought up with the latest commit.

> Do you happen to have the benchmark numbers of the before and after? Would be nice to have them posted here as a reference.

Yes let's do that.  Here's the results of running the benchmarks on my workstation.   The benchmarks were always fairly noisy for Python/Kotlin, I would guess a margin of error of about 5%.  But the overall effect seems clear:

```
Benchmarking callbacks/python-callbacks-basic
Benchmarking callbacks/python-callbacks-basic: Warming up for 3.0000 s
Benchmarking callbacks/python-callbacks-basic: Collecting 100 samples in estimated 10.073 s (278k iterations)
Benchmarking callbacks/python-callbacks-basic: Analyzing
callbacks/python-callbacks-basic
                        time:   [36.224 µs 36.242 µs 36.259 µs]
                        change: [-20.429% -19.597% -18.867%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high severe
Benchmarking callbacks/python-callbacks-void-return
Benchmarking callbacks/python-callbacks-void-return: Warming up for 3.0000 s
Benchmarking callbacks/python-callbacks-void-return: Collecting 100 samples in estimated 10.037 s (1.1M iterations)
Benchmarking callbacks/python-callbacks-void-return: Analyzing
callbacks/python-callbacks-void-return
                        time:   [8.8727 µs 8.8992 µs 8.9484 µs]
                        change: [-57.711% -57.572% -57.369%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low mild
  5 (5.00%) high severe
Benchmarking callbacks/python-callbacks-no-args-void-return
Benchmarking callbacks/python-callbacks-no-args-void-return: Warming up for 3.0000 s
Benchmarking callbacks/python-callbacks-no-args-void-return: Collecting 100 samples in estimated 10.010 s (4.3M iterations)
Benchmarking callbacks/python-callbacks-no-args-void-return: Analyzing
callbacks/python-callbacks-no-args-void-return
                        time:   [2.3429 µs 2.3476 µs 2.3525 µs]
                        change: [-57.244% -57.095% -56.963%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```
```
Benchmarking callbacks/kotlin-callbacks-basic
Benchmarking callbacks/kotlin-callbacks-basic: Warming up for 3.0000 s
Benchmarking callbacks/kotlin-callbacks-basic: Collecting 100 samples in estimated 10.025 s (742k iterations)
Benchmarking callbacks/kotlin-callbacks-basic: Analyzing
callbacks/kotlin-callbacks-basic
                        time:   [13.192 µs 13.430 µs 13.659 µs]
                        change: [-42.337% -39.112% -35.106%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe
Benchmarking callbacks/kotlin-callbacks-void-return
Benchmarking callbacks/kotlin-callbacks-void-return: Warming up for 3.0000 s
Benchmarking callbacks/kotlin-callbacks-void-return: Collecting 100 samples in estimated 10.005 s (2.9M iterations)
Benchmarking callbacks/kotlin-callbacks-void-return: Analyzing
callbacks/kotlin-callbacks-void-return
                        time:   [3.2238 µs 3.3074 µs 3.3915 µs]
                        change: [-79.606% -78.129% -76.674%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
Benchmarking callbacks/kotlin-callbacks-no-args-void-return
Benchmarking callbacks/kotlin-callbacks-no-args-void-return: Warming up for 3.0000 s
Benchmarking callbacks/kotlin-callbacks-no-args-void-return: Collecting 100 samples in estimated 10.004 s (3.9M iterations)
Benchmarking callbacks/kotlin-callbacks-no-args-void-return: Analyzing
callbacks/kotlin-callbacks-no-args-void-return
                        time:   [2.5148 µs 2.5967 µs 2.6783 µs]
                        change: [-81.435% -80.294% -79.059%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild
```

```
callbacks/swift-callbacks-basic
                        time:   [8.2316 µs 8.2596 µs 8.2898 µs]
                        change: [-5.7728% -5.2739% -4.8039%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild
callbacks/swift-callbacks-void-return
                        time:   [6.1453 µs 6.1978 µs 6.2637 µs]
                        change: [-4.5813% -4.1072% -3.6176%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) high mild
  5 (5.00%) high severe
callbacks/swift-callbacks-no-args-void-return
                        time:   [419.52 ns 420.48 ns 421.65 ns]
                        change: [-41.218% -40.224% -38.874%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  6 (6.00%) high mild
  6 (6.00%) high severe
```

uniffi_core/src/ffi/foreigncallbacks.rs
```diff
@@ -195,9 +200,139 @@ impl ForeignCallbackInternals {
...
+                handle,
+                method,
+                args.data_pointer(),
+                args.len() as i32,
+                ret_rbuf,
+            )
+        }
+    }
+
+    // TODO: Refactor the callback code so that we don't need to have 3 different functions here
```
badboy, line 226: 👍

badboy (approved):

Issue #200, opened 2020-08-04
Add support to document the bindings, copied from in-IDL documentation


jhugman:
Getting the foreign language devs closer to the documentation.

If the IDL parser gives us access to the comments, we should try hard to copy it over to the bindings.

eoger:
See also: https://github.com/mozilla/uniffi-rs/issues/43

thunderbiscuit:
@eoger I'm wondering if this has been solved, as I see that it's been "closed as completed" in https://github.com/mozilla/application-services/issues/4254 and https://github.com/mozilla/application-services/issues/3436.

We've been using a sort of patch system where we apply a patch to the bindings file every time we build a new one, but most often not all hunks in the patch can be applied and much manual work has to be done. It's very error prone, and I don't like the idea of us poking around in the auto-generated bindings file too much. On the other hand, if we don't, we have no API docs we can generate in Dokka, nor do we get hints with auto-completion and the like. Would love to find a better, long-term solution to this.

mhammond:
I don't think this is done, and those other issues were probably closed as "completed" in error - I think the intent was to just not track them over there.

I also suspect that this is likely to never happen - afaik we can't get the comments out of the UDL, but more importantly, we are probably going to move away from UDL files entirely in favor of #1257, so it's unlikely we will spend much more effort improving UDL support unless strictly necessary.

thunderbiscuit:
Oh that's very exciting! I see some of that code has been merged to `main` so I assume it's passed the design phase and is moving full steam ahead? From reading the related issues/PR I see it might be possible to do partial migration; we'd be happy to be early testers of that. I'll keep an eye out for those changes.

arg0d:
Do you think its possible to squeeze in documentation to `.udl` files via attributes or something like that? It doesn't have to be pretty, just good enough to work until proc-macros are ready.

mhammond:
I'm sure many things are possible, but to be frank, this isn't something mozilla is going to prioritize in the short term.

arg0d:
I'm going to try and make a contribution on this. My current plan is to introduce a new extended attribute `Attribute::Doc`, and parse it for all nodes that desire documentation, e.g. `Function`, `Enum`, `Error`, `Interface`, `Constructor`, `Method`, `Record`. WebIDL parser supports multiline strings, so its possible to write multiline attributes. `textwrap` crate can be used to normalize indentation in multiline string, and to indent by desired level in generated bindings.

### Sample 1
[...]

PR #1498, opened 2023-03-24
Support for generation of documentation comments for bindings

zduny:
[We](https://github.com/eigerco)'ve developed a solution for attaching documentation comments for generated bindings.

Doc-comment are extracted from accompanying intermediary Rust code, attached to `ComponentInterface` and then attached to bindings in templates.

See [example](https://github.com/eigerco/uniffi-rs/tree/main/examples/documentation).

Related issue: https://github.com/mozilla/uniffi-rs/issues/200


examples/documentation/tests/bindings/test_documentation.py
```diff
@@ -0,0 +1,19 @@
...
+    def test_get_name(self):
+        self.assertEqual(documentation.Person("Daniel").get_name(), "Daniel", "Should be Daniel")
+        
+class TestAdd(unittest.TestCase):
+    def test_add(self):
+        self.assertEqual(documentation.add(2, 4), 6, "Should be 6")
+
+
+if __name__ == '__main__':
+    unittest.main()
```
bendk, line 19: It seems like this is testing the functionality rather than the doc generation.

What if we made these tests work more like the trybuild tests?  We check source files that have the classes/functions/whatever, with the docstrings, but with stub code for their bodies.  Then test that the generated code has docstrings that match the expected ones.  I think that's testing this change better and also users can browse the checked in files to get a feeling of what the generated docs look like.

tarikeshaq:
I'm offline for a little bit so forwarding review to @bendk :)

uniffi_bindgen/src/bindings/python/templates/MethodDocsTemplate.py
```diff
@@ -0,0 +1,22 @@
...
+    {%- endif -%}
+
+    {%- match docs.return_description -%}
+      {% when Some with (desc) %}
+
+        Returns: {{ desc }}
+      {%- when None %}
+    {%- endmatch %}        """
+  {% when None %}
+{%- endmatch %}
```
bendk, line 22: I assume this code is good, but the amount of hoops askama is putting you through seems very sad.  One of the experiments I tried in PR #1353 was to implement a new template system that understood whitespace better.  Maybe it would be worth it to extract that code and see if it improved things.

This can be follow up work, but in the meantime I really want some way to verify the results of this template code, because I can't follow it in my head.  That one reason why I like the testing suggestion.

uniffi_docs/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+keywords = ["ffi", "bindgen", "docs"]
+
+[dependencies]
+anyhow = "1"
+syn = { version = "1.0", features = ["full"] }
+pulldown-cmark = { version = "0.9.2"}
+
+[dev-dependencies]
+indoc = "2"
+quote = "1.0.23"
```
bendk, line 18: Does this need to be a separate crate?  It seems like it could be a submodule of `uniffi_bindgen`, maybe with a feature flag to enable/disable it.

uniffi_docs/src/lib.rs
```diff
@@ -0,0 +1,518 @@
...
+pub struct Documentation {
+    pub functions: HashMap<String, Function>,
+    pub structures: HashMap<String, Structure>,
+}
+
+/// Extract doc comment from attributes.
+///
+/// Rust doc comments are silently converted (during parsing) to attributes of form:
+/// #[doc = "documentation comment content"]
+fn extract_doc_comment(attrs: &[Attribute]) -> Option<String> {
```
bendk, line 132: How do you see this interacting with the proc-macros?  I don't love parsing the source code again when we already parse it in those macros.  Could this code be called directly from inside the macros?

bendk (changes requested):
Overall it looks solid to me, the one thing I'd like is a different set of tests for this.

Very interested to hear what other UniFFI devs think about this one.

uniffi_docs/Cargo.toml
bendk: Actually, never mind about this one, a separate crate seems good since it allows both `uniffi_bindgen` and `uniffi_macros` to depend on it.

examples/documentation/tests/bindings/test_documentation.py
mhammond: I agree - as written this test doesn't seem to have any value - it's testing things with already good coverage and doesn't seem to be testing the new functionality added here.

mhammond (commented):
As usual I agree with everything @bendk said :) In particular, I really like the idea of ensuring the approach and layout is sane for procmacros, and depending on the discussion in #1493, that it's suitable for feeding docs extracted from UDL into.

examples/documentation/tests/bindings/test_documentation.py
arg0d: I have implemented a "unit" test to ensure that generated bindings contain appropriate docstrings. The test is written in a fixture. Inside the UDL file, each uniffi node (`function`, `object`, `enum`, etc..) has a different docstring attached to it. The test generates bindings from the UDL file. Then, the test does `cat|grep` to ensure that docstrings are being generated for all nodes. While this test doesn't check that docstrings are placed in the appropriate location, it does at least ensure that all nodes attempt to generate some form of their docstrings.

You can check the code here: https://github.com/NordSecurity/uniffi-rs/compare/kristupas/docstrings...NordSecurity:uniffi-rs:kristupas/docstring-templates?expand=1

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -778,6 +778,65 @@ impl ComponentInterface {
         }
         Ok(())
     }
+
+    /// Attach documentation to structs/"objects"/enums/functions.
+    ///
+    /// Documentation comments in the resulting bindings are based on this information.
+    pub fn attach_documentation(&mut self, mut documentation: uniffi_docs::Documentation) {
+        for object in self.objects.iter_mut() {
```
jplatte, line 786:
```suggestion
        for object in &mut self.objects {
```

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -778,6 +778,65 @@ impl ComponentInterface {
...
+
+                for method in object.methods.iter_mut() {
+                    if let Some(function) = methods.remove(method.name()) {
+                        method.documentation = Some(function);
+                    }
+                }
+            }
+        }
+
+        for (_, record) in self.records.iter_mut() {
```
jplatte, line 806:
```suggestion
        for (_, record) in &mut self.records {
```

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -778,6 +778,65 @@ impl ComponentInterface {
...
+
+                for variant in enum_.variants.iter_mut() {
+                    if let Some(member) = members.remove(variant.name()) {
+                        variant.documentation = Some(member);
+                    }
+                }
+            }
+        }
+
+        for function in self.functions.iter_mut() {
```
jplatte, line 834:
```suggestion
        for function in &mut self.functions {
```

uniffi_docs/src/lib.rs
jplatte: Yeah, for this reason I think your comment above about this being a separate crate is actually very much relevant. We should be doing this as part of the existing macros, not in an entirely different pass that does filesystem I/O.

jplatte (commented):
Nice PR! I think the approach isn't exactly right, but fortunately large parts of the code should be reusable anyways. See comments below.

bendk:
> However, I'm not so sure about being forced to write documentation in Rust code, while the definitions resides in UDL file. My PR makes it possible to parse docstrings from UDL file, so lets cooperate and combine our PRs! Would you be willing to switch to docstrings in UDL file, or would you rather continue to write docstrings in Rust?

It seems like these approaches could be combined if both the UDL parser and proc-macro use the same system.  I think this means agreeing on how the `docstring`/`documentation` field should work.

uniffi_docs/src/lib.rs
```diff
@@ -0,0 +1,518 @@
...
+
+use anyhow::Result;
+use pulldown_cmark::{Event, HeadingLevel::H1, Parser, Tag};
+use syn::Attribute;
+
+/// Function documentation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Function {
+    pub description: String,
+    pub arguments_descriptions: HashMap<String, String>,
```
mhammond, line 15: I wonder if a `Vec<(String, String)>` or similar might be better so the args can be rendered in the order they are seen? Ditto the other hashmaps here, eg, an enum would probably want to render docs for an enum in the order the variants are declared?

zduny: Rendered order is based on `arguments()` function of `uniffi_bindgen::interface::function::Function`:
https://github.com/eigerco/uniffi-rs/blob/3e04e2f60fe8f1af136110c2e1f3bdac800c34fe/uniffi_bindgen/src/bindings/kotlin/templates/FunctionDocsTemplate.kt#L8-L12

uniffi_docs/src/lib.rs
```diff
@@ -0,0 +1,518 @@
...
+#[derive(Debug, PartialEq, Eq)]
+pub struct Documentation {
+    pub functions: HashMap<String, Function>,
+    pub structures: HashMap<String, Structure>,
+}
+
+/// Extract doc comment from attributes.
+///
+/// Rust doc comments are silently converted (during parsing) to attributes of form:
+/// #[doc = "documentation comment content"]
```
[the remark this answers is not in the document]
zduny: [`syn`](https://crates.io/crates/syn) crate I used for parsing rust code seems to do the same - i.e. it represents doc comments that way in its AST.

examples/documentation/tests/bindings/test_documentation.py
zduny: Unit test is just a part of an example - it does not test doc generation.
IMO doc comments generation should be tested separately, outside of examples (you shouldn't be confusing `uniffi-rs` users with `uniffi-rs` internal functionality testing inside examples) - how to do it is still TBD.

zduny:
Hi, as I mentioned in related discussion - documentation comment generation feature for `uniffi-rs` was just a part of larger project we're currently working on and sadly we don't have much capacity at the moment to provide deeper focus to further improvements of our PR.

Would you be willing to provide a list of minor things that we should fix ourselves and then takeover from that point?

I guess it would be best if you could commit inside this PR, but then sadly there is this [issue](https://github.com/orgs/community/discussions/5634).

mhammond:
> sadly we don't have much capacity at the moment to provide deeper focus to further improvements of our PR.

Same here :) And there's no urgency from our POV either.

MeerKatDev:
Hi @mhammond @bendk,

we're reviving the work with UniFFI and I was wondering whether we could give this PR another chance. 

What is missing here in order to proceed with the integration? Very eager to work together!

Cheers,
Luca

arg0d:
@MeerKatDev would you be wiling to accept my changes for parsing docstrings from UDL into your changes? I think it should be pretty straight forward. I could make a PR into your fork. https://github.com/mozilla/uniffi-rs/pull/1493

MeerKatDev:
> @MeerKatDev would you be wiling to accept my changes for parsing docstrings from UDL into your changes? I think it should be pretty straight forward. I could make a PR into your fork. #1493

Hi @arg0d yes, please make open a PR to our fork if you may. I'm still not convinced by comments in UDL since as I understood the plan is to dump UDL files on the way, but if the community likes them, then please feel free to work on a PR to add them in.

MeerKatDev:
> > I'm still not convinced by comments in UDL since as I understood the plan is to dump UDL files on the way,
> 
> FWIW, Mozilla has no plans to drop UDL, and there seems a chance that some capabilities UDL has will never be available in procmacros. This means that a document generation system which excludes UDL is unlikely to be useful to Mozilla - which obviously isn't meant to imply we wouldn't take a patch which didn't support UDL, but does mean our help and involvement would probably be limited.

Thank you @mhammond hadn't understand that point very well - then @arg0d it'd be very desirable to get those features merged to our branch of course!

MeerKatDev:
Today I synced our fork with mozilla's main, I'm going to integrate the suggested changes from here on.

mhammond:
UDL docs are going to be quite difficult now. While I think it is true that a doc tool without support for UDL is unlikely to be that interesting to Mozilla, I still suggest that you ignore it in the first instance for the reasons mentioned by @arg0d. Once a core is in place we will be in a better position to assess whether we should add docs to the metadata items, or whether to create entirely new structs and perform a separate pass over the UDL purely for the purposes of generating the docs.

arg0d:
I have updated [my original UDL docstrings PR](https://github.com/mozilla/uniffi-rs/pull/1493), and also created a [PR to merge my UDL changes into this PR](https://github.com/eigerco/uniffi-rs/pull/12).

Overall I would say the changes in my PR are more straightforward. The docstrings are rendered in the most basic way - what you type is the UDL is what you get in bindings. The changes in https://github.com/mozilla/uniffi-rs/pull/1493 are covered by both unit tests and fixture tests to some degree. If Rust backend/templates miss a docstring somewhere, fixture tests should catch that.

Changes in this PR https://github.com/mozilla/uniffi-rs/pull/1498 are not enough for our use case. Docstrings are missing in some key places, i.e. callbacks, errors, object methods, namespace. I have included everything we need in https://github.com/eigerco/uniffi-rs/pull/12. If https://github.com/eigerco/uniffi-rs/pull/12 is merged, then any of the 2 PRs are fine by us.

MeerKatDev:
Today I merged again the codebase, it looks ok so far. I also applied the suggestions from @jplatte (gonna commit that one soon). I am also looking forward to merge the work by @arg0d .
There's also an example called `documentation`. 

What are we missing to merge this one?

arg0d:
This PR is not up to date with mozilla/uniffi-rs:main

MeerKatDev:
> This PR is not up to date with mozilla/uniffi-rs:main

Now it is.

MeerKatDev:
> This PR is not up to date with mozilla/uniffi-rs:main

now it is!

uniffi_docs/src/lib.rs
jplatte: I'm guessing that since this review thread has not received new comments, nothing has changed in this regard?

uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -240,9 +247,11 @@ impl AsType for Enum {
 /// Represents an individual variant in an Enum.
 ///
 /// Each variant has a name and zero or more fields.
-#[derive(Debug, Clone, Default, PartialEq, Eq, Checksum)]
+#[derive(Debug, Clone, PartialEq, Eq, Default, Checksum)]
```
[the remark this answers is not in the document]
MeerKatDev: I'm not sure when or how that happened, I will eliminate it

MeerKatDev:
@jplatte it [this PR](https://github.com/eigerco/uniffi-rs/pull/10) going to lengthen the merge/review process?

badboy:
Note that I just reviewed #1493 and personally favor that approach. UniFFI folks have a meeting tomorrow anyway, so I'll bring it up there.

MeerKatDev:
> Note that I just reviewed #1493 and personally favor that approach. UniFFI folks have a meeting tomorrow anyway, so I'll bring it up there.

@badboy please notice that I have a PR including that approach [here](https://github.com/eigerco/uniffi-rs/pull/12), so we could include both approaches with this PR.

skeet70:
> > Note that I just reviewed #1493 and personally favor that approach. UniFFI folks have a meeting tomorrow anyway, so I'll bring it up there.
> 
> @badboy please notice that there is [a PR](https://github.com/eigerco/uniffi-rs/pull/12) entering this one, which includes #1493 , so we could include both approaches with this PR.
> 
> Moreover, I don't understand favoring that approach. It forces developers to use UDL, while the [proc-macros approach](https://mozilla.github.io/uniffi-rs/proc_macro/index.html) is clearly a step towards NOT using UDL. Meaning that if I want some comments on a structure, I have to write down the UDL spec, while the Rust implementation will always be there. It also precludes from various other tools which may be developed for Rust, having quite a larger community of developers than the one of UDL. It's also more natural to write comments on Rust code (using `///`!) than on UDL.

Same question here too. It's great to have both as options in my opinion. We don't have a UDL file in our project, would prefer to keep it that way to avoid duplication, and want the Rust library to be the canonical single source of documentation. This PR accomplishes the doc goal along with the others, while the other one is only really useful for those already committed to UDL.

badboy:
> > Note that I just reviewed #1493 and personally favor that approach. UniFFI folks have a meeting tomorrow anyway, so I'll bring it up there.
> 
> @badboy please notice that there is [a PR](https://github.com/eigerco/uniffi-rs/pull/12) entering this one, which includes #1493 , so we could include both approaches with this PR.
> 
> Moreover, I don't understand favoring that approach. It forces developers to use UDL, while the [proc-macros approach](https://mozilla.github.io/uniffi-rs/proc_macro/index.html) is clearly a step towards NOT using UDL. Meaning that if I want some comments on a structure, I have to write down the UDL spec, while the Rust implementation will always be there. It also precludes from various other tools which may be developed for Rust, having quite a larger community of developers than the one of UDL. It's also more natural to write comments on Rust code (using `///`!) than on UDL.

Sorry for the miscommunication there. That's on me.
I was under the impression that it's either that PR or this one; not aware of the fact that we can get both together as well.
Especially with that other PR up it shouldn't be much more of a deal to pull both things in.
I'll give this one another look tomorrow and then we can move forward.

jplatte:
@badboy Please note that there is still a major unresolved thread here: https://github.com/mozilla/uniffi-rs/pull/1498#discussion_r1147586564

MeerKatDev:
> @badboy Please note that there is still a major unresolved thread here: [#1498 (comment)](https://github.com/mozilla/uniffi-rs/pull/1498#discussion_r1147586564)

is there a way we can help? I'm also unsure whether I fully understand the issue: 

What I got is that it is a problem that the uniffi docs is a separate crate, because you would like to parse both proc macros and documentation in one pass - but isn't it better for them to be kept separated, since they deal with totally different issues? We could integrate it, but won't it be more of a burden to maintain?

Also, I'd expect documentation parsing to be much quicker than expansion of proc macros. Is the performance really going to be an issue here? Let's also take into consideration that docs generation is easy to switch on and off.

arg0d:
As I understand there are 3 use cases when using docstrings.
1. Use UDL file, write docstrings in UDL file. This is covered by my PR.
2. Use proc-macros without any UDL file, write docstrings in Rust code. I assume this PR covers this use case?
3. Use UDL file without any proc macros, write docstrings in Rust code. I think this works with this PR.

I find 3. to be quite an interesting use case. Somebody had mentioned before that this might be a preferred method for some, since documentation in Rust code may be more up to date than in UDL files. Supporting this use case is only possible with changes implemented in this PR. Since proc macros are not used, the only way to extract documentation from Rust code is to parse the Rust code in a separate pass. I can definitely imagine how this would be useful in some projects.

From brief comments by @jplatte, supporting 2. should be trivial with changes from my PR. Some changes are required to extract doctsring from `#[uniffi::export]`, store and parse docstring to/from metadata section, but thats basically it.  If 3. is not a desired use case, than the Rust parsing logic can be safely dropped.

@MeerKatDev Which use case do you prefer?

jplatte:
I would expect doc collection to be a regular part of processing both UDL & the proc-macro input. That would also mean that mixing UDL and proc-macros is supported for docs in exactly the same ways it is for bindings generation. I have noticed that this PR diverges from that, but I don't understand the details of how it actually works other than that it processes Rust files with `syn` independently. Is there a write-up of that somewhere?

jplatte:
One possibly noteworthy feature that proc-macros support naturally is target-specific code (e.g. `#[cfg(target_os = "ios")]`). I think it would be very weird if bindings generation handled these correctly, but documentation generation didn't.

MeerKatDev:
@arg0d This PR was written with the 3. use case in mind, when still the 2. wasn't available. I - and I think many other developers - will prefer to write the documentation in Rust, so I prefer 3. (is it what you were asking?)

@jplatte that would require a much tighter integration though, and additional (mandatory) maintenance on the way, from your side. As we understood, this had to end up as a modular library, and not everyone will want to bother with documentation. I think that this reasoning would be enough to lean on the side of making separate passes.

arg0d:
I think uniffi could support all 3 use cases. The only problem I can imagine is lack of tests to cover all 3 use cases for docstrings. Assuming there are enough tests, supporting all 3 use case seems quite trivial. There isn't mutable state to worry about, its just moving strings from A to B.

I don't see a problem with conditional compilation, assuming 2nd option is fully supported. When using proc-macros, conditional compilation would work fine by using proc macros to extract docstrings. When using this 3rd option to extract docstrings from Rust code, conditional compilation wouldn't work correctly, but I don't see that as a big problem, especially if its clearly documented somewhere how the documentation is extract with this mode.

I think it would be most appropriate to do all this over the course of multiple PRs. The 1st option is covered by #1493. The 2nd is not currently covered, but it can be covered in future PR. The 3rd option is covered by this PR. I think we could merge both of these PRs sequentially.

I'm not very particular about the order in which PRs are merged. I just think merging either one of these PRs is going to get the the process moving again. The only difference between these is naming in Rust code / templates. #1493 uses `docstring`, and this PR uses `documentation`. Since conflicts need to be resolved between these PRs, the first PR to be merged will determine the naming going forward.

jplatte:
> that would require a much tighter integration though

Yes, this is a PR to the main UniFFI repo afterall, not a separate one.

> and additional (mandatory) maintenance on the way, from your side.

I'm not clear on why this would be more complicated maintenance-wise

> As we understood, this had to end up as a modular library, and not everyone will want to bother with documentation.

I don't think anybody would be forced to write documentation comments with the alternative solution? Also, it might still be worth it to generate documentation without doc comments, just so the structure of the generated bindings is clearer.

All of this is just my personal opinion as a contributor / collaborator though, I'm not a maintainer; this needs input from somebody from the actual team maintaining this.

uniffi_bindgen/src/interface/enum_.rs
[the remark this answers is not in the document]
MeerKatDev: reverted!

MeerKatDev:
@badboy Hi, I saw you guys added support for the UDL docstrings, is this PR being considered for merging too?

badboy:
> @badboy Hi, I saw you guys added support for the UDL docstrings, is this PR being considered for merging too?

We're discussing it tomorrow, at which point I'll get back to you.

mhammond:
> is this PR being considered for merging too?

I suspect #1493 being merged will have a significant impact on this PR, right? Eg, `EnumTemplate.kt` now does `{%- call kt::docstring(variant, 4) %}` as opposed to your `{% include "EnumVariantDocsTemplate.kt" %}`.

On one hand I don't want to ask you to do work which might not get accepted, but OTOH this patch now looks much bigger than it probably needs to be, which makes it difficult to see the bigger picture

badboy:
> > is this PR being considered for merging too?
> 
> I suspect #1493 being merged will have a significant impact on this PR, right? Eg, `EnumTemplate.kt` now does `{%- call kt::docstring(variant, 4) %}` as opposed to your `{% include "EnumVariantDocsTemplate.kt" %}`.
> 
> On one hand I don't want to ask you to do work which might not get accepted, but OTOH this patch now looks much bigger than it probably needs to be, which makes it difficult to see the bigger picture

There's a PR on the other repo that does merge in the code from 1493, thus making this a bit smaller overall. It would still need a rebase. (I also had some other comments on the code that I did not yet submit)

badboy:
Some of us maintainers discussed this last week.
As pointed out #1493 gets us the "docs in UDL" part (use case 1).
In the meantime #1862 was posted and gets us "docs in Rust code in proc-macro mode" (and I also have been working on that solution) and that solves use case 2.

We do not think that having a third way, to write docs in Rust but have all definitions in UDL, is beneficial.
Jonas already pointed out some things, like the re-parsing of Rust code and the unclarity of how target-specific code docs are handled; and I agree with those points.
I'm also not sure how we would handle having docs in both UDL and Rust (accidentally or intentionally).

Therefore I will close this PR.
I'm sorry it took us so long to decide, without earlier feedback. We appreciate your work on this!

Side note: There's a couple of details in this PR that would need changes anyway, e.g. the markdown parsing is not something we would do as-is. As implemented here it will break on real-world docs and relies on a documentation convention that is not adhered to by all Rust code (nor properly documented). We could discuss how to tackle that issue separately though.

PR #1501, opened 2023-03-30
Fix swift custom type

nesium:
Hi there. I've noticed that when declaring a custom type with a custom conversion, where the type is based on a built-in type which isn't [lowered into a RustBuffer](https://mozilla.github.io/uniffi-rs/internals/lifting_and_lowering.html#lowered-types) (in my case a `i64`) the generated Swift code doesn't compile. This PR is my attempt to fix that.

Furthermore when writing the included test I noticed typos in the Swift floating point types. Since that seemed very much related I've included fixes for that issue as well.


badboy (approved):

badboy:
Not sure why CircleCI complains here. You might need to login, or I can re-push this PR to run it.
I did run the Swift tests for the changed example locally and it worked.

nesium:
Fixed. Thanks for the quick review @badboy!

PR #1509, opened 2023-04-05
Post-#1469 cleanup

jplatte:
When trying to upgrade to the latest git version of UniFFI, I found that there was some confusion around `r#`-prefixed vs. unprefixed type names that resulted in "has no definition" errors. This fixes that, and also cleans up some other minor issues I found.


bendk (approved):
Nice cleanup, thanks for fixing this.

PR #1510, opened 2023-04-06
Fully qualify uses of `Result` in scaffolding/macro code

ijc:
PR #1469 introduced an incompatibility with a local `type Result<T> = std::result::Result<T, MyError>` type alias. As can be seen with the included test without the fix applied:

```
error[E0107]: this type alias takes 1 generic argument but 2 generic arguments were supplied
  --> /mounted_workdir/target/debug/build/unary-result-alias-f408c03f1955dadd/out/unary-result-alias.uniffi.rs:79:1
   |
79 | / #[::uniffi::ffi_converter_error(
80 | |     tag = crate::UniFfiTag,
81 | |     flat_error,
82 | |
83 | | )]
   | |  ^
   | |  |
   | |__expected 1 generic argument
   |    help: remove this generic argument
   |
note: type alias defined here, with 1 generic parameter: `T`
  --> fixtures/regressions/unary-result-alias/src/lib.rs:7:10
   |
7  | pub type Result<T> = std::result::Result<T, MyError>;
   |          ^^^^^^ -
   = note: this error originates in the macro `::uniffi::ffi_converter_default_return` which comes from the expansion of the attribute macro `::uniffi::ffi_converter_error` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Fixing the macros in `uniffi_core` to use `::std::result::Result` fixes that.

This exposes a further case in the generated code:

```
error[E0107]: this type alias takes 1 generic argument but 2 generic arguments were supplied
   --> /mounted_workdir/target/debug/build/unary-result-alias-f408c03f1955dadd/out/unary-result-alias.uniffi.rs:108:40
    |
108 |     uniffi::rust_call(call_status, || <Result<(), r#MyError> as ::uniffi::FfiConverter<crate::UniFfiTag>>::lower_return(
    |                                        ^^^^^^     --------- help: remove this generic argument
    |                                        |
    |                                        expected 1 generic argument
    |
note: type alias defined here, with 1 generic parameter: `T`
   --> fixtures/regressions/unary-result-alias/src/lib.rs:7:10
    |
7   | pub type Result<T> = std::result::Result<T, MyError>;
    |          ^^^^^^ -
```

So the same change is required to the output produced by bindgen.

### DIFF
--- a/uniffi_bindgen/src/scaffolding/mod.rs
+++ b/uniffi_bindgen/src/scaffolding/mod.rs
@@ -108,7 +108,7 @@ mod filters {
         };
         Ok(match callable.throws_type() {
             Some(e) => format!(
-                "<Result<{}, {}> as ::uniffi::FfiConverter<crate::UniFfiTag>>",
+                "<::std::result::Result<{}, {}> as ::uniffi::FfiConverter<crate::UniFfiTag>>",
                 result_type,
                 type_rs(&e)?
             ),
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ members = [
   "fixtures/regressions/missing-newline",
   "fixtures/regressions/swift-callbacks-omit-labels",
   "fixtures/regressions/swift-dictionary-nesting",
+  "fixtures/regressions/unary-result-alias",
   "fixtures/uitests",
   "fixtures/uniffi-fixture-time",
   "fixtures/version-mismatch",
--- a/fixtures/regressions/unary-result-alias/Cargo.toml
+++ b/fixtures/regressions/unary-result-alias/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "unary-result-alias"
+version = "0.1.0"
+edition = "2021"
+license = "MPL-2.0"
+publish = false
+
+[lib]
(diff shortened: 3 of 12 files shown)


ijc:
@bendk FYI.

uniffi_bindgen/src/scaffolding/mod.rs
```diff
@@ -108,7 +108,7 @@ mod filters {
         };
         Ok(match callable.throws_type() {
             Some(e) => format!(
-                "<Result<{}, {}> as ::uniffi::FfiConverter<crate::UniFfiTag>>",
+                "<std::result::Result<{}, {}> as ::uniffi::FfiConverter<crate::UniFfiTag>>",
```
bendk, line 111: How about using `::std` here as well?

bendk (changes requested):
Looks great to me, thanks for the tests as well.  I think there's one more `::` to add and everything else looks good.

uniffi_bindgen/src/scaffolding/mod.rs
ijc: Agreed. Done.

bendk (approved):

bendk:
Thanks for the submission!

PR #1512, opened 2023-04-06
Ensure bridging header function prototypes are valid

ijc:
Compiling my swift project I am seeing:

```
Generated/mycrateFFI.h:75:45: error: this function declaration is not a prototype [-Werror,-Wstrict-prototypes]
```

for the `ffi_mycrate_uniffi_contract_version` and
`uniffi_mycrate_checksum_func_myfun` declarations.

This is because they take no arguments which is emitted as `()`, for C `(void)` is required, C++ is happy with either.

Add a specific case to `swift::arg_list_ffi_decl` to handle this. The resulting diff is:

```
--- mycrateFFI.h	2023-04-06 10:38:25.645130502 +0100
+++ mycrateFFI.h	2023-04-06 11:01:12.415958817 +0100
@@ -73,9 +73,11 @@ RustBuffer ffi_mycrate_rustbuffer_reserv
 );

 uint32_t ffi_mycrate_uniffi_contract_version(
+    void

 );

 uint16_t uniffi_mycrate_checksum_func_myfun(
+    void

 );
```

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ members = [
   "fixtures/simple-iface",
   "fixtures/swift-omit-labels",
   "fixtures/futures",
+  "fixtures/swift-bridging-header-compile",
 ]
 
 resolver = "2"
--- a/fixtures/swift-bridging-header-compile/Cargo.toml
+++ b/fixtures/swift-bridging-header-compile/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "swift-bridging-header-compile"
+version = "0.1.0"
+edition = "2021"
+license = "MPL-2.0"
+publish = false
+
+[lib]
+name = "uniffi_swift_bridging_header_compiler"
+crate-type = ["lib", "cdylib"]
+
+[dependencies]
+uniffi = { path = "../../uniffi" }
+
+[build-dependencies]
+uniffi = {path = "../../uniffi", features = ["build"] }
+
+[dev-dependencies]
+uniffi = {path = "../../uniffi", features = ["bindgen-tests", "cli"] }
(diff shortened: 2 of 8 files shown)


ijc:
I think these functions were newly added in #1469 (/cc @bendk) and uniffi didn't have any `FfiFunction`'s which had an empty argument list before that point which is why this went unnoticed until now.

ijc:
> Compiling my swift project I am seeing:

Actually, that's inaccurate: it's a hybrid swift + objc project and i think the prototype error was coming from building a `.m` file which was including the bridging header and therefore the FFI.h.

mhammond:
Any chance you can demonstrate the problem in one of our [fixtures](https://github.com/mozilla/uniffi-rs/tree/main/fixtures)?

ijc:
Yes. I'm trying to figure that out now, there don't seem to be any existing fixtures which demonstrate running a C/obj-c compiler over the FFI.h.

ijc:
> Any chance you can demonstrate the problem in one of our [fixtures](https://github.com/mozilla/uniffi-rs/tree/main/fixtures)?

Done.

bendk (approved):
Wow, C is so weird.  This seems correct to me based on https://en.wikipedia.org/wiki/Compatibility_of_C_and_C%2B%2B.

badboy:
> Wow, C is so weird. This seems correct to me based on [en.wikipedia.org/wiki/Compatibility_of_C_and_C%2B%2B](https://en.wikipedia.org/wiki/Compatibility_of_C_and_C%2B%2B).

Yes. I can confirm `void` as a signal for "no argument" is correct and empty arguments technically allows any number of arguments. Bit annoying that we need the additional conditional in the template, but it is what it is.

Issue #1507, opened 2023-04-03
Foreign async code can destroy the future waker while there's still a callback pending


bendk:
Rust can call the waker callback after the future is completed, which our foreign bindings code doesn't handle well.

The case I'm thinking of goes something like this:
- A Rust function creates 2 futures, let's call them `child-a` and `child-b`
- Rust races them with [select_all()](https://docs.rs/futures/latest/futures/future/fn.select_all.html), creating a new future, let's call it `parent`
- The foreign bindings awaits `parent`
- `child-a` completes and calls the foreign waker callback
- The foreign waker polls `parent`finds that it's ready
- The foreign code destroys the waker and returns the result
- `child-b` completes and calls the waker callback -- except this is now a dangling pointer so this will probably crash the system.

https://github.com/bendk/uniffi-rs/commit/28a907b1bd3ea0e7c28860a8fb3a562dbe06cb0e has some tests that show this in action.

bendk:
I believe #1515 fixes this one.

PR #1515, opened 2023-04-17
Async refactor

bendk:
Refactored the async code:
  - Rust is now responsible for polling/managing the Future
  - There's a generic API for scheduling Rust tasks using a foreign executor, this is what Rust uses to schedule its polls.

There's a few reasons for this:
 - #1507 me think that that managing ownership of the wakers over the FFI is going to be a nightmare.  I think it means we would need to make an FFI call for each `clone()` and `drop()` call on the waker.  It seems much simpler to me to keep the Future/Wakers in Rust.
 - I want the foreign executor code in general to add to our async story.  In particular, I want to allow the Rust code to schedule blocking calls on the foreign executor, turning sync APIs into async ones and avoiding the need for adding wrapper code to do that (e.g. [SyncableLoginStorage.kt](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/service/sync-logins/src/main/java/mozilla/components/service/sync/logins/SyncableLoginsStorage.kt#L98))
 - I want to add async callback interface support and I think the same general process and the `FutureCallback` type can be used there.
 - I think this code is simpler overall.


bendk:
@Hywan I can't officially add you as a reviewer, but I'd love to know what you think of this.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -3,19 +3,21 @@
 // Error runtime.
 @Structure.FieldOrder("code", "error_buf")
 internal open class RustCallStatus : Structure() {
-    @JvmField var code: Int = 0
+    @JvmField var code: Byte = 0
```
bendk, line 6: I think this has been the wrong type since it was introduced :exploding_head:

mhammond (approved):
There's nothing not to like here 👍

Hywan:
I was on holidays. Will check that Monday :-).

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -175,112 +107,6 @@
...
-//! must “cut” the flow, so that Rust code can continue to run as expected, and
-//! after that, the _poll function_ must be called.
-//!
-//! Put differently, the waker function must call the _poll function_ _as
-//! soon as possible_, not _immediately_. It actually makes sense: The waker
-//! must signal the executor to schedule a poll for a specific `Future` when
-//! possible; it's not an inline operation. The implementation of the waker
-//! must be very careful of that.
-//!
-//! With a diagram (because this comment would look so much clever with a diagram),
```
Hywan, line 244: That's the sadest part of this PR: the removal of this diagram.

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -289,318 +115,429 @@
...
+            callback,
+            callback_data,
+        })
+    }
 
-        assert_eq!(mem::size_of::<RustFuture::<bool>>(), rust_future_size);
-        assert_eq!(mem::size_of::<RustFuture::<u64>>(), rust_future_size);
+    /// Wake up soon and poll our future.
+    ///
+    /// This method ensures that a call to `do_wake()` is scheduled.  One one call will be scheduled
```
Hywan, line 204:
```suggestion
    /// This method ensures that a call to `do_wake()` is scheduled.  Only one call will be scheduled
```

Hywan (approved):
Overall it looks way nicer than my approach. It's a really good improvement. I see that the “waker re-entry” problem is solved on the Rust side instead of the foreign side, and I don't know what to think what this yet. If all the tests I've written are passing, it should be fine. Let's try it out!

fixtures/foreign-executor/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+                });
+            });
+        });
+    }
+
+    fn get_last_result(&self) -> Option<TestResult> {
+        self.last_result.lock().unwrap().take()
+    }
+}
+#[derive(uniffi::Record)]
```
jplatte, line 64:
```suggestion
}

#[derive(uniffi::Record)]
```

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -319,6 +321,12 @@ impl CodeOracle for KotlinCodeOracle {
             },
             FfiType::ForeignBytes => "ForeignBytes.ByValue".to_string(),
             FfiType::ForeignCallback => "ForeignCallback".to_string(),
+            FfiType::ForeignExecutorHandle => "USize".to_string(),
+            FfiType::ForeignExecutorCallback => "UniFfiForeignExecutorCallback".to_string(),
+            FfiType::FutureCallback { return_type } => {
+                format!("UniFfiFutureCallback{}", return_type.canonical_name(),)
```
jplatte, line 327:
```suggestion
                format!("UniFfiFutureCallback{}", return_type.canonical_name())
```

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -405,6 +416,22 @@ pub mod filters {
...
+            None => "void".into(),
+        };
+        let throws_string = match &result_type.throws_type {
+            Some(t) => t.canonical_name().to_snake_case(),
+            None => "void".into(),
+        };
+        Ok(format!(
+            "uniffi_async_callback_{}__{}",
+            return_string, throws_string
+        ))
```
jplatte, line 432:
```suggestion
        Ok(format!("uniffi_async_callback_{return_string}__{throws_string}"))
```

uniffi_bindgen/src/bindings/kotlin/templates/ForeignExecutorTemplate.kt
```diff
@@ -0,0 +1,60 @@
...
+        if (rustTask == null) {
+            FfiConverterForeignExecutor.drop(handle)
+        } else {
+            val coroutineScope = FfiConverterForeignExecutor.lift(handle)
+            coroutineScope.launch {
+                if (delayMs > 0) {
+                    delay(delayMs.toLong())
+                }
+                rustTask.invoke(rustTaskData)
+            }
```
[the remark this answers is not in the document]
bendk: I think this should be handled on the [Rust side](https://github.com/mozilla/uniffi-rs/blob/5290b35a4f68eae207ec3beedda83a2d3fac247b/uniffi_core/src/ffi/rustfuture.rs#L202-L287).  That code ensures that only one callback is scheduled at once, even if `wake()` is called multiple times from different threads.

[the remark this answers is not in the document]
Hywan: Yeah, I've seen that it was caught on the Rust side. Simpler. Good job.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
bendk: It doesn't matter much, but we define it as a `i8` in the `rustcalls.rs` file.  I think the main reason for `i8` was Kotlin's historical issues with unsigned types.  I don't think that's much of an issue anymore, at some point we could consider changing a lot of the FFI types to be unsigned.

uniffi_bindgen/src/bindings/python/templates/RustBufferTemplate.py
```diff
@@ -137,6 +137,8 @@ def readFloat(self):
     def readDouble(self):
         return self._unpack_from(8, ">d")
 
+    def readCSizeT(self):
+        return self._unpack_from(ctypes.sizeof(ctypes.c_size_t) , "@N")
```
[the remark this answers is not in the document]
bendk: It think that should be fixed, I'll try to follow up after we merge this.

fixtures/foreign-executor/tests/test_generated_bindings.rs
```diff
@@ -1,5 +1,8 @@
 uniffi::build_foreign_language_testcases!(
     "tests/bindings/test_foreign_executor.py",
-    "tests/bindings/test_foreign_executor.kts",
-    "tests/bindings/test_foreign_executor.swift",
+    // Disable due to the Docker image being outdated for now.
+    // Please see https://github.com/mozilla/uniffi-rs/pull/1409#issuecomment-1437170423
+    //
+    // "tests/bindings/test_foreign_executor.kts",
+    // "tests/bindings/test_foreign_executor.swift",
```
Hywan, line 7: This is really annoying :-(

badboy: If I don't get the mozilla/uniffi-ci image enabled by the end of the week, we can move over to another user-owned container for the time being.

Hywan: :+1:

Hywan: \o/

jplatte:
Trying this out, we are getting the following two errors a bunch of times:

```
/Users/runner/work/matrix-rust-sdk/matrix-rust-sdk/bindings/apple/generated/swift/matrix_sdk_ffi.swift:2734:21: error: cannot find 'FfiConverterForeignExecutor' in scope
                    FfiConverterForeignExecutor.lower(UniFfiForeignExecutor()),
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/runner/work/matrix-rust-sdk/matrix-rust-sdk/bindings/apple/generated/swift/matrix_sdk_ffi.swift:2734:55: error: cannot find 'UniFfiForeignExecutor' in scope
                    FfiConverterForeignExecutor.lower(UniFfiForeignExecutor()),
                                                      ^~~~~~~~~~~~~~~~~~~~~
```

(from [CI logs](https://github.com/matrix-org/matrix-rust-sdk/actions/runs/4957419625/jobs/8869190388))

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -289,318 +115,429 @@
...
+///
+/// RustFuture is always stored inside a Pin<Arc<>>.  The `Arc<>` allows it to be shared between
+/// wakers and Pin<> signals that it must not move, since this would break any self-references in
+/// the future.
+pub struct RustFuture<F, T, UT>
+where
+    // The future needs to be `Send`, since it will move to whatever thread the foreign executor
+    // chooses.  However, it doesn't need to be `Sync', since we don't share references between
+    // threads (see do_wake()).
+    F: Future<Output = T> + Send,
```
ijc, line 151: I've just bumped to the version of main with this PR merged and now I'm seeing a lot of "future returned by `FOO` is not `Send`" on my `[uniffi::export]`s. e.g. from things such as uses of `itertools::IntoChunks` within the function (and an async call using a chunk) and similar things.

I've not worked through all the cases we are seeing but some of them are looking like they will tricky to resolve, especially in a code base which also targets wasm (which is inherently single threaded so various types, such as `JsValue` and event `JsFuture` aren't `Send` etc). and it looks like our use of the Actix test stuff is going to run into trouble too since e.g. `actix_web::dev::Service` isn't `Send`.

I'm unsure if this is actually a new requirement or if it's always been this way under the hood but wasn't made visible to Rust until now, so the compiler wasn't catching it. Can we see anyway that the requirement can be relaxed? I assume it's tricky (or even impossible) due to the nature of the foreign executor?

bendk:
Sorry for the fallout from this one.

@jplatte I believe https://github.com/mozilla/uniffi-rs/pull/1537 should fix the build issue you're seeing, although I wasn't able to directly test.

uniffi_core/src/ffi/rustfuture.rs
bendk: This is a tricky one.  When I wrote the code I was under the impression that it was always a requirement and we should have had the bound all along.  I think that's true in the general case, but I also think you're correct that with certain setups it won't be needed.

I made https://github.com/mozilla/uniffi-rs/issues/1538 to discuss this more.

Hywan:
\o/

PR #1516, opened 2023-04-23
Library mode bindings generation

bendk:
Implemented a new mode for the `generate` command.   The new mode is enabled by passing `--library [path-to-library-file]` instead of a UDL file.  UniFFI will then use the exported metadata symbols in the library + cargo metadata to figure out what to generate.  Bindings will be generated for all UniFFIed crate that went into the library and output them to a single dir.

I think this should improve use-cases like the app-services megazord, where multiple UniFFI crates, including support crates, are compiled into one large library.  It makes it much simpler to add new crates or to move types into a support crate and use the external types feature:

  - No need to add a new `uniffi-bindgen generate` task to the build process
  - No need to add a `uniffi.toml` config just to spcify the dylib name.
  - No need to figure out the `external_packages` map for external packages.


bendk:
@badboy: I'm hoping to switch app-services over to using this feature for bindings generation.  Can you foreesee any Glean-specific issues?

fixtures/ext-types/lib/src/ext-types-lib.udl
```diff
@@ -21,26 +21,26 @@ namespace imported_types_lib {
...
 // And re-use the `custom-types` example - this exposes `Url` and `Handle`
-[External="custom-types"]
+[External="custom_types"]
 typedef extern Url;
 
-[External="custom-types"]
+[External="custom_types"]
 typedef extern Handle;
 
 // And a new type here to tie them all together.
```
bendk, line 46: Maybe I could add call this out in the changelog, but it seems like the example was just wrong from the start.  I don't think `custom-types` is a legal crate name.  It is a legal cargo package name, but the docs say "crate name" and in the custom types example it's not actually the package name, it's just the crate name with `_` replaced with `-`.

When specifying package maps in the config file, any name works as long as you're consistent, which is why the test passed before.  But I think we should still say you should use the crate name and anything else leads to UB.

uniffi_bindgen/src/crate_mode.rs
```diff
@@ -0,0 +1,244 @@
...
+// A single source that we generate bindings for
+#[derive(Debug)]
+pub struct Source {
+    pub package: Package,
+    pub crate_name: String,
+    pub ci: ComponentInterface,
+    pub config: Config,
+}
+
+fn calc_sources(
```
badboy, line 102: Not too happy with the name, as to me it's not really clear what we "calculate".
But I couldn't come up with a better name either.
`load_sources`? `parse_sources`? None really read much better, so maybe `calc_sources` is as good as any other.

uniffi_bindgen/src/lib.rs
```diff
@@ -97,84 +97,68 @@ const BINDGEN_VERSION: &str = env!("CARGO_PKG_VERSION");
...
-
-    /// Get default config values from the `ComponentInterface`
+/// BindingsConfigs are initially loaded from `uniffi.toml` file.  Then the trait methods are used
+/// to fill in missing values.
+pub trait BindingsConfig: DeserializeOwned {
+    /// key in the `bindings` table from `uniffi.toml` for this configuration
+    const TOML_KEY: &'static str;
+
+    /// Update missing values using the `ComponentInterface`
+    #[allow(unused)]
```
badboy, line 124: Why are those needed?

uniffi_bindgen/src/lib.rs
```diff
@@ -199,28 +183,27 @@ fn load_bindings_config_toml<BC: BindingGeneratorConfig>(
...
-    type Config: BindingGeneratorConfig;
+    /// Handles configuring the bindings
+    type Config: BindingsConfig;
 
     /// Writes the bindings to the output directory
     ///
     /// # Arguments
     /// - `ci`: A [`ComponentInterface`] representing the interface
-    /// - `config`: A instance of the BindingGeneratorConfig associated with this type
+    /// - `config`: A instance of the [`BindingsConfig`] associated with this type
```
badboy, line 206:
```suggestion
    /// - `config`: An instance of the [`BindingsConfig`] associated with this type
```

uniffi/Cargo.toml
```diff
@@ -13,6 +13,11 @@ license = "MPL-2.0"
 edition = "2021"
 keywords = ["ffi", "bindgen"]
 
+[[bin]]
+name = "uniffi"
+path = "src/main.rs"
+required-features = ["cli"]
```
badboy, line 19: This means we're shipping a binary again, which we only recently removed.
Is that required? Shouldn't it still be the consumer building the binary to run to avoid versionining issues?

uniffi_bindgen/src/lib.rs
```diff
@@ -97,84 +97,68 @@ const BINDGEN_VERSION: &str = env!("CARGO_PKG_VERSION");
 use anyhow::{anyhow, bail, Context, Result};
 use camino::{Utf8Path, Utf8PathBuf};
 use fs_err::{self as fs, File};
-use serde::{Deserialize, Serialize};
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
```
badboy, line 100: (can't comment on unchanged lines)

Line 75 above still uses the `include!(concat!(...))` part in the docs.
This should also be changed to the `uniffi::include_scaffolding!` usage.

badboy (changes requested):
> @badboy: I'm hoping to switch app-services over to using this feature for bindings generation. Can you foreesee any Glean-specific issues?

I don't expect any issues, as long as the old way still works.
Right now it does, but the crate mode does fail for Glean right now, due to our weird naming I think. I would like to trace that down quickly tomorrow, at least to know if this inherently breaks or can be fixed in a simple followup.

Lots of smaller things called out inline already.
Otherwise this looks quite solid already.

docs/manual/src/tutorial/foreign_language_bindings.md
```diff
@@ -38,24 +38,50 @@ creating a binary for each crate that uses UniFFI.  You can avoid this by creati
...
+From the `example/arithmetic` directory, run:
+```
+cargo run --bin uniffi-bindgen generate --crate uniffi-example-arithmetic --language kotlin --out-dir out
+```
+
+The check out the `out` directory.
+
+When using crate mode:
+  - If the any dependent crates that use UniFFI, will also have bindings generated for them.
+  - The `cdylib_name` and `kotlin.external_packages` fields do not need to be specified
```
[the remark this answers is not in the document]
bendk: The config file values override the calculated values.  I think if someone manually sets those values we should trust them.

uniffi/Cargo.toml
bendk: Yeah, that's a good point.  It was useful for testing, but I'll remove that commit from this PR.

uniffi_bindgen/src/bindings/mod.rs
```diff
@@ -133,3 +122,10 @@ pub fn write_bindings(
     }
     Ok(())
 }
+
+pub fn write_file(path: &Utf8Path, contents: String) -> Result<()> {
+    println!("{path}");
```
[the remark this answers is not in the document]
bendk: This `println!()` was actually intended.  It makes it easier to debug issues if you can see what's being generated.  I can split that out into a separate PR if you have concerns.  Otherwise I can add a comment to make it more clear.

uniffi_bindgen/src/lib.rs
```diff
@@ -264,16 +247,11 @@ pub fn generate_external_bindings(
 // such as the `extern "C"` function definitions and record data types.
 pub fn generate_component_scaffolding(
     udl_file: &Utf8Path,
-    config_file_override: Option<&Utf8Path>,
+    _config_file_override: Option<&Utf8Path>,
```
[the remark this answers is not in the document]
bendk: My thinking was that it would be useful if we ever did want to load the config file, for example if we added support for a `scaffolding` section.  But that feature has never been used and it's not clear how it would work with proc-macros anyways.  Let's remove it all.

uniffi_bindgen/src/lib.rs
```diff
@@ -97,84 +97,68 @@ const BINDGEN_VERSION: &str = env!("CARGO_PKG_VERSION");
...
-    /// Get default config values from the `ComponentInterface`
+/// BindingsConfigs are initially loaded from `uniffi.toml` file.  Then the trait methods are used
+/// to fill in missing values.
+pub trait BindingsConfig: DeserializeOwned {
+    /// key in the `bindings` table from `uniffi.toml` for this configuration
+    const TOML_KEY: &'static str;
+
+    /// Update missing values using the `ComponentInterface`
+    #[allow(unused)]
+    fn update_from_ci(&mut self, ci: &ComponentInterface) {}
```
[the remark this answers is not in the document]
bendk: Sounds good to me, I removed them.

bendk:
Pushed a couple commits that address the comments.  I removed all the printlns.  I thought it would be nice to be more verbose and I didn't want to bring in a logging implementation.  Maybe I'll make a separate PR to explore that, but it's not very important either way IMO.

docs/manual/src/tutorial/foreign_language_bindings.md
```diff
@@ -38,24 +38,48 @@ creating a binary for each crate that uses UniFFI.  You can avoid this by creati
...
 
 ### Swift
 
 Run
 ```
-cargo run --bin uniffi-bindgen generate src/math.udl --language swift
+cargo run --bin uniffi-bindgen generate src/arithmetic.udl --language swift
 ```
-then check out `src/math.swift`
+then check out `src/arithmetic.swift`
```
mhammond, line 59: I noticed the path here and above at line 51 were different, so decided to actually try this command out - and it failed: `error: no bin target named uniffi-bindgen.` - which I guess is somewhat expected given the discussion above re uniffi-bindgen, but it still seems unfortunate - I wonder if we should get an issue on file for this? It seems a reasonable expectation that it's clear to new users exactly how to run example commands in our manual?

docs/manual/src/tutorial/foreign_language_bindings.md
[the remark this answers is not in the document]
mhammond: Maybe leave this sentence out entirely? This document doesn't talk about uniffi.toml fields at all. Indeed, the manual itself doesn't - `cdylib_name` appears exactly once - manual/src/swift/configuration.md 🤷 Maybe yet another followup is to better explain uniffi.toml and which one is used when?

docs/manual/src/tutorial/foreign_language_bindings.md
```diff
@@ -38,24 +38,48 @@ creating a binary for each crate that uses UniFFI.  You can avoid this by creati
...
+Then check out the `out` directory.
+
+When using crate mode:
+  - If any dependent crates use UniFFI, they will also have bindings generated for them.
+  - The `cdylib_name` and `kotlin.external_packages` fields do not need to be specified
+    `uniffi.toml`.
+
+Crate mode adds some extra requirements:
+  - It must be run from within the cargo workspace of your project
+  - There must be exactly 1 UDL file used when compiling the Rust library.  However, crates can have
```
mhammond, line 82: s/There must be/Each crate must have/ ? The entire point of this mode is to better support multiple UDL files, so this confused me.

docs/manual/src/tutorial/foreign_language_bindings.md
```diff
@@ -38,24 +38,48 @@ creating a binary for each crate that uses UniFFI.  You can avoid this by creati
...
+    multiple UDL files as long as they ensure only one is used for any particular build,
+    e.g. by using feature flags.
+  - Rust sources must use `uniffi::include_scaffolding!` to include the scaffolding code.
```
mhammond, line 85: What else might they do? I guess I'm wondering is this line also should just die?

docs/manual/src/udl/ext_types_external.md
```diff
@@ -13,12 +13,12 @@ dictionary DemoDict {
 };
 ```
 
-And further, assume that you have another crate called `consuming-crate` which would like to use
-this dictionary. Inside `consuming-crate`'s UDL file you can reference `DemoDict` by using a
+And further, assume that you have another crate called `consuming_crate` which would like to use
```
mhammond, line 16: huh - I actually liked the `-` over the `_`, because Rust does magic with them. ie, if the crate *really* was `consuming-crate`, the distinction between `-` and `_` below becomes important.

docs/manual/src/udl/ext_types_external.md
```diff
@@ -66,10 +66,10 @@ which varies slightly for each language:
 
 ### Kotlin
 
-For Kotlin, the generated code needs to import the external types from the
-Kotlin module that corresponds to the Rust crate.  By default, UniFFI assumes
-that the Kotlin module name matches the Rust crate name, but this can be
-configured in `uniffi.toml` with an entry like this:
+For Kotlin, "crate mode" generation with `generate --crate [crate-name]` is recommended when using external types.
```
mhammond, line 69: this makes me think this PR doesn't go far enough :)

mhammond (commented):
Sorry, but I didn't get very far at all here, but the note "is recommended when using external types." made me wonder if we should go all-in on this - ie, make it the default? When would we recommend to not use it? Should we make the "old" way deprecated?

(Sorry if this is a dumb question - I didn't get far enough through today to actually answer some of this for myself, but it does seem that at a high-level, there's no need for a radical distinction between these modes)

badboy:
> > @badboy: I'm hoping to switch app-services over to using this feature for bindings generation. Can you foreesee any Glean-specific issues?
> 
> I don't expect any issues, as long as the old way still works. Right now it does, but the crate mode does fail for Glean right now, due to our weird naming I think. I would like to trace that down quickly tomorrow, at least to know if this inherently breaks or can be fixed in a simple followup.

I looked into this a bit further.
Over in Glean we have a crate called `glean-bundle`, which depends on `glean-core`.
`glean-bundle` is the one used for the mobile targets.
Running the crate-mode bindgen against `glean-bundle` results in a panic:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: cargo metadata returned 0 packages for crate name glean_core', /Users/jer/mozilla/src/uniffi-rs/uniffi/src/lib.rs:36:2
```

somewhere along the lines `glean-core` is translated into `glean_core` (package name vs crate-internal name?)
Naively renaming just the `glean-core` package to `glean_core` results in another panic:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /Users/jer/mozilla/src/uniffi-rs/uniffi_bindgen/src/crate_mode.rs:73:10
```

So somewhere internally this uses the underscored name where it should probably use whatever is correctly defined in a `Cargo.toml`.
Note that this is not a blocker for Glean really, because we would stay on the non-crate-mode anyway for now.
But this does point to a bug of name handling internally, so this might affect others eventually too.

docs/manual/src/tutorial/foreign_language_bindings.md
bendk: Yeah, I actually added a binary named `uniffi`, but @badboy points out that would mean that we now distributed a binary again, which means it's much more likely that users would use it and get version mismatches.

Maybe this could be solved by naming?  If we called the binary `uniffi-internal-bindgen` or something that should hopefully discourage users from using it.  We can make it easy to call inside the workspace with local cargo configuration.

bendk:
> Sorry, but I didn't get very far at all here, but the note "is recommended when using external types." made me wonder if we should go all-in on this - ie, make it the default? When would we recommend to not use it? Should we make the "old" way deprecated?

I think this might be the way to go and the next release will already have breaking changes, so maybe it's a good time.  We could make the old method of specifying a UDL file require a flag rather than the other way around.


> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: cargo metadata returned 0 packages for crate name glean_core', /Users/jer/mozilla/src/uniffi-rs/uniffi/src/lib.rs:36:2

Hmm, I'll take a look at this when I get back.  I think that's probably just a bug in the code.

docs/manual/src/tutorial/foreign_language_bindings.md
badboy: We have [uniffi-bindgen-cli](https://github.com/mozilla/uniffi-rs/tree/main/examples/app/uniffi-bindgen-cli) as the internal binary for that, used in the sample app. Maybe we should just lift this up as a `uniffi-internal-bindgen` crate we don't publish?

uniffi_bindgen/src/crate_mode.rs
```diff
@@ -0,0 +1,242 @@
...
+        .collect();
+
+    match cdylib_files.len() {
+        1 => Ok(cdylib_files[0].to_owned()),
+        n => bail!("Found {n} cdylib files for {}", package.name),
+    }
+}
+
+fn build_messages(main_crate_name: &str) -> Result<Vec<Message>> {
+    let mut child = Command::new(env!("CARGO"))
```
mhammond, line 176: In slack you mentioned you run `cargo build` to find the cdylib - isn't it possible that some people will have run their "real" (or "desired"?) `cargo build` with options that will make this be incorrect - eg, things like cross-compiling? I guess I'm wondering if it might be more robust to make the consumer specify this?

mhammond (commented):
Another random thought is that "crate mode" doesn't quite seem like the right terminology - it could almost be argued that the existing mechanism, where each crate is done independently is "crate mode" and this is more like a "library mode" or similar?

uniffi_bindgen/src/crate_mode.rs
bendk: I was thinking about this and I think it requires a different approach.  Maybe instead of the user specifying the crate and UniFFI building it, the user would need to specify the final library file.  It's a bit more work for them, but they're going to have to copy that library sooner or later, so I don't think it's that big of a deal.

Alternatively, we could make this command a wrapper around `cargo build` and let the user specify all the args they want to pass to it.  I'm not sure exactly how to do this though.

bendk:
I finally got back to this one and pushed an update.  This changes "crate mode" to "library mode" and requires the user to specify the built dylib file rather than a crate name.  The main reason is [Mark's comment](https://github.com/mozilla/uniffi-rs/pull/1516#discussion_r1186283271) which points out that things will break if the command line that UniFFI uses to build the crate doesn't match the one used to build the library.

A nice side benefit of this is that we no longer need to spawn `cargo build` and pipe the output through the `cargo_metadata` crate, which is always a bit tricky to get right.  I'm pretty sure this is why the `glean-bundle` was failing, for some reason we weren't seeing all the build messages.  I just tested and it seems to work correctly now.

I also made some small changes to the docs to indicate that we expect library mode to be the default in some future UniFFI version.  Does this seem like a good plan to everyone?

mhammond (approved):
The plan and new version looks great, thanks!

fixtures/ext-types/lib/src/ext-types-lib.udl
bendk: The old names still work with the old mode, so I think I'm going to add writing up a changelog for this.

bendk:
Added a few more commits and got this is more-or-less working in [app-services](https://github.com/mozilla/application-services/pull/5592) (there's a couple build failures that I think is related to the fact that this isn't in a real release yet).  @badboy did you want to more time to review or should we merge this one?

badboy:
> Added a few more commits and got this is more-or-less working in [app-services](https://github.com/mozilla/application-services/pull/5592) (there's a couple build failures that I think is related to the fact that this isn't in a real release yet). @badboy did you want to more time to review or should we merge this one?

I can give this another look tomorrow (my) morning and merge it if everything looks ok.
(though right now there's still a merge conflict)

badboy (approved):
👍 and it works for Glean!

PR #1523, opened 2023-04-27
Update `docker/cargo-docker.sh` to new image

ijc:
Following b18505488dc2 ("Switch to updated container image").

The changes to the README are a bit silly/pointless but I figured they ought to be in sync.

/cc @badboy.


badboy (changes requested):
I hope to switch over to the mozilla-owned image next week anyway (and the name here is not correct atm), so we can just wait with the fix until we have the final name right.

PR #1528, opened 2023-04-29
Ensure unused `#[derive(uniffi::Object)]`'s don't break things

### LINKED ISSUE #1527: Objects exposed by procmacros but which aren't actually used fail
With the diff below applied, the test fails with:

```
  File "/Users/skip/src/moz/uniffi-rs/fixtures/proc-macro/tests/bindings/test_proc_macro.py", line 5, in <module>
    from uniffi_proc_macro import *
AttributeError: module 'uniffi_proc_macro' has no attribute 'Unused'
```

The problem seems to be that `add_known_type()` is only called when the object is actually used (eg, when iterating function args, record elts, etc. As a result, Python ends up adding the name to `__all__` but nothing for that type is generated.

```
diff --git a/fixtures/proc-macro/src/lib.rs b/fixtures/proc-macro/src/lib.rs
index c5f18c2a..47970d7c 100644
--- a/fixtures/proc-macro/src/lib.rs
+++ b/fixtures/proc-macro/src/lib.rs
@@ -26,6 +26,17 @@ pub struct Three {
     obj: Arc<Object>,
 }
 
+#[derive(uniffi::Object)]
+pub struct Unused;
+
+#[uniffi::export]
+impl Unused {
+    #[uniffi::constructor]
+    fn new() -> Arc<Self> {
+        Arc::new(Self)
+    }
+}
+
 #[derive(uniffi::Object)]
 pub struct Object;
 
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-256)
┆Issue Number: UNIFFI-256

mhammond:
Fixes #1527

The other alternative is to lean into the fact it's unused meaning we optimize the object out entirely, but I'm not sure that's wise.

### DIFF
--- a/fixtures/proc-macro/src/lib.rs
+++ b/fixtures/proc-macro/src/lib.rs
@@ -26,6 +26,19 @@ pub struct Three {
     obj: Arc<Object>,
 }
 
+// An object that's not used anywhere (ie, in records, function signatures, etc)
+// should not break things.
+#[derive(uniffi::Object)]
+pub struct Unused;
+
+#[uniffi::export]
+impl Unused {
+    #[uniffi::constructor]
+    fn new() -> Arc<Self> {
+        Arc::new(Self)
+    }
+}
+
 #[derive(uniffi::Object)]
 pub struct Object;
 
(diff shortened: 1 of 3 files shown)


jplatte (approved):
This actually explains why some objects weren't showing up to it us! Thanks for figuring this one out so quickly 🙂

PR #1529, opened 2023-05-01
Get traits working with procmacros

mhammond:
This turned out fairly well - some unfortunate copy/paste which was hard to completely untangle due to visibility restrictions but I was excited to work out how to do this :)

### DIFF
--- a/uniffi_macros/src/export.rs
+++ b/uniffi_macros/src/export.rs
@@ -2,8 +2,8 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use proc_macro2::TokenStream;
-use quote::quote_spanned;
+use proc_macro2::{Ident, Span, TokenStream};
+use quote::{quote, quote_spanned};
 use syn::{visit_mut::VisitMut, Item, Type};
 
 mod attributes;
@@ -15,6 +15,8 @@ use self::{
     item::{ExportItem, ImplItem},
     scaffolding::{gen_constructor_scaffolding, gen_fn_scaffolding, gen_method_scaffolding},
 };
+use crate::{object::interface_meta_static_var, util::ident_to_string};
+use uniffi_meta::free_fn_symbol_name;
 
 // TODO(jplatte): Ensure no generics, …
 // TODO(jplatte): Aggregate errors instead of short-circuiting, wherever possible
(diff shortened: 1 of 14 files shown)


uniffi_macros/src/export.rs
```diff
@@ -42,12 +44,48 @@ pub(crate) fn expand_export(
...
+            let free_tokens: syn::Result<TokenStream> = Ok(quote! {
+                #[doc(hidden)]
+                #[no_mangle]
+                pub extern "C" fn #free_fn_ident(
+                    ptr: *const ::std::ffi::c_void,
+                    call_status: &mut ::uniffi::RustCallStatus
+                ) {
+                    uniffi::rust_call(call_status, || {
+                        assert!(!ptr.is_null());
+                        drop(unsafe { Box::from_raw(ptr as *mut std::sync::Arc<dyn #self_ident>) });
```
jplatte, line 67:
```suggestion
                        drop(unsafe { ::std::boxed::Box::from_raw(ptr as *mut std::sync::Arc<dyn #self_ident>) });
```

uniffi_macros/src/export.rs
```diff
@@ -42,12 +44,48 @@ pub(crate) fn expand_export(
...
+            let impl_tokens = items.into_iter().map(|item| match item? {
+                ImplItem::Method(sig) => {
+                    gen_method_scaffolding(sig, &mod_path, &self_ident, &args, true)
+                }
+                _ => unreachable!("traits have no constructors"),
+            });
+            let meta_static_var = interface_meta_static_var(&self_ident, true, &mod_path);
+            let macro_tokens = Ok(quote! {
+                ::uniffi::ffi_converter_trait_decl!(dyn #self_ident, stringify!(#self_ident), crate::UniFfiTag);
+            });
```
jplatte, line 82: Wouldn't it be easier to call into the right proc-macro code directly, rather than doing a roundtrip through the compiler?

uniffi_macros/src/export.rs
```diff
@@ -42,12 +44,48 @@ pub(crate) fn expand_export(
...
+            let item_tokens: TokenStream = vec![meta_static_var, free_tokens, macro_tokens]
+                .into_iter()
+                .chain(impl_tokens)
+                .collect::<syn::Result<_>>()?;
```
jplatte, line 86: What's the idea behind doing all the work, then returning the first error found?

uniffi_macros/src/export.rs
mhammond: Can you explain what you mean here?

uniffi_macros/src/export.rs
mhammond: TBH I initially just copy-pasted that from a few lines above, then rolled with it. I *think* I've made the change you suggest.

uniffi_macros/src/export.rs
jplatte: Oh, I thought `ffi_converter_trait_decl` was a proc-macro like most of the macros we have, in which case it would be more efficient / easy to follow to have the proc-macro code call the codegen for that directly, rather than emitting a macro call that's then resolved and executed by the compiler.

uniffi_macros/src/export.rs
```diff
@@ -42,12 +44,53 @@ pub(crate) fn expand_export(
...
+                ) {
+                    uniffi::rust_call(call_status, || {
+                        assert!(!ptr.is_null());
+                        drop(unsafe { ::std::boxed::Box::from_raw(ptr as *mut std::sync::Arc<dyn #self_ident>) });
+                        Ok(())
+                    });
+                }
+            };
+
+            let impl_tokens: Vec<TokenStream> = items
```
jplatte, line 73:
```suggestion
            let impl_tokens: TokenStream = items
```

uniffi_macros/src/export.rs
```diff
@@ -42,12 +44,53 @@ pub(crate) fn expand_export(
...
+            let meta_static_var = interface_meta_static_var(&self_ident, true, &mod_path)?;
+            let macro_tokens = quote! {
+                ::uniffi::ffi_converter_trait_decl!(dyn #self_ident, stringify!(#self_ident), crate::UniFfiTag);
+            };
+
+            let item_tokens: TokenStream = vec![meta_static_var, free_tokens, macro_tokens]
+                .into_iter()
+                .chain(impl_tokens)
+                .collect();
+            Ok(quote_spanned! { self_ident.span() => #item_tokens })
```
jplatte, line 92:
```suggestion
            Ok(quote_spanned! { self_ident.span() =>
                #meta_static_var
                #free_tokens
                #macro_tokens
                #impl_tokens
            })
```

PR #1530, opened 2023-05-01
Fix breakage from conflict introduced with trait support

mhammond:
Oops, a conflict that didn't actually conflict means main is broken - this fixes it.


skhamis (approved):
:shipit:

PR #1534, opened 2023-05-09
Spike/Experiment: Add JSON helper methods for Python `dictionary` types

jeddai:
Okay so this is very much the definition of a draft, I figure we can use it as a point for discussion if there's interest or a reference for posterity if we decide to just close it.

### Explanation
Long story short, I was [setting up a library's UDL file](https://github.com/jeddai/application-services/blob/OLD-udl-full/components/nimbus/src/cirrus.udl) in such a way that it was going to be _very_ strongly typed across the FFI boundary. This library is going to sit behind a REST API, so converting JSON to Python types and vice-versa would be somewhat important. 

UDL `dictionary` and `enum` types are converted to Python `class`es, and **(in my limited experience with Python)** I could not for the life of me get `json.<loads|dumps>` to work with the typed classes. I tried a few things before going the route I went with this set of changes:
* Having the classes be generated as `dataclass`es
* Having the classes inherit from Pydantic's `BaseModel` (used by FastAPI)

_Author's note: to say I am inexperienced with Python would be an understatement, so there may be some simple solution to this with which I'm not familiar._

### Still needs work
* Some of the UDL types are not implemented yet, I focused on the ones I'd need to function for the library I was working on first
* Tests (there's no testing around these changes atm)
* Refactoring the processing of enums that are not flat
    * At the time I wasn't aware of how `serde_json` mapped non-flat enums with the enum name in the object-def, so the try/except setup I used in this changeset should be replaced in favor of that.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/gen_python/mod.rs
+++ b/uniffi_bindgen/src/bindings/python/gen_python/mod.rs
@@ -69,18 +69,24 @@ static KEYWORDS: Lazy<HashSet<String>> = Lazy::new(|| {
     HashSet::from_iter(kwlist.into_iter().map(|s| s.to_string()))
 });
 
+fn default_false() -> bool {
+    false
+}
+
 // Config options to customize the generated python.
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 pub struct Config {
     cdylib_name: Option<String>,
     #[serde(default)]
     custom_types: HashMap<String, CustomTypeConfig>,
+    #[serde(default = "default_false")]
+    json_support: bool,
 }
 
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 pub struct CustomTypeConfig {
-    // This `CustomTypeConfig` doesn't have a `type_name` like the others -- which is why we have
-    // separate structs rather than a shared one.
+    into_type: Option<String>,
+    from_type: Option<String>,
     imports: Option<Vec<String>>,
     into_custom: TemplateExpression,
     from_custom: TemplateExpression,
@@ -101,6 +107,7 @@ impl From<&ComponentInterface> for Config {
         Config {
             cdylib_name: Some(format!("uniffi_{}", ci.namespace())),
             custom_types: HashMap::new(),
+            json_support: false,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -69,18 +69,24 @@ static KEYWORDS: Lazy<HashSet<String>> = Lazy::new(|| {
...
+    #[serde(default = "default_false")]
+    json_support: bool,
 }
 
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 pub struct CustomTypeConfig {
-    // This `CustomTypeConfig` doesn't have a `type_name` like the others -- which is why we have
-    // separate structs rather than a shared one.
+    into_type: Option<String>,
+    from_type: Option<String>,
```
jeddai, line 89: Please ignore these, they're from a previous attempt at this and I forgot to remove them 😬

jeddai:
Totally! I did find that worked for simpler types, but the structure of the types I was generating and using didn't really lend to using that solution to any measurable degree.

mhammond:
Hi Charlie,
  Do you mind expanding on your motivation some more?

> Long story short, I was [setting up a library's UDL file](https://github.com/jeddai/application-services/blob/OLD-udl-full/components/nimbus/src/cirrus.udl) in such a way that it was going to be _very_ strongly typed across the FFI boundary. This library is going to sit behind a REST API, so converting JSON to Python types and vice-versa would be somewhat important.

The way I'd picture that working is all these would be structs on the rust side, then converted from json by serde, all in Rust. Then the "normal" Python bindings should generate code for Python objects for each of the struct. IOW, no json across the FFI at all.

I guess you have something else in mind, but I'm not quite sure what that is yet.

jeddai:
> The way I'd picture that working is all these would be structs on the rust side, then converted from json by serde, all in Rust. Then the "normal" Python bindings should generate code for Python objects for each of the struct. IOW, no json across the FFI at all.

Yep I totally agree, no JSON across the FFI. This is for the generated Python code to be able to convert JSON it receives from external places (API requests, config files, etc...) into the UniFFI'd Rust types without ever crossing the FFI.

I previously spoke with @bendk about it, and Ben's suggestion was to write Rust helper functions that would accept JSON strings and convert them into the Rust types. While that works no problem, I had already made progress on this solution and just wanted to put it somewhere and describe it for posterity's sake more than anything else. 

As mentioned in the description, happy to close this out if it's more something that should happen on the implementing code's side.

Issue #1536, opened 2023-05-12
Swift Foreign executor tests intermittently failing on CI


bendk:
I'm not sure exactly why this is happening, but here's an example of it: https://app.circleci.com/pipelines/github/mozilla/uniffi-rs/2865/workflows/5b366557-6d99-4a77-bedf-903e099fd44a/jobs/11502.

I think we should disable the tests until we figure this out.

conanoc:
```Swift
assert(result.callHappenedInDifferentThread)
```
I guess this assertion can fail because `Task`s in Swift guarantee nothing about threads. So, the code run by `Task.detached(priority: TaskPriority.background) { ... }` can run on the same thread as the caller's thread.
I wonder why this assertion is needed for the ForeignExecutor.

PR #1537, opened 2023-05-12
Fixups for the async refactor

bendk:
This fixes several issues that popped up after this was merged to main:
  - Require bytes `1.3`, which was the version that added support for the `_ne` methods.
  - Add the `ForeignExecutor` type for async methods as well as functions.
  - Disable the swift foreign executor test (#1536)
  - Put the `use` statements together in `uniffi_core/src/ffi_converter_impls.rs`.  This didn't cause any isuses but it was just weird.
  - Fixed a couple of `trivial-cast` warnings that I saw when trying to compile the matrix code.

### DIFF
--- a/fixtures/foreign-executor/tests/test_generated_bindings.rs
+++ b/fixtures/foreign-executor/tests/test_generated_bindings.rs
@@ -1,5 +1,7 @@
 uniffi::build_foreign_language_testcases!(
     "tests/bindings/test_foreign_executor.py",
     "tests/bindings/test_foreign_executor.kts",
-    "tests/bindings/test_foreign_executor.swift",
+    // Disabled because of intermittent CI failures
+    // (https://github.com/mozilla/uniffi-rs/issues/1536)
+    // "tests/bindings/test_foreign_executor.swift",
 );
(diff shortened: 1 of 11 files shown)


tarikeshaq (approved):
Big rubber stamp as I'm unfamiliar with the code here, but this is a fixup and our tests are passing so happy to see it land if it'll fix breakage

jplatte:
This did indeed fix the errors about `ForeignExecutor` not being found 👍🏼

fixtures/foreign-executor/tests/test_generated_bindings.rs
```diff
@@ -1,5 +1,7 @@
 uniffi::build_foreign_language_testcases!(
     "tests/bindings/test_foreign_executor.py",
     "tests/bindings/test_foreign_executor.kts",
-    "tests/bindings/test_foreign_executor.swift",
+    // Disabled because of intermittent CI failures
+    // (https://github.com/mozilla/uniffi-rs/issues/1536)
+    // "tests/bindings/test_foreign_executor.swift",
```
Hywan, line 6: It must be a joke :-D. There is a malediction around those tests ;-).

PR #1539, opened 2023-05-12
Rename the Swift ffi type filter functions (#1106)

bendk:
The current names are very confusing to me.  The new ones make it clear which to use when.


bendk:
From the issue:

> I wonder if this should prompt us to break the file in to a gen_swift and gen_header files, with two different filters modules?

I wouldn't be against this, but I didn't do it in this PR since there's currently just 1 filter function used in the header template.

jhugman (approved):
This looks fine to me.

PR #1543, opened 2023-05-16
Introduce `bytes` type

### LINKED ISSUE #1299: Python bindings: use `bytes` for `sequence<u8>`
currently, `sequence<u8>` is mapped to a list of integers, while it may make more sense to use `bytes` or `byterarray`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-181)
┆friendlyId: UNIFFI-181

### LINKED ISSUE #1541: Add a `bytes` type to UDL
Some languages, such as e.g. Python, distinguish between arrays of 8-bit integers (`List[int]` in Python) and byte buffers (`bytes` in Python).

This would IMO neatly solve #1299.

Kotlin: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-byte-array/
Python: https://docs.python.org/3/library/stdtypes.html#bytes

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-262)
┆Issue Number: UNIFFI-262

heinrich5991:
This type is mapped to the native bytestring type in the various languages, i.e.

- Kotlin: `ByteArray`
- Python: `bytes`
- Ruby: `String` with `Encoding::BINARY`
- Swift: `Data`

Also add tests exercising this new type.

Fixes #1299.
Fixes #1541.


heinrich5991:
Where/how would I add tests?

This is just a draft so far, Python works, the rest of the languages aren't implemented yet. What do you think about the current approach?

CC #1299 #1541

mhammond (commented):
This looks like a good start. I think I prefer just "bytes" instead of "bytestring" though. For tests you should look at, say, fixtures/coverall and add a "bytes" object to each of the types supported, and as results and args to methods/functions etc, then each language should test them too.

heinrich5991:
@mhammond I added support for Kotlin, Ruby and Swift and added tests. For Kotlin, I wasn't sure whether to use the `UByteArray` or the `ByteArray` type, `UByteArray` seems newer and not as well integrated as `ByteArray`.

The PR should be ready for review now.

fixtures/coverall/tests/bindings/test_coverall.kts
```diff
@@ -39,6 +39,34 @@ createSomeDict().use { d ->
     assert(d.coveralls!!.getName() == "some_dict")
 }
 
+createNoneDict().use { d ->
```
badboy, line 42: Where's this from? Seems unrelated to the `bytes` part

fixtures/coverall/tests/test_generated_bindings.rs
```diff
@@ -3,5 +3,5 @@ uniffi::build_foreign_language_testcases!(
     "tests/bindings/test_coverall.kts",
     "tests/bindings/test_coverall.rb",
     "tests/bindings/test_coverall.swift",
-    "tests/bindings/test_handlerace.kts",
+    //"tests/bindings/test_handlerace.kts",
```
badboy, line 6: Why is this commented out now?

badboy (commented):
I haven't given this a full review yet, but will do that asap.

fixtures/coverall/tests/bindings/test_coverall.kts
heinrich5991: This test was part of the other language's tests, so I added it to Kotlin as well. Should I move it to a different commit/remove it altogether?

fixtures/coverall/tests/test_generated_bindings.rs
heinrich5991: Oops, that was just used while developing, will add it again.

fixtures/coverall/tests/bindings/test_coverall.kts
badboy: Interesting. Yeah, I don't mind having that added, but as a separate commit would be good then.

heinrich5991: Done.

badboy:
> `contentEquals`

Because `==` is just `a.equals(b)` and `equals` is not overriden for those collections.

uniffi_bindgen/src/bindings/kotlin/templates/UByteArrayHelper.kt
```diff
@@ -0,0 +1,16 @@
+@OptIn(kotlin.ExperimentalUnsignedTypes::class)
+public object FfiConverterUByteArray: FfiConverterRustBuffer<UByteArray> {
```
linabutler, line 2: Thanks so much for putting this PR together!

How would you (and other folks reviewing the PR) feel about using `ByteArray` instead of `UByteArray` in Kotlin?

It looks like it's possible to use `UByteArray.asByteArray()` to turn a `UByteArray` into a `ByteArray` without an extra copy, but many Java and Android interfaces work with `ByteArray` exclusively, and it might be nice to avoid the `toUByteArray`/`toByteArray` copies in `read`/`write`, and also avoid having to opt-in to experimental unsigned types every time...

heinrich5991: > and it might be nice to avoid the `toUByteArray`/`toByteArray` copies in `read`/`write`

I didn't see the `UByteArray.asByteArray` function when composing the PR, it seems there's also the inverse `ByteArray.asUByteArray`, so this could all be done without copies, and the performance point is moot.

> How would you (and other folks reviewing the PR) feel about using `ByteArray` instead of `UByteArray` in Kotlin?

I don't know enough about Kotlin to have feelings about this, really. I went with `UByteArray` because it's unsigned in the other languages, too, but I'll happily change it to `ByteArray` if that's more in the way of Kotlin.

heinrich5991: Done.

badboy (approved):
Some minimal nits regarding the tests, but otherwise this should be good to go.

heinrich5991:
Done.

heinrich5991:
Test failure seems unrelated.

mhammond (commented):
Fantastic! It looks great, and sorry I didn't mention this before, but this really needs a changelog entry and a change to the docs

heinrich5991:
> this really needs a changelog entry and a change to the docs

Done, I think. Did I find all the relevant places?

mhammond (approved):
Thanks!

PR #1545, opened 2023-05-17
Remove `./` from paths

heinrich5991:
Make `CODE_OF_CONDUCT.md` non-executable and remove unneeded `./` from paths. `./` doesn't do anything.


mhammond:
I like the first and last commits, but don't understand what harm the "./" is doing?

heinrich5991:
> I like the first and last commits, but don't understand what harm the "./" is doing?

No harm, just superfluous. What value does "./" add? I can drop the commit if wanted.

mhammond (approved):
Ok, I'm sold :)

PR #1546, opened 2023-05-17
Check for integer overflows in Python bindings

### LINKED ISSUE #1540: Integer overflows not checked (at least for Python)
If I follow [the tutorial](https://mozilla.github.io/uniffi-rs/tutorial/udl_file.html) and end up with a UDL like this:
```udl
namespace math {
  u32 add(u32 a, u32 b);
};
```
I can call this function from Python using `math.add(2**32, 0)` and get no error. Instead, the first integer is silently truncated to 32 bits, becoming 0.

I'd expect this to raise an error instead.

Looking into the source code, it seems this is done already if the integer is part of a larger structure; in this case the `struct` Python module is used which throws an exception for out-of-range integers.

heinrich5991:
Not spelling out the numbers as numbers is fine, CPython optimizes it out:
```py
>>> import dis
>>> dis.dis("not -2**63 <= value < 2**63")
  0           0 RESUME                   0

  1           2 LOAD_CONST               0 (-9223372036854775808)
              4 LOAD_NAME                0 (value)
              6 SWAP                     2
              8 COPY                     2
             10 COMPARE_OP               1 (<=)
             16 JUMP_IF_FALSE_OR_POP     5 (to 28)
             18 LOAD_CONST               1 (9223372036854775808)
             20 COMPARE_OP               0 (<)
             26 JUMP_FORWARD             2 (to 32)
        >>   28 SWAP                     2
             30 POP_TOP
        >>   32 UNARY_NOT
             34 RETURN_VALUE
```

The exception thrown imitates the one of the `struct` module:
```py
>>> import struct
>>> struct.pack("<I", 2**32)
  File "<stdin>", line 1, in <module>
struct.error: 'I' format requires 0 <= number <= 4294967295
```

Also add some tests exercising these code paths.

Fixes #1540.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/templates/UInt8Helper.py
+++ b/uniffi_bindgen/src/bindings/python/templates/UInt8Helper.py
@@ -1,4 +1,10 @@
 class FfiConverterUInt8(FfiConverterPrimitive):
+    @classmethod
+    def lower(cls, value):
+        if not 0 <= value < 2**8:
+            raise ValueError("u8 requires {} <= value < {}".format(0, 2**8))
+        return super().lower(value)
+
     @staticmethod
     def read(buf):
         return buf.readU8()
--- a/uniffi_bindgen/src/bindings/python/templates/UInt8Helper.py
+++ b/uniffi_bindgen/src/bindings/python/templates/UInt8Helper.py
@@ -1,9 +1,7 @@
-class FfiConverterUInt8(FfiConverterPrimitive):
-    @classmethod
-    def lower(cls, value):
-        if not 0 <= value < 2**8:
-            raise ValueError("u8 requires {} <= value < {}".format(0, 2**8))
-        return super().lower(value)
+class FfiConverterUInt8(FfiConverterPrimitiveInt):
+    CLASS_NAME = "u8"
+    VALUE_MIN = 0
+    VALUE_MAX = 2**8
 
     @staticmethod
     def read(buf):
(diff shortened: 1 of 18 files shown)


heinrich5991:
Test failure seems unrelated.

uniffi_bindgen/src/bindings/python/templates/UInt8Helper.py
```diff
@@ -1,4 +1,10 @@
 class FfiConverterUInt8(FfiConverterPrimitive):
+    @classmethod
+    def lower(cls, value):
+        if not 0 <= value < 2**8:
+            raise ValueError("u8 requires {} <= value < {}".format(0, 2**8))
```
bendk, line 5: I'd love to reduce the code duplication here. What about this?
  - Define `class FfiConverterInt(FfiConvererPrimitive)`
  - `FfiConverterInt` overrides lower and adds the bounds check, but doesn't define the upper/lower bounds.
  - Each integer type inherits from `FfiConverterInt` and defines the bounds.

bendk (changes requested):
The code looks correct to me and thanks for adding the tests!  I had coding suggestion, but feel free to push back on it if there's something I'm missing.

bendk:
I guess there's also the possibility of float overflows, but I'm not sure how big of an issue that is in practice.

uniffi_bindgen/src/bindings/python/templates/UInt8Helper.py
heinrich5991: That would work, it'd come with a (tiny) performance penalty, but I guess that's okay.

heinrich5991: Done.

bendk (approved):

PR #1549, opened 2023-05-17
Callback interface returns

### LINKED ISSUE #1547: `callback interface` returning `void` causes "reading the unit type is not allowed"
A `callback interface` method which returns void is resulting in `reading the unit type is not allowed` when lifting the result.

This started in #1515 and persists at current main 58e0a24dbcd2785c71dd2ccffd99f892abf9e785.

The commit message for 884403ddb0e994e1807b49119e9ec2090e2ac27e mentions using a `u8` as a placeholder for `void` returns from async functions, but these are not async so I'm not sure what is going wrong but maybe it's to do with those changes.

I have pushed a reproducer to: https://github.com/mozilla/uniffi-rs/compare/main...ijc:uniffi-rs:add-test-for-callback-returning-void (first commit is based at a93713ac0c87d7a5dc69763e143cc2f39e34f3c6 where it passes, then master is merged ontop to demonstrate the failure). I'll convert that to a PR if I can spot the issue.

/cc @bendk.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-266)
┆Issue Number: UNIFFI-266

bendk:
Refactoring the callback interface return code
    
Added a `FfiConverter` method to handle returning values from callback interface methods.  This removes the need for multiple `invoke_callback_*` methods and fixes #1547.

Removed the swift code that wrote an string to the RustBuffer for flat errors.  No other bindings do that and we just ignored it on the Rust side.

Implemented the lift/lower/read/write for the unit struct.  We now call these methods in the callback interface code.

### DIFF
--- a/fixtures/callbacks/src/callbacks.udl
+++ b/fixtures/callbacks/src/callbacks.udl
@@ -26,6 +26,8 @@ callback interface ForeignGetters {
   string? get_option(string? v, boolean arg2);
   [Throws=SimpleError]
   sequence<i32> get_list(sequence<i32> v, boolean arg2);
+  [Throws=SimpleError]
+  void get_nothing(string v);
 };
 
 /// These objects are implemented in Rust, and call out to `ForeignGetters`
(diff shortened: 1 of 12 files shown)


mhammond (approved):
I've been poking around some of this recently so I'll steal the review - looks sane to me!

ijc:
Thanks for this, I can confirm that this has fixed my application's use of Uniffi.

PR #1550, opened 2023-05-19
Kotlin: Add an upper-bound for UniFfiHandleMap

bendk:
Without this a user could pass a null value, which isn't correct. This prevents compiler errors when I try to use the new UniFFI version with application-services.


skhamis (approved):
LGTM!

Issue #1286, opened 2022-06-28
Override modifier in Kotlin


kirillzh:
Consider this UDL interface:
```udl
interface Test {
  string to_string();
};
```

This generates Kotlin code that does not compile because `toString()` function is missing an `override` modifier since this is an open, member function declared in the `Any` class, which all classes implement in Kotlin:
```kotlin
interface Test {
    // This won't compile: 'toString' hides member of supertype 'Any' and needs 'override' modifier
    fun toString(): String
}
```

Correct way to fix generated Kotlin code is to add missing `override` modifier:
```kotlin
interface Test {
    override fun toString(): String
}
```

As far as I can tell, uniffi doesn't currently provide a way to generate `override` modifier. Would it be reasonable to add support for it, perhaps, with an UDL syntax like this?
```udl
interface Test {
  override string to_string();
};
```

Meanwhile, an alternative workaround here is to use different API naming.

bendk:
I don't know how to do this in general, since which functions need `override` is language-specific.  However, a `to_string` method is very common and it could be nice to support.  I think this would mean:
  - Having a way to flag the function.  Maybe this is an attribute, or maybe we special-case the name `to_string()`
  - For languages that support it (which I think is all the current languages), generate a stringify function with the proper name and signature.

quad:
@kirillzh Is there any situation in which `toString` _wouldn't_ need an associated `override` in Kotlin?

If not, then @bendk, would you accept a PR that adds that behaviour to the Kotlin binding templates?

It would be akin to the existing code to do fixups for things like reserved words.

quad:
[There are only three methods on Kotlin's `Any` interface:](https://github.com/JetBrains/kotlin/blob/2f75ea585ef6165b11138aaadd3efecc4a3d4828/core/builtins/native/kotlin/Any.kt)
- equals
- hashCode
- toString

I strongly suspect that `toString ` is the only reasonable method for an FFI definition to override.

quad:
**(Potential) Implementation notes**

1. Add a function to [`CodeOracle`](https://github.com/mozilla/uniffi-rs/blob/cce0145527e3f7f251e4435f0749af1ab91dfb84/uniffi_bindgen/src/backend/oracle.rs), perhaps something like `fn_name_on_base_class`
2. Implement it for each binding; `kotlin` will use it, `python` and `swift` won't, `ruby` _could_ but overriding methods is nicer there
[...]

PR #1551, opened 2023-05-21
Implement magic "trait methods"

mhammond:
This is a POC for supporting "magic" methods in foreign bindings. In the included test, we show Python generating `__repr__`, `__str__`, `__eq__` and `__hash__` for a uniffi exposed object, each with the correct semantics. I haven't implemented any languages other than Python, but hopefully it's clear how Kotlin might work as requested in #1286.

It's a bit hacky, bit it turned out better than I was expecting.
* The "trait names" are just strings, which we assume are among a "well known" set and currently match Rust trait names (eg, `Display`, `PartialEq`, etc). 
* In all cases, we generate "custom" ffi signatures in the rust crate, and these signatures are assumed to be known by the bindings - ie, there's no way to communicate that the ffi hash function has a signature of `fn(&self)->u64`, but the foreign side is expected to just know that.
* No support for macros yet, but that should be relatively modest.

@jplatte @bendk @badboy WDTY?

### DIFF
--- a/fixtures/uitests/tests/ui/trait_methods_no_trait.stderr
+++ b/fixtures/uitests/tests/ui/trait_methods_no_trait.stderr
@@ -1,0 +1,28 @@
+error[E0277]: `TraitMethods` doesn't implement `std::fmt::Display`
+ --> $OUT_DIR[uniffi_uitests]/trait_methods.uniffi.rs
+  |
+  | ...   uniffi::deps::static_assertions::assert_impl_all!(r#TraitMethods: std::fmt::Display); // This object has a trait method which requi...
+  |                                                         ^^^^^^^^^^^^^^ `TraitMethods` cannot be formatted with the default formatter
+  |
+  = help: the trait `std::fmt::Display` is not implemented for `TraitMethods`
+  = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
+note: required by a bound in `uniffi_trait_methods_fn_method_traitmethods_uniffi_trait_display::{closure#0}::_::{closure#0}::assert_impl_all`
+ --> $OUT_DIR[uniffi_uitests]/trait_methods.uniffi.rs
+  |
+  |             uniffi::deps::static_assertions::assert_impl_all!(r#TraitMethods: std::fmt::Display); // This object has a trait method which...
+  |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `uniffi_trait_methods_fn_method_traitmethods_uniffi_trait_display::{closure#0}::_::{closure#0}::assert_impl_all`
+  = note: this error originates in the macro `uniffi::deps::static_assertions::assert_impl_all` (in Nightly builds, run with -Z macro-backtrace for more info)
+
+error[E0277]: `TraitMethods` doesn't implement `std::fmt::Display`
+ --> $OUT_DIR[uniffi_uitests]/trait_methods.uniffi.rs
+  |
+  | /                 match<std::sync::Arc<r#TraitMethods> as ::uniffi::FfiConverter<crate::UniFfiTag>>::try_lift(r#ptr) {
+  | |                     Ok(ref val) => val,
+  | |                     Err(err) => panic!("Failed to convert arg '{}': {}", "ptr", err),
+  | |                 }
+  | |_________________^ `TraitMethods` cannot be formatted with the default formatter
+  |
+  = help: the trait `std::fmt::Display` is not implemented for `TraitMethods`
+  = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
+  = help: the trait `std::fmt::Display` is implemented for `Arc<T>`
+  = note: this error originates in the macro `$crate::__export::format_args` which comes from the expansion of the macro `format` (in Nightly builds, run with -Z macro-backtrace for more info)
--- a/uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
@@ -1,13 +1,13 @@
 {%- let obj = ci.get_object_definition(name).unwrap() %}
 
 class {{ type_name }}:
-    {%- match obj.primary_constructor() %}
-    {%- when Some with (cons) %}
+{%- match obj.primary_constructor() %}
+{%-     when Some with (cons) %}
     def __init__(self, {% call py::arg_list_decl(cons) -%}):
         {%- call py::setup_args_extra_indent(cons) %}
         self._pointer = {% call py::to_ffi_call(cons) %}
-    {%- when None %}
-    {%- endmatch %}
+{%-     when None %}
+{%- endmatch %}
 
     def __del__(self):
         # In case of partial initialization of instances.
@@ -24,44 +24,33 @@ class {{ type_name }}:
         inst._pointer = pointer
         return inst
 
-    {% for cons in obj.alternate_constructors() -%}
+{%- for cons in obj.alternate_constructors() %}
+
     @classmethod
     def {{ cons.name()|fn_name }}(cls, {% call py::arg_list_decl(cons) %}):
         {%- call py::setup_args_extra_indent(cons) %}
         # Call the (fallible) function before creating any half-baked object instances.
         pointer = {% call py::to_ffi_call(cons) %}
         return cls._make_instance_(pointer)
-    {% endfor %}
+{% endfor %}
 
-    {% for meth in obj.methods() -%}
-    {% if meth.is_async() %}
+{%- for meth in obj.methods() -%}
+    {%- call py::method_decl(meth.name()|fn_name, meth) %}
+{% endfor %}
 
-    async def {{ meth.name()|fn_name }}(self, {% call py::arg_list_decl(meth) %}):
-        {%- call py::setup_args_extra_indent(meth) %}
-        return await rust_call_async(
-            _UniFFILib.{{ func.ffi_func().name() }},
-            {{ func.result_type().borrow()|async_callback_fn }},
-            self._pointer,
-            {% call py::arg_list_lowered(func) %}
-        )
-
-    {% else %}
-    {%- match meth.return_type() -%}
-
-    {%- when Some with (return_type) -%}
-    def {{ meth.name()|fn_name }}(self, {% call py::arg_list_decl(meth) %}):
-        {%- call py::setup_args_extra_indent(meth) %}
-        return {{ return_type|lift_fn }}(
-            {% call py::to_ffi_call_with_prefix("self._pointer", meth) %}
-        )
-
-    {%- when None -%}
-    def {{ meth.name()|fn_name }}(self, {% call py::arg_list_decl(meth) %}):
-        {%- call py::setup_args_extra_indent(meth) %}
-        {% call py::to_ffi_call_with_prefix("self._pointer", meth) %}
-    {% endmatch %}
-    {% endif %}
-    {% endfor %}
+{%- for tm in obj.uniffi_traits() -%}
+{%-     match tm %}
+{%-         when UniffiTrait::Debug { fmt } %}
+            {%- call py::method_decl("__repr__", fmt) %}
+{%-         when UniffiTrait::Display { fmt } %}
+            {%- call py::method_decl("__str__", fmt) %}
+{%-         when UniffiTrait::Eq { eq, ne } %}
+            {%- call py::method_decl("__eq__", eq) %}
+            {%- call py::method_decl("__ne__", ne) %}
+{%-         when UniffiTrait::Hash { hash } %}
+            {%- call py::method_decl("__hash__", hash) %}
+{%      endmatch %}
+{% endfor %}
 
 
 class {{ ffi_converter_name }}:
--- a/uniffi_bindgen/src/interface/attributes.rs
+++ b/uniffi_bindgen/src/interface/attributes.rs
@@ -33,6 +33,7 @@ pub(super) enum Attribute {
     Name(String),
     SelfType(SelfType),
     Throws(String),
+    Traits(Vec<String>),
     // `[External="crate_name"]` - We can `use crate_name::...` for the type.
     External {
         crate_name: String,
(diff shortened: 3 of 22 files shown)


uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
```diff
@@ -1,13 +1,13 @@
 {%- let obj = ci.get_object_definition(name).unwrap() %}
 
 class {{ type_name }}(object):
-    {%- match obj.primary_constructor() %}
-    {%- when Some with (cons) %}
+{%- match obj.primary_constructor() %}
```
mhammond, line 4: I was having real trouble reading the askama control flow, so I re-indented all control flow in this file, which I think makes it far more readable, but lmk if you think I've made things worse here.

jplatte:
> Note however a "name" is still required for them - it will be used on the rust side of the world

It looks like the scaffolding code generation already special-cases some of these methods and just uses the right trait methods, rather than calling the method of the specified name. I think we should do that consistently, and not pretend that there's a meaningful name on the Rust side. Then the UDL could also look very much like the suggested proc-macro solution:

```webidl
[Magic=Display, Magic=Debug, Magic=Hash]
interface Magic {
    // ...
}
```

jplatte:
> and by foreign bindings which don't support magic methods (although maybe we should arrange for such languages to ignore them?)

I think we should leave it up to the bindings generator for each language what to map each of these FFI functions to. Ignoring them would likely not be what users expect, I think the best solution would be to use whatever is the most idiomatic name for such a method in the target language when there is no standard interface / protocol / whatever to be implemented.

uniffi_bindgen/src/scaffolding/templates/macros.rs
```diff
@@ -6,6 +6,35 @@
...
+    {
+        // the magic method "hash" - described as `fn(&self) -> i64`
+        use ::std::hash::{Hash, Hasher};
+        let mut s = ::std::collections::hash_map::DefaultHasher::new();
+        match<std::sync::Arc<{{ obj_name }}> as ::uniffi::FfiConverter<crate::UniFfiTag>>::try_lift(r#ptr) {
+            Ok(ref val) => val,
+            Err(err) => panic!("Failed to convert arg '{}': {}", "ptr", err),
+        }.hash(&mut s);
+        s.finish()
+    }
```
bendk, line 21: It seems fine to me that this requires some extra work to since the signature doesn't match.  If it was easy to make this a generic function in `uniffi_core`, then I'd say we should move it there.

bendk (commented):
This is pretty exciting.  I had some bikeshedding comments, but in general this seems like a good idea and we can implement it without too much complexity.

uniffi_bindgen/src/scaffolding/templates/macros.rs
```diff
@@ -6,6 +6,35 @@
...
+{#  Some magic methods need a uniffi-compatible shim #}
+{% if meth.magic() == "hash" %}
+    {
+        // the magic method "hash" - described as `fn(&self) -> i64`
+        use ::std::hash::{Hash, Hasher};
+        let mut s = ::std::collections::hash_map::DefaultHasher::new();
+        match<std::sync::Arc<{{ obj_name }}> as ::uniffi::FfiConverter<crate::UniFfiTag>>::try_lift(r#ptr) {
+            Ok(ref val) => val,
+            Err(err) => panic!("Failed to convert arg '{}': {}", "ptr", err),
+        }.hash(&mut s);
```
bendk, line 19: I think this should be fully-qualified (something like `<val as Hash>::hash(&val, &mut s)`) just to be sure there's no name conflict.  The same goes for magic methods where the signature matches.

mhammond:
I saw Ben's comments too late, so haven't renamed "magic" to "traits" - but I think that makes alot of sense the way it is now shaped. I took almost all @jplatte's suggestions though - this is looking quite nice now, lmk what you think!

uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
```diff
@@ -24,19 +24,33 @@ def _make_instance_(cls, pointer):
...
+{%-          let method_name = "__repr__".to_string() %}
+{%-      else if meth.magic() == "str" %}
+{%-          let method_name = "__str__".to_string() %}
+{%-      else if meth.magic() == "eq" %}
+{%-          let method_name = "__eq__".to_string() %}
+{%-      else if meth.magic() == "hash" %}
+{%-          let method_name = "__hash__".to_string() %}
+{%-       else %}
+{%           let method_name = meth.name()|fn_name %}
+{%-       endif %}
```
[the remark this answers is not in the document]
mhammond: I guess we have that requirement now - for `PartialEq` to be correct it should expose 2 methods. I guess the new `TraitMethods` struct could have a `Vec<Methods>` and part of the "contract" with uniffi is that the method names are listed in the order they appear in the trait (ie, `[0]` is `__eq__`, `[1]` is `__ne__`). This doesn't sound too bad, as there is already an implicit contract re the function signatures.

badboy:
I like it.
[I managed to implement support for it in the Swift part easily](https://github.com/badboy/uniffi-rs/commit/20f824b2).

jplatte:
Looking pretty good! I think for scaffolding it might make sense to go for a macro-based solution that can be reused by the proc-macro frontend, but we can refactor things later as well.

mhammond:
I ended up supporting multiple methods per traits given it's technically a requirement today for PartialEq, plus the realization that if we needed to add it in the future it would end up as a breaking change for bindings, whereas this patch is not - ie, this PR should not have broken swift, but the change to multiple methods *would* have broken it if @badboy's swift implementation landed. So sorry @badboy I broke it :)

I agree we can refactor a little for procmacros, but think this is ready to land!

mhammond:
I can make the traits an enum now, so I'll poke at that

mhammond:
enum was easy :) I named it `UniffiTrait` and implemented it in `uniffi_meta`, then used that from uniffi_bindgen. Happy to bikeshed on the trait name, and @jplatte, lmk if you think that will work OK for procmacros.

badboy:
> I ended up supporting multiple methods per traits given it's technically a requirement today for PartialEq, plus the realization that if we needed to add it in the future it would end up as a breaking change for bindings, whereas this patch is not - ie, this PR should not have broken swift, but the change to multiple methods _would_ have broken it if @badboy's swift implementation landed. So sorry @badboy I broke it :)

No worries, I had a little bit of time and just explored how these traits work in Swift (and in doing so learned about `CustomDebugStringConvertible` and `CustomStringConvertible`). I fully expected it to need an update once this PR finds its final form anyway.

uniffi_bindgen/src/scaffolding/templates/macros.rs
```diff
@@ -62,3 +62,22 @@ r#{{ func.name() }}({% call _arg_list_rs_call(func) -%})
...
+    uniffi::rust_call(call_status, || {
+        {{ meth|return_ffi_converter }}::lower_return(
+{%- endmacro %}
+
+{%- macro method_decl_postscript(meth) %}
+            {% if meth.throws() %}.map_err(Into::into){% endif %}
+        )
+    })
+}
+{% endmacro %}
```
jplatte: Another missing final newline.

jplatte (commented):
How bad of a breaking change would it be to introduce extra arguments to an existing of these trait methods later (for example a flag for enabling pretty-printing in `Debug`)?

jplatte:
I think `Display` and `Debug` are sufficiently different in target languages (`__str__` vs `__repr__`, `CustomStringConvertible` vs. ` CustomDebugStringConvertible`) that we should not try to have one interface for everything.

That actually leads me to another question: Do the target languages supported by UniFFI directly have any sort of pretty-printing / "alternate" flag for debug strings like Rust does? If not, maybe it's not worth adding support for this in any way.

mhammond:
> I think Display and Debug are sufficiently different in target languages (__str__ vs __repr__, CustomStringConvertible vs.  CustomDebugStringConvertible) that we should not try to have one interface for everything.

I agree from the foreign side, but from the rust side if we have 2 functions that differ only by the static format string they pass to `format!()` it does seem a bit of a waste - especially if we ended up allowing one of them to use a different format string based on an argument. But it's a trivial optimization so I don't care.

> Do the target languages supported by UniFFI directly have any sort of pretty-printing / "alternate" flag for debug strings like Rust does?

Python does support overriding `__format__` but a quick look shows it's probably not useful for this use-case - not sure about others. But IMO it's OK that a work around to get pretty printing is to say `obj._trait_format(true)`?

mhammond:
Enabling pretty-printing didn't work out - without major work, supporting that would mean Python ends up generating `__repr__(self, want_pretty=DEFAULT)` or similar - which works fine for Python, but would not work OK for strongly typed languages. It's probably possible to work around that, but I think that can be a followup.

So I'm done here!

uniffi_bindgen/src/interface/object.rs
```diff
@@ -602,6 +698,29 @@ impl APIConverter<Method> for weedle::interface::OperationInterfaceMember<'_> {
     }
 }
 
+#[derive(Debug, Clone, Checksum)]
+pub struct TraitMethod {
+    pub(super) id: UniffiTrait,
+    // The list of methods exposed by this trait. Most will be 1, but eg, `PartialEq` will have `eq` and `ne`.
+    // The order is defined as being the same order as defined in the rust trait - eg, it's understood that
+    // `eq` is 0, `ne` is 1.
+    pub(super) methods: Vec<Method>,
```
bendk, line 707: I don't love this `methods` field and the hard-coded indexes.  What if `TraitMethod` was an enum?

```
enum TraitMethod {
   Display(Method),
   Debug(Method),
   PartialEq {
       eq: Method,
       ne: Method,
   },
   ...
}
```

I also had some naming suggestions that I don't feel that strongly about:
  - Prefix it with `Uniffi` to make it consistent with `UniffiTrait`
  - s/Method/Methods/

fixtures/uitests/tests/ui/trait_methods_no_trait.stderr
```diff
@@ -0,0 +1,28 @@
+error[E0277]: `TraitMethods` doesn't implement `std::fmt::Display`
```
bendk, line 1: Nice.

uniffi_meta/src/lib.rs
```diff
@@ -174,6 +174,27 @@ impl MethodMetadata {
...
+    Hash,
+}
+
+impl TryFrom<&str> for UniffiTrait {
+    type Error = anyhow::Error;
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        match value {
+            "Debug" => Ok(UniffiTrait::Debug),
+            "Display" => Ok(UniffiTrait::Display),
+            "PartialEq" => Ok(UniffiTrait::PartialEq),
```
bendk, line 191: Should this be `Eq`?  It seems weird to define the `==` operator, but not have it transitive/reflexive/symmetric.  I'm not really sure about this one though.

uniffi_bindgen/src/interface/object.rs
mhammond: > What if TraitMethod was an enum?

I decided to avoid that in the interest if keeping procmacros sane - IIUC, it will need to "emit" these, and emitting an array of MethodMetadata is probably easier than emitting an enum with struct fields?

Happy to do that though if that consideration doesn't actually apply

uniffi_meta/src/lib.rs
mhammond: Me either - PartialEq is what actually defines the methods we use though, but I guess `Eq` is probably closer to what other languages actually need here (at least, is a ore conservative choice and likely to support all languages). So yeah, I guess `Eq` is better

uniffi_bindgen/src/interface/object.rs
jplatte: I don't think the enum solution will be more complex for proc-macros. I'm in favor of this change.

jplatte (approved):
Just want to say I have no additional concerns, happy to see this merge whenever bendk is happy 👍🏼

uniffi_meta/src/lib.rs
mhammond: To put it another way: if a struct defines `PartialEq` but not `Eq`, and the definition isn't transitive, IIUC that would meet the requirement for rust, but unlikely to meet the requirement for foreign languages. Rust implementing `Eq` would.

uniffi_bindgen/src/interface/object.rs
mhammond: >s/Method/Methods/

I think the singular "Method" would be more consistent (eg, `Attribute` is singular), but `UniffiTrait` is probably better, yeah

uniffi_bindgen/src/interface/attributes.rs
```diff
@@ -284,6 +311,7 @@ impl TryFrom<&weedle::attribute::ExtendedAttributeList<'_>> for InterfaceAttribu
             Attribute::Enum => Ok(()),
             Attribute::Error => Ok(()),
             Attribute::Trait => Ok(()),
+            Attribute::Traits(_) => Ok(()),
```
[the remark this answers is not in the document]
mhammond: Can't see why note!

mhammond:
As we discussed, I needed to move the enum back into uniffi_bindgen, but I think this is much cleaner.

mhammond:
Oh, I meant to say I'm so close to this I can't remember what we said were good names and what actually is - so feel free to bikeshed names still!

bendk (approved):
Looks great to me, I just had some very minor suggestions that you can take or leave.

uniffi_bindgen/src/interface/object.rs
```diff
@@ -170,6 +179,20 @@ impl Object {
...
+                    .flat_map(|ut| match ut {
+                        UniffiTrait::Display { fmt: m }
+                        | UniffiTrait::Debug { fmt: m }
+                        | UniffiTrait::Hash { hash: m } => vec![m],
+                        UniffiTrait::Eq { eq, ne } => vec![eq, ne],
+                    })
+                    .collect::<Vec<&Method>>()
+                    .into_iter()
+                    .map(|m| &m.ffi_func)
+                    .into_iter(),
```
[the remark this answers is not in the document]
jplatte: Pretty sure this is an identity conversion.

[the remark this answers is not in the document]
jplatte: And the `.collect::<Vec<_>>().into_iter()` is also unnecessary I think?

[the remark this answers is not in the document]
mhammond: Yeah, you are both correct. I have to admit I spent too long battling that block and as soon as I had it working I celebrated and moved on :)

PR #1552, opened 2023-05-23
Don't explicitly derive from `object` in Python

heinrich5991:
That's an artifact from Python 2, Python 3 does that automatically.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/templates/EnumTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/EnumTemplate.py
@@ -19,7 +19,7 @@ class {{ type_name }}:
 
     # Each enum variant is a nested class of the enum itself.
     {% for variant in e.variants() -%}
-    class {{ variant.name()|enum_variant_py }}(object):
+    class {{ variant.name()|enum_variant_py }}:
         def __init__(self,{% for field in variant.fields() %}{{ field.name()|var_name }}{% if loop.last %}{% else %}, {% endif %}{% endfor %}):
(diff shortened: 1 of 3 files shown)


bendk:
I think there was an intention to support Python 2 at one point, but I feel like now could be a good time to stop supporting it.  I think @mhammond and @badboy are the two main Python devs, what do you think?

In any case, please add a breaking change entry to `CHANGELOG.md` saying that we're dropping support for Python 2

badboy (approved):
Python 2 is dead, long live Python 3!

(but what bendk said: a changelog entry would be nice)

mhammond:
I don't really object to dropping support for python 2, but what harm is caused by the current behaviour? If python 2 support ends up being killed just as a side-effect of a purely aesthetic change I'd be inclined to keep the status quo until dropping it becomes a necessity.

heinrich5991:
I very much doubt that Python 2 support works if it's not explicitly tested. And indeed, e.g. running the coverall test:

```
  File "/home/u/git/uniffi-rs/fixtures/coverall/tests/bindings/test_coverall.py", line 6, in <module>
    from datetime import datetime, timezone
ImportError: cannot import name timezone
test uniffi_foreign_language_testcase_test_coverall_py ... FAILED

failures:

---- uniffi_foreign_language_testcase_test_coverall_py stdout ----
Error: running `python2` failed
```

The [docs on `timezone`](https://docs.python.org/3/library/datetime.html#timezone-objects) say they're supported since Python 3.2.

Issue #1426, opened 2022-12-15
Allow things to be renamed in the bindings config


bendk:
We should allow functions, types, and other objects to be renamed in the bindings config, overriding the generated names.

The biggest use case for this is Swift, which doesn't support namespaces.  This means that if you UniFFI a collection of rust crates, the names of each type needs to be unique or else you can name collisions -- for example if 2 crates defined an `Error` type and you tried to import both generated modules.

This can be worked around by keeping each type name unique across all the UDL files, but that's not so idiomatic for languages whose modules form namespaces -- For example in Python `foo.Error` and `bar.Error` is usually more natural than `foo.FooError` and `bar.BarError`.

This means that the in many cases it would be good to allow types to share a name, but rename them for Swift to add a prefix.  I think the renames could be specified in `uniffi.toml` like this:

```
[bindings.swift.names]
SomeType = "RenamedType"
some_function = "renamed_function"
```

arg0d:
Our Apple team is also interesting in something like this. They would like to use Swift Uniffi interface throughout their code base. However, they have their own interface naming convention, so they would like generated Uniffi interfaces to be named according to their own interface naming convention. Right now the only way to accomplish this would be to fork Uniffi and change the interface naming in template files directly.

kriswuollett:
> I think the renames could be specified in `uniffi.toml` like this:
> 
> ```
> [bindings.swift.names]
> SomeType = "RenamedType"
> some_function = "renamed_function"
> ```

While I wouldn't want to discourage support the option of renaming individual types, the resulting config files may likely become very unwieldy if renaming things one at a time. I'd assume the most common case for Swift would be to namespace with a **per-crate prefix for everything** like back in the days of Objective-C. The issues with the above are:

1.  Need to fully qualify types potentially due to name conflicts with transitive dependencies, **and that** quoted identifiers/keys are needed since [punctuation isn't allowed in raw keys](https://github.com/toml-lang/toml/blob/dc5af251cd348ee053d9355359743767e3d0b123/toml.md?plain=1#L108), but this seems unavoidable in TOML,  e.g.:

```toml
[bindings.swift.names]
"my_crate::foo::SomeType" = "renamedType"
```

[...]

PR #1556, opened 2023-05-24
Improve error outputs for Python code

### LINKED ISSUE #1555: Error messages in Python are verbose
```
>>> h.decompress(b"")
[…]
libtw2_huffman.UniFFIExceptionTmpNamespace.InvalidInput: DecompressionError.InvalidInput('input is not a valid huffman compression')
```

The `UniFFIExceptionTmpNamespace` thing ideally wouldn't appear, the error message probably shouldn't repeat the error type, and Python errors containing a single string usually just print the message verbatim, without quoting.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-269)
┆Issue Number: UNIFFI-269

heinrich5991:
Fixes #1555.

Before:
```
>>> libtw2_huffman.decompress(b"")
[…]
libtw2_huffman.UniFFIExceptionTmpNamespace.InvalidInput: DecompressionError.InvalidInput('input is not a valid huffman compression')
```

After:
```
>>> libtw2_huffman.decompress(b"")
[…]
libtw2_huffman.DecompressionError.InvalidInput: input is not a valid huffman compression
```

### DIFF
--- a/fixtures/coverall/tests/bindings/test_coverall.kts
+++ b/fixtures/coverall/tests/bindings/test_coverall.kts
@@ -171,7 +171,7 @@ Coveralls("test_complex_errors").use { coveralls ->
         assert(e.code == 10.toShort())
         assert(e.extendedCode == 20.toShort())
         assert(e.toString() == "uniffi.coverall.ComplexException\$OsException: code=10, extendedCode=20") {
-            "Unexpected ComplexError.OsError.toString() value: ${e.toString()}"
+            "Unexpected ComplexException.OsError.toString() value: ${e.toString()}"
         }
     }
 
@@ -181,12 +181,21 @@ Coveralls("test_complex_errors").use { coveralls ->
     } catch(e: ComplexException.PermissionDenied) {
         assert(e.reason == "Forbidden")
         assert(e.toString() == "uniffi.coverall.ComplexException\$PermissionDenied: reason=Forbidden") {
-            "Unexpected ComplexError.PermissionDenied.toString() value: ${e.toString()}"
+            "Unexpected ComplexException.PermissionDenied.toString() value: ${e.toString()}"
         }
     }
 
     try {
         coveralls.maybeThrowComplex(3)
+        throw RuntimeException("Expected method to throw exception")
+    } catch(e: ComplexException.UnknownException) {
+        assert(e.toString() == "uniffi.coverall.ComplexException\$UnknownException: ") {
+            "Unexpected ComplexException.UnknownException.toString() value: ${e.toString()}"
+        }
+    }
+
+    try {
+        coveralls.maybeThrowComplex(4)
         throw RuntimeException("Expected method to throw exception")
     } catch(e: InternalException) {
         // Expected result
--- a/uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
@@ -3,47 +3,44 @@
 # {{ type_name }}
 # We want to define each variant as a nested class that's also a subclass,
 # which is tricky in Python.  To accomplish this we're going to create each
-# class separated, then manually add the child classes to the base class's
+# class separately, then manually add the child classes to the base class's
 # __dict__.  All of this happens in dummy class to avoid polluting the module
 # namespace.
-class UniFFIExceptionTmpNamespace:
-    class {{ type_name }}(Exception):
-        pass
-    {% for variant in e.variants() %}
-    {%- let variant_type_name = variant.name()|class_name %}
+class {{ type_name }}(Exception):
+    pass
 
+UniFFITemp{{ type_name }} = {{ type_name }}
+
+class {{ type_name }}:
+    {%- for variant in e.variants() -%}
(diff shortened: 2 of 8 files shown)


uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -3,47 +3,42 @@
...
+# class separately, then manually add the child classes to the base class's
 # __dict__.  All of this happens in dummy class to avoid polluting the module
 # namespace.
-class UniFFIExceptionTmpNamespace:
-    class {{ type_name }}(Exception):
-        pass
+class {{ type_name }}(Exception):
+    pass
+
+{{ type_name }}Temp = {{ type_name }}
```
bendk, line 12: - I think this name should be deleted at the end of the definitions to avoid polluting the namespace
- The name should be something that's guaranteed not to clash with any other user names.  Users may define a `Foo` and `FooTemp` class.  It's not 100% foolproof, but I usually use a prefix that starts with "uniffi" (maybe `UniFFiTemp{{ type_name }}` in this case.

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -3,47 +3,42 @@
...
-                '{{ field.name()|var_name }}={!r}'.format(self.{{ field.name()|var_name }}),
-                {%- endfor %}
-            ]
-            return "{{ type_name }}.{{ variant_type_name }}({})".format(', '.join(field_parts))
-            {%- else %}
-            return "{{ type_name }}.{{ variant_type_name }}()"
-            {%- endif %}
     {%- endif %}
 
     {{ type_name }}.{{ variant_type_name }} = {{ variant_type_name }}
```
bendk, line 38: I think this should assign to `{{ type_name}}Temp.{{ variant_type_name }}` (or whatever temp name we end up going with).  That way users can do use `FooError.Bar` to refer to the variant, which matches the Rust syntax.

...Or maybe we should just give up on this altogether.  It feels more Pythonic to make the variants top-level classes.  The one issue there is that 2 errors could have variants with the same name, but maybe #1426 could address that.

fixtures/coverall/tests/bindings/test_coverall.kts
```diff
@@ -181,13 +181,22 @@ Coveralls("test_complex_errors").use { coveralls ->
     } catch(e: ComplexException.PermissionDenied) {
         assert(e.reason == "Forbidden")
         assert(e.toString() == "uniffi.coverall.ComplexException\$PermissionDenied: reason=Forbidden") {
-            "Unexpected ComplexError.PermissionDenied.toString() value: ${e.toString()}"
+            "Unexpected ComplexException.PermissionDenied.toString() value: ${e.toString()}"
```
bendk, line 184: This seems like an improvement but I don't understand how the Kotlin message changed without any changes to `bindings::kotlin`.

bendk (changes requested):

bendk:
Thank you for giving this a pass.  I'm not sure what the final product should be, but I don't think the current code is correct and I'm glad someone is pushing for changes.

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
heinrich5991: Did both.

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
heinrich5991: > I think this should assign to `{{ type_name}}Temp.{{ variant_type_name }}` (or whatever temp name we end up going with). That way users can do use `FooError.Bar` to refer to the variant, which matches the Rust syntax.

Hmm, interesting, why did it work before? Apparently the new class's name isn't in scope when declaring the class (which is probably the reason for this whole construct), so this still worked.

I changed it anyway, for clarity.

fixtures/coverall/tests/bindings/test_coverall.kts
heinrich5991: This just changed the test's error message to match the rest of the code. Note that the same error is also called `ComplexException` a line above and also three lines above.

uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -3,47 +3,42 @@
...
-            {%- if variant.has_fields() %}
-            field_parts = [
-                {%- for field in variant.fields() %}
-                '{{ field.name()|var_name }}={!r}'.format(self.{{ field.name()|var_name }}),
-                {%- endfor %}
-            ]
-            return "{{ type_name }}.{{ variant_type_name }}({})".format(', '.join(field_parts))
-            {%- else %}
-            return "{{ type_name }}.{{ variant_type_name }}()"
-            {%- endif %}
```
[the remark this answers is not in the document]
heinrich5991: Done.

bendk (approved):
Looks great, thanks for this.

PR #1560, opened 2023-05-25
Add support for field defaults to proc-macro frontend

jplatte:
Based on #1559.

### DIFF
--- a/docs/manual/src/proc_macro/index.md
+++ b/docs/manual/src/proc_macro/index.md
@@ -104,6 +104,12 @@ will fail).
 pub struct MyRecord {
     pub field_a: String,
     pub field_b: Option<Arc<MyObject>>,
+    // Fields can have a default value.
+    // Currently, only string, integer, float and boolean literals are supported as defaults.
+    #[uniffi(default = "hello")]
+    pub greeting: String,
+    #[uniffi(default = true)]
+    pub some_flag: bool,
 }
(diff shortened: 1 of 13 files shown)


fixtures/metadata/src/tests.rs
```diff
@@ -173,10 +174,14 @@ mod test_metadata {
...
                         name: "name".into(),
                         ty: Type::String,
+                        default: Some(Literal::Str {
+                            value: "test".to_owned(),
+                        }),
                     },
                     FieldMetadata {
                         name: "age".into(),
                         ty: Type::U16,
+                        default: None,
```
mhammond, line 184: missed opportunity here for `default: Default::default()` :)

docs/manual/src/proc_macro/index.md
```diff
@@ -104,6 +104,12 @@ will fail).
 pub struct MyRecord {
     pub field_a: String,
     pub field_b: Option<Arc<MyObject>>,
+    // Fields can have a default value.
+    // Currently, only string, integer, float and boolean literals are supported as defaults.
```
mhammond, line 108: I wonder if it is worth adding something like "these work the same way as documented for UDL files" or similar (the docs need a restructure, but udl/structs.md does describe this in some detail)

mhammond (approved):
👍

Issue #1531, opened 2023-05-02
Can't use procmacro derived types as external types


mhammond:
I started looking at this but need to put it down for a while - so throwing this up in case anyone can offer clues.

With the patch at the end, I get:

    % cargo test -p uniffi-fixture-ext-types
    error[E0119]: conflicting implementations of trait `uniffi::FfiConverter<UniFfiTag>` for type `uniffi_one::UniffiOneProcMacroEnum`
       --> /Users/skip/src/moz/uniffi-rs/target/debug/build/uniffi-fixture-ext-types-23c75fdeb14707b4/out/ext-types-lib.uniffi.rs:446:1
        |
    446 | ::uniffi::ffi_converter_forward!(r#UniffiOneProcMacroEnum, ::r#uniffi_one::UniFfiTag, crate::UniFfiTag);
        | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        |
        = note: conflicting implementation in crate `uniffi_one`:
                - impl<T> FfiConverter<T> for UniffiOneProcMacroEnum;
        = note: this error originates in the macro `::uniffi::ffi_converter_forward` (in Nightly builds, run with -Z macro-backtrace for more info)I doubt this is limited to enums, but this is as far as I got.

```diff
diff --git a/fixtures/ext-types/lib/src/ext-types-lib.udl b/fixtures/ext-types/lib/src/ext-types-lib.udl
index 78d7c81b..6fe169bd 100644
--- a/fixtures/ext-types/lib/src/ext-types-lib.udl
+++ b/fixtures/ext-types/lib/src/ext-types-lib.udl
@@ -26,6 +26,9 @@ typedef extern UniffiOneType;
 // An enum in the same crate
 [External="uniffi-one"]
 typedef extern UniffiOneEnum;
+// An enum derived via a procmacro in that same crate
+[External="uniffi-one"]
+typedef extern UniffiOneProcMacroEnum;
 // An interface in the same crate
 [ExternalInterface="uniffi-one"]
 typedef extern UniffiOneInterface;
diff --git a/fixtures/ext-types/lib/src/lib.rs b/fixtures/ext-types/lib/src/lib.rs
index c8de7b72..f9f57e1e 100644
--- a/fixtures/ext-types/lib/src/lib.rs
[...]

PR #1561, opened 2023-05-25
Adding some documentation for external types

bendk:
These are some small changes related to a couple PRs:

 I was hoping to implement the system described in  https://github.com/mozilla/uniffi-rs/issues/1531, but I realized that
 wouldn't be easy for UDL files.  Currently, any `record` or `enum`  defined in a UDL file can be a remote type and there's no distiction  made between the two.  I don't want to put in the work to change that, so I just tried to document things and explain the differences.

 I realized that we introduced a couple breaking changes in corner cases with external types and documented them:
   - We discussed this some in https://github.com/mozilla/uniffi-rs/pull/1516#discussion_r1198960326, but I realize that the `ExternalTypes.rs` template also uses the external type crate name.  So I decided that it does deserve a line in the changelog and that line can be worded a bit simpler since it applies to current usage, not only to users that switch to library mode.
   - I realized we added the requirement for exteranl types type in the crate root when I tried using the new UniFFI code with app-services.

### DIFF
--- a/uniffi_macros/src/util.rs
+++ b/uniffi_macros/src/util.rs
@@ -212,6 +212,26 @@ mod kw {
 
 #[derive(Default)]
 pub(crate) struct CommonAttr {
+    /// Specifies the `UniFfiTag` used when implementing `FfiConverter`
+    ///   - When things are defined with proc-macros, this is `None` which means create a blanket
+    ///     impl for all types.
+    ///   - When things are defined with UDL files this is `Some(crate::UniFfiTag)`, which means only
+    ///     implement it for the local tag in the crate
+    ///
+    /// The reason for this split is remote types, i.e. types defined in remote crates that we
+    /// don't control and therefore can't define a blanket impl on because of the orphan rules.
+    ///
+    /// With UDL, we handle this by only implementing `FfiConverter<crate::UniFfiTag>` for the
+    /// type.  This gets around the orphan rules since a local type is in the trait, but requires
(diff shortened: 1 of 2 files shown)


uniffi_macros/src/util.rs
```diff
@@ -212,6 +212,26 @@ mod kw {
 
 #[derive(Default)]
 pub(crate) struct CommonAttr {
+    /// Specifies the `UniFfiTag` used when implementing `FfiConverter`
+    ///   - When things are defined in the proc-macros, this is `None` which means create a blanket
+    ///     impl for all types.
```
jplatte, line 217: This is not entirely true, I think right now it's still allowed to specify a tag when using the derive. I can post a PR to remove that possibility though.

bendk: Please do, I was wondering about that.

uniffi_macros/src/util.rs
```diff
@@ -212,6 +212,26 @@ mod kw {
 
 #[derive(Default)]
 pub(crate) struct CommonAttr {
+    /// Specifies the `UniFfiTag` used when implementing `FfiConverter`
+    ///   - When things are defined in the proc-macros, this is `None` which means create a blanket
```
mhammond, line 216: Not sure what terminology we are using for this in general, but I'd suggest s/in the/with/ ?

mhammond (approved):

PR #1562, opened 2023-05-25
macros: Disallow specifying tag type for derives

jplatte:
As [requested](https://github.com/mozilla/uniffi-rs/pull/1561#discussion_r1205719813).

Should we also enforce that the tag *is* specified for the `ffi_converter_*` attribute macros?


bendk (approved):
Looks good.

> Should we also enforce that the tag is specified for the ffi_converter_* attribute macros?

I don't feel strongly either way since those are not intended to be public facing.  Let's merge this one and we can consider doing that as a follow-up.

PR #1564, opened 2023-05-29
Clarify the semantics of Enums and Errors

mhammond:
The `is_flat` semantics for enums and errors was confusing me (mainly because that's entirely the wrong term which should be used), so I tried to capture the semantics and expand some comments on it.

It mainly adds additional coverall tests for enums and error, and adds matching enum and error definitions via proc_macros, and ensures the semantics are the same.

### DIFF
--- a/uniffi_bindgen/src/interface/enum_.rs
+++ b/uniffi_bindgen/src/interface/enum_.rs
@@ -92,7 +92,23 @@ use super::{APIConverter, ComponentInterface};
 pub struct Enum {
     pub(super) name: String,
     pub(super) variants: Vec<Variant>,
-    // "Flat" enums do not have variants with associated data.
+    // NOTE: `flat` is a misleading name and to make matters worse, has 2 different
+    // meanings depending on the context :(
+    // * When used as part of Rust scaffolding generation, it means "is this enum
+    //   used with an Error, and that error should we lowered to foreign bindings
+    //   by converting each variant to a string and lowering the variant with that
+    //   string?". In that context, it should probably be called `lowered_as_string` or
+    //   similar.
+    // * When used as part of bindings generation, it means "does this enum have only
+    //   variants with no associated data"? The foreign binding generators are likely
+    //   to generate significantly different versions of the enum based on that value.
+    //
+    // The reason it is described as "has 2 different meanings" by way of example:
+    // * For an Enum described as being a flat error, but the enum itself has variants with data,
+    //   `flat` will be `true` for the Enum when generating scaffolding and `false` when
+    //   generating bindings.
+    // * For an Enum not used as an error but which has no variants with data, `flat` will be
+    //   false when generating the scaffolding but `true` when generating bindings.
     pub(super) flat: bool,
 }
 
@@ -156,7 +172,9 @@ impl APIConverter<Enum> for weedle::EnumDefinition<'_> {
                     })
                 })
                 .collect::<Result<Vec<_>>>()?,
-            // Enums declared using the `enum` syntax can never have variants with fields.
+            // Enums declared using the `enum` syntax can never have variants with fields described in the UDL.
+            // So regardless of whether this enum has variants with fields or not, we lower it with
+            // a string representation of itself.
             flat: true,
         })
     }
--- a/uniffi_macros/src/error.rs
+++ b/uniffi_macros/src/error.rs
@@ -79,7 +79,8 @@ fn error_ffi_converter_impl(ident: &Ident, enum_: &DataEnum, attr: &ErrorAttr) -
 
 // FfiConverters for "flat errors"
 //
-// These are errors where we only expose the variants, not the data.
+// These are errors where we only lower the to_string() value, rather than any assocated data.
+// We lower the to_string() value unconditionally, whether the enum has associated data or not.
 fn flat_error_ffi_converter_impl(
     ident: &Ident,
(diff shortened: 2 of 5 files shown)


uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -92,7 +92,16 @@ use super::{APIConverter, ComponentInterface};
...
+    // NOTE: `flat` is a misleading name and to make matters worse, has 2 different
+    // meanings depending on the context :(
+    // * When used as part of Rust scaffolding generation, it means "is this enum
+    //   used with an Error, and that error should we lowered to foreign bindings
+    //   by converting each variant to a string and lowering the variant with that
+    //   string?". In that context, it should probably be called `lowered_as_string` or
+    //   similar.
+    // * When used as part of bindings generation, it means "does this enum have only
+    //   variants with no associated data"? The foreign binding generators are likely
+    //   to generate significantly different versions of the enum based on that value.
```
bendk, line 104: The scaffolding part seems right to me.  I'm not sure that it has a different meaning for the bindings generation though.  The way I think of it, the bindings generation lifts the string lowered by Rust and uses that as the error message, rather than lifting the associated data and storing that.

uniffi_macros/src/error.rs
```diff
@@ -79,7 +79,7 @@ fn error_ffi_converter_impl(ident: &Ident, enum_: &DataEnum, attr: &ErrorAttr) -
 
 // FfiConverters for "flat errors"
 //
-// These are errors where we only expose the variants, not the data.
+// These are errors where we only expose to_string of the variants, not the data.
```
bendk, line 82: This one still seems a bit unclear to me.  How about "These are errors where we only lower the `to_string()` value, rather than any assocated data".

bendk (changes requested):
This could definitely be made more clear, thanks for writing this up.  I had a few suggestions on the docs, although it's hard to come up with any description since the concept itself is a bit confusing.

uniffi_bindgen/src/interface/enum_.rs
mhammond: Thanks - I tried to clarify what I meant here, which is that the exact same enum will have a different value for `flat` when used to generate scaffolding than when used to generate bindings.

mhammond:
> Alternatively, we could consider dropping is_flat support. The only drawback that I can see is that it would require consumers to update their UDL files in some cases. Are there other drawbacks?

These are used today primarily to allow errors with variants which aren't supported by uniffi - eg, `OsError` or similar.

In fact I think we should go the other way - support is_flat for regular enums too, which the same basic semantics as for errors - it would mean that enum can only be lowered into foreign bindings (which is a restriction that already exists for error enums), but that might still be a reasonable option for some use-cases.

uniffi_bindgen/src/interface/enum_.rs
bendk: This is really weird indeed.  How does the `flat` value change depending on the context?  Doesn't the same UDL-parsing happen in both cases?

mhammond: Yes and no - for bindings, it goes via uniffi_macros and metadata. In particular, [these semantics](https://github.com/mozilla/uniffi-rs/blob/154739df1867d619ad257d3c02b289c23427fb9f/uniffi_bindgen/src/interface/enum_.rs#L127) don't match what UDL parsing does

jplatte:
> Alternatively, we could consider dropping `is_flat` support. The only drawback that I can see is that it would require consumers to update their UDL files in some cases. Are there other drawbacks?

I'd be favor of doing this, eventually. Currently this is a very convenient way of handling errors, and I think there's no clear replacement right now, but I think we might be able to figure something out (in particular, interface-types-as-errors + an exported `Display` impl should help a lot).

bendk:
> These are used today primarily to allow errors with variants which aren't supported by uniffi - eg, `OsError` or similar.
> 
> In fact I think we should go the other way - support is_flat for regular enums too, which the same basic semantics as for errors - it would mean that enum can only be lowered into foreign bindings (which is a restriction that already exists for error enums), but that might still be a reasonable option for some use-cases.

I keep thinking that custom types could be the solution here.  What if you used custom types to either:
  - Add a custom type that lowers `OsError` to a unit struct or some other placeholder value
  - Add a custom type that lowers the error enum to a different one which has the same variant names, but doesn't have fields that can't be lowered.

In both cases, the Rust `try_lift`/`try_read` methods would panic, just like how flat errors can't be passed from the foreign side back to Rust.  Maybe we could add some way to configure the custom type so that UniFFI would refuse to generate code that tried to do that.

If this worked, it seems like a nice generalization.  It works for for errors, enums, and also records.  It also doesn't allows you to avoid lowering some fields, but still lower others.

bendk:
> (in particular, interface-types-as-errors + an exported `Display` impl should help a lot)

I totally agree, especially the recent changes to allow exporting `Display`.  Half of what `flat_error` did was work around fields that can't be lowered.  The other half was to allow the errors to still have custom messages without those fields.

mhammond:
While these discussions are valuable, I'd still like to land this PR - it doesn't change any semantics, so will help ensure that any semantics we *do* change are done intentionally.

bendk (approved):
Landing this PR seems good.  I think it accurately describes the current behavior.

PR #1568, opened 2023-05-31
Support Into::into for function results to be consistent with errors

mhammond:
Uniffi [already supports mapping "internal" errors to public ones](https://github.com/mozilla/uniffi-rs/blob/755049ebaa967fdfbb192d5b9c111353a7fee99f/fixtures/coverall/src/lib.rs#L23-L37). This PR does the same thing for function results.

### DIFF
--- a/fixtures/coverall/src/coverall.udl
+++ b/fixtures/coverall/src/coverall.udl
@@ -6,6 +6,11 @@ namespace coverall {
 
     sequence<TestTrait> get_traits();
 
+    MaybeSimpleDict get_maybe_simple_dict(i8 index);
+    MaybeSimpleDict get_maybe_simple_internal_dict(string? text);
+    [Throws=CoverallError]
+    MaybeSimpleDict fallible_get_maybe_simple_internal_dict(string? text);
+
     // void returning error throwing namespace function to catch clippy warnings (eg, #1330)
     [Throws=CoverallError]
     void println(string text);
@@ -88,6 +93,10 @@ interface Coveralls {
 
     [Throws=CoverallError]
     void fallible_panic(string message);
+
+    MaybeSimpleDict get_maybe_simple_internal_dict(string? text);
+    [Throws=CoverallError]
+    MaybeSimpleDict fallible_get_maybe_simple_internal_dict(string? text);
 
     // *** Test functions which take either `self` or other params as `Arc<Self>` ***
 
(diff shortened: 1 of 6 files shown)


jplatte:
I've noticed that UniFFI *sometimes* does this for errors (though not consistently, semi-recently I saw a few functions break on upgrade because they returned `anyhow::Error` when the UDL said they have some uniffi-compatible error type). I'm not a really a fan. But this only affects UDL anyways, since with the proc-macros there is nowhere to put the "actual" type that is used for FFI conversion, right?

mhammond:
> I've noticed that UniFFI _sometimes_ does this for errors (though not consistently, semi-recently I saw a few functions break on upgrade because they returned `anyhow::Error` when the UDL said they have some uniffi-compatible error type).

I'd love to see an example of that, because that is exactly the use-case I'm working towards - the Rust API returns `anyhow::Error()` but the error exposed to consumers is something that anyhow can `into()` to. I've actually got that seamlessly working for interfaces described in UDL and am working towards proc-macro support.

> I'm not a really a fan tbh. But this only affects UDL anyways, since with the proc-macros there is nowhere to put the "actual" type that is used for FFI conversion, right?

My intention is to do procmacros very soon, probably via an attribute. I actually started on that for errors (for the reason I noted above), but then thought I might as well do the same thing for actual return values at the same time, so rounded back to the UDL first.

mhammond:
(To be clear, I'm working towards a function with an `anyhow::Result` being able to have the `Err` turn into an *interface* on the foreign side, with the interface having methods like cause, backtrace, etc)

jplatte:
> (To be clear, I'm working towards a function with an `anyhow::Result` being able to have the `Err` turn into an _interface_ on the foreign side, with the interface having methods like cause, backtrace, etc)

Can we discuss this in an issue or on Matrix?

mhammond:
@jplatte and I had a discussion in matrix, where we agreed this might not be the best mechanism to use for this, particularly for procmacros. However, I think we do broadly agree that the same mechanism should be used for errors as for non-error return values, and because we already have `into` support for errors, I think this PR has value as it aligns error and non-error return values using the existing mechanism. This should also make any future changes to the mechanism easier to rationalize about. @jplatte, do you agree with that assessment?

jplatte:
Yeah, makes sense. I just want to note that we don't have `into` support for errors with the proc-macros, and I'd prefer to explore other solutions for proc-macros in general (not just for errors) before "porting" the `into` support to proc-macros.

PR #1572, opened 2023-06-01
Check for integer overflows in Ruby bindings

heinrich5991:
The exception thrown is [`RangeError`](https://docs.ruby-lang.org/en/master/RangeError.html), like e.g. when doing `256.chr`.

```
> 256.chr
(irb):1:in `chr': 256 out of char range (RangeError)
        from (irb):1:in `<main>'
        from /usr/lib/ruby/gems/3.0.0/gems/irb-1.4.2/exe/irb:11:in `<top (required)>'
        from /usr/bin/irb:25:in `load'
        from /usr/bin/irb:25:in `<main>'
```

### DIFF
--- a/uniffi_bindgen/src/bindings/ruby/templates/Helpers.rb
+++ b/uniffi_bindgen/src/bindings/ruby/templates/Helpers.rb
@@ -1,0 +1,5 @@
+def uniffi_in_range(i, type_name, min, max)
+  i = i.to_i
+  raise RangeError, "#{type_name} requires #{min} <= value < #{max}" unless (min <= i && i < max)
+  i
+end
--- a/uniffi_bindgen/src/bindings/ruby/templates/Helpers.rb
+++ b/uniffi_bindgen/src/bindings/ruby/templates/Helpers.rb
@@ -1,4 +1,4 @@
-def uniffi_in_range(i, type_name, min, max)
+def self.uniffi_in_range(i, type_name, min, max)
   i = i.to_i
   raise RangeError, "#{type_name} requires #{min} <= value < #{max}" unless (min <= i && i < max)
   i
--- a/fixtures/type-limits/tests/bindings/test_type_limits.rb
+++ b/fixtures/type-limits/tests/bindings/test_type_limits.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+
+# This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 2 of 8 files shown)


heinrich5991:
The method `uniffi_in_range` is currently placed in the top-level, which is probably not desirable. I couldn't figure out how to move it to the `Uniffi<ModuleName>` because it has to be accessible from the outside, too. I couldn't name the `Uniffi<ModuleName>` from `fn coerce_rb`, so I couldn't just place it in there.

Any ideas how this could be solved?

heinrich5991:
@saks Perhaps you have an idea to the above question? The matrix channel told me that you're the one who authored these bindings.

saks:
@heinrich5991 , thanks for reaching out. I'll get back to you in a few hours.

saks:
@heinrich5991 , thanks for reaching out. I'll get back to you in a few hours.

I like the idea to put this method into some module, the one defined as uniffi namespace should do. I've created this PR (https://github.com/heinrich5991/uniffi-rs/pull/1/files) to illustrate the idea. Please have a look and let me know

uniffi_bindgen/src/bindings/ruby/templates/Helpers.rb
```diff
@@ -0,0 +1,7 @@
+def uniffi_in_range(i, type_name, min, max)
+    i = i.to_i
```
badboy, line 2: side-note: Ruby typically uses 2-space indentation.

heinrich5991:
@saks
> I like the idea to put this method into some module, the one defined as uniffi namespace should do. I've created this PR (https://github.com/heinrich5991/uniffi-rs/pull/1/files) to illustrate the idea. Please have a look and let me know

Thanks for the PR, I added the commit.

uniffi_bindgen/src/bindings/ruby/templates/Helpers.rb
```diff
@@ -0,0 +1,7 @@
...
+    if not (min <= i and i < max)
+        raise RangeError.new "#{type_name} requires #{min} <= value < #{max}"
```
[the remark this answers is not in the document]
heinrich5991: Done.

uniffi_bindgen/src/bindings/ruby/templates/Helpers.rb
heinrich5991: Done.

badboy (approved):

PR #1573, opened 2023-06-02
Proc macro callbacks

bendk:
This took a lot longer than I expected, but I think it's working now.  I ended up refactoring a lot of related code, hopefully for the better but tell me if I made anything worse.


docs/manual/src/proc_macro/index.md
```diff
@@ -228,6 +228,56 @@ fn do_http_request() -> Result<(), MyApiError> {
...
+To avoid panics, callback interfaces can use `Result<T, E>` types for all return values.  If the callback
+interface implementation throws the exception that corresponds to the `E` parameter, `Err(E)` will
+be returned to the Rust code.  However, in most languages it's still possible for the implementation
+to throw other exceptions.  To avoid panics in those cases, the error type must be wrapped
+with the `#[uniffi(handle_unknown_callback_error)]` attribute and
+`From<UnexpectedUniFFICallbackError>` must be implemented:
+
+```rust
+#[derive(uniffi::Error)]
+#[uniffi(handle_unknown_callback_error)]
```
bendk, line 266: I don't love this name, but couldn't think of anything better.  Definitely open to suggestions on this one.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -218,6 +218,42 @@ impl ComponentInterface {
         self.errors.get(name)
     }
 
+    /// Get an Enum definition by name or panic
+    pub fn get_enum_definition_unchecked(&self, name: &str) -> &Enum {
```
jplatte, line 224: "unchecked" is generally used for unsafe functions that have undefined behavior in the unexpected case. But maybe it's fine to use the same naming here.

uniffi_macros/src/export/item.rs
```diff
@@ -82,9 +83,14 @@ impl ExportItem {
 
                 let attrs = ExportedImplFnAttributes::new(&impl_fn.attrs)?;
                 let item = if attrs.constructor {
-                    ImplItem::Constructor(ConstructorSignature::new(impl_fn.sig)?)
+                    let sig = FnSignature::new_constructor(self_ident.clone(), impl_fn.sig)?;
+                    if sig.is_async {
+                        return sig.syn_err("Async constructors are not supported");
+                    }
+                    ImplItem::Constructor(sig)
```
jplatte, line 90: Why is the `is_async` check not part of `new_constructor`? Also, having a `syn_err` method seems kind of weird to me.

CHANGELOG.md
```diff
@@ -33,15 +33,13 @@
...
-
 - Traits can be exposed as a UniFFI `interface` by using a `[Trait]` attribute in the UDL.
   See [the documentation](https://mozilla.github.io/uniffi-rs/udl/interfaces.html#exposing-traits-as-interfaces).
-
 - The `bytes` primitive type was added, it represents an array of bytes. It maps to `ByteArray` in Kotlin, `bytes` in Python, `String` with `Encoding::BINARY` in Ruby and `Data` in Swift.
 - Shortened `str()` representations of errors in Python to align with other exceptions in Python. Use `repr()` or the `{!r}` format to get the old representation back.
-
 - Methods implemented by standard Rust traits, such as `Debug`, `Display`, `Eq` and `Hash` can now be exposed over the FFI and bindings may implement special methods for them.
   See [the documentation](https://mozilla.github.io/uniffi-rs/udl/interfaces.html#exposing-methods-from-standard-rust-traits).
+- Implemented proc-macro callback inteface support
```
jplatte, line 42: typo: inteface -> interface

docs/manual/src/proc_macro/index.md
```diff
@@ -228,6 +228,56 @@ fn do_http_request() -> Result<(), MyApiError> {
 }
 ```
 
+## The `#[uniffi::callback_interface]` attribute
+
+`#[uniffi::callback_interface]` can be used to export a [callback interface](../udl/callback_interfaces.html) definition
```
jplatte, line 233: Missing period at the end of the sentence.

docs/manual/src/proc_macro/index.md
```diff
@@ -228,6 +228,56 @@ fn do_http_request() -> Result<(), MyApiError> {
...
+This allows the foreign bindings to implement the interface and pass an instance to the Rust code.
+
+```rust
+#[uniffi::callback_interface]
+pub trait Person {
```
jplatte, line 238: Hm, why a separate macro? I think `#[uniffi::export]` or (`#[uniffi::export(callback_interface)]` if trait declarations can have a different meaning to UniFFI) would be more "natural", but of course that's very subjective.

docs/manual/src/proc_macro/index.md
```diff
@@ -228,6 +228,56 @@ fn do_http_request() -> Result<(), MyApiError> {
...
+Most languages allow arbitrary exceptions to be thrown, which presents issues for callback
+interfaces.  If a callback interface function returns a non-Result type, then any exception will
+result in a panic on the Rust side.
+
+To avoid panics, callback interfaces can use `Result<T, E>` types for all return values.  If the callback
+interface implementation throws the exception that corresponds to the `E` parameter, `Err(E)` will
+be returned to the Rust code.  However, in most languages it's still possible for the implementation
+to throw other exceptions.  To avoid panics in those cases, the error type must be wrapped
+with the `#[uniffi(handle_unknown_callback_error)]` attribute and
+`From<UnexpectedUniFFICallbackError>` must be implemented:
```
jplatte, line 262: Should this really be opt-in? I think an opt-out solution might be better.

docs/manual/src/proc_macro/index.md
jplatte: If we make it opt-out instead of opt-in, we could call it `panic_on_unknown_error`.

docs/manual/src/proc_macro/index.md
bendk: This was a very incisive comment and made me rethink a lot of how the code was organized.  I updated the code so not only does it share the same macro name, but the underlying uniffi macro code works more similarly.  In particular, callback interface traits and non-callback interface traits work very similarly which I think is a big win.

docs/manual/src/proc_macro/index.md
bendk: For callback interface errors I agree, but some (most?) errors are never used in callback interfaces and making them opt-out doesn't seem correct to me.

jplatte: Right, of course.

uniffi_macros/src/enum_.rs
```diff
@@ -202,3 +211,25 @@ pub fn variant_metadata(enum_: &DataEnum) -> syn::Result<Vec<TokenStream>> {
...
+        quote! {
+            fn handle_callback_unexpected_error(e: ::uniffi::UnexpectedUniFFICallbackError) -> Self {
+                <Self as ::std::convert::From<::uniffi::UnexpectedUniFFICallbackError>>::from(e)
+            }
+        }
+    } else {
+        // Use the default function
+        quote! {}
+    }
+}
```
jplatte, line 235: `Option<TokenStream>` can be interpolated the same way as `TokenStream`, so no code outside this function has to change if you use `bool::then` here:

```suggestion
) -> Option<TokenStream> {
    handle_unknown_callback_error.then(|| quote! {
        fn handle_callback_unexpected_error(e: ::uniffi::UnexpectedUniFFICallbackError) -> Self {
            <Self as ::std::convert::From<::uniffi::UnexpectedUniFFICallbackError>>::from(e)
        }
    })
}
```

(I think the comment in the else branch isn't really needed, the same thing is stated at the end of the doc comment)

uniffi_macros/src/export/callback_interface.rs
```diff
@@ -0,0 +1,134 @@
...
+    items: &[ImplItem],
+) -> syn::Result<TokenStream> {
+    let trait_name = ident_to_string(trait_ident);
+    let trait_impl_methods = items
+        .iter()
+        .map(|item| match item {
+            ImplItem::Method(sig) => gen_method_impl(sig, internals_ident),
+            _ => unreachable!("traits have no constructors"),
+        })
+        .collect::<syn::Result<Vec<_>>>()?;
```
jplatte, line 28:
```suggestion
        .collect::<syn::Result<TokenStream>>()?;
```

uniffi_macros/src/export/callback_interface.rs
```diff
@@ -0,0 +1,134 @@
...
+                #internals_ident.invoke_callback::<(), crate::UniFfiTag>(
+                    self.handle, uniffi::IDX_CALLBACK_FREE, Default::default()
+                )
+            }
+        }
+
+        ::uniffi::deps::static_assertions::assert_impl_all!(#ident: Send);
+
+        impl #trait_ident for #ident {
+            #(#trait_impl_methods)*
```
jplatte, line 54:
```suggestion
            #trait_impl_methods
```

uniffi_macros/src/export/callback_interface.rs
```diff
@@ -0,0 +1,134 @@
...
+            handle: u64,
+        }
+
+        impl #ident {
+            fn new(handle: u64) -> Self {
+                Self { handle }
+            }
+        }
+
+        impl Drop for #ident {
```
jplatte, line 43:
```suggestion
        impl ::std::mem::Drop for #ident {
```

uniffi_macros/src/export/callback_interface.rs
```diff
@@ -0,0 +1,134 @@
...
+            let mut #buf_ident = ::std::vec::Vec::new();
+            #(#write_exprs;)*
+        }
+    } else {
+        quote! { let #buf_ident = ::std::vec::Vec::new(); }
+    };
+
+    Ok(quote! {
+        fn #ident(&self, #(#params),*) -> #return_ty {
+            #construct_args_buf
```
jplatte, line 103:
```suggestion
    let mut write_exprs = sig.write_exprs(&buf_ident);

    Ok(quote! {
        fn #ident(&self, #(#params),*) -> #return_ty {
            #[allow(unused_mut)]
            let mut #buf_ident = ::std::vec::Vec::new();
            #(#write_exprs;)*
```

jplatte (approved):

PR #1574, opened 2023-06-02
Fix intermittent failures in the futures Python fixture

mhammond:
I can't reproduce the kotlin failure we sometimes see on CI, but Python regularly fails for me. The test seems to assume that sleeping for X ms means we will wake up almost exactly after X ms, which isn't really how things work. There are still a couple of questionable tests left IMO, but didn't want to go *too* far - LMK if you think I should just kill the ones I flagged as being suspect too, or we can just wait and kill them if they remain intermittent on CI?

### DIFF
--- a/fixtures/futures/tests/bindings/test_futures.py
+++ b/fixtures/futures/tests/bindings/test_futures.py
@@ -9,25 +9,13 @@ def now():
 class TestFutures(unittest.TestCase):
     def test_always_ready(self):
         async def test():
-            t0 = now()
-            result = await always_ready()
-            t1 = now()
-
-            t_delta = (t1 - t0).total_seconds()
-            self.assertTrue(t_delta < 0.1)
-            self.assertEqual(result, True)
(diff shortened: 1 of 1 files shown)


fixtures/futures/tests/bindings/test_futures.py
```diff
@@ -14,7 +14,8 @@ async def test():
             t1 = now()
 
             t_delta = (t1 - t0).total_seconds()
-            self.assertTrue(t_delta < 0.1)
+            # XXX - this is suspect - nothing guarantees this!
+            self.assertLess(t_delta, 0.1)
```
tarikeshaq, line 18: Yeah I agree, let's kill the assertions that seem like they could cause intermittent failures again

tarikeshaq (approved):
LGTM with a comment, generally cautious of using time based tests but the remaining ones seem easy to convince myself that they won't be intermittent, and we're testing the mechanism itself here

PR #1577, opened 2023-06-02
Added warning about Send + Sync and callback interfaces

bendk:
When working on proc-macros I thought about this and I'm pretty sure it's an issue.  Am I missing something here or is this something we should have been warning about from the start?

### DIFF
--- a/docs/manual/src/udl/callback_interfaces.md
+++ b/docs/manual/src/udl/callback_interfaces.md
@@ -23,15 +23,20 @@ pub trait Keychain: Send + Sync + Debug {
 }
 ```
 
-### Why Send + Sync?
+### Send + Sync?
 
-The concrete types that UniFFI generates for callback interfaces implement `Send`, `Sync`, and `Debug`, so it's safe to
+The concrete types that UniFFI generates for callback interfaces implement `Send`, `Sync`, and `Debug`, so it's legal to
 include these as supertraits of your callback interface trait.  This isn't strictly necessary, but it's often useful.  In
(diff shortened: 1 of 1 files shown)


bendk:
> This seems an improvement, but I think that last sentence seems a big wrong - IIUC, their callback implementation could _logically_ require `Send + Sync` but Rust isn't going to insist on it. I wonder if you should hand-wave with something like "... even though Rust will never insist on the implementation being `Send + Sync`, you might still have a contract with the foreign binding which means it must _logically_ be `Send + Sync`. You must make sure to inform your library consumers of any such requirements"?

That's what I was trying to get at, but I don't think I made it clear on the first pass.  What do you think of the updated language?

mhammond:
I think the new wording is great. But I think the point I was trying to make is that the wording:

> If you add `Send + Sync` to your callback interface

seems odd - nothing will ever insist they do, and nothing will change if they do, right? It might be that I'm missing something though.

bendk:
> seems odd - nothing will ever insist they do, and nothing will change if they do, right? It might be that I'm missing something though.

The paragraph above is suggesting that it's something you might want to do and gives reasons for it.  Maybe I should change the wording to "If you add `Send + Sync` as supertraits to your callback interface Trait"

mhammond:
I think what I'm struggling with is that I'm not sure at what point Rust is going to insist on `Send + Sync`, or at which point the foreign implementation must be conceptually `Send + Sync` but nothing is supply guard-rails. But now I'm not even sure what I'm talking about :) Landing what you have is fine!

PR #1579, opened 2023-06-04
Fix custom type demo usage of deprecated URL constructor in kotlin

mhammond:
Somehow I've ended up with java 20 or something, and this example is failing due to deprecated use of the url constructor. Hopefully this new way works with the older versions.


stan-irl:
im also hitting this. using jenv to override the java version doesnt fix it either

bendk (approved):

PR #1587, opened 2023-06-05
Re-introduce distinction between CodeType and CodeTypeDispatch

mhammond:
This fixes a source of significant confusion in the code-oracle support.

On one hand, it has the concept of a `CodeType` which you get back from an Oracle via a `CodeDispatch` - so a code-type is closely tied to the oracle that supplied it. But:

* Every `CodeType` method takes a `CodeOracle` param - which is odd because the `CodeType` came from the oracle - it doesn't need the oracle to work.

* There's a blanket implementation of `CodeType` made available for `Type` - so not only do we now have a `CodeType` that didn't come from an Oracle, so clearly can't know the language semantics the Oracle abstracts away, there's now *two* implementations of `CodeType` for each type, with each behaving differently.

This cleans all that up:

* CodeType is only supplied by Oracles - there's no blanket implementation for Type.

* Therefore `CodeType` no longer takes Oracle's as args.

* Filters now take a `CodeTypeDispatch` - which *are* implemented for type, and explicitly use the oracle to get a `CodeType`

### DIFF
--- a/uniffi_bindgen/src/backend/types.rs
+++ b/uniffi_bindgen/src/backend/types.rs
@@ -11,50 +11,40 @@
 //! (the FfiType), and foreign language expressions that call into the machinery. This helper code
 //! can be provided by a template file.
 //!
-//! A `CodeDeclaration` is needed for each type that is declared in the UDL file. This has access to
-//! the [crate::interface::ComponentInterface], which is the closest thing to an Intermediate Representation.
-//!
-//! `CodeDeclaration`s provide the target language's version of the UDL type, including forwarding method calls
-//! into Rust. It is likely if you're implementing a `CodeDeclaration` for this purpose, it will also need to cross
-//! the FFI, and you'll also need a `CodeType`.
-//!
-//! `CodeDeclaration`s can also be used to conditionally include code: e.g. only include the CallbackInterfaceRuntime
-//! if the user has used at least one callback interface.
-//!
-//! Each backend has a wrapper template for each file it needs to generate. This should collect the `CodeDeclaration`s that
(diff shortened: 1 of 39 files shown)


uniffi_bindgen/src/backend/types.rs
```diff
@@ -65,46 +55,46 @@ pub trait CodeType {
...
+    fn initialization_fn(&self) -> Option<String> {
         None
     }
 }
 
 /// This trait is used to implement `CodeType` for `Type` and type-like structs (`Record`, `Enum`, `Field`,
 /// etc).  We forward all method calls to a `Box<dyn CodeType>`, which we get by calling
 /// `CodeOracle.find()`.
-pub trait CodeTypeDispatch {
+pub trait CodeTypeDispatch: Debug {
```
bendk, line 97: I would love to replace this trait with something like:

```
pub trait AsType {
   fn as_type(&self) -> Type;
}
```

The having filters that looked like:
```
pub fn type_name(t: &impl AsType) -> Result<String, askama::Error> {
    Ok(oracle.find(t.as_type()).type_label())
}
```

`AsType` is more general than `CodeTypeDispatch` and overall the code seems simpler to me.  Maybe this should be a separate PR though.

bendk (approved):
These changes look good to me.  Maybe we should consider the `AsType` suggestion here, but I'm also fine with merging this one and discussing that separately.

uniffi_bindgen/src/backend/types.rs
mhammond: That's actually a great idea and I agree it's much cleaner - I put up #1588, let me know if that's not what you had in mind.

mhammond:
Done in #1588

PR #1588, opened 2023-06-07
Cleanup the CodeOracle by removing CodeTypeDispatch entirely

mhammond:
Over in #1587 Ben had what I think's a great idea, so here it is. I put it as a new PR/branch just incase we decided to land the other one anyway.

### DIFF
--- a/uniffi_bindgen/src/interface/record.rs
+++ b/uniffi_bindgen/src/interface/record.rs
@@ -52,7 +52,7 @@ use super::{
     convert_type,
     literal::{convert_default_value, Literal},
 };
-use super::{APIConverter, ComponentInterface};
+use super::{APIConverter, AsType, ComponentInterface};
 
 /// Represents a "data class" style object, for passing around complex values.
 ///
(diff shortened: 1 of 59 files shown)


uniffi_bindgen/src/interface/types/mod.rs
```diff
@@ -252,6 +252,28 @@ impl From<&&Type> for FfiType {
...
 }
 
+// A trait so various things can turn into a type.
+pub trait AsType: core::fmt::Debug {
+    fn as_type(&self) -> Type;
+}
+
+impl AsType for Type {
+    fn as_type(&self) -> Type {
+        self.clone()
```
bendk, line 262: This is a drawback to this approach, but a really minor one.  I think one of the reasons for the initial `CodeDispatch` system was that I was too afraid of the cloning.

uniffi_bindgen/src/interface/record.rs
```diff
@@ -141,6 +137,12 @@ impl Field {
     }
 }
 
+impl AsType for Field {
+    fn as_type(&self) -> Type {
+        self.type_.clone()
+    }
+}
```
bendk, line 144: I'll just mention that another way to do this would be `impl Into<Type> for &Field`, but that doesn't feel like it's in the spirit of `Into` to me.  I like a separate trait better.

bendk (approved):
Looks great to me.  The new system seems simpler overall and we even are net decreasing our LOC count.  I didn't inspect all the template changes since they seemed very mechanical.

PR #1591, opened 2023-06-10
Remove the CodeOracle

mhammond:
This is now purely a local consideration for bindings, there's no need for a shared trait. This removes a lot of unnecessary code.

I also snuck in some Kotlin improvements which are a good example of why not having the shared trait makes sense. It's probably easiest/best to look at the 2 commits individually.


jplatte:
Haven't looked at the diff, but this sounds great! I previously found the amount of indirection / abstraction hard to grok, and jump-to-definition bringing me to a trait method declaration instead of an implementation was a big part of that.

bendk:
I'm mixed on this one.  The nice thing about the shared code was that it kept the backends more consistent.  For example, the Kotlin backend is now going to handle the enums differently from all the other ones.  OTOH, I do agree that the indirection was hard to follow.  Also, it can be easier to evolve the code if you don't have to change all backends at once.

Overall I think I'm for, but I'd like to hear from @jhugman who understands the vision more.

mhammond:
>  The nice thing about the shared code was that it kept the backends more consistent.

I think it's worth pointing out that the current code is setup such that the `CodeOracle` never actually escapes the binding (ie, no implementations outside the bindings ever take an Oracle as an arg etc).

So now, when a binding has a new requirement, it has 2 choices:
* Add a new method to `CodeOracle` which is specific to the binding, but force all other bindings to learn about it, reinforcing the fact that the `CodeOracle` is really just a union of requirements from all bindings.
* Ignore it completely, just add a new method to the impl of the code oracle and call it directly instead of by the trait

(which is exactly what I've done in #1592)

IMO, artificially trying to force bindings to use a trait that's entirely unused, unnecessary and which adds significant complexity, all in the interests of consistency seems the wrong tradeoff.

> Also, it can be easier to evolve the code if you don't have to change all backends at once.

Yeah, this is the big win IMO, especially for external bindings. Ruby never used the code oracle, which is the reason it hasn't needed as much churn as the other bindings recently.

jhugman:
The point of the oracle was to reduce the mega-`match`es over the `Type` enum, by putting the lookup to an object that could do the `lift`/`lower` for each variant.

I think the mistake was to put the `CodeType` methods (e.g. `type_label`) onto the oracle. This seemed to be a cause of too much indirection, and over-coupled the oracle to `CodeType`.

This patch seems to move the lookup functionality of the oracle into the `AsCodeType` trait, and correct that mistake.

So, all in all, this PR looks good to me.

Should there be a different or the same `CodeType` trait per language? Does it drive forward all the backends at a time? Is it the cause of the dreaded churn? I don't know. I'm not sure.

I think consistency of implementation in the binding backends should be prized: adding the same thing to Kotlin as you've just added to Swift has been a lot easier because they look the same.

jhugman (approved):

Issue #1553, opened 2023-05-23
`uniffi_foreign_language_testcase_test_futures` test seems flaky


heinrich5991:
Two failed test runs in an unrelated PR:
- [Kotlin](https://app.circleci.com/pipelines/github/mozilla/uniffi-rs/2906/workflows/eccdbbb2-63f8-4bda-aeaf-fec5d02f471a/jobs/11685): timeout
- [Python](https://app.circleci.com/pipelines/github/mozilla/uniffi-rs/2906/workflows/4213abe8-25dd-4950-887c-5a3b32c9acbb/jobs/11678): "FATAL: exception not rethrown"

JssDWt:
Can confirm this flakiness occurs in our CI too.
Did you manage to address the issue in your CI somehow?

PR #1597, opened 2023-06-15
Add type checking to strings/bytes for Python/Ruby

heinrich5991:
Also fixes that bytes had to be valid UTF-8 due to #1595.


heinrich5991:
Test failure is the unrelated #1553.

uniffi_bindgen/src/bindings/python/templates/BytesHelper.py
```diff
@@ -8,5 +8,9 @@ def read(buf):
 
     @staticmethod
     def write(value, buf):
+        try:
```
mhammond, line 11: Why not `isinstance()` here? I assume it's because you want to support anything exposing the buffer protocol - so why do bytes get "anything bytes-like" but strings aren't "anything string-like"?

fixtures/type-limits/tests/bindings/test_type_limits.py
```diff
@@ -195,11 +195,48 @@ def test_special_floats(self):
...
 
+    def test_non_string(self):
+        self.assertRaises(TypeError, lambda: take_string(None))
+        self.assertRaises(TypeError, lambda: take_string(False))
+        self.assertRaises(TypeError, lambda: take_string(True))
+        self.assertRaises(TypeError, lambda: take_string(0))
+        self.assertRaises(TypeError, lambda: take_string(0.0))
+        self.assertRaises(TypeError, lambda: take_string(b""))
+
+        class A:
```
mhammond, line 206: It's not clear to me what the current semantics are - does this work without this patch? If so, what's the motivation for preventing this from working?

mhammond (commented):
I like the idea of better test coverage, but I'm not quite clear on some of the decisions made (and sorry if I missed a discussion elsewhere, please point me to it!)

fixtures/type-limits/tests/bindings/test_type_limits.py
mhammond: (I mean, I guess "does the object have `__str__`?" isn't really a great indicator of "this object wants to be seen as string like", but my question is really more about "do we want to exclude objects which want to be string like?"

heinrich5991: > It's not clear to me what the current semantics are - does this work without this patch?

I didn't change the current semantics except for throwing `TypeError` instead of `AttributeError`. So no, it does not work without this patch.

In this way, this test is trying to test for the current behavior.

heinrich5991:
> but I'm not quite clear on some of the decisions made (and sorry if I missed a discussion elsewhere, please point me to it!)

You haven't missed a discussion on this. I probably should have expanded on the choices I made in the commit message.

mhammond:
Thanks for the explanation. I'm fine with this, but I think you should wait for @badboy to say he is :)

PR #1598, opened 2023-06-15
Python: For eq/ne handle types differing from the one we're implementing

badboy:
`NotImplemented` is what these methods should return, see https://docs.python.org/3/library/constants.html#NotImplemented The interpreter is then responsible for doing the right thing.

Unfortunately this means "manually" implementing these methods in the template instead of relying on the convenient macros.


mhammond (approved):
oops, I should read the docs more :)

badboy:
> oops, I should read the docs more :)

I didn't know either until just now and initially implemented them to just return true or false for those other cases.

PR #1599, opened 2023-06-15
Format Python names according to PEP8

### LINKED ISSUE #1557: Python modules export a lot of implementation details
For a simple library with just two functions, I get the following tab-completion:

```
Python 3.11.3 (main, Apr  5 2023, 15:52:25) [GCC 12.2.1 20230201] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import libtw2_huffman
>>> libtw2_huffman.
libtw2_huffman.DEFAULT
libtw2_huffman.DecompressionError(
libtw2_huffman.DecompressionErrorTemp(
libtw2_huffman.FOREIGN_CALLBACK_T(
libtw2_huffman.FfiConverterBytes()
libtw2_huffman.FfiConverterPrimitive()
libtw2_huffman.FfiConverterPrimitiveInt()
libtw2_huffman.FfiConverterRustBuffer()
libtw2_huffman.FfiConverterString()
libtw2_huffman.FfiConverterTypeDecompressionError()
libtw2_huffman.ForeignBytes(
libtw2_huffman.InternalError(
libtw2_huffman.Path(
libtw2_huffman.RustBuffer(
libtw2_huffman.RustBufferBuilder()
libtw2_huffman.RustBufferStream(
libtw2_huffman.RustCallStatus(
libtw2_huffman.UNIFFI_FOREIGN_EXECUTOR_CALLBACK_T(
libtw2_huffman.UNIFFI_RUST_TASK(
libtw2_huffman.UniFfiPointerManager()
libtw2_huffman.UniFfiPointerManagerCPython()
libtw2_huffman.UniFfiPointerManagerGeneral()
libtw2_huffman.compress(
libtw2_huffman.contextlib
libtw2_huffman.contextvars
libtw2_huffman.ctypes
libtw2_huffman.datetime
libtw2_huffman.decompress(
libtw2_huffman.enum
libtw2_huffman.loadIndirect()
libtw2_huffman.os
libtw2_huffman.platform
libtw2_huffman.rust_call(
libtw2_huffman.rust_call_async(
libtw2_huffman.rust_call_with_error(
libtw2_huffman.struct
libtw2_huffman.sys
libtw2_huffman.uniffi_check_api_checksums(
libtw2_huffman.uniffi_check_call_status(
libtw2_huffman.uniffi_check_contract_api_version(
libtw2_huffman.uniffi_future_callback_t(
```

This could be fixed by prefixing all internal symbols with `_` (marking them "private" by Python standards, see [PEP 8](https://peps.python.org/pep-0008/#method-names-and-instance-variables)) or by adding another module reexporting only the relevant parts of the library.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-270)
┆Issue Number: UNIFFI-270

heinrich5991:
All internal symbols now start with `_Uniffi` or `_uniffi` so that name clashes are unlikely. Additionally, this means that they are no longer included in typical autocompletion. This is quite a heavy-handed approach because it requires all changes to the Python bindings to be aware that names have to start with an underscore.

Fixes #1557.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/gen_python/mod.rs
+++ b/uniffi_bindgen/src/bindings/python/gen_python/mod.rs
@@ -303,17 +303,17 @@ impl PythonCodeOracle {
             FfiType::Float64 => "ctypes.c_double".to_string(),
             FfiType::RustArcPtr(_) => "ctypes.c_void_p".to_string(),
             FfiType::RustBuffer(maybe_suffix) => match maybe_suffix {
-                Some(suffix) => format!("RustBuffer{suffix}"),
-                None => "RustBuffer".to_string(),
+                Some(suffix) => format!("_UniffiRustBuffer{suffix}"),
+                None => "_UniffiRustBuffer".to_string(),
             },
-            FfiType::ForeignBytes => "ForeignBytes".to_string(),
-            FfiType::ForeignCallback => "FOREIGN_CALLBACK_T".to_string(),
+            FfiType::ForeignBytes => "_UniffiForeignBytes".to_string(),
+            FfiType::ForeignCallback => "_UNIFFI_FOREIGN_CALLBACK_T".to_string(),
             // Pointer to an `asyncio.EventLoop` instance
             FfiType::ForeignExecutorHandle => "ctypes.c_size_t".to_string(),
-            FfiType::ForeignExecutorCallback => "UNIFFI_FOREIGN_EXECUTOR_CALLBACK_T".to_string(),
+            FfiType::ForeignExecutorCallback => "_UNIFFI_FOREIGN_EXECUTOR_CALLBACK_T".to_string(),
             FfiType::FutureCallback { return_type } => {
                 format!(
-                    "uniffi_future_callback_t({})",
+                    "_uniffi_future_callback_t({})",
                     Self::ffi_type_label(return_type),
                 )
             }
@@ -385,7 +385,7 @@ pub mod filters {
     }
 
     pub fn ffi_converter_name(as_ct: &impl AsCodeType) -> Result<String, askama::Error> {
-        Ok(as_ct.as_codetype().ffi_converter_name())
+        Ok(String::from("_Uniffi") + &as_ct.as_codetype().ffi_converter_name()[3..])
     }
 
     pub fn canonical_name(as_ct: &impl AsCodeType) -> Result<String, askama::Error> {
@@ -419,7 +419,7 @@ pub mod filters {
             None => "void".into(),
         };
         Ok(format!(
-            "uniffi_async_callback_{return_string}__{throws_string}"
+            "_uniffi_async_callback_{return_string}__{throws_string}"
         ))
     }
 
--- a/uniffi_bindgen/src/bindings/python/templates/ForeignExecutorTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/ForeignExecutorTemplate.py
@@ -3,33 +3,33 @@
 {{ self.add_import("asyncio") }}
 
 class {{ ffi_converter_name }}:
-    _pointer_manager = UniFfiPointerManager()
+    _pointer_manager = _UniffiPointerManager()
 
     @classmethod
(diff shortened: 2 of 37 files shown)


heinrich5991:
@mhammond I'd also like to hear your opinion on this.

heinrich5991:
Tab completion now (I assume it also works in PyCharm):
```py
>>> import uniffi_type_limits
>>> uniffi_type_limits.
uniffi_type_limits.InternalError(  uniffi_type_limits.take_i8(
uniffi_type_limits.take_f32(       uniffi_type_limits.take_string(
uniffi_type_limits.take_f64(       uniffi_type_limits.take_u16(
uniffi_type_limits.take_i16(       uniffi_type_limits.take_u32(
uniffi_type_limits.take_i32(       uniffi_type_limits.take_u64(
uniffi_type_limits.take_i64(       uniffi_type_limits.take_u8(
```

uniffi_bindgen/src/bindings/python/templates/ForeignExecutorTemplate.py
```diff
@@ -1,35 +1,35 @@
 # FFI code for the ForeignExecutor type
 
-{{ self.add_import("asyncio") }}
```
mhammond, line 3: I think the `add_import` machinery should continue to be used, but more generally, this doesn't really seem idiomatic to me - I've almost never seen people rename every single import they do to support this use-case. Same as for `ctypes`. If every single module designed to be used by other modules was forced to have *every single* import be of the form `import mod as _mod` I think we'd all go a little crazy.

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -378,7 +378,7 @@ pub mod filters {
     }
 
     pub fn ffi_converter_name(as_ct: &impl AsCodeType) -> Result<String, askama::Error> {
-        Ok(as_ct.as_codetype().ffi_converter_name())
+        Ok(String::from("_Uniffi") + &as_ct.as_codetype().ffi_converter_name()[3..])
```
mhammond, line 381: IIUC this is transforming, say, `FfiConverterFoo` to `_UniffiConverterFoo`, right? In some ways I kinda prefer `UniffiFfiConverter` - it's mildly redundant, but it means the fact it's an `ffi converter` is more obvious, and best I can tell it probably doesn't hurt, right?

Alternatively, it's not particularly important for everything to have a `_Uniffi` prefix is it? eg, `_rust_call` functions below didn't get one - they just got a leading `_` - so is there any reason `_FfiConverterFoo` isn't just as good for the problem being solved here? eg, I had to go down a strange path to work out why the `3..` existed here, whereas just tacking on a leading `_` seems far less likely to have people scratching their head.

mhammond (commented):
I'm OK with this - I think I'd prefer to have less magic in the ffi converter but really not too bothered (although if we stick with the magic `3..` I think a comment makes sense.

However, I'd still like to see how @badboy feels before mashing the "approve" button, but I approve in principle :)

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
heinrich5991: > eg, I had to go down a strange path to work out why the `3..` existed here, whereas just tacking on a leading `_` seems far less likely to have people scratching their head.

Ah yea, that's a hack. I guess I should fix that before merging…

> eg, `_rust_call` functions below didn't get one - they just got a leading `_`

Ah, I probably just missed those… Having everything get a `uniffi` prefix makes it quite clear that nothing will conflict and also sets a standard for future symbols, I'd call that a win. I guess I'll add `uniffi` to those symbols as well.

mhammond:
Sorry for the delay - do you mind updating the conflicts and we can get this merged.

heinrich5991:
> Sorry for the delay - do you mind updating the conflicts and we can get this merged.

Rebased.

mhammond (approved):

PR #1600, opened 2023-06-15
Add proc-macro external type support (#1531)

bendk:
This allows proc-macro based libraries to use types defined in UDL files.  It also paves the way for proc-macros to support remote types from non-UniFFI types.

I'm hoping we can sneak this in to the `0.24` release.

### DIFF
--- a/fixtures/metadata/src/tests.rs
+++ b/fixtures/metadata/src/tests.rs
@@ -132,12 +132,15 @@ mod test_type_ids {
     #[test]
     fn test_user_types() {
         check_type_id::<Person>(Type::Record {
+            module_path: "uniffi_fixture_metadata".into(),
             name: "Person".into(),
         });
         check_type_id::<Weapon>(Type::Enum {
+            module_path: "uniffi_fixture_metadata".into(),
             name: "Weapon".into(),
         });
         check_type_id::<Arc<Calculator>>(Type::ArcObject {
+            module_path: "uniffi_fixture_metadata".into(),
             object_name: "Calculator".into(),
             is_trait: false,
         });
@@ -243,6 +246,7 @@ mod test_metadata {
                         fields: vec![FieldMetadata {
                             name: "result".into(),
                             ty: Type::Record {
+                                module_path: "uniffi_fixture_metadata".into(),
                                 name: "Person".into(),
                             },
                             default: None,
@@ -303,6 +307,7 @@ mod test_metadata {
                             fields: vec![FieldMetadata {
                                 name: "weapon".into(),
                                 ty: Type::Enum {
+                                    module_path: "uniffi_fixture_metadata".into(),
                                     name: "Weapon".into(),
                                 },
                                 default: None,
@@ -399,12 +404,14 @@ mod test_function_metadata {
                     FnParamMetadata {
                         name: "person".into(),
                         ty: Type::Record {
+                            module_path: "uniffi_fixture_metadata".into(),
                             name: "Person".into(),
                         },
                     },
                     FnParamMetadata {
                         name: "weapon".into(),
                         ty: Type::Enum {
+                            module_path: "uniffi_fixture_metadata".into(),
                             name: "Weapon".into(),
                         },
                     },
@@ -443,9 +450,11 @@ mod test_function_metadata {
                 is_async: false,
                 inputs: vec![],
                 return_type: Some(Type::Enum {
+                    module_path: "uniffi_fixture_metadata".into(),
                     name: "State".into(),
                 }),
                 throws: Some(Type::Enum {
+                    module_path: "uniffi_fixture_metadata".into(),
                     name: "FlatError".into(),
                 }),
                 checksum: UNIFFI_META_CONST_UNIFFI_FIXTURE_METADATA_FUNC_TEST_FUNC_THAT_THROWS
@@ -465,6 +474,7 @@ mod test_function_metadata {
                 inputs: vec![],
                 return_type: None,
                 throws: Some(Type::Enum {
+                    module_path: "uniffi_fixture_metadata".into(),
                     name: "FlatError".into(),
                 }),
                 checksum:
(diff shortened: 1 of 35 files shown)


fixtures/proc-macro/application-services-m2.tar.gz
```diff
@@ -0,0 +1,15 @@
+{
```
badboy, line 1: Accidentally included?

uniffi_macros/src/util.rs
```diff
@@ -211,6 +211,20 @@ pub(crate) fn tagged_impl_header(
...
+    Path {
+        leading_colon: None,
+        segments: [
+            quote::format_ident!("crate"),
+            quote::format_ident!("UniFfiTag"),
+        ]
+        .into_iter()
+        .map(PathSegment::from)
+        .collect(),
+    }
```
jplatte, line 225: Could just do `syn::parse_quote!(crate::UniFfiTag)`.

uniffi_macros/src/record.rs
```diff
@@ -55,6 +55,10 @@ pub(crate) fn record_ffi_converter_impl(
 ) -> TokenStream {
     let impl_spec = tagged_impl_header("FfiConverter", ident, tag);
     let name = ident_to_string(ident);
+    let mod_path = match mod_path() {
+        Ok(p) => p,
+        Err(e) => return e.into_compile_error(),
+    };
```
jplatte, line 61: Maybe make this function return `syn::Result<TokenStream>` instead and handle `into_compile_error` one level higher? Then it should just be a matter of `.unwrap_or_else(syn::Error::into_compile_error)`.

stan-irl:
@bendk I just tried to build with this branch and hit `unsupported type for error: External` when generating kotlin: https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs#L238

bendk:
@stan-irl 

Thanks for testing this out with some real-world examples.  It's really nice to get concrete feedback.

- I believe the error issue is a general one.  I got the same issue with UDL files and opened #5679 for it.  If possible, I'll try to get a fix in for that one soon after this is merged.
- I believe the issue with duplicate definitions is solved by https://github.com/mozilla/uniffi-rs/pull/1404, but that's gotten fairly bitrotted and I need to do a rebase pass on it.
- I just pushed a commit to fix the visibility issue with `lift`.

stan-irl:
thankyou!!

fixtures/metadata/src/tests.rs
```diff
@@ -132,12 +132,15 @@ mod test_type_ids {
     #[test]
     fn test_user_types() {
         check_type_id::<Person>(Type::Record {
+            crate_name: "uniffi_fixture_metadata".into(),
```
mhammond, line 135: I think uniffi meta calls the same concept `module_path`, which seems yet more general than this - any reason to not use that same concept - IIUC that would mean we just use a name from a sub-module, whereas `crate_name` implies only the root?

uniffi_meta/src/child_types.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::Type;
+
+/// Iterate over all child types as mutable references
+///
+/// This is used to find and convert external type
+pub trait ChildTypesMut {
+    fn child_types_mut(&mut self) -> Vec<&mut Type>;
+}
+
+impl ChildTypesMut for crate::Metadata {
```
mhammond, line 14: I've nearly turned [`TypeIterator`](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/interface/types/mod.rs#L127) into a trait before, which would make even more sense here I think? Is there any reason to not do that here? ie, define the trait as a more general `IterTypes`and later we might let it be used in more places (eg, in that example above)

Although really, I don't quite understand how this differs from the above example? I guess because we don't actually have a struct to represent `Type::External`?

(hrmph - the fact this is `mut` also seems odd. I'll look at this in more detail again soon)

mhammond (commented):
I don't understand how a future multi-crate world looks:

* my module_path comment before is likely wrong - I guess it assumes a module *inside* a crate - ie, the top segment of the path probably is not a crate name.
* However, ISTM that a ComponentInterface has always been a representation of a crate, and the `MetadataGroup` struct reinforces that - it's a kind of "here are all the items in this crate/namespace, please explode a CI up.
* Or maybe not? It's adding them to a CI, not creating a new one, in which case this patch does make some sense, although ISTM that the various Metadata items and the uniffi_bindgen::interface:: structs would *also* need them at some point?
* And it further means simple name resolution can't be done with just a name - ie, you can't reject 2 interfaces with the name `Foo` unless they are in the same crate - which isn't handled here (but you *would* want to reject builtin names without considering the module name, and presumably you would *not* want these types duplicated per crate)

uniffi_meta/src/lib.rs
```diff
@@ -235,19 +238,24 @@ pub enum Type {
     ForeignExecutor,
     SystemTime,
     Enum {
+        crate_name: String,
         name: String,
     },
     Record {
+        crate_name: String,
```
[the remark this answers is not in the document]
bendk: Changing `crate_name` to `module_path` makes sense to me.

I think it's reasonable to encode the crate_name/module path in the `Type` metadata is that metadata is shared between all crates.  Functions from one crate should see that type as a `Type::Record`, while functions from another crate should see it as a `Type::External`.

However, maybe a better system would be to lean into the concept of type identifiers and only store a path to the type definition.  I'm picturing something like:
  - There's a `TypeDefinition` enum with variants for record, object, etc.
  - There's a `TypeDefinitionUniverse` that maps full module paths, including the crate name, to `TypeDefinition`s
  - The `TYPE_RECORD`, `TYPE_ENUM`, etc. constants that are used for function signature arguments are replaced with `TYPE_USER`, which is followed by a module path string that keys into `TypeDefinitionUniverse`.
  - All of this can be used to construct a `ComponentInterface` for each crate

Side note: currently `module_path` is just a crate name, because of some reasons that I forgot.  I believe it should be possible to make it a full path though.

bendk:
@mhammond I rebased and implemented the changes that I could remember us discussing.  Is there anything more to do for this PR?  (I'll make sure to squash before merging).

mhammond (approved):

PR #1602, opened 2023-06-16
Test string edge cases for Kotlin and Swift

heinrich5991:


heinrich5991:
I don't like the behavior around lone surrogates in Kotlin:
```kt
assert(takeString("\ud800") == "?") // surrogate
```

I think this behavior is default in Java/Kotlin though. I'd prefer to throw an error. What do you think?

mhammond:
> I think this behavior is default in Java/Kotlin though. I'd prefer to throw an error. What do you think?

A simple kotlin version of this is something like:
```
fun takeStringNative(s: String): String {
    return s
}

val got = takeStringNative("\ud800");
```

right? For me, that returns the surrogate as-is, and not an ascii '?'. Python, Ruby and Swift all reject the string entirely, and the Rust docs imply it should too - at least for `char` - https://doc.rust-lang.org/std/primitive.char.html

Can you help explain what's going on here for Kotlin, as I'm really not sure :)

heinrich5991:
I meant to say that this is the expected behavior when transcoding between character sets and encountering invalid sequences. Java/Kotlin usually replace the faulty characters instead of throwing an exception.

> Python, Ruby and Swift all reject the string entirely

Python does not reject this string but it throws when trying to convert it to UTF-8. Ruby, Swift and Rust do indeed not support lone surrogates (which is the correct way IMO).

> Can you help explain what's going on here for Kotlin, as I'm really not sure :)

We're calling the transcoding function [String.getBytes](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#getBytes-java.nio.charset.Charset-) (through some abstractions). That function (IMO) represents the way Java usually deals with invalid string encodings, it inserts replacement characters (in this case ASCII question marks). I'd prefer to throw an error if the input string is invalid.

heinrich5991:
@bendk Do you have an opinion on this?

bendk:
I'm not much of a Kotlin expert.  Maybe @jhugman?

jhugman:
> I don't like the behavior around lone surrogates in Kotlin:

What does Kotlin do natively? My preference would be that the embedding application language should not be surprised by the behaviour of the FFI:

i.e. 
```kt
assert((takeString("\ud800") == "?") == ("\ud800" == "?")) 
```

jhugman:
@heinrich5991 

> That's not possible to achieve.

Thanks for laying it out.

My preference, in that case, would be option 1.

heinrich5991:
@mhammond Should be ready for review now. I implemented @jhugman's (and my) preferred solution.

fixtures/type-limits/tests/bindings/test_type_limits.kts
```diff
@@ -0,0 +1,17 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import uniffi.uniffi_type_limits.*;
+
+// test_strings
+try {
+    takeString("\ud800")
+    throw RuntimeException("Should have thrown an CharacterCodingException exception!")
+    assert(takeString("") == "")
```
mhammond, line 11: it looks like this should be removed?

uniffi_bindgen/src/bindings/kotlin/templates/StringHelper.kt
```diff
@@ -38,6 +46,7 @@ public object FfiConverterString: FfiConverter<String, RustBuffer.ByValue> {
     }
 
     override fun write(value: String, buf: ByteBuffer) {
+        check(value)
```
mhammond, line 49: ouch - is there really no way to implement this other than to do one full pass over the string to check for the invalid sequence, then another to actually perform the encoding?

fixtures/type-limits/tests/bindings/test_type_limits.kts
heinrich5991: Done.

uniffi_bindgen/src/bindings/kotlin/templates/StringHelper.kt
```diff
@@ -38,6 +46,7 @@ public object FfiConverterString: FfiConverter<String, RustBuffer.ByValue> {
...
         val byteArr = value.toByteArray(Charsets.UTF_8)
```
linabutler, line 50: Thanks so much for putting this together, @heinrich5991! Let's validate and encode in a single pass, though, like @mhammond suggests.

Since you already have the encoder, you're almost there! I think we can do something like this to make it work:

```suggestion
        val byteArr = Charsets.UTF_8.newEncoder().run {
            onMalformedInput(java.nio.charset.CodingErrorAction.REPORT)
            encode(java.nio.CharBuffer.wrap(value)).array()
        }
```

(Let's also add imports for `java.nio.charset.CodingErrorAction` and `java.nio.CharBuffer` to `wrapper.kt`, so that we don't need to write `java.nio` every time).

Just to note for completeness, we could make this even more efficient by having the encoder write directly into the `RustBuffer`, instead of allocating and copying from `byteArr`. I _definitely_ don't think it's worth doing that now, though. We'd need to know the encoded size to allocate the `RustBuffer`, which means either `maxBytesPerChar()`—overallocating by 4x for almost every string—or `averageBytesPerChar()`, and reallocating if we run out of room.

But validating and encoding in one pass is an easier optimization to make now! 🚀

linabutler (changes requested):

uniffi_bindgen/src/bindings/kotlin/templates/StringHelper.kt
heinrich5991: Thanks for the help, I was a little stuck. The provided source fails for some non-ASCII strings, but it's fixed by respecting the returned `ByteBuffer`'s `limit()`.

uniffi_bindgen/src/bindings/kotlin/templates/StringHelper.kt
heinrich5991: Done, with the help of @linabutler.

linabutler (approved):
Fantastic! Thanks for fixing that up, and for catching the non-ASCII inputs—this looks great to me, and I really like that it aligns Kotlin's behavior with Rust's and our other target languages'.

mhammond (approved):
Thanks for your persistence @heinrich5991 and for your guidance @linabutler!

Issue #1608, opened 2023-06-19
Write some documentation for async/futures


badboy:
Looks like we only have examples, but nothing on https://mozilla.github.io/uniffi-rs/

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-287)
┆Issue Number: UNIFFI-287

mhammond:
> Looks like we only have examples

I don't think we even have examples - we have fixtures, but they aren't particularly approachable for understanding how async works here.

badboy:
> > Looks like we only have examples
> 
> I don't think we even have examples - we have fixtures, but they aren't particularly approachable for understanding how async works here.

`s/examples/tests` -- is what I meant. And yeah, they are definitely not good enough for users.

PR #1609, opened 2023-06-19
Update changelog with notable changes since the last release

badboy:
We might be missing some docs here and there, though that we can do in followups without blocking the release.

* [ ] https://github.com/mozilla/uniffi-rs/issues/1608

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@
 ### ⚠️ Breaking Changes ⚠️
 - ABI: Implemented a new callback-interface ABI that significantly improves performance on Python and Kotlin.
   - UniFFI users will automatically get the benefits of this without any code changes.
-  - External bindings authors will need to update their bindings code.  See PR #1494 for details.
-- ABI: Changed API checksum handling.  This affects external bindings authors who will need to update their code to work with the new system.  See PR #1469 for details.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -17,8 +17,8 @@
 ### ⚠️ Breaking Changes ⚠️
 - ABI: Implemented a new callback-interface ABI that significantly improves performance on Python and Kotlin.
   - UniFFI users will automatically get the benefits of this without any code changes.
-  - External bindings authors will need to update their bindings code.  See PR #1494 for details.
-- ABI: Changed API checksum handling.  This affects external bindings authors who will need to update their code to work with the new system.  See PR #1469 for details.
+  - External bindings authors will need to update their bindings code.  See [PR #1494](https://github.com/mozilla/uniffi-rs/pull/1494) for details.
```
mhammond, line 20: I know you only added the link, but I don't think that PR is particularly useful - nothing actually landed in that PR. Maybe you could add https://github.com/mozilla/uniffi-rs/commit/323a4976992aff207db7946fc1f1cea614838f46 (which is roughly the outcome of the discussion in that PR) to the list below and change this para to just say "please see "Guidance for external bindings below" here?

mhammond (approved):

CHANGELOG.md
badboy: yup, I asked in channel. That commit looks much better to link to indeed.

PR #1613, opened 2023-06-20
Add ADR-8 -Allowing more types as errors

mhammond:
[Rendered](https://github.com/mhammond/uniffi-rs/blob/adr-8/docs/adr/0008-allowing-more-types-as-errors.md)

@bendk and I agreed that assuming we want to support interfaces as errors we should probably try and form consensus around the semantics. Here's my attempt at documenting what I see as the options.

Please comment on whether you support this in general, and if so, which of the sub-options you prefer. Personally, I'm somewhere between (B) and (C).

### DIFF
--- a/docs/adr/0008-allowing-more-types-as-errors.md
+++ b/docs/adr/0008-allowing-more-types-as-errors.md
@@ -1,0 +1,142 @@
+# Allowing more types to be used as errors.
+
+* Status: proposed
+* Deciders: Hopefully consensus on [#1613](https://github.com/mozilla/uniffi-rs/pull/1613)
+
+## Current UniFFI error handling
+
+UniFFI functions currently must report error returns via an `Enum` - ie, the `E` in `Result<T, E>` must be an Enum.
+
+This ADR proposes moving towards a model where more types can be an `E`, and specfically, proposes allowing any uniffi "interface".
+
+Assuming this isn't controversial, it will also make decisions about how such errors should be represented in the foreign bindings we maintain.
+
+### The current situation
+When an `Enum` is used as an error, on the foreign side it's not an `Enum` at all - it's a distinct type -
+it derives from the local `Exception` class and isn't usable as the actual `Enum` behind it.
+
+### The options for interfaces.
+Consider the UDL
+```idl
+interface ErrorDetails { string message(); };
+[Throws=ErrorDetails] void do_it();
+``````
+and Rust
+```rust
+struct ErrorDetails {}
+impl ErrorDetails {
+    fn message(&self) -> String { todo!() }
+}
+fn do_it() -> Result<(), ErrorDetails> { Err(ErrorDetails {}) }
+```
+
+How should this be captured on the foreign side?
+
+### A) Create an "error only" type
+
+This means an interface used as an error behaves a lot like an enum used as en error. It means:
+* There's no "non error" version of the interface. (just like there's no non error version of an enum declared as an error)
+* A struct simply appearing as an error changes how the foreign side sees that struct significantly.
+* The type can't be used exactly as it could in non-error contexts.
+
+ie, it's a bit of "either or, but not both". But it makes for more natural foreign code:
+
+```
+try:
+    do_it()
+except ErrorDetailsException as e:
+    # There's no `ErrorDetails` type in this world.
+    # ie, `e' is an  `ErrorDetailsException`
+    # `ErrorDetailsException` derives from `Exception`
+    e.message()
+```
+
+### B) Create an original type and an error type
+
+We create 2 types - the `interface` itself unchanged from how the struct would be generated now,
+and another exception "wrapper" for the type.
+It means you *catch* a slightly different name from the interface, but the non-error version of the interface is what you deal with.
+
+```
+try:
+    do_it()
+except ErrorDetailsException as e:
+    # There are 2 types here:
+    # `e` is an `ErrorDetailsException`
+    # `e.inner` (or any other name) is an `ErrorDetails` - does not derive from `Exception`
+    # We can make it look natural via delegation.
+    e.message()
+    # But also use the original interface
+    HandleErrorReportingProblem(e.inner)
+```
+
+To clarify, you could imagine generated Python code looking something like:
+```
+class ErrorDetails:
+    ...
+class ErrorDetailsException(Exception):
+    ...
+```
+### C) Create and original type with an error type nested within the original
+
+Very similar to the above - there are still 2 unique types created, but only one type is a top-level
+type - the error type is created inside the exception.
+
+The code catching exceptions might look something like:
+```
+try:
+    do_it()
+except ErrorDetails.Exception as e:
+    # same block above as for B
+```
+
+ie, there's no top-level `ErrorDetailsException` type, it's accessed via `ErrorDetails.Exception`
+
+To clarify, you could imagine generated Python code looking something like:
+```
+class ErrorDetails:
+    # snip regular interface.
+    class Exception(Exception):
+        # snip error specific code
+```
+
+## Decision Outcome
+
+Chosen option: "A) Create an "error only" type" - it's more natural to foreign authors, and
+the use-cases and pro/con list can't identify a compelling benefit to having multiple objects.
+
+## Pros and Cons of the Options
+
+### 1: Do nothing - don't support interfaces as errors.
+
+Con: Seems a bit negative :)
+
+### 2: Implement Option A above
+
+- Pro: no magic, there's exactly 1 type for the type in the UDL/macro
+- Con: That type gets "tainted" by being used as an error. Using the interface in other contexts is less natural.
+- Con: It might be tricky to use as an external type - all external consumers must use as an error, or none of them must.
+
+### 3: Implement Option B above:
+
+- Pro: a more "pure" system - the interface can be used naturally in contexts other than being an error.
+- Pro: It mirrors Rust features in that it "adds" a capability rather then transforms it; it's more natural from a Rust viewpoint.
+- Con: It does not mirror the standard behavior of any of our bindings; it's less natural from a foreign viewpoint.
+- Con: The wrapping might get in the way, and it's more complexity for a possibly non-existing use-case.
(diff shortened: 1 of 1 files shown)


docs/adr/0008-allowing-more-types-as-errors.md
```diff
@@ -0,0 +1,139 @@
...
+## Pros and Cons of the Options
+
+### 1: Do nothing - don't support interfaces as errors.
+
+Con: Seems a bit negative :)
+
+### 2: Implement Option A above
+
+- Pro: no magic, there's exactly 1 type for the type in the UDL/macro
+- Con: That type gets "tainted" by being used as an error. Using the interface in other contexts is less natural.
```
bendk, line 116: I think this is both pro and con.  It's unnatural from the Rust POV, but natural from the foreign binding POV.

docs/adr/0008-allowing-more-types-as-errors.md
```diff
@@ -0,0 +1,139 @@
+# Allowing more types to be used as errors.
+
+* Status: proposed
+* Deciders: Hopefully consensus on #1613
+
+## Current UniFFI error handling
+
+UniFFI functions currently must report error returns via an `Enum` - ie, the `E` in `Result<T, E>` must be an Enum.
+
+This ADR proposes moving towards a model where any type could be an `E`, and specfically, proposes allowing any uniffi "interface"
```
bendk, line 10: Maybe `s/any type could/more types can/`.  I don't think we ever want ints or floats to be an error type.

docs/adr/0008-allowing-more-types-as-errors.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+How should this be captured on the foreign side?
+
+### A) Create an "error only" type
+
+This is like the status quo re enums. It means:
+* There's no "non error" version of the interface.
+* A struct simply appearing as an error changes how the foreign side sees that struct significantly.
+* The type can't be used exactly as it could in non-error contexts.
```
bendk, line 40: I don't think this is a requirement.  It seems possible to allow these types to be returned, inputted as arguments, etc.  It also seems possible to let the foreign code call into their API outside of a `catch` block.

I think the real issue is the bullet above.  It would be very weird for the foreign code to have a type that can be used both as a regular object and an exception.

docs/adr/0008-allowing-more-types-as-errors.md
```diff
@@ -0,0 +1,139 @@
...
+### 2: Implement Option A above
+
+- Pro: no magic, there's exactly 1 type for the type in the UDL/macro
+- Con: That type gets "tainted" by being used as an error. Using the interface in other contexts is less natural.
+- Con: It might be tricky to use as an external type - all external consumers must use as an error, or none of them must.
+
+### 3: Implement Option B above:
+
+- Pro: a more "pure" system - the interface can be used naturally in contexts other than being an error.
+- Pro: It mirrors Rust features in that it "adds" a capability rather then transforms it.
```
bendk, line 122: I agree that this feels more pure/natural from a Rust POV.  But in practice, aren't error types treated somewhat specially by the library?

docs/adr/0008-allowing-more-types-as-errors.md
```diff
@@ -0,0 +1,139 @@
...
+### 4: Implement Option C above:
+
+- Pro and Cons roughly the same for Option B plus:
+- Pro: less global types
+- Pro: the relationship between the interface and error is cleaner
+- Con: Might feel less natural to foreign authors.
+
+### 5: Something else?
+
+* What would exposing a `dictionary` (ie, a struct with only data) mean? What other types might become errors, and how does that inform us?
```
bendk, line 134: I think a dictionary makes sense.  It would work like a single enum variant, but it's base class would be `Exception` rather than the enum type.

bendk (commented):
I'm on the side of `A`, but I wouldn't mind `B` or `C` too much.

docs/adr/0008-allowing-more-types-as-errors.md
mhammond: We'd then get to choose A, B or C here too. Would we want to force there to be exactly 1 record type which derives from `Exception` and not have a plain-old record available?

Something I also meant to mention above is a compatibilty concern: imagine you use a struct/record today as *not* an error. Trying to use them as an error would be a breaking change if it meant the name and layout changed *everywhere* it was used just because a single function in a single crate wanted to use it an an error.

docs/adr/0008-allowing-more-types-as-errors.md
mhammond: I think it depends on the context: today, if I see a `struct Foo` in Rust, I know how that's going to be represented by UniFFI in bindings. With option A, that's no longer true - I now need to search for every single place `Foo` is used, in all crates, and if I can identify even a single one as being an `E` in a `Result<>`, then that representation changes. I think that's really my biggest objection to A - that how `Foo` is used in a place which is far far away from where `Foo` is defined changes what you understand about `Foo` in a significant and global way.

docs/adr/0008-allowing-more-types-as-errors.md
```diff
@@ -0,0 +1,139 @@
...
+    fn message(&self) -> String { todo!() }
+}
+fn do_it() -> Result<(), ErrorDetails> { Err(ErrorDetails {}) }
+```
+
+How should this be captured on the foreign side?
+
+### A) Create an "error only" type
+
+This is like the status quo re enums. It means:
```
badboy, line 37: Status Quo but more types, not just enums.

docs/adr/0008-allowing-more-types-as-errors.md
badboy: If we still require an `[Error]` annotation on that type it's clear, isn't it?

docs/adr/0008-allowing-more-types-as-errors.md
badboy: The renaming is indeed not great. My first instinct would be to not rename user types. 
Yes, that means you either don't have the `Exception` suffix in languages where it is an exception or you need to use the `Exception` suffix in Rust code, which also feels awkward.

badboy: A dictionary translates into a class/struct in the foreign language, so it could inherit/implement the corresponding error/exception class/trait.
That still works in A and except from its inherit it doesn't change much, does it?

badboy (commented):
I'm definitely in favor of the overall premise: (nearly?) any type should be usable as an error type.

I'm leaning towards A.
To me it feels more natural to have the error type as an exception directly.
Those error types would still be annotated, wouldn't they?

docs/adr/0008-allowing-more-types-as-errors.md
mhammond: What I meant here was that (a) is like the status for enums - once an enum is marked as an error it's no longer useful as an enum.

docs/adr/0008-allowing-more-types-as-errors.md
mhammond: We don't need an `[Error]` annotation. I guess we could insist on one anyway, but that is a bit of a smell for me - ie, with (a) "we would insist on an error attribute so that other side-effects (ie, the fact this is used somewhere as an error causes the object to change significantly) become less surprising" vs (b) and (c) "no side effects so no need for additional annotations"

docs/adr/0008-allowing-more-types-as-errors.md
mhammond: We already have precedents for renaming error types - I think both Swift and Kotlin will rename "Error" to "Exception" and I *think* linters will complain otherwise.

(b) and (c) allows you to just append "Error" or "Exception" so you get this for free.

bendk: > We'd then get to choose A, B or C here too. Would we want to force there to be exactly 1 record type which derives from Exception and not have a plain-old record available?

I think so

> Something I also meant to mention above is a compatibilty concern: imagine you use a struct/record today as not an error. Trying to use them as an error would be a breaking change if it meant the name and layout changed everywhere it was used just because a single function in a single crate wanted to use it an an error.

I would rephrase this as you simply couldn't repurpose the type as an error type, which doesn't seem so bad.  From the perspective of a foreign language consume, types are either exceptions or they're not and this restriction is completely natural.

This limitation would also be easy to overcome once we had more flexibility.  If I had a `Foo` type, I could always add a `FooError` type that was a dictionary with a single `foo` field.

mhammond:
The consensus seems to be (a). A nice part of this is that the choice between (a), (b) and (c) is purely in bindings - it will not change the core of the implementation, so different bindings could make different choices, and some bindings might make it optional. I'm not going to propose that happen, just nothing that it could :)

I'll put a PR up, and if we are happy with it I'll update this PR to reflect the decision.

docs/adr/0008-allowing-more-types-as-errors.md
mhammond: I added a few comments around this.

docs/adr/0008-allowing-more-types-as-errors.md
mhammond: I changed this to reflect that even if we support other types, the same key considerations are also going to apply to them.

docs/adr/0008-allowing-more-types-as-errors.md
mhammond: I reworded this a little.

mhammond:
I've updated this PR to reflect the consensus that (A) is the preferred option, and also clarified a few other things based on the comments. I think this is ready to rock and roll!

bendk (approved):
Looks great to me, let's land this.

ddurst:
@mhammond @badboy if this is still r+, can we close it out?

mhammond:
On reflection I think we should just kill this - it's not a bug enough decision to actually warrant an ADR, but it did serve a useful purpose for deciding on an implementation strategy.

Issue #1604, opened 2023-06-18
UniFFiing types and metadata: splitting out UDL


mhammond:
If agreed, the outcome of this issue will be the following:

* Split the UDL parsing code into its own crate, having it emit `uniffi_meta` structs
* Have `uniffi_bindgen` depend only on `uniffi_meta` - ie, it knows nothing about weedle, udl, etc
* `uniffi_bindgen` however does remain responsible for generating "scaffolding" (ie, the rust code to support what's defined in the UDL)

The end result is that:

* UDL is "demoted" (ie, no longer a first-class citizen of uniffi_bindgen)
* `uniffi_macros` is "promoted" to the canonical source for `uniffi_bindgen`.
* `uniffi_bindgen` continues to use its "interface" objects - eg, all bindings continue to use a `uniffi_bindgen::interface::Object`, it just that these `Object`s are always instantiated from `uniffi_meta` (whereas now they are constructed via _both_ meta _and_ weedle types)
* A downside of this is that `uniffi_meta` will need to grow to accomodate metadata specific only to bindings - eg, the boolean to indicate whether an argument is `ByRef` will be added to `uniffi_meta` even though bindings and the procmacros never need it.
* `uniffi_meta` will similarly need to grow support for everything binding specific that can be expressed in UDL, even if `uniffi_macros` can't yet do that.

I propose:

# Step 1
* Move core type, such as `Type`, `Literal` and `FfiType`, into `uniffi_meta`, removing the existing equivalent types.
  * Even though the uniffi_meta versions of these structs are arguably cleaner, we stick with the OG type struct to avoid breaking bindings.

# Step 2
* A new crate, `uniffi_udl`: All UDL specific code in `uniffi_bindgen` is moved to this crate.
* `uniffi_udl` depends on `uniffi_meta`, but not `uniffi_bindgen` - thus it deals exclusively with `uniffi_meta`.
* `uniffi_meta` grows support for fields that are not yet supported by `uniffi_macros` but which `uniffi_bindgen` needs.
* `uniffi_bindgen` then springs a `ComponentInterface` and `TypeUniverse` into existence only via `uniffi_meta` types.

# Steps N++
* See if there's further scope to have the scaffolding and binding generation become more specialised.

@jplatte @bendk any thoughts or objections?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-285)
┆Issue Number: UNIFFI-285

[...]

PR #1615, opened 2023-06-20
Move Type and related structs/enums to uniffi_meta.

mhammond:
This is what "step 1" from #1604 might look like.


uniffi_bindgen/src/interface/function.rs
```diff
@@ -381,20 +375,17 @@ mod test {
         let func2 = ci.get_function_definition("rich").unwrap();
         assert_eq!(func2.name(), "rich");
         assert_eq!(
-            func2.return_type().unwrap().canonical_name(),
```
mhammond, line 384: So yeah, when I moved canonical_name into the Ruby bindings, I didn't notice that function *was* used in tests, so the version in the Ruby bindings were actually being used here. I think these test changes are improvements though - we are now checking the *actual* type.

uniffi_meta/src/lib.rs
```diff
@@ -218,53 +223,7 @@ pub struct FnParamMetadata {
...
-    },
-    HashMap {
-        key_type: Box<Type>,
-        value_type: Box<Type>,
-    },
-}
-
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize)]
-pub enum Literal {
+pub enum LiteralMetadata {
```
mhammond, line 226: I think the literals are going to need to change some more, but this seems reasonable for now

mhammond:
To hopefully make this clearer:

> uniffi_bindgen can then depend on uniffi_udl

The way I see it, `uniffi_udl` ends up with something like `pub fn parse_udl(udl_file: &Utf8Path) -> Result<uniffi_meta::MetadataGroup>`

While `uniffi_bindgen` depends on `uniffi_udl`, it's really just for this function - that would be (roughly) the only `pub` thing exported by `uniffi_udl`. It is via this function that `uniffi_bindgen` could continue to export a function `pub fn parse_udl(udl_file: &Utf8Path) -> Result<ComponentInterface>`

mhammond:
> It also feels a bit weird that uniffi_meta now handles FfiType.

Yeah, good call. I moved this so that I could easily keep `Type::ffi_type(&self)` - which I really only did to try and avoid yet more breakage in the bindings. However, I've come up with a way to keep `FfiType` in `uniffi_bindgen` where it belongs, and make the bindings a little bit cleaner, even though we do need to suck up some minor changes to the bindings.

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -420,6 +420,10 @@ pub mod filters {
         Ok(as_ct.as_codetype().literal(literal))
     }
 
+    pub fn ffi_type(type_: &impl AsType) -> Result<FfiType, askama::Error> {
+        Ok(type_.as_type().into())
+    }
```
bendk, line 425: I think this could live in https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/backend/filters.rs and not have to be duplicated for each bindings language.

bendk:
> The way I see it, `uniffi_udl` ends up with something like `pub fn parse_udl(udl_file: &Utf8Path) -> Result<uniffi_meta::MetadataGroup>`
> 
> While `uniffi_bindgen` depends on `uniffi_udl`, it's really just for this function - that would be (roughly) the only `pub` thing exported by `uniffi_udl`. It is via this function that `uniffi_bindgen` could continue to export a function `pub fn parse_udl(udl_file: &Utf8Path) -> Result<ComponentInterface>`

Couldn't this be done right now, relying on the `uniffi_meta::Type` -> `uniffi_bindgen::Type` conversion that exists?  I'm not saying we shouldn't do this, I'm just wondering what depends on what.

uniffi_meta/src/types.rs
```diff
@@ -0,0 +1,147 @@
...
+    },
+    ForeignExecutor,
+    // Types defined in the component API, each of which has a string name.
+    Record(String),
+    Enum(String),
+    CallbackInterface(String),
+    // Structurally recursive types.
+    Optional(Box<Type>),
+    Sequence(Box<Type>),
+    Map(Box<Type>, Box<Type>),
```
bendk, line 96: Can this be `Map { key_type: Box<Type>, value_type: Box<Type> }`?

I don't mind the tuple-style for Optional and Sequence, but once we get multiple child types I think named fields are better.

bendk (approved):
Now that FfiType is back in `uniffi_bindgen`, this seems like an improvement to me overall.

mhammond:
> Couldn't this be done right now, relying on the uniffi_meta::Type -> uniffi_bindgen::Type conversion that exists? I'm not saying we shouldn't do this, I'm just wondering what depends on what.

Yeah, I guess it could - that would mean everything weedle related which refers to `Type` would need to change to refer to the slightly differently shaped `Type`, which could be done if necessary, but IMO it's mildly better to just unify these types. 

See also https://github.com/mhammond/uniffi-rs/pull/5

uniffi_meta/src/types.rs
mhammond: sure! How about just `k` and `v`?

mhammond:
The `module_path` changes caused a fair bit of churn, in particular, we must always set it to the namespace in uniffi_bindgen. This in turn meant changing how we get the namespace. It all works out in the end though, and my other PR cleans some of that up. @bendk do you mind another look?

bendk (approved):
This still looks good to me.

PR #1619, opened 2023-06-21
Delay contract checks until after all functions are defined

badboy:
Otherwise we might call it with the wrong ABI.
This so far wasn't an issue on platforms UniFFI tests on, but can be observed on Windows (at leats in Wine), where suddenly the checksums won't match, e.g.
39784 is expected, but the call gets 33135464 back. The former is 0x9b68 and fits well within a 16-bit unsigned integer. The latter is 0x1f99b68 which does not fit (but it has the same suffix, guess why...)


bendk (commented):
Sorry for this one, it's like the endian issue all over again.  The fix looks good though.

bendk (approved):

PR #1620, opened 2023-06-22
Add a version to the example/fixture dependencies on uniffi

glandium:
This should allow to vendor the crates without hacks like https://searchfox.org/mozilla-central/rev/fb55a4ee44a9a95d099aa806ca494eb988252ded/Cargo.toml#192-197

The downside is that it requires updating the version when the uniffi version bumps in a semver-incompatible way.


badboy:
> The downside is that it requires updating the version when the uniffi version bumps in a semver-incompatible way.

Luckily our `cargo-release` tooling takes care of that.

badboy (approved):

PR #1626, opened 2023-06-29
Callback interface via export missing method argument types

ijc:
Using `#[uniffi::export(callback_interface)]` on a trait which has methods taking `Option<T>` arguments currently fails because `FfiConverterOptional$T` is not defined in the generated Kotlin code (at least).

Here a failing test is added to the `proc-macro` fixtures and then fixed by registering the types in the CB related path of `ComponentInterface::add_trait_method_meta`.

I did try just making it call `ComponentInterface::add_method_meta` but this hit the "object ... not found" assertion in there. Perhaps I should have done a deeper refactoring to also allow more consolidation?

@bendk you recently consolidated some code in ba28e9d50cef1ed857599e939e39782eddc7365b which I think is related, although by inspection I think the issue was already present before that.

### DIFF
--- a/uniffi_bindgen/src/interface/mod.rs
+++ b/uniffi_bindgen/src/interface/mod.rs
@@ -754,7 +754,12 @@ impl ComponentInterface {
                     meta.index,
                 );
             }
-            cbi.methods.push(meta.into());
+            let defn: Method = meta.into();
+            for arg in defn.iter_types() {
+                self.types.add_known_type(arg);
+            }
+
+            cbi.methods.push(defn);
         } else {
             self.add_method_meta(meta)?;
         }
--- a/fixtures/proc-macro/src/callback_interface.rs
+++ b/fixtures/proc-macro/src/callback_interface.rs
@@ -8,5 +8,6 @@ use crate::BasicError;
 pub trait TestCallbackInterface {
     fn do_nothing(&self);
     fn add(&self, a: u32, b: u32) -> u32;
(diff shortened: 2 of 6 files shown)


uniffi_bindgen/src/interface/mod.rs
```diff
@@ -754,7 +754,15 @@ impl ComponentInterface {
                     meta.index,
                 );
             }
-            cbi.methods.push(meta.into());
+            let defn: Method = meta.into();
+            for arg in &defn.arguments {
```
mhammond, line 758: I know this hasn't been done consistently everywhere else, but `defn.iter_types()` is a slightly more compact way of doing this.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -754,7 +754,15 @@ impl ComponentInterface {
                     meta.index,
                 );
             }
-            cbi.methods.push(meta.into());
+            let defn: Method = meta.into();
+            for arg in defn.iter_types() {
+                self.types.add_known_type(arg);
+            }
+            if let Some(ty) = &defn.return_type {
```
mhammond, line 761: thanks - but `iter_types()` also gives the return type, so this block should be unnecessary

ijc: How convenient! Adjusted, thanks

mhammond (approved):

mhammond:
Thanks!

PR #1627, opened 2023-06-29
Add Interoptopus under alternative tools

FeldrinH:
Another alternative for generating multi-language bindings that I've come across is Interoptopus. I figured it ought to be listed here for completeness.


badboy (approved):
Neat, I didn't know about it

PR #1629, opened 2023-07-03
Split UDL parts of uniffi_bindgen into a new uniffi_udl crate.

mhammond:
"step 2" from https://github.com/mozilla/uniffi-rs/issues/1604

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
   "uniffi_macros",
   "uniffi_meta",
   "uniffi_testing",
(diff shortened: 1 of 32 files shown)


bendk (approved):
Yes please.  I'll just repeat my comment I made in the PR in your repo:  I know there's a lot of churn, but ISTM that if we're going to move in this general direction, then there's no avoiding that. I think the direction is definitely correct, so let's go for it.

I guess the one thing that should be fixed before merging is the doc lint issue.

PR #1632, opened 2023-07-03
Don't require a UDL file to generate scaffolding

mhammond:
I was adding a new example which didn't need a UDL and figured I'd see how hard it was to avoid needing one. I think this looks reasonable.


bendk:
We have dueling PRs for this.  I made https://github.com/mozilla/uniffi-rs/pull/1633/files, which goes a bit further by removing the need for `build.rs` and allows the scaffolding to be generated without using Askama at all.

I can't really see a downside to that, but I also think this PR fine.  I'd be happy to land this one now if that one needs discussion.

mhammond:
Let's take #1633 instead!

PR #1633, opened 2023-07-04
Allows crates to use only proc-macros and not UDL

bendk:
- Added a `setup_scaffolding` proc-macro.  This does things like generate `crate::UniffiTag` and export the rustbuffer scaffolding functions.
- Updated the Askama scaffolding template to call `setup_scaffolding!()`
- Removed the scaffolding check that `uniffi_bindgen` and `uniffi` have the same version.  I don't think this is needed anymore now that users only depend on `uniffi` and get their `uniffi_bindgen` dependency transitively.  Note: we still need a similar check in the bindings, but this commit only changes the scaffolding generation.
- Updated `fixtures/simple-iface` to test the new system
- Updated docs

### DIFF
--- a/uniffi_bindgen/src/library_mode.rs
+++ b/uniffi_bindgen/src/library_mode.rs
@@ -17,16 +17,16 @@
 ///     package maps.
 use crate::{
     bindings::{self, TargetLanguage},
-    macro_metadata, parse_udl, ComponentInterface, Config, Result,
+    macro_metadata, ComponentInterface, Config, Result,
 };
 use anyhow::{bail, Context};
 use camino::Utf8Path;
 use cargo_metadata::{MetadataCommand, Package};
 use std::{
-    collections::{BTreeSet, HashMap, HashSet},
+    collections::{HashMap, HashSet},
     fs,
 };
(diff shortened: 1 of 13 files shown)


mhammond (approved):
This is great! I like the `load_component_interface` -> `load_udl_metadata` refactor in mine, but everything else about yours is better :)

bendk:
I like that refactor too, just updated the code to use it.

uniffi_bindgen/src/library_mode.rs
```diff
@@ -167,28 +169,35 @@ fn find_package_by_crate_name(
...
-    crate_name: &str,
+fn load_udl_metadata(
+    group: &MetadataGroup,
     crate_root: &Utf8Path,
-    metadata: &BTreeSet<uniffi_meta::Metadata>,
-) -> Result<ComponentInterface> {
-    let udl_items = metadata
+    crate_name: &str,
+) -> Result<Option<MetadataGroup>> {
+    //            macro_metadata::add_group_to_ci(&mut ci, group)?;
```
mhammond, line 177: nano-nit: kill this?

Issue #1631, opened 2023-07-03
Failing to compile after a recent commit


0xOmarA:
I first want to start by thanking the team maintaining UniFFI. I've been looking into this tool for a little over a week now and have been surprised by how capable and how great the generated bindings are. We have faced the same difficulties with code sharing that the various teams over at Mozilla seem to have faced and UniFFI has so far been a pleasure to use! Thank you for writing this tool!

I'm using UniFFI as a git dependency as that's what the UniFFI book [used to recommend](https://github.com/mozilla/uniffi-rs/blob/f0041800c48293d631a2f425861726a5440e245b/docs/manual/src/proc_macro/index.md?plain=1#L15-L17) before the proc-macros feature moved to being non-experimental. More specifically, I'm depending on commit 862e653a9ab8e7b91592fa140cdeb8f1d48a7bb7 of UniFFI-rs.

In my production code I have an enum of 29 variants where each variant has between 2-4 fields each. Some of those fields are of the same data type and some are of differing data types. My enum derives the `uniffi::Enum` trait through UniFFI's `uniffi::Enum` proc-macro. So, I do not have any hand-written definitions in the udl file.

I can generate bindings with no problem when depending on the commit hash 4f39e5d4046d301d75fb4f16457a996bbda05447. However, when updating to f0041800c48293d631a2f425861726a5440e245b, all of a sudden I'm no longer able to compile my Rust crate with the UniFFI API, I get the following error:

```
   Compiling package v0.1.0 (/Users/0xOmarA/Desktop/package)
error[E0080]: evaluation of constant value failed
   --> /Users/0xOmarA/.cargo/git/checkouts/uniffi-rs-6f89edd2a1ffa4bd/f004180/uniffi_core/src/metadata.rs:122:9
    |
122 |         assert!(self.size + other.size <= BUF_SIZE);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'assertion failed: self.size + other.size <= BUF_SIZE', /Users/0xOmarA/.cargo/git/checkouts/uniffi-rs-6f89edd2a1ffa4bd/f004180/uniffi_core/src/metadata.rs:122:9
    |
note: inside `MetadataBuffer::concat`
   --> /Users/0xOmarA/.cargo/git/checkouts/uniffi-rs-6f89edd2a1ffa4bd/f004180/uniffi_core/src/metadata.rs:122:9
    |
[...]

PR #1636, opened 2023-07-05
Bump the size of the metadata buffer

0xOmarA:
As discussed in #1631, this PR doubles the size of the metadata buffer to address the problem discussed in the issue.


bendk (approved):
Looks good to me!  Thanks for opening the PR.

PR #1644, opened 2023-07-07
Improve async-related sanity checks in scaffolding generation

jplatte:
This allows

```rust
impl Foo {
    fn just_a_regular_fn(&self) { /* ... */ }
    async fn my_async_fn(&self) { /* run within the context of a tokio runtime */ }
}
```

which used to complain that "this attribute is only allowed on async functions" (pointing to the `"tokio"` string).

### DIFF
--- a/fixtures/futures/src/lib.rs
+++ b/fixtures/futures/src/lib.rs
@@ -171,6 +171,23 @@ impl Megaphone {
     }
 }
 
+// The async_runtime attribute used to error when *any* function in the impl block was not async,
+// now it should work as long as at least one function *is* async.
+#[uniffi::export(async_runtime = "tokio")]
+impl Megaphone {
+    /// A sync method that yells something immediately.
+    pub fn say_now(&self, who: String) -> String {
+        format!("Hello, {who}!").to_uppercase()
+    }
+
+    /// An async method that yells something after a certain time.
+    ///
+    /// Uses tokio's timer functionality.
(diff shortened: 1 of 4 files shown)


mhammond:
Any chance of a test? And how would such an impl have 2 async functions, one with a tokio executor and one with a foreign one?

jplatte:
> Any chance of a test?

Can add one tomorrow. 

> And how would such an impl have 2 async functions, one with a tokio executor and one with a foreign one?

For now, as before, not at all.

mhammond:
> > And how would such an impl have 2 async functions, one with a tokio executor and one with a foreign one?
> 
> For now, as before, not at all.

Right, but currently it can be selected per function. Wouldn't it make sense to do it similarly, allow each method to choose?

jplatte:
I think if one depends on tokio either way, it's extremely unlikely you'd want to enter the tokio runtime context only for some specific async functions, but not for others.

mhammond:
Wouldn't that apply to functions too?

jplatte:
Sure, but there is not really a nice way of configuring proc-macros on a whole-crate basis.

mhammond:
Yeah, whole-crate configuration belongs in uniffi.toml. If it's crate-wide, surely we can work out how to avoid the attribute entirely?

mhammond (approved):
So while I'm getting abstract and wondering about how the bindings might do more here, this is a nice improvement, thanks.

PR #1646, opened 2023-07-08
structs with data don't work?

mhammond:
Not sure what I'm doing wrong here.


jplatte:
AFAIK `usize` is simply not a type that UniFFI supports?

PR #1660, opened 2023-07-22
Fix export of Swift custom types which don't have a foreign type

mhammond:
If a Custom Type doesn't have a custom foreign type (as specified in uniffi.toml), Swift fails to export the custom type meaning it can't be used externally.

In this example, Swift is treating the `Guid` as a plain string, which didn't export the public FFI functions.


bendk (approved):
Fix looks good to me.

PR #1661, opened 2023-07-22
Allow procmacros to define custom types

mhammond:
Includes docs and examples.

### DIFF
--- a/docs/manual/src/proc_macro/index.md
+++ b/docs/manual/src/proc_macro/index.md
@@ -185,6 +185,45 @@ impl Foo {
 }
 ```
 
+## The `uniffi::CustomType` derive
+
+The `CustomType` derive registers a type as a custom type and can be used on any struct that wants to be exposed
+as a builtin type. You must specify the builtin type using either the `builtin` or `newtype` attribute.
+
+If you specify the `builtin` attribute, you must manually implement the `UniffiCustomTypeConverter` for your
+type as described in the [UDL documentation for Custom Types](../udl/custom_types.md). However, if you
+specify the `newtype` attribute UniFFI will generate the implementation of this trait assuming the type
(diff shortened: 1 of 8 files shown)


docs/manual/src/proc_macro/index.md
```diff
@@ -185,6 +185,45 @@ impl Foo {
...
+}
+```
+
+Whereas this type uses the "new type" idiom, so `UniffiCustomTypeConverter` is automatically derived
+
+```rust
+#[derive(uniffi::CustomType)]
+#[uniffi(newtype=i64)]
+pub struct NewtypeHandle(i64);
+```
```
bendk, line 225: This is awesome

uniffi_macros/src/lib.rs
```diff
@@ -228,6 +251,16 @@ pub fn scaffolding_ffi_converter_callback_interface(tokens: TokenStream) -> Toke
...
     let name = syn::parse_macro_input!(component_name as LitStr);
+    // A trait that's in our crate for our external wrapped types to implement.
+    let custom_type_converter = quote! {
+        #[allow(unused)]
+        trait UniffiCustomTypeConverter {
+            type Builtin;
+            fn into_custom(val: Self::Builtin) -> uniffi::Result<Self> where Self: Sized;
+            fn from_custom(obj: Self) -> Self::Builtin;
+        }
+    };
```
bendk, line 262: I think this should also be included added in `setup_scaffolding!()`.

bendk (approved):
Looks great to me, except my one comment about `setup_scaffolding()`.  It's nice that the new handling is even more ergonomic than the old one and it also seems like we're finally at a place where we can expand the proc-macro functionality without having to touch the basic system.

PR #1662, opened 2023-07-24
Allow interfaces to be errors.

mhammond:
This is an implementation of adr-0008, as proposed in #1613. It has support for interfaces described in UDL and procmacros. This will remains as a draft until #1613 lands, but I think this is actually ready for review now.

It only supports Python - other bindings can come later and existing bindings are not impacted (unless you actually try and use the feature there).

### DIFF
--- a/docs/manual/src/udl/errors.md
+++ b/docs/manual/src/udl/errors.md
@@ -42,3 +42,85 @@ interface ArithmeticError {
   IntegerOverflow(u64 a, u64 b);
 };
 ```
+
+## Interfaces as errors
+
+** This is not avaiable in all bindings **
+
+It's possible to use an `interface` (ie, a rust struct impl) as an error;
+the thrown object will have methods instead of fields.
+This can be particularly useful when working with `anyhow` style errors, where
+an enumeration can't represent the various error states.
+
+In your UDL:
+```
+namespace error {
+  [Throws=MyError]
+  void bail(string message);
+}
+
+[Traits=(Debug)]
+interface MyError {
+  string message();
+};
+```
+and Rust:
+```rs
+#[derive(Debug)]
+pub struct MyError {
+    e: anyhow::Error,
+}
+
+impl MyError {
+    fn message(&self) -> String> { ... }
+}
+
+// anything using `anyhow::Error` can throw `MyError`.
+impl From<anyhow::Error> for MyError {
+    fn from(e: anyhow::Error) -> Self {
+        Self { e }
+    }
+}
+
+fn bail(message: String) -> anyhow::Result<()> {
+    anyhow::bail!("{message}");
+}
+```
+then Python might then do:
+```py
+try:
+  bail("oh no")
+except MyError as e:
+  print("oops", e.message())
+```
+
+This works for procmacros too, although without the builtin type coercion:
+
+```rs
+#[derive(Debug, uniffi::Error)]
+pub struct ProcErrorInterface {
+    e: String,
+}
+
+#[uniffi::export]
+impl ProcErrorInterface {
+    fn message(&self) -> String {
+        self.e.clone()
+    }
+}
+
+impl std::fmt::Display for ProcErrorInterface {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "ProcErrorInterface {}", self.e)
+    }
+}
+
+#[uniffi::export]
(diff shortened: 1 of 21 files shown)


docs/manual/src/udl/errors.md
```diff
@@ -42,3 +42,85 @@ interface ArithmeticError {
   IntegerOverflow(u64 a, u64 b);
 };
 ```
+
+## Interfaces as errors
+
+** This is not avaiable in all bindings **
+
+It's possible to use an `interface` (ie, a rust struct impl) as an error;
```
bendk, line 50: I'm not sure if "rust struct impl" is the right term for this, but I don't have anything much better.  Maybe just linking to the interface documents?

docs/manual/src/udl/errors.md
```diff
@@ -42,3 +42,85 @@ interface ArithmeticError {
...
+interface MyError {
+  string message();
+};
+```
+and Rust:
+```rs
+#[derive(Debug)]
+pub struct MyError {
+    e: anyhow::Error,
+}
```
bendk, line 72: Would it be possible to return `anyhow::Error` directly, maybe together with a crate-local trait that supplies the `message()` method?

docs/manual/src/udl/errors.md
```diff
@@ -42,3 +42,85 @@ interface ArithmeticError {
...
+impl std::fmt::Display for ProcErrorInterface {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "ProcErrorInterface {}", self.e)
+    }
+}
+
+#[uniffi::export]
+fn throw_proc_error(e: String) -> Result<(), Arc<ProcErrorInterface>> {
+    Err(Arc::new(ProcErrorInterface { e }))
+}
```
bendk, line 121: The `Arc::new` is kind of annoying.  I wonder if we could make it so we could return the struct directly and have UniFFI wrap it with an `Arc<>`.  It seems orthogonal to this work, but maybe I'll experiment with this.

uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
```diff
@@ -60,9 +63,32 @@ def __ne__(self, other: object) -> {{ ne.return_type().unwrap()|type_name }}:
...
 {%-         when UniffiTrait::Hash { hash } %}
             {%- call py::method_decl("__hash__", hash) %}
-{%      endmatch %}
-{% endfor %}
+{%-      endmatch %}
+{%- endfor %}
+
+{%- if ci.is_name_used_as_error(name) %}
+{# Due to some mismatches in the ffi converter mechanisms, errors are forced to be a RustBuffer #}
+class {{ ffi_converter_name }}__as_error(_UniffiConverterRustBuffer):
```
bendk, line 71: How about defining something like Kotlin's [lowerIntoRustBuffer()](https://github.com/mozilla/uniffi-rs/blob/d15d530439b65e9d8f9c6e0f8847306388ac7a4a/uniffi_bindgen/src/bindings/kotlin/templates/FfiConverterTemplate.kt#L34-L47) rather than a second FFIConverter class.  As that method mentions, it's also what we need to do for [callback interfaces](https://github.com/mozilla/uniffi-rs/blob/d15d530439b65e9d8f9c6e0f8847306388ac7a4a/uniffi_bindgen/src/bindings/python/templates/CallbackInterfaceTemplate.py#L37-L39), so having that function could simplify that code a bit to.

uniffi_bindgen/src/scaffolding/templates/ObjectTemplate.rs
```diff
@@ -55,7 +55,7 @@ pub extern "C" fn {{ ffi_free.name() }}(ptr: *const std::os::raw::c_void, call_s
         uniffi::rust_call(call_status, || {
             {{ cons|return_ffi_converter }}::lower_return(
                 {%- if cons.throws() %}
-                {{ obj.rust_name() }}::{% call rs::to_rs_call(cons) %}.map(::std::sync::Arc::new).map_err(Into::into)
+                {{ obj.rust_name() }}::{% call rs::to_rs_call(cons) %}.map(::std::sync::Arc::new)
```
bendk, line 58: I feel good about this change, but maybe it breaks backwards comparability?  If so, it should have a changelog entry.

bendk (commented):
Overall this looks super promising.  There's maybe a bit of bikeshedding around how to return these errors, I really hope the proc-macro-based code can be simplified.  But this seems very nice overall.

PR #1666, opened 2023-07-25
Hack: Copy metadata module to avoid uniffi_core dependency

badboy:
In m-c we're running into linking issues in `uniffi_macros`, where it's missing symbols coming from `uniffi_core`:

    /bin/ld: uniffi_foreign_executor_callback_set: undefined version:
    /bin/ld: failed to set dynamic section sizes: bad value

That's most likely this issue on Rust:
https://github.com/rust-lang/rust/issues/111888
It's called out that this likely actually broke because of another PR: https://github.com/rust-lang/rust/pull/99944

Despite this bug there's a bit of an issue in UniFFI to begin with: We're exporting `extern "C"` functions from a crate that is a dependency of some other crates, including `uniffi_macros`, and thus the symbols land in a part where they are not needed.

However for `uniffi_meta` in particular we don't need much from `uniffi_core`, so for now we just copy the necessary bits to get it all working.

---

Same as https://github.com/mozilla/uniffi-rs/pull/1665, but against the `release-v0.24.x` branch

### DIFF
--- a/uniffi_meta/Cargo.toml
+++ b/uniffi_meta/Cargo.toml
@@ -14,4 +14,3 @@ bytes = "1.3"
 serde = { version = "1.0.136", features = ["derive"] }
 siphasher = "0.3"
 uniffi_checksum_derive = { version = "0.24.1", path = "../uniffi_checksum_derive" }
-uniffi_core = { path = "../uniffi_core", version = "=0.24.1" }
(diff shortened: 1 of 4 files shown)


bendk (approved):
Nice workaround!  Let's merge.

travis79 (approved):

PR #1667, opened 2023-07-25
Failing test case for future cancellation

jplatte:
Output from CI:

```
[fixtures/futures/src/lib.rs:285] &*MUTEX = Mutex {
    data: (),
}
[fixtures/futures/src/lib.rs:285] &*MUTEX = Mutex {
    data: <locked>,
}
test uniffi_foreign_language_testcase_test_futures_kts has been running for over 60 seconds

Too long with no output (exceeded 10m0s): context deadline exceeded
```

### DIFF
--- a/fixtures/futures/src/lib.rs
+++ b/fixtures/futures/src/lib.rs
@@ -276,4 +276,16 @@ pub async fn broken_sleep(ms: u16, fail_after: u16) {
     .await;
 }
 
+use once_cell::sync::Lazy;
+use tokio::sync::Mutex as TokioMutex;
+
+#[uniffi::export(async_runtime = "tokio")]
+pub async fn use_shared_resource() {
+    static MUTEX: Lazy<TokioMutex<()>> = Lazy::new(|| TokioMutex::new(()));
+    let guard = dbg!(&*MUTEX).lock().await;
+    tokio::time::sleep(Duration::from_millis(100)).await;
+    drop(guard);
+    dbg!();
+}
+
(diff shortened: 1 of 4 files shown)


fixtures/futures/src/lib.rs
```diff
@@ -276,4 +276,16 @@ pub async fn broken_sleep(ms: u16, fail_after: u16) {
...
 }
 
+use once_cell::sync::Lazy;
+use tokio::sync::Mutex as TokioMutex;
+
+#[uniffi::export(async_runtime = "tokio")]
+pub async fn use_shared_resource() {
+    static MUTEX: Lazy<TokioMutex<()>> = Lazy::new(|| TokioMutex::new(()));
+    let guard = dbg!(&*MUTEX).lock().await;
+    tokio::time::sleep(Duration::from_millis(100)).await;
```
bendk, line 286: I'm confused about how this works at all since I don't see any code to start up the tokio runtime. @Hywan how does this work for the `say_after_with_tokio` test?

jplatte: The tokio support for uniffi's async support was added by me, not Hywan 😉

It works by using the `async-compat` crate, which contains a lazily-initialized single-threaded tokio runtime. When an async fn with `#[uniffi::export(async_runtime = "tokio")]` is called for the first time, the runtime is initialized.

Hywan: Tokio isn't the important it here. We could achieve the same with a custom `Sleep` future and an async lock.

PR #1671, opened 2023-07-25
Allow different crate types when using library mode

bendk:
This makes library mode work when the dependent crate doesn't include "lib" in their `crate-type` list.  I think that check was leftover from older code and isn't needed anymore.

### DIFF
--- a/uniffi_bindgen/src/library_mode.rs
+++ b/uniffi_bindgen/src/library_mode.rs
@@ -158,9 +158,9 @@ fn find_package_by_crate_name(
         .packages
         .iter()
         .filter(|p| {
-            p.targets.iter().any(|t| {
-                t.name.replace('-', "_") == crate_name && t.crate_types.iter().any(|ct| ct == "lib")
-            })
+            p.targets
+                .iter()
+                .any(|t| t.name.replace('-', "_") == crate_name)
         })
         .collect();
     match matching.len() {


bendk:
@jplatte I think this should fix your issue with library mode, can you test it against your branch?

As a workaround in the meantime, I think you should be able to use a regular release if you add `lib` to the `crate-type` list for `matrix-sdk-ffi`

jplatte:
I'm getting a new, somewhat cryptic, error message now:

```
     Running `target/debug/uniffi-bindgen generate --language kotlin --language swift --library target/debug/libmatrix_sdk_ffi.a --out-dir target/generated-bindings`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Generate bindings for kotlin requires a cdylib, but target/debug/libmatrix_sdk_ffi.a was given', /home/jplatte/.cargo/git/checkouts/uniffi-rs-6f89edd2a1ffa4bd/088e565/uniffi/src/lib.rs:38:21
```

bendk:
@jplatte Could use use `target/debug/libmatrix_sdk_ffi.so` instead?

Maybe that check should be removed, but the idea was to ensure that you use the same dylib that the bindings are eventually going to open.

jplatte:
Huh? Consumers could use either, the static or dynamic library. Anyways, I totally missed that I had given the *static* lib, so the error message actually was pretty clear 🤦🏼

jplatte:
Anyways, it works when passing the `.so` file! 🎉

aringenbach:
> @jplatte I think this should fix your issue with library mode, can you test it against your branch?
> 
> As a workaround in the meantime, I think you should be able to use a regular release if you add `lib` to the `crate-type` list for `matrix-sdk-ffi`

Just an FIY for people who might use this workaround, for some reason adding the `lib` type to the ffi bindings crate did increase the size of static libraries by quite a bunch (40MB to 80MB on one of my projects), so using the actual fix as soon as possible is definitely bette :).

PR #1672, opened 2023-07-26
Don't require objects to always be wrapped in `Arc<>`

bendk:
When reviewing #1662 I felt like I would be very annoyed to always have to wrap error objects with `Arc::new()`, this is an attempt to see if can avoid that.  It contains a general `FfiConverter` implementation for the struct itself, that simply wraps the value in `Arc::new()` then forwards the call to the normal `Arc<>` impl.

For constructors and other methods that return an object, UniFFI can wrap the object in an `Arc<>` just fine, the library doesn't need to do this.  Allowing the library to skip the wrapping can improve ergonomics, especially if we start to allow interfaces as errors.

### DIFF
--- a/uniffi_macros/src/object.rs
+++ b/uniffi_macros/src/object.rs
@@ -62,6 +62,7 @@ pub fn expand_object(input: DeriveInput, udl_mode: bool) -> syn::Result<TokenStr
 pub(crate) fn interface_impl(ident: &Ident, udl_mode: bool) -> TokenStream {
     let name = ident_to_string(ident);
     let impl_spec = tagged_impl_header("FfiConverterArc", ident, udl_mode);
+    let lower_return_impl_spec = tagged_impl_header("LowerReturn", ident, udl_mode);
     let lift_ref_impl_spec = tagged_impl_header("LiftRef", ident, udl_mode);
     let mod_path = match mod_path() {
         Ok(p) => p,
@@ -133,6 +134,16 @@ pub(crate) fn interface_impl(ident: &Ident, udl_mode: bool) -> TokenStream {
                 .concat_str(#mod_path)
                 .concat_str(#name)
                 .concat_bool(false);
+        }
+
+        unsafe #lower_return_impl_spec {
+            type ReturnType = <Self as ::uniffi::FfiConverterArc<crate::UniFfiTag>>::FfiType;
+
+            fn lower_return(obj: Self) -> Result<Self::ReturnType, ::uniffi::RustBuffer> {
+                Ok(<Self as ::uniffi::FfiConverterArc<crate::UniFfiTag>>::lower(::std::sync::Arc::new(obj)))
+            }
+
+            const TYPE_ID_META: ::uniffi::MetadataBuffer = <Self as ::uniffi::FfiConverterArc<crate::UniFfiTag>>::TYPE_ID_META;
         }
 
         unsafe #lift_ref_impl_spec {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 - Add support for [docstrings in UDL](https://mozilla.github.io/uniffi-rs/udl/docstrings.html)
 - Ability for UDL to use external trait interfaces [#1831](https://github.com/mozilla/uniffi-rs/issues/1831)
 - Add support for docstrings via procmacros [#1862](https://github.com/mozilla/uniffi-rs/pull/1862)
+- Objects can now be returned from functions/constructors/methods without wrapping them in an `Arc<>`.
 
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.25.2...HEAD).
 
--- a/docs/manual/src/proc_macro/index.md
+++ b/docs/manual/src/proc_macro/index.md
@@ -55,9 +55,9 @@ struct MyObject {
 #[uniffi::export]
 impl MyObject {
     // Constructors need to be annotated as such.
-    // As of right now, they must return `Arc<Self>`, this might change.
-    // If the constructor is named `new`, it is treated as the primary
-    // constructor, so in most languages this is invoked with `MyObject()`.
+    // The return value can be either `Self` or `Arc<Self>`
+    // It is treated as the primary constructor, so in most languages this is invoked with
+    `MyObject()`.
     #[uniffi::constructor]
     fn new(argument: String) -> Arc<Self> {
         // ...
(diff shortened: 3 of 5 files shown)


uniffi_macros/src/object.rs
```diff
@@ -138,6 +143,43 @@ pub(crate) fn interface_impl(ident: &Ident, tag: Option<&Path>) -> TokenStream {
...
+            fn write(obj: Self, buf: &mut Vec<u8>) {
+                <Self as ::uniffi::FfiConverterArc<#uniffi_tag>>::write(::std::sync::Arc::new(obj), buf)
+            }
+
+            fn lower_return(v: Self) -> ::std::result::Result<Self::FfiType, ::uniffi::RustBuffer> {
+                <Self as ::uniffi::FfiConverterArc<#uniffi_tag>>::lower_return(::std::sync::Arc::new(v))
+            }
+
+            fn try_lift(v: Self::FfiType) -> ::uniffi::Result<Self> {
+                panic!("Can't lift")
```
bendk, line 165: This is the main issue with this change.  If we implement `FfiConverter` on a type, then UniFFI-exported functions can input that type as an argument.  However, it's not possible to input the object type directly if it's already wrapped in an arc ([into_inner](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.into_inner) doesn't seem right).

Right now this is just implemented as a panic, which is unfortunate because if a library does try to input an object type it would be better to catch the issue at compile time.  I don't think there's a static assertion to check that a function is never called though.

One solution would be to split up `FfiConverter` into 2 traits: one for lifting and one for lowering.  That seems like a big change though, is there anything easier?

jplatte:
I agree that this should be supported, but I don't think anything should change / needs to change about the traits. This can be 100% a function of the scaffolding code generation: If a `#[uniffi::constructor]` returns exactly `Self` or the same `Ident` that's found in the impl header, we can add `Arc::new` to the generated scaffolding, otherwise omit it like we do now. Wdyt?

bendk:
> I agree that this should be supported, but I don't think anything should change / needs to change about the traits. This can be 100% a function of the scaffolding code generation: If a `#[uniffi::constructor]` returns exactly `Self` or the same `Ident` that's found in the impl header, we can add `Arc::new` to the generated scaffolding, otherwise omit it like we do now. Wdyt?

That's a neat idea, but I'm also concerned about non-constructurs that return interfaces.  In particular, if we want to use interfaces as errors, it's going to be annoying to have to add `map_err(Arc::new)` to every returned Result.

jplatte:
I see what you mean. Well in that case, I'm fully in support of splitting the trait in two, which I wanted anyways. We could allow users to opt into this for regular types as well, to reduce the amount of generated code when a certain type is either only ever used as (part of) a parameter, or only ever used as (part of) a return value.

bendk:
I wonder what the right split is.  In addition to `FfiLift` and `FfiLower` traits, I think it would also be nice to have an `FfiCallbackError` trait since errors used in callback interfaces have [extra requirements](https://mozilla.github.io/uniffi-rs/udl/callback_interfaces.html#extra-requirements-for-errors-used-in-callback-interfaces).  Maybe we should just go really fine grained and split out traits like `FfiRead`, `FfiWrite`, `FfiReturn`, etc.

jplatte:
`FfiReturn` is something I had previously, maybe you remember. If we have that separately, things like `Option<()>` that are not representable in many languages will be disallowed which they currently aren't (because `()` will only implement `FfiReturn`, none of the other traits).

bendk:
We have split up the `FfiConverter` trait so now this one works great.   By implementing `LowerReturn`, users can return an object from a constructor/function/method and UniFFI will wrap it in an `Arc<>` for you.

A step further would be to implement `Lower`, which would allow you to pass an object into a callback interface method without wrapping it.  That feels a bit too magical to me though, so I didn't do it.

bendk:
Another option would be to restrict this even more and only support it for constructors.  That feels like the least magical of all, but it would require adding a new FFI trait, maybe we could call it `LowerNew`.

mhammond:
> Another option would be to restrict this even more and only support it for constructors. That feels like the least magical of all,

I'm torn here. Allowing all methods to return an object and have the arc wrapping done does seem convenient and might even mean consumers need to make fewer changes to their existing APIs to allow UniFFI to expose it. I don't think it will mis-lead too many people - our docs make it quite clear *everything* is an Arc.

OTOH, I am sympathetic to "explicit is better than implicit", so expect people will disagree and respect that.

However, I *am* struggling a little with the concept that is makes sense for constructors but not elsewhere - both arguments above apply equally in both cases IMO.

(See also #1063 where there was strong objection to doing this for constructors. My memory is faulty, because for some reason I thought we supported that everywhere *except* constructors, but I don't think that's true)

bendk:
> However, I am struggling a little with the concept that is makes sense for constructors but not elsewhere - both arguments above apply equally in both cases IMO.

I don't disagree, but one detail is that if we want to support returning `T` objects, then what about `Vec<T>`, `Option<T>`, etc.  If we only implement `LiftReturn` then those won't be supported but users might expect them to be.

That's an argument for the two extreme cases: either limit the auto-arc-wrapping to constructors or support it anytime we lift a value.

mhammond (approved):
I was thinking out loud before - I should try to stop that :)

I don't think it's worth extra effort to *not* do something in the name of "purity" when it's actually quite useful.

mhammond:
So whether we *can* is different to whether we *should* :) A big part of that for me is how magic it seems and whether it helps or hurts the mental model users have with uniffi.

IMO, the direct question here (returning objects from constructors without the Arc<>) has clear utility and IMO isn't going to confuse anyone's mental model of the world. The fact functions also get this "for free" unless we take additional action to prevent it can be classed a "happy accident" - it might come in useful and I don't really see footguns there, so I'm not sure it's worth additional effort to *not* support it. In the same way, I doubt I'd be that keen on spending additional effort *to* support it if it didn't come for free. It could certainly be argued that these limitations *are* going to be confusing for users, and I suspect a good argument could persuade me if anyone feels strongly about it - but at face value I currently thing it's fine.

However, moving beyond this starts to move into uncertain territory for me. `Vec<>`, `Option<>` etc seem much more of a gray zone. What about hashmaps? What does an example hashmap look like in this scenario? Is the user also going to expect a degree of magic when interacting with objects in dictionaries etc? How would we explain *some* embedded `Arc`s can omit the Arc (eg, in options/hashmaps) but not others (eg, when returning a struct literal which itself has object members)

So without thinking too much about it, I'm kinda +1 on adding this capability to constructors, +1 on leaving a happy accident that other functions get the simple case for free too, and -0.5 on expanding this too far at this stage, because it's much easier to add this capability in the future than it would be to take it away if it turns out to be confusing for people.

bendk:
After talking with Mark, I'm not sure if implementing `Lower` is right or not so I reverted it back to the version that only implemented `LowerReturn`.

Going to merge this one since that change was already approved.

PR #1675, opened 2023-07-26
Make the foreign executor API public

bendk:
Technically it was always `pub` since it was used in the generated scaffolding, but this makes it more usable in consumer code:
  - Make the `RunFuture.task` field Sync, otherwise the `ForeignExecutor` functions are not really usable.
  - Added macros to improve the ergonomics
  - Added tests and documentation

### DIFF
--- a/docs/manual/src/futures.md
+++ b/docs/manual/src/futures.md
@@ -36,11 +36,62 @@ if __name__ == '__main__':
 
 This code uses `asyncio` to drive the future to completion, while our exposed function is used with `await`.
 
-In Rust `Future` terminology this means the foreign bindings supply the "executor" - think event-loop, or async runtime. In this example it's `asyncio`. There's no requirement for a Rust event loop.
-
 There are [some great API docs](https://docs.rs/uniffi_core/latest/uniffi_core/ffi/rustfuture/index.html) on the implementation that are well worth a read.
 
-See the [foreign-executor fixture](https://github.com/mozilla/uniffi-rs/tree/main/fixtures/foreign-executor) for more implementation details.
+## Foreign executors
+
+UniFFI futures are normally usually driven by a foreign executor - think event-loop, or async runtime.
+In the example above it's `asyncio`.
+There's no requirement for a Rust executor like tokio, although you can manually start one and use it alongside the foreign executor.
+Rust code can also use foreign executors to schedule tasks, for example to schedule work in a background thread.
+
+### Constructing ForeignExecutor instances in foreign languages
+
+- Kotlin uses a `CoroutineScope` (for example: `CoroutineScope(Dispatchers.IO)`)
+- Python uses an `EventLoop` (for example: `asyncio.get_running_loop()`)
+- Swift uses the UniFFI-defined class `UniFfiForeignExecutor` (for example: `UniFfiForeignExecutor(priority: TaskPriority.background)`)
+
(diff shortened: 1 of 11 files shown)


docs/manual/src/futures.md
```diff
@@ -36,11 +36,16 @@ if __name__ == '__main__':
...
 
-See the [foreign-executor fixture](https://github.com/mozilla/uniffi-rs/tree/main/fixtures/foreign-executor) for more implementation details.
+## Foreign executors
+
+UniFFI futures are normally usually driven by a foreign executor - think event-loop, or async runtime.
+In the example above it's `asyncio`.
+There's no requirement for a Rust executor like tokio, although you can manually start one and use it alongside the foreign executor.
+
+Rust code can also use foreign executors directly to schedule tasks, a typical use case is scheduling work in a background thread.
+For more info see the [ForeignExecutor API](https://docs.rs/uniffi_core/latest/uniffi_core/ffi/foreignexecutor/struct.ForeignExecutor.html) and the [futures example crate](https://github.com/mozilla/uniffi-rs/tree/main/examples/futures).
```
bendk, line 48: The `ForeignExecutor` docs.rs API will get the python/kotlin/swift examples from the struct docstring once this gets merged to main.

fixtures/futures/src/lib.rs
```diff
@@ -276,4 +279,54 @@ pub async fn broken_sleep(ms: u16, fail_after: u16) {
...
+        // background thread
+        42
+    }
+
+    /// Schedule incrementing a counter value using the foreign executor
+    ///
+    /// This tests the schedule() method which schedules a task to run "fire-and-forget" style,
+    /// where we don't get the result back
+    pub fn schedule_increment(self: Arc<Self>) {
+        Arc::clone(&self).background_executor.schedule(0, move || {
```
bendk, line 322: The `Arc::clone()` call is not really needed, but I couldn't find a way around it.  If you change it to `self.background_executor.schedule()` then Rust complains because the closure moves `self` while it's still borrowed for the `schedule()` call.

In practice I don't think there's any threading issues, because the closure won't execute until later on, but I can't figure out a way to tell Rust that.

uniffi_core/src/ffi/foreignexecutor.rs
```diff
@@ -110,6 +108,124 @@ impl ForeignExecutor {
...
+            //   - Note that the only code that will run in this thread is the
+            //     `executor.run()` call.  You can't implement Drop for references, so there's
+            //     no chance that dropping the reference will cause some other code to run.
+            //   - The owned value won't be dropped in this thread by the `executor.run()`
+            //     call.
+            //   - The owned value won't be dropped by another thread.  We currently have a
+            //     shared reference, so no other thread should be able to create a mutable
+            //     reference.
+            let executor = &*(executor as *const ForeignExecutor);
+            executor.run($delay, $closure)
```
bendk, line 178: I'm pretty sure this is correct, but please double-check the logic here.

bendk:
I think I found a way to have decent ergonomics here, at the cost of some unsafe code.  I think this is ready to be merged once someone double-checks that code.

bendk:
I'm going to close this one in favor of a larger PR that shows more of the "big picture".  That one will also include async callback interfaces and more example code.   It's hard to evaluate this one on its own without knowing how it relates to the larger async story.

PR #1676, opened 2023-07-27
Combine the 'external-types' and 'ext-types' fixtures.

mhammond:
The 'external-types' fixture was testing that a type defined in another crate could be used in UDL in this crate. It was a little confusing to have two fixtures with such similar names, so the two fixtures are now one.

I'm often mildly confused which fixture I'm after, so what do we think about merging them?


bendk (approved):
Yes this is currently pretty confusing.  I'm totally in favor of merging them.

PR #1677, opened 2023-07-27
[DO NOT MERGE] Try and work out what's causing kotlin test timeouts

mhammond:


arg0d:
The best way to reproduce this is to run the tests locally using `docker/cargo-docker.sh test`, and to wrap `test_futures.kts` file into `repeat(1000) { .. }` block to repeatedly run the tests many times. This way the hanging can be reproduced within a minute.

arg0d:
I did more digging and it looks like `callbackHolder` is not actually retaining the callback from being GCed. I created a global variable to retain the callback, and the tests stopped hanging. This is not a viable solution, since it wouldn't work if the same function is called multiple times in parallel.
https://github.com/mozilla/uniffi-rs/blob/7b18fa787336aa738f205a44f7b2df728e83a097/uniffi_bindgen/src/bindings/kotlin/templates/TopLevelFunctionTemplate.kt#L15

Hywan:
Fixed by #1684.

Issue #1678, opened 2023-07-27
Error `the trait FfiConverterArc<UniFfiTag> is not implemented for RustGlobalViewModel` when using only proc macros (without UDL file)


praveenperera:
NOTE: I tried this on master https://github.com/mozilla/uniffi-rs/commit/93634167f234eda2e891c45f0c7a9b9310fe73dc. On `0.24.2` I was getting `setup_scaffolding` not found in `uniffi`

I tried to remove the udl file completely and switch to using only proc macros. I was able to remove everything accept for the constructor.

## Broken

lib.rs
```rust
pub mod view_models;

uniffi::setup_scaffolding!();
```

global.rs
```rust
pub struct RustGlobalViewModel;

#[uniffi::export]
impl RustGlobalViewModel {
    #[uniffi::constructor]
    pub fn new() -> Arc<Self> {
        Arc::new(Self)
    }
 }
```

I get the following error

```
61 | #[uniffi::export]
                              │     | ^^^^^^^^^^^^^^^^^ the trait `FfiConverterArc<UniFfiTag>` is not implemented for `RustGlobalViewModel`
                              │     |
                              │     = help: the trait `FfiConverter<UT>` is implemented for `Arc<T>`
                              │     = note: required for `Arc<RustGlobalViewModel>` to implement `FfiConverter<UniFfiTag>`
                              │     = note: this error originates in the attribute macro `uniffi::export` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I was able to get it working by adding the UDL file with just the constructor 

## Working

kube_viewer.udl
```
namespace kube_viewer {};

interface RustGlobalViewModel {
  constructor();
};
```

lib.rs
```rust
pub mod view_models;

uniffi::include_scaffolding!("kube_viewer");
```

global.rs
```rust
impl RustGlobalViewModel {
    pub fn new() -> Self {
        Self
    }
}
```

build.rs
```rust
fn main() {
    uniffi::generate_scaffolding("./src/kube_viewer.udl").unwrap();
}
```

mhammond:
Fairly sure you need `#[derive(uniffi::Object)]` on the struct, eg, https://github.com/mozilla/uniffi-rs/blob/93634167f234eda2e891c45f0c7a9b9310fe73dc/fixtures/proc-macro/src/lib.rs#L36

praveenperera:
[...]

PR #1679, opened 2023-07-27
Update docs for proc-macro constructors

praveenperera:
With this change code copied from the docs will work.

Ref: #1678


mhammond (approved):
Thanks!

PR #1680, opened 2023-07-31
Avoid using a derive for custom types to support types defined elsewh…

mhammond:
You need a struct definition to use a derive, but CustomTypes are designed to work even when the strict is defined externally. This PR allows for that to work.

### DIFF
--- a/docs/manual/src/proc_macro/index.md
+++ b/docs/manual/src/proc_macro/index.md
@@ -198,11 +198,12 @@ implements the "new type" idiom (ie, a tuple struct with the builtin as the only
 
 For example, this type does not use the "new type" idiom, so must specify the complete implementation.
 ```rust
-#[derive(uniffi::CustomType)]
-#[uniffi(builtin=String)]
 pub struct Uuid {
     val: String,
 }
+
+// Use `url::Url` as a custom type, with `String` as the Builtin
+uniffi::impl_ffi_converter_custom_type!(Url, String);
 
 impl UniffiCustomTypeConverter for Uuid {
     type Builtin = String;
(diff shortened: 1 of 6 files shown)


bendk:
`FfiConverter` supports this by having that `UniFfiTag` parameter.  Could `UniffiCustomTypeConverter` also have that?  I don't have a strong opinion on this one, just wondering out loud.

docs/manual/src/proc_macro/index.md
```diff
@@ -198,12 +198,13 @@ implements the "new type" idiom (ie, a tuple struct with the builtin as the only
...
 For example, this type does not use the "new type" idiom, so must specify the complete implementation.
 ```rust
-#[derive(uniffi::CustomType)]
-#[uniffi(builtin=String)]
 pub struct Uuid {
     val: String,
 }
 
+// Use `url::Url` as a custom type, with `String` as the Builtin
+::uniffi::impl_ffi_converter_custom_type!(Url, String);
```
jplatte, line 206:
```suggestion
uniffi::impl_ffi_converter_custom_type!(Url, String);
```

docs/manual/src/proc_macro/index.md
```diff
@@ -217,11 +218,11 @@ impl UniffiCustomTypeConverter for Uuid {
...
 ```
 
-Whereas this type uses the "new type" idiom, so `UniffiCustomTypeConverter` is automatically derived
+Whereas this type uses the "new type" idiom, so we can use `impl_ffi_converter_custom_newtype` to
+automatically implement `UniffiCustomTypeConverter`.
 
 ```rust
-#[derive(uniffi::CustomType)]
-#[uniffi(newtype=i64)]
+impl_ffi_converter_custom_newtype(NewTypeHandle, i64);
```
jplatte, line 225:
```suggestion
impl_ffi_converter_custom_newtype!(NewTypeHandle, i64);
```

bendk (approved):
Looks good to me, just had one naming suggestion.

PR #1681, opened 2023-07-31
Include why we paniced in our panic message.

mhammond:
A couple of times I've seen uniffi panic without telling me the underlying reason for the panic. Turns out that's easy to fix.


jplatte (commented):
I feel like the `Display` output is better suited than `Debug` when it comes to panic messages, but overall agree that the error should be printed.

jplatte (approved):

PR #1684, opened 2023-08-02
feat: Cancellable `Future`

### LINKED ISSUE #1669: Rust future seems to be leaked when cancelling Kotlin coroutine
Here's a failing test case, there's also a PR for it at #1667:

```rust
use once_cell::sync::Lazy;
use tokio::sync::Mutex as TokioMutex;

pub async fn use_shared_resource() {
    static MUTEX: Lazy<TokioMutex<()>> = Lazy::new(|| TokioMutex::new(()));
    let guard = dbg!(&*MUTEX).lock().await;
    tokio::time::sleep(Duration::from_millis(100)).await;
    drop(guard);
    dbg!();
}
```

```kotlin
runBlocking {
    val job = launch {
        useSharedResource()
    }

    delay(50)
    // this should drop the future and thus release the mutex
    job.cancel()

    // but it doesn't so this doesn't succeed either
    withTimeout(1000) {
        useSharedResource()
    }
}
```

stderr output:

```
[fixtures/futures/src/lib.rs:285] &*MUTEX = Mutex {
    data: (),
}
[fixtures/futures/src/lib.rs:285] &*MUTEX = Mutex {
    data: <locked>,
}
```

showing the second `dbg!()` is never reached due to cancellation, and the second time the function is called, the mutex is locked.

re. the "doesn't succeed" at the end, there seems to be an additional bug that makes `withTimeout` block rather than throw an exception.

cc @bendk @Hywan

Hywan:
This patch should ideally be reviewed commit-by-commit for an easier reviewer experience. All commits contain a descriptive message, so that I don't have to repeat everything here.

It fixes #1669 and closes #1667.

The most important patches are probably 9cf9f111f9c0a30cbe2a447fbb0052099a73f2fa and e488828a71e8c075662e169fdea8309b24febf7f.

In a nutshell: this patch ensures that it's possible to cancel or drop a `RustFuture` along with its inner `Future` properly. A new FFI function is generated per async function, `<name>_uniffi_drop`. I'm fully aware that this patch “breaks” the black-box model of the `RustFuture` from the foreign language point of view. It's however absolutely necessary, otherwise it's impossible to trigger a cancellation or a proper drop of a `RustFuture`.

I also have hope that this patch will fix https://github.com/mozilla/uniffi-rs/pull/1677#issuecomment-1660663156, as now there is a double reference to the `callbackHolder`, which must ensure that the garbage collector won't collect it prematurely.

Finally, this patch focuses on Kotlin, and I'm starting to take a look at the Swift part now. I may update this PR to update Swift if that's necessary. Python will follow.

Follow up feature: We may simplify `Pin<Arc<RustFuture>>` to simply `Arc<RustFuture>` once this PR has been merged, because now `RustFuture::future` contains the `Pin`. Edit: that's actually done in db62d750964dfe27701de44d29a0f23a8d583216.


Hywan:
cc @mhammond since Ben is in holidays.

jmartinesp:
I tried running the tests mentioned in #1677 (repeating those tests 100, 1000 times) which were getting stuck before at the 2nd-3rd iteration and with these changes I got until >50 iterations of repeating the same tests before I canceled them.

uniffi_bindgen/src/bindings/kotlin/templates/AsyncTypes.kt
```diff
@@ -1,12 +1,14 @@
...
-{{ self.add_import("kotlin.coroutines.Continuation") }}
 {{ self.add_import("kotlin.coroutines.resume") }}
 {{ self.add_import("kotlin.coroutines.resumeWithException") }}
+{{ self.add_import("kotlinx.coroutines.suspendCancellableCoroutine") }}
+{{ self.add_import("kotlinx.coroutines.CancellableContinuation") }}
+{{ self.add_import("kotlinx.coroutines.CompletionHandler") }}
 
 {# We use these in the generated functions, which don't have access to add_import() -- might as well add it here #}
-{{ self.add_import("kotlin.coroutines.suspendCoroutine") }}
+// {{ self.add_import("kotlin.coroutines.suspendCoroutine") }}
```
jplatte, line 11: Can this be removed?

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -429,13 +434,11 @@ mod tests {
         fn rust_future_weak(&self) -> Weak<TestRustFuture> {
             // It seems like there's not a great way to get an &Arc from a Pin<Arc>, so we need to
             // do a little dance here
-            Arc::downgrade(&Pin::into_inner(Clone::clone(&self.rust_future)))
+            Arc::downgrade(&Arc::clone(&self.rust_future))
```
jplatte, line 437:
```suggestion
            Arc::downgrade(&self.rust_future)
```

uniffi_bindgen/src/bindings/kotlin/templates/AsyncTypes.kt
Hywan: Sure, thanks!

uniffi_core/src/ffi/rustfuture.rs
Hywan: 🤦 of course! Fixed.

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -236,10 +228,11 @@ where
         // Store 1 in `waker_counter`, which we'll use at the end of this call.
         self.wake_counter.store(1, Ordering::Relaxed);
 
-        // `Pin<&mut>`` from our `UnsafeCell`.  `&mut` it is safe, since this is the only reference we
+        // `Pin<&mut>` from our `UnsafeCell`.  `&mut` it is safe, since this is the only reference we
```
mhammond: nano-nit which already existed: extraneous 'it' after `&mut`

Hywan: 👍

Hywan:
Apparently, in the Kotlin implementation, the GC is doing something weird again. It collects `rustFuturePtr` prematurely. I'm working on this.

Hywan:
Problem solved on Kotlin.

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -434,7 +434,7 @@ mod tests {
         fn rust_future_weak(&self) -> Weak<TestRustFuture> {
             // It seems like there's not a great way to get an &Arc from a Pin<Arc>, so we need to
```
mhammond, line 435: is this comment still correct?

mhammond: heh - well, I guess the comment probably remains correct, but I mean "relevant" :)

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -77,6 +77,7 @@ private inline fun <U> rustCall(callback: (RustCallStatus) -> U): U {
 public class USize(value: Long = 0) : IntegerType(Native.SIZE_T_SIZE, value, true) {
     // This is needed to fill in the gaps of IntegerType's implementation of Number for Kotlin.
     override fun toByte() = toInt().toByte()
+    @Deprecated("`toInt().toChar()` is deprecated")
```
bendk, line 80: Can you link to the jetbrains issue that will make this unneeded?

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -286,36 +287,41 @@ where
...
         }
     }
 
+    /// Cancel or drop the `Future` contained inside `Self`.
+    pub fn cancel_or_drop_inner_future(self: Arc<Self>) {
+        // Increase the `wake_counter` to ensure the future cannot be awaken again.
+        self.wake_counter.fetch_add(1, Ordering::Relaxed);
+
+        // Take the `Future`, leaving `None` in its place.
+        let _future = self.future.write().unwrap().take();
```
bendk, line 308: This ensures `self.future` is dropped, but what about the rest of `self`?  Are we sure that the `Arc<Self>` will eventually be dropped.

One scenario I'm thinking about is if there's already a `wake_callback` scheduled.  In that case, we've called `Arc::into_raw()` with the expectation of calling `Arc::from_raw()` later, but that's not going to happen.  It seems dirty, but maybe if `wake_counter` wasn't 0 before the `fetch_add` call, we should manually decrement the count.

bendk: Thinking about this a bit more, it could be a tricky issue.  On Kotlin, it seems that the scheduled callbacks are never called after a `CoroutineScope` is cancelled, but I don't know if that's true for all the foreign bindings.

bendk:
> I also have hope that this patch will fix https://github.com/mozilla/uniffi-rs/pull/1677#issuecomment-1660663156, as now there is a double reference to the callbackHolder, which must ensure that the garbage collector won't collect it prematurely.

Can you explain this one more?  I don't see where the extra reference gets stored.

bendk (commented):
I like a lot about this PR: the doc updates, the fix for #1677.  But I wonder if we should take a different approach to fixing the cancellation issue.

It seems to me that there's currently a fundamental problem with the ForeignExecutor scheduling: some scheduled calls never happen.  The main cause seems to be that once the executor (AKA the CoroutineScope in Kotlin) is canceled then future scheduled calls will be ignored.  Another cause that I'm not 100% sure about is that a call is successfully scheduled, but the executor is cancelled before that call fires.  Because of this, we're leaking Arc pointers because we schedule a wake callback, but the callback never happens.

If that analysis is correct, then I'd rather update the ForeignExecutor API to handle this.  I think there's 2 changes that need to be made:
  - `ForeignExecutorCallback` should return a value indicating if the callback was scheduled successfully.  We can then check this value in `schedule_do_wake()` and if we see a failure then unleak the Arc pointer.
  - The callbacks themselves should get an additional arg specifying if the executor was cancelled.  `wake_callback()` should check that and if there was a cancellation, it should still call `Arc::from_raw()` to unleak the pointer, but not call `do_wake()`.

I think this will require a lot less changes on the Rust side.  I also think this should be do-able from the Kotlin side.  I think the first bullet is just a matter of checking `isCancelled`.  The second bullet might be a bit more tricky, but it seems to me that there has to be a way to do that.

The main reason I'm suggesting this is that I want to start using the ForeignExecutor for more features (#1675).  I think we're going to run into the same issue there if we don't fix things at the ForeignExecutor level here.  Also, in general, it feels like this PR is working around a broken primitive operation rather than fixing the primitive itself.

bendk:
I'm not totally sure that analysis is correct, feel free to push back if I missed something.

If you do think it's correct, I'd be happy to work on the Rust ForeignExecutor side of things if that would help since I'm already working with that code for #1675.

fixtures/futures/tests/bindings/test_futures.kts
```diff
@@ -203,3 +203,27 @@ runBlocking {
 
     assertApproximateTime(time, 500, "broken sleep")
 }
+
+// Test a future that uses a lock and that is cancelled.
+runBlocking {
+    val job = launch {
+        useSharedResource(100U)
```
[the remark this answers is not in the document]
Hywan: If the cancellation doesn't work, the `withTimeout` will raise a timeout exception. That's how I've been testing this :-).

uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -57,19 +57,43 @@ class {{ type_name }}(
...
+        var rustFuturePtr: Pointer? = null
+        val completionHandlerLock = Mutex(locked = true)
+
         return coroutineScope {
             val scope = this
-            return@coroutineScope suspendCoroutine { continuation ->
+            val completionHandler: CompletionHandler = { _ ->
+                runBlocking {
+                    completionHandlerLock.withLock {
+                        if (rustFuturePtr != null) {
```
jmartinesp, line 76:
```kotlin
rustFuturePtr?.let { ptr -> ... } 
```
Is more idiomatic and you avoid having to force unwrap inside the if.

uniffi_core/src/ffi/rustfuture.rs
Hywan: Comment removed :-).

uniffi_core/src/ffi/rustfuture.rs
Hywan: > Thinking about this a bit more, it could be a tricky issue. On Kotlin, it seems that the scheduled callbacks are never called after a `CoroutineScope` is cancelled, but I don't know if that's true for all the foreign bindings.

Yeah, the way Kotlin handles that is really not simple to fit in the current design of `ForeignExecutor`.

Hywan:
> Can you explain this one more? I don't see where the extra reference gets stored.

I'm not sure. I was surprised to see that this patch fixes the problem, but it wasn't my first intent.

uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
Hywan: Solved, thanks.

Hywan:
> I like a lot about this PR: the doc updates, the fix for #1677. But I wonder if we should take a different approach to fixing the cancellation issue.
> 
> It seems to me that there's currently a fundamental problem with the ForeignExecutor scheduling: some scheduled calls never happen. The main cause seems to be that once the executor (AKA the CoroutineScope in Kotlin) is canceled then future scheduled calls will be ignored. Another cause that I'm not 100% sure about is that a call is successfully scheduled, but the executor is cancelled before that call fires. Because of this, we're leaking Arc pointers because we schedule a wake callback, but the callback never happens.
> 
> If that analysis is correct, then I'd rather update the ForeignExecutor API to handle this. I think there's 2 changes that need to be made:
> 
>     * `ForeignExecutorCallback` should return a value indicating if the callback was scheduled successfully.  We can then check this value in `schedule_do_wake()` and if we see a failure then unleak the Arc pointer.
> 
>     * The callbacks themselves should get an additional arg specifying if the executor was cancelled.  `wake_callback()` should check that and if there was a cancellation, it should still call `Arc::from_raw()` to unleak the pointer, but not call `do_wake()`.
> 
> 
> I think this will require a lot less changes on the Rust side. I also think this should be do-able from the Kotlin side. I think the first bullet is just a matter of checking `isCancelled`. The second bullet might be a bit more tricky, but it seems to me that there has to be a way to do that.
> 
> The main reason I'm suggesting this is that I want to start using the ForeignExecutor for more features (#1675). I think we're going to run into the same issue there if we don't fix things at the ForeignExecutor level here. Also, in general, it feels like this PR is working around a broken primitive operation rather than fixing the primitive itself.

To clarify, I did open the PR because we are hitting this problem hard, and it's blocking us. I know it's not the best fix possible, and I'm OK if you close this and start something from scratch :-). I'm aware this is a dressing on a deeper problem.

Hywan:
@bendk Your suggestion would work well with https://developer.apple.com/documentation/swift/withtaskcancellationhandler(operation:oncancel:) for Swift too. I've a branch to work with this, and it fits pretty well with the current design of the PR. It would fit pretty well with your suggestion too.

Hywan:
@bendk

> The callbacks themselves should get an additional arg specifying if the executor was cancelled. wake_callback() should check that and if there was a cancellation, it should still call Arc::from_raw() to unleak the pointer, but not call do_wake().

The biggest problem with Kotlin is that you don't know when a `Job` has been cancelled, except if your `Continuation` has configured a `invokeOnCancellation` completion handler (`CompletionHandler`). There is no `isCancelled` property, like you have in Swift.

jmartinesp:
> @bendk
> 
> > The callbacks themselves should get an additional arg specifying if the executor was cancelled. wake_callback() should check that and if there was a cancellation, it should still call Arc::from_raw() to unleak the pointer, but not call do_wake().
> 
> The biggest problem with Kotlin is that you don't know when a `Job` has been cancelled, except if your `Continuation` has configured a `invokeOnCancellation` completion handler (`CompletionHandler`). There is no `isCancelled` property, like you have in Swift.

Actually, there is: on one hand, if you already have a reference to the `Job` you can just check [Job.isCancelled](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/is-cancelled.html). If you're inside a coroutine (or a suspend function) you can just check the underlying [coroutineContext.isActive](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/is-active.html) property, it will be false if it's either completed, cancelling or cancelled it I'm not mistaken.

OtaK:
Just a heads-up, this PR (in its current state) seems to introduce a regression for the following situation: 

```rust
// error: future cannot be sent between threads safely
// the trait `Send` is not implemented for `*const c_void`
// note: captured value is not `Send`
// required by a bound in `RustFuture::<T, UT>::new`

#[derive(uniffi::Object)]
pub struct MyFirstObj;

#[uniffi::export]
impl MyFirstObj {
    pub async fn do_thing(&self, second_obj: Arc<MySecondObj>) {}
    //                           ^^^^^^^^^^ has type `*const c_void` which is not `Send`
}

#[derive(uniffi::Object)]
pub struct MySecondObj;

```

Such code used to compile without issues on 0.24.3.

jplatte:
@OtaK as per the discussion above, this PR is unlikely to be merged anyways. But it's interesting that we don't have a test case like that, it should probably be added to the next PR that aims to fix the cancellation problem.

bendk:
> @bendk
> 
> > I like a lot about this PR: the doc updates, the fix for #1677. But I wonder if we should take a different approach to fixing the cancellation issue.
> > It seems to me that there's currently a fundamental problem with the ForeignExecutor scheduling: some scheduled calls never happen. The main cause seems to be that once the executor (AKA the CoroutineScope in Kotlin) is canceled then future scheduled calls will be ignored. Another cause that I'm not 100% sure about is that a call is successfully scheduled, but the executor is cancelled before that call fires. Because of this, we're leaking Arc pointers because we schedule a wake callback, but the callback never happens.
> > If that analysis is correct, then I'd rather update the ForeignExecutor API to handle this. I think there's 2 changes that need to be made:
> > 
> > * `ForeignExecutorCallback` should return a value indicating if the callback was scheduled successfully.  We can then check this value in `schedule_do_wake()` and if we see a failure then unleak the Arc pointer.
> > * The callbacks themselves should get an additional arg specifying if the executor was cancelled.  `wake_callback()` should check that and if there was a cancellation, it should still call `Arc::from_raw()` to unleak the pointer, but not call `do_wake()`.
> > 
> > I think this will require a lot less changes on the Rust side. I also think this should be do-able from the Kotlin side. I think the first bullet is just a matter of checking `isCancelled`. The second bullet might be a bit more tricky, but it seems to me that there has to be a way to do that.
> > The main reason I'm suggesting this is that I want to start using the ForeignExecutor for more features (#1675). I think we're going to run into the same issue there if we don't fix things at the ForeignExecutor level here. Also, in general, it feels like this PR is working around a broken primitive operation rather than fixing the primitive itself.
> 
> To clarify, I did open the PR because we are hitting this problem hard, and it's blocking us. I know it's not the best fix possible, and I'm OK if you close this and start something from scratch :-). I'm aware this is a dressing on a deeper problem.

That makes sense.  I've started some work on this and hoepfully will have a PR soon.

fixtures/futures/tests/bindings/test_futures.kts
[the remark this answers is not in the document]
bendk: I'm seeing weird behavior with this.  Instead of seeing any exception, the program just hangs.  I've had slightly better success if I set the timeout in Rust.  The program still hangs, but I'm able to print out something before.

bendk:
@Hywan are you able to reproduce this issue on swift?  It seems to work for me when I try to write a similar unit test.

bendk:
> @Hywan are you able to reproduce this issue on swift? It seems to work for me when I try to write a similar unit test.

Nevermind about this, I just needed to adjust my timeouts.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
badboy: https://github.com/mozilla/uniffi-rs/issues/1685

We should get that fixed outside of this lengthy PR I guess

bendk: That makes a lot of sense, #1692 does this

bendk:
> To clarify, I did open the PR because we are hitting this problem hard, and it's blocking us. I know it's not the best fix possible, and I'm OK if you close this and start something from scratch :-). I'm aware this is a dressing on a deeper problem.

I just made #1697 to address the deeper problem.  How about closing this one?

PR #1687, opened 2023-08-07
Fix Kotlin callback handler reference handling

bendk:
When we execute a future on Kotlin, we create a `jni.Callback` instance that Rust calls when the Future is ready.  We need to hold a reference to that object or else the JVM will finalize it and when Rust tries to invoke the callback bad things will happen. In the unit tests this causes them to stall because the Kotlin coroutine continutation is never completed.

The `callbackHolder` variable was intended to hold a reference, but it doesn't seem to be working (see [arg0d's analysis](https://github.com/mozilla/uniffi-rs/pull/1677#issuecomment-1660663156)). I belive the issue is that variable isn't actually used anywhere so Kotlin frees it before the coroutine has completed and the reason #1684 fixes the issue is that it uses the variable in the completion handler (https://github.com/mozilla/uniffi-rs/pull/1684/files#diff-f564b605a6ba32d19bcb47128dc9c5c2f01169e2a41fa60ffaaf0b6eea13845cR34)

This commit changes things so instead of using a local variable, there's a global set that stores all active callbacks.  It seems to fix the issue for me when using arg0d's `repeat(1000) { ... }` test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@
 - [Custom Types](https://mozilla.github.io/uniffi-rs/proc_macro/index.html#the-unifficustomtype-derive) are now supported for proc-macros, including a very
   low-friction way of exposing types implementing the new-type idiom.
 
+### What's Fixed
+
+- Kotlin: Fixed low-level issue with exported async APIs
+
 ## v0.24.3 (backend crates: v0.24.3) - (_2023-08-01_)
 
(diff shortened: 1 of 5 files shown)


bendk:
@arg0d can you test if this fixes things for you?

skhamis (approved):
Makes sense to me. Thanks for diving into this!

arg0d:
I'm on vacation right now and don't have my workspace right now, but this looks like should solve the issue.

Issue #1685, opened 2023-08-04
kotlin 1.9 fails to compile bindings


mkobetic:
Attempting to compile kotlin bindings generated with uniffi `v0.24.3` using kotlin 1.9 fails with the following error:
```
error: warnings found and -Werror specified
target/tmp/bindings_ffi-87ed72fb9be5acd3/uniffi/xmtpv3/xmtpv3.kt:272:18: warning: this declaration overrides deprecated member but not marked as deprecated itself. Please add @Deprecated annotation or suppress. See https://youtrack.jetbrains.com/issue/KT-47902 for details
    override fun toChar() = toInt().toChar()
                 ^
test uniffi_foreign_language_testcase_test_android_kts ... FAILED

failures:

---- uniffi_foreign_language_testcase_test_android_kts stdout ----
Creating testing out_dir: /Users/martin/rs/libxmtp/bindings_ffi/target/tmp/bindings_ffi-87ed72fb9be5acd3
Error: running `kotlinc` failed


failures:
    uniffi_foreign_language_testcase_test_android_kts

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 7.55s
```

The warning refers to toChar function in the `USize` integer type:

```
// IntegerType that matches Rust's `usize` / C's `size_t`
public class USize(value: Long = 0) : IntegerType(Native.SIZE_T_SIZE, value, true) {
    // This is needed to fill in the gaps of IntegerType's implementation of Number for Kotlin.
    override fun toByte() = toInt().toByte()
    override fun toChar() = toInt().toChar()
    override fun toShort() = toInt().toShort()
    ...
```

It seems that toChar needs to be either omitted or annotated as deprecated as per https://youtrack.jetbrains.com/issue/KT-47902.

```
% kotlin -version
Kotlin version 1.9.0-release-358 (JRE 20.0.1)
```

mhammond:
Fixed by #1692

PR #1693, opened 2023-08-08
Kotlin: Add deprecated annotation to override of deprecated function

richardhuaaa:
`toChar()` is deprecated in the latest Kotlin, and without this annotation, compiling the generated bindings will generate a warning (which causes an error with `-WError`). More details in https://github.com/mozilla/uniffi-rs/issues/1685

Verified this fixes the issue on my local repo.


richardhuaaa:
Oops, just saw this is already fixed in https://github.com/bendk/uniffi-rs/commit/153a09b2fd9a026ba748dad0020a524028cc57cd

PR #1694, opened 2023-08-08
Followup to #1680, update macro names for custom type macros.

mhammond:
I forgot to re-push before landing that PR. I think the new names are better and I'd like to get this in before contributors start playing with it.


bendk (approved):
:ship:

Issue #1638, opened 2023-07-06
Ability to specify foreign type when exporting via proc macro (e.g. bytes for Vec<u8>)


ijc:
I have a struct  with `Vec<u8>` fields which is exported with `#[derivce(uniffi::Record)]` as well as exported methods which receive `Vec<u8>`. In Kotlin these get rendered as `List<UByte>`, as if they were `sequence u8` in UDL.

If I were using UDL then I could name the type `bytes` and this would be rendered as `ByteArray` in Kotlin which is a little easier to work with Kotlin side (AIUI it's more idiomatic generally but it also avoids signed vs unsigned bytes disconnect from Rust v Kotlin -- right now I seem to have to write `foo.toUByteArray().toByteArray()` on the Kotlin side).

I don't think that proc macros currently allow picking the types.

Would it be possible/desirable to have the proc macros default to bytes for `Vec<u8>` as a special case? (I don't think this makes sense as a per language choice, but I suppose that's an option too).

If not, could there be some other method for overriding the types, e.g. a `#[uniffi(type = bytes)]` type attribute on struct fields and method arguments perhaps? (not sure how that works for return types though?). This seems more general but is surely harder to implement.

mhammond:
This seems tricky.

> Would it be possible/desirable to have the proc macros default to bytes for Vec<u8> as a special case?

Allowing a default in uniffi.toml or via some other mechanism makes some sense, but probably still leaves the general question open - how to specify when you *don't* want the default used?

Supporting *only* `Bytes` seems a bit heavy-handed, and (IMO) is something we'd want to do in UDL too, and for some consumers might be a significant b/w compat burden. Maybe there's a path to that future which can be made less painful though?

`#[uniffi(type = bytes)]` gets a bit ugly for `Option`, `Vec` and `Map` - but maybe that's doable?

I had a vague idea of using a type alias - eg, `type UniffiBytes = Vec<u8>;` and saying that you need to use that, but that only goes so far - there are still parts which are unable to differentiate given it's a simple alias - specifically, it will be (IIUC) impossible in the current code to have a different `FfiConverter` for `UniffiBytes` than for `Vec<u8>` which I think would be a requirement to make this work. Using a "new type" (ie, `type UniffiBytes(Vec<u8>);`) isn't going to fly either.

ijc:
> #[uniffi(type = bytes)] gets a bit ugly for Option, Vec and Map - but maybe that's doable?

[...]

PR #1696, opened 2023-08-09
Render `Vec<u8>` via proc macros as `bytes` not `sequence<u8>`.

ijc:
Based on the discussion in #1638, I'm not sure what the overall consensus there was (we got a little sidetracked by the concept of a default) but this is the very simplest thing which happens to satisfy my use-case in the hopes of making things a little more concrete.

Based on the discussion I think these are the other possibilities which were in the mix, in increasing order of (what I think is the) complexity

1. Plumb through the selection to a global option in `uniffii.toml` affecting all languages (conceptually simple but there's a lot of layers to plumb through to get to `MetadataReader::new()` and pass in the choice I think).
2. Plumb through the selection to a per-language option in `uniffii.toml` (not much different to 1 in terms of complexity I think?)
3. Add a `metadata::codes::TYPE_BYTES` and emit it unconditionally for any instance of a `Vec<u8>` (but how to recognise that in a proc macro in the face of aliases etc?)
4. Add a `metadata::codes::TYPE_BYTES` and do $something (an attribute? but then what about function return types?) in the proc macros to allow per field/argument choice "bytes or sequence" for `Vec<u8>` fields), which would feed into a choice when outputing `TYPE_ID_META` for the field/argument in question.
5. As for 3 but allow an arbitrary UDL type to be given.

Personally I don't need 1, 2, 4 or 5 (and 3 is really just an alternative implementation of this PR I think)

On reflection I think 5 is a huge can of worms wrt lifting and lowering and while it seems conceptually nice and general I can't actually think of a use case beyond bytes anyway.

4 is only a potential option because `bytes` and `sequence<u8>` map to the same thing in the FFI buffer (4 bytes length + that many bytes), I think that is by design and could be relied on?

I think @mhammond was concerned about backwards compatibility for consumers of libraries using this, 1 or 2 might satisfy this? Or would 4 be needed? I think, in general an API which mixes bytes and lists of u8 would seem a bit odd, but backwards compat might be a motivator for that.

I've not touched the UDL side of things, there `sequence<u8>` remains a list of `u8`'s. Mark mentioned perhaps wanting to do this for UDL too? It looks like maybe https://github.com/mozilla/uniffi-rs/blob/1696344a8f1643cfa1a36a3fac201340b0b62a98/uniffi_udl/src/resolver.rs#L120 is where that could happen (and then the utility of the `bytes` type in UDL becomes a bit questionable).

### DIFF
--- a/fixtures/metadata/src/tests.rs
+++ b/fixtures/metadata/src/tests.rs
@@ -151,8 +151,9 @@ mod test_type_ids {
         check_type_id::<Option<u8>>(Type::Optional {
             inner_type: Box::new(Type::UInt8),
         });
-        check_type_id::<Vec<u8>>(Type::Sequence {
-            inner_type: Box::new(Type::UInt8),
+        check_type_id::<Vec<u8>>(Type::Bytes);
+        check_type_id::<Vec<u16>>(Type::Sequence {
+            inner_type: Box::new(Type::UInt16),
         });
         check_type_id::<HashMap<String, u8>>(Type::Map {
             key_type: Box::new(Type::String),
--- a/fixtures/proc-macro/src/callback_interface.rs
+++ b/fixtures/proc-macro/src/callback_interface.rs
@@ -2,13 +2,14 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use crate::{BasicError, Object};
+use crate::{BasicError, Object, RecordWithBytes};
 
 #[uniffi::export(callback_interface)]
 pub trait TestCallbackInterface {
     fn do_nothing(&self);
     fn add(&self, a: u32, b: u32) -> u32;
     fn optional(&self, a: Option<u32>) -> u32;
+    fn with_bytes(&self, rwb: RecordWithBytes) -> Vec<u8>;
     fn try_parse_int(&self, value: String) -> Result<u32, BasicError>;
     fn callback_handler(&self, h: std::sync::Arc<Object>) -> u32;
 }
--- a/fixtures/proc-macro/src/lib.rs
+++ b/fixtures/proc-macro/src/lib.rs
@@ -31,6 +31,11 @@ pub struct Three {
     obj: Arc<Object>,
 }
 
+#[derive(uniffi::Record, Debug, PartialEq)]
+pub struct RecordWithBytes {
+    some_bytes: Vec<u8>,
+}
+
 // An object that's not used anywhere (ie, in records, function signatures, etc)
 // should not break things.
 #[derive(uniffi::Object)]
@@ -98,11 +103,22 @@ fn take_two(two: Two) -> String {
 }
 
 #[uniffi::export]
+fn take_record_with_bytes(rwb: RecordWithBytes) -> Vec<u8> {
+    rwb.some_bytes
+}
+
+#[uniffi::export]
 fn test_callback_interface(cb: Box<dyn TestCallbackInterface>) {
     cb.do_nothing();
     assert_eq!(cb.add(1, 1), 2);
     assert_eq!(cb.optional(Some(1)), 1);
     assert_eq!(cb.optional(None), 0);
+    assert_eq!(
+        cb.with_bytes(RecordWithBytes {
+            some_bytes: vec![9, 8, 7],
+        }),
+        vec![9, 8, 7]
+    );
     assert_eq!(Ok(10), cb.try_parse_int("10".to_string()));
     assert_eq!(
         Err(BasicError::InvalidInput),
@@ -124,6 +140,13 @@ pub struct Zero {
 fn make_zero() -> Zero {
     Zero {
         inner: String::from("ZERO"),
+    }
+}
+
+#[uniffi::export]
+fn make_record_with_bytes() -> RecordWithBytes {
+    RecordWithBytes {
+        some_bytes: vec![0, 1, 2, 3, 4],
     }
 }
 
--- a/fixtures/proc-macro/tests/bindings/test_proc_macro.kts
+++ b/fixtures/proc-macro/tests/bindings/test_proc_macro.kts
@@ -10,6 +10,9 @@ assert(one.inner == 123)
 val two = Two("a")
 assert(takeTwo(two) == "a")
 
+val rwb = RecordWithBytes(byteArrayOf(1,2,3).toUByteArray().toList())
+assert(takeRecordWithBytes(rwb).toUByteArray().toByteArray().contentEquals(byteArrayOf(1, 2, 3)))
+
 var obj = Object()
 obj = Object.namedCtor(1u)
 assert(obj.isHeavy() == MaybeBool.UNCERTAIN)
@@ -20,6 +23,7 @@ assert(enumIdentity(MaybeBool.TRUE) == MaybeBool.TRUE)
 val three = Three(obj)
 
 assert(makeZero().inner == "ZERO")
+assert(makeRecordWithBytes().someBytes.toUByteArray().toByteArray().contentEquals(byteArrayOf(0, 1, 2, 3, 4)))
 
 try {
     alwaysFails()
@@ -42,6 +46,8 @@ class KtTestCallbackInterface : TestCallbackInterface {
     override fun add(a: UInt, b: UInt) = a + b
 
     override fun optional(a: UInt?) = a ?: 0u
+
+    override fun withBytes(rwb: RecordWithBytes) = rwb.someBytes
 
     override fun tryParseInt(value: String): UInt {
         if (value == "force-unexpected-error") {
--- a/fixtures/proc-macro/tests/bindings/test_proc_macro.kts
+++ b/fixtures/proc-macro/tests/bindings/test_proc_macro.kts
@@ -10,8 +10,8 @@ assert(one.inner == 123)
 val two = Two("a")
 assert(takeTwo(two) == "a")
 
-val rwb = RecordWithBytes(byteArrayOf(1,2,3).toUByteArray().toList())
-assert(takeRecordWithBytes(rwb).toUByteArray().toByteArray().contentEquals(byteArrayOf(1, 2, 3)))
+val rwb = RecordWithBytes(byteArrayOf(1,2,3))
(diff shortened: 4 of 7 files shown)


mhammond:
It looks like this means procmacros can no longer use `Vec<u8>`, is that correct? It doesn't really make sense to me to support it in UDL but not procmacros - if we did this I think I'd prefer a clean break for UDL too, but do wonder what others think. Our use of bytes is fairly limited, so on one hand any change like this is unlikely to hurt us much, but on the other hand we aren't really the best to comment on  the pros and cons. If we did take this path, I *think* I mildly prefer (3) because it resolves an ambiguity (there are 2 ways to represent bytes, only one of which actually is used). But as said, I'm really not sure on the direction.

@heinrich5991 created the PR for bytes - I wonder if they have any thoughts?

heinrich5991:
Basically this: https://github.com/mozilla/uniffi-rs/issues/1638#issuecomment-1650317536.

I think it'd be bad if uniffi would be restricted by Rust's type system. Note that e.g. serde also allows annotations to fields that say that these should be serialized as bytes instead of sequence u8 (https://docs.rs/serde_bytes/0.11.12/serde_bytes/):
```rs
use serde::{Deserialize, Serialize};

#[derive(Deserialize, Serialize)]
struct Efficient<'a> {
    #[serde(with = "serde_bytes")]
    bytes: &'a [u8],

    #[serde(with = "serde_bytes")]
    byte_buf: Vec<u8>,
}
```

bendk:
I think there's some more options, maybe they're good?  What do others think?
  - Define a `Bytes` type that wraps `Vec<u8>`.  In some ways this is the cleanest solution, but it might require a lot of changes to the Rust code.
  - Use annotations in the function signatures, similar to the serde example.  This is the best syntax I can come up with:
  
```
pub fn scramble_bytes(#[bytes] input: Vec<u8>) -> #[bytes] Vec<u8> {
 ...
}
```
 - Define the alias `type Bytes = Vec<u8>` in UniFFI.  If the proc-macros see a that `Vec<u8>` is spelled out as `Bytes` by matching the last component of the type path, we force the metadata code to be `TYPE_BYTES`.  This seems kind of nice, but maybe too magical.

heinrich5991:
> * Define a `Bytes` type that wraps `Vec<u8>` in a newtype.  In some ways this is the cleanest solution, but it might require a lot of changes to the Rust code.  (Actually maybe not, as long as `Bytes` derefs to `Vec<u8>`)

This doesn't work so well, as it might require changes to data structures. E.g. a `Vec<Vec<u8>>` cannot easily be transformed into a `Vec<Bytes>`. Essentially every time it's not purely the `Vec<u8>` type, problems are going to arise.

> * Use annotations in the function signatures, similar to the serde example.  This is the best syntax I can come up with:

Now that I think about it, this also doesn't work for nested types very well. How do you annotate a `Option<Vec<u8>>`? Or a `(Vec<u8>, Vec<u8>)` where one `Vec<u8>` is supposed to map to `Bytes` and the other is not?

ijc:
> I think it'd be bad if uniffi would be restricted by Rust's type system.

Specifically on bytes vs sequence<u8> do you have a usecase in mind where a "List<u8>" kind of construct would be preferred over the language specific bytes/data type (`ByteArray` in Kotlin, `Data` in Swift, `bytes` in Python)? If not then it could be justifiable that `Vec<u8>` is "special".

The closest thing Uniffi has to the serde-level flexibility is the `custom` type stuff where you can [specify per-language conversions](https://mozilla.github.io/uniffi-rs/udl/custom_types.html#custom-types-in-the-bindings-code) in `uniffi.toml`. That sort of implies using newtypes, which I think would work with @bendk's "Define a Bytes type that wraps Vec<u8> in a newtype" suggestion -- given that you could use `bindings.kotlin.custom_types.Data` in `uniffi.toml`. I'm unclear if that can be as efficient as the builtin bytes handling though (e.g. without special handling I think it would have to go through the `List<u8>` on the way to `ByteArray`)

> > Use annotations in the function signatures, similar to the serde example. 

> Now that I think about it, this also doesn't work for nested types very well. How do you annotate a Option<Vec<u8>>? Or a (Vec<u8>, Vec<u8>) where one Vec<u8> is supposed to map to Bytes and the other is not?

I think this way lies full support for a UDL-like type syntax -- e.g. `#[uniffi(as = "bytes?"` for `Option<Vec<u8>>`.  (IIRC UDL doesn't support tuples so I'm not sure what that would look like)

Can `#[...]` attributes be attached to function arguments and return type though? I don't know about arguments but I thought not for return type? This is a case which serde doesn't hit, so we don't have prior art there. The only solution I can think of is an attribute on the fn itself with the full prototype spelled out (again, in a UDL like syntax).

ijc:
> I'm unclear if that can be as efficient as the builtin bytes handling though (e.g. without special handling I think it would have to go through the List<u8> on the way to ByteArray)

FWIW I tried this in my project, it's a bit of a tangent and long, so behind the fold.

<details>

Given:

```rust
/// An HTTP body (request or response)
#[derive(Default)]
pub struct HttpBody(Vec<u8>);

uniffi::custom_newtype!(HttpBody, Vec<u8>);
```
and

```toml
[bindings.kotlin.custom_types.HttpBody]
# Name of the type in the Kotlin code
type_name = "ByteArray"
# Classes that need to be imported
imports = []
# Expression to convert the builtin type the custom type.
# List<u8> -> UByteArray -> ByteArray
into_custom = "{}.toUByteArray().toByteArray()"
# Expression to convert the custom type to the builtin type.
# ByteArray -> UByteArray -> List<u8>
from_custom = "{}.toUByteArray().toList()"


[bindings.swift.custom_types.HttpBody]
# Name of the type in the Kotlin code
type_name = "Data"
# Classes that need to be imported
imports = []
# Expression to convert the builtin type the custom type.
# [UInt8] -> Data
into_custom = "Data({})"
# Expression to convert the custom type to the builtin type.
# Data -> [UInt8]
from_custom = "Array({})"
```

we end up generating:

```kotlin
    override fun lift(value: RustBuffer.ByValue): HttpBody {
        val builtinValue = FfiConverterSequenceUByte.lift(value)
        return builtinValue.toUByteArray().toByteArray()
    }
    override fun lower(value: HttpBody): RustBuffer.ByValue {
        val builtinValue = value.toUByteArray().toList()
        return FfiConverterSequenceUByte.lower(builtinValue)
    }
```
and
```swift
    public static func read(from buf: inout (data: Data, offset: Data.Index)) throws -> HttpBody {
        let builtinValue = try FfiConverterSequenceUInt8.read(from: &buf)
        return Data(builtinValue)
    }

    public static func write(_ value: HttpBody, into buf: inout [UInt8]) {
        let builtinValue = Array(value)
        return FfiConverterSequenceUInt8.write(builtinValue, into: &buf)
    }
```

I think in each case that's at lease one more "step" (perhaps including an allocation?) than using `FfiConverterByteArray` or `FfiConverterData`?

</details>

mhammond:
> I think it'd be bad if uniffi would be restricted by Rust's type system.

UniFFI is certainly restricted by Rust's type system, and it infiltrates to every corner. I can agree with a more nuanced take, which I think is the point you were actually making, but there aren't many absolutes here, it's all trade-offs.

Is this summary of the current situation correct:

1) No one has articulated a great use-case for the foreign code wanting a distinction between "bytes" and Vec<u8> - just a slight sense of it being a "shame" if we were unable to have both.
2) Annotations on items works in simple cases, but falls apart on recursive types.
3) We might be able to simulate this using "custom types", although that will create unnatural code in some cases (eg, some times you just want to declare the type of a param as being "bytes" rather than use an invented name to work around the problem)
4) We can probably work around this by using Rust type alias and "conventions" which UniFFI procmacros can identify, meaning the Rust code is more natural in most cases, although not as flexible as some may like.

Is there anything I missed?

ijc:
> Is there anything I missed?

I think that's pretty accurate. I think your point 1 is pretty key, since if there is a usecase for the vec of bytes then that would mean a very different set of trade-offs needed to be considered.

One thing which I think hasn't been mentioned is that the real driver for wanting bytes over vec is really Kotlin/JVM. The JVM having signed bytes by default which means that Kotlin code receiving a `foo List<u8>` needs to do things like `foo.toUByteArray().toByteArray()` (and vice-versa when calling methods) in order to then use standard Kotlin functions for dealing with `bytes`.

AFAICT most other platforms/languages have unsigned bytes so e.g. for Swift a `sequence<u8>` is an `Array(UInt8)` (or maybe a `[UInt8]` -- I'm unclear what the distinction is in Swift). These are generally less problematic in terms of ergonomics for the Swift coders consuming the API, I'm unclear on the runtime perf implications of `Array` vs `Data` (which is what bytes maps to)..

I'm not sure about Python (I@ve not used Uniffi there myself), I think there might be a runtime efficiency win for bytes over a list of u8's, but since the language is duck-typed there is less of an ergonomic issue for coders? 

Does Uniffi deal with any other platforms where bytes are signed?

I don't know what the implications of this "Kotlin-centric" view might be...

mhammond:
A few of us had a chat about this in-person and where we landed was:

* While fully supporting both use-cases was seen as ideal, no one could come up with a credible use-case for someone truly wanting a sequence of u8 integers, and we felt that even in the case where that was actually wanted, the pain of moving from a bytes type really wasn't that bad.
* Hackery to work around it (eg, assuming type-aliasing and handling a special specific string such as `Bytes` or similar) really didn't seem worthwhile given the lack of clear use-cases for the distinction.
* We agreed Kotlin a key driver, but Python and JS both had clear use-cases for supporting bytes over list/array of ints, and assumed that external and future languages probably would too, so we shouldn't consider this a kotlin-specific issue.

So while it would be ideal to support both, we can't see a sane path towards that. Therefore we felt that a patch which caused procmacro-backed code to *only* support bytes would be reasonable if such a patch was possible. It would be fine for udl to continue as-is (ie, support both), although there might be a case to have udl and procmacros converge after some deprecation period.

How does everyone feel about this?

mhammond (approved):
which was a long-winded way of saying thanks for the patch!

ijc:
FWIW this all sounds good to me.

But, sadly, as of today I'm on a new team internally and my immediate usecase for this (indeed uniffi generally) has gone away :sob:. I fully expect to be using Uniffi again in the future though and would expect that nice handling of bytes would be of interest when that day comes.

I've rebased the (trivial) conflict away).

PR #1697, opened 2023-08-09
Avoid memory leaks with async cancellation

bendk:
This adds better support for cancelling Futures (#1669).  The basic issue is that our Rust code leaks an Arc/Box when scheduling calls, expecting to reverse that leakage when the scheduled call is invoked.  However if the foreign executer is cancelled, then the scheduled call is never invoked and the Rust Future is leaked in the middle of the operation.  If that Future is holding on to some shared resource like a Mutex then deadlocks will start happening.

Now, the foreign language code now informs Rust that a scheduled call has been cancelled and the Rust code drops the Arc/Box.

When implementing this I noticed some more issues with our basic async system:
  - Swift uses a new `Task` to schedule calls, which means that we give up on structured concurrency.  Cancelling the top-level task doesn't cancel the Rust future task.  This means the tests are currently passing by accident.
  - Python uses an `EventLoop` to schedule calls, which doesn't have a simple way of informing you if the call was cancelled.

I think the solution to both of these issues will require some more fundamental changes in a future PR

### DIFF
--- a/fixtures/futures/tests/bindings/test_futures.kts
+++ b/fixtures/futures/tests/bindings/test_futures.kts
@@ -207,25 +207,31 @@ runBlocking {
 
 // Test a future that uses a lock and that is cancelled.
 runBlocking {
-    val job = launch {
-        useSharedResource(SharedResourceOptions(releaseAfterMs=100U, timeoutMs=1000U))
+    val time = measureTimeMillis {
+        val job = launch {
+            useSharedResource(SharedResourceOptions(releaseAfterMs=100U, timeoutMs=1000U))
+        }
+
+        // Wait some time to ensure the task has locked the shared resource
+        delay(50)
+        // Cancel the job before the shared resource has been released.
+        job.cancel()
+
+        // Try accessing the shared resource again.  The initial task should release the shared resource
+        // before the timeout expires.
+        useSharedResource(SharedResourceOptions(releaseAfterMs=0U, timeoutMs=1000U))
     }
-
-    // Wait some time to ensure the task has locked the shared resource
-    delay(50)
-    // Cancel the job before the shared resource has been released.
-    job.cancel()
-
-    // Try accessing the shared resource again.  The initial task should release the shared resource
-    // before the timeout expires.
-    useSharedResource(SharedResourceOptions(releaseAfterMs=0U, timeoutMs=1000U))
+    println("useSharedResource: ${time}ms")
 }
 
 // Test a future that uses a lock and that is not cancelled.
 runBlocking {
-    useSharedResource(SharedResourceOptions(releaseAfterMs=100U, timeoutMs=1000U))
+    val time = measureTimeMillis {
+        useSharedResource(SharedResourceOptions(releaseAfterMs=100U, timeoutMs=1000U))
 
-    useSharedResource(SharedResourceOptions(releaseAfterMs=0U, timeoutMs=1000U))
+        useSharedResource(SharedResourceOptions(releaseAfterMs=0U, timeoutMs=1000U))
+    }
+    println("useSharedResource (not canceled): ${time}ms")
 }
 
 // Test that we properly cleaned up future callback references
--- a/fixtures/futures/tests/bindings/test_futures.kts
+++ b/fixtures/futures/tests/bindings/test_futures.kts
@@ -208,25 +208,24 @@ runBlocking {
 // Test a future that uses a lock and that is cancelled.
 runBlocking {
     val job = launch {
-        useSharedResource(100U)
+        useSharedResource(SharedResourceOptions(releaseAfterMs=100U, timeoutMs=1000U))
     }
 
+    // Wait some time to ensure the task has locked the shared resource
     delay(50)
     // Cancel the job before the shared resource has been released.
     job.cancel()
 
-    withTimeout(1000) {
-        useSharedResource(100U)
-    }
+    // Try accessing the shared resource again.  The initial task should release the shared resource
+    // before the timeout expires.
+    useSharedResource(SharedResourceOptions(releaseAfterMs=0U, timeoutMs=1000U))
 }
 
 // Test a future that uses a lock and that is not cancelled.
 runBlocking {
-    useSharedResource(100U)
+    useSharedResource(SharedResourceOptions(releaseAfterMs=100U, timeoutMs=1000U))
 
-    withTimeout(1000) {
-        useSharedResource(100U)
-    }
+    useSharedResource(SharedResourceOptions(releaseAfterMs=0U, timeoutMs=1000U))
 }
 
 // Test that we properly cleaned up future callback references
--- a/fixtures/futures/tests/bindings/test_futures.kts
+++ b/fixtures/futures/tests/bindings/test_futures.kts
@@ -205,5 +205,29 @@ runBlocking {
 }
 
 
+// Test a future that uses a lock and that is cancelled.
+runBlocking {
+    val job = launch {
+        useSharedResource(100U)
+    }
+
+    delay(50)
+    // Cancel the job before the shared resource has been released.
+    job.cancel()
+
+    withTimeout(1000) {
+        useSharedResource(100U)
+    }
+}
+
+// Test a future that uses a lock and that is not cancelled.
+runBlocking {
+    useSharedResource(100U)
+
+    withTimeout(1000) {
+        useSharedResource(100U)
+    }
+}
+
 // Test that we properly cleaned up future callback references
 assert(uniffiActiveFutureCallbacks.size == 0)
--- a/fixtures/futures/tests/bindings/test_futures.swift
+++ b/fixtures/futures/tests/bindings/test_futures.swift
@@ -232,4 +232,33 @@ Task {
 	counter.leave()
 }
 
+// Test a future that uses a lock and that is cancelled.
+counter.enter()
+Task {
+	let task = Task {
+	    try! await useSharedResource(options: SharedResourceOptions(releaseAfterMs: 100, timeoutMs: 1000))
+	}
+
+	// Wait some time to ensure the task has locked the shared resource
+	try await Task.sleep(nanoseconds: 50_000_000)
+	// Cancel the job task the shared resource has been released.
+        //
+        // FIXME: this test currently passes because `test.cancel()` doesn't actually cancel the
+        // operation.  We need to rework the Swift async handling to handle this properly.
+	task.cancel()
+
+	// Try accessing the shared resource again.  The initial task should release the shared resource
+	// before the timeout expires.
+	try! await useSharedResource(options: SharedResourceOptions(releaseAfterMs: 0, timeoutMs: 1000))
+	counter.leave()
+}
+
+// Test a future that uses a lock and that is not cancelled.
+counter.enter()
+Task {
+	try! await useSharedResource(options: SharedResourceOptions(releaseAfterMs: 100, timeoutMs: 1000))
+	try! await useSharedResource(options: SharedResourceOptions(releaseAfterMs: 0, timeoutMs: 1000))
+	counter.leave()
+}
+
 counter.wait()
--- a/fixtures/futures/tests/bindings/test_futures.swift
+++ b/fixtures/futures/tests/bindings/test_futures.swift
@@ -242,9 +242,9 @@ Task {
 	// Wait some time to ensure the task has locked the shared resource
 	try await Task.sleep(nanoseconds: 50_000_000)
 	// Cancel the job task the shared resource has been released.
-        //
-        // FIXME: this test currently passes because `test.cancel()` doesn't actually cancel the
-        // operation.  We need to rework the Swift async handling to handle this properly.
+	//
+	// FIXME: this test currently passes because `test.cancel()` doesn't actually cancel the
+	// operation.  We need to rework the Swift async handling to handle this properly.
 	task.cancel()
 
 	// Try accessing the shared resource again.  The initial task should release the shared resource
--- a/fixtures/futures/tests/bindings/test_futures.swift
+++ b/fixtures/futures/tests/bindings/test_futures.swift
@@ -1,4 +1,4 @@
-import uniffi_futures
+import futures
 import Foundation // To get `DispatchGroup` and `Date` types.
 
 var counter = DispatchGroup()
--- a/uniffi_bindgen/src/bindings/swift/templates/BridgingHeaderTemplate.h
+++ b/uniffi_bindgen/src/bindings/swift/templates/BridgingHeaderTemplate.h
@@ -32,7 +32,7 @@ typedef struct RustBuffer
 typedef int32_t (*ForeignCallback)(uint64_t, int32_t, const uint8_t *_Nonnull, int32_t, RustBuffer *_Nonnull);
 
 // Task defined in Rust that Swift executes
-typedef void (*UniFfiRustTaskCallback)(const void * _Nullable);
+typedef void (*UniFfiRustTaskCallback)(const void * _Nullable, int8_t);
 
 // Callback to execute Rust tasks using a Swift Task
 //
@@ -41,7 +41,7 @@ typedef void (*UniFfiRustTaskCallback)(const void * _Nullable);
 //   delay: Delay in MS
 //   task: UniFfiRustTaskCallback to call
 //   task_data: data to pass the task callback
-typedef void (*UniFfiForeignExecutorCallback)(size_t, uint32_t, UniFfiRustTaskCallback _Nullable, const void * _Nullable);
+typedef int8_t (*UniFfiForeignExecutorCallback)(size_t, uint32_t, UniFfiRustTaskCallback _Nullable, const void * _Nullable);
 
 typedef struct ForeignBytes
 {
--- a/uniffi_bindgen/src/bindings/swift/templates/ForeignExecutorTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/ForeignExecutorTemplate.swift
@@ -1,3 +1,9 @@
+private let UNIFFI_RUST_TASK_CALLBACK_SUCCESS: Int8 = 0
+private let UNIFFI_RUST_TASK_CALLBACK_CANCELLED: Int8 = 1
+private let UNIFFI_FOREIGN_EXECUTOR_CALLBACK_SUCCESS: Int8 = 0
+private let UNIFFI_FOREIGN_EXECUTOR_CALLBACK_CANCELED: Int8 = 1
+private let UNIFFI_FOREIGN_EXECUTOR_CALLBACK_ERROR: Int8 = 2
+
 // Encapsulates an executor that can run Rust tasks
 //
 // On Swift, `Task.detached` can handle this we just need to know what priority to send it.
@@ -35,7 +41,7 @@ fileprivate struct FfiConverterForeignExecutor: FfiConverter {
 }
 
 
-fileprivate func uniffiForeignExecutorCallback(executorHandle: Int, delayMs: UInt32, rustTask: UniFfiRustTaskCallback?, taskData: UnsafeRawPointer?) {
+fileprivate func uniffiForeignExecutorCallback(executorHandle: Int, delayMs: UInt32, rustTask: UniFfiRustTaskCallback?, taskData: UnsafeRawPointer?) -> Int8 {
     if let rustTask = rustTask {
         let executor = try! FfiConverterForeignExecutor.lift(executorHandle)
         Task.detached(priority: executor.priority) {
@@ -43,12 +49,14 @@ fileprivate func uniffiForeignExecutorCallback(executorHandle: Int, delayMs: UIn
                 let nanoseconds: UInt64 = numericCast(delayMs * 1000000)
                 try! await Task.sleep(nanoseconds: nanoseconds)
             }
-            rustTask(taskData)
+            rustTask(taskData, UNIFFI_RUST_TASK_CALLBACK_SUCCESS)
         }
-
+        return UNIFFI_FOREIGN_EXECUTOR_CALLBACK_SUCCESS
+    } else {
+        // When rustTask is null, we should drop the foreign executor.
+        // However, since its just a value type, we don't need to do anything here.
+        return UNIFFI_FOREIGN_EXECUTOR_CALLBACK_SUCCESS
     }
-    // No else branch: when rustTask is null, we should drop the foreign executor. However, since
-    // its just a value type, we don't need to do anything here.
 }
 
 fileprivate func uniffiInitForeignExecutor() {
--- a/uniffi_core/src/ffi/foreignexecutor.rs
+++ b/uniffi_core/src/ffi/foreignexecutor.rs
@@ -114,8 +114,8 @@ impl ForeignExecutor {
 
 /// Low-level schedule interface
 ///
-/// When using this function, take care to ensure that the ForeignExecutor that holds the
-/// ForeignExecutorHandle has not been dropped.
+/// When using this function, take care to ensure that the `ForeignExecutor` that holds the
+/// `ForeignExecutorHandle` has not been dropped.
 pub(crate) fn schedule_raw(
     handle: ForeignExecutorHandle,
     delay: u32,
--- a/uniffi_core/src/ffi/foreignexecutor.rs
+++ b/uniffi_core/src/ffi/foreignexecutor.rs
@@ -39,19 +39,77 @@ unsafe impl Sync for ForeignExecutorHandle {}
 /// bindings should release the reference to the executor that was reserved for Rust.
 ///
 /// This callback can be invoked from any thread, including threads created by Rust.
+///
+/// The callback should return one of the `ForeignExecutorCallbackResult` values.
 pub type ForeignExecutorCallback = extern "C" fn(
     executor: ForeignExecutorHandle,
     delay: u32,
     task: Option<RustTaskCallback>,
     task_data: *const (),
-);
+) -> i8;
+
+/// Result code returned by `ForeignExecutorCallback`
+#[repr(i8)]
+#[derive(Debug, PartialEq, Eq)]
+pub enum ForeignExecutorCallbackResult {
+    /// Callback was scheduled successfully
+    Success = 0,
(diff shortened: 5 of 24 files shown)


fixtures/futures/tests/bindings/test_futures.kts
```diff
@@ -205,5 +205,28 @@ runBlocking {
...
+    // before the timeout expires.
+    useSharedResource(SharedResourceOptions(releaseAfterMs=0U, timeoutMs=1000U))
+}
+
+// Test a future that uses a lock and that is not cancelled.
+runBlocking {
+    useSharedResource(SharedResourceOptions(releaseAfterMs=100U, timeoutMs=1000U))
+
+    useSharedResource(SharedResourceOptions(releaseAfterMs=0U, timeoutMs=1000U))
+}
```
Hywan, line 229: You may want to measure the timing here, otherwise it's hard to tell when the test is failing.

fixtures/futures/tests/bindings/test_futures.kts
```diff
@@ -205,5 +205,28 @@ runBlocking {
...
+
+    // Wait some time to ensure the task has locked the shared resource
+    delay(50)
+    // Cancel the job before the shared resource has been released.
+    job.cancel()
+
+    // Try accessing the shared resource again.  The initial task should release the shared resource
+    // before the timeout expires.
+    useSharedResource(SharedResourceOptions(releaseAfterMs=0U, timeoutMs=1000U))
+}
```
Hywan, line 222: You may want to measure the timing here, otherwise it's hard to tell when the test is failing.

fixtures/futures/tests/bindings/test_futures.swift
```diff
@@ -232,4 +232,33 @@ Task {
...
+	let task = Task {
+	    try! await useSharedResource(options: SharedResourceOptions(releaseAfterMs: 100, timeoutMs: 1000))
+	}
+
+	// Wait some time to ensure the task has locked the shared resource
+	try await Task.sleep(nanoseconds: 50_000_000)
+	// Cancel the job task the shared resource has been released.
+        //
+        // FIXME: this test currently passes because `test.cancel()` doesn't actually cancel the
+        // operation.  We need to rework the Swift async handling to handle this properly.
```
Hywan, line 247:
```suggestion
    //
    // FIXME: this test currently passes because `test.cancel()` doesn't actually cancel the
    // operation.  We need to rework the Swift async handling to handle this properly.
```

fixtures/futures/tests/bindings/test_futures.swift
```diff
@@ -232,4 +232,33 @@ Task {
 	counter.leave()
 }
 
+// Test a future that uses a lock and that is cancelled.
+counter.enter()
+Task {
+	let task = Task {
+	    try! await useSharedResource(options: SharedResourceOptions(releaseAfterMs: 100, timeoutMs: 1000))
+	}
```
Hywan, line 240: Are we sure the `Task` runs the code inside the block as soon as it's built? Or should we have a `await task.value` to ensure the `Task` runs?

uniffi_bindgen/src/bindings/swift/templates/BridgingHeaderTemplate.h
```diff
@@ -32,7 +32,7 @@ typedef struct RustBuffer
 typedef int32_t (*ForeignCallback)(uint64_t, int32_t, const uint8_t *_Nonnull, int32_t, RustBuffer *_Nonnull);
 
 // Task defined in Rust that Swift executes
-typedef void (*UniFfiRustTaskCallback)(const void * _Nullable);
+typedef void (*UniFfiRustTaskCallback)(const void * _Nullable, int8_t);
```
Hywan, line 35: May we prefer a `uint8_t` here?

```suggestion
typedef void (*UniFfiRustTaskCallback)(const void * _Nullable, uint8_t);
```

uniffi_bindgen/src/bindings/swift/templates/BridgingHeaderTemplate.h
```diff
@@ -41,7 +41,7 @@ typedef void (*UniFfiRustTaskCallback)(const void * _Nullable);
 //   delay: Delay in MS
 //   task: UniFfiRustTaskCallback to call
 //   task_data: data to pass the task callback
-typedef void (*UniFfiForeignExecutorCallback)(size_t, uint32_t, UniFfiRustTaskCallback _Nullable, const void * _Nullable);
+typedef int8_t (*UniFfiForeignExecutorCallback)(size_t, uint32_t, UniFfiRustTaskCallback _Nullable, const void * _Nullable);
```
Hywan, line 44: May we prefer a `uint8_t` here?

```suggestion
typedef uint8_t (*UniFfiForeignExecutorCallback)(size_t, uint32_t, UniFfiRustTaskCallback _Nullable, const void * _Nullable);
```

uniffi_bindgen/src/bindings/swift/templates/ForeignExecutorTemplate.swift
```diff
@@ -1,3 +1,9 @@
+private let UNIFFI_RUST_TASK_CALLBACK_SUCCESS: Int8 = 0
+private let UNIFFI_RUST_TASK_CALLBACK_CANCELLED: Int8 = 1
+private let UNIFFI_FOREIGN_EXECUTOR_CALLBACK_SUCCESS: Int8 = 0
+private let UNIFFI_FOREIGN_EXECUTOR_CALLBACK_CANCELED: Int8 = 1
+private let UNIFFI_FOREIGN_EXECUTOR_CALLBACK_ERROR: Int8 = 2
```
Hywan, line 5:
```suggestion
private let UNIFFI_RUST_TASK_CALLBACK_SUCCESS: UInt8 = 0
private let UNIFFI_RUST_TASK_CALLBACK_CANCELLED: UInt8 = 1
private let UNIFFI_FOREIGN_EXECUTOR_CALLBACK_SUCCESS: UInt8 = 0
private let UNIFFI_FOREIGN_EXECUTOR_CALLBACK_CANCELED: UInt8 = 1
private let UNIFFI_FOREIGN_EXECUTOR_CALLBACK_ERROR: UInt8 = 2
```

uniffi_bindgen/src/bindings/swift/templates/ForeignExecutorTemplate.swift
```diff
@@ -35,20 +41,22 @@ fileprivate struct FfiConverterForeignExecutor: FfiConverter {
 }
 
 
-fileprivate func uniffiForeignExecutorCallback(executorHandle: Int, delayMs: UInt32, rustTask: UniFfiRustTaskCallback?, taskData: UnsafeRawPointer?) {
+fileprivate func uniffiForeignExecutorCallback(executorHandle: Int, delayMs: UInt32, rustTask: UniFfiRustTaskCallback?, taskData: UnsafeRawPointer?) -> Int8 {
```
Hywan, line 44:
```suggestion
fileprivate func uniffiForeignExecutorCallback(executorHandle: Int, delayMs: UInt32, rustTask: UniFfiRustTaskCallback?, taskData: UnsafeRawPointer?) -> UInt8 {
```

uniffi_core/src/ffi/foreignexecutor.rs
```diff
@@ -39,19 +39,40 @@ unsafe impl Sync for ForeignExecutorHandle {}
...
-);
+) -> i8;
+
+/// Returned by `ForeignExecutorCallback` if the callback was scheduled successfully
+pub const FOREIGN_EXECUTOR_CALLBACK_SUCCESS: i8 = 0;
+/// Returned by `ForeignExecutorCallback` if the callback couldn't be scheduled because the
+/// foreign executor is canceled/closed.
+pub const FOREIGN_EXECUTOR_CALLBACK_CANCELED: i8 = 1;
+/// Returned by `ForeignExecutorCallback` for all other errors
+pub const FOREIGN_EXECUTOR_CALLBACK_ERROR: i8 = 2;
```
Hywan, line 57:
```suggestion
) -> u8;

/// Returned by `ForeignExecutorCallback` if the callback was scheduled successfully
pub const FOREIGN_EXECUTOR_CALLBACK_SUCCESS: u8 = 0;
/// Returned by `ForeignExecutorCallback` if the callback couldn't be scheduled because the
/// foreign executor is canceled/closed.
pub const FOREIGN_EXECUTOR_CALLBACK_CANCELED: u8 = 1;
/// Returned by `ForeignExecutorCallback` for all other errors
pub const FOREIGN_EXECUTOR_CALLBACK_ERROR: u8 = 2;
```

uniffi_core/src/ffi/foreignexecutor.rs
```diff
@@ -39,19 +39,40 @@ unsafe impl Sync for ForeignExecutorHandle {}
...
 // Option<RustTaskCallback> should use the null pointer optimization and be represented in C as a
 // regular pointer.  Let's check that.
 static_assertions::assert_eq_size!(usize, Option<RustTaskCallback>);
 
 /// Callback for a Rust task, this is what the foreign executor invokes
-pub type RustTaskCallback = extern "C" fn(*const ());
+///
+/// The task will be passed the `task_data` passed to `ForeignExecutorCallback` in addition to one
+/// of the `RUST_TASK_CALLBACK_*` codes.
+pub type RustTaskCallback = extern "C" fn(*const (), i8);
```
Hywan, line 67:
```suggestion
pub type RustTaskCallback = extern "C" fn(*const (), u8);
```

uniffi_core/src/ffi/foreignexecutor.rs
```diff
@@ -39,19 +39,40 @@ unsafe impl Sync for ForeignExecutorHandle {}
...
+
+/// Passed to a RustTaskCallback function when the executor successfully invokes them
+pub const RUST_TASK_CALLBACK_SUCCESS: i8 = 0;
```
Hywan, line 70:
```suggestion
pub const RUST_TASK_CALLBACK_SUCCESS: u8 = 0;
```

uniffi_core/src/ffi/foreignexecutor.rs
```diff
@@ -39,19 +39,40 @@ unsafe impl Sync for ForeignExecutorHandle {}
...
+/// Passed to a RustTaskCallback function when the executor is cancelled.
+///
+/// This signals that any progress using the executor should be halted.  In particular, Futures
+/// should not continue to progress.
+pub const RUST_TASK_CALLBACK_CANCELED: i8 = 1;
```
Hywan, line 75:
```suggestion
pub const RUST_TASK_CALLBACK_CANCELED: u8 = 1;
```

uniffi_core/src/ffi/foreignexecutor.rs
```diff
@@ -114,20 +135,31 @@ impl ForeignExecutor {
...
-/// ForeignExecutorHandle has not been dropped.
+/// When using this function, take care to ensure that the `ForeignExecutor` that holds the
+/// `ForeignExecutorHandle` has not been dropped.
 pub(crate) fn schedule_raw(
     handle: ForeignExecutorHandle,
     delay: u32,
     callback: RustTaskCallback,
     data: *const (),
-) {
+) -> i8 {
```
Hywan, line 145:
```suggestion
) -> u8 {
```

uniffi_core/src/ffi/foreignexecutor.rs
```diff
@@ -114,20 +135,31 @@ impl ForeignExecutor {
...
     (get_foreign_executor_callback())(handle, delay, Some(callback), data)
 }
 
+// Log errors when we don't schedule_raw() doesn't return FOREIGN_EXECUTOR_CALLBACK_SUCCESS
+pub(super) fn log_schedule_raw_error(result: i8) {
```
Hywan, line 150:
```suggestion
pub(super) fn log_schedule_raw_error(result: u8) {
```

uniffi_core/src/ffi/foreignexecutor.rs
Hywan, line 70: I'm wondering if that would not be better to have a:

```rust
#[repr(u8)]
pub enum RustTaskCallbackResult {
    Success = 0,
    Canceled = 1,
}
```

Thoughts?

Hywan (commented):
Overall it looks good. I think it's important to measure the timing in the tests, that's the most accurate way to know whether the locks inside the “share resource” function are released or not.

I'm also surprised you don't need [`withTaskCancellationHandler`](https://developer.apple.com/documentation/swift/withtaskcancellationhandler(operation:oncancel:)) in the Swift code, but maybe I'm misunderstanding the whole picture :-).

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -54,10 +54,10 @@
 //! ) {
 //!     ::uniffi::call_with_output(uniffi_call_status, || {
 //!         let uniffi_rust_future = RustFuture::<_, bool, crate::UniFFITag,>::new(
-//!             future: hello(),
-//!             executor,
-//!             callback,
-//!             callback_data,
+//!             future: hello(), // the future!
```
Hywan, line 57:
```diff
- future: hello(), // the future!
+ hello(), // the future!
```

fixtures/futures/tests/bindings/test_futures.swift
bendk: The task is scheduled as soon as it's built and should run during the `Task.sleep()` call below.  I don't think we can `await` the task itself, since we want to cancel it.

uniffi_bindgen/src/bindings/swift/templates/BridgingHeaderTemplate.h
bendk: The reason we've historically used signed ints is because Kotlin only had experimental support (which was inherited from Java).  I think that Kotlin has dropped the "experimental" label, but I think I prefer sticking with signed ints for now.  Maybe in a future commit we can switch them all to unsigned.

uniffi_core/src/ffi/foreignexecutor.rs
bendk: Seems good to me.  I used constants mostly just to match the other FFI code, but I don't think it's important to do that.

bendk:
Thanks for the review, just pushed some updates to this.

> I'm also surprised you don't need [withTaskCancellationHandler](https://developer.apple.com/documentation/swift/withtaskcancellationhandler(operation:oncancel:)) in the Swift code, but maybe I'm misunderstanding the whole picture :-).

I think we're going to need it eventually, but it's working right now by accident.  The way the Swift code currently works, there's no way to cancel the future -- it always runs to completion.  If you cancel the parent task, the Rust code will continue to move forward, releasing the lock in this case, then it will try to complete the canceled Swift task which is basically a no-op.

I hope to fix this soon, but I think this PR already has enough changes.

fixtures/futures/tests/bindings/test_futures.swift
Hywan: OK!

uniffi_bindgen/src/bindings/swift/templates/BridgingHeaderTemplate.h
Hywan: OK, I didn't know, thanks for the explanation.

bendk:
> Hence the need to add measure the timing in the test. Cancelling a `Task` should be immediate. In this particular case, the lock is freed because of a timer, but the lock might never be freed if the `Future` isn't properly dropped.

Could you explain this one more?  I'm not sure about what the potential issue is and how measuring timing addresses it.

Hywan:
In the Swift case, `Future` cancellation isn't supported. However, the test still passes because the mutex used in the test case will always be released. It the mutex is released after 1s, then 2 calls to the function locking the mutex, takes 2s. If the first calls is immediately cancelled, then it takes 1s to execute the test, which means that the `Future` is indeed cancelled properly, and that the mutex is released properly. Without any time measurement, for the Swift case, the test can pass but it doesn't mean that the `Future` is cancelled as expected. In this very particular case, the mutex is released after a certain amount of time for testing convenience, which can be misleading. In real life, a mutex is unlikely to be released after a certain amount of time :-).

jplatte:
@Hywan I think your concern was already adressed in the PR description. There will be a follow-up PR for cancellation in Swift.

> * Swift uses a new `Task` to schedule calls, which means that we give up on structured concurrency. Cancelling the top-level task doesn't cancel the Rust future task. This means the tests are currently passing by accident.
> * Python uses an `EventLoop` to schedule calls, which doesn't have a simple way of informing you if the call was cancelled.
> 
> 
> I think the solution to both of these issues will require some more fundamental changes in a future PR

Maybe to avoid confusion, the PR title should be updated to "Async cancellation for Kotlin".

Also, small note regarding the eventual Swift implementation, there's also the static variable [`Task.isCancelled`](https://developer.apple.com/documentation/swift/task/iscancelled-swift.type.property). I'm not super familiar with the current async implementation, but if it's somewhat similar to the previous impl where there was a poll method invoked from foreign code periodically, checking that static var right before each poll and cancelling (dropping) the future in case it's `false` should be all that's necessary for Swift cancellation.

Hywan (approved):

bendk:
> In the Swift case, Future cancellation isn't supported. However, the test still passes because the mutex used in the test case will always be released. It the mutex is released after 1s, then 2 calls to the function locking the mutex, takes 2s. If the first calls is immediately cancelled, then it takes 1s to execute the test, which means that the Future is indeed cancelled properly, and that the mutex is released properly. Without any time measurement, for the Swift case, the test can pass but it doesn't mean that the Future is cancelled as expected. In this very particular case, the mutex is released after a certain amount of time for testing convenience, which can be misleading. In real life, a mutex is unlikely to be released after a certain amount of time :-).

I think this is the case for all languages actually.  AFAICT the sequence of events is something like:

  - 0ms: Foreign code starts a child thread
  - 1ms: child thread locks the Mutex
  - 2ms child thread sleeps for 100ms
  - 50ms: main thread cancels the parent task /coroutine context / event loop.
     - Note: there's nothing signalling the Rust code to cancel and release the mutex.  If a library wanted to do this they would need to setup a some sort of side channel for this and manually handle it themselves.  For example, setting up a shared `tokio::CancellationToken` or a boolean flag that is periodically checked in a work loop.  I can't think of any generic way for UniFFI to handle this.
  - 50ms: main thread awaits `use_shared_resource` in the main thread
  - 100ms: child thread wakes up and the function completes
  - 101ms: child thread future waker is invoked which schedules a call to the foreign executor to poll the future
    - Since the foreign executor has been cancelled, this call will never go through.
 - 102ms: child thread Future is dropped, unlocking the mutex
      -  Before, bugs in the Kotlin/Python code would prevent this because they would leak a pointer when scheduling the call with a cancelled foreign executor.
 - 103ms: The main thread `use_shared_resource` call locks the mutex.  
 - 104ms: the main thread `use_shared_resource` call completes, unlocking the mutex and completing the test.

I think the issue with Swift would only come up in more complicated cases where the future was awaiting multiple child futures or something like that.  In that case, I think the Kotlin/Python code could notice earlier that the executor was cancelled and stop trying to advance the future, but the Swift code would just keep driving it forward until completion.  IOW, it has to be a case where there are multiple calls to `Future::poll()` after the parent task is cancelled.

bendk:
> Also, small note regarding the eventual Swift implementation, there's also the static variable [Task.isCancelled](https://developer.apple.com/documentation/swift/task/iscancelled-swift.type.property).

I think this is the right approach, but it would mean that the executor handle needs to reference the parent `Task` rather than the `UniFffiForeignExecutor` struct that we define.  This is getting into the fundamental changes I mentioned.  I think I tried to make `ForeignExecutor` do too much and we need to split it into 2 different types:
  - `AsyncContext` represents the foreign context where the async Rust call is being invoked from.  On Swift this would be the parent `Task` value.
  - `DispatchQueue` a worker queue / thread pool that runs synchronous code and returns a Future for the result.  On Swift this would be a struct, that works like the current `ForeignExecutor`.

Python would also benefit from this split.  `AsyncContext` would correspond to `EventLoop` and `DispatchQueue` would correspond to `futures.Executor`.  However, I think on Kotlin they would both correspond to `CoroutineContext`

bendk:
I updated the title to reflect what I think is really the heart of this change.  Any concerns with merging this one?

mhammond (approved):

jplatte:
> Note: there's nothing signalling the Rust code to cancel and release the mutex.  If a library wanted to do this they would need to setup a some sort of side channel for this and manually handle it themselves.  For example, setting up a shared `tokio::CancellationToken` or a boolean flag that is periodically checked in a work loop.  I can't think of any generic way for UniFFI to handle this.

There doesn't need to be anything "signalling" the Rust code to cancel. Unless the future is currently being polled on another thread¹, you just drop it and that's how it is cancelled.

¹ obviously you must not drop any value, future or not, while it is being accessed by another thread

> > Also, small note regarding the eventual Swift implementation, there's also the static variable [Task.isCancelled](https://developer.apple.com/documentation/swift/task/iscancelled-swift.type.property).
> 
> I think this is the right approach, but it would mean that the executor handle needs to reference the parent `Task` rather than the `UniFffiForeignExecutor` struct that we define.

It's a **static** variable that's automatically available within every coroutine. There is also a non-static variable `isCancelled` that you can check on a task object from outside of it, though I would have assumed in our case the Rust future is always polled from within a coroutine.

bendk:
> It's a **static** variable that's automatically available within every coroutine. There is also a non-static variable `isCancelled` that you can check on a task object from outside of it, though I would have assumed in our case the Rust future is always polled from within a coroutine.

That last part is the issue.  Suppose that the Rust future spins up a thread or awaits a future that's running in a tokio runtime.  Then the waker is likely to be woken up in those Rust-created threads and `Task.isCancelled` is probably not going to work.

jplatte:
Okay, I can kind of see what you mean; I'll trust you on that. What do you think about the first part of my previous message?

bendk:
> Okay, I can kind of see what you mean; I'll trust you on that. What do you think about the first part of my previous message?

:) I feel like I only kind of understand it.  Async is complicated.

On the first part: I think agree with you, there doesn't need to be any way to signal a cancellation.  However, I think that some people might intuitively believe that when you cancel the parent task from the foreign language the Rust code stops running and that's not always going to be the case and in some cases you're going to want to setup that side-channel for cancellation.

jplatte:
Okay, let me try to make it a bit clearer then: I am 100% certain that we *can* make it so that when the (foreign) parent task is cancelled, the Rust future is cancelled properly. All we need to do when the parent task is cancelled is make sure that the future isn't currently being polled by another thread (I see it is held inside an `UnsafeCell`, so I'm not sure how to this would be achieved), drop it, and ensure nothing else retains a reference to the now-dropped future.

bendk:
My point is that "cancelled properly" can mean a lot of things to a lot of people and there's times when we won't be cancelling properly according to some definition.  For example, suppose you spin up a thread like in the [tokio::oneshot](https://docs.rs/tokio/latest/tokio/sync/oneshot/index.html#examples) example.  There's no way for UniFFI to tell that thread that it should end early.

That said, maybe we could be doing better.   In the example, the `tokio::timeout` call in that this example runs is still active when the parent task is cancelled.  That means tokio will hold a reference to our waker, which keeps the future alive -- at least I think that's how it works.  Part of me thinks that this means the future must stay alive until the timer fires.  But writing this out, I wonder if the waker could hold a weakref and that would fix things.

Here's another example that I'm very confused about.  Let's assume we add support for async callback interface methods and we have the following "stack":
  1. An async function is running on the foreign side
  2. That function calls an async Rust function
  3. The Rust function calls an async callback interface method

If 1 is canceled, can we ensure that 3 is also cancelled?  I think it's possible if to set up some kind of parent-child relationship between the various tasks, but that would probably need to lean on static methods like [Task.withUnsafeCurrentTask](https://developer.apple.com/documentation/swift/withunsafecurrenttask(body:)#parameters).  But then what happens if the async callback method was called from a tokio thread?  In that case, I don't think there is a "current task" in swift.  Maybe if we want to set up that kind of structured concurrency, we have to start assuming that everything is happening inside the foreign executor, and any calls into tokio or other runtimes would need some sort of bridging function.  I'm not really sure what's best.

jplatte:
> My point is that "cancelled properly" can mean a lot of things to a lot of people and there's times when we won't be cancelling properly according to some definition.

I don't get what you want to express with this. Cancelling a future has a well-defined meaning in Rust, it is exactly equivalent to dropping the future. 

> For example, suppose you spin up a thread like in the [tokio::oneshot](https://docs.rs/tokio/latest/tokio/sync/oneshot/index.html#examples) example, but it's actually doing some real work in a loop.  There's no way for UniFFI to tell that thread that it should end early.

Yes, tokio task `JoinHandle`s have detach-on-drop semantics, same as `std`'s thread `JoinHandle`s. If people want to spawn a task that cancellation is propagated to, they can wrap the `JoinHandle` in a type that cancels it on drop, and keep that wrapper around on the future's stack. This is not something that UniFFI should be concerned with though, I think.

jplatte:
Possibly helpful reading: https://blog.yoshuawuyts.com/async-cancellation-1/#cancelling-futures

bendk:
> Cancelling a future has a well-defined meaning in Rust, it is exactly equivalent to dropping the future.

Yeah, but the question is how to bridge between that and foreign languages that use cancellation callbacks.  I think the system should be:
  - The foreign language holds the only reference to the future.  When the parent task is cancelled, it drops the reference.  This means the waker can only hold a weak reference to the future.
  - When we implement async callback interface methods, we should store a cancellation handle/callback alongside the future.  If the future is dropped before it's ready, then we cancel the foreign task.

I think this should handle both the examples we're talking about here and my async callback interface example above.  It's also much closer to the design that Hywan initially implemented and also suggested initially for cancellation.  I pushed too hard trying to make Rust handle everything.

jplatte:
> The foreign language holds the only reference to the future.  When the parent task is cancelled, it drops the reference.  This means the waker can only hold a weak reference to the future.

Sounds okay.

(I don't really understand why wakers need to reference the RustFuture, since in pure Rust w/o foreign languages they always reference a task and our RustFuture is never the toplevel future of a foreign task AFAIU, but that's not too say that I doubt the implementation, just an observation.)

Issue #1460, opened 2023-01-26
Reduce Swift warnings


notmandatory:
Due to #1350 which escapes all Swift identifiers (except for namespace functions) users of generated Swift bindings now get the follow warnings: 

```
Keyword '<name>' does not need to be escaped in argument list
```

There's currently no way in Swift to suppress these warnings and I was unable to find any active proposals to add a way to do it.

One way to avoid these is to only escape known Swift keywords. The best list of keywords I've found is in the blog post: ["All the Reserved Keywords in Swift"](https://betterprogramming.pub/all-the-reserved-keywords-in-swift-17efcfaa3f3e). 

I'm opening this issue for feedback on if this is worth fixing and as a starting point for myself or anyone else who wants to do it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-234)
┆Issue Number: UNIFFI-234

mhammond:
The warnings annoy me too, but not enough to actually make the change :) I think having a list of keywords would be fine - [that's what Python does](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/python/gen_python/mod.rs#L30)

notmandatory:
Thanks for the link, good to see hard-coding a word list isn't too hack-y a way to handle it.

badboy:
I'm still mad that there's no way in Swift to disable warnings. Wargh.

mhammond:
yeah, I dug into this looking for an easy answer a little while ago - but nope :(

ijc:
Between #1701, #1703 and #1704 there are no more Swift warnings in the Uniffi test fixtures.

Specifically #1701 addresses the issue mention here regarding unnecessary quoting of non-keywords.

PR #1701, opened 2023-08-10
swift: Only quote keywords in contexts where that is required

ijc:
Determined from https://docs.swift.org/swift-book/documentation/the-swift-programming-language/lexicalstructure/#Keywords-and-Punctuation

Previously lots of warning were generated of the form:
```
/tmp/target/tmp/uniffi-example-custom-types-2b18f735d09116c0/custom_types.swift:427:13: warning: keyword 'url' does not need to be escaped in argument list
            `url`: FfiConverterTypeUrl.read(from: &buf),
            ^   ~
```

When compiling generated Swift code.

Swift appears to be fairly tollerant of keyswords in some contexts, in particular:
* only a small subset of keywords need to be quoted when a name is used as an argument label when calling a function.
* enum variant names do not need to be quoted when they are used, only when they are defined.

Fixes: #1460

---

TBH I'm a bit worried about this PR re the chances of regressions: I'm not a Swift expert in any way and the contexts where quoting is required vs not required is a little bit of a mystery to me... I can say that the tests all still pass and in particular `uniffi-fixture-keywords-swift` works, but I don't know how comprehensive the coverage is there. One way to mitigate might be to add some config options to the `bindings.swift` bit of `uniffi.toml` to add keywords to the various lists (including maybe adding one for `enum_variant_swift` which in this PR does no quoting since it isn't needed it seems), i.e. a chicken switch of sorts..

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/gen_swift/mod.rs
+++ b/uniffi_bindgen/src/bindings/swift/gen_swift/mod.rs
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use once_cell::sync::Lazy;
 use std::borrow::Borrow;
 use std::cell::RefCell;
 use std::collections::{BTreeSet, HashMap, HashSet};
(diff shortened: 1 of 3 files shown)


mhammond:
Looks good to me, but that might just be because I don't know Swift very well :) James, are you able to have a look at this?

badboy (approved):
I ran this against the Glean code base and it got rid of all warnings related to keywords. r+

Issue #1637, opened 2023-07-06
Inconsistent renaming of `*Error` to `*Exception` in Kotlin bindings


ijc:
Since Uniffi does not yet support async callback interfaces we have an API which is passing a callback to a sync Kotlin function, the handler is then called when some async work dispatched by the Kotlin code completes. We then wrap that in a `Future` on the Rust side.

So the API is something like:

```rust

#[derive(uniffi::Record)]
pub struct Success { ... }

#[derive(Debug, thiserror::Error, uniffi::Error)]
pub enum Error {
    Foo,
    Bar,
    Unexpected { reason: String }, // Could be exposed as From<uniffi::UnexpectedUniFFICallbackError>
}

#[derive(uniffi::Object)]
pub struct Handler { ... }

#[uniffi::export]
impl Handler {
    fn success(&self, s: Success) { ... }

    fn error(&self, err: MyError) { ... }
}

#[uniffi::export(callback_interface)]
pub trait CB: Send + Sync {
    fn make_request(
        &self,
        handler: Arc<Handler>,
    );
}
```

Plus some scaffolding which lets rust code `.await` on the `Handler` being called after it is passed to `make_request` (too much to reproduce here, I can provide if it helps clarify though).

This is simulating:

```rust
#[uniffi::export(callback_interface)]
pub trait CB: Send + Sync {
    async fn make_request(
        &self,
    ) -> Result<Success, MyError>;
}
```

Note that `MyError` is also used as a `-> Result<_, MyError>` result from other regular methods exposed via UniFFi. (Aside, I think we'd also like to use `#[uniffi(handle_unknown_callback_error)]` so that `make_request(...)-> Result<(), MyError>` works in case that method throws before it manages to dispatch whatever async work it is doing on the Kotlin side, right now we have a `catch` on the Kotlin side which calls `handler.error` in that case to avoid that throw panicing in the bindings).

The problem is that `MyError` is being rewritten to `MyException` in some but not all cases, i.e. [`error_name`](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs#L268) is inconsistently applied.

<details>

<summary>Reproducer in `fixtures/proc-macro` using `BasicError` (sometimes `BasicException` in Kotlin)</summary>

```diff
diff --git a/fixtures/proc-macro/src/callback_interface.rs b/fixtures/proc-macro/src/callback_interface.rs
[...]

PR #1706, opened 2023-08-11
kotlin: Use correct name when generating method which takes an error as an ar…

ijc:
…gument

The name must be renamed `FooError` -> `FooException` to match transforms done elsewhere.

Without the other changes the test changes failed with:

```
uniffi-fixture-proc-macro-fb98904aa2d05965/uniffi/fixture/proc_macro/uniffi_proc_macro.kt:819:26: error: unresolved reference: BasicError
    fun `takeError`(`e`: BasicError): UInt
                         ^
uniffi-fixture-proc-macro-fb98904aa2d05965/uniffi/fixture/proc_macro/uniffi_proc_macro.kt:878:35: error: unresolved reference: BasicError
    override fun `takeError`(`e`: BasicError): UInt =
                                  ^
```

We need to apply the renaming to arg lists too.

Fixes: #1637

---

/cc @mhammond -- thanks for the tip re ci.is_name_used_as_error()!

### DIFF
--- a/fixtures/proc-macro/src/callback_interface.rs
+++ b/fixtures/proc-macro/src/callback_interface.rs
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use crate::BasicError;
+use crate::{BasicError, Object};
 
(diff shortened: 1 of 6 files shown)


mhammond (approved):
This makes sense to me, thanks!

PR #1707, opened 2023-08-13
swift: Add cargo swift to "External resources" section

antoniusnaumann:
I built a little cargo plugin, that helps with packaging a UniFFI project as Swift Package. The motivation behind this is to allow using Rust code in iOS and macOS apps without requiring additional build scripts or configuration. Building a Swift Package seems to be a good solution for this, since the Swift Package Manager is well integrated into XCode.

Additionally, this enables to distribute Rust + UniFFI crates as Swift Packages for use by other people unfamiliar with Rust or UniFFI.

I would appreciate it if my plugin would be listed in the README.


badboy (approved):

badboy:
Hey! I just tried this out and wow, this is great! It just ... works! Thanks for adding it.

Issue #1708, opened 2023-08-14
Methods named "invoke" for Kotlin callbacks using jna.Callback might be safer if called "callback"


gmulhearn-anonyome:
hi, this is the first time i've ran into this, and only occurred with the introduction of `UniFfiForeignExecutorCallback`. When running debug android tests, i get this exception when UniFFI instance initializes:
```
java.lang.IllegalArgumentException: Callback must implement a single public method, or one public method named 'callback'
at com.sun.jna.CallbackReference.getCallbackMethod(CallbackReference.java:418)
at com.sun.jna.CallbackReference.getCallbackMethod(CallbackReference.java:388)
at com.sun.jna.CallbackReference.<init>(CallbackReference.java:283)
at com.sun.jna.CallbackReference.getFunctionPointer(CallbackReference.java:506)
at com.sun.jna.CallbackReference.getFunctionPointer(CallbackReference.java:483)
at com.sun.jna.Function.convertArgument(Function.java:558)
at com.sun.jna.Function.invoke(Function.java:345)
at com.sun.jna.Library$Handler.invoke(Library.java:265)
at java.lang.reflect.Proxy.invoke(Proxy.java:1006)
at $Proxy7.uniffi_foreign_executor_callback_set(Unknown Source)
at com.package.FfiConverterForeignExecutor.register$package_debug(UniffiWrapper.kt:2981)
at com.package._UniFFILib$Companion$INSTANCE$2.invoke(UniffiWrapper.kt:396)
at com.package._UniFFILib$Companion$INSTANCE$2.invoke(UniffiWrapper.kt:391)
```

I added some break lines in `CallbackReference`, essentially it has the following behaviour:
1. if there is only 1 public method (i.e. `UniFfiForeignExecutorCallback.invoke`) then register that as the callback handler
2. if there is more than 1 pub method, than find the public method called `"callback"` and use that,
3. if "callback" is not found, throw the exception seen above ^

I added some breaklines, and found that when running debug tests, jacoco is injecting another method into the UniFfiForeignExecutorCallback object: `$jacocoInit` https://github.com/jacoco/jacoco/issues/946 .

this occurs in at least jna@5.10 and jna@5.13 from my testing.

as that jacoco issue suggests, jna should really be ignoring synethic methods so that this error does not occur, however i wonder if UniFFI could improve by calling it's Callback methods `"callback"` instead? I've tried changing the generated file to `callback` and had success.

thnx

[...]

PR #1709, opened 2023-08-14
Issue 1708: jna callback method "invoke" -> "callback"

gmulhearn-anonyome:
see #1708


bendk (approved):
This looks good to me.  I ran the tests locally and they're passing, so I think we can merge this one.

PR #1714, opened 2023-08-18
swift: add quotes as required for enum variants (fixes #1713)

### LINKED ISSUE #1713: Regression due to missing quotes in generated Swift bindings
This Rust code generates Swift bindings that don't compile anymore since https://github.com/mozilla/uniffi-rs/pull/1701 has been merged:

```
uniffi::include_scaffolding!("api");

#[derive(uniffi::Enum)]
pub enum State {
    Init,
    Jeej,
}
```

This results in the following Swift code and compile error:

```
// Note that we don't yet support `indirect` for enums.
// See https://github.com/mozilla/uniffi-rs/issues/396 for further discussion.
public enum State {
    
    case `init`
    case jeej
}

public struct FfiConverterTypeState: FfiConverterRustBuffer {
    typealias SwiftType = State

    public static func read(from buf: inout (data: Data, offset: Data.Index)) throws -> State {
        let variant: Int32 = try readInt(&buf)
        switch variant {
        
        case 1: return .init
        
        case 2: return .jeej
        
        default: throw UniffiInternalError.unexpectedEnumCase
        }
    }

    public static func write(_ value: State, into buf: inout [UInt8]) {
        switch value {
        
        
        case .init: // ERROR IS HERE
            writeInt(&buf, Int32(1))
        
        
        case .jeej:
            writeInt(&buf, Int32(2))
        
        }
    }
}
```

```
math.swift:362:25: error: 'State' cannot be constructed because it has no accessible initializers
        case 1: return .init
                       ~^~~~
math.swift:374:15: error: member 'init(nilLiteral:)' expects argument of type '()'
        case .init:
              ^
```

bnjbvr:
This fixes the issue I've spotted in #1713, by adding quotes until the generated swift bindings compile.

### DIFF
--- a/fixtures/keywords/swift/src/keywords.udl
+++ b/fixtures/keywords/swift/src/keywords.udl
@@ -4,6 +4,7 @@ namespace keywords_swift {
 
 enum case {
     "internal",
+    "init",
 };
 
 [Enum]
--- a/fixtures/keywords/swift/src/lib.rs
+++ b/fixtures/keywords/swift/src/lib.rs
@@ -7,6 +7,7 @@
 #[allow(non_camel_case_types)]
 pub enum r#case {
     r#internal,
+    r#init,
 }
 
 #[allow(non_camel_case_types)]
--- a/uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
@@ -13,7 +13,7 @@ public struct {{ ffi_converter_name }}: FfiConverterRustBuffer {
         let variant: Int32 = try readInt(&buf)
         switch variant {
         {% for variant in e.variants() %}
-        case {{ loop.index }}: return .{{ variant.name()|enum_variant_swift }}{% if variant.has_fields() %}(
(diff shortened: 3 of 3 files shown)


mhammond:
Any change you could make a change to one of the fixtures which demonstrates the problem?

bnjbvr:
> Any chance you could make a change to one of the fixtures which demonstrates the problem?

I've done just that, with an enum that contained an `Init` variant.

mhammond:
I'm very sorry, but I was confused as to what the problem was. I believe you can reproduce the problem even easier and in a place better suited to this test - the patch below should also fail:

```
--- a/fixtures/keywords/kotlin/src/keywords.udl
+++ b/fixtures/keywords/kotlin/src/keywords.udl
@@ -35,6 +35,7 @@ interface false {
 [Error]
 enum class {
     "object",
+    "init",
 };
 
 [Error]
```

The coverall change *is* what I asked for :( But the fact it only covers a keyword for one language isn't clear from the addition, so would you mind reverting this coverall change, and verify that this both reproduces the problem and that your patch fixes it?

Again, sorry for the run-around!

bnjbvr:
No worries, done (the issue appeared with Swift, so I put the new keyword there instead of in the Kotlin fixture file as shown in your comment), and the CI seems to be happy about it!

mhammond (approved):
Thanks!

PR #1717, opened 2023-08-22
implemented buildMap todo for kotlin

Dmitry-Borodin:
there was todo to use buildMap, it is stable since Kotlin 1.6
implemented todo

note - I was not able to run tests from doker image with 
```
error: failed to open: /mounted_workdir/target/debug/.cargo-lock
Caused by:
  Permission denied (os error 13)
```
But I hope they will run in PR flow now.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/MapTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/MapTemplate.kt
@@ -5,8 +5,8 @@ public object {{ ffi_converter_name }}: FfiConverterRustBuffer<Map<{{ key_type_n
         return buildMap<{{ key_type_name }}, {{ value_type_name }}> {
             val len = buf.getInt()
             repeat(len) {
-                val k = { { key_type|read_fn } }(buf)
-                val v = { { value_type|read_fn } }(buf)
+                val k = {{ key_type|read_fn }}(buf)
+                val v = {{ value_type|read_fn }}(buf)
                 this[k] = v
             }
         }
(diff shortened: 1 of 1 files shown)


uniffi_bindgen/src/bindings/kotlin/templates/MapTemplate.kt
```diff
@@ -2,15 +2,14 @@
...
     override fun read(buf: ByteBuffer): Map<{{ key_type_name }}, {{ value_type_name }}> {
-        // TODO: Once Kotlin's `buildMap` API is stabilized we should use it here.
-        val items : MutableMap<{{ key_type_name }}, {{ value_type_name }}> = mutableMapOf()
-        val len = buf.getInt()
-        repeat(len) {
-            val k = {{ key_type|read_fn }}(buf)
-            val v = {{ value_type|read_fn }}(buf)
-            items[k] = v
+        return buildMap<{{ key_type_name }}, {{ value_type_name }}> {
+            val len = buf.getInt()
```
jmartinesp, line 6: There is a `buildMap` variant that takes a `capacity` parameter and should be more performant, maybe the `len` value could be used for that?

```kotlin
val len = buf.getInt()
return buildMap<{{ key_type_name }}, {{ value_type_name }}>(len) {
```

Dmitry-Borodin: Thank you, didn't notice that. Fixed.

Dmitry-Borodin:
build failing because of rust dependencies
```
error: package `clap_derive v4.4.0` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.69.0
Either upgrade to rustc 1.70.0 or newer, or use
cargo update -p clap_derive@4.4.0 --precise ver
where `ver` is the latest version of `clap_derive` supporting rustc 1.69.0
```

badboy:
If you rebase on `main` CI will pass (https://github.com/mozilla/uniffi-rs/pull/1720 landed)

bendk (approved):
CI is passing, let's take this.

Issue #1153, opened 2022-01-20
Add top-level extern "C" functions for RustBuffer


bendk:
We should define top-level extern "C" functions for rust buffers.  This would be all the functions that are generated for the scaffolding code, but without the hash included:

  - `rust_buffer_alloc()`
  - `rust_buffer_free()`
  - `rust_buffer_from_bytes()`
  - `rust_buffer_reserve()`

Having these as top-level functions would allow the bindings code to access them without depending on any particular scaffolding.  This would allow us to define a `RustBuffer` wrapper in the bindings language that could be shared between all UniFFIed components.  I'm hoping to do this for the desktop-JS bindings.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-130)
┆epic: Rust Components on Desktop
┆friendlyId: UNIFFI-130
┆sprintEndDate: 2022-02-18

rfk:
> This would allow us to define a RustBuffer wrapper in the bindings language that could be shared
> between all UniFFIed components

(Hi, just continuing to drop in from time to time with historical context! 👋🏻)

One of the problems we had with this sort of thing in the past, was that we wanted to be able to support loading different Rust components from different `.so` files, for a "non-megazorded" development flow. In that sort of situation it's very important to use the individual allocation functions associated with each individual component, because each `.so` would be maintaining its own separate allocation state.

I don't *think* we have this problem any more, because we expect to always be deploying in a way that all Rust components have been compiled into the same library, and are thus sharing a single Rust allocator. Still, it's worth carefully thinking about whether there are any edge-cases where you might accidentally use the wrong allocator, or try to free memory from one allocator in a different one, or similar.

bendk:
The historical context is amazing, thanks so much.

This change is for the second attempt at desktop-JS bindings.  For those bindings, everything is going to be in the same library -- UniFFI, all the components, the rust scaffolding, and he C++ JS code.  No dynamic loading at all.  In this situation I think a single set of `RustBuffer` functions makes sense.  But I agree that it makes a lot less sense when you're dynamically loading from `.so` libraries.

So I think what I'll try is to have these functions as `extern "C"` functions in the uniffi library, but continue to have the scaffolding code have their functions prefixed with the component name and UDL hash.

PR #1718, opened 2023-08-23
Mangle rust buffer functions

arg0d:
Not mangling these functions cause duplicate symbol errors when attempting to link 2 static libraries containing different uniffi versions. The error also appears when linking 2 static libraries containing the same uniffi version, but when the libraries are compiled on different architecture machies, e.g. MacOS Intel and MacOS Arm.

These Rust buffer functions are already exposed with a namespace prefix in RustBuffer.rs, and foreign language bindings use these prefixed functions.

### DIFF
--- a/uniffi_core/src/ffi/rustbuffer.rs
+++ b/uniffi_core/src/ffi/rustbuffer.rs
@@ -207,11 +207,7 @@ impl Default for RustBuffer {
 /// to the foreign-language code as a `RustBuffer` struct. Callers must eventually
 /// free the resulting buffer, either by explicitly calling [`uniffi_rustbuffer_free`] defined
 /// below, or by passing ownership of the buffer back into Rust code.
-#[no_mangle]
-pub extern "C" fn uniffi_rustbuffer_alloc(
-    size: i32,
-    call_status: &mut RustCallStatus,
-) -> RustBuffer {
+pub fn uniffi_rustbuffer_alloc(size: i32, call_status: &mut RustCallStatus) -> RustBuffer {
     rust_call(call_status, || {
         Ok(RustBuffer::new_with_size(size.max(0) as usize))
     })
@@ -225,8 +221,7 @@ pub extern "C" fn uniffi_rustbuffer_alloc(
 /// # Safety
 /// This function will dereference a provided pointer in order to copy bytes from it, so
 /// make sure the `ForeignBytes` struct contains a valid pointer and length.
-#[no_mangle]
-pub unsafe extern "C" fn uniffi_rustbuffer_from_bytes(
+pub fn uniffi_rustbuffer_from_bytes(
     bytes: ForeignBytes,
     call_status: &mut RustCallStatus,
 ) -> RustBuffer {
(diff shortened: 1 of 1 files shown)


badboy:
Now that I looked at this I was wondering why we did this in the first place.
Then I read the [comment just above it](https://github.com/mozilla/uniffi-rs/blob/0ffa4a7d1e9b786f30922fc6f2e8f96415fccd3b/uniffi_core/src/ffi/rustbuffer.rs#L202-L204).

And as it turns out [we indeed make use of that in mozilla-central](https://searchfox.org/mozilla-central/rev/f979f15eaeef504bfdcd27f033323d62b51986cd/toolkit/components/uniffi-js/OwnedRustBuffer.cpp#26-27), so landing this as is will break our use case.

jplatte:
I find this part of the comment that you linked very confusing:

> which is needed in the case where we create multiple dylib artifacts since each dylib will have its own allocator.

Is this for the case that the dylib is compiled with a custom [`global_allocator`](https://doc.rust-lang.org/reference/runtime.html#the-global_allocator-attribute)? I was only aware of there being linking problems with the no-mangle symbols from a non-root crate in older Rust versions. I guess it makes sense though (which is sad because that means we'll never be able to get rid of `uniffi::setup_scaffolding!()`).

arg0d:
Is there anyway that you could introduce these no-mangle rust buffer functions in mozilla central project, instead of in uniffi? I.e., do something similar like `RustBuffer.rs` does.

These changes are quite important for us. Considering that right now these are very minor changes, we can maintain them in our own fork. Hopefully this won't become more complicated in the future versions. I was checking 0.24.0, and there was an additional no-mangle function `uniffi_foreign_executor_callback_set`, that looks like it could be problematic. I explored the code a bit, and the only way to fix that would be a breaking change to `0.24.0`. This is problematic, because currently we base `uniffi-bindgen-cs` and `uniffi-bindgen-go` on our fork, and if our fork is not compatible with upstream, that means C# and Go generators would not integrate well into uniffi ecosystem. I also checked `main` branch, and I didn't see this function anymore, so I hope this will not be a problem for 0.25.0.

For the record, I'm copy-pasting a message from the chat explaining why its important for us to be able to link static libraries containing different uniffi versions:
> Our product contains 3 libraries which are versioned separately. The libraries are developed by 3 separate teams, and the teams run their own release schedules. The libraries are consumed by siloed product-application teams (Apple, Windows, Android, Linux), which integrate the libraries at their own pace. Library consumers sometimes want to delay updating one of the libraries for their own reasons, e.g. bugs or simply lack of priority in the moment. Other times they want to update one of the libraries quickly, e.g. bug fixes. If we went with megazord approach for all our consumers, then we would have to create tailored megazord configuration for each consumer, and the consumers would have to maintain this configuration file containing their desired library versions.

arg0d:
> I also checked main branch, and I didn't see this function anymore, so I hope this will not be a problem for 0.25.0.

I was wrong, `uniffi_foreign_executor_callback_set` still exists in `main`. I was checking in the wrong branch.

badboy:
> Is there anyway that you could introduce these no-mangle rust buffer functions in mozilla central project, instead of in uniffi? I.e., do something similar like `RustBuffer.rs` does.

yes, I will look into it this week and see what the best way forward is.

badboy:
So for context: these functions were added back in https://github.com/mozilla/uniffi-rs/issues/1153
There's additional context there that indeed we avoided this in the past due to the clashes.
We thus landed only these new functions, but kept the wrappers for everything else.

IMO a way forward here is to have the `no_mangle` part behind another feature, so that m-c can enable it, but others won't be affected.

`uniffi_foreign_executor_callback_set` is another thing though, that's used by the bindings. Maybe it should be part of the scaffolding instead? I need to look into that.

bendk:
Thinking through this more, I think your plan of moving it to the scaffolding could work.  IIUC: There still would be a global function in `uniffi_core`, but the `export "C"` function would be defined in the bindings and would call into that.  That seems fine to me.

badboy:
> Thinking through this more, I think your plan of moving it to the scaffolding could work. IIUC: There still would be a global function in `uniffi_core`, but the `export "C"` function would be defined in the bindings and would call into the `uniffi_core` one. That seems fine to me.

Yup, essentially the same wrapping as we do for the rustbuffer functions.

PR #1719, opened 2023-08-24
Avoid using serde just for printing debug representations

mhammond:
There were a lot of places where we derived `Serialize` just so a command-line tool could dump a JSON repr of a CI. Everything dumped also derives `Debug`, so now we just prefer using `{:#?}` and avoid using serde except where we actually need it (which is largely limited to parsing toml)


bendk (approved):
Seems good to me.  As long as the output is human-readable it should be fine.

mhammond:
https://github.com/clap-rs/clap/issues/5088

epage:
RE 5088

> Clap is working as-expected.
> 
> Libraries should not be making policy decisions for their dependents by constraining upper bounds on version requirements. In fact doing so can be harmful to the entire ecosystem. We only do it for `clap`/`clap_builder`/`clap_derive` due to a special case related to proc macros. See https://doc.rust-lang.org/nightly/cargo/reference/specifying-dependencies.html#multiple-requirements for more details.
> 
> Instead, if you care about MSRV, you likely should be committing your lockfile. Note that Cargo has changed its guidance (blog post on this is upcoming). See the new guidance on the nightly docs: https://doc.rust-lang.org/nightly/cargo/faq.html#why-have-cargolock-in-version-control

uniffi/Cargo.toml
```diff
@@ -20,7 +20,8 @@ uniffi_core = { path = "../uniffi_core", version = "=0.24.1" }
 uniffi_macros = { path = "../uniffi_macros", version = "=0.24.1" }
 anyhow = "1"
 camino = { version = "1.0.8", optional = true }
-clap = { version = "4", features = ["cargo", "std", "derive"], optional = true }
+clap = { version = "=4.3.19", features = ["cargo", "std", "derive"], optional = true }
+clap_lex = "<=0.5.0" # work around https://github.com/clap-rs/clap/issues/5088, can be removed when MSR is 1.70.
```
epage, line 24: Looks like this is intended to be used as a library.  Putting upper bounds on version requirements can break people's applications.

See https://doc.rust-lang.org/nightly/cargo/reference/specifying-dependencies.html#multiple-requirements

mhammond: Thanks - but I don't really understand the advice here - without the upper bound we get the error about not being on 1.70. Committing the lockfile might be something we can investigate, but the implications for things like mozilla-central need to be considered, and isn't something we've considered given existing advice.

epage: > isn't something we've considered given existing advice.

To clarify, is this referring to the previous advice that cargo gave compared to the new/current advice?

> I don't really understand the advice here - without the upper bound we get the error about not being on 1.70. 

Committing lockfiles is the primary solution for this.

mhammond: > To clarify, is this referring to the previous advice that cargo gave compared to the new/current advice?

The previous advice. I'm not sure how we would have discovered the "current" advice given it seems to be behind a URL which implies nightly and as you mentioned, a blog post with the new advice remains pending.

> Committing lockfiles is the primary solution for this.

So projects using the previous advice are expected to be broken in this way simply due to the new clap release and should unbreak themselves by discovering the new advice and deciphering that advice into realizing they must now commit a Cargo.lock? And that's considered "expected"?

epage: Many of us had given up on the old advice for a while now because of the challenges in maintaining MSRV. One of the reasons I pushed for the change in advice is so there wasn't a divide between those that felt comfortable ignoring it and those contorting things to make the advice work (especially by doing damaging changes like to oddball version requirements) but to instead reflect the reality of how to make things work.

epage: btw we did a first pass at raising awareness by ensuring TWiR included our FCP on this (which they don't normally notice our FCPs)

badboy (approved):

PR #1720, opened 2023-08-25
Commit the Cargo.lock file to build with known working versions of dependencies

badboy:
This offloads _some_ care to consumers, as by default they might now pull in transitive dependencies that break UniFFI's MSRV. What they will need to do is `cargo update -p $dep --precise $ver` where `$ver` is a version that is MSRV-compatible (and is probably the one uniffi's Cargo.lock file contains)

---

Alternative to the pinning in #1719 
Briefly discussed with Mark in chat.
cargo is changing recommendations around that: https://doc.rust-lang.org/nightly/cargo/faq.html#why-have-cargolock-in-version-control

I certainly had confusion in the past due to different versions use locally than in CI until I realized a mismatch in dependencies.
I'm fairly confident this is fine getting pulled into m-c. The lock file will be ignored and only m-c's lockfile counts. m-c will need to ensure transitive dependencies are within MSRV. 
Yes, that offloads that responsibility to other places.
But other consumers with higher MSRVs can use those updated transitive dependencies.


epage:
For more background and analysis on this change in advice, see https://github.com/rust-lang/cargo/pull/12382 and the issue it closed.

mhammond (approved):

PR #1723, opened 2023-08-29
fix: Bindgen not picking up correct symbols

OtaK:
Hi! 

I don't have a repro repo at hand this time (unfortunately, the repro env is quite complex to produce and I don't have much time :( ).
So this is a complicated one but in short: 

* In a certain configuration (see our FFI here https://github.com/wireapp/core-crypto/tree/6169072fe65fc3ccebfbe583db3bc101b88180e4/crypto-ffi)
* and in a pure proc-macro environment,

*Some* symbols were incorrectly named in the generated bindings; It seems the issue was centered around callback interfaces

* Given a library symbol that gets emitted as `uniffi_core_crypto_ffi_fn_init_callback_corecryptocallbacks`
	* Defined here https://github.com/wireapp/core-crypto/blob/6169072fe65fc3ccebfbe583db3bc101b88180e4/crypto-ffi/src/generic.rs#L649
* The resulting Swift, Python and Kotlin bindings were calling the following symbol: `uniffi_CoreCrypto_fn_init_callback_corecryptocallbacks`

This was most likely because of a disparity between the namespace "name" and its "ffi-name" - The issue being that both methods `namespace()` and `ffi_namespace()` were resolving to the same thing. 

My conclusion is that UniFFI "forgot" the actual name of the library (i.e. in my example `core_crypto_ffi`) and was using only the library namespace (i.e. `CoreCrypto`), resulting in the correct `.so` being generated but the bindings calling a non-existent symbol.

Feel free to ask any questions, and if really needed I can take the time to setup a repro repo in the next few days.

Cheers!

### DIFF
--- a/uniffi_bindgen/src/interface/mod.rs
+++ b/uniffi_bindgen/src/interface/mod.rs
@@ -114,12 +114,18 @@ impl ComponentInterface {
     /// Add a metadata group to a `ComponentInterface`.
     pub fn add_metadata(&mut self, group: uniffi_meta::MetadataGroup) -> Result<()> {
         if self.types.namespace.is_empty() {
-            self.types.namespace = group.namespace.name.clone();
-        } else if self.types.namespace != group.namespace.name {
+            self.types.namespace = group.namespace.clone();
+        } else if self.types.namespace.crate_name.is_empty() {
+            self.types.namespace.crate_name = group.namespace.crate_name.clone();
+        } else if self.types.namespace.name.is_empty() {
+            self.types.namespace.name = group.namespace.name.clone();
+        } else if self.types.namespace != group.namespace {
             bail!(
-                "Namespace mismatch: {} - {}",
+                "Namespace mismatch: {}::{} - {}::{}",
+                group.namespace.crate_name,
                 group.namespace.name,
-                self.types.namespace
+                self.types.namespace.crate_name,
+                self.types.namespace.name,
             );
         }
         // Unconditionally add the String type, which is used by the panic handling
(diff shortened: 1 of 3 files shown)


OtaK:
Seems I broke all UDL things with my PR, I'll fix it and then mark my PR as ready.

mhammond:
Are you mixing UDL and procmacros? If so, I think you are hitting the same problem I'm looking at, which is summarized as "the namespace and the crate name must be identical" and easily demonstrated by changing the [namespace in this fixture](https://github.com/mozilla/uniffi-rs/blob/main/fixtures/proc-macro/src/proc-macro.udl#L1C1-L2C1) to be anything else other than the value it has.

OtaK:
@mhammond No, our environment is pure proc-macros, leading me to think that *something* that usually gets informed correctly by UDL (for example, a `namespace` block) breaks when in a proc-macro environment without a UDL file to support it.

And your theory makes sense, as the `#[uniffi::Object]` we export (equivalent to a UDL `namespace`) is `CoreCrypto` while the crate name/cdylib name is `core_crypto_ffi`

mhammond:
hrm, I see you call `uniffi::setup_scaffolding!("CoreCrypto");`, so the namespace should be set to that with the crate name being `core_crypto_ffi` - I suspect you are hitting a variant of the same problem I described though - but it might be easier to fix in your case than it is when UDL is actually involved, but I'm not sure. It looks like none of our fixtures/examples has a call to `setup_scaffolding()` which specifies a name different from the crate name, so maybe that's a key to the STR.

OtaK:
> It looks like none of our fixtures/examples has a call to setup_scaffolding() which specifies a name different from the crate name, so maybe that's a key to the STR.

Indeed it seems it's the trick. We did a change on our side to provide `core_crypto_ffi` as namespace and then rename the output Swift/Kotlin binding from `core_crypto_ffi.[swift|kt]` to `CoreCrypto.[swift|kt]` and it seems to do the trick. 

Anyway I'll close this PR as my "fix" actually breaks everything and doesn't even work for us. 
It breaks the `uniffi_contract_version` FFI symbols as well.

PR #1724, opened 2023-08-29
Fixup Cargo.lock to reflect changes in how serde is used

mhammond:
I could have sworn I updated the new Cargo.lock, but apparently not!


skhamis (approved):
lol maybe from the recent add of Cargo.lock here? https://github.com/mozilla/uniffi-rs/pull/1720/files

in any case LGTM! :shipit:

PR #1725, opened 2023-08-29
Remove unused RustBuffer.rs scaffolding template

mhammond:
Apparently this template is no longer used, but it confused me when I tried to modify it.


bendk (approved):
Looks good to me.  @badboy, will this affect you work in #1718?

badboy:
> Looks good to me. @badboy, will this affect you work in #1718?

Good question. I did not have the time to finalize my work on that yet, but will do so this week.

badboy (approved):
This is fine. Its use was removed in f847c32ec945447be6476301fcd88ac056b5a6dc. The macros now generate this scaffolding inline.
It won't affect my work on the no_mangle stuff.

PR #1735, opened 2023-09-06
Mangle rustbuffer & foreign_executor_callback_set functions

badboy:
Not mangling these functions cause duplicate symbol errors when
attempting to link 2 static libraries containing different uniffi
versions. The error also appears when linking 2 static libraries
containing the same uniffi version, but when the libraries are
compiled on different architecture machies, e.g. MacOS Intel and MacOS
Arm.

These Rust buffer functions are already exposed with a namespace prefix
in RustBuffer.

---

This is the alternative to #1718 that should help us land this in m-c as well (I still need to test this).


bendk (approved):
Looks great, thanks for doing this.

badboy:
I now tested this in m-c and it works (we need to add `uniffi_core` as a dep of `uniffi-bindgen-gecko-js` and enable the `extern-rustbuffer` feature)

PR #1740, opened 2023-09-07
Require Send + Sync for object types

jplatte:
Without this, it is possible to send types that contain non-threadsafe constructs like `RefCell` over FFI, and call methods that use them from multiple threads.

Both Send and Sync are required on the inner type for Arc to implement Send itself, which is what we must guarantee to be able to move an Arc over FFI, since we can't guarantee it is kept on the same thread inside the calling code.

### DIFF
--- a/fixtures/uitests/tests/ui/interface_not_sync_and_send.stderr
+++ b/fixtures/uitests/tests/ui/interface_not_sync_and_send.stderr
@@ -11,12 +11,11 @@ note: required because it appears within the type `Counter`
   |
 9 | pub struct Counter {
   |            ^^^^^^^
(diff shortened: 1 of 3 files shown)


Hywan (approved):

bendk (approved):
Definitely a good idea.  I thought we had this bound, maybe I accidentally removed it when doing the `FfiConverterArc` refactory?  In any case, let's merge.

bendk:
Looks like we have several examples/fixtures that aren't following this rule.  Can you update them to pass?

jplatte:
I think it was just the UI tests that had their output changed. Tests are passing locally after running with `TRYBUILD=overwrite`.

Issue #1734, opened 2023-09-05
Foreign DispatchQueue type


bendk:
A common pattern we use in [application-services](https://github.com/mozilla/application-services/) is to run SQLite queries using a foreign-managed task queue (`Dispatchers.IO` on Kotlin, `DispatchQueue` on Swift, etc).  Could UniFFI support this as a first-class type?

This would require:
  - Moving it across the FFI (only supporting Foreign -> Rust is probably fine)
  - Adding way to execute Rust synchronous code using the foreign DispatchQueue in two modes:
     - async mode where you await the result
     - fire-and-forget mode where you just want the code to run and don't care about the result.

bendk:
I think `BackgroundQueue` is a nicer name than `DispatchQueue`, since the feature is really about running tasks in a background thread.  I feel like there could be a better name though, open to any suggestions.

mhammond:
"background thread" implies "low priority thread" to me, which I don't think is the actual intent. The actual use-cases here tend to be more about thread blocking than thread priority. I was going to suggest something like `TaskQueue`, but depending on the content the `Task` part of that might itself carry baggage. `WorkQueue`? But I don't think it matters that much though.

bendk:
You're right, it really is about blocking more than anything.  `tokio` separates their threads into "worker threads" and "blocking threads".  What about `BlockingQueue` or `BlockingTaskQueue`?

mhammond:
using the same terminology as tokio is certainly appealing!

PR #1742, opened 2023-09-08
Updated Future FFI

bendk:
The initial motivation for this was cancellation.  PR#1697 made it so if an async function was cancelled we would eventually release the resources. However, it would be better if we could immedately release the resources.  In order to implement that, I realized I needed to change the future FFI quite a bit.

The new FFI is simpler overall and supports cancel and drop operations. Cancel ensures that the foreign code will resume and break out of its async code.  Drop ensures that all resources from the wrapped future are relased.

The new code does not use ForeignExecutor anymore, so that code is in a state of limbo.  I hope to repurpose it for foreign dispatch queues (#1734).  If that doesn't work out, we can just delete it.

The FFI calls need some care since we pass a type-erased handle to the foreign code, while RustFuture is generic over an anonymous Future type:
  - The concrete RustFuture type implements the `RustFutureFFI<F>` where F is the FFI type.
  - We give the foreign side a raw `Box<Arc<dyn RustFutureFFI<F>>>>`.  This makes it easy to call the FFI from the scaffolding code.  The extra Box turns the wide pointer into a normal sized pointer, which simplifies the foreign code.
  - We brute-force monomorphize the scaffolding functions, generating a poll, cancel, complete, and free function for each possible FFI type.
  
More changes:
- Updated the futures fixture tests for this to hold on to the mutex longer in the initial call.  This makes it so they will fail unless the future is dropped while the mutex is still locked.  Before they would only succeed as long as the mutex was dropped once the timeout expired.
- Updated `RustCallStatus.code` field to be an enum.  Added `Cancelled` as one of the variants.  `Cancelled` is only used for async functions.
- Removed the FutureCallback and invoke_future_callback from `FfiConverter`.
- New syncronization handling code in RustFuture that's hopefully clearer, more correct, and more understandable than the old stuff.
- Updated `UNIFFI_CONTRACT_VERSION` since this is an ABI change
- Removed the `RustCallStatus` param from async scaffolding functions. These functions can't fail, so there's no need.
- Added is_async() to the Callable trait.


uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -8,68 +8,26 @@
...
+//!   - Suspend the function until the [rust_future_poll] continuation function is called
+//!   - If the continuation was function was called with [RustFuturePoll::Ready], then break
+//!     otherwise continue.
+//! 2b. If the async function is cancelled, then call [rust_future_cancel].  This causes the
+//!     continuation function to be called with [RustFuturePoll::Ready] and the [RustFuture] to
+//!     enter a cancelled state.
+//! 3. Call [rust_future_complete] to get the result of the future.
+//! 4. Call [rust_future_free] to free the future, ideally in a finally block.  This:
+//!    - Releases any resources held by the future
+//!    - Calls any continuation callbacks that have not been called yet
```
bendk, line 25: This PR could be called @Hywan was right.  This new API is basically the one he initially proposed with just some tweaks to simplify the FFI and bindings code.

fixtures/futures/tests/bindings/test_futures.py
```diff
@@ -150,21 +150,14 @@ async def test():
...
-        # should release the shared resource before the timeout expires.
-        asyncio.run(use_shared_resource(SharedResourceOptions(release_after_ms=0, timeout_ms=1000)))
+        async def test():
+            task = asyncio.create_task(use_shared_resource(
+                SharedResourceOptions(release_after_ms=5000, timeout_ms=100)))
+            # Wait some time to ensure the task has locked the shared resource
+            await asyncio.sleep(0.05)
+            task.cancel()
+            await use_shared_resource(SharedResourceOptions(release_after_ms=0, timeout_ms=1000))
+        asyncio.run(test())
```
bendk, line 160: Note: that this is changing the test to check what happens when you call `task.cancel()`, rather than closing an `EventLoop`.  This matches real-world use much better, but it is true that we won't clean up resources if the user manually closes the eventloop with Rust futures still running.

I personally don't think we need to address that or even document it.  Please tell me if you think we should be doing either of those things.

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -8,68 +8,26 @@
...
-//!     })
-//! }
-//! ```
-//!
-//! Rust will continue to poll the future until it's ready, after that.  The callback will
-//! eventually be invoked with these arguments:
-//!   - callback_data
-//!   - FfiConverter::ReturnType (the type that would be returned by a sync function)
-//!   - RustCallStatus (used to signal errors/panics when executing the future)
-//! - Rust will stop polling the future, even if it's waker is invoked again.
```
bendk, line 72: I really liked this example code as a reader, but it was really hard to keep up to date as I was making changes.  Hopefully the pseudocode version will provide close to the same benefit without being so hard to maintain.  Does that make sense to others?

uniffi_macros/src/export/scaffolding.rs
```diff
@@ -188,25 +188,10 @@ fn gen_ffi_function(
...
-                        #future_expr,
-                        uniffi_executor_handle,
-                        uniffi_callback,
-                        uniffi_callback_data
-                    );
-                    uniffi_rust_future.wake();
-                    Ok(())
-                });
+                #pre_fn_call;
+                ::uniffi::rust_future_new(async move { #future_expr.await }, crate::UniFfiTag)
```
bendk, line 194: Is new `async move {}` actually fixes an important bug.  If you have a [async method that inputs `&self`](https://github.com/bendk/uniffi-rs/blob/3b7180f374fcd27f2abc66c8f3c1c3787ee57e04/fixtures/metadata/src/tests.rs#L384-L386), then the generate future has the same lifetime as `self`.  Things would break if you destroyed the object then continued to poll the future, which was possible with the old system.

I found this out because Rust complained when compiling the `fixtures/metadata` test.  In general, it seems like the code in `rustfutures.rs` relies less on unsafe code and our intuition and therefore Rust is able to check things better.  It was really annoying adding all those `Send + 'static` bounds, but that's already paid off.

uniffi_bindgen/src/interface/function.rs
```diff
@@ -264,6 +271,10 @@ impl Callable for Function {
     fn throws_type(&self) -> Option<Type> {
         self.throws_type().cloned()
     }
+
+    fn is_async(&self) -> bool {
+        self.is_async()
```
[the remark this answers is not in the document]
bendk: Good point, I changed it to use the field directly.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -413,6 +413,120 @@ impl ComponentInterface {
...
+                ),
+                _ => unimplemented!("Async functions for {t:?}"),
+            },
+            None => format!("ffi_{}_rust_future_complete_void", self.ffi_namespace()),
+        };
+        FfiFunction {
+            name,
+            is_async: false,
+            arguments: vec![FfiArgument {
+                name: "handle".to_string(),
```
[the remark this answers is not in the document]
bendk: Updated the code.  I agree that's the best method name for this conversion.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -413,6 +413,120 @@ impl ComponentInterface {
...
 
+    /// Builtin FFI function to set the Rust Future continuation callback
+    pub fn ffi_rust_future_continuation_callback_set(&self) -> FfiFunction {
+        FfiFunction {
+            name: format!(
+                "ffi_{}_rust_future_continuation_callback_set",
+                self.ffi_namespace()
+            ),
+            arguments: vec![FfiArgument {
+                name: "callback".into(),
```
jplatte, line 424:
```suggestion
                name: "callback".to_owned(),
```

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -116,476 +76,627 @@
...
+/// - The `T` param must correctly correspond to the [rust_future_new] call.  It must
+///   be `<Output as FfiConverter<UT>>::ReturnType`
+pub unsafe fn rust_future_complete<T: FfiDefault>(
+    handle: RustFutureHandle,
+    out_status: &mut RustCallStatus,
+) -> T {
+    let future = &*(handle.0 as *mut Arc<dyn RustFutureFFI>);
+    let mut return_value = T::ffi_default();
+    let out_return = mem::transmute::<&mut T, &mut ()>(&mut return_value);
+    future.ffi_complete(out_return, out_status);
```
jplatte, line 195: This looks very wrong. Why are safe reference types being used here?

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -116,476 +76,627 @@
...
+                log::error!(
+                    "store: observed Self::Set state, is poll() being called from multiple threads at once?"
+                );
+                call_continuation(old_data, RustFuturePoll::Ready);
+            }
+        }
+    }
+
+    fn send(&mut self) {
+        if matches!(self, Self::Cancelled) {
```
jplatte, line 248: For consistency, should also use `if let`, or `fn store` above should also use `matches!`.

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -116,476 +76,627 @@
...
-    /// This function gets called when a `RawWaker` gets dropped.
-    /// This function gets called when a `RawWaker` gets dropped.
-    unsafe fn raw_drop(self_ptr: *const ()) {
-        drop(Self::from_raw(self_ptr))
+/// RustFuture FFI trait.  This allows `Arc<RustFuture<_, _, _>>` to be cast to
+/// `Arc<dyn RustFutureFFI>`, which is needed to implement the public FFI API.  In particular, this
+/// allows you to use RustFuture functionality without knowing the concrete Future type, which is
+/// unnamable.
+#[doc(hidden)]
+trait RustFutureFFI {
```
jplatte, line 467:
```suggestion
trait RustFutureFfi {
```

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -116,476 +76,627 @@
...
-    }
-
-    /// This function gets called when a `RawWaker` gets dropped.
-    /// This function gets called when a `RawWaker` gets dropped.
-    unsafe fn raw_drop(self_ptr: *const ()) {
-        drop(Self::from_raw(self_ptr))
+/// RustFuture FFI trait.  This allows `Arc<RustFuture<_, _, _>>` to be cast to
+/// `Arc<dyn RustFutureFFI>`, which is needed to implement the public FFI API.  In particular, this
+/// allows you to use RustFuture functionality without knowing the concrete Future type, which is
+/// unnamable.
```
jplatte, line 465: It's not really clear to me why we have this separate trait instead of storing a `dyn Future` in `RustFuture`, but that's not something I want to dig into right now. So just leaving this comment here as a note that `dyn Future` is also a thing.

uniffi_core/src/ffi/rustfuture.rs
```diff
@@ -116,476 +76,627 @@
...
+    fn ffi_cancel(&self) {
+        self.cancel()
+    }
+
+    unsafe fn ffi_complete(&self, return_value: &mut (), call_status: &mut RustCallStatus) {
+        // Unsafely transmute return_value.  This works as long as the foreign code calls the
+        // correct `rust_future_complete_*` function.
+        let return_value = std::mem::transmute::<&mut (), &mut T::ReturnType>(return_value);
+        self.complete(return_value, call_status)
+    }
```
jplatte, line 494: I really think `&mut ()` should be `*mut ()` or `NonNull<()>`. `&mut` has aliasing guarantees that we are almost definitely violating. Also, transmute should be the absolute last thing to reach for when other casting methods are not applicable.

jplatte (changes requested):
I'm pretty certain that some of the code in there is UB. We should probably have some unit tests that Miri can run.

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@
 - [Custom Types](https://mozilla.github.io/uniffi-rs/proc_macro/index.html#the-unifficustomtype-derive) are now supported for proc-macros, including a very
   low-friction way of exposing types implementing the new-type idiom.
 - Proc-macros: Added support for ByRef arguments
+- Swift async functions/methods now always throw, since they can throw CancellationError when the parent
+  task is cancelled.
```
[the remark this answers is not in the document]
bendk: I have no clue :)  The only alternative I could see was a runtime error, which can't be a general solution.  Maybe users could configure which behavior they wanted?

uniffi_bindgen/src/bindings/swift/templates/Async.swift
```diff
@@ -0,0 +1,67 @@
...
+            pollResult = await withUnsafeContinuation {
+                {{ ci.ffi_rust_future_poll().name() }}(
+                    rustFuture,
+                    ContinuationHolder($0).toOpaque()
+                )
+            }
+        } while pollResult != UNIFFI_RUST_FUTURE_POLL_READY
+    } onCancel: {
+        {{ ci.ffi_rust_future_cancel().name() }}(rustFuture)
+    }
```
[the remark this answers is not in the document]
jplatte: I see what you mean. I didn't review the Rust code in detail, but if there is a mutex that should indeed be fine.

uniffi_core/src/ffi/rustfuture.rs
bendk: I don't understand this one.  Both `rust_future_complete` and `ffi_complete` are unsafe functions precisely because they transmute the references.  The safety requirements are listed above.  If those are followed, I think the code should work because you'll end up with something like:
```
let mut return_value: T::ReturnType = T::ffi_default();
let out_return = mem::transmute::<&mut T, &mut ()>(&mut return_value);
let out_return = std::mem::transmute::<&mut (), &mut T::ReturnType>(out_return);
```

That should be okay right?

uniffi_core/src/ffi/rustfuture.rs
bendk: Storing `dyn Future` would be possible, but then you'd still have the other generics (`RustFuture<T, UT>`).

I think that would mean that each exported async function would need its own poll/drop/cancel/complete scaffolding functions, rather than the current situation where we have a single poll/drop/cancel, and one complete function per FFI type.  I went with this one, but it's a close call and I'm definitely open to the other way.

bendk: I think that if each async function had its own complete function, we could drop that transmuting code, which would be nice.

uniffi_core/src/ffi/rustfuture.rs
bendk: That's very interesting, although I still don't follow it all.  Wouldn't aliasing need to involve 2 pointers/references?  Is the issue that some of these transmutes are copying the `&ref` so you end up with 2 of them?

RE provenance: does it work differently between pointers and references?  The RawWaker API casts `*const T` to `*const ()`, so I figured I could do the same with references.  My rough mental model is that provenance is set when the pointer is allocated and sticks around through a cast/transmute.

In any case, all of this is making me want to rework things we keep the generic `T` around and avoid all of these issues.  I think that means we need to define a poll/cancel/complete/free scaffolding function for each exported async function.  That sounds a bit sad, but worth it keep things simple.

jplatte: > That's very interesting, although I still don't follow it all. Wouldn't aliasing need to involve 2 pointers/references? Is the issue that some of these transmutes are copying the `&ref` so you end up with 2 of them?

Oh, right transmute moves the `&mut` so maybe it is fine in that regard.

> RE provenance: does it work differently between pointers and references? The RawWaker API casts `*const T` to `*const ()`, so I figured I could do the same with references. My rough mental model is that provenance is set when the pointer is allocated and sticks around through a cast/transmute.

Yes, provenance is different between pointers and references, at least under stacked borrows. I believe it's different with the tree borrows model. Under stacked borrows, AFAIK a (non-DST) reference only ever has provenance over as many bytes as there are in the type it references, so `&mut ()` has provenance over nothing. Pointers retain provenance when cast.

> In any case, all of this is making me want to rework things we keep the generic `T` around and avoid all of these issues. I think that means we need to define a poll/cancel/complete/free scaffolding function for each exported async function. That sounds a bit sad, but worth it keep things simple.

That's what Ivan's initial implementation did, right?

bendk:
My current thinking is that the unsafe transmute code needs to go, but I also really want to avoid having to define 4 future FFI functions for each exported async function.  My quick estimate is that's 1kb of binary space per function.

I just pushed a commit that tries to balance things by making `RustFutureFfi` generic on `ReturnType`.  That makes it so we only define the 4 functions per FFI return type, which is currently 13 types.

uniffi_core/src/ffi/rustfuture.rs
Hywan: Thanks for recognizing it 👼.

uniffi_core/src/ffi/rustfuture.rs
Hywan: Indeed, that was part of my initial implem', and also part of my last PR.

CHANGELOG.md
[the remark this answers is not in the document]
Hywan: Let's ask good Swift developers, like @stefanceriu or @pixlwave?

fixtures/futures/tests/bindings/test_futures.swift
```diff
@@ -236,7 +236,8 @@ Task {
 counter.enter()
 Task {
 	let task = Task {
-	    try! await useSharedResource(options: SharedResourceOptions(releaseAfterMs: 100, timeoutMs: 1000))
+            // Use try? to ignore cancellation errors
```
Hywan, line 239:
```suggestion
        // Use try? to ignore cancellation errors
```

Akazm:
CancellationErrors are not guaranteed to be thrown by `async func`s, but only by Swift's `Task`s instead.

[See here, for example](https://forums.swift.org/t/cancellationerror/60009)

bendk:
> CancellationErrors are not guaranteed to be thrown by `async func`s, but only by Swift's `Task`s instead.

But they can be right?  If the consumer will sometimes cancel the parent task, then I believe the async function should be marked `throws`.

  - Does that sound right to you?
  - When we are generating an async function we don't know if the consumer intends it cancel the parent task or not.  Should we always mark it `throws`, never mark it `throws`, make it configurable somehow?
  - Is there a way to avoid the issue?  If the parent task is cancelled, then I guess it doesn't really matter what we return.  But I'm not sure what code to generate for that case.

CHANGELOG.md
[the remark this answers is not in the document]
stefanceriu: I'm not sure I get the full context here but that code looks just fine 🤷‍♂️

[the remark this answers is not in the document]
jplatte: It's about [this](https://github.com/mozilla/uniffi-rs/pull/1742/files#diff-c96520ae7c5e213facf72cbbad8297cdc353c894f23f4aae30dc298463471c9a) diff where every `await` is now prefixed with `try!` or `try?`.

bendk:
Here's the concrete situation I'm thinking of:
- Swift code starts a task
- The task runs the async function `foo()`
- `foo()` calls the generated async function `bar()`
- `bar()` doesn't naturally throw, but it's possible that the task is cancelled.  So what should it do?
  - Mark itself as `throw` and throw `CancellationError` if the parent task is cancelled.
  - Don't mark itself as `throw`, but still throw an error.  If the parent task is cancelled, I believe we're going to end up with a `RuntimeError`.
  - Some other option that I can't figure out.

Akazm:
> But they can be right? If the consumer will sometimes cancel the parent task, then maybe the async function should be marked `throws`.

> * Does that sound right to you?

In idiomatic terms: no.

> * When we are generating an async function we don't know if the consumer intends it cancel the parent task or not.  Should we always mark it `throws`, never mark it `throws`, make it configurable somehow?

Configurable sounds like a good trade-off, but without having looked at the code (!!!), I don't really get what the issue exactly is here. You'd typically cancel a task top-down, not the other way around. As stated in the example, Swift only guarantees providing information about the cancellation status via a Task object. I assume this is exactly because of the issue we're dealing with here. Kotlin Coroutines handle this differently, where cancellation is done via Exceptions.

Without further ado by the developer, Swift tasks just get cancelled. You could consider this opt-in.

> * Is there a way to avoid the issue? 

As stated above - If you'd really need to be aware of cancellation when calling a specific non-throwing async function, you could always write an extension for it.

CHANGELOG.md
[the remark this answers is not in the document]
stefanceriu: When you're dealing with throwing functions you need to mark them with some variation of `try`. When you're in the context of a Task:
* simple `try` will rethrow the error up to the task level, where it will transform into a `Task<(), Error>`
* `try?` will make the result an optional which you will need to handle locally. The parent task will become `Task<()?, Never>`
* and `try!` will make the runtime crash with a `Fatal error: 'try!' expression unexpectedly raised an error:`. The parent task will become `Task<(), Never>` (notice the optional)

I'm not sure how that testing jig is setup and how fatal errors are handled but it seems that tasks that are expected to throw are correctly marked with `try?`. For the rest it might be simpler to just use `try`, throw errors up the task level and expect the `counter` to be wrong and fail the test.

Hope this answers your question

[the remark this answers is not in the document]
bendk: I think it's about the ergonomics for consumers.  We probably don't want to require them to add `try!` to all of their async function calls, because they could be cancelled in principle even if the application never actually cancels them.

I think you suggested configuration as a solution to this issue and that seems like it strikes a good balance.  By default, generated async functions will not check for cancellation and will only throw if the underlying Rust function returns a `Result`.  Swift users can opt-in to cancellation checks in the `uniffi.toml` file, which means they will always need to use one of the `try` variations when calling the function.

bendk:
Pushed two commits. One implements configurable cancellation on Swift like we discussed.  The second one fixes an issue I realized after rebasing over the #1747 changes: futures would fail to have a `Send` bound if they lifted objects, since that involves a raw pointer.

I made two new commits for easy review, but if this is approved I'll squash the final product into 1 commit.  I'll also add @jplatte and @Hywan  as co-authors, thanks for all the help with this work.

Is this ready to merge?

docs/manual/src/futures.md
```diff
@@ -41,3 +41,21 @@ In Rust `Future` terminology this means the foreign bindings supply the "executo
...
+## Swift Cancellation
+
+Swift async functions normally do not check for cancellation of their parent task.  You can opt in
+to this using the `[bindings.swift.cancellable]` array in the `uniffi.toml` file:
+
+```toml
+[bindings.swift]
+cancellable = [
+   # Make the `foo` function cancellable
+   "foo",
```
bendk, line 54: I'm open to suggestions for a better name for this config value.

jplatte:
So how does it work with the configurable cancellation now if the task is cancelled? I'm guessing that fns that aren't marked as throwing can still throw in Swift, similar to the old checked vs. unchecked exceptions in Java? And the cancellation makes the Swift scaffolding code throw, which is then propagated up to the root of the task? So anything that depends on that task will itself see the error but in a more obvious way (joining a task being checked-fallible)?

Akazm:
> So how does it work with the configurable cancellation now if the task is cancelled? I'm guessing that fns that aren't marked as throwing can still throw in Swift, similar to the old checked vs. unchecked exceptions in Java? 

Any Error thrown from a function that is not explicitly declared as throwing will result in a crash/panic. Cancelling a task, however, does not throw an error. Hence I'm confused about this in general, but I didn't take a look at the code.

> And the cancellation makes the Swift scaffolding code throw, 

It should not, as stated above, at least in idiomatic terms.

bendk:
>>   And the cancellation makes the Swift scaffolding code throw,

> It should not, as stated above, at least in idiomatic terms.

It's hard making this work idiomatically for all languages.  Here's the ways I would consider approaching this:
  - Don't make any Swift functions cancellable.  Add a way to set a cancellation flag in the Rust interface that the Rust code checks for.  This feels slightly odd for Rust async since you usually just drop the future, but it would work alright.  We could create a way to have Rust check the Swift `Task.isCancelled` flag, but making an FFI call for that doesn't feel right to me and wouldn't work for Kotlin/Python.
  - Wrap the Rust async function with a Swift async function.  It can check for `CancellationError` and do the right thing.

One last possibility would be to treat optional / void returns specially.  If those get cancelled, we could make the async function return `nil` or `()` instead of throwing.  I'm guessing that's how most of the Swift wrappers are going to look.  Maybe that could be a third configuration option down the line.

> So how does it work with the configurable cancellation now if the task is cancelled? I'm guessing that fns that aren't marked as throwing can still throw in Swift, similar to the old checked vs. unchecked exceptions in Java?

True, but but the async functions that are configured for cancellation will be marked `throws` so the calling function will need to explicitly `try!` for this to happen.

> And the cancellation makes the Swift scaffolding code throw, which is then propagated up to the root of the task? So anything that depends on that task will itself see the error but in a more obvious way (joining a task being checked-fallible)?

I believe that's what will happen, although I think most applications will end up some sort of wrapper so that cancellation gets handled before it propagates too much.

bendk:
> I just realized there are unit tests in uniffi-core. I decided to run them through miri. After disabling the foreign executor (since that API is no longer used anymore), I got this on your branch

I would expect tons of these errors since we're doing things like inputing raw pointers from the foreign side.  Do the other callback cells generate the same errors?  Maybe the difference is just test coverage.  In any case, I do hope to lessen the number of raw pointers soon.

jplatte:
> I would expect tons of these errors since we're doing things like inputing raw pointers from the foreign side.

Err, what? You expect the code to have undefined behavior??

bendk:
> > I would expect tons of these errors since we're doing things like inputing raw pointers from the foreign side.
> 
> Err, what? You expect the code to have undefined behavior??

I guess I don't understand miri.   If a pointer comes is passed from across the C layer, then how can it know that it's non-dangling, well-aligned, or has any provenance?

bendk:
Maybe we should think through the Swift cancellation code in a separate PR and try to land the rest of this which has some nice improvements to the FFI that I'd like to take.

jplatte:
Sure, I'm fine with this being merged, then iterated on in follow-up PRs. I'm interested in trying to make Miri happy.

bendk:
https://github.com/mozilla/uniffi-rs/pull/1767 contains these changes, without commiting to a system for Swift cancellation, and with some fixes that make the test pass when run in miri.

bendk:
And cancellation is in https://github.com/mozilla/uniffi-rs/pull/1768.  I'm going to close this one.  Let's continue to discuss this on those PRs.

PR #1746, opened 2023-09-11
Added ByRef support for proc-macros

bendk:
The nice thing about this change is that users don't need to mark their arguments in some special way.  If you take a `&Object` things will work.

The not-so-nice parts is that we need to go back to token parsing to implement this, so type aliases can make it fail.  Also it means that we can't support things like `&Enum` or `&Record`.  I thought about trying some tricks like `TYPE_ID_META` or introducing a new trait, but I couldn't think of any way to make it work.  Given that users used to have to mark the argument with `[ByRef]` and this hasn't historically worked with enums/records, I think it should be okay.


docs/manual/src/proc_macro/index.md
```diff
@@ -186,6 +186,19 @@ impl Foo {
...
+Exported functions can input object arguments as either an `Arc<>` or reference.
+```rust
+// Input foo as an Arc and bar as a reference
+fn call_both(foo: Arc<Foo>, bar: &Foo) {
+  foo.method_a();
+  bar.method_rba();
+```
+
+There are a couple limitations when using references for arguments:
+  - They can only be used with interface types: objects, trait interfaces, or callback interfaces.
```
bendk, line 198: Note: I've only actually tested this with objects, but I'm pretty sure the other two possibilities should work as well.

jplatte (approved):
Other than the documentation nitpick, LGTM

PR #1747, opened 2023-09-12
Use proc-macros for all exports

bendk:
Reworked UDL-based scaffolding generation so that it forwards the work to the proc-macro code.  This means that almost all scaffolding generation now flows through the proc-macros and we no longer need to keep 2 implementations of it.  The only major thing left is built-in traits.

The scaffolding templates now generate stub interfaces and wrap them with a `[uniffi::export]` macro. Added the `udl_mode` argument to the `export` macro to support this.

- Added support for [ByArc] in callback interfaces
- Moved all keywords into a single module.
- Renamed the existing macros that did similar work to `derive_[thing]_for_udl`.
- The derive macros now also support the `#[uniffi(udl_mode)]` attribute. I doubt any users will use this, but the code seemed cleaner this way so why not?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,10 +28,11 @@
     be used to avoid this and use the specified value.
 
 - Crates can now use proc-macros without UDL files to export their interface.  See the "Procedural Macros: Attributes and Derives" manual section for details.
-
 - [Custom Types](https://mozilla.github.io/uniffi-rs/proc_macro/index.html#the-unifficustomtype-derive) are now supported for proc-macros, including a very
   low-friction way of exposing types implementing the new-type idiom.
 - Proc-macros: Added support for ByRef arguments
+- Proc-macros: Implemented custom type conversion error handling (https://mozilla.github.io/uniffi-rs/udl/custom_types.html#error-handling-during-conversion)
+- Error types must now implement `Error + Send + Sync + 'static`.
 
 ### What's Fixed
 
--- a/docs/manual/src/proc_macro/index.md
+++ b/docs/manual/src/proc_macro/index.md
@@ -91,13 +91,45 @@ trait MyTrait {
 
 ```
 
-Most UniFFI [builtin types](../udl/builtin_types.md) can be used as parameter and return types.
-When a type is not supported, you will get a clear compiler error about it.
 
-User-defined types are also supported in a limited manner: records (structs with named fields,
-`dictionary` in UDL) and enums can be used when the corresponding derive macro is used at
-their definition. Opaque objects (`interface` in UDL) can always be used regardless of whether they
-are defined in UDL and / or via derive macro; they just need to be put inside an `Arc` as always.
+All owned [builtin types](../udl/builtin_types.md) and user-defined types can be used as arguments
+and return types.
+
+Arguments and receivers can also be references to these types, for example:
+
+```rust
+// Input data types as references
+#[uniffi::export]
+fn process_data(a: &MyRecord, b: &MyEnum, c: Option<&MyRecord>) {
+    ...
+}
+
+#[uniffi::export]
+impl Foo {
+  // Methods can take a `&self`, which will be borrowed from `Arc<Self>`
+  fn some_method(&self) {
+    ...
+  }
+}
+
+// Input foo as an Arc and bar as a reference
+fn call_both(foo: Arc<Foo>, bar: &Foo) {
+  foo.some_method();
+  bar.some_method();
+}
+```
+
+The one restriction is that the reference must be visible in the function signature.  This wouldn't
+work:
+
+```rust
+type MyFooRef = &'static Foo;
+
+// ERROR: UniFFI won't recognize that the `foo` argument is a reference.
+#[uniffi::export]
+fn do_something(foo: MyFooRef) {
+}
+```
 
 ## The `uniffi::Record` derive
 
@@ -185,19 +217,6 @@ impl Foo {
     }
 }
 ```
-
-Exported functions can input object arguments as either an `Arc<>` or reference.
-```rust
-// Input foo as an Arc and bar as a reference
-fn call_both(foo: Arc<Foo>, bar: &Foo) {
-  foo.method_a();
-  bar.method_rba();
-```
-
-There are a couple limitations when using references for arguments:
-  - They can only be used with objects and trait interfaces
-  - The reference must be visible in the function signature.
-    If you have a type alias `type MyFooRef<'a> = &'a Foo`, then `fn do_something(foo: MyFooRef<'_>)` would not work.
 
 ## The `uniffi::custom_type` and `uniffi::custom_newtype` macros
 
--- a/uniffi_core/src/ffi_converter_traits.rs
+++ b/uniffi_core/src/ffi_converter_traits.rs
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use std::sync::Arc;
+use std::{borrow::Borrow, sync::Arc};
 
 use crate::{
     try_lift_from_rust_buffer, FfiDefault, MetadataBuffer, Result, RustBuffer, RustCallStatus,
@@ -90,6 +90,16 @@ pub unsafe trait FfiConverter<UT>: Sized {
     /// thrown exceptions on the foreign code, serialize the error into a RustBuffer and return
     /// `Err(buf)`
     fn lower_return(obj: Self) -> Result<Self::ReturnType, RustBuffer>;
+
+    /// If possible, get a serialized error for failed argument lifts
+    ///
+    /// By default, we just panic and let `rust_call` handle things.  However, for `Result<_, E>`
+    /// returns, if the anyhow error can be downcast to `E`, then serialize that and return it.
+    /// This results in the foreign code throwing a "normal" exception, rather than an unexpected
+    /// exception.
+    fn handle_failed_lift(arg_name: &str, e: anyhow::Error) -> RustBuffer {
+        panic!("Failed to convert arg '{arg_name}': {e}")
+    }
 
     /// Lift a rust value of the target type, from an FFI value of type Self::FfiType.
     ///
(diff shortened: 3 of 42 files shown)


bendk:
One nice thing about this PR is that exposes all the bits of inconsistency left between the proc-macro and UDL-based generation.   I made https://github.com/mozilla/uniffi-rs/labels/proc-macro-udl-inconsistency to capture what I think are the remaining issues here.

ObjectTemplate.rs
```diff
@@ -0,0 +1,142 @@
+// For each Object definition, we assume the caller has provided an appropriately-shaped `struct T`
```
badboy, line 1: Mistakenly commited this file at the top-level?

bendk:
I realized yesterday that the first commit handled methods and constructors, but not top-level functions.  Once I started using proc-macros for those I realized that there were two more issues that needed handling:
  - Custom type conversion errors
  - ByRef args in any position and with any type

This is making me feel pretty good about the approach.  Making the code to flow through proc-macros is forcing them to gain feature parity with UDL.

uniffi_core/src/ffi_converter_traits.rs
```diff
@@ -253,3 +263,16 @@ where
...
+/// This is needed because if we see a `&T` parameter, it's not clear which FfiConverter to use to
+/// lift it.  Normally it's just `T`, but for interfaces it's `Arc<T>` and for callback interfaces
+/// it's `Box<T>`.
+///
+/// This trait provides the proc-macros with the functionality they need to handle this
+pub trait LiftRef<UT> {
+    type LiftType: FfiConverter<UT>;
+
+    fn as_ref(v: &Self::LiftType) -> &Self;
+}
```
bendk, line 278: I'm hoping `LiftRef` is the first step towards deconstructing `FfiConverter` into smaller traits, maybe those smaller traits could be named `Lift`, `Lower`, and `Return`.

I tried to avoid having to define the `as_ref` method.  I thought I could use the standard `Deref` or `AsRef`, but I couldn't get them to work for both `&T` and `&Arc<T>`, since the latter needs `&**` to convert it to the proper reference type.  If anyone else can get this to work, we can drop this method and the trait gets very simple.

jplatte: I think the `Borrow` trait could work here.

CHANGELOG.md
```diff
@@ -17,10 +17,11 @@
 ### What's new
 
 - Crates can now use proc-macros without UDL files to export their interface.  See the "Procedural Macros: Attributes and Derives" manual section for details.
-
 - [Custom Types](https://mozilla.github.io/uniffi-rs/proc_macro/index.html#the-unifficustomtype-derive) are now supported for proc-macros, including a very
   low-friction way of exposing types implementing the new-type idiom.
 - Proc-macros: Added support for ByRef arguments
+- Proc-macros: Implemented custom type conversion error handling (https://mozilla.github.io/uniffi-rs/udl/custom_types.html#error-handling-during-conversion)
+- Proc-macros: Error types must now implement `Error + Send + Sync + 'static`, like they do for UDL.
```
bendk, line 24: I don't think this should be a large burden for library consumers, but please speak up if it will affect you.

uniffi_core/src/ffi_converter_traits.rs
bendk: That's beautiful, thanks.

CHANGELOG.md
badboy: This applies to UDL code though too now.
At least I just pulled in your changes to compile Glean against and noticed that we _don't_ yet implement `std::error::Error`. Easy to solve and we already satisfy the other bounds.

docs/manual/src/proc_macro/index.md
```diff
@@ -91,13 +91,43 @@ trait MyTrait {
 
 ```
 
-Most UniFFI [builtin types](../udl/builtin_types.md) can be used as parameter and return types.
-When a type is not supported, you will get a clear compiler error about it.
+All UniFFI types can be used as parameter and return types.
```
badboy, line 94: What are `UniFFI types`? ~~Also we don't support returning references, but we can take them in.~~ (this is made more clear in the next sentence)
So probably this sentence should be more precise.

badboy (approved):
I gave this a cursory look and also compiled Glean against it. Minimal changes necessary for us and we still pass our test suite. Looking at the generated code I was initially confused how it even works by redefining everything, but because it attaches all those macros it of course works.
Overall I like it. I didn't dive into all code changes (it's a lot), but I'm still confident you have everything covered.

(and hey, the generated code could act as the base work to migrate to the proc-macro approach \o/)

docs/manual/src/proc_macro/index.md
bendk: Good point, I just pushed an update.

CHANGELOG.md
bendk: We could keep the bound as `Display + Debug + Send + Sync + 'static`, but I thought it was simpler to spell out the requirement as `Error + Send + Sync + 'static` and maybe we'll need the `Error` bound sometime in the future.

uniffi_macros/src/export/scaffolding.rs
```diff
@@ -114,27 +135,41 @@ impl ScaffoldingBits {
...
         let params: Vec<_> = sig.args.iter().map(NamedArg::scaffolding_param).collect();
         let param_lifts = sig.lift_exprs();
+        let mut rust_fn_call = quote! { #self_ident::#ident(#(#param_lifts,)*) };
+        if udl_mode {
+            if sig.looks_like_result {
+                rust_fn_call = quote! { #rust_fn_call.map(::std::sync::Arc::new).map_err(::std::convert::Into::into) };
+            } else {
+                rust_fn_call = quote! { ::std::sync::Arc::new(#rust_fn_call) };
+            }
+        }
```
jplatte, line 167: No need to make this `mut`, can just move the initial value for `rust_fn_call` into an `else`.

uniffi_macros/src/export/scaffolding.rs
```diff
@@ -114,27 +135,41 @@ impl ScaffoldingBits {
             .chain(sig.scaffolding_params())
             .collect();
         let param_lifts = sig.lift_exprs();
+        let mut rust_fn_call = quote! { uniffi_self.#ident(#(#param_lifts,)*) };
+        if udl_mode && sig.looks_like_result {
+            rust_fn_call = quote! { #rust_fn_call.map_err(::std::convert::Into::into) }
+        }
```
jplatte, line 141: Same here.

uniffi_macros/src/export/scaffolding.rs
```diff
@@ -87,19 +99,28 @@ struct ScaffoldingBits {
...
 impl ScaffoldingBits {
-    fn new_for_function(sig: &FnSignature) -> Self {
+    fn new_for_function(sig: &FnSignature, udl_mode: bool) -> Self {
         let ident = &sig.ident;
         let params: Vec<_> = sig.args.iter().map(NamedArg::scaffolding_param).collect();
         let param_lifts = sig.lift_exprs();
+        let mut rust_fn_call = quote! { #ident(#(#param_lifts,)*) };
+        if udl_mode && sig.looks_like_result {
+            rust_fn_call = quote! { #rust_fn_call.map_err(::std::convert::Into::into) };
+        }
```
jplatte, line 109: And here.

uniffi_macros/src/export/callback_interface.rs
```diff
@@ -61,13 +61,19 @@ pub(super) fn trait_impl(
...
-    let impl_spec = tagged_impl_header("FfiConverter", &quote! { Box<dyn #trait_ident> }, tag);
-    let tag = match tag {
-        Some(t) => quote! { #t },
-        None => quote! { T },
+    let impl_spec = tagged_impl_header(
+        "FfiConverter",
+        &quote! { ::std::boxed::Box<dyn #trait_ident> },
+        udl_mode,
+    );
+    let lift_ref_impl_spec = tagged_impl_header("LiftRef", &quote! { dyn #trait_ident }, udl_mode);
```
jplatte, line 72:
```suggestion
    let dyn_trait = quote! { dyn #trait_ident };
    let box_dyn_trait = quote! { ::std::boxed::Box<#dyn_trait };
    let impl_spec = tagged_impl_header("FfiConverter", &box_dyn_trait, udl_mode);
    let lift_ref_impl_spec = tagged_impl_header("LiftRef", &dyn_trait, udl_mode);
```

uniffi_macros/src/export/callback_interface.rs
```diff
@@ -115,6 +121,10 @@ pub fn ffi_converter_callback_interface_impl(
             .concat_str(#mod_path)
             .concat_str(#name);
         }
+
+        #lift_ref_impl_spec {
+            type LiftType = ::std::boxed::Box<dyn #trait_ident>;
```
jplatte, line 126:
```suggestion
            type LiftType = #box_dyn_trait;
```

jplatte (approved):
Haven't reviewed thoroughly, but it's clear that this goes in the right direction! :)

uniffi_macros/src/fnsig.rs
```diff
@@ -374,6 +406,20 @@ impl NamedArg {
     }
 }
 
+fn looks_like_result(return_type: &ReturnType) -> bool {
+    if let ReturnType::Type(_, ty) = return_type {
+        if let Type::Path(p) = &**ty {
+            if let Some(seg) = p.path.segments.last() {
+                if seg.ident == "Result" {
```
skeet70, line 413: You mentioned in a different comment that this would fail for aliased results, but am I reading wrong that this would still work for the most comment aliased result of `Result<T>`?

skeet70 (approved):
I know my approval doesn't matter much, but this looks great, thanks.

mhammond (approved):
Nice! Thanks for killing "udl mode" :)

badboy:
> I know my approval doesn't matter much, but this looks great, thanks.

still appreciated.

uniffi_macros/src/fnsig.rs
bendk: I think it would, but it doesn't matter either way.  This is only used to implement the UDL-only behavior of automatically mapping errors using Into (#1749).  So we only need to make sure that the UDL-based generator outputs the expected tokens.

Depending on the decision in 1749, we could consider using this for proc-macros as well, but I think there's probably a way that works without parsing tokens.

PR #1751, opened 2023-09-12
Add note about coroutines dependency

skeet70:
Only a doc change, no tests needed. I noticed this dependency while setting up a nix flake to run my own tests generated using your macros.


bendk (approved):
Looks great, thanks for the contribution.

PR #1752, opened 2023-09-14
Catch special case of `Error` enum variant

skeet70:
This seems reasonable to me because if you're handling an `Enum` you're treating these special names as identifiers. At the same time it feels very hacky, and I couldn't tell where further up the chain was causing `Error` to be treated specially here. I tried testing with other `NonAnyTypes` like `Promise` and they end up in the same code as `Identifier` correctly.

### DIFF
--- a/uniffi_udl/src/converters/mod.rs
+++ b/uniffi_udl/src/converters/mod.rs
@@ -54,10 +54,19 @@ impl APIConverter<VariantMetadata> for weedle::interface::OperationInterfaceMemb
         }
         let name: String = {
             use weedle::types::{
-                NonAnyType::Identifier, ReturnType, SingleType::NonAny, Type::Single,
+                NonAnyType::{self, Identifier},
+                ReturnType,
+                SingleType::NonAny,
+                Type::Single,
             };
             match &self.return_type {
                 ReturnType::Type(Single(NonAny(Identifier(id)))) => id.type_.0.to_owned(),
+                // Using recognized/parsed types as enum variant names can lead to the bail error because they match
+                // before `Identifier`. `Error` is one that's likely to be common, so we're circumventing what is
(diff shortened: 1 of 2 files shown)


uniffi_udl/src/converters/mod.rs
```diff
@@ -54,10 +54,14 @@ impl APIConverter<VariantMetadata> for weedle::interface::OperationInterfaceMemb
...
             use weedle::types::{
-                NonAnyType::Identifier, ReturnType, SingleType::NonAny, Type::Single,
+                NonAnyType::{self, Identifier},
+                ReturnType,
+                SingleType::NonAny,
+                Type::Single,
             };
             match &self.return_type {
                 ReturnType::Type(Single(NonAny(Identifier(id)))) => id.type_.0.to_owned(),
+                ReturnType::Type(Single(NonAny(NonAnyType::Error(_)))) => "Error".to_string(),
```
jeddai, line 64: Yep this is exactly what I did when I was trying to build/test a fix for this locally. Great to see the test with it 💯

mhammond:
> I tried testing with other `NonAnyTypes` like `Promise` and they end up in the same code as `Identifier` correctly.

huh - so `Error` is defined as `Error(MayBeNull<term!(Error)>),`, while `Promise` is `Promise(PromiseType<'a>),` - and indeed that works. Something closer to Error is `Object(MayBeNull<term!(object)>),` - which also works. Closer yet is `DataView(MayBeNull<term!(DataView)>),` - which does not! A quick look doesn't help me understand why, but the case for using the failing items as names doesn't seem as strong as error.

@skeet70, how do you feel about updating the uniffi_udl change to include a comment which reflects the above, then I think we can just merge it with support for `Error` and open another issue should someone else want to do something more comprehensive?

PR #1754, opened 2023-09-15
Update rust to 1.72

mhammond:
Had a 1.71 update unpublished in my local branches, so dusted it off. Not at all bothered if we want to wait before landing this.


jhugman (approved):
This looks like a rubber stamp: LGTM. r+

PR #1761, opened 2023-09-22
Provide Kotlin companion objects for generated types

Akazm:
Since associated functions are unsupported as of now, an alternative to associated functions would be using (prefixed) global functions instead. A developer could then utilize the generated, global functions and wrap them within target-language specific type-associated extension functions.

While it's possible without any further ado to simply extend Protocols, Classes & Structs in Swift even with associated functions and generated properties, any Kotlin type is required to provide a companion object in order to mimic or enable such extensions (there's no static or associated in Kotlin, hence a companion object is required).
Kotlin does not provide a capibility to attach a companion object from anywhere but the types's declaration itself.

This PR therefore attaches a `companion object` to any generated `data class`, `class` or `enum class` by default.

**Edit**: Add `interface` companions as well.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceTemplate.kt
@@ -17,6 +17,7 @@ public interface {{ type_name }} {
     {%- else -%}
     {%- endmatch %}
     {% endfor %}
+    companion object
 }
 
 // The ForeignCallback that is passed to Rust.
--- a/uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
@@ -11,6 +11,7 @@ enum class {{ type_name }} {
     {% for variant in e.variants() -%}
     {{ variant|variant_name }}{% if loop.last %};{% else %},{% endif %}
     {%- endfor %}
+    companion object
 }
 
 public object {{ e|ffi_converter_name }}: FfiConverterRustBuffer<{{ type_name }}> {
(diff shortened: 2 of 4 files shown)


mhammond:
Would it be possible to add a test/demo of the capabilities this unlocks? Or at least a thorough description here? In particular, I'm wondering if that idiom already makes sense for other languages we support (meaning Kotlin is the current outlier) or whether it doesn't actually make sense for these other languages (meaning Kotlin becomes an outlier)?

Akazm:
> Or at least a thorough description here? In particular, I'm wondering if that idiom already makes sense for other languages we support (meaning Kotlin is the current outlier) or whether it doesn't actually make sense for these other languages (meaning Kotlin becomes an outlier)?

Ruby & Python allow monkey runtime patching, Swift provides the keyword `extension` for any `protocol`, `class` or `struct`. This allows a Python, Ruby or Swift developer to extend uniffi's generated code the way (s)he desires with _associated_* functions and properties.

*the term _associated_ might be misleading when it comes to Python & Ruby, hence italic

Kotlin, on the other hand, requires an _opt-in_ to enable this.

I'll let example code do the talking here, where we pretend that `Foo` is a class generated by uniffi **without** this PR and Bar is a class generated by uniffi **containing** the changes contained in this PR:

```Kotlin
class Foo
class Bar {
    companion object
}
// This compiles, but to no effect, as we'll see later
fun Foo.helloWorld() { println("Hello, world!") }

// "This is fine"
fun Bar.Companion.helloWorld() { println("Hello, world!") }

fun main(args: Array<String>) {
    Foo.helloWorld() // Unresolved reference: helloWorld
    Bar.Companion.helloWorld() // Everything's fine!
}
```

Does that _clear the fog_?

mhammond:
Thanks, it does! I've no objection, but also know almost nothing about Swift (as you saw!) so I'll let this sit for a few days in case someone more knowledgeable chimes in, otherwise we can merge it next week.

mhammond (approved):

PR #1766, opened 2023-09-25
Split up FfiConverter

bendk:
We've been discussing this for a while, here's my attempt at it.  This makes it so the macro code stops using `FfiConverter` and starts using more specialized traits `Return`, `Lift`, etc.  The `FfiConverter` trait lives on, but just as a simple way to implement all the traits that we actually use in the generated code.

This is rebased on top of the async-cancellation branch because that one had some changes to how we handled callback returns.  Ignore all commits except the last.  If we end up not taking that branch, I think it should be pretty easy to rebase over main too.

### DIFF
--- a/uniffi_core/src/ffi_converter_impls.rs
+++ b/uniffi_core/src/ffi_converter_impls.rs
@@ -22,17 +22,18 @@
 /// consumer crates.  To do this, it defines blanket impls like `impl<UT> FFIConverter<UT> for u8`.
 /// "UT" means an abitrary `UniFfiTag` type.
 use crate::{
-    check_remaining, ffi_converter_default_return, ffi_converter_rust_buffer_lift_and_lower,
-    lower_into_rust_buffer, metadata, try_lift_from_rust_buffer, FfiConverter, LiftRef,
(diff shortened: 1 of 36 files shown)


uniffi_core/src/ffi_converter_impls.rs
```diff
@@ -333,84 +287,123 @@ unsafe impl<UT, T: FfiConverter<UT>> FfiConverter<UT> for Option<T> {
...
-/// the key must always be of the String type.
-///
-/// HashMaps are currently always passed by serializing to a buffer.
-/// We write a `i32` entries count followed by each entry (string
-/// key followed by the value) in turn.
-/// (It's a signed type due to limits of the JVM).
-unsafe impl<K, V, UT> FfiConverter<UT> for HashMap<K, V>
+// Support for associative arrays via the FFI.
+// Note that because of webidl limitations,
+// the key must always be of the String type.
```
jplatte, line 351: … if UDL is used?

jplatte (approved):
Very nice! I have some ideas for follow-ups, but want to experiment with that myself after this is merged :)

mhammond (approved):
I agree, very nice!

Hywan (commented):
Really nice!

PR #1768, opened 2023-09-26
Swift cancellation

bendk:
Added a way for users to opt-in to async cancellation checks on Swift. This means that if the parent Task is cancelled, the exported Rust function will raise a `CancellationError`.  It's opt-in because this means that the function will always be marked `throws` forcing users to add `try` to all calls.
    
Reworked the `rustCall*` functions a bit:
  * Split `uniffiRustCallAsync` into a cancellable and non-cancellable
    version.
  * Added the `uniffi` prefix `rustCall` and `rustCallWithError`.
  * `uniffiRustCall` is no longer marked `throws`, since it's not
    expected to throw.  `uniffiRustCallWithError` is still marked
    throws.
  * `uniffiRustCallAsync` now inputs a complete closure that calls the
    FFI complete function with either `uniffiRustCall` or
    `uniffiRustCallWithError`
  * The result of this is that we can now use `rethrows` and Swift will
    figure out if an async function will throw or not.


bendk:
I personally could go either way on this one.  If I was writing an async API, I think I would handle cancellation at the Rust layer: give users a method to set a cancellation flag, periodically check that in your async function, and return a placeholder value if you see it.  On the other hand, I could see someone wanting to handle cancellation this way and it only affects users who opt in.

jplatte:
I think it heavily depends on what's the usual thing to do in Swift.. Would a lower-level async function (continuation?) throw a cancellation error when it notices the task has been cancelled? Would it be declared as throwing such that consumers of the API have to deal with that fact?

stefanceriu:
Talked this over with @pixlwave, we basically want both:

1) Let's take for example NSURLSession which will throw a cancellation error if the request times out i.e.
```
func data(
    from url: URL,
    delegate: (URLSessionTaskDelegate)? = nil
) async throws -> (Data, URLResponse)
```

This is useful when the task can get cancelled internally, without the user manually calling cancel on it or releasing the reference to the main task

2) Tasks that we create and cancel ourselves from the outside generally do not need to inform us back with the cancellation confirmation. We just cancel and forget. On the swift side we can check `Task.isCancelled` and bail out early, cleanup etc.

3) Continuations don't play a part in this as they're only useful for making non-async methods async


That being said we don't necessarily believe that having all async rust functions throw is much of a problem

pixlwave:
@jplatte Just to explore your response a little bit, let me ask a question:

What if the responsibility on whether it throws comes down to the Rust side of FFI? For example, when a task is cancelled it is the exported function's responsibility to check for the cancelation (is that even possible in Rust, I don't know futures). Then either of the following is possible.
- The function sees the cancellation and the author decides it isn't important to communicate back through the FFI so it isn't written as returning a `Result` (so Uniffi exports as non-throwing).
- The function sees the cancellation but the author wants to communicate the cancellation back through the FFI so it will be written as returning a `Result` and in turn, Uniffi exports it with `throws` on the Swift side.

This is fundamentally how it works on the Swift-side. It's up to the API author to decide what to do with the cancellation within a Task and to throw or not which I guess is why its hard to know how the 2 fit together.

All this said, we're happy with option 2 if the above doesn't make sense in the context of Rust.

jplatte:
Yeah unfortunately the above doesn't make sense in Rust, as Rust futures are cancelled by dropping them; there is no standardized way in Rust to "kindly" request a future to be cancelled, but giving it a chance to do some cleanup work. The only way to do cleanup is hooking into the destruction of the future via an `impl Drop for MyFuture` (which is itself only possible for hand-written futures, not `async fn`s, and also causes some problems but that's where we are at).

Akazm:
> Talked this over with @pixlwave, we basically want both:
> 
> 1. Let's take for example NSURLSession which will throw a cancellation error if the request times out i.e.

Does it, though? I cannot find any documentation about that. What I can find, however, is an `URLError` within Apple's documentation - which provides a [static property `timedOut`](https://developer.apple.com/documentation/foundation/urlerror/2293002-timedout) to allow catched errors to be identified as caused by a timeout.

> Continue running, likely doing some pointless work and waiting for other Swift code higher up the stack to short-circuit what it is doing.

I second this. Checking Task cancellation is, in my experience, a) semantical in nature and b) an explicit opt-in (see what Apple Engineer John McCall has to say about it [here](https://forums.swift.org/t/what-happens-if-a-task-is-not-canceled/55299/2)).

This means: In case Rust performs too much of work here at once without checking cancellation, the solution might be to split the workload into different subtasks after all, each of them itself performing their own cancellation checks. As you guys probably all now, checking whether the current Task is cancelled or not is done via static funcs and properties. If there's no other way, I might have to pass some kind of callback interface to Rust instead, allowing me to access anything required to do cancellation checks.

So without attempting to offend anyone, maybe the goal is entirely off of uniffi's scope here after all?

Akazm:
> I'm not a fan of bringing Swift's cancellation design to Rust, as it will almost always only be skin-deep, i.e. it will either be ignored entirely or people will write (very ugly) boilerplate to cancel the "entire" Rust future (not propagating the cancel-flag inside sub-futures) by dropping it upon noticing that the Swift task was cancelled.

It's just my $0.02, of course.

Kotlin Native does suffer from the same issue without providing a solution for timeouts in general, at least when it comes to networking, so there's that - It's a complicated issue and there'll be no solution without disadvantages.

jplatte:
Absolutely, thanks for sharing. Just wanted to make clear why I don't like that solution, didn't intend to make it sound like it's not an option at all or something like that.

bendk:
There doesn't seem to be much energy for pushing this forward, is there anyone out there wants this feature?  If not, I guess we should close this.

TheBlueMatt-macbook:
Is it possible to only support cancellation in the Swift -> Rust direction? Cancellation can be rather trivially mapped to the Rust `Future` simply being `Drop`ped and no longer `poll`ed, which is native to Rust `Future`s. This avoids the question of how to handle cancellation when Rust `Future`s call into async `Swift` code.

bendk:
> Is it possible to only support cancellation in the Swift -> Rust direction? Cancellation can be rather trivially mapped to the Rust `Future` simply being `Drop`ped and no longer `poll`ed, which is native to Rust `Future`s. This avoids the question of how to handle cancellation when Rust `Future`s call into async `Swift` code.

It seems possible to me, I think it would mean catching the cancellation error [here](https://github.com/mozilla/uniffi-rs/blob/0b50fa388d282f67ef12403ac1e70881921d2db3/uniffi_bindgen/src/bindings/swift/templates/Async.swift#L69-L71) and also in the `uniffiTraitInterfaceCallAsync` function below, then calling `rust_future_cancel` in those cases.  One question I have, is what if a switch function is awaiting the Rust Future?  I guess it would just never complete in that case, but maybe that's acceptable for some use-cases.

If you're interested in this functionality, please open an issue or better yet a PR.

PR #1773, opened 2023-09-30
Allow uniffi_traits to be used with procmacros.

### LINKED ISSUE #1548: Allow exporting Debug implementations
I think it would be very nice to be able to export `Debug` implementations for object types as a special builtin method that hooks into the target languages most fitting debug-formatting mechanism, e.g.

* For Python, generate a `__repr__` method
* For Swift, generate a `CustomDebugStringConvertible` implementation

I imagine it would look something like this in Rust:

```rust
#[derive(Debug, uniffi::Object)]
#[uniffi(export(Debug))]
struct Foo {
    // fields
}
```

Allowing the same for `Display` would be interesting as well, though of course not all the target languages necessarily have two builtin string conversion functionalities like Rust. Maybe target languages that don't could then just generate a regular `to_string` / `toString` method for `Display` (and something like a `debug` method if there's no inherent mechanism for that)?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-265)
┆Issue Number: UNIFFI-265

### LINKED ISSUE #1750: Implement builtin traits for proc-macros
After #1747 is merged, [builtin traits](https://mozilla.github.io/uniffi-rs/udl/interfaces.html#exposing-methods-from-standard-rust-traits) will be the last bits of scaffolding that are generated by the Askama templates and not the proc-macros.  We should fix that so that a) proc-macros support those traits and b) we have a single code-path for scaffolding code.

mhammond:
The proc-macro mechanism is also used for UDL which allows a lot of scaffolding to be removed.

Fixes #1750, fixes #1548.

### DIFF
--- a/uniffi_macros/src/export/utrait.rs
+++ b/uniffi_macros/src/export/utrait.rs
@@ -1,0 +1,168 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use proc_macro2::{Ident, TokenStream};
+use quote::quote;
+use syn::ext::IdentExt;
+
+use super::{attributes::ExportAttributeArguments, gen_ffi_function};
+use crate::fnsig::FnSignature;
+use uniffi_meta::UniffiTraitDiscriminants;
+
+pub(crate) fn expand_uniffi_trait_export(
+    self_ident: Ident,
+    uniffi_traits: Vec<UniffiTraitDiscriminants>,
+) -> syn::Result<TokenStream> {
+    let udl_mode = false;
+    let mut impl_items = Vec::new();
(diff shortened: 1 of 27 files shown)


uniffi_macros/src/export/utrait.rs
```diff
@@ -0,0 +1,168 @@
...
+    let syn::Item::Fn(item) = item else {
+        unreachable!()
+    };
+
+    let ffi_func = gen_ffi_function(
+        &FnSignature::new_method(self_ident.clone(), item.sig.clone())?,
+        &ExportAttributeArguments::default(),
+        udl_mode,
+    )?;
+    // metadata for the trait itself.
```
bendk, line 165: This is the same as the comments above, but I think it should point out the differences between the two types of metadata.

bendk (approved):
Looks great to me, I just had a docstring suggestion.

It's kind of amazing how much this removes from uniffi-bindgen.

PR #1780, opened 2023-10-05
Allow library mode to be used by external generators

arg0d:
The dependency between `lib.rs` and `library_mode.rs` is a bit convoluted, but I think this turned out quite elegant.


mhammond (approved):
This looks great, thanks.

Issue #1352, opened 2022-09-29
Generated functions should use interfaces when possible (arg, return types)


jmartinesp:
Hi all, first thanks for the amazing work on the bindings generator, we're using it as a base for our shared library across Android, iOS and web and it's been incredibly helpful so far.

Today I was trying to create a fake of one of our Rust classes in Kotlin and I saw that for every exported type there is a "_TypeName_Interface" interface or protocol created. I started creating the fake for this interface but I had to stop as soon as I got to a function that returned a struct with some impl from our project since the function returned the concrete generated type that extends from `FFIObject` instead of the interface.

```udl
interface Bar {
   void bar();
}

interface Foo {
    Bar foo();
}
```

This would create:

```kotlin
class Bar(pointer: Pointer) : FFIObject(pointer), BarInterface {
    override fun `bar`() { ... }
}

class Foo(pointer: Pointer) : FFIObject(pointer), FooInterface {
    override fun `foo`(): Bar { ... }
}
```

* If I tried to create a fake of that class, I couldn't do it because the `Foo` class is not open, so you can't extend it.
* There is not way to return a fake that extends from the interface since the return type is the concrete class, not the interface.

The problem seems to be the same on iOS, but with protocols instead of interfaces. Maybe I'm doing something wrong?

It's probably worth noting that while this _might work_ for unit testing using some mocking library, which I haven't really tested yet, my current use case is that one of our Android views indirectly calls our Rust code to retrieve some values and this causes some tools such as the UI layout preview to crash when they try to load the generated native library that the concrete generated classes use underneath. So my goal is to have those fakes not only for testing, but for other use cases where I can't load the native library.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-202)
┆Issue Number: UNIFFI-202

mhammond:
It does make sense that the interface should be described in terms of other interfaces where possible, but it just hasn't bitten us yet. IOW, I think a PR for this is likely to be accepted.

jmartinesp:
Fixed by https://github.com/mozilla/uniffi-rs/pull/1815.

Issue #1578, opened 2023-06-02
Support foreign implements of trait interfaces


bendk:
We recently added support for [trait interfaces](https://mozilla.github.io/uniffi-rs/udl/interfaces.html#exposing-traits-as-interfaces) (AKA sending Box<dyn Trait> from Rust to the foreign code).  We should extend that code to also allow the foreign code to pass classes that implement the interface to APIs that expect a trait object.  In the manual example, that means if they define a `ForeignButton` class that implements `Button`, they can pass that to `press()`..

This would create some nice symmetry and combine well with #1576.

It also addresses a real use-case we have in application-services.  We have Rust sync engines for our components, but for historical reasons we also have sync engines defined on the foreign side and don't want to transition away right now.  It would be really great if we could make it so `SyncManager` could seamlessly work with both types of engines at once.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-277)
┆Issue Number: UNIFFI-277

bendk:
The main question I have about this is if we should give some support for using concrete types rather than always using type-erased versions.  The main issue I see is when Rust passes a foreign object back to the foreign code.  In this case, it might be nice for UniFFI to automagically convert it back to the concrete object, rather than passing back some sort of type-erased trait object.  This eliminates a lot of overhead, since you can directly call methods on the object.  With a trait object, you'd have to lower/lift the arguments across the FFI to pass them to the scaffolding function, then the scaffolding function would lower/lift them again to call into the callback interface.

I think that case could be handled if UniFFI inserted a `uniffi_lower` method into the trait.  The default implementation would do what we currently do and create a `Box<Arc<dyn trait>>` and pass that over the FFI.  But for the type that we create for callback interfaces, we would pass something different back across the FFI.  Then the foreign code could distinguish between the 2 cases and sometimes recover the concrete object.

This sounds like more complexity than it's worth to me, but I wanted to mention for completeness.

Another way to address this would be to create an actual vtable for the trait interface.  Foreign objects would use a vtable would have functions on the foreign side.  This avoids the round-tripping issue that I mentioned above, although I think you would still need to lower/lift the arguments once.

PR #1782, opened 2023-10-08
Kotlin: use interfaces in bindings

jmartinesp:
Some time ago, I created https://github.com/mozilla/uniffi-rs/issues/1352 because we had some issues when testing code in unit tests in Kotlin, since all the values sent to and returned from the bindings were the concrete implementations in Rust, not their interfaces. I tried a first attempt to create this PR at the time, but I had very little knowledge of UniFFi so I couldn't make it work as I wanted to.

In this PR I try to make every function that exposes a type with a Kotlin interface in its bindings use that interface instead of the concrete implementation, so it's easier to replace it with fakes or alternative implementations in the Kotlin code.

For that I created the `fn protocol_label(&self) -> String` and `fn has_protocol(&self) -> bool` in `Type`, and set the right values in the `Type::Object` case, and the needed changes in the different Kotlin templates. I added a test for Map, since it wasn't being tested with anything other than primitive types, but I'm unsure if further tests are needed. 

This means that some impl like this:
```rust
impl Foo {
  fn use_bar(bar: Arc<Bar>): Foo
}
impl Bar {}
```

Will now produce:

```kotlin
interface FooInterface : Disposable {
  fun useBar(bar: BarInterface): FooInterface
}
interface BarInterface : Disposable {}

// And their class counterparts:
class Foo(pointer: Pointer): FFIObject(pointer), FooInterface {
  override fun useBar(bar: BarInterface): FooInterface
  ...
}
```

Record templates and enums should also return bindings with interfaces instead of classes.

Also, the interfaces for the generated class types now implement the `Disposable` interface, to try to ease their usage without casting them to the concrete implementations in the receiving code: any fake or alternate implementation should just leave the `destroy` funs empty. However, while this was done to prevent API breakage, it won't be possible to prevent it completely: since all functions previously returned concrete classes instead of the interfaces, any code that used that will have to move to using the interfaces instead. When I tried it with our codebase, it meant 15 minutes of adding the needed `Interface` suffixes. I guess it would be great if this output could be configurable, but I'm not sure how to do that.

Also, I'm not an expert in either Rust or templating, so any suggestions to improve the code are welcome 😅 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
 - Proc-macros: The `handle_unknown_callback_error` attribute is no longer needed for callback
   interface errors
 - Foreign types can now implement trait interfaces
-- Kotlin binginds now use interfaces wherever possible.
+- Kotlin bindings now use interfaces wherever possible.
 
 ### What's Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - Proc-macros: The `handle_unknown_callback_error` attribute is no longer needed for callback
   interface errors
 - Foreign types can now implement trait interfaces
+- Kotlin binginds now use interfaces wherever possible.
 
 ### What's Fixed
 
--- a/uniffi_bindgen/src/backend/types.rs
+++ b/uniffi_bindgen/src/backend/types.rs
@@ -16,10 +16,6 @@ pub trait CodeType: Debug {
     /// method signatures and property declarations.
     fn type_label(&self) -> String;
 
-    fn protocol_label(&self) -> String {
-        self.type_label()
-    }
-
     /// A representation of this type label that can be used as part of another
     /// identifier. e.g. `read_foo()`, or `FooInternals`.
     ///
@@ -74,9 +70,5 @@ pub trait CodeType: Debug {
     /// Function to run at startup
     fn initialization_fn(&self) -> Option<String> {
         None
-    }
-
-    fn has_protocol(&self) -> bool {
-        false
     }
 }
--- a/uniffi_bindgen/src/backend/types.rs
+++ b/uniffi_bindgen/src/backend/types.rs
@@ -16,6 +16,8 @@ pub trait CodeType: Debug {
     /// method signatures and property declarations.
     fn type_label(&self) -> String;
 
+    fn protocol_label(&self) -> String { self.type_label() }
+
     /// A representation of this type label that can be used as part of another
     /// identifier. e.g. `read_foo()`, or `FooInternals`.
     ///
--- a/uniffi_bindgen/src/backend/types.rs
+++ b/uniffi_bindgen/src/backend/types.rs
@@ -71,4 +71,6 @@ pub trait CodeType: Debug {
     fn initialization_fn(&self) -> Option<String> {
         None
     }
+
+    fn has_abstraction(&self) -> bool { false }
 }
--- a/uniffi_bindgen/src/backend/types.rs
+++ b/uniffi_bindgen/src/backend/types.rs
@@ -16,7 +16,9 @@ pub trait CodeType: Debug {
     /// method signatures and property declarations.
     fn type_label(&self) -> String;
 
-    fn protocol_label(&self) -> String { self.type_label() }
+    fn protocol_label(&self) -> String {
+        self.type_label()
+    }
 
     /// A representation of this type label that can be used as part of another
     /// identifier. e.g. `read_foo()`, or `FooInternals`.
@@ -74,5 +76,7 @@ pub trait CodeType: Debug {
         None
     }
 
-    fn has_abstraction(&self) -> bool { false }
+    fn has_protocol(&self) -> bool {
+        false
+    }
 }
--- a/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs
@@ -27,7 +27,7 @@ impl CodeType for ObjectCodeType {
     fn protocol_label(&self) -> String {
         if self.has_protocol() {
             let mut label = self.type_label();
-            let index = self.type_label().find("?").unwrap_or(label.len());
+            let index = self.type_label().find('?').unwrap_or(label.len());
             label.insert_str(index, "Interface");
             label
         } else {
--- a/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs
@@ -20,8 +20,20 @@ impl ObjectCodeType {
 }
 
 impl CodeType for ObjectCodeType {
+
     fn type_label(&self) -> String {
         super::KotlinCodeOracle.class_name(&self.name)
+    }
+
+    fn protocol_label(&self) -> String {
+        if self.id == "Object" {
+            self.type_label()
+        } else {
+            let mut label = self.type_label();
+            let index = self.type_label().find("?").unwrap_or(label.len());
+            label.insert_str(index, "Interface");
+            label
+        }
     }
 
     fn canonical_name(&self) -> String {
--- a/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs
@@ -20,19 +20,18 @@ impl ObjectCodeType {
 }
 
 impl CodeType for ObjectCodeType {
-
     fn type_label(&self) -> String {
         super::KotlinCodeOracle.class_name(&self.name)
     }
 
     fn protocol_label(&self) -> String {
-        if self.id == "Object" {
-            self.type_label()
-        } else {
+        if self.has_protocol() {
             let mut label = self.type_label();
             let index = self.type_label().find("?").unwrap_or(label.len());
             label.insert_str(index, "Interface");
             label
+        } else {
+            self.type_label()
         }
     }
 
@@ -51,5 +50,7 @@ impl CodeType for ObjectCodeType {
         }
     }
 
-    fn has_abstraction(&self) -> bool { true }
+    fn has_protocol(&self) -> bool {
+        self.id != "Object"
+    }
 }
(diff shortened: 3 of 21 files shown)


CHANGELOG.md
```diff
@@ -36,6 +36,7 @@
 - Error types must now implement `Error + Send + Sync + 'static`.
 - Proc-macros: The `handle_unknown_callback_error` attribute is no longer needed for callback
   interface errors
+- Kotlin binginds now use interfaces wherever possible.
```
badboy, line 39:
```suggestion
- Kotlin bindings now use interfaces wherever possible.
```

badboy (approved):
Code-wise this looks fine.
I'm unsure about the implications of `Disposable` though and in general would like someone elses look on this.

@mhammond would you take a look?

uniffi_bindgen/src/backend/types.rs
```diff
@@ -16,6 +16,10 @@ pub trait CodeType: Debug {
     /// method signatures and property declarations.
     fn type_label(&self) -> String;
 
+    fn protocol_label(&self) -> String {
```
mhammond, line 19: "protocol" doesn't mean anything to other bindings IIUC, and there's a chance that it means something *different* in some external languages. Do you think it might be possible to keep this abstraction in the Kotlin bindings? It seems like that's where some types get an overridden value, so if we can encapsulate the defaults there too it seems like it might be an improvement.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs
```diff
@@ -20,11 +20,26 @@ impl CodeType for ObjectCodeType {
...
+
     fn canonical_name(&self) -> String {
         format!("Type{}", self.id)
     }
 
     fn literal(&self, _literal: &Literal) -> String {
         unreachable!();
     }
+
+    fn has_protocol(&self) -> bool {
```
mhammond, line 42: `id` here was recently renamed to `name` to avoid confusion about what `id` actually means - so checking the name of an object being exactly "Object" doesn't seem correct?

uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -3,7 +3,7 @@
 {{- self.add_import("java.util.concurrent.atomic.AtomicLong") }}
 {{- self.add_import("java.util.concurrent.atomic.AtomicBoolean") }}
 
-public interface {{ type_name }}Interface {
+public interface {{ type_name }}Interface : Disposable {
```
mhammond, line 6: Can you please explain a little more about why `Disposable` is now needed on interfaces. I think we are probably going to move *away* from `Disposable`, albeit with an opt-in for `AutoCloseable`

uniffi_meta/src/types.rs
```diff
@@ -141,6 +141,19 @@ impl Type {
         };
         Box::new(std::iter::once(self).chain(nested_types))
     }
+
+    pub fn needs_unchecked_upcast_in_return(&self) -> bool {
```
mhammond, line 145: I don't quite follow this - why does only a `Map<Object, Object>` need this?

mhammond (changes requested):
I'd really like to see more of this exercised in one of the kotlin fixtures - probably `test_coverall.kts` - in particular, demonstrating the benefits of this in that fixture - eg, supplying a "mock" impl of the interface instead of the concrete implementations, and in as many contexts as possible (eg, maps, vecs, in args and return values etc). This will help demonstrate the value of the change and help those of us without much Kotlin experience to understand the attraction.

After the next iteration I'll probably also ask @jhugman to have a poke too!

uniffi_meta/src/types.rs
jmartinesp: It's actually a Map where either key or value or both types are objects. In Kotlin, `Map<Int, FooInterface>` and `Map<Int, Foo>` are different types, so when we read those from Rust through `lift` we get a `Map<Int, Foo>` but the return type of the function is actually `Map<Int, FooInterface>` and it can't be casted implicitly.

uniffi_bindgen/src/backend/types.rs
jmartinesp: I can try, maybe I can make it work with some filter, but it depends on what info I have available there.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs
jmartinesp: That's unfortunate. Thing is, we need to add the `Interface` suffix to any object that comes from the Rust SDK, but sometimes there are functions that return `java.lang.Object` and will try to also append the suffix. Is there any way to check the fully qualified name?

uniffi_bindgen/src/backend/types.rs
mhammond: You might be able to add it to the `KotlinCodeOracle` too? That can get quite confusing, feel free to ping us if you get stuck.

uniffi_meta/src/types.rs
jmartinesp: Surprisingly, I didn't find this same issue with `List`. I'll double check, just in case.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs
jmartinesp: I know I ran into this issue so I had to add this exception, but I can't remember where. I'll remove it and see where the tests fail.

bendk:
This overlaps with the work that I was doing for #1578.  For that issue, I added the [object_names](https://github.com/mozilla/uniffi-rs/blob/ca35eda00bba3ae316154dfd7d9faa3a68d20c75/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs#L327-L337) function, which returns both the interface name and the concrete class name.  These complement `type_name` which is the general function to get the name of some type.  Note that there's two cases here:

- For objects, `concrete_name == type_name` and `interface_name` is different
- For trait interfaces, `interface_name == type_name` and `concrete_name` is different.

This matches up pretty well with how lowering works.  For regular objects, we can only lower the concrete type.  Any function that inputs them needs to input the concrete type, a dict/enum/vec that stores the type should probably store the concrete type, since otherwise that dict/enum/vec wouldn't be lowerable, etc.  For trait objects, it's the opposite.  All functions can input any type that implements the interface, dicts/enums/vecs should store the interface not the concrete type, etc.

So, while there are some particular cases where you could replace the concrete type with the protocol, like a Rust function return, in general I think this is going to be hard.  There are other options though:

  - We could make the concrete type open.  It's slightly risky because if you don't have an actual handle to a Rust object, then calling a UniFFI function will fail.  But I would be okay with saying that users accept that risk if they extend the generated class.  For your use case of only creating the objects if the native library isn't available, it seems like it should work fine.
  - How hard would it be to refactor things so your application's functions inputted the interface type instead of the regular type?  This seems like a "correct" way to handle this issue, but maybe it would require a lot of boilerplate to wrap everything.
  - You could use the trait interfaces.  The idea here is you define a Rust trait that can be implemented by either Rust or the foreign side and both of those can be used interchangeably.  Your objects wouldn't really be fakes anymore, they would just be an alternate implementation.  Be warned that the initial implementation just landed yesterday and I'm still actively working on it, but I'm pretty confident it will be ready by the time of the next release.  Also, even if it works perfectly, it means the Rust code would have to change to take `Arc<dyn Trait>` instead of `Arc<ConcreteType>` and you'd pay the performance penalty of dynamic dispatch.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/object.rs
jmartinesp: Oh, my mistake. It was a struct called `Object`, it was the test that needed to change.

jmartinesp:
> We could make the concrete type open. It's slightly risky because if you don't have an actual handle to a Rust object, then calling a UniFFI function will fail. But I would be okay with saying that users accept that risk if they extend the generated class. For your use case of only creating the objects if the native library isn't available, it seems like it should work fine.

That might actually be the easiest way to achieve this. But in that case, having the `{class_name}Interface` interface for structs doesn't really have any use, I think? Unless I missed some internal usage of these interfaces we could just generate the class, which being now open could be extended. Extending has the slight drawback of not forcing you to provide new implementations for existing methods, so you could end up calling actual FFI code way more easily than using interfaces.

> How hard would it be to refactor things so your application's functions inputted the interface type instead of the regular type? This seems like a "correct" way to handle this issue, but maybe it would require a lot of boilerplate to wrap everything.

That's what I tried after the first implementation of this PR. For a large project it took me ~15min to replace all usages of concrete Rust types with their interfaces. I also had to replace the `use { ... }` functions we have been using with another import, but it was quite easy.

> You could use the trait interfaces. The idea here is you define a Rust trait that can be implemented by either Rust or the foreign side and both of those can be used interchangeably. Your objects wouldn't really be fakes anymore, they would just be an alternate implementation. Be warned that the initial implementation just landed yesterday and I'm still actively working on it, but I'm pretty confident it will be ready by the time of the next release. Also, even if it works perfectly, it means the Rust code would have to change to take Arc<dyn Trait> instead of Arc<ConcreteType> and you'd pay the performance penalty of dynamic dispatch.

That's what I was using the generated interfaces for, so it seems a bit redundant to have an actual class, and interface that apparently does nothing (again, I might be wrong) and then an interface that can be implemented. It also means more work for our Rust team if I'm not mistaken, since they'd have to re-define every object they pass to us to use these trait interfaces, including those inside `dictionaries` or `enums`.

------

In any case, I'll upload some new changes rebasing the previous work in `main` and using `KotlinCodeOracle` for deciding when to use the interface version. Those seem to work with the current tests, I'll try creating a few new ones later. If I have the time, I'll also try the 1st approach about using `open class` in a new branch.

uniffi_bindgen/src/backend/types.rs
jmartinesp: Done in https://github.com/mozilla/uniffi-rs/pull/1782/commits/451480c7ac8fcbda5b0d0cf83305ebcb351fa447

jmartinesp:
Closing this, as https://github.com/mozilla/uniffi-rs/pull/1815 was merged as a simpler implementation of what was proposed here.

PR #1783, opened 2023-10-09
Update docs re uniffi.toml and tweaks to a couple of examples

mhammond:
Better docs for `uniffi.toml`, in preparation for a Python enhancement.


bendk (approved):
Looks good!

PR #1784, opened 2023-10-09
Add support for `external_packages` to Python so Python packages can …

### LINKED ISSUE #1776: External types seem to prevent (easy) distribution
Here's the gist of it by example: I have two crates (rust_ffi and tiny_lib). rust_ffi depends on tiny_lib. Both have UDLs. rust_ffi references a type from the tiny_lib. When I generate the Python code for rust_ffi, it has imports pointing to tiny_lib (e.g. `from tiny_lib import TinyLibType`). I can generate tiny_lib's Python code as well and that seems like it works in general, but after being packaged together in a wheel the import breaks; when the import statement runs in `rust_ffi.py` it searches for a package named `tiny_lib` and fails to find it, completely ignoring the adjacent file with the same name.

I have made a [small example](https://github.com/DrGodCarl/uniffi-example) of this problem. Its README is a bit out of date from what I've described above, but the setup and problem are essentially the same. I have also tried using a megazord and generating using library mode but the Python files generate the same way and exhibit the same behavior. When I modify the `rust_ffil.py` in the wheel to import from `.tiny_lib` it works exactly as expected, though that's not really sustainable for a larger project.

It's possible I've missed how I can package these up in the docs somewhere. If that's the case, I'd love to be pointed to those docs and I apologize for opening an issue.

mhammond:
…be used.

When using external types, the expectation is that most consumers will want to use a Python package which holds all the modules - therefore, imports for external types defaults to `from .module import name`. This behaviour can be changed to support any package name, including just doing a regular top-level module import - which we leverage for some tests.

Fixes #1776

(While this seems sensible to me, feel free to push back on either the implementation or the default)

### DIFF
--- a/docs/manual/src/python/configuration.md
+++ b/docs/manual/src/python/configuration.md
@@ -8,10 +8,41 @@ The generated Python modules can be configured using a `uniffi.toml` configurati
 | ------------------ | -------  |------------ |
 | `cdylib_name`      | `uniffi_{namespace}`[^1] | The name of the compiled Rust library containing the FFI implementation (not needed when using `generate --library`). |
 | `custom_types`      | | A map which controls how custom types are exposed to Python. See the [custom types section of the manual](../udl/custom_types#custom-types-in-the-bindings-code)|
+| `external_packages` | | A map which controls the package name used by external packages. See below for more.
 
+## External Packages
 
-## Example
+When you reference external modules, uniffi will generate statements like `from module import Type`
+in the referencing module. The `external_packages` configuration value allows you to specify how `module`
+is formed in such statements.
 
+The value is a map, keyed by the crate-name and the value is the package name which will be used by
+Python for that crate. The default value is an empty map.
+
+When looking up crate-name, the following behavior is implemented.
+
+### Default value
+If no value for the crate is found, it is assumed that you will be packaging up your library
+as a simple Python package, so the statement will be of the form `from .module import Type`,
+where `module` is the namespace specified in that crate.
+
+Note that this is invalid syntax unless the module lives in a package - attempting to
+use the module as a stand-alone module will fail. UniFFI just generates flat .py files; the
+packaging is up to you. Eg, a build process might create a directory, create an `__init__.py`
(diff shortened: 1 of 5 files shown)


uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -94,6 +96,16 @@ impl Config {
...
+
+    /// Get the package name for a given external namespace.
+    pub fn module_for_namespace(&self, ns: &str) -> String {
+        let ns = ns.to_string().to_snake_case();
+        match self.external_packages.get(&ns) {
+            None => format!(".{ns}"),
+            Some(value) if value.is_empty() => ns,
+            Some(value) => format!("{value}.{ns}"),
+        }
+    }
```
jhugman, line 108: Neat.

jhugman (approved):
This looks good to me! Also: nice docs!

docs/manual/src/python/configuration.md
```diff
@@ -8,10 +8,34 @@ The generated Python modules can be configured using a `uniffi.toml` configurati
 | ------------------ | -------  |------------ |
 | `cdylib_name`      | `uniffi_{namespace}`[^1] | The name of the compiled Rust library containing the FFI implementation (not needed when using `generate --library`). |
 | `custom_types`      | | A map which controls how custom types are exposed to Python. See the [custom types section of the manual](../udl/custom_types#custom-types-in-the-bindings-code)|
+| `external_packages` | `.` | A map which controls the package name used by external packages.
```
badboy, line 11: This is a bit confusing in the docs, the default is given as `.` but the text says this should be a map.
I would expect that I can set `external_packages = "."` to get the same default behavior, but that doesn't work.
And thus now from only the docs it's not clear how to configure this. It's then explained in the next lines, maybe the table should call this out? "Look further down for details"

badboy (approved):
small note, but otherwise good to go!

docs/manual/src/python/configuration.md
mhammond: yeah, thanks, I made a few other tweaks to make that clearer too.

PR #1785, opened 2023-10-09
Fix async wake behavior

bendk:
If wake() is called while we are in the middle of polling a future, then we should immediately call our continuation function to schedule another poll.


jplatte (approved):

jplatte:
cc @Hywan

jplatte:
Empirically, this fixes all of our problems with the rewritten async methods.

Hywan (approved):
LGTM :-).

PR #1786, opened 2023-10-10
Normalize 'generate_bindings' with 'generate_external_bindings'

arg0d:
`generate_external_bindings` was lacking `crate_name` field. This PR should ensure that in the future `generate_external_bindings` has all the same fields and functions identical to ` generate_bindings`.


arg0d:
Same here, I'm not too familiar with these options. I'm just trying to maintain consistent CLI interface in external generators.

arg0d:
Now that you mention it, `crate_name` CLI description doesn't seem to match what its actually doing. The description seems like it was correct for 0.24, but not anymore. I did a quick trace through the code in `main`, and it seems `crate_name` is included in FFI function names?

mhammond:
Yeah, "crate name" and "module path" are used somewhat interchangeably, but that's the primary use. My main "problem" with library mode and crate_name is that when using library mode there are potentially *many* crate names involved, and that in general, the crate that actually created the library isn't really any more special than any other.

Issue #1394, opened 2022-11-02
Revisit decision to not support destructors for interface objects on Kotlin.


bendk:
Early on, we made a decision to use explicit destructors for Kotlin (#8).  I think we should revisit that decision and consider if it's still what we want.

The main biggest issue with the destructors is that they are quite a footgun.  If any consumer forgets to call `destroy()`, then we're leaking memory.  As a concrete example, what kind of care are we taking to ensure that app-services [places connections](https://github.com/mozilla/application-services/blob/main/components/places/src/places.udl#L31) are properly destroyed?  I've never once thought about it while working on this code.  I did some quick grepping and AFAICT, the android-components code never calls `destroy()`, but luckily it caches the connections which ends up preventing the leaks

This gets more footgunny when you deal with nested objects.  If you have a record that has on interface object field, then you also need to call `destroy()`.  The same would go for Enums, except we have code that panics if you actually try define an Enum with a variant that contains an interface object (#1372).  I wonder if the reason for that panic is because we didn't want to deal with this issue.  Errors would be even more problematic, since errors might only get caught by a top-level catch-all function (OTOH, I think preventing errors from containing objects is reasonable).  Lastly, if you had a function that returned a sequence/map of interface objects, then the caller would need to iterate and call `destroy()` on each element.

The secondary issue is that it adds complexity to our code and our decision making.  Tracking which objects could maybe contain an interface object is not trivial.  Also, each time we consider lifting a restriction on what can contain an interface object, it's harder because we need to consider what to do about `destroy()` on Kotlin.

I suspect the limitations from 2020 might not apply anymore.  Can we use the `Cleaner` object on Android now?  Or at least have a non-buggy `finalize()` method?  I don't know, but maybe someone with knowledge can chime in here.

I personally think we should move to a more flexible system and let the user choose.   In `uniffi.toml` they could choose if they wanted the current system, a `Cleaner`-based system, or a `finalize()` based system.  I would also consider allowing one of the latter systems + an explicit `destroy()` method as an option.

What do others think?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-210)
┆Components: UniFFI
┆Epic: Maintenance collector
┆Issue Number: UNIFFI-210
┆Sprint: Backlog

mhammond:
Thanks for bringing this up - I don't have a strong opinion yet, but think we do need to tread carefully. 

[...]

PR #1787, opened 2023-10-10
ADR-0008: Freeing objects and resources

bendk:
We've been discussing this one quite a bit lately and it seems like a great time to make a decision.  Here's the start of an ADR which lays out all the possibilities I've heard so far, I'd love to hear what others think.  I'm including everyone that I can remember having a discussion about this, but all voices are welcome.

### DIFF
--- a/docs/adr/0008-freeing-objects-and-resources.md
+++ b/docs/adr/0008-freeing-objects-and-resources.md
@@ -1,0 +1,158 @@
+# Freeing objects and resources
+
+* Status: Accepted
+* Deciders: Uniffi developers, as approved in PR-1787
+* Date: 2023-10-18
+
+Prevous discussion: [PR 1394](https://github.com/mozilla/uniffi-rs/issues/1394), [Issue 8](https://github.com/mozilla/uniffi-rs/issues/8)
+
+## Context and Problem Statement
+
+UniFFI objects always hold a reference that needs to be released and sometimes hold a resource that needs to be closed.
+The current design conflates the two issues.
+We can address both concerns better and improve ergonomics by dealing with the two separately.
+
+This ADR is written in language-neutral terms so that the issue can be discussed in general.
+However it's most clear in the Kotlin implementation and this ADR will affect that code the most, see Appendix 1 for the proposed changes.
+
+## Releasing References
+When a Rust object is passed to the foreign side, the foreign code stores a reference to a Rust `Arc<>`.
+Conversely, foreign objects passed into Rust (eg, callbacks or trait implementations) cause Rust to hold a reference to the foreign object.
+
+UniFFI needs to ensure that when the generated objects are destroyed, the reference on the other side of the FFI is released.
+It is acceptable for this release to be delayed and/or non-deterministic as long as it eventually happens.
+In particular, it is acceptable for it to be handled by garbage collection.
+
+## Closing Resources
+
+When an object holds a significant resource (eg, some scarce operating-system owned resource), it may not be acceptable to couple the resource to the lifetime of the foreign object.
+In this case, it MUST be to possible close the resource while the foreign object is still alive and before its reference is released.
+
+These same considerations apply in both directions across the FFI.
+If a garbage-collected foreign object stores a resource, then there MUST be a way to close the resource before the garbage collection happens.
+If a Rust object holds a resource, there MUST be a way to close that resource even if there may be other `Arc<>` references.
+
+## Object containers
+
+The discussion above has focused on individual objects, but in any non-trivial application many objects interact.
+For example, consider a simple `Enum` or `dictionary` - in the usual case, these objects hold references to a non-resource holding objects.
+Given the discussion above, it's acceptable for these references to be released using the standard memory management of the foreign language.
+
+However, the `Enum`, `dictionary`, `Vec`, `HashMap`, etc. may contain an object that itself contains a resource.
+While the component author is able to offer methods to close the resource on the objects themselves, they are not able to introduce methods to all the container objects which might hold the object.
+So the question here is whether UniFFI itself should try and offer such capabilities.
+
+## Foreign "sugar" for closing resources.
+
+Objects which hold significant resources offer the capability to closing those resources even while the object remains alive.
+For example, most foreign languages which expose a "file" object tend to provide a `close()` method that ensures the handle is released to the operating system.
+
+For almost all languages it is possible use something like a `try/finally` pattern to ensure the object is closed as soon as practical.
+This pattern is so common that some languages offer syntactic sugar for this express purpose.
+For example, Python offers a "context manager", Kotlin offers the `use` extension for objects that implement `Closable`, etc.
+
+Ideally, UniFFI would offer a way so consumers can opt-in to implementing this capability.
+This must be opt-in because that opt-in process must tell UniFFI exactly how the resource should be closed - while it will be very common for this to be via a `close()` method, UniFFI can't assume this (eg, a Python context manager doesn't insist on any particular technique)
+
+## Decision Drivers
+
+* UniFFI should present simple and clear abstractions to users.
+
+* UniFFI should let users explicity decide how they want to deal with resources.
+  There is no way UniFFI can automatically differentiate between objects that hold significant resources versus those that do not.
+  Thus, components must "opt in" somehow to this facility, where the opt-in could be anywhere from fully manual closing (eg, by supplying a close method and documenting it must be called ASAP) or something more formal (ie, telling UniFFI this is such an object and UniFFI supplying some builtin support for this concept)
+
+* There is nothing UniFFI can do by itself to ensure that Rust references are released when foreign objects are destroyed.
+  This responsibility must fall on the foreign language in one way or another.
+
+* It seems unreasonable for UniFFI to be expected to track object containers and propagate any close operations as this would quickly become an ergonomic and technical quagmire (eg, would be give it to all objects just in-case such an object might be added in the future?
+  If not, how would the user (or even UniFFI itself) know whether the container actually held such an object? etc)
+
+## Considered Options
+
+### [Option 1] UniFFI handles releasing references, closing resources is largely left up to the user
+
+* Rely on all languages having *some* strategy for automatically closing resources as they are no longer reachable.
+  For most languages this will be something like a destructor, or some other facility with destructor-like semantics.
+  If a language has no such capability, then the bindings will be forced to implement some special facility for each and every object to be explicitly freed, but this strategy is unlikely to be useful in practice.
+
+* Don't automatically implement any methods to close a resource; if an object has resources that should be closed while the object remains alive, the component author should supply a mechanism to do that.
+
+* Investigate providing a facility for objects to opt-in to "foreign sugar" for closing the resources.
+  The exact mechanics for this are beyond the scope of the ADR.
+
+* Offer some general advice summarizing this in the documents, like "define a close() method and document it", but otherwise leave this problem for users to solve.
+
+* Don't do anything special for object containers.
+  If the field of a record contains a resource, then it's the user's responsibility to properly close that field.
+
+### [Option 2] Resource and Object as separate types
+
+* Require UDL users to label their interfaces as either a `Resource` or `Object` and proc-macro users to derive either `Resource` or `Object`.
+* `Object` instances are assumed to not contain significant resources and we rely on the foreign memory management for releasing the reference.
+* `Resource` instances would have a formal mechanism for closing their resource and releasing their reference at the same time.
+  Optionally, we may also use the foreign memory management as a fallback to close the resource and release the reference if the user did not do so explicitly.
+
+### [Option 3] Make Object do both (ie, assume all objects should get close methods)
+
+* Like option 2, but treat all objects as though they have significant resources to be freed.
+
+## Pros and Cons of the Options
+
+### [Option 1] UniFFI handles releasing references, closing resources is largely left up to the user
+
+* Good, because gives the user control over their object namespace.  UniFFI doesn't define any methods or assume any methods exist, etc
+* Bad, because it relies on a contractual agreement to ensure resource leaks are avoided and doesn't help if those obligations aren't met.
+* Bad, because it requires the "closeable" object to deal with some extra state - for example, if an object provides a `close()` method
+  which may be called before the object is destroyed, the object must deal with (a) the possibility that the object continues to be used
+  in some capacity after this close and (b) that as the object is destroyed `close()` may or may not have been called.
+* Good, because the same general solution works for when Rust is holding a foreign reference.
+* Bad, because generating "foreign sugar" means each object must opt-in to this behaviour instead of getting it for free.
+* Good, because making resource closing the user's responsibility extends to object containers as well.
+  If a library authors decides that it makes sense to define an enum variant that contains some resource-holding object, then it's the library consumer's responsibility to ensure that resource is properly closed.
+
+### [Option 2] Resource and Object as separate types
+
+* Good, because users explicitly decide on their interface behavior.
+* Good, because it means we could infer the "foreign sugar" situation and automatically implement special interfaces like `AutoCloseable`
+  in Kotlin for a Context Manager in Python.
+* Bad, because it means all objects must use the same signature for closing objects, whereas, eg, a connection to a database
+  object might prefer the method to be called `close_connection` and might even allow parameters.
+* Bad, because it implicitly defines some methods (eg, `close()`), preventing users from defining methods with those names that have different semantics.
+* Bad, because it doesn't offer a reasonable solution for when Rust is holding a foreign reference.
+* Good, because the distinction makes deciding how to handle object containers easier.
+  For example, enums can contain `Object` fields, but not `Resource` fields.
+
+### [Option 3] Make Object do both (ie, assume all objects should get close methods)
+
+* Good, because it's the easiest to implement.
+* Good, because we can easily continue to implement special traits like `AutoCloseable`.
+* Bad, for all the reasons Option 1 is bad.
+* Bad, because it's not clear how to handle object containers in all cases.  Should enums be allowed to contain objects?
+
+## Decision Outcome
+
+Chosen option: **[Option 1] UniFFI handles releasing references, closing resources is largely left up to the user**
+
+# Appendix 1 - what does this mean for our foreign bindings?
+
+The above is all quite abstract - what does this mean in practice for our language bindings:
+
+## Python
+
+* Python bindings are already hooked up to `__del__` so nothing changes here.
+* Once we have the "Foreign Sugar" opt-in mechanism in place we will generate context managers for such objects.
+
+## Swift
+
+* Existing `deinit` implementation is suitable so nothing changes here.
+
+## Kotlin
+
+* UniFFI will implement a mechanism that automically frees Rust references when the corresponding foreign object is destroyed.
+  [Effective Java, 3rd edition](https://www.oreilly.com/library/view/effective-java-3rd/9780134686097/) recommends never using destructors.
+  It recomends avoiding the [Cleaner API](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Cleaner.html), but does say that it may be appropriate for freeing a "native peer", like the UniFFI reference.
+  The Android docs recommend using the [ReferenceQueue](https://www.android-doc.com/reference/java/lang/ref/ReferenceQueue.html) API, which is the low-level class that `Cleaner` uses (https://www.android-doc.com/reference/java/lang/Object.html#finalize())
+
+* We will no longer generate the `Closable` or `AutoClosable` interface for objects.
+* Once we have the "Foreign Sugar" opt-in mechanism in place we will generate the `Closable` interface for all such objects (maybe `AutoClosable` as well although that is mostly only useful for Java interoperability).


docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Decision Drivers
+
+* If an object does not hold a resource, then we should be able to use the GC to free it.
+* The solution should work for both Rust objects passed to the foreign language and foreign objects
+  passed to Rust.
+
+## Considered Options
+
+### [Option 1] User-defined close method
```
bendk, line 42: Sammy suggested this one to me recently and I have to admit that I'm biased towards it.  Please point out anything I'm missing because of that bias, but I like it because a) it's the least opinionated and b) AFAICT, it's the only solution that also works for callback interfaces.

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,109 @@
...
+
+We should come up with a general approach for resource-holding objects and apply it consistently
+across all languages.
+
+
+## Decision Drivers
+
+* If an object does not hold a resource, then we should be able to use the GC to free it.
+* The solution should work for both Rust objects passed to the foreign language and foreign objects
+  passed to Rust.
```
bendk, line 38: IMO at least, but this isn't a hard requirement.  We could chose one option for objects and a different option for callabck interfaces (although [trait interfaces](https://github.com/mozilla/uniffi-rs/issues/1578) could get tricky).

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,109 @@
...
+### [Option 2] Resource and Object as separate types
+
+* Require UDL users to label their interfaces as either a `Resource` or `Object` and proc-macro
+  users to derive either `Resource` or `Object`.
+* Each type would have slightly different handling.  For example on Kotlin `Resource` would work
+  like an `Object` does today and `Object` would work like the `Object` from [1].
+* For other languages, we could either introduce similar distinctions or make both types behave
+  the same.
+
+### [Option 2] Make Object do both
```
jplatte, line 70:
```suggestion
### [Option 3] Make Object do both
```

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,109 @@
...
+  languages (unlike Rust, where dropping an object releases any resources).
+* Good, because it's the most explicit.
+* Good, because gives the user control over their object namespace.  UniFFI doesn't define any methods for them.
+* Bad, because it's not idiomatic to call `close()` on Rust or other RAII-based languages.
+* Bad, because it requires Rust and other RAII-based language implementations to manually check if
+  the resource object has been closed or not.
+* Bad, because it relies on the `Cleaner` Java API, which is not available on very old versions of
+  Android.
+
+### [Option 2] Resource and Object as separate types
```
jplatte, line 94: I'd say it also has the downside of adding more complexity. The other two options don't really introduce any new concepts in the public API from my POV, this one does.

docs/adr/0008-freeing-objects-and-resources.md
jplatte: I'm actually against this, I think in contexts where RAII is available, we should use it. But maybe I'm misunderstanding something as I wasn't even aware that freeing of callback interface objects was ever a problem, and this ADR doesn't really provide any details as to *why* it's a problem (I would have thought dropping the `Box<dyn Trait>` would do the right thing,though now I'm wondering how it's even possible that Rust gets unique ownership over such an object, I guess it actually doesn't and the `Box` points to a GC-managed object?).

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,109 @@
...
+  implement `AutoCloseable`.
+* In addition, Kotlin objects also use the `Cleaner` API to ensure they are destroyed if the
+  user doesn't manually call it.
+
+## Pros and Cons of the Options
+
+### [Option 1] User-defined resource close method
+
+* Good, because it's the only option that supports callback interfaces (and trait interfaces in the
+  future).  If a foreign object is passed into Rust and that object holds a resource, Rust needs to
```
arg0d, line 82: Foreign object holding a resource would be a callback implementation on foreign language side, right? Any resources being held by the callback are user defined, right? So it should be possible to generate a "close" method for the callback that is called when Rust invokes callback with `0` to remove the callback from handle map. User could then implement the appropriate logic to cleanup any resources being held by the callback. Closing the callback seems to be already implemented, only `close` method is lacking in the interface to allow foreign bindings to close further resources.

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,109 @@
...
+
+Currently, our Kotlin bindings try to address this problem for Objects by adding a `destroy()` method that
+manually drops the Rust value and implementing the `AutoCloseable` interface which allows users to
+ensure this is called via the `use` extension function.  However, this is not sufficient for several
+reasons:
+
+  * It is specific to Kotlin, but many languages face similar issues.
+  * All objects get this handling, when only some need it.
+  * This system is tied to decision to *not* support freeing UniFFI objects via GC-based systems
+    like `Cleaner`, which is not correct for some interfaces (see #1394)
```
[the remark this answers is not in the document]
bendk: I linked #1394 for the opposite reason: to show that sometimes we should rely on the GC.  The sentence is confusing because there's a weird double-negative in there though.  I'll try to clean that up.

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,109 @@
...
+
+* Status: Propsed
+* Deciders: ?
+* Date: 2023-10-10
+
+Prevous discussion: [PR 1394](https://github.com/mozilla/uniffi-rs/issues/1394), [Issue 8](https://github.com/mozilla/uniffi-rs/issues/8)
+
+## Context and Problem Statement
+
+UniFFI must support objects that hold resources like database connections, file handles, network sockets, etc.  This is difficult, because objects like this are handled differently by different languages:
```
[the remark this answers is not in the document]
arg0d: > Java/Kotlin/Python programmers make a distinction between memory and other types of resources

this sentence clears up ideas in this ADR, you should include it in the context.

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,105 @@
...
+
+* Status: Propsed
+* Deciders: ?
+* Date: 2023-10-10
+
+Prevous discussion: [PR 1394](https://github.com/mozilla/uniffi-rs/issues/1394), [Issue 8](https://github.com/mozilla/uniffi-rs/issues/8)
+
+## Context and Problem Statement
+
+UniFFI must support objects that hold resources that should be cleaned up promptly like database connections, file handles, network sockets, large memory allocations, etc.
```
mhammond, line 11: Thanks heaps for driving this forward Ben - it's a discussion we keep having that never seems to make progress!

Initially I'm just putting down some thoughts on the problem statement, as I believe tightening this up might help the rest.

#1394 says we could add "finalizer support to make a 'best effort' attempt at freeing what it can while leaving the auto-closeable pattern in place" - I'm not aware of us exploring that, nor how this ADR might change if that turned out to be successful.

This section starts off talking about all bindings in general, but then fairly quickly pivots into just Kotlin. While Kotlin is important, I think it's still worth a bit more of an explanation about how the other "important" bindings work. For example, it seems that Python also shares the same general problem - a context-manager is available, but this doesn't really work for long lived references. #8 implies Swift does not have this problem. C# and Go seem worth mentioning too as I suspect they might end up being very popular UniFFI bindings.

Upon re-reading this ADR, I think your "Option 1" can be restated as "Do nothing. It is up to the designer of the API exposed via UniFFI to take care of this. An option might be to add explicit "close()" methods, but that's all on you, we don't care". If that's accurate, maybe update the wording to be more explicit/brutal? We've also been talking about "transitive requirements (eg, enum holding a dictionary which holds a resource - there's nothing generic you can do on the enum to clean this tree up) - if the intention is also to offer no help there and leave that to the API designer it might be worth calling that out.

I also don't really understand the callback example - or more specifically, how the callback example is any different from the general problem. If you consider a pure-python program which takes some arbitrary object, there's also no generic way for the holder of that object to release the resources held by that arbitrary object - but that's not generally considered a problem - the solutions are either (a) know the type, so, eg, if it's a file you know you can .close() it, or (b) don't do that. IOW, the responsibility for this on the designer of the API so that either (a) the API has an explicit close-type method, or (b) don't require the Rust code to keep a long-lived reference to an object holding resources you care about?

tl;dr - I think I'm also in favour of Option 1 if I understand it correctly - and if I do, I'd suggest:
* Making that option even more explicit and "brutal" - ie, explicitly call out "this is your problem, UniFFI doesn't help"
* Explicitly call out the transitive case (ie, if your enums hold objects which manage resources, you are probably going to have a bad time, but UniFFI isn't going to try and stop you)
* Either remove most of the callback language (ie, replace it with something like "this is just a slightly different view of the exact same problem") or clarify how callbacks differ.
* Make the problem statement slightly less Kotlin specific, or explain how, eg, Python is different.

And probably most importantly, work out which Kotlin people to put this in-front of. Most UniFFI core contributors know enough Kotlin to be dangerous, but probably not quite enough to know how viable this will be in practice.

docs/adr/0008-freeing-objects-and-resources.md
mhammond: > Foreign object holding a resource would be a callback implementation on foreign language side, right?

I think we want to generalize callbacks into more generic traits though - eg, wrap them in an `Arc<>` such that Rust code can be blissfully unaware whether a trait implementation is foreign or rust implemented.

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,105 @@
...
+### [Option 1] User-defined close method
+
+* Don't automatically implement any methods to close a resource.
+* Advise users to define a `close()` or similar method for resource-holding objects / callback
+  interfaces / trait interfaces and ensure it's called.  This also likely means defining an error
+  variant like `Error::Closed` that gets returned any methods that use the resource after the
+  `close()` method is called.  Alternatively, handle this by panicking or throwing an error.
+* Use the `Cleaner` API on Kotlin to automatically free UniFFI pointers for Kotlin objects.
+  Note that this combines with the previous bullet.  `close()` releases the resource, the Cleaner
+  action frees the heap allocation.
```
jplatte, line 47: I didn't know anything about this API, nor much about "oldschool" finalizers, but https://stackoverflow.com/questions/52879761/should-java-9-cleaner-be-preferred-to-finalization points towards `Cleaner` really being the best choice here.

docs/adr/0008-freeing-objects-and-resources.md
bendk: Thanks, I think these are great suggestions.  I "brutalized" option 1 a bit, I could have gone farther, but I think it makes the point pretty well now.  I added some text about the transitive case and also other languages.  I agree it shouldn't be Kotlin specific.

I moved callbacks to their own section and tried to elaborate a bit more.  On the one hand, I think that it's all part of the same problem, but I also think they're important because they provide extra complications for any automatically generated solution.

Next step is to find some more Kotlin people...

docs/adr/0008-freeing-objects-and-resources.md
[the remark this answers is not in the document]
linabutler: I also had the same question at first, and wanted to share how I thought through it, in case it's helpful for context or other folks!

IIUC, from the perspective of the foreign code, a UniFFI object is "a pointer to some memory that exists outside of the GC heap". Because it's outside the GC heap, it can't be GCed automatically.

That's why these objects need some kind of destructor, either:

1. An explicit method, like `close()` or `destroy()`, or...
2. A finalizer that's automatically called by the foreign language's runtime. In the JVM, that's `java.lang.Object.finalize()` or `Cleaner`; in Python, that's `__del__`; in Go, that's `runtime.SetFinalizer`, and so on.

On the Rust side, the object might have a handle to a "resource"—like a database connection, a file handle, etc.—or it might not, but that's totally opaque to the foreign code.

In that sense, I don't think UniFFI's `AutoCloseable` machinery today is concerned about "resources" specifically—it's a destructor that tells Rust to drop the object, and free up the memory it uses. Whether or not dropping the object also happens to release "resources" on the Rust side—for example, a rusqlite connection implementing `Drop` to automatically close the connection—is an implementation detail from Kotlin's perspective.

jhugman:
> Objects can hold resources that should be cleaned up promptly like database connections, file handles, network sockets, large memory allocations, etc.
> Many languages, like Kotlin, Java and Python make a distinction between these types of resources and GC heap allocations that is reflected in the API.
> Programmers typically expose a `close` method and languages often offer constructs to assist in using these like context managers on Python or the `use` extension function on Kotlin.
> However, on Rust and other RAII-based languages, these resources are typically released in the destructor rather than with a dedicated API.

I'm not sure that this is the problem to which the solution proposed is solving.

The problem we have been discussing is this:

- most language expose some kind of finalizer where we can inform Rust to free up memory when the language no longer needs it:
  - In Swift we implement `deinit`
  - In Python we implement `__del__`
  - In Ruby we define with `ObjectSpace.define_finalizer`

In Kotlin, we might implement `java.lang.Object.finalize` ([spec](https://docs.oracle.com/javase/specs/jls/se17/html/jls-12.html#jls-12.6)). However:
 - it has been [deprecated since JDK 9](https://bugs.openjdk.org/browse/JDK-8165641),
 - anecdotally, it has been observed as the source of crashes on Android. 
 - before it was deprecated, it was the subject of heavy warnings against using it– so is a source of JVM diversity and little usage.

Indeed, the [Android documentation on `java.lang.Object#finalize`](https://www.android-doc.com/reference/java/lang/Object.html#finalize()) is very instructive here, and worth quoting extensively:

> Note that objects that override finalize are significantly more expensive than objects that don't. Finalizers may be run a long time after the object is no longer reachable, depending on memory pressure, so it's a bad idea to rely on them for cleanup. Note also that finalizers are run on a single VM-wide finalizer thread, so doing blocking work in a finalizer is a bad idea. A finalizer is usually only necessary for a class that has a native peer and needs to call a native method to destroy that peer. Even then, it's better to provide an explicit close method (and implement [Closeable](https://www.android-doc.com/reference/java/io/Closeable.html)), and insist that callers manually dispose of instances. This works well for something like files, but less well for something like a BigInteger where typical calling code would have to deal with lots of temporaries. Unfortunately, code that creates lots of temporaries is the worst kind of code from the point of view of the single finalizer thread.
>
> If you must use finalizers, consider at least providing your own [ReferenceQueue](https://www.android-doc.com/reference/java/lang/ref/ReferenceQueue.html) and having your own thread process that queue.

For our case, the mitigations from (in chronological order) Josh Bloch's Effective Java, the JDK deprecation of finalize and the Android doc all point in the same direction:

- implement `Closeable` and `AutoCloseable`, and insist our users manually dispose of their instances.
- implement a Cleaner, or something else that polls a `ReferenceQueue`.
- don't rely on either happening as soon as the object drops out of scope.

So far, we have implemented `AutoCloseable` and `Closeable`; and this is the source of impedance mismatch and confusion. 

1. As @linabutler [alludes](https://github.com/mozilla/uniffi-rs/pull/1787#discussion_r1355939897), JVM heap memory isn't something that Java/Kotlin programmers need to worry about, so most things that needs a finalizer are open and closeable, e.g. sockets, files, databases. However non-heap memory doesn't fall cleanly into either of these categories: it does need cleanup, but isn't openable and closeable. It feels strange to `close` an only-in-memory object, but that's currently what we advise to avoid memory leaks.

2. To the untrained eye, `AutoCloseable` seems to do something useful, however it is really only useful in a specific use case: namely interacting with the so-called [`try-with-resources`](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html). 

Kotlin doesn't have a try-with-resources, but does implement [`use`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html) which does the same job. The [implementation of `use`](https://github.com/JetBrains/kotlin/blob/7a7d392b3470b38d42f80c896b7270678d0f95c3/libraries/stdlib/jvm/src/kotlin/io/Closeable.kt#L21-L44) is worth looking at, if only to see how little the tooling around `AutoCloseable` and `Closeable` gives use.

3. Part of uniffi's value proposition is that we try and expose Rust crates to foreign languages in such a way as using them is indistinguishable from using a foreign language equivalent. We should look at the existence of `Closeable` and forcing users to explicitly call `close` on `interface` structs is a leak in the abstraction we should seek to close.

---
**TL;DR**: 

1. We should move away from using `Closeable` and `AutoCloseable`, in favour of a cleaner based approach, to bring up to parity with the other languages we support.
2. We should not replicated `close` methods for other languages. If `__del__` is deprecated in Python we should investigate alternatives (e.g. `ContextManager`).

jhugman (commented):
> Objects can hold resources that should be cleaned up promptly like database connections, file handles, network sockets, large memory allocations, etc.
> Many languages, like Kotlin, Java and Python make a distinction between these types of resources and GC heap allocations that is reflected in the API.
> Programmers typically expose a `close` method and languages often offer constructs to assist in using these like context managers on Python or the `use` extension function on Kotlin.
> However, on Rust and other RAII-based languages, these resources are typically released in the destructor rather than with a dedicated API.

I'm not sure that this is the problem to which the solution proposed is solving.

The problem we have been discussing is this:

- most language expose some kind of finalizer where we can inform Rust to free up memory when the language no longer needs it:
  - In Swift we implement `deinit`
  - In Python we implement `__del__`
  - In Ruby we define with `ObjectSpace.define_finalizer`

In Kotlin, we might implement `java.lang.Object.finalize` ([spec](https://docs.oracle.com/javase/specs/jls/se17/html/jls-12.html#jls-12.6)). However:
 - it has been [deprecated since JDK 9](https://bugs.openjdk.org/browse/JDK-8165641),
 - anecdotally, it has been observed as the source of crashes on Android. 
 - before it was deprecated, it was the subject of heavy warnings against using it– so is a source of JVM diversity and little usage.

Indeed, the [Android documentation on `java.lang.Object#finalize`](https://www.android-doc.com/reference/java/lang/Object.html#finalize()) is very instructive here, and worth quoting extensively:

> Note that objects that override finalize are significantly more expensive than objects that don't. Finalizers may be run a long time after the object is no longer reachable, depending on memory pressure, so it's a bad idea to rely on them for cleanup. Note also that finalizers are run on a single VM-wide finalizer thread, so doing blocking work in a finalizer is a bad idea. A finalizer is usually only necessary for a class that has a native peer and needs to call a native method to destroy that peer. Even then, it's better to provide an explicit close method (and implement [Closeable](https://www.android-doc.com/reference/java/io/Closeable.html)), and insist that callers manually dispose of instances. This works well for something like files, but less well for something like a BigInteger where typical calling code would have to deal with lots of temporaries. Unfortunately, code that creates lots of temporaries is the worst kind of code from the point of view of the single finalizer thread.
>
> If you must use finalizers, consider at least providing your own [ReferenceQueue](https://www.android-doc.com/reference/java/lang/ref/ReferenceQueue.html) and having your own thread process that queue.

For our case, the mitigations from (in chronological order) Josh Bloch's Effective Java, the JDK deprecation of finalize and the Android doc all point in the same direction:

- implement `Closeable` and `AutoCloseable`, and insist our users manually dispose of their instances.
- implement a Cleaner, or something else that polls a `ReferenceQueue`.
- don't rely on either happening as soon as the object drops out of scope.

So far, we have implemented `AutoCloseable` and `Closeable`; and this is the source of impedance mismatch and confusion. 

1. As @linabutler [alludes](https://github.com/mozilla/uniffi-rs/pull/1787#discussion_r1355939897), JVM heap memory isn't something that Java/Kotlin programmers need to worry about, so most things that needs a finalizer are open and closeable, e.g. sockets, files, databases. However non-heap memory doesn't fall cleanly into either of these categories: it does need cleanup, but isn't openable and closeable. It feels strange to `close` an only-in-memory object, but that's currently what we advise to avoid memory leaks.

2. To the untrained eye, `AutoCloseable` seems to do something useful, however it is really only useful in a specific use case: namely interacting with the so-called [`try-with-resources`](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html). 

Kotlin doesn't have a try-with-resources, but does implement [`use`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html) which does the same job. The [implementation of `use`](https://github.com/JetBrains/kotlin/blob/7a7d392b3470b38d42f80c896b7270678d0f95c3/libraries/stdlib/jvm/src/kotlin/io/Closeable.kt#L21-L44) is worth looking at, if only to see how little the tooling around `AutoCloseable` and `Closeable` gives use.

3. Part of uniffi's value proposition is that we try and expose Rust crates to foreign languages in such a way as using them is indistinguishable from using a foreign language equivalent. We should look at the existence of `Closeable` and forcing users to explicitly call `close` on `interface` structs is a leak in the abstraction we should seek to close.

---
**TL;DR**: 

1. We should move away from using `Closeable` and `AutoCloseable`, in favour of a cleaner based approach, to bring up to parity with the other languages we support.
2. We should not replicated `close` methods for other languages. If `__del__` is deprecated in Python we should investigate alternatives (e.g. `ContextManager`).

WDYT?

mhammond:
Thanks for the detailed reply James! A couple of questions:

Re (1), I always thought that `Closeable` and `AutoCloseable` were not mutually exclusive - so if a Kotlin object holds a database connection, `AutoCloseable` might be a good way to ensure the database connection is closed even while the object remains reachable. So while I understand the motivation of moving from `Closeable` to the cleaner, I don't see why it follows that we should also move away from `AutoCloseable`?

Re (2) I don't quite understand what you are saying there. Eg, re Python, are you proposing something *should* change re `__del__`, or just speculating what might happen if `__del__` was deprecated? But even if the latter, the reference to a `ContextManager` confuses me a little - IIUC, a `ContextManager` is analogous with `AutoClosable` in that it's a way to force reclamation of resources even while the object remains reachable. IIUC, `__del__` being deprecated *could not* naively be replaced with a context manager in the same way that Kotlin can't rely exclusively on `AutoCloseable` because, as you mention, it requires the consumer to use a very specific pattern for it to be effective.

Indeed, combining these points, my take away would be:
1) Python already has `__del__`, Swift already has `deinit`, the sanest thing Kotlin could do for parity there is the cleaner.
2) For more fine-grained control (ie, so there's a sane path to reclaiming resources even while objects remain reachable) our objects should *also* implement a `ContextManager` for Python, `AutoCloseable` for Kotin, (and ??? for Swift?*)
3) Given Rust uses `Arc<>`, dropping the Rust reference might not have the desired effect in practice, so we'd encourage UniFFI consumers to implement resource reclamation as part of their API. But (1) and (2) above means it's generally *not* necessary for *every* interface exposed via UniFFI to have explicit user-defined "close" methods, just ones with "expensive" resources associated with them.

(*I'm even less familiar with Swift than I am Kotlin, but IIUC swift objects use reference counting, so I can't see why the same arguments don't apply there - I've just no idea if there's a protocol or some other technique that might be suitable)

bendk:
@mhammond I agree with basically everything you said, except for one part:

> For more fine-grained control (ie, so there's a sane path to reclaiming resources even while objects remain reachable) our objects should also implement a ContextManager for Python, AutoCloseable for Kotin, (and ??? for Swift?*)

> Given Rust uses Arc<>, dropping the Rust reference might not have the desired effect in practice, so we'd encourage UniFFI consumers to implement resource reclamation as part of their API.

I think there's a tension between these 2 statements.  If we're suggesting to the user to define a `close()`-style method in some cases, then how can we implement `Autocloseable` for all objects?  I think the only way it works is our current system, where we assume that dropping the arc does release the resource.  That's likely true for Objects, but once you start mixing in callback interfaces / trait interfaces, I don't think it's true anymore.

I think we should start by simply not trying to implement AutoClosable, ContextManager, etc.

Later on, we could let users opt-in to this with something similar to how we all users to implement standard Rust traits.  If an interface defines `close` and they flag it as being "Closeable", then we implement AutoClosable, ContextManager, etc.

mhammond:
> I think there's a tension between these 2 statements. If we're suggesting to the user to define a close()-style method in some cases, then how can we implement Autocloseable for all objects?

I guess I was suggesting that we can document something simple like: A reference to the corresponding Rust object will be dropped when the object is garbage collected. You can force the reference to be dropped earlier by using the object in a `try/with` / `ContextManager`. Note however that dropping the Rust reference doesn't guarantee anything about the Rust object behind the reference, so you are encouraged to add close capabilities to your API".

> Later on, we could let users opt-in to this with something similar to how we all users to implement standard Rust traits.

Are they mutually exclusive? We are always going to be dealing with an `Arc<>`, so I guess I can't really see why a `ContextManager` can't *both* call some magic method *and* drop the reference?

OTOH, I guess an alternative is to stop supporting the ability for the Rust object to be "disconnected" from the foreign object entirely. According to the template code, after `Autocloseable::close()` is called consumers should expect to see `IllegalStateException` interacting with the object. Maybe the better thing to do is never actually disconnect and instead let that kind of exception bubble up from the "still connected but now logically closed" Rust object?

(Apologies if that was the intent of James's proposal which I missed!)

jhugman:
Question (1) 

> `AutoCloseable` might be a good way to ensure the database connection is closed even while the object remains reachable

For clarity: we don't need to implement `AutoCloseable` to do this. `AutoCloseable` is only to integrate with `try-with-resources`, which is only good for interacting with in a one-shot-in-the-same-block manner. As far as I can tell, we don't use `try-with-resources` (a Java only language feature) anywhere in Mozilla. 

Similarly, the only thing `Closeable` does is let us interact with the `use` function, which is pretty easy to re-implement.

@bendk is correct, both interfaces are almost identical: they each ask you to implement a single `close` method, which let you interact with a limited set of libraries which don't suit our use case.

Question (2) 

I must admit, my python is not what it should be. `__del__` looks exactly what we need at the moment. My cursory read of `ContextManager`s would be that it was slightly more capable as `try-with-resources`: it offers enter-block and exit-block semantics, where we actually need a longer living lifecycle and orthogonal to blocks.

If `__del__` did become deprecated (in favor of something else), I'd guess we'd use the something else, rather than `ContextManager`.

My point however still stands: we should bring Kotlin support to the same level we have for Python, Swift and Ruby— implicit memory handling is done implicitly, and any use-case-specific resource closing (where necessary/desired) is done explicitly in userland. 

> Maybe the better thing to do is never actually disconnect and instead let that kind of exception bubble up from the "still connected but now logically closed" Rust object?

This is the point @linabutler so eloquently made this week– by defauit we should avoid the state where we have orphaned/dangling Foreign Language objects that are still reachable, but effectively dead because the backing Rust object has been invalidated by an explict call from the foreign side; this is facility that the users can build for themselves, but we shouldn't enable it by default.

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,141 @@
...
+
+Chosen option: 
+
+# Appendix 1 - what does this mean for our foreign bindings?
+
+The above is all quite abstract - what does this mean in practice for our language bindings:
+
+## Python
+
+* Python bindings are already hooked up to __del__ so nothing changes here.
```
jplatte, line 126:
```suggestion
* Python bindings are already hooked up to `__del__` so nothing changes here.
```

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,141 @@
+# Freeing objects and resources
+
+* Status: Propsed
```
jplatte, line 3:
```suggestion
* Status: Proposed
```

Maybe you should add a [typos](https://github.com/crate-ci/typos) CI job 😉

jplatte (approved):
Very nice rewrite of the ADR text!

jplatte:
Given the previous discussion and updated ADR text, I also support option 1.

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,149 @@
...
+When a Rust object is  passed to the foreign side, the foreign code stores reference to a Rust `Arc<>`.
+Conversely, foreign objects passed into Rust (eg, callbacks or trait implementations) cause Rust to hold a reference to the foreign object.
+
+UniFFI needs to ensure that when the generated objects are destroyed, the reference on the other side of the FFI is released.
+It is acceptable for this release to be delayed and/or non-deterministic as long as it eventually happens.
+In particular, it is acceptable for it to be handled by garbage collection.
+
+## Closing Resources
+
+When an object holds a significant resource (eg, some scarce operating-system owned resource), it may not be acceptable to couple  the resource to the lifetime of the foreign object.
```
mhammond, line 21: nit, double-spaces here.

mhammond (commented):
Thanks heaps Ben! I have entirely nits (and a type CI job is an interesting idea :)

I guess we also need to consider making a decision here. I believe you and I are clearly on the "option 1" fence, and I doubt that will be controversial, so I guess we should actually fill that part of it in, then get formal approval from a few people. "Deciders" should probably just be something like "Uniffi developers, as approved in PR-XXXX" or similar? Either way, thanks for staying with this!

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Swift
+
+* Existing `deinit` implementation is suitable so nothing changes here.
+* Once we have the "Foreign Sugar" opt-in mechanism in place we will generate context managers for such objects.
+
+## Kotlin
+
+* We need some mechanism for destructors or destructor-like semantics so Rust can be notified as objects die.
+  [Effective Java, 3rd edition](https://www.oreilly.com/library/view/effective-java-3rd/9780134686097/) recommends avoiding destructors and using the [Cleaner API](Cleaner API: https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Cleaner.html) instead.
```
badboy, line 144:
```suggestion
  [Effective Java, 3rd edition](https://www.oreilly.com/library/view/effective-java-3rd/9780134686097/) recommends avoiding destructors and using the [Cleaner API](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Cleaner.html) instead.
```

mhammond (approved):

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,146 @@
...
+## Swift
+
+* Existing `deinit` implementation is suitable so nothing changes here.
+* Once we have the "Foreign Sugar" opt-in mechanism in place we will generate context managers for such objects.
+
+## Kotlin
+
+* We need some mechanism for destructors or destructor-like semantics so Rust can be notified as objects die.
+  [Effective Java, 3rd edition](https://www.oreilly.com/library/view/effective-java-3rd/9780134686097/) recommends avoiding destructors and using the [Cleaner API](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Cleaner.html) instead.
+* We will no longer generate the `Closable` or `AutoClosable` interface for objects.
```
jhugman, line 145: 🥂 \O/

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,146 @@
...
+
+Prevous discussion: [PR 1394](https://github.com/mozilla/uniffi-rs/issues/1394), [Issue 8](https://github.com/mozilla/uniffi-rs/issues/8)
+
+## Context and Problem Statement
+
+## Releasing References
+When a Rust object is  passed to the foreign side, the foreign code stores reference to a Rust `Arc<>`.
+Conversely, foreign objects passed into Rust (eg, callbacks or trait implementations) cause Rust to hold a reference to the foreign object.
+
+UniFFI needs to ensure that when the generated objects are destroyed, the reference on the other side of the FFI is released.
```
jhugman, line 15: Form nit: Is this the problem statement? 

I had to read the Context and Problem statement at least twice to identify this sentence as what the ADR considers the nub of the problem. I'm still unsure.

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,146 @@
...
+For example, most foreign languages which expose a "file" object tend to provide a `close()` method that ensures the handle is released to the operating system.
+
+For almost all languages it is possible use something like a `try/finally` pattern to ensure the object is closed as soon as practical.
+This pattern is so common that some languages offer syntactic sugar for this express purpose.
+For example, Python offers a "context manager", Kotlin offers the `use` extension for objects that implement `Closable`, etc.
+
+Ideally, UniFFI would offer a way so consumers can opt-in to implementing this capability.
+This must be opt-in because that opt-in process must tell UniFFI exactly how the resource should be closed - while it will be very common for this to be via a `close()` method, UniFFI can't assume this (eg, a Python context manager doesn't insist on any particular technique)
+
+## Decision Drivers
```
jhugman, line 50: Form nit: what are the things that make a good option?

e.g. Simplicity? Non-leaky abstractions? Performance?

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,146 @@
...
+
+* Python bindings are already hooked up to `__del__` so nothing changes here.
+* Once we have the "Foreign Sugar" opt-in mechanism in place we will generate context managers for such objects.
+
+## Swift
+
+* Existing `deinit` implementation is suitable so nothing changes here.
+* Once we have the "Foreign Sugar" opt-in mechanism in place we will generate context managers for such objects.
+
+## Kotlin
```
jhugman, line 141: Nit: Burying the lede. Considering we don't generate a `close` method for any other language except Kotlin, and every other language has destructor/finalizer arrangements that we already use, it is surprising that Kotlin is only spoken about in the last 5 lines of the ADR.

jhugman (approved):
A few comments on the form and problem statement of the ADR, but nothing to stand in the way of the action proposed.

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,146 @@
...
+
+* We need some mechanism for destructors or destructor-like semantics so Rust can be notified as objects die.
+  [Effective Java, 3rd edition](https://www.oreilly.com/library/view/effective-java-3rd/9780134686097/) recommends avoiding destructors and using the [Cleaner API](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/lang/ref/Cleaner.html) instead.
```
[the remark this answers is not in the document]
bendk: I updated the wording, WDTY?

docs/adr/0008-freeing-objects-and-resources.md
bendk: Good point, I added a bullet point and reworked another one to try to address this.

docs/adr/0008-freeing-objects-and-resources.md
bendk: I think there's a tension her.e.  As a practical matter, this ADR is essentially just suggestion changes to the Kotlin code.  As a theoretical matter, I think it's important to discuss the topic in general and separate it from all the details about finalizers, cleaners, etc.

I tried to strike a balance by mentioning this at the top of the doc, do you think that addresses the concern?

bendk:
I don't think there's been any concerns with the proposal, just suggestions about the ADR wording.  If anyone wants to argue for one of the other options, please speak up.

bendk:
I just realized something about the current Kotlin system: Since user code can cause the underlying handle to be destroyed, it's risky to ever pass it to Rust.  Most of the [ObjectRuntime.kt](https://github.com/mozilla/uniffi-rs/blob/39fc189c2e4b2441efd0198e2e99ec894c8332d0/uniffi_bindgen/src/bindings/kotlin/templates/ObjectRuntime.kt) file is dedicated to avoiding this for method receivers, but doesn't the same issue apply to function/method arguments?  IOW, if I pass an object to a function, what's to stop another thread from calling `destroy()` while we're in the middle of lowering everything?

tarikeshaq:
oops not sure why GitHub removed review from @arg0d, wanted remove the request for myself only 😄

skhamis (approved):
Super late to this party, but option 1 was something I supported and so LGTM!

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,159 @@
...
+
+UniFFI objects always hold a reference that needs to be released and sometimes hold a resource that needs to be closed.
+The current design conflates the two issues.
+We can address both concerns better and improve ergonomics by dealing with the two separately.
+
+This ADR is written in language-neutral terms so that the issue can be discussed in general.
+However it's most clear in the Kotlin implementation and this ADR will affect that code the most, see Appendix 1 for the proposed changes.
+
+## Releasing References
+When a Rust object is  passed to the foreign side, the foreign code stores reference to a Rust `Arc<>`.
```
linabutler, line 19:
```suggestion
When a Rust object is passed to the foreign side, the foreign code stores a reference to a Rust `Arc<>`.
```

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,159 @@
...
+## Foreign "sugar" for closing resources.
+
+Objects which hold significant resources offer the capability to closing those resources even while the object remains alive.
+For example, most foreign languages which expose a "file" object tend to provide a `close()` method that ensures the handle is released to the operating system.
+
+For almost all languages it is possible use something like a `try/finally` pattern to ensure the object is closed as soon as practical.
+This pattern is so common that some languages offer syntactic sugar for this express purpose.
+For example, Python offers a "context manager", Kotlin offers the `use` extension for objects that implement `Closable`, etc.
+
+Ideally, UniFFI would offer a way so consumers can opt-in to implementing this capability.
```
linabutler, line 54: I really like your separation of references and resources, and having UniFFI release references makes a lot of sense to me! But I'm not sure I follow the conclusion that UniFFI needs to concern itself with resources, and was wondering if we could talk more about that.

If a UniFFI consumer wants to define an `interface` type `T` that holds a resource—like a file or a database connection—they could manually define a `close` method, and have the Rust implementation of `T` hold a `Mutex<Option<Resource>>` that the implementation of `T::close()` method would `take()` to close the resource early. They could also `impl Drop for T`, which would take care of closing the resource once the last `Arc<T>` reference was released.

It's definitely true that, without an opt-in sugar, the UniFFI-generated Kotlin binding for `T` couldn't implement `Closeable` or `AutoCloseable`—`T` would just have a plain, vanilla `close()` method. (And, similarly, the Python binding would have a regular `close` method, and wouldn't be a context manager). But I wonder if that's okay?

I'm also wondering what this opt-in sugar would look like for languages like Swift (and maybe Ruby?) that have different idioms for resources. For example, Swift's `deinit` is the closest, but ARC doesn't guarantee precise lifetimes, and Swift seems to prefer the "`with`-prefixed function that takes a closure" idiom for resources ([`withLock`](https://developer.apple.com/documentation/os/osallocatedunfairlock/4061628-withlock), [`withExtendedLifetime`](https://developer.apple.com/documentation/swift/withextendedlifetime(_:_:)-31nv4), [`withUnsafe{Mutable}{Pointer, Bytes}`](https://developer.apple.com/documentation/swift/manual-memory-management)).

Sorry that was so wordy! I guess I'm wondering—would the opt-in sugar for resources offer enough benefits as a UniFFI feature, than having consumers handle resource management themselves?

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Python
+
+* Python bindings are already hooked up to `__del__` so nothing changes here.
+* Once we have the "Foreign Sugar" opt-in mechanism in place we will generate context managers for such objects.
+
+## Swift
+
+* Existing `deinit` implementation is suitable so nothing changes here.
+* Once we have the "Foreign Sugar" opt-in mechanism in place we will generate context managers for such objects.
```
linabutler, line 149: Hmmm, IIUC, Swift doesn't really have "context managers", but I was wondering if this was a typo / copy-paste from the Python section, or if there's something specific you meant by it?

linabutler (approved):
Thanks so much, @bendk!

docs/adr/0008-freeing-objects-and-resources.md
bendk: Nope, just a copy and paste error.

docs/adr/0008-freeing-objects-and-resources.md
bendk: Great question.  Right now, I fall on the side of probably implementing `Closeable` and/or `AutoClosable` on Kotlin, maybe implementing context managers on Python, and that's it.  I'm not sure what others feel like or even if that's where I'll end up once we actually test it out.

I think the line that says "The exact mechanics for this are beyond the scope..." was supposed to communicate that we don't really have a final decision here, however the other line does seem to imply that we want to do *something*.  I softened that a bit to just say, we'll investigate it and decide later.

docs/adr/0008-freeing-objects-and-resources.md
```diff
@@ -0,0 +1,158 @@
...
+* Python bindings are already hooked up to `__del__` so nothing changes here.
+* Once we have the "Foreign Sugar" opt-in mechanism in place we will generate context managers for such objects.
+
+## Swift
+
+* Existing `deinit` implementation is suitable so nothing changes here.
+
+## Kotlin
+
+* UniFFI will implement a mechanism that automically frees Rust references when the corresponding foreign object is destroyed.
```
bendk, line 152: I think one of the last issues is to update this statement so that it's clear that UniFFI would handle this behind-the-scenes and neither the library author or consumer would need to write any code to make that happen.  Does this new wording make that clear?

bendk:
Looks like we're all agreed on the approach.  Is there anything outstanding to discuss?  If not, I'll aim to merge this tomorrow.

jplatte:
Not something that should have stopped merging the ADR, but your [concern about the Kotlin implementation](https://github.com/mozilla/uniffi-rs/pull/1787#issuecomment-1788966499) seems like something that needs more investigation, in case you had forgotten about it again because nobody reacted.

bendk:
> Not something that should have stopped merging the ADR, but your [concern about the Kotlin implementation](https://github.com/mozilla/uniffi-rs/pull/1787#issuecomment-1788966499) seems like something that needs more investigation, in case you had forgotten about it again because nobody reacted.

I think that gets fixed once freeing the reference is tied to `PhantomQueue`, `Cleaner`, or some similar method.  Those won't free the reference while it's still reachable, which I think fixes the issue.

PR #1788, opened 2023-10-11
Merge override config with `uniffi.toml`

arg0d:
Instead of replacing `uniffi.toml` with override config entirely, merge the override config into `uniffi.toml`. The merge recursively upserts TOML keys.

This strategy seems pretty standard everywhere when it comes to configuration. Merging allows the configuration to be selectively changed, without having to duplicate the entire default configuration.

The main use case for this would be to allow "global" configuration across multiple crates in a library. Some configuration options should be the same for all crates in a library, without having to duplicate these options across all crates. Right now I can think of a few use cases of such configuration:

- In Swift generator, `omit_argument_labels`. I'm not entirely sure if anyone is actually using this, but it seems like an option that someone might potentially want to configure for all crates in a library.

- In `uniffi-bindgen-go`, `go_mod` configuration key is used to control the prefix for external type imports. Multiple crates in a library using external types share a common prefix when it comes to importing the external types, e.g. `com.example.uniffi.*`. This should configured as a single property for all crates.

- In `uniffi-bindgen-cs`, `visibility_modifier` configuration key is used to control the visibility modifier for "public" types. In C# visibility is applied to "assembly" level, not "file" level. So if bindings consumers wishes to package the bindings into his own "assembly" (library?), then "public" fields would be public for the consumers of the wrapper aswell https://github.com/NordSecurity/uniffi-bindgen-go/pull/13. This option would also benefit from a global config, since you would expect this value to remain the same for all crates in Rust library.

- External binding generators can use the merge option to inject additional configuration into `custom_types` fixture. This one is a bit sketchy, because external binding generators might want to modify the configuration for specific fixture/example crates in upstream, instead of doing global configuration for all crates. Still, global configuration solves the immediate issue of configuring `custom_types` crate to including `Url` custom type for external bindings.

### DIFF
--- a/uniffi/src/cli.rs
+++ b/uniffi/src/cli.rs
@@ -35,7 +35,9 @@ enum Commands {
         #[clap(long, short)]
         no_format: bool,
 
-        /// Path to the optional uniffi config file. If not provided, uniffi-bindgen will try to guess it from the UDL's file location.
+        /// Path to optional uniffi config file. This config will be merged on top of default
+        /// `uniffi.toml` config in crate root. The merge recursively upserts TOML keys into
+        /// the default config.
         #[clap(long, short)]
         config: Option<Utf8PathBuf>,
 
@@ -95,15 +97,17 @@ pub fn run_main() -> anyhow::Result<()> {
                 if lib_file.is_some() {
                     panic!("--lib-file is not compatible with --library.")
                 }
-                if config.is_some() {
-                    panic!("--config is not compatible with --library.  The config file(s) will be found automatically.")
-                }
                 let out_dir = out_dir.expect("--out-dir is required when using --library");
                 if language.is_empty() {
                     panic!("please specify at least one language with --language")
                 }
                 uniffi_bindgen::library_mode::generate_bindings(
-                    &source, crate_name, &language, &out_dir, !no_format,
+                    &source,
+                    crate_name,
+                    &language,
+                    config.as_deref(),
+                    &out_dir,
+                    !no_format,
                 )?;
             } else {
                 uniffi_bindgen::generate_bindings(
--- a/uniffi/src/cli.rs
+++ b/uniffi/src/cli.rs
@@ -35,9 +35,7 @@ enum Commands {
         #[clap(long, short)]
         no_format: bool,
 
-        /// Path to optional uniffi config file. This config will be merged on top of default
-        /// `uniffi.toml` config in crate root. The merge recursively upserts TOML keys into
-        /// the default config.
+        /// Path to optional uniffi config file. This config is merged with the `uniffi.toml` config present in each crate, with its values taking precedence.
         #[clap(long, short)]
         config: Option<Utf8PathBuf>,
 
--- a/uniffi_bindgen/src/lib.rs
+++ b/uniffi_bindgen/src/lib.rs
@@ -417,22 +417,53 @@ fn format_code_with_rustfmt(path: &Utf8Path) -> Result<()> {
     Ok(())
 }
 
+/// Load TOML from file if the file exists.
+fn load_toml_file(source: Option<&Utf8Path>) -> Result<Option<toml::value::Table>> {
+    if let Some(source) = source {
+        if source.exists() {
+            let contents =
+                fs::read_to_string(source).with_context(|| format!("read file: {:?}", source))?;
+            return Ok(Some(
+                toml::de::from_str(&contents)
+                    .with_context(|| format!("parse toml: {:?}", source))?,
+            ));
+        }
+    }
+
+    Ok(None)
+}
+
+/// Load the default `uniffi.toml` config, merge TOML trees with `config_file_override` if specified.
 fn load_initial_config<Config: DeserializeOwned>(
     crate_root: &Utf8Path,
     config_file_override: Option<&Utf8Path>,
 ) -> Result<Config> {
-    let path = match config_file_override {
-        Some(cfg) => Some(cfg.to_owned()),
-        None => crate_root.join("uniffi.toml").canonicalize_utf8().ok(),
-    };
-    let toml_config = match path {
-        Some(path) => {
-            let contents = fs::read_to_string(path).context("Failed to read config file")?;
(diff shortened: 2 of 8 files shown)


arg0d:
After writing the motivation for this change, it looks to be quite niche and I'm not sure anymore if its worth supporting this, at least with the current configuration options. Still, being able to selectively override configuration is pretty standard and it sounds like this could be useful in the future.

arg0d:
I can't think of a better way to handle `go_mod` and `visibility_modifier` configuration, so I think merging configs is a sane way to go. I think in the future this could pave way for adding more nuanced configuration options. https://github.com/NordSecurity/uniffi-bindgen-go/pull/13#issuecomment-1759462333

uniffi_bindgen/src/lib.rs
```diff
@@ -417,22 +417,53 @@ fn format_code_with_rustfmt(path: &Utf8Path) -> Result<()> {
...
+
+    let override_config = load_toml_file(config_file_override).context("override config")?;
+    if let Some(override_config) = override_config {
+        merge_toml(&mut config, override_config);
+    }
+
+    Ok(toml::Value::from(config).try_into()?)
+}
+
+fn merge_toml(a: &mut toml::value::Table, b: toml::value::Table) {
```
mhammond, line 453: I think we should have some tests here.

uniffi/src/cli.rs
```diff
@@ -35,7 +35,9 @@ enum Commands {
         #[clap(long, short)]
         no_format: bool,
 
-        /// Path to the optional uniffi config file. If not provided, uniffi-bindgen will try to guess it from the UDL's file location.
+        /// Path to optional uniffi config file. This config will be merged on top of default
```
mhammond, line 38: This docstring could probably be improved - it describes the implementation but not really the UI - ie, maybe it should be described in terms of being the default config or similar? It's also worth adding some docs to https://github.com/mozilla/uniffi-rs/blob/main/docs/manual/src/bindings.md.

mhammond (commented):
I think Mozilla could take advantage of this too - eg, 3 random uniffo.toml files in our repo:

https://github.com/mozilla/application-services/blob/8a7493212e0beacd4849d4c43cde9a30ca312f36/components/nimbus/uniffi.toml
https://github.com/mozilla/application-services/blob/8a7493212e0beacd4849d4c43cde9a30ca312f36/components/push/uniffi.toml
https://github.com/mozilla/application-services/blob/8a7493212e0beacd4849d4c43cde9a30ca312f36/components/sync_manager/uniffi.toml

Shows a fair bit of duplication, including things that really don't make sense to have in each component - eg, it doesn't really make sense for the sync_manager component to know it is in a cdylib named megazord, because that's just a decision made by the *consumer* of the component, not the component itself.

So all in all I think I'm +1 on this, although I might try and see how it works in practice for us. In the meantime I've a couple of suggestions...

bendk:
> The main use case for this would be to allow "global" configuration across multiple crates in a library. Some configuration options should be the same for all crates in a library, without having to duplicate these options across all crates. 

I think this would be great, but there's one thing I've never been able to figure out: How would this interact with external crates?  Suppose UniFFI popularity continues to increase and libraries like `serde`, `anyhow`, `log`, etc. add a feature to export UniFFI bindings.  Would the global configuration apply to them?  If not, how do you distinguish between the crates your project is bundling together and external crates?

mhammond:
> Suppose UniFFI popularity continues to increase and libraries like `serde`, `anyhow`, `log`, etc. add a feature to export UniFFI bindings. Would the global configuration apply to them? If so, is this a safe thing to do? If not, how do you distinguish between the crates your project is bundling together and external crates?

That's an interesting question! TBH, if that happened, I'd imagine it would only be for the Rust scaffolding - eg, I could imagine (say) them offering an `FfiConverter`, but can't picture them offering much in the way of Kotlin binding support. Thus, I doubt they'd actually rely on anything in uniffi.toml. But ultimately, even if they did offer binding support, I'd think that yes, we *would* want the global uniffi.toml to apply at least in some cases - eg, there's no possible way they could specify a meaningful `cdylib_name` and without a facility like offered in this PR, there'd be no reasonable way for us to supply it when building local bindings for that crate?

uniffi/src/cli.rs
arg0d: > This docstring could probably be improved - it describes the implementation but not really the UI - ie, maybe it should be described in terms of being the default config or similar? It's also worth adding some docs to https://github.com/mozilla/uniffi-rs/blob/main/docs/manual/src/bindings.md.

I will update the markdown fil, but I don't agree about the description. I would agree that its quite verbose, but its not wrong. This config lets you override default config values in crate config. I'm not really sure how else to describe this config.

arg0d: Reading your comment again, I see your point. Describing it in terms of being the default config sounds very terse. But this does not correctly map to the behaviour. It does not work like the default config. It works as an override to the config in each crate.

Now I'm thinking if its worth changing the implementation to make this the "default" config for all crates, instead of overriding.
- If the config acts as a default, then what you see in a crate's config config is what you get in the bindings. This sounds convenient when thinking from individual crates perspective. Setting `foo: foo` in crate config, I can be sure that this is what I will always get.
- If the config acts as an override, then what you see in the override config is what you get in all crate's bindings. This sounds convenient when thinking from bindings generation perspective. Setting `foo: bar` in override config, I can be sure that all crates will share the same value, regardless of their config.

I'm still leaning towards the override option. Thinking about a theoretical option `force_snake_case` for Kotlin (not sure why anyone would want that, but hey), I want a way to ensure that all my crates use the same configuration value, regardless of their crate configuration. I don't want to have to go and grep all crate config files, to ensure that none of the crates using this configuration key.

I've reworded the docstring, and updated documentation file. Let me know what you think.

uniffi_bindgen/src/lib.rs
arg0d: Added `test_merge_toml`

mhammond (approved):
This looks great to me, thanks.

PR #1790, opened 2023-10-11
Make From<UnexpectedUniFFICallbackError> optional

bendk:
This is in preparation of #1578 -- allowing foreign languages to implement traits.

One issue there is that if we want to allow foreign languages to implement traits, then any Result return types need to implement `LiftReturn`, and that impl currently has a bound on `E: From<UnexpectedUniFFICallbackError`.  IOW, it's only implemented for errors that can be converted from UnexpectedUniFFICallbackError.

This makes sense when the trait is actually used by the foreign languages, however some traits are only intended to be implemented by Rust code.  For that matter, some libraries may be fine panicking in the the face of unexpected callback errors.

By using the auto-ref specialization hack, we're able to convert the error if there's a `From<UnexpectedUniFFICallbackError>` impl, but use a panicking version if not.  The trait bound is no longer required to implement `LiftReturn`.

### DIFF
--- a/uniffi_macros/src/util.rs
+++ b/uniffi_macros/src/util.rs
@@ -216,11 +216,30 @@ pub(crate) fn tagged_impl_header(
     }
 }
 
-pub(crate) fn derive_ffi_traits(ty: &Ident, udl_mode: bool) -> TokenStream {
+pub(crate) fn derive_all_ffi_traits(ty: &Ident, udl_mode: bool) -> TokenStream {
     if udl_mode {
         quote! { ::uniffi::derive_ffi_traits!(local #ty); }
     } else {
         quote! { ::uniffi::derive_ffi_traits!(blanket #ty); }
+    }
+}
+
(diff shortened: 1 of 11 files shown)


uniffi_macros/src/util.rs
```diff
@@ -216,14 +216,33 @@ pub(crate) fn tagged_impl_header(
     }
 }
 
-pub(crate) fn derive_ffi_traits(ty: &Ident, udl_mode: bool) -> TokenStream {
+pub(crate) fn derive_all_ffi_traits(ty: &Ident, udl_mode: bool) -> TokenStream {
```
mhammond, line 219: These names kinda bother me a little - `derive_all_ffi_traits` seems to be deriving for a single ident where as `derive_ffi_traits` takes an array of identifiers. It would be great if you could maybe come up with alternatives to make this a little clearer?

mhammond (approved):

uniffi_macros/src/util.rs
mhammond: oh right, doh! Swapping those 2 args sounds great, thanks!

Issue #1730, opened 2023-09-05
Better system for handles


bendk:
There are many situations where we manage opaque FFI handles: callback interfaces, futures, callback data, etc.  The basic requirement is to create a handle for an object, pass it across the FFI, use the handle for FFI calls, then finally pass the handle to some function to free the object.

We generally use pointer/usize handles for this and have several ad-hoc systems to handle it.  There are several issues with these systems:

 - Pointers vary in size between platforms
 - Pointers require extra level of boxing when used with `Arc<dyn Trait>` (and currently [oneshot::Sender](https://github.com/faern/oneshot/pull/29))
 - Not all foreign languages support pointers directly (JS even has issues with 64-bit integers)
 - Managing the reference count in the foreign language is hard

I think we can come up with a general system for handles that avoids these issues.  The starting point can be the [slab](https://github.com/tokio-rs/slab/blob/master/src/lib.rs) crate, which is pretty simple and essentially creates `usize` handles to manage object allocations.  I think we can update it to use `u32` handles and catch use-after-free errors in some case (maybe most).

PR #1791, opened 2023-10-12
Foreign trait interfaces

bendk:
This PR enables foreign code to implement trait interfaces.  The code seems to be working for me, but there are some issues that should be resolved before this is merged:

* It's based on #1790, we should approve that one first
* As mentioned in the last commit, the way we handle the `FfiType` is inefficient and also feels too risky to me.  I think we should implement #1730 to fix this, but we could also just try to clean it up with the current system.
* We need to come up with a system for freeing the Rust pointers on Kotlin (#1787).

I also think the example could be improved some.


bendk:
Looks like this one is maybe ready to merge.  The code isn't ideal, but it's working.  The only way it's worse than the current code is when when a Rust trait object is passed back and forth across the FFI, in which case it will be less efficient.  We just released `0.25` and I am confident we can have a fix for that before the next release.

mhammond (approved):
This is fantastic! @bendk have been talking about and working towards this for well over a year and I think it opens a whole new world of possibilities for UniFFI - thanks Ben 🎉 🚢

PR #1794, opened 2023-10-18
Add an example of and better docs for foreign traits.

mhammond:
The main point of that is to somewhat de-emphasize callbacks and put them on an even keel with foreign traits. It doesn't go so far as to deprecate callbacks, but I think ultimately that's where we are likely to end up.

### DIFF
--- a/docs/manual/src/foreign_traits.md
+++ b/docs/manual/src/foreign_traits.md
@@ -1,0 +1,146 @@
+# Foreign traits
+
+UniFFI traits can be implemented by foreign code.
+This means traits implemented in Python/Swift/Kotlin etc can provide Rust code with capabilities not easily implemented in Rust, such as:
+
+ * device APIs not directly available to Rust.
+ * provide glue to clip together Rust components at runtime.
+ * access shared resources and assets bundled with the app.
+
+# Example
+
+To implement a Rust trait in a foreign language, you might:
+
+## 1. Define a Rust trait
+
+This toy example defines a way of Rust accessing a key-value store exposed
+by the host operating system (e.g. the key chain).
+
+All methods of the Rust trait should return a `Result<>` with the error half being
+a [compatible error type](./udl/errors.md) - see below for more on error handling.
+
+For example:
+
+```rust,no_run
+pub trait Keychain: Send + Sync + Debug {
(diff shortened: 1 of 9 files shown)


docs/manual/src/foreign_traits.md
```diff
@@ -0,0 +1,143 @@
...
+For example:
+
+```rust,no_run
+pub trait Keychain: Send + Sync + Debug {
+  fn get(&self, key: String) -> Result<Option<String>, KeyChainError>;
+  fn put(&self, key: String, value: String) -> Result<(), KeyChainError>;
+}
+```
+
+If you didn't add `#[uniffi::export]` above the trait, you will need in your UDL:
```
badboy, line 31: What about:
```suggestion
If you are using macros add `#[uniffi::export]` above the trait.
Otherwise define this trait in your UDL file:
```

docs/manual/src/foreign_traits.md
```diff
@@ -0,0 +1,143 @@
...
+  pub fn login(&self) {
+    let username = self.keychain.get("username".into());
+    let password = self.keychain.get("password".into());
+  }
+}
+```
+
+## 3. Create a foreign language implementation of the trait
+
+Here's a Kotlin implementation.
```
badboy, line 76:
```suggestion
Here's a Kotlin implementation:
```

badboy (approved):

docs/manual/src/foreign_traits.md
```diff
@@ -0,0 +1,144 @@
...
+# Error handling
+
+We must handle foreign code failing, so all methods of the Rust trait should return a `Result<>` with a [compatible error type](./udl/errors.md) otherwise these errors will panic.
+
+## Unexpected Error handling.
+
+So long as your function returns a `Result<>`, it's possible for you to define how "unexpected" errors
+(ie, errors not directly covered by your `Result<>` type, panics, etc) are converted to your `Result<>`'s `Err`.
+
+If your code defines a `From<uniffi::UnexpectedUniFFICallbackError>` impl for your error type, then those errors will be converted into your error type which will be returned to the Rust caller instead of panicing.
```
bendk, line 141: This doc looks great, but maybe call out the panic slightly more.  Something like: "...your error type which will be returned to the Rust caller.  If they do not, then the generated code  will panic."

Issue #1798, opened 2023-10-23
Type error in Kotlin generated code when using external types in async functions in 0.25.0


dani-garcia:
Not sure what information can help with this, or the exact cause of the error, but I have some reproduction project here: 
https://github.com/dani-garcia/uniffi_025_repro

The project is structured as such:
- One inner crate, that exposes a uniffi library using proc-macros and defines a external type conversion, in this case `chrono::DateTime`
- One outer crate, that also exposes a uniffi library using proc-macros, imports `inner` and just reexports it's functions and external types (with `ffi_converter_forward`).

Note that the only function that these two crates expose is `get_time`.

With both functions defined as sync, the generated code works fine, but when we change the function in `outer` to `async fn`, we get the following error in the generated kotlin code:

```
Type mismatch: inferred type is com.example.outer.RustBuffer.ByValue but com.example.inner.RustBuffer.ByValue was expected
```

Code generated for the function without `async`:
```kotlin
    override fun `getTime`(): DateTime =
        callWithPointer {
            rustCall() { _status ->
                _UniFFILib.INSTANCE.uniffi_outer_fn_method_outerclient_get_time(
                    it,

                    _status
                )
            }
        }.let {
            FfiConverterTypeDateTime.lift(it)
        }
```

Code generated for the function with `async`:
```kotlin
    override suspend fun `getTime`() : DateTime {
        return uniffiRustCallAsync(
            callWithPointer { thisPtr ->
                _UniFFILib.INSTANCE.uniffi_outer_fn_method_outerclient_get_time(
                    thisPtr,
                    
                )
            },
            { future, continuation -> _UniFFILib.INSTANCE.ffi_outer_rust_future_poll_rust_buffer(future, continuation) },
            { future, status -> _UniFFILib.INSTANCE.ffi_outer_rust_future_complete_rust_buffer(future, status) },
            { future -> _UniFFILib.INSTANCE.ffi_outer_rust_future_free_rust_buffer(future) },
            // lift function
            { FfiConverterTypeDateTime.lift(it) }, // <------- Here is where it's failing
            // Error FFI converter
            NullCallStatusErrorHandler,
        )
    }
```

The function in outer just calls the inner function:
```rust
use std::sync::Arc;
use chrono::Utc;
uniffi::setup_scaffolding!();

#[derive(uniffi::Object)]
pub struct OuterClient();

#[uniffi::export]
impl OuterClient {
    #[uniffi::constructor]
[...]

PR #1803, opened 2023-10-25
Fixed a couple issues with async code and multiple crates

bendk:
* Kotlin: fixed external types with async functions (#1798)
* Swift: fixed compile error when multiple crates define async functions


mhammond (approved):

bendk:
I believe there's still an issue with this on Kotlin, and maybe Pythons other than CPython: https://github.com/mozilla/uniffi-rs/issues/1809.  I think this can get fixed soon and should be part of `0.25.1`.

PR #1810, opened 2023-10-27
Fix for multiple crates defining async functions

bendk:
A single global callback won't work if there are multiple foreign modules.  Instead, input the callback in `poll()`.

Added some tests for this in `fixtures/ext-types/proc-macro-lib`. Things worked on Swift and Python for me because that code was based on leaking a pointer and it happened to work if the pointer came from another module.  It just halted on Kotlin though and I think the same would happen if I wasn't on CPython and the alternate PointerMangager was used.


skhamis (approved):
The test and updated code LGTM!

PR #1811, opened 2023-10-27
Fix doc generation - they are now generated in docs/manual/html

mhammond:
Docs used to be generated directly in `docs/manual/book`. Now that there's a new "target" for link-checking we have `docs/manual/book/html` and `docs/manual/book/linkcheck`

(at least I think that's what's going on :)


bendk (approved):
Looks good to me, let's see if this fixes things.

PR #1812, opened 2023-10-27
Backporting fix for trait methods that throw exceptions

bendk:
This is part of #1791, which was merged just after the `0.25.0` release was made.  We wanted to give that code a bit of time in main before releasing it.  However, this part is useful for traits implemented in Rust.


mhammond (approved):
rs=me!

PR #1814, opened 2023-10-28
Adding `FfiType::Callback`

bendk:
This is sort-of in the same sprit as #1808, but just focused on foreign callback functions.  The goal is to have a single `FfiType` that we can reuse whenever we want to add a new callback function type.

With this change, if you want to add a new callback function type you just add an item in `ffi_callback_definitions()`.  I think this should make the process much easier.  It  can help avoid FFI type mismatches, like how I made the Kotlin signature input a i16 instead of an i8.

Use the new type for the future continuation callback.  I have another plan for callback interface callbacks.  I didn't touch the foreign executor callback, I hope to work on that very soon.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
@@ -274,14 +274,19 @@ impl KotlinCodeOracle {
         }
     }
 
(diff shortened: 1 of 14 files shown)


bendk:
I thought I was going to need to get some more PRs merged before moving to the second step of this, but I realized I was wrong.  I just put up #1818 which includes this change and more.  Let's discuss in that PR, I don't think there's much reason to merge half of the changes without the other.

PR #1815, opened 2023-10-28
[Kotlin] Generate open classes and methods

jmartinesp:
This PR is an alternative implementation of #1782 , with the 1st approach @bendk proposed [here](https://github.com/mozilla/uniffi-rs/pull/1782#issuecomment-1771432020):

> We could make the concrete type open. It's slightly risky because if you don't have an actual handle to a Rust object, then calling a UniFFI function will fail. But I would be okay with saying that users accept that risk if they extend the generated class. For your use case of only creating the objects if the native library isn't available, it seems like it should work fine.

The implementation changes are:

- `FFIObject.pointer` is now nullable. I know it's not great, but it's necessary as far as I can tell and I haven't seen any issues with it so far.
- `FFIObject` now has an alternate `constructor(noPointer: NoPointer)`. I tried using an empty constructor or one with `pointer: Pointer?` as the only parameter, but it clashed with the code in generated classes.
- `FFIObject.freeRustArcPtr` implementation in the different classes now does a null check before trying to free the pointer, so you don't need to override it in the fakes, and methods using `callWithPointer` will unwrap the nullable value and throw a `NullPointerException` if it doesn't exist.
- I removed the interfaces for concrete types because they don't seem to have any use, since you can't implement some `FooInterface` and just pass it to a `fn bar(foo: Foo)` function without a `ClassCastException`, and the open class can now act as a kind of interface (again, I might be wrong about this, they could be restored). I did keep the interface for the trait interfaces and their impls.

I added a few tests displaying how these open classes can be extended, used and **not used** in a real project.

I feel like this approach isn't as correct as the one in #1782 where you'd be forced to provide an alternative implementation for the interfaces that doesn't crash, but it's way simpler to maintain and the drawbacks I see as a lib user aren't that bad.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
@@ -7,7 +7,15 @@ open class {{ impl_class_name }} : FFIObject{% if obj.is_trait_interface() %}, {
 
     constructor(pointer: Pointer): super(pointer)
 
-    protected constructor(noPointer: NoPointer): super(noPointer)
+    /**
+     * This constructor can be used to instantiate a fake object.
+     *
+     * **WARNING: Any object instantiated with this constructor cannot be passed to an actual Rust-backed object.**
+     * Since there isn't a backing [Pointer] the FFI lower functions will crash.
+     * @param noPointer Placeholder value so we can have a constructor separate from the default empty one that may be
+     *   implemented for classes extending [FFIObject].
+     */
+    constructor(noPointer: NoPointer): super(noPointer)
 
     {%- match obj.primary_constructor() %}
     {%- when Some with (cons) %}
--- a/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
@@ -3,7 +3,9 @@
 {%- let (interface_name, impl_class_name) = obj|object_names %}
 {%- let methods = obj.methods() %}
 
-open class {{ impl_class_name }} : FFIObject{% if obj.is_trait_interface() %}, {{ interface_name }} {% endif %} {
+{% include "Interface.kt" %}
+
+open class {{ impl_class_name }} : FFIObject, {{ interface_name }} {
 
     constructor(pointer: Pointer): super(pointer)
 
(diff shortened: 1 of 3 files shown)


jmartinesp:
Hmm... I see this could clash with #1808. Maybe the `UniffiHandleWrapper` could be an enum with `Some/None` cases instead of the nullable `Pointer` used here.

bendk:
> Hmm... I see this could clash with #1808. Maybe the `UniffiHandleWrapper` could be an enum with `Some/None` cases instead of the nullable `Pointer` used here.

I haven't checked this code, but you can use `-1` to represent a `null` handle.  In general, the negative numbers are reserved for special cases.

uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -3,11 +3,19 @@
 {%- let (interface_name, impl_class_name) = obj|object_names %}
 {%- let methods = obj.methods() %}
 
-{% include "Interface.kt" %}
```
bendk, line 6: I'm not sure, but I believe the interface definition enables delegation and some users are using this.  Can we keep it in?

uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -3,11 +3,19 @@
...
-) : FFIObject(pointer), {{ interface_name }}{
+    /**
+     * This constructor can be used to instantiate a fake object.
+     *
+     * **WARNING: Any object instantiated with this constructor cannot be passed to an actual Rust-backed object.**
+     * Since there isn't a backing [Pointer] the FFI lower functions will crash.
+     * @param noPointer Placeholder value so we can have a constructor separate from the default empty one that may be
+     *   implemented for classes extending [FFIObject].
+     */
+    constructor(noPointer: NoPointer): super(noPointer)
```
bendk, line 18: Great docstring and I like the use of the singleton object to differentiate this from the regular constructor.

bendk (changes requested):
Looks great overall, but can you add back in the interface definition?  I think that's still useful.

uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
jmartinesp: Ah, I didn't realise that was an option. We don't rely on delegation on our project as we try to destroy and map any Rust backed object as soon as possible, but it could be useful to add an intermediate wrapper, even though you'd still need to pass Rust backed objects to any functions that require them, although you could try to minimise these by tweaking the binding API.

I'll add back the interfaces then.

bendk (approved):
Looks great, thanks for the contribution!

PR #1817, opened 2023-10-30
Rust traits `Display`, `Hash` and `Eq` exposed to Kotlin and Swift.

mhammond:
This brings those language to parity with Python.

Note that Kotin is using `Display` for `toString()` and there's no obvious way to expose `Debug`, so it isn't. Swift does seem to have a sensible way of exposing both `Debug` and `Display`.

Neither language seems to need `ne` so that's unused.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
@@ -4,7 +4,21 @@
 
 {% include "Protocol.swift" %}
 
-public class {{ impl_class_name }}: {{ protocol_name }} {
+public class {{ impl_class_name }}:
+    {%- for tm in obj.uniffi_traits() %}
+    {%-     match tm %}
+    {%-         when UniffiTrait::Display { fmt } %}
+    CustomStringConvertible,
+    {%-         when UniffiTrait::Debug { fmt } %}
+    CustomDebugStringConvertible,
+    {%-         when UniffiTrait::Eq { eq, ne } %}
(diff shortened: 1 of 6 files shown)


badboy:
> Swift does seem to have a sensible way of exposing both `Debug` and `Display`.

What do you mean by this?
From the code you don't seem to special-case anything there and you use both `CustomDebugStringConvertible` and `CustomStringConvertible`, so I'm a bit confused by your description here.

mhammond:
Sorry for the confusion: I was trying to say that Kotlin is able to get `Display` output but unable to get `Debug`, whereas Swift, like Python (`__str__` vs `__repr__`) is able to get both.

badboy (approved):

uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
```diff
@@ -88,6 +102,38 @@ public class {{ impl_class_name }}: {{ protocol_name }} {
...
+    {%-         when UniffiTrait::Hash { hash } %}
+    public func hash(into hasher: inout Hasher) {
+        let val = {% call swift::try(hash) %} {{ hash.return_type().unwrap()|lift_fn }}(
+            {% call swift::to_ffi_call_with_prefix("self.pointer", hash) %}
+        )
+        hasher.combine(val)
+    }
+    {%-         else %}
+    {%-    endmatch %}
+    {%- endfor %}
```
jhugman, line 135: Nice!

jhugman (approved):
LGTM!

PR #1820, opened 2023-10-30
Split up the rustfuture module

bendk:
That module was getting pretty long and I want to add some more functionality to it.  Let's split it up into smaller files.

This commit just moves things around and renames `ContinuationDataCell` to `Scheduler`.


uniffi_core/src/ffi/rustfuture/scheduler.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    pub(super) fn is_cancelled(&self) -> bool {
+        matches!(self, Self::Cancelled)
+    }
+}
+
+// ContinuationDataCell is Send + Sync as long we handle the *const () pointer correctly
+
+unsafe impl Send for Scheduler {}
+unsafe impl Sync for Scheduler {}
```
jplatte, line 94: I think at least for `Send` the comment is wrong (and `Sync` would be trivially obtained by wrapping the `*const ()` in `sync_wrapper::SyncWrapper` which would not restrict any of the uses here).

For `Send` to hold for regular `&reference`s, the type being referenced needs to be `Sync` so I think this also needs to hold for any data we store in this pointer from any backend, and that should be documented (here, but also in any docs we have for backend authors).

bendk: Good point.  I just updated that comment.  I'm also going to do a general rework on handles in another PR, I'll make sure to mention something like this there.  In general, I think whenever either side passes a handle over the FFI, it must be `Send` and the handled object must by `Sync`.

jplatte (approved):

mhammond (approved):

PR #1821, opened 2023-10-31
Fully qualify Send and Sync trait in generated code

dani-garcia:
Other standard library types were already using a fully qualified name, but that wasn't the case for the `Send` and `Sync` bounds, which will cause issues in any code that has types named `Send` or `Sync` and is using the `uniffi` proc macros.


jplatte (approved):
Thanks!

mhammond:
Not sure what's up with CI, but this can't have broken it!

dani-garcia:
Strange, tried rebasing on master and force pushing in case it was a temporary fluke with CI, but it seems to be stuck there.

PR #1823, opened 2023-11-01
Handles

### LINKED ISSUE #1797: Object.lower() can result in use-after-free errors

bendk:
This PR contains the parts of #1808 that I think are most beneficial and least controversial:

- **Consolidate the handle code.**  Use one `FfiType` variant for all handles.  Implement one system
  in Rust and one system in each foreign language rather than multiple systems.  This makes it
  easier to add new object/handle types.
- **Handles should be clonable.** This is needed to fix #1797.
- **Rust and foreign handles should be distinguishable.**  This will make the trait interface implementation more efficient.

This PR sets things up to fix #1797, but doesn't actually fix it.  That's going to take some more code and I want to do that in a separate PR.
 
It also doesn't implement the trait interface changes, because when I did that I hit #1797.  In the [traits example](https://github.com/bendk/uniffi-rs/blob/484fc4526144113e38ff2abe48f26d1246c8dab5/examples/traits/tests/bindings/test_traits.py#L15), we pass a trait interface into Rust, then Rust immediately returns it. This is what happens Rust returns it:

* It copies the handle in `lower()`
* It drops the struct, invoking the `IDX_CALLBACK_FREE` method.
* It then returns the handle back to the foreign language, which is use-after-free.


tarikeshaq (commented):
Apologies this fell through my review queue, I'll redirect to @mhammond as he's been more recently involved and I'd most likely not provide the best review without playing catchup

uniffi_core/src/ffi/handle.rs
```diff
@@ -0,0 +1,53 @@
...
+///
+/// Rust handles are generated by leaking a raw pointer
+/// Foreign handles are generated with a handle map that only generates odd values.
+/// For all currently supported architectures and hopefully any ones we add in the future:
+/// * 0 is an invalid value.
+/// * The lowest bit will always be set for foreign handles and never set for Rust ones (since the
+///   leaked pointer will be aligned).
+///
+/// Foreign handles for internal bindings languages look like this:
+///   * Bits 0-48 are the map key.  Keys are always odd, so the lowest bit is always set.
```
arg0d, line 18: Nice one, I was thinking it would be MSB to check range `[0; u64::max/2)`, but checking LSB seems less error prone. Also LSB is friendly with map identifier bits.

bendk:
This is really old, some of it already has been implemented like #1797.  Let's close this one, I'm going to try to open a new version.

PR #1825, opened 2023-11-01
Replaced CancellationError with fatalError

### LINKED ISSUE #1824: 'CancellationError' is only available in macOS 10.15 or newer
In https://github.com/mozilla/uniffi-rs/pull/1767, `CancellationError` was added to [Helpers.swift](https://github.com/mozilla/uniffi-rs/blob/39fc189c2e4b2441efd0198e2e99ec894c8332d0/uniffi_bindgen/src/bindings/swift/templates/Helpers.swift#L68-L100).

When using this for macOS swift bindings, [CancellationError](https://developer.apple.com/documentation/swift/cancellationerror/) is only available for macOS 15 (macOS 14 came out ~5 weeks ago).

I could add some test code but I think it might be clear?

simlay:
Closes #1824.

This fix was discussed in briefly in matrix. https://github.com/mozilla/uniffi-rs/pull/1768 has some related work.

I've tested this work fix in https://github.com/liveview-native/liveview-native-core/pull/43

Is there a CI check I should add to prevent this in the future?

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
@@ -93,7 +93,7 @@ private func uniffiCheckCallStatus(
             }
 
         case CALL_CANCELLED:
-                throw CancellationError()
+                throw fatalError("Cancellation not supported yet")
 
         default:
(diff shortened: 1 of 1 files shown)


bendk (approved):
Looks good to me, thanks for putting up this PR.

PR #1829, opened 2023-11-02
Removing redundant `throw` in `Helpers.swift`

bendk:
`fatalError()` just causes a crash, there's no need to throw it.  This removes some warnings when building the swift code.


simlay (commented):
My bad. Looking at my exact example, I see the warning mentioned and should have dealt with it in #1825.

badboy (approved):

PR #1833, opened 2023-11-06
procmacro implemented types can be used in UDL.

mhammond:
People who want to slowly move away from UDL, or wish to stick with UDL but use some features available only to procmacros are a little stuck as currently it's impossible for UDL to reference types defined by procmacros in this crate. This PR makes that possible.


bendk (approved):
LGTM.

It seems like the only question is how the attribute works, and I think this solution makes sense.  It's slightly different than external types which uses `ExternalInterface`, but I like this system better.

PR #1834, opened 2023-11-06
Adding UDL async support

### LINKED ISSUE #1716: UDL support for async methods
The User's Guide only tells how to define async functions when using attributes. How can I define async functions when using the .udl file?

bendk:
If I realized it was this easy, I would have done this before.  `uniffi_udl` is great. Fixes #1716.


mhammond (approved):
great!

PR #1836, opened 2023-11-07
Add support for immutable records in Kotlin and Swift

dani-garcia:
This isn't so much a finalised PR but more of a proposal and request for comments on how to enable immutable records in Kotlin and Swift. There's some previous discussion on why the records were changed to mutable by default in #471.

I agree that allowing mutability by default is the most flexible option, but it can cause issues with certain frameworks like Jetpack Compose in Kotlin, for example:

```kotlin
data class Sample(var num1: Int)

class SampleViewModel: ViewModel() {
   // Create a state object. When used in a UI widget, any changes to sampleData will be automatically reflected in the UI
   var sampleData by mutableStateOf(Sample(num1 = 0))
      private set

   fun onChange(newNum: Int) {
      // This works, by copying the record and reassigning the reference, changes to sampleData are detected
      sampleData = sampleData.copy(num1 = newNum)
      // This won't work, sampleData has the same reference as before and mutableStateOf will not detect any changes
      sampleData.num1 = newNum
   }
}
```

It's because of quirks like this that immutable data classes are preferred by default in Kotlin. By using them, we make this type of mistakes impossible. When we want to change a value in a data class, Kotlin provides the very useful `.copy()` method to make copies of records while updating some of their properties, so the need to mutate a record in place is almost entirely negated.

As far as I know this change would be Kotlin and Swift specific, so adding a way of tagging records as immutable in UDL/proc-macros, even though it would be considerably more flexible, seems to me like a little too much effort. Instead, I'm proposing an option in `uniffi.toml` for the two languages to always generate immutable records. The changes to the code base this way would be minimal, and at least for my use case (Kotlin+Compose, Swift+SwitUI), more than enough to solve the issue mentioned.

I'd appreciate your thoughts on this topic, and let me know if this is a direction you'd want to go in, or if you have another alternative in mind.

### DIFF
--- a/docs/manual/src/kotlin/configuration.md
+++ b/docs/manual/src/kotlin/configuration.md
@@ -8,6 +8,7 @@ The generated Kotlin modules can be configured using a `uniffi.toml` configurati
 | ------------------ | -------  |------------ |
 | `package_name`     |  `uniffi` | The Kotlin package name - ie, the value used in the `package` statement at the top of generated files. |
 | `cdylib_name`      | `uniffi_{namespace}`[^1] | The name of the compiled Rust library containing the FFI implementation (not needed when using `generate --library`). |
+| `generate_immutable_records` | `false` | Whether to generate records with immutable fields (`val` instead of `var`). |
 | `custom_types`      | | A map which controls how custom types are exposed to Kotlin. See the [custom types section of the manual](../udl/custom_types.md#custom-types-in-the-bindings-code)|
 | `external_packages` | | A map of packages to be used for the specified external crates. The key is the Rust crate name, the value is the Kotlin package which will be used referring to types in that crate. See the [external types section of the manual](../udl/ext_types_external.md#kotlin)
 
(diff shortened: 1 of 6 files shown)


dani-garcia:
At least for our Kotlin developers, I know they prefer using immutable records everywhere and just doing `new_record = old_record.copy(property = new_value)` whenever they need to update a value. I don't see such a strong opinion from the Swift devs, but I figured it was worth adding the option for symmetry's sake.

I wouldn't mind having a way to select this per-type, but it's a bit harder to integrate that nicely, maybe just having a `mutable_records` map as an option in the `toml`? Something like this:
```toml
[bindings.kotlin.mutable_records]
"*" = false                       # By default we want all records immutable
"MyRecordThatNeedsToBeMut" = true # This one specifically needs to be mutable
```

Any thoughts?

dani-garcia:
Also sidenote because I've never seen that CI error before:
```
CircleCI Pipeline Could not find a usable config.yml, you may have revoked the CircleCI OAuth app.
```

Seems I had a previous personal account on CircleCI using Github SSO, but revoked its permission years ago, I tried to login again on CircleCI and give it permissions, and that seems to have solved the issue, in case it happens to anyone else :)

mhammond:
This seems fine to me and the fact it's specified in a config means it doesn't really risk breaking anyone. But could you please document this [here](https://github.com/mozilla/uniffi-rs/blob/main/docs/manual/src/kotlin/configuration.md)

mhammond (approved):

PR #1837, opened 2023-11-09
Blocking task queue

bendk:
This adds support for running blocking tasks in an async function, using a foreign-provided dispatcher.  For example, running SQLite queries inside a `CoroutineContext` / `DispatchQueue` / `futures.Executor`.  I've tried to implement this a couple ways, I like this one the best by far because it allows the Rust closure to reference values in the outer scope (it works like  tokio's `run_in_place()` rather than `spawn_blocking()`).

This removes the old `ForeignExecutor` code which tried to do this, but in a worse way.

This depends on #1820 and #1823, we need to agree on those before merging this one.


docs/manual/src/futures.md
```diff
@@ -46,4 +46,62 @@ In Rust `Future` terminology this means the foreign bindings supply the "executo
...
+This assumption, combined with cooperative scheduling, allows for a large number of futures to be handled by a small number of threads.
+Foreign executors make similar assumptions and sometimes more extreme ones.
+For example, the Python eventloop is single threaded -- if any task spends a long time between `await` points, then it will block all other tasks from progressing.
+
+This raises the question of how async code can interact with blocking code.
+"blocking" here means code that preforms blocking IO, long-running computations without `await` breaks, etc.
+To support this, UniFFI defines the `BlockingTaskQueue` type, which is a foreign object that schedules work on a thread where it's okay to block.
+
+On Rust, `BlockingTaskQueue` is a UniFFI type that can safely run blocking code.
+It's `run_blocking` method works like tokio's [block_in_place](https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html) function.
```
bendk, line 61: I almost want to name this `block_in_place` to match tokio's name.  The reason I didn't is the "in_place" part doesn't make as much sense when there's not a `spawn_blocking()` method to contrast with.

uniffi_core/src/ffi/rustfuture/scheduler.rs
```diff
@@ -2,11 +2,72 @@
...
+/// will be ready.
+///
+/// There's one tricky issue here: how can we ensure that when the top-level task is run in the
+/// blocking task queue, this future will be polled?  What happens this future is a child of `join!`,
+/// `FuturesUnordered` or some other Future that handles its own polling?
+///
+/// We start with an assumption: if we notify the waker then this future will be polled when the
+/// top-level task is polled next.  If a future does not honor this then we consider it a broken
+/// future.  This seems fair, since that future would almost certainly break a lot of other future
+/// code.
```
bendk, line 42: I'm pretty sure this is a good assumption, but I'd love a double check on this one.

When I mention "other future code", I was thinking of something like a lock.  If your async lock type locks a mutex and notifies the waker, then any intermediate futures must poll your type or else you're in a bad state.  Unless I'm missing something, async lock authors just have assume they will get polled and we can too.

docs/manual/src/futures.md
```diff
@@ -45,3 +45,63 @@ This code uses `asyncio` to drive the future to completion, while our exposed fu
...
+Foreign executors make similar assumptions and sometimes more extreme ones.
+For example, the Python eventloop is single threaded -- if any task spends a long time between `await` points, then it will block all other tasks from progressing.
+
+This raises the question of how async code can interact with blocking code that performs blocking IO, long-running computations without `await` breaks, etc.
+UniFFI defines the `BlockingTaskQueue` type, which is a foreign object that schedules work on a thread where it's okay to block.
+
+On Rust, `BlockingTaskQueue` is a UniFFI type that can safely run blocking code.
+It's `execute` method works like tokio's [block_in_place](https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html) function.
+It inputs a closure and runs it in the `BlockingTaskQueue`.
+This closure can reference the outside scope (i.e. it does not need to be `'static`).
```
bendk, line 62: I'm not sure what this method should be named.  I chose `execute` simply because I hadn't seen it used in other places so it didn't have the baggage of other names.  I almost want to name in `block_in_place` since it matches the semantics of the tokio method.  But the "in_place" part feels weird since we don't have a `spawn_blocking` method to contrast with.

uniffi_core/src/ffi/rustfuture/mod.rs
```diff
@@ -28,10 +31,23 @@ pub enum RustFuturePoll {
 ///
 /// The Rust side of things calls this when the foreign side should call [rust_future_poll] again
 /// to continue progress on the future.
-pub type RustFutureContinuationCallback = extern "C" fn(callback_data: *const (), RustFuturePoll);
+///
+/// WARNING: the call to [rust_future_poll] must be scheduled to happen soon after the callback is
+/// called, but not inside the callback itself.  If [rust_future_poll] is called inside the
+/// callback, some futures will deadlock and our scheduler code might as well.
```
bendk, line 37: This isn't really related to the PR, but it's something I noticed when trying to refactor some things.

bendk:
Updated this code to not depend on the handles work, because I didn't want that to block this.  This means the foreign handle-map code is a bit of a mess, but #1823 should clean that up.

At the same time, I did make it depend on the [ffi-vtables branch](https://github.com/mozilla/uniffi-rs/pull/1818) because that code was extremely useful for implementing `Drop` for `BlockingTaskQueue`  (I also used it to implement `Clone`).  It would be possible to land this without those changes, but I think it would be much more complicated.

tarikeshaq:
Removing myself from the review, will also remove myself from auto-assign - if I take more UniFFI work in the future I'll add myself back

bendk:
Rebased this one now that the vtable code has been merged.  It's looking pretty good now so I removed the `do-not-merge` tag.

bendk:
Removed this one from the list of PRs for `0.27` since it seems too early to tell how we should handle it.  See #2040 for a discussion.

PR #1842, opened 2023-11-13
Fix regression in Kotlin error/exception names.

### LINKED ISSUE #1828: v0.25 breaks generation of nested error variants in Kotlin bindings
There is a problem when we use `[Error]` enum as a variant value for another `[Error]` one for example.

UDL code:
```
[Error]
interface Sub1Error {
    One(string error);
    Two(string error);
};

[Error]
interface Sub2Error {
    One(string error);
    Two(string error);
};

[Error]
interface RootError {
    Sub1(Sub1Error error);
    Sub2(Sub2Error error);
};
```

Rust code:
```rust
#[derive(Debug, thiserror::Error)]
pub enum RootError {
    #[error(transparent)]
    Sub1 {
        #[from]
        error: Sub1Error,
    },
    #[error(transparent)]
    Sub2 {
        #[from]
        error: Sub2Error,
    },
}

#[derive(Debug, thiserror::Error)]
pub enum Sub1Error {
    #[error("{error}")]
    One { error: String },
    #[error("{error}")]
    Two { error: String },
}

#[derive(Debug, thiserror::Error)]
pub enum Sub2Error {
    #[error("{error}")]
    One { error: String },
    #[error("{error}")]
    Two { error: String },
}
```

Error:
```
uniffi-rs/target/tmp/uniffi-example-traits-74e2e027c7e1d172/uniffi/traits/traits.kt:1004:22: error: unresolved reference: Sub1Error
        val `error`: Sub1Error
                     ^
uniffi-rs/target/tmp/uniffi-example-traits-74e2e027c7e1d172/uniffi/traits/traits.kt:1011:22: error: unresolved reference: Sub2Error
        val `error`: Sub2Error
```

The example above works with versions 0.23 and 0.24 and earlier I think, but beginning with this change (https://github.com/mozilla/uniffi-rs/commit/5e3dea51f17ae59a695a40e23479d57262968bb6#diff-522f0ef157bc0a32453bafd790a9058a35cbb8a721df58a9b51b1ec6b74db5caL456) or at least in version 0.25 it doesn't.

Similar to https://github.com/mozilla/uniffi-rs/issues/1020 and https://github.com/mozilla/uniffi-rs/issues/1062

mhammond:
Fixes #1828.

I'm frustrated with myself that I left this so broken - I think I was trying to avoid this larger Kotlin refactor that is sadly necessary. I ended up forking `CodeType` as only Kotlin needs error renaming.


uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -401,43 +440,41 @@ pub mod filters {
     use super::*;
     pub use crate::backend::filters::*;
 
-    pub fn type_name(as_ct: &impl AsCodeType) -> Result<String, askama::Error> {
+    pub(super) fn type_name(as_ct: &impl AsCodeType) -> Result<String, askama::Error> {
```
bendk, line 443: It's not a big deal either way, but I'd lean towards keeping these `pub`.

For one, it'll make the PR smaller, but I also generally prefer that functions are either public or private.  If we don't want outside code to access these methods, then we can make `gen_python` a private module.  If we need more complicated access control, we can selectively import its public functions into the parent module.

bendk (approved):
Looks good to me.  I had one nit about the module visibility, but I think that can go either way.

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
mhammond: Ah yeah - this was needed before I "unforked" otherwise I got errors about leaking private types. Note `AsCodeType` is again pub I can remove this for python!

mhammond: oops, I was wrong - I unforked that trait, but Python is still using the fork. Without this I get:

```
385 | trait AsCodeType {
    | ---------------- `gen_python::AsCodeType` declared as private
...
443 |     pub fn type_name(as_ct: &impl AsCodeType) -> Result<String, askama::Error> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak private trait
```

(which TBH I don't quite understand seeing as the filters module, while nominally `pub` isn't actually exported anywhere.)

mhammond: (and trying to make the trait itself pub then confuses even more things)

bendk: Ahh I see, then `pub(super)` makes sense to me.

kp-timo-beyel:
This is still broken in 0.25.2 if the nested error is declared in an external crate. 

`#[derive(uniffi::Error, thiserror::Error, PartialEq)]
pub enum TokenError {
    /// There was an error performing the API Request
    #[error("tokens.error.apirequest")]
    ApiError {
        #[from]
        source: kp_api_requests::error::ApiRequestError,
    },`
    
    
This results in the type to be renamed from ApiRequestError to ApiRequestException
`sealed class ApiRequestException: Exception() {`
   
but the import not reflecting this

`   import com.sdk.api_requests.ApiRequestError`

PR #1847, opened 2023-11-13
Implement `Lift` for flat errors

bendk:
We noticed a regression in `0.25.x` where dictionaries containing flat errors caused code not to be compiled.  We never officially supported compound types that stored errors, but it seems better to allow the code to compile in this case.

The reason this isn't supported is that we can't lift flat errors, so therefore we can't lift the dictionary.  A better fix for this would be to not implement `Lift` on the dictionary in this case, which would cause a compile error rather than a runtime panic.  However, that's actually not so simple as noted in the comment so I went for an easier way.

This brings this very close to how they worked in `0.24.x` where we would generate an `FfiConverter` implementation for flat errors, but the lift half of it would just be panics.

### DIFF
--- a/uniffi_macros/src/error.rs
+++ b/uniffi_macros/src/error.rs
@@ -87,6 +87,7 @@ fn flat_error_ffi_converter_impl(
 ) -> TokenStream {
     let name = ident_to_string(ident);
     let lower_impl_spec = tagged_impl_header("Lower", ident, udl_mode);
+    let lift_impl_spec = tagged_impl_header("Lift", ident, udl_mode);
     let derive_ffi_traits = derive_ffi_traits(ident, udl_mode, &["ConvertError"]);
     let mod_path = match mod_path() {
         Ok(p) => p,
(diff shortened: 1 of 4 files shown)


uniffi_macros/src/error.rs
```diff
@@ -157,7 +157,31 @@ fn flat_error_ffi_converter_impl(
...
+            //
+            // Note: it would be better to not derive `Lift` for dictionaries containing flat
+            // errors, but getting the trait bounds and derived impls there would be much harder.
+            // For now, we just fail at runtime.
+            #[automatically_derived]
+            unsafe #lift_impl_spec {
+                type FfiType = ::uniffi::RustBuffer;
+
+                fn try_read(buf: &mut &[::std::primitive::u8]) -> ::uniffi::deps::anyhow::Result<Self> {
+                    panic!("Can't lift flat errors")
```
mhammond, line 174: is it worth opening an issue around getting these errors at buildtime instead of a runtime panic?

mhammond (commented):
This looks great, thanks, but shouldn't we have a test that "accidentally" exercises the panic?

uniffi_macros/src/error.rs
bendk: Good point, I made https://github.com/mozilla/uniffi-rs/issues/1850 for this.

bendk:
> This looks great, thanks, but shouldn't we have a test that "accidentally" exercises the panic?

Maybe we could test it from Rust, but I'd be worried about trying to exercise the panic through the foreign code.  Lowering a flat error from the foreign side is not allowed and implementations can choose to not implement it, segfault, or anything else.  I believe all of our implementations fail in a predictable way that could be tested, but I'd be worried that adding the test could break an external bindings generator.

mhammond (approved):
Thanks!

PR #1848, opened 2023-11-14
Improve docs re: proc macros

### LINKED ISSUE #1816: Improve docs around building / scaffolding when using proc macros
When investigating + patching my own project as a result of #1804, I came across the following areas for improvement. 

1. As far as I can tell, it is not strictly necessary to have a `build.rs` file with a call to `generate_scaffolding` anymore (see [Rust Scaffolding](https://github.com/mozilla/uniffi-rs/blob/eb97592f8c48a7f5cf02a94662b8b7861a6544f3/docs/manual/src/tutorial/Rust_scaffolding.md)) if you are using proc macros exclusively rather than UDL.
2. The [proc macros build workflow](https://github.com/mozilla/uniffi-rs/blob/eb97592f8c48a7f5cf02a94662b8b7861a6544f3/docs/manual/src/proc_macro/index.md?plain=1#L24C7-L24C7) doesn't make it clear how to generate mixed UDL and proc macro bindings. I never did figure this out myself, but I assume it's possible.
3. Closely related to 2, I never figured out how to generate the bindings without switching to library mode. I glossed over that section initially because it didn't register as important, but it turns out to be crucial at least as far as I can tell once you drop UDL completely.

I'd be happy to open a PR if the team could confirm the details, but for now they are all a bit fuzzy to me.

ianthetechie:
Closes #1816.

### DIFF
--- a/docs/manual/src/tutorial/Rust_scaffolding.md
+++ b/docs/manual/src/tutorial/Rust_scaffolding.md
@@ -38,7 +38,9 @@ uniffi::include_scaffolding!("math");
 ### Setup for crates using only proc macros
 
 If you are only using proc macros, you can skip `build.rs` entirely!
-All you need to do is add this to the top of `lib.rs`:
+All you need to do is add this to the top of `lib.rs`
+NOTE: This function takes an optional parameter, the [`namespace`](../udl/namespace.md) used by the component.
+If not specified, the crate name will be used as the namespace.
 
 ```rust
 uniffi::setup_scaffolding!();
--- a/docs/manual/src/tutorial/Rust_scaffolding.md
+++ b/docs/manual/src/tutorial/Rust_scaffolding.md
@@ -14,7 +14,12 @@ uniffi = "0.XX.0"
 uniffi = { version = "0.XX.0", features = ["build"] }
 ```
 
-Then create a `build.rs` file next to `Cargo.toml` that uses `uniffi` to generate the Rust scaffolding code.
+As noted in [Describing the interface](udl_file.md), UniFFI currently supports two methods of interface definitions: UDL files and proc macros.
+If you are using only proc macros, you can skip some boilerplate in your crate setup as well.
+
+### Setup for crates using UDL
+
+Crates using UDL need a `build.rs` file next to `Cargo.toml`. This uses `uniffi` to generate the Rust scaffolding code.
 
 ```rust
 fn main() {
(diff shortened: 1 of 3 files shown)


docs/manual/src/tutorial/Rust_scaffolding.md
```diff
@@ -30,7 +35,16 @@ uniffi::include_scaffolding!("math");
...
 **Note:** The file name is always `<namespace>.uniffi.rs`.
 
-Great! `add` is ready to see the outside world!
+### Setup for crates using only proc macros
+
+If you are only using proc macros, you can skip `build.rs` entirely!
+All you need to do is add this to the top of `lib.rs`:
+
+```rust
+uniffi::setup_scaffolding!();
```
mhammond, line 44: please add something like: The setup_scaffolding function takes one optional parameter, the ["namespace"](../udl/namespace,md] used by the component. If not specified, the name of the crate will be used as the namespace.

docs/manual/src/tutorial/foreign_language_bindings.md
```diff
@@ -38,11 +38,11 @@ creating a binary for each crate that uses UniFFI.  You can avoid this by creati
 
 Then your can run `uniffi-bindgen` from any create in your project using `cargo run -p uniffi-bindgen [args]`
 
-## Running uniffi-bindgen using a library file
+## Running uniffi-bindgen using a library file (RECOMMENDED)
 
 Use `generate --library` to generate foreign bindings by using a cdylib file built for your library.
 This flag was added in UniFFI 0.24 and can be more convenient than specifying the UDL file -- especially when multiple UniFFI-ed crates are built together in one library.
-The plan is to make library mode the default in a future UniFFI version.
+The plan is to make library mode the default in a future UniFFI version, and it is highly recommended to specify the flag for now.
```
mhammond, line 45: maybe add here "because some features simply don't work otherwise"?

docs/manual/src/tutorial/Rust_scaffolding.md
ianthetechie: Good call; that's super helpful.

docs/manual/src/tutorial/foreign_language_bindings.md
ianthetechie: Done.

mhammond (approved):
Thank you!

Issue #1854, opened 2023-11-18
External type defined in UDL can't be referenced in record defined by a procmacro


mhammond:
In our ext-types fixture we [import a type named UniffiOneInterface](https://github.com/mozilla/uniffi-rs/blob/802e4afa963e906823e29049da2984ff2f5696f2/fixtures/ext-types/lib/src/ext-types-lib.udl#L34-L35) which is [defined via UDL in another crate](https://github.com/mozilla/uniffi-rs/blob/802e4afa963e906823e29049da2984ff2f5696f2/fixtures/ext-types/uniffi-one/src/uniffi-one.udl#L12)

While that crate works, it fails if that same crate tries to refer to that type in a dictionary defined by a procmacro, eg:
```
#[derive(Default, uniffi::Record)]
pub struct ObjectsType {
    pub maybe_interface: Option<Arc<UniffiOneInterface>>,
}
```

fails with:
```
thread 'uniffi_foreign_language_testcase_test_imported_types_swift' panicked at 'assertion failed: `(left == right)`
  left: `External { module_path: "uniffi_one", name: "UniffiOneInterface", namespace: "uniffi_one_ns", kind: Interface, tagged: false }`,
 right: `External { module_path: "uniffi_one", name: "UniffiOneInterface", namespace: "uniffi_one_ns", kind: Interface, tagged: true }`', uniffi_bindgen/src/interface/universe.rs:49:17
```

The problem is that things get confused about whether the external type is tagged or not. This is another example of where having some types be tagged and others not causes problems.

PR #1855, opened 2023-11-18
Fix support for external objects and traits.

mhammond:
Some tweaks to Kotlin where necessary and apart from discovering #1854 things tended to work fine.


bendk (approved):
This makes me really want to implement ADR-0008.  But in the meantime, this change looks good.

Issue #1827, opened 2023-11-02
Foreign trait interface implementations can break when libraries expect a trait to be Rust-only


bendk:
In `0.25.0` this worked and the `bar` argument was treated as a normal `[ByRef]` argument.

```
#[uniffi::export]
pub trait Foo: Send + Sync {
     fn foo(&self, bar: &Bar);
}
```

However this currently fails in `main` because we now also generate a callback interface for each trait.  `ByRef` arguments are not supported there.

bendk:
Maybe we could add support for `ByRef` args in callback interfaces:
- Primitive types can be copied
- Types that are serialized to `RustBuffer`, probably don't need any special treatment.
- For objects/trait interfaces, could we clone the Arc?

If we could get this to work, then there's still a question of if the semantics are correct.  If the Rust code is passing a reference they find copying the data into buffer or cloning the Arc surprising.

mhammond:
> a question of if the semantics are correct.

Yeah, I'm a bit torn here. I don't like breaking consumers, but OTOH, I don't think we ever promised this would work, did we? It seems we didn't have test coverage for it otherwise we'd have noticed it before now. I agree that making it *look* like we are passing by reference just for us to turn it into a copy behind their back isn't ideal and "un-rusty". Is supporting this something we will regret as a foot-gun in a few years?

skeet70:
We have code that does this, and don't care about external consumers being able to implement my traits. They're used internally to cement the API and externally they let a consumer specify they need that interface (which my library presents multiple implementers of). If supporting callbacks with `ByRef` isn't a semantically good direction, we would want the ability to label the trait so a callback interface isn't generated for it.

mhammond:
yeah, we do need an attribute here.

mhammond:
> Maybe we could add support for `ByRef` args in callback interfaces:
> ...
> If we could get this to work, then there's still a question of if the semantics are correct. If the Rust code is passing a reference they find copying the data into buffer or cloning the Arc surprising.

Actually I think this is fine and preferable. Rust code calling the trait will not be impacted at all. Foreign code calling in will mean we have a newly created `Bar` we pass a reference to, but I don't see how that's surprising to anyone. Am I missing other cases?

What's involved in supporting byrefs there?

skeet70:
[...]

PR #1862, opened 2023-11-26
Docstrings for proc macros

ptitjes:


uniffi_macros/src/export/callback_interface.rs
```diff
@@ -182,18 +182,21 @@ pub(super) fn metadata_items(
...
-    );
+
+    let mut metadata_expr = quote! {
+        ::uniffi::MetadataBuffer::from_code(::uniffi::metadata::codes::CALLBACK_INTERFACE)
+            .concat_str(#module_path)
+            .concat_str(#trait_name)
+    };
+    if let Some(docstring) = docstring {
+        metadata_expr.extend(quote! { .concat_str(#docstring) })
+    }
```
jplatte, line 196: If adding a docstring is optional, don't we have to encode a bool for whether there is a docstring or not?

uniffi_macros/src/object.rs
```diff
@@ -131,17 +134,24 @@ pub(crate) fn interface_meta_static_var(
...
+
+    let mut metadata_expr = quote! {
+            ::uniffi::MetadataBuffer::from_code(::uniffi::metadata::codes::INTERFACE)
+                .concat_str(#module_path)
+                .concat_str(#name)
+                .concat_bool(#is_trait)
+    };
+    if let Some(docstring) = docstring {
+        metadata_expr.extend(quote! { .concat_str(#docstring) });
+    }
```
jplatte, line 149: Same here.

jplatte (commented):
Thanks for working on this!

Maybe you're already aware, but there is an alternate PR at #1498 for supporting something like this. As per comments there, I would prefer an approach like this though (as a frequent contributor, not a maintainer) so I'm glad there's now a PR for it :)

ptitjes:
> Thanks for working on this!

You're welcome! Please not that I am a very new Rust coder.

> Maybe you're already aware, but there is an alternate PR at #1498 for supporting something like this. As per comments there, I would prefer an approach like this though (as a frequent contributor, not a maintainer) so I'm glad there's now a PR for it :)

No, I missed it. But it indeed seems the approach used here is more straightforward.

fixtures/docstring-proc-macro/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+    One,
+    /// <docstring-enum-variant-2>
+    Two,
+}
+
+/// <docstring-associated-enum>
+#[derive(uniffi::Enum)]
+pub enum AssociatedEnumTest {
+    /// <docstring-associated-enum-variant>
+    Test { code: i16 },
```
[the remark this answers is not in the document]
ptitjes: Yeah I told myself that, as the code paths for fields handled that, and then I forgot.

uniffi_macros/src/export/callback_interface.rs
ptitjes: Well, I should really pass a String and not an Option<String>, as this is only called for callback interfaces, and `MetadataReader::read_callback_interface` always read it.

I should unwrap the option in the `ExportItem::Trait` match case of `expand_export`. I will fix that.

uniffi_macros/src/object.rs
ptitjes: Yeah I got a little bit confused here too.

uniffi_macros/src/export/callback_interface.rs
ptitjes: Fully fixed. No more Options.

uniffi_macros/src/object.rs
ptitjes: Fully fixed. No more Options.

jplatte:
cc @bendk @badboy

badboy:
oh yey, I did the same work in the past couple of days and it looks very similar. Given this was posted before I could even get to it I will spend some time on reviewing this instead.

fixtures/docstring-proc-macro/README.md
```diff
@@ -0,0 +1,3 @@
+# A basic test for uniffi components
+
+This test covers docstrings.
```
badboy, line 3:
```suggestion
This test covers docstrings in proc-macro mode.
```

fixtures/docstring-proc-macro/src/docstring-proc-macro.udl
```diff
@@ -0,0 +1,2 @@
+/// <docstring-namespace>
```
badboy, line 1: side-note: fwiw, when I implemented this I didn't use the UDL at all and only then realized that of course we can't get the crate's top-level docs as the namespace docs. But I guess that's fine.

uniffi_bindgen/src/bindings/swift/templates/macros.swift
```diff
@@ -57,14 +57,15 @@
 // Note the var_name and type_name filters.
 -#}
 {% macro field_list_decl(item) %}
-    {%- for field in item.fields() -%}
+    {% for field in item.fields() -%}
```
badboy, line 60: Minimal nit: I don't think you need to remove the `-` here. Then again it saves us on some whitespace and that's really not critical.

fixtures/docstring-proc-macro/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+        /// <docstring-variant-field>
+        code: i16,
+    },
+    /// <docstring-associated-enum-variant-2>
+    Test2 { code: i16 },
+}
+
+/// <docstring-error>
+#[derive(uniffi::Error, Debug, thiserror::Error)]
+enum ErrorTest {
```
badboy, line 30: If you add `#[uniffi(flat_error)]` here, the `message` field is exposed and the test file would be the same as for the `docstring` fixture.

fixtures/docstring-proc-macro/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+
+[dependencies]
+thiserror = "1.0"
+uniffi = { path = "../../uniffi" }
+
+[build-dependencies]
+uniffi = { path = "../../uniffi", features = ["build"] }
+
+[dev-dependencies]
+camino = "1.0.8"
```
badboy, line 21: camino is not used in the tests here.

```suggestion
```

badboy (approved):
Except for some naming and ordering this is the same as my implementation that I did 2 days ago and didn't yet push.

So ... good job! Minimal nits.

.gitignore
```diff
@@ -2,6 +2,7 @@ target
 .*.swp
 *.jar
 .vscode
+.idea
```
badboy, line 5: Normally I would reject this btw, but it's right after a line of `.vscode` so .. uh, guess we accepted that a while ago.

jplatte: What's the problem with it? It just helps contributors not accidentally commit unwanted files.

badboy: It tends to accumulate a list of editor-specific files over time, just because one more person used a different setup.
These should go into one's global gitignore (IMO); but again: I didn't reject it here and it's fine.

jplatte: I always forget that global `.gitignore` is a thing :)

(though I wonder if it's all that practical, there are projects where I've wanted to check in editor configuration though nowadays I prefer a `contrib/ide` directory instead)

mhammond:
This is great, thanks! I've a minor preference that we reused the existing docstring fixture for procmacros too, just so as things move forward it's much easier to test and rationalize about any differences between procmacros and UDL - so if you think that makes sense and have the energy to do that it would be great, but I'm not going to try and block this landing for that.

ptitjes:
> oh yey, I did the same work in the past couple of days and it looks very similar. Given this was posted before I could even get to it I will spend some time on reviewing this instead.

I am really sorry if stepped on your toes. There was no issue open, so I jumped on it as I needed it.

uniffi_bindgen/src/bindings/swift/templates/macros.swift
ptitjes: Indeed, fixed it.

fixtures/docstring-proc-macro/src/lib.rs
ptitjes: Nice, I missed that. Read the documentation too fast...

fixtures/docstring-proc-macro/Cargo.toml
ptitjes: Indeed.

ptitjes:
> just so as things move forward it's much easier to test and rationalize about any differences between procmacros and UDL

Not that I am too lazy, but I really think it is better to have separate fixture to easily compare the generated code with UDL or procmacros. As the symbols have the same names, you can easily diff and spot the differences. For example:

```diff
--- uniffi-fixture-docstring-49c2dd0b0f909ef/uniffi/fixture/docstring/uniffi_docstring.kt	2023-11-27 19:27:01.335778828 +0100
+++ uniffi-fixture-docstring-proc-macro-a3df46f603a93fe/uniffi/fixture/docstring/proc/macro/uniffi_docstring_proc_macro.kt	2023-11-27 19:27:12.867338022 +0100

...

@@ -999,6 +999,9 @@
      */
     data class Test(
         
+        /**
+         * <docstring-variant-field>
+         */
         val `code`: Short
         ) : AssociatedEnumTest() {
         companion object
```

ptitjes:
@badboy @mhammond @jplatte  All changes done.

fixtures/docstring-proc-macro/tests/bindings/test_docstring.py
```diff
@@ -0,0 +1,52 @@
...
+# Test objects
+assert ObjectTest.__doc__ == "<docstring-object>"
+assert ObjectTest.__init__.__doc__ == "<docstring-primary-constructor>"
+assert ObjectTest.new_alternate.__doc__ == "<docstring-alternate-constructor>"
+assert ObjectTest.test.__doc__ == "<docstring-method>"
+
+# Test records
+assert RecordTest.__doc__ == "<docstring-record>"
+
+# `__doc__` is not supported for class fields
```
skeet70, line 47: I can't test this on my project due to #1827, but to clarify the docs are still generated for the things with these comments, it's only that `__doc__` doesn't work so the test can't be done?

ptitjes:
I forgot to commit the `Cargo.lock`. Now we should be good.

skeet70 (commented):
Traits and trait methods get docstrings, and in the test code the implementer of the callback trait inherits the `CallbackTest` docstring. Does the same happen with a

```rust
#[uniffi::export]
pub trait NotCallbackTest {
  /// <docstring-trait-method>
  fn other_test(&self);
}

#[uniffi::export]
impl NotCallbackTest for Object {
  fn other_test(&self) {}
}
```

or does the method that shows up from the `impl` require its own duplicate docstring be defined?

fixtures/docstring-proc-macro/tests/bindings/test_docstring.py
ptitjes: I am not a Pythonista but as I understand it, it is a limitation of `__doc__`. After reading the comments, I did not try to improve the test on this.

uniffi_macros/src/util.rs
```diff
@@ -276,3 +276,20 @@ impl Parse for ExternalTypeItem {
         })
     }
 }
+
+pub(crate) fn extract_docstring(attrs: &[Attribute]) -> syn::Result<String> {
```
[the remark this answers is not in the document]
mhammond: also space before "foo", but yeah, we should. IMO we should also check udl does the same?

fixtures/docstring-proc-macro/tests/bindings/test_docstring.py
ptitjes: OK I got what you mean. So the answer is yes. You can look at the `EnumTemplate.py` changes I made in this PR, where I add a docstring to fields of associated enums. However it was not possible for errors.

uniffi_macros/src/util.rs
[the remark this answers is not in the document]
badboy: to speed this along I'm gonna merge and tackle that separately

badboy:
GitHub complained about merge conflicts even after I resolved them in the web UI, so I opted to rebase manually and force-push. All I did was clear up the merge conflict in changelog.md.

uniffi_macros/src/util.rs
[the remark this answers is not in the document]
ptitjes: > Turns out: not a problem in generated code because of the use of textwrap::dedent (and textwrap::indent)

However, useless spaces would be included in the metadata in library files, right?
So maybe adding a trim here would be a good thing:
https://github.com/mozilla/uniffi-rs/blob/main/uniffi_macros/src/util.rs#L288

Do you want me to do a mini-PR?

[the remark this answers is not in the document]
badboy: True. I'll gladly accept a PR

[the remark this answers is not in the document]
ptitjes: @badboy https://github.com/mozilla/uniffi-rs/pull/1878

PR #1866, opened 2023-11-26
Add README to published crates

### LINKED ISSUE #1860: Add README.md to all published crates
eg, https://crates.io/crates/uniffi_udl says "uniffi_udl v0.25.2 appears to have no README.md file" - all crates we publish probably should.

ptitjes:
Fixes #1860.

### DIFF
--- a/uniffi/Cargo.toml
+++ b/uniffi/Cargo.toml
@@ -12,6 +12,7 @@ authors = ["Firefox Sync Team <sync-team@mozilla.com>"]
 license = "MPL-2.0"
 edition = "2021"
 keywords = ["ffi", "bindgen"]
+readme = "../README.md"
 
 [dependencies]
 uniffi_bindgen = { path = "../uniffi_bindgen", version = "=0.25.2", optional = true }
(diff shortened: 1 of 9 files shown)


uniffi/Cargo.toml
```diff
@@ -12,6 +12,7 @@ authors = ["Firefox Sync Team <sync-team@mozilla.com>"]
 license = "MPL-2.0"
 edition = "2021"
 keywords = ["ffi", "bindgen"]
+readme = "../README.md"
```
jplatte, line 15: I don't think this will actually work, since nothing higher up than the `Cargo.toml` is included in the crate tarball.. Unless maybe mentioning a file in the manifest makes Cargo include it? I'm skeptical.

ptitjes: It does. As you can see, this is what I did (following the documentation) here:
https://gitlab.com/trixnity/uniffi-kotlin-multiplatform-bindings/-/blob/main/bindgen/Cargo.toml

And it does the job:
https://crates.io/crates/uniffi_bindgen_kotlin_multiplatform

As per the documentation:
> The readme field should be the path to a file in the package root (relative to this Cargo.toml) that contains general information about the package. This file will be transferred to the registry when you publish.
https://doc.rust-lang.org/cargo/reference/manifest.html#the-readme-field

jplatte: Ah, great!

mhammond (approved):
Nice, thanks!

PR #1869, opened 2023-11-27
Implement the Cleaner for kotlin Objects

jhugman:
This PR adds:

- `java.lang.ref.Cleaner` to perform cleanup of Rust Arcs shortly after kotlin Garbage Collection.
- An Android specific cleaner, with configuration to switch between them.

It leaves intact the `AutoCloseable` infrastructure previously used, so as to allow developers the choice of 

1. Prompt clean up when they are finished with the object
2. An automatic backup, sometime later.

Some further work may be necessary: 

1. removing `Disposable` in favour of [`java.lang.AutoCloseable`](https://developer.android.com/reference/kotlin/java/lang/AutoCloseable).
2. maybe: putting `AutoCloseable` behind a `[Traits=]` trait declaration.

For ease of review, and backwards compatibility, I'd suggest keeping these separate from this PR.


fixtures/coverall/tests/bindings/test_coverall.kts
```diff
@@ -488,3 +509,7 @@ Coveralls("using_fakes_with_real_objects_crashes").use { coveralls ->
     }
     assert(exception != null)
 }
+
+// This is from an earlier GC test; ealier, we made 1000 new objects.
+// By now, the GC has had time to clean up, and now we should see 0 alive.
+assert(getNumAlive() == 0UL) { "Num alive is ${getNumAlive()}. GC/Cleaner hasn't run fully" };
```
badboy, line 515: I guess there's no guarantee that the GC actually ran at this point, right? Will this lead to annoying intermittents on CI?

uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -29,18 +29,16 @@ open class {{ impl_class_name }} : FFIObject, {{ interface_name }} {
...
-     */
-    override protected fun freeRustArcPtr() {
-        this.pointer?.let { ptr ->
-            rustCall() { status ->
-                _UniFFILib.INSTANCE.{{ obj.ffi_object_free().name() }}(ptr, status)
+    override val cleanable: Cleaner.Cleanable = _UniFFILib.CLEANER.register(this, _CleanAction(pointer))
+
+    // Use a static inner class instead of a closure so as not to accidentally
+    // capture `this` as part of the cleanable's action.
+    private class _CleanAction(private val pointer: Pointer?) : Runnable {
```
bendk, line 36: Nit: I like using `Uniffi` whenever we need a unique prefix.

uniffi_bindgen/src/bindings/kotlin/templates/ObjectRuntime.kt
```diff
@@ -71,13 +74,23 @@
...
+//
+// In this case, `destroy`ing manually may be a better solution.
 //
-// Sigh...all of this for want of a robust finalization mechanism.
+// The cleaner can live side by side with the manual calling of `destroy`. In the order of preference, uniffi objects
+// with Rust peers can be reclaimed:
+//
+// 1. By calling the `destroy` method of the object, which calls `rustObject.free()`. If that doesn't happen:
+// 2. When the object becomes unreachable, AND the Cleaner thread gets to call `rustObject.free()`. If the thread is starved then:
+// 3. The memory is reclaimed when the process terminates.
```
bendk, line 93: I prefer relying on `2` and would love to get rid of the `destroy()` method.  It gets around [corner cases](https://github.com/mozilla/uniffi-rs/pull/1787#issuecomment-1788966499) and feels more consistent with GC-based language.

How serious is the risk of cleaner thread starvation?  I hope it's something most applications can ignore.

I don't think this needs to be a blocker.  I'd be happy to merge this one as-is and discuss in a future PR.

bendk (commented):
Looks good to me.  The only reason I didn't check approve is I also wonder about the tests failing intermittently.

fixtures/coverall/tests/bindings/test_coverall.kts
jhugman: Perhaps. On my slow machine, this test never failed. 

The test failure on first push (not this one) suggested that the thread was working rather better on CI than my machine.

The alternative: putting in ~500ms sleep– which I was trying to avoid.

jmartinesp:
I tested this on our Android app and the app instantly crashed. 

It turns out both [java.lang.ref.Cleaner](https://developer.android.com/reference/java/lang/ref/Cleaner) and [android.system.SystemCleaner](https://developer.android.com/reference/android/system/SystemCleaner) are only available from API 33+, which is only used in [around 36% of all available devices](https://apilevels.com/). Given some input from my team, I thought this class would be desugared and added to our Kotlin sources from the Java 9+ APIs, but it turns out it's not. However, JNA does have its own Cleaner class (`com.sun.jna.internal.Cleaner`).

So maybe for the Android version, something like this could be used?

```kotlin
val cleaner = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
    SystemCleaner.cleaner()
} else {
    com.sun.jna.internal.Cleaner.getCleaner()
}
```

The problem with these is, these 2 `Cleaner` instances come from different classes, so you'd have to either write a wrapper and cast it accordingly or do that in-place for every call to this cleaner. Otherwise, maybe a custom `Cleaner`  implementation based on either `java.lang.ref.Cleaner` or `SystemCleaner` could be provided.

jmartinesp:
In https://github.com/jmartinesp/uniffi-rs/tree/android-cleaner-wrapper I created an example wrapper for Android based on the `android_cleaner` config value:

- If `android_cleaner` is disabled, the Java 9+ Cleaner is used (the tests pass, so this should work as expected).
- If `android_cleaner` is enabled, the `AndroidCleaner` interface is used instead:
    - On API 33+: `AndroidSystemCleaner` is implemented, which uses the same Java 9+ Cleaner since it's available.
    - On older APIs, `AndroidJnaCleaner` is implemented instead, which uses the JNA Cleaner implementation.

I checked both Android implementations work using devices with different API versions, so maybe you can use something like this as a base for the implementation.

fixtures/coverall/tests/bindings/test_coverall.kts
jmartinesp: What about manually calling `System.gc()` before? It doesn't actually guarantee the GC will be done at that point, but it should make it more likely.

jhugman:
> In https://github.com/jmartinesp/uniffi-rs/tree/android-cleaner-wrapper I created an example wrapper for Android based

Oh, amazing. Thank you! 

I'm pretty busy this week, so probably won't get to this until next week. Do you want to take this over, or can it wait until then?

jmartinesp:
> I'm pretty busy this week, so probably won't get to this until next week. Do you want to take this over, or can it wait until then?

Take it over as in creating a new PR with these changes plus mine? If so, I think the work should probably continue in this PR, maybe cherry picking my `AndroidCleaner` implementation if it's good enough. 

Also, I don't think anyone is in a hurry for this, since anyone that uses UniFFI with Kotlin is already freeing the memory manually, so we can keep doing that until this is ready. That said, it would be super useful to have it 😅 .

uniffi_bindgen/src/bindings/kotlin/templates/ObjectRuntime.kt
jhugman: (1) is the escape hatch for when thread starvation happens. As you say, most apps won't have to worry about it.

We should deprecate and then remove the `Disposable` interface, and the `destroy` method, though this isn't done here, because we should go through the formal deprecation process.

jhugman (commented):
@jmartinesp would you mind having another look at this sometime? I haven't been able to test it on Android, either with or without `android_cleaner = true`.

jmartinesp:
> > @jmartinesp would you mind having another look at this sometime? I haven't been able to test it on Android, either with or without `android_cleaner = true`.
> 
> I tested it with the current contents and it's working fine on an Android 14 device but it's failing on an Android 7 emulator because it's trying to use `SystemCleaner`, which isn't available. Are you sure the commits were correctly cherry-picked? The contents of `ObjectRuntime.kt` look different.

Just to confirm, I re-tested the branch on my fork and it works well for both APIs, so something definitely went wrong when cherry-picking its contents.

jmartinesp:
Sorry, I deleted the comments above because I just realised I was testing an outdated version of the branch 🤦 .

With the latest commits I can confirm it works both on the newest API (API 34, Android 14) and old ones (API 23, Android 7) using their respective versions of the cleaner.

jmartinesp:
It also works well with the workaround you added with `try/catch` when `java.lang.ref.Cleaner` cannot be found, it defaults to the JNA Cleaner in older APIs, and it uses the actual java cleaner when it is available. 

However, maybe some kind of message should be logged here? I wouldn't have realised what was happening until I took a look at the implementation details.

jhugman:
> However, maybe some kind of message should be logged here?

There doesn't seem to be anything logged by the Kotlin FFI at this point. Where do you want to log to in (a potentially but not definitely non-Android) context? e.g. `java.util.logging.Logger`, `println`.

I'm happy to log if this is a solved problem already, but I'd also be happy to not.

Last commit fixes up documentation.

jmartinesp:
> Where do you want to log to in (a potentially but not definitely non-Android) context? e.g. `java.util.logging.Logger`, `println`.

Good question. I think `Logger` won't work with logcat on Android, so maybe `println` would be better. In any case, if this creates further problems maybe it's better not to do it until you decide on how to print logs from the library separately.

bendk (approved):
Looks good to me.  I'm mostly relying on @jmartinesp's review, but I'm happy to rubber stamp.

julian-baumann:
I am getting the following error:
```
Unresolved reference: Cleaner
```
in this generated code:
```kotlin
// The fallback Jna cleaner, which is available for both Android, and the JVM.
private class UniffiJnaCleaner : UniffiCleaner {
    private val cleaner = com.sun.jna.internal.Cleaner.getCleaner()

    override fun register(value: Any, cleanUpTask: Runnable): UniffiCleaner.Cleanable =
        UniffiJnaCleanable(cleaner.register(value, cleanUpTask))
}

private class UniffiJnaCleanable(
    private val cleanable: com.sun.jna.internal.Cleaner.Cleanable,
) : UniffiCleaner.Cleanable {
    override fun clean() = cleanable.clean()
}
```

I have set `android = true` and `android_cleaner = true` in my `uniffi.toml` config, but the issue persists. Do I have to add some sort of dependency for `com.sun.jna.internal.Cleaner`?

jmartinesp:
JNA's `Cleaner` was added in `v5.12.0` if I'm not mistaken. Could your JNA version be previous to that one?

uniffi_bindgen/src/bindings/kotlin/templates/ObjectRuntime.kt
```diff
@@ -71,13 +81,19 @@
...
 //
-// In the future we may be able to replace some of this with automatic finalization logic, such as using
-// the new "Cleaner" functionality in Java 9. The above scheme has been designed to work even if `destroy` is
-// invoked by garbage-collection machinery rather than by calling code (which by the way, it's apparently also
-// possible for the JVM to finalize an object while there is an in-flight call to one of its methods [1],
-// so there would still be some complexity here).
+// This makes a cleaner a better alternative to _not_ calling `destroy()` as
+// and when the object is finished with, but the abstraction is not perfect: if the Rust object's `drop`
+// method is slow, and/or there are many objects to cleanup, and it's on a low end Android device, then the cleaner
+// thread may be starved, and the app will leak memory.
```
mgeisler, line 87: Hi there :wave:

I'm looking to use UniFFI for a library which will be used in Android. The paragraph here gave me pause: will an application actually leak memory in an unrecoverable way, or will it "just" hold on to the memory longer than necessary?

I'm asking since the preconditions seem to be that `drop` is slow or that it takes a long time to cleanup all the objects. That makes me think that the app will be slower than ideal, but that it will still (eventually) free the memory?

Even if the phone runs out of RAM, the OS will eventually clean things up as mentioned further up in the comment.

What am I missing here?

mhammond: I'm far far from an expert in Kotlin, but my understanding of the situation is that the memory held by the Rust object will not be free'd until the generated object's generated `destroy()` method is called. However, the JS garbage collector is both non-deterministic and makes no real attempt to keep memory usage low. The "modern" way of cleaning stuff up is apparently via this "cleaner" mechanism, but even this isn't perfect as it runs on its own thread and nothing guarantees that thread doesn't get starved. In that situation you will have what looks very much like a leak even though it's not really a leak - there's nothing stopping these things getting collected, it's just that they have not been.

Firefox for Android doesn't really exercise much of this - the objects we use tend to be long lived and small in number, whereas most data is stored in records which doesn't have this consideration as records don't have an "attached" Rust object.

mgeisler: Hi @mhammond, thanks a lot for the quick reply!

> I'm far far from an expert in Kotlin,

Me neither, for the record 😅 I learned Java some 20 years ago and I broadly hope Kotlin is a nicer version of that.

> In that situation you will have what looks very much like a leak even though it's not really a leak - there's nothing stopping these things getting collected, it's just that they have not been.

That makes sense! Perhaps it's just a small matter of terminology: I'm used to thinking of a leak as "ups, I threw away the pointer!" and as a result, the allocation cannot be freed.

The indeterministic behavior of a garbage collector is something else for me — even if it results in the program using more memory than intended.

I guess a reference cycle in Rust also falls under the same category of leaks now that I think of it.

mhammond: > That makes sense! Perhaps it's just a small matter of terminology: I'm used to thinking of a leak as "ups, I threw away the pointer!" and as a result, the allocation cannot be freed.
> 
> The indeterministic behavior of a garbage collector is something else for me — even if it results in the program using more memory than intended.

Me too! However, our Android engineers are very vocal in treating it far more seriously than that. I can only assume that's because so many bugs in our Android apps are due to our processes being killed due to memory pressure - I guess Android doesn't simply run a GC to try and reclaim memory before unilaterally killing it.

mgeisler: I see... I actually work in Android, but in the Rust part, so I don't really know anything about how apps work. But I can see how people will want stern warnings about this when it leads to the app being killed at random times.

Thanks for clearing it up for me!

PR #1875, opened 2023-11-30
Support docstrings longer than 255 characters

### LINKED ISSUE #1874: Docstrings for proc macros compile error when > 255 characters
I saw the the proc macro support for docstrings was merged recently so I decided to test it in some projects, and I noticed that any  docstring longer than 255 characters will fail to compile with the following error:

```
error[E0080]: evaluation of constant value failed
   --> /Users/dani/Desktop/uniffi-rs/uniffi_core/src/metadata.rs:178:9
    |
178 |         assert!(string.len() < 256);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'assertion failed: string.len() < 256'
    |
note: inside `MetadataBuffer::concat_str`
```

This is the code used to test this:
```rust
/// This is a very long multi line test docstring that exceeds 255 characters.
/// This is a very long multi line test docstring that exceeds 255 characters.
/// This is a very long multi line test docstring that exceeds 255 characters.
/// This is a very long multi line test docstring that exceeds 255 characters.
#[uniffi::export]
pub fn test2() {
}
```

This seems to be caused by the way `MetadataBuffer` stores strings, by encoding their length as a `u8`. This limits their max length to 255, which is easily exceeded if the docstring contains some complex explanation or code example.

My first thought would be to update `MetadataBuffer::concat_str` to encode the length as a `u16` instead, and change `MetadataReader` to do the same, but I'm not sure if that will have some adverse effects somewhere else.

dani-garcia:
Fixes #1874

Docstrings are now using a u16 for length encoding to remove the 255 character limit. This should be more than enough for the buffer size that is used.

Implemented using separate long_string encoding to avoid increasing the size of the rest of the metadata.

### DIFF
--- a/fixtures/docstring-proc-macro/src/lib.rs
+++ b/fixtures/docstring-proc-macro/src/lib.rs
@@ -94,4 +94,11 @@ pub trait CallbackTest {
     fn test(&self);
 }
 
+/// This is a very long multi line test docstring that exceeds 255 characters.
+/// This is a very long multi line test docstring that exceeds 255 characters.
+/// This is a very long multi line test docstring that exceeds 255 characters.
+/// This is a very long multi line test docstring that exceeds 255 characters.
+#[uniffi::export]
+pub fn test_long_docstring() {}
+
 uniffi::include_scaffolding!("docstring-proc-macro");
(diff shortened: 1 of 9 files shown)


badboy (approved):
> This should be more than enough for the buffer size that is used.

famous last words (especially given that the buffer itsels is limited to 4096 bytes right now).

PR #1878, opened 2023-12-01
fix(proc-macros): trim docstring lines

ptitjes:
We avoid storing useless spaces in proc-macros metadata.

### DIFF
--- a/uniffi_macros/src/util.rs
+++ b/uniffi_macros/src/util.rs
@@ -285,7 +285,7 @@ pub(crate) fn extract_docstring(attrs: &[Attribute]) -> syn::Result<String> {
             let name_value = attr.meta.require_name_value()?;
             if let Expr::Lit(expr) = &name_value.value {
                 if let Lit::Str(lit_str) = &expr.lit {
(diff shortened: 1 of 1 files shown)


uniffi_macros/src/util.rs
```diff
@@ -285,7 +285,7 @@ pub(crate) fn extract_docstring(attrs: &[Attribute]) -> syn::Result<String> {
             let name_value = attr.meta.require_name_value()?;
             if let Expr::Lit(expr) = &name_value.value {
                 if let Lit::Str(lit_str) = &expr.lit {
-                    return Ok(lit_str.value());
+                    return Ok(lit_str.value().trim().to_string());
```
jplatte, line 288:
```suggestion
                    return Ok(lit_str.value().trim().to_owned());
```

same thing, but clearer IMHO

arg0d:
Wouldn't this break indentation?

badboy:
> Wouldn't this break indentation?

I had the same thought over the weekend. I'll add a test case later so we properly capture (and handle) this.

PR #1879, opened 2023-12-01
Updated lower validation for Python/Ruby

### LINKED ISSUE #1797: Object.lower() can result in use-after-free errors

bendk:
I want to fix #1797 by cloning the object handle in lower().  However, this presents an issue in the dynamic languages because they perform various failable checks in lower(): type checks, bounds checks, etc.  If the check fails, then the cloned handle will be leaked.

Prevent this by adding a `check()` method.  We do the check first then once we get to `lower()`, nothing can fail.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/gen_python/mod.rs
+++ b/uniffi_bindgen/src/bindings/python/gen_python/mod.rs
@@ -456,6 +456,10 @@ pub mod filters {
         Ok(format!("{}.lift", ffi_converter_name(as_ct)?))
     }
 
+    pub(super) fn check_lower_fn(as_ct: &impl AsCodeType) -> Result<String, askama::Error> {
+        Ok(format!("{}.check_lower", ffi_converter_name(as_ct)?))
+    }
+
     pub(super) fn lower_fn(as_ct: &impl AsCodeType) -> Result<String, askama::Error> {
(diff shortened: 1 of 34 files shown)


mhammond (approved):
LGTM, but the name 'check' isn't particularly clear about *what* is being checked. How do you feel about a more explicit name, eg, `check_lowerable()`?

PR #1880, opened 2023-12-01
Clone objects when lowering them (#1797)

bendk:
(This PR is on top of #1879, let's merge that one first)

Currently, `lower()` always returns a borrow of the object handle.  This is fine for function arguments, since you know the object is still alive on the stack while the function is being called.  However, for function returns this is not correct.

To fix this: clone the handle in `lower()`

Added a test for this -- it was surprisingly easy to cause a segfault with the current behavior.


uniffi_macros/src/export/trait_interface.rs
```diff
@@ -30,15 +29,32 @@ pub(super) fn gen_trait_scaffolding(
...
+                let ptr = ptr as *mut std::sync::Arc<dyn #self_ident>;
+                let arc = unsafe { ::std::sync::Arc::clone(&*ptr) };
+                Ok(::std::boxed::Box::into_raw(::std::boxed::Box::new(arc)) as  *const ::std::ffi::c_void)
+            })
+        }
+
         #[doc(hidden)]
         #[no_mangle]
-        pub extern "C" fn #free_fn_ident(
+        pub unsafe extern "C" fn #free_fn_ident(
```
bendk, line 57: Clippy complained about the lack of unsafe in the clone function.  I think free is just as unsafe, but clippy doesn't recognize it, so I added the keyword here to.  It shouldn't make a difference, these functions are only called across the FFI anyways.

uniffi_bindgen/src/interface/object.rs
```diff
@@ -99,6 +97,8 @@ pub struct Object {
     //    hash value we're trying to calculate here, so excluding it
     //    avoids a weird circular dependency in the calculation.
     #[checksum_ignore]
+    pub(super) ffi_func_clone: FfiFunction,
```
mhammond, line 100: needs a short comment like the others

uniffi_macros/src/export/trait_interface.rs
mhammond: please add a one-liner of that as a comment :)

Issue #1792, opened 2023-10-13
Ability to define underlying type for enums


skhamis:
Allow the Enum's underlying type to be defined so we can leverage the foreign languages built-in functions.

## The issue

Defining an enum in rust as so:

Rust
```rust
#[repr(u8)]
#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
pub enum VisitType {
    Link = 1,
    Typed = 2,
}
```

UDL
```
enum VisitType {
 "Link",
 "Typed",
}
```

Generates

swift
```swift
public enum VisitType {
    case link
    case typed
}
```
kotlin
```kotlin
enum class VisitType {
    LINK, TYPED
}
```

The problem is at least in swift we lose the ability to init the enum as a raw value or turn the enum into a raw value without recreating those values via `extension VisitType...`

# Proposed solution

a decorated in the udl (or proc macro) to define the underlying type for the Enum. Using the above example:
UDL
```
[Int]
enum VisitType {
 "Link",
 "Typed",
}
```
would generate

swift
```swift
public enum VisitType: Int {
    case link
    case typed
}
```
kotlin
```kotlin
enum class VisitType(var type: Int) {
    LINK, TYPED
}

Thus swift can take advantage of using `VisitType(rawValue: v)` as well as `let value = visitType.rawValue` and Kotlin as well.

linabutler:
Thanks for filing this! Over in https://github.com/mozilla/application-services/pull/5867#pullrequestreview-1676696796, @ncloudioj was just commenting how nice it would have been to pass a bitset instead of a `Vec` of enum variants.

I think that's a little more involved than your suggestion—a bitset's variants would need to have specific raw values, not just implicitly assigned ones—but it would be wonderful to be able to support both!

skhamis:
> Thanks for filing this! Over in [mozilla/application-services#5867 (review)](https://github.com/mozilla/application-services/pull/5867#pullrequestreview-1676696796), @ncloudioj was just commenting how nice it would have been to pass a bitset instead of a `Vec` of enum variants.
> 
> I think that's a little more involved than your suggestion—a bitset's variants would need to have specific raw values, not just implicitly assigned ones—but it would be wonderful to be able to support both, as well as other underlying types like strings!

yes! I'm definitely okay with either starting out implicitly assigning or just go the full manual assigning. Though as you mentioned that might involved a little thinking of designing how the proc-macro/udl would define that

mhammond:
[...]

PR #1885, opened 2023-12-04
Add enum variant discriminator metadata and fill it in for procmacros.

mhammond:
Also implements exposing the discriminator for Python.

This is the most complex part on the path to #1792.


uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -202,6 +202,28 @@ impl Enum {
         &self.variants
     }
 
+    // Get the literal value to use for the specified variant's discriminant.
+    // Tries to follow Rust's rules when mixing specified and unspecified values,
+    // but only promises to get specified values correct.
```
bendk, line 207: It seems like it follows the logic to me, which cases would it get wrong?  If there are cases where we get it wrong, we should add a note in the docs about this.

Also, even if this doesn't match Rust's rules, each discriminant must be unique.   Otherwise the foreign enums will probably not work.  Can you add a note about that and maybe even check in the code?

mhammond: > It seems like it follows the logic to me, which cases would it get wrong? 

I read the Rust docs but not the implementation. The docs aren't as precise as I'd like, hence me trying to not oversell this implementation. I guess I should just have a bit more faith :)

> Also, even if this doesn't match Rust's rules, each discriminant must be unique. Otherwise the foreign enums will probably not work.

I should add a comment about this, but my assumption here is that we can leave that checking to Rust. If there is a variant with a duplicate discriminate then we can just assume the Rust compilation will fail, therefore we can assume there are no dupes?

> Can you add a note about that and maybe even check in the code?

I wanted to avoid that is it would mean I need to check *every* variant even when just checking the first instance, and probably would need a hashmap to remember what I'd already seen. It seems far far simpler to assume the Rust compiler checks this.

WDYT?

mhammond: Oh, the other thing it doesn't check is "overflow" - eg, the Rust compiler will complain if a `repr(u8)` overflows (eg, when one variant specified `0xFF` and the next variant specifies no value). That code doesn't even *know* the repr(), so that too is something I'm fine assuming the Rust compiler enforces.

mhammond:
IIRC, Ben and I agreed this really just needs the comments updated. Let me know if I forgot something!

bendk (approved):
That's what I remember too.  The new comment looks great to me.

Issue #1736, opened 2023-09-06
oneshot transitively depends on `windows`


badboy:
https://github.com/mozilla/uniffi-rs/commit/d4a6e652a122bd0a669cf002c99c6026b8f60997 pulled in the `oneshot` dependency, which itself depends on loom under a `cfg`: https://github.com/faern/oneshot/blob/80f4c3f0ef6bf2051bf3e59166194fde3c1d2699/Cargo.toml#L26-L27
`loom` depends on the `windows` crate which we don't want to pull into m-c.
unfortunately the dependency resolution is not smart enough and _does_ pull that in, making m-c fail on an import of uniffi with this.
The hacky way is to patch `oneshot` (remove the testing parts) locally.

I wonder if there's a better way to handle that in uniffi instead?
Definitely annoying but maybe a short term solution is to inline that crate?

cc @bendk

jplatte:
What do you mean by "pull in"? IIRC, cfg-conditional dependencies get recorded in the lockfile by design (such that you have a stable lockfile across platforms and such), but only built when the cfg is actually active. I'm not sure when they're downloaded.

bendk:
I'm guessing "pull in" means copied by `cargo vendor`.

That's annoying indeed.  I think inlining should be okay, the crate itself is very small.  Another option would be to implement the API ourselves using a Mutex like we had before.  The main reason to pull in that crate was that it had a nice API, performance was secondary.

I can take this on.  Do you have a preference between the two?

badboy:
@jplatte ah yes, sorry; as bendk said: by cargo vendor and we have additional tooling making sure that we don't vendor some crates (and `windows` is huge so it's currently not allowed).

bendk:
I forked the oneshot repo and made this commit that removes the `loom` target: https://github.com/bendk/oneshot/commit/1f3c657c8073aec4f0b6ebac7be33b4851644745

I think we can use this to keep the dependency simple in UniFFI and avoid blowing up the vendor directory in moz-central.

mgeisler:
Hi @badboy and @bendk,

I'm currently in the process of vendoring the UniFFI crates for use in Android. The forked oneshot crate is making my life a little interesting here. Not a blocker, but I thought you might be interested in hearing about it: https://r.android.com/3055962.

[...]

PR #1886, opened 2023-12-05
Depend on the patched oneshot version (#1736)

bendk:
After trying a couple times to vendor our code into moz-central using a patch in `Cargo.toml`, I think this is a better solution.  I'm hoping to merge this into the `0.25.x` branch, release `0.25.3`, and use that.

This will test our release strategy a bit.  We should be able to upgrade moz-central to using `0.25.3` without changing Glean or app-services.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -649,19 +649,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "generator"
-version = "0.7.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5cc16584ff22b460a382b7feec54b23d2908d858152e5739a120b949293bd74e"
-dependencies = [
- "cc",
- "libc",
- "log",
- "rustversion",
- "windows",
-]
-
-[[package]]
 name = "gimli"
 version = "0.28.0"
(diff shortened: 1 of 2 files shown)


bendk:
> you cannot release `uniffi_core` to crates.io with a git dependency.
> One way would be to publish a `bendk-oneshot` crate and depend on that.

TIL.  I named the crate `oneshot-uniffi` and updated `Cargo.toml`.

mhammond:
What changes have we made to oneshot? Have we tried upstreaming them (eg, via a feature)? Someone will have to do a fair bit of sleuthing to even *find* the sources to the forked version, so this situation doesn't seem ideal longer term - I'm wondering what we are doing to prevent it being a long term thing?

bendk:
Here's the change: https://github.com/bendk/oneshot/commit/a95e9eda47f8680921039025627b56181594a059

The issue is that that the `loom` dependency is only used for the `loom` target, but that's still an issue for moz-central because the vendoring process brings in all dependencies for all possible target/feature combinations.

It's not a great situation, but I can't think of a better solution.  Maybe the long-term solution is to wait for `cargo vendor` to add support for disabling certain targets (something like https://github.com/rust-lang/cargo/issues/7058)

PR #1887, opened 2023-12-07
V0.25.3 changes

bendk:
Changes for the `0.25.3` release.

For most consumers this will not change anything.  For consumers like mozilla-central that vendor in `uniffi`, this will mean they don't need to vendor in `loom` and its dependencies, which would never be used in practice.

I ran `cargo release` with the `--no-publish` flag so that we can approve the PR before I make the official release.  This seems like a better order of operations to me.  If this works and people like it, I'll update the release docs .


mhammond (approved):
`--no-publish` makes sense to me! I think we can also avoid those `chore` checkins in most cases? Can we update the docs/process for that too?

PR #1889, opened 2023-12-07
Updating the docs tooling around the release process

bendk:
The main change is that we will now use the `cargo release --no-publish` flag to avoid publishing to `crates.io` until the release PR is approved.

Added the `--execute` flag by default.  I don't think we need to do dry runs since we're now only creating a commit and waiting for PR approval.

Updated `release-process.md` to reflect the new process.

I left the current wording about separating `uniffi` releases from the rest of the crates, although I noticed that in practice we haven't been doing that.  All of our major changes in the last year bumped `UNIFFI_CONTRACT_VERSION`, so we never had a chance to, for example, bump the `unffi-bindgen` version without also bumping the `uniffi` version.  Maybe we should revisit this decision, but I didn't want to do that in this change.

A consequence of the last note is that we still need to make the 2 `Chore: Release` commits.


tools/publish-release.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -ex
+# Note: make sure these are ordered so that dependencies come before the crates that depend on them
```
mhammond, line 4: IIUC we never expect this to run in automation, so do you think it makes sense to, say, show the output of `git status` and check the user really does want to publish something? eg, I'm worried someone will execute, say, `publish-release.sh --help`

bendk: Good point, I added a check that `git status` is clean and also added a confirmation prompt.

tools/publish-release.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+
+if [[ `git status --porcelain` ]]; then
+  echo "Uncommited changes."
+  echo "Please commit changes and run this command again"
+  echo
+  echo "See docs/release.md for details"
+  exit
```
badboy, line 8:
```suggestion
  exit 2
```

Make this more explicitly fail?

badboy (changes requested):

badboy (approved):

PR #1899, opened 2023-12-13
Properly dedent doc strings after parsing in the proc-macro

### LINKED ISSUE #1884: Properly strip/dedent parsed comments
Initially we took the raw comments in proc-macro code, which means the leading spaces (and trailing) are included.
That's not so much a problem for printing them, because we do the dedent-indent dance.
But since #1878 we now trim all input, which removes too much.

Let's add a proper test case and fix this.

badboy:
Fixes #1884

---

I changed both docstring tests. We can unify those tests in a followup.

### DIFF
--- a/fixtures/docstring-proc-macro/tests/bindings/test_docstring.py
+++ b/fixtures/docstring-proc-macro/tests/bindings/test_docstring.py
@@ -36,7 +36,13 @@ assert AssociatedErrorTest.Test.__doc__ == "<docstring-associated-error-variant>
 assert AssociatedErrorTest.Test2.__doc__ == "<docstring-associated-error-variant-2>"
 
 # Test objects
-assert ObjectTest.__doc__ == "<docstring-object>"
+expected_object_doc = """
+    <docstring-object>
+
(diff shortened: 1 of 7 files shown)


fixtures/docstring-proc-macro/tests/bindings/test_docstring.py
```diff
@@ -36,7 +36,13 @@
 assert AssociatedErrorTest.Test2.__doc__ == "<docstring-associated-error-variant-2>"
 
 # Test objects
-assert ObjectTest.__doc__ == "<docstring-object>"
+expected_object_doc = """
```
mhammond, line 39: it wasn't immediately clear to me why this was expected. Can you please see if a test in fixtures/procmacros isn't too difficult, and there you can test the string without the language-specific indents added by languages, and also comment about exactly why the value is what it is (ie, describe in words the intent of the dedent). Ideally here you'd also explain the same (but in this case, IIUC, Python has added an indent to match the indent of the object definition in the generated .py code?)

mhammond (changes requested):

PR #1901, opened 2023-12-13
Don't invoke the RustFutureContinuationCallback while holding the mutex

bendk:
As described in the comments, this could lead to a deadlock if the foreign code immediately polled the future.  This doesn't currently happen with any of the core bindings, but it might with other bindings and it seems like a footgun in general.

To accomplish this, I made a separate class to handle the state and moved the mutex to only wrap that state.  The state mutation happens with the mutex locked and the callback invocation happens after it's unlocked.


bendk:
Actually, let's not do this.

I ran into this issue when working on #1837 and I believe this change fixes things on our side.  However, I'm still getting deadlocks.  I believe the issue is that the futures themselves don't expect to be polled while they are calling the waker function and will deadlock if that happens.

Instead of fixing this in code, I'll just update the docs in that branch to point this out.

PR #1902, opened 2023-12-14
Make foreign trait implementations opt-in (#1827)

bendk:
This changes the default behavior for exported trait interfaces to not generate a foreign callback interface implementation for them.  Users can opt-in to this using the `TraitWithCallbackInterface` attribute for UDL and the `with_callback_interface` for proc-macros.

The main reason for this change is that callback interfaces and trait interfaces aren't completely compatible.  For example, `ByRef` is not supported for callback interfaces, but it is supported for trait interfaces that only have Rust implementations.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
@@ -75,7 +75,7 @@ class {{ impl_name }}:
 {%      endmatch %}
 {% endfor %}
 
-{%- if obj.is_trait_interface() %}
+{%- if obj.has_callback_interface() %}
 {%- let callback_handler_class = format!("UniffiCallbackInterface{}", name) %}
 {%- let callback_handler_obj = format!("uniffiCallbackInterface{}", name) %}
 {%- let ffi_init_callback = obj.ffi_init_callback() %}
@@ -83,7 +83,7 @@ class {{ impl_name }}:
 {%- endif %}
 
 class {{ ffi_converter_name }}:
-    {%- if obj.is_trait_interface() %}
+    {%- if obj.has_callback_interface() %}
     _handle_map = ConcurrentHandleMap()
     {%- endif %}
 
@@ -93,7 +93,7 @@ class {{ ffi_converter_name }}:
 
     @staticmethod
     def check_lower(value: {{ type_name }}):
-        {%- if obj.is_trait_interface() %}
+        {%- if obj.has_callback_interface() %}
         pass
         {%- else %}
         if not isinstance(value, {{ impl_name }}):
@@ -102,14 +102,13 @@ class {{ ffi_converter_name }}:
 
     @staticmethod
     def lower(value: {{ protocol_name }}):
-        {%- match obj.imp() %}
-        {%- when ObjectImpl::Struct %}
+        {%- if obj.has_callback_interface() %}
+        return {{ ffi_converter_name }}._handle_map.insert(value)
+        {%- else %}
         if not isinstance(value, {{ impl_name }}):
             raise TypeError("Expected {{ impl_name }} instance, {} found".format(type(value).__name__))
         return value._uniffi_clone_pointer()
-        {%- when ObjectImpl::Trait %}
-        return {{ ffi_converter_name }}._handle_map.insert(value)
-        {%- endmatch %}
+        {%- endif %}
 
     @classmethod
     def read(cls, buf: _UniffiRustBuffer):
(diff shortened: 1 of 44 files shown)


bendk:
I think this is the simplest way to fix the issue and making a distinction between these two things will be good going forward.  Totally open to bikeshedding on the attribute names.

mhammond:
This patch is fantastic, loads of tests and exactly what needs to be done here, thanks!

And sorry for the slow followup to your question in #1827, but as we discussed, I'm allowed to bike-shed names etc :)

For the UniFFI consumer: I think we should move away from "callback". That's not really accurate for many use-cases with generic traits. Maybe just "foreign"? I propose for UDL:
```
[ForeignTrait]
interface SimCard { ... };
```

and something like `#[uniffi::export(with_foreign)]` for procmacros? There's no good reason to be consistent with `#[uniffi::export(callback_interface)]`; that should fade into obscurity and maybe formally deprecated - callbacks and procmacros are kinda "recent" anyway and I suspect not widely used - @jplatte, do you use them?

Internally: "callback" seems fine, "foreign" might marginally be better for consistency, whatever :) But it still seems very verbose - what about if you dropped the word "interface" everywhere if comes after "callback" in this patch - eg, all CallBackInterface/callback_interface/CALLBACK_INTERFACE etc just need "callback"? (or "foreign" ;) 

Internally is just nit-picking though, the "public" spelling is the important bit. I'd like to see other bike-sheds :)

bendk:
> For the UniFFI consumer: I think we should move away from "callback". That's not really accurate for many use-cases with generic traits. Maybe just "foreign"?

This sounds good to me.  The one tweak I'd suggest is with the UDL keyword.  `[ForeignTrait]` could imply that it's foreign-only.  I'd rather something that feels more additive.  What about `[AllowForeign]` or `[WithForeign]`?

mhammond:
Sounds great! I think I mildly prefer `with` over `allow` - the latter tends to imply some *other* action is required to actually enable it. Regardless though, it would be obviously ideal if procmacros and udl attributes used identical terminology.

mhammond:
OTOH, I'm not sure plain-old `[WithForeign]` is perfect stand-alone in a UDL as there's no indication it's a trait - were you thinking `[Trait, WithForeign]`? If not, maybe `[TraitWithForeign]`?

bendk:
> OTOH, I'm not sure plain-old `[WithForeign]` is perfect stand-alone in a UDL as there's no indication it's a trait - were you thinking `[Trait, WithForeign]`? If not, maybe `[TraitWithForeign]`?

Good point.  I think either would work, but I'm currently liking `[Trait, WithForeign]` the best.

uniffi_macros/src/export/trait_interface.rs
```diff
@@ -108,11 +119,36 @@ pub(super) fn gen_trait_scaffolding(
...
+        }
+    } else {
+        quote! {
+            fn try_lift(v: Self::FfiType) -> ::uniffi::deps::anyhow::Result<::std::sync::Arc<Self>> {
+                unsafe {
+                    Ok(*::std::boxed::Box::from_raw(v as *mut ::std::sync::Arc<Self>))
+                }
+            }
+        }
+    };
```
skeet70, line 146: This is definitely the part I wasn't going to figure out how to do :P.

skeet70 (approved):

skeet70:
Any of the names seem like they work to me. The one that'd make it the clearest to me would be `[Trait, AllowForeignImpl]` and `export(allow_foreign_impl)`.

bendk:
I ended up going with `[WithForeign]`.  I don't think there's a big difference between the two, but it's shorter and also  users might get confused and think that `[AllowForeignImpl]` would not have an effect if they didn't actually implement anything on the foreign side.

jplatte:
> callbacks and procmacros are kinda "recent" anyway and I suspect not widely used - @jplatte, do you use them?

Not sure if this Q is still relevant but yes, `#[uniffi::export(callback_interface)]` is used by the Matrix Rust SDK (though I am no longer working on it much as my employment at Element is ending).

uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
```diff
@@ -1,3 +1,4 @@
+# IMP: {{ "{:?}"|format(obj.imp()) }}
```
mhammond, line 1: nit: either remove this line or make it a little less terse :)

mhammond (approved):
:shipit:

PR #1903, opened 2023-12-14
Implement LiftRef for str and [T]

bendk:
This is needed to allow these to work with proc-macros.

This is not needed for UDL, because there we generate a scaffolding function than inputs `&String`, then pass that value to the Rust function than inputs `&str`.  Rust can handle that conversion automatically.  However, with proc-macros the scaffolding function types are always exactly the same as the source function types.

### DIFF
--- a/fixtures/proc-macro/src/lib.rs
+++ b/fixtures/proc-macro/src/lib.rs
@@ -270,4 +270,9 @@ fn get_externals(e: Option<Externals>) -> Externals {
     e.unwrap_or_default()
 }
 
+#[uniffi::export]
+pub fn join(parts: &[String], sep: &str) -> String {
+    parts.join(sep)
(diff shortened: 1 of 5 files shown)


fixtures/proc-macro/src/lib.rs
```diff
@@ -270,4 +270,9 @@ fn get_externals(e: Option<Externals>) -> Externals {
     e.unwrap_or_default()
 }
 
+#[uniffi::export]
+pub fn join(parts: &[String], sep: &str) -> String {
```
bendk, line 274: Sadly, still no support for `&[&str]`, that would require a lot more work I think.

badboy (approved):

jplatte (approved):
Nice

PR #1904, opened 2023-12-15
Be consistent with variable name prefixes

bendk:
We've used a bunch of different prefix strategies to prevent our variable names from conflicting with the user's.  Let's try to make things consistent by using `uniffi`, `UNIFFI`, or `Uniffi`.

The main difference is removing the leading `_` character, which seems redundant.


bendk:
I'd also be happy to always use the `_` on Python.   There's some logic to that too.  My main goal is to make things consistent.

bendk:
Let's make this more module.  I just opened 3 new PRs that each focus on one language.  Let's use those instead of this one.

PR #1907, opened 2023-12-20
Be more consistent with Kotlin variable prefixes

bendk:
Let's always use `uniffi`, `Uniffi`, or `UNIFFI`. Before some of the variable names had leading underscores, which seems redundant and not idomatic Kotlin.  My guess is that it was copy and pasted from Python.


bendk:
Just pushed some more changes and added a README.  The new changes add the `Uniffi` prefix to more classes to avoid name collisions.  In particular, `Handle` can easily collide with a user name and should be prefixed.

mhammond (approved):

PR #1908, opened 2023-12-20
Be more consistent with Python variable prefixes

bendk:
Always use `uniffi`, `Uniffi`, or `UNIFFI` for module-level names.  Add a leading underscore for private class members.  I think this is the most Pythonic naming style. Because we define `__all__` for modules, there's no chance that users will accidentally import the UniFFI items with `from module import *`

Before there were lots of leading underscores, but I don't see the point and it doesn't feel Pythonic to me.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/templates/Async.py
+++ b/uniffi_bindgen/src/bindings/python/templates/Async.py
@@ -57,7 +57,7 @@ async def _uniffi_rust_call_async(rust_future, ffi_poll, ffi_complete, ffi_free,
                 break
 
         return lift_func(
-            _rust_call_with_error(error_ffi_converter, ffi_complete, rust_future)
+            _uniffi_rust_call_with_error(error_ffi_converter, ffi_complete, rust_future)
         )
     finally:
         ffi_free(rust_future)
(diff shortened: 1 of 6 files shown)


uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -351,14 +351,14 @@ impl PythonCodeOracle {
             FfiType::Float64 => "ctypes.c_double".to_string(),
             FfiType::RustArcPtr(_) => "ctypes.c_void_p".to_string(),
             FfiType::RustBuffer(maybe_suffix) => match maybe_suffix {
-                Some(suffix) => format!("_UniffiRustBuffer{suffix}"),
-                None => "_UniffiRustBuffer".to_string(),
+                Some(suffix) => format!("UniffiRustBuffer{suffix}"),
+                None => "UniffiRustBuffer".to_string(),
```
badboy, line 355: The underscores were actually _added_ in #1599

bendk: Doh!

I just updated the PR to always use the underscores.  PEP 8 seems to say we don't need them if we use `__all__`, but the autocomplete example from 1599 is pretty convincing to me.

I also updated the names for a few more functions.  `_rust_call` is now `_uniffi_rust_call`

bendk:
I also added a README for the templates that goes over naming.

bendk:
I've been trying to clean up my old branches and finally got around to this one.  There's not that much left in this PR, but I think would still be good to merge.  WDYT?

PR #1909, opened 2023-12-20
Be more consistent with Ruby variable prefixes

bendk:
Let's always use `uniffi`, `Uniffi`, or `UNIFFI`. Before some of the variable names had leading underscores, which seems redundant and not idomatic Kotlin.  My guess is that it comes from copy and pasted Python code.

### DIFF
--- a/uniffi_bindgen/src/bindings/ruby/gen_ruby/mod.rs
+++ b/uniffi_bindgen/src/bindings/ruby/gen_ruby/mod.rs
@@ -265,7 +265,7 @@ mod filters {
     pub fn check_lower_rb(nm: &str, type_: &Type) -> Result<String, askama::Error> {
         Ok(match type_ {
             Type::Object { name, .. } => {
-                format!("({}._uniffi_check_lower {nm})", class_name_rb(name)?)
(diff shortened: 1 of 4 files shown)


bendk:
@saks what do you think?  Like I said, I'm willing to use any naming scheme that makes sense to Ruby programmers.  I'm just hoping to make things consistent.

saks:
> @saks what do you think?  Like I said, I'm willing to use any naming scheme that makes sense to Ruby programmers.  I'm just hoping to make things consistent.

Thanks for reaching out, I'll have a look later today.

saks:
hey @bendk! These changes are looking very reasonable. I agree with the suspicion, those underscore prefixes are indeed were copied from python bindings and are not quite in agreement with ruby naming conventions. Thanks for making these changes. I've double verified that there are no other methods with underscore prefixes and found nothing. :+1:

saks (approved):
LGTM!

bendk:
Thanks for the quick review!  It's great to see you're still following the project.

mhammond (approved):

PR #1912, opened 2023-12-22
Fix docs and typos

Sajjon:
Fix typos in 30+ files, mostly by using the amazing `typos` crate CLI tool: https://github.com/crate-ci/typos 

We should use it part of CI and/or git pre-commit hook! Like I do in [`identified_vec` crate](https://github.com/Sajjon/identified_vec/blob/main/.pre-commit-config.yaml#L2) very convenient!

It is customisable, so I added a file which removes some noise.

Kotlin tests are failing, because I did not have Kotlin installed, I'm installing it now and will re-run locally...


Sajjon:
CI seems to pass, so I guess PR is ready to be merged.

Kotlin tests fail locally:
```sh
/Users/sajjon/Developer/uniffi-rs/target/tmp/unary-result-alias-e56f907d5e23476c/uniffi/unary_result_alias/unary_result_alias.kt:492:73: error: unresolved reference: Pointer
    fun ffi_uniffi_unary_result_alias_rust_future_cancel_void(`handle`: Pointer,
                                                                        ^
/Users/sajjon/Developer/uniffi-rs/target/tmp/unary-result-alias-e56f907d5e23476c/uniffi/unary_result_alias/unary_result_alias.kt:494:71: error: unresolved reference: Pointer
    fun ffi_uniffi_unary_result_alias_rust_future_free_void(`handle`: Pointer,
                                                                      ^
/Users/sajjon/Developer/uniffi-rs/target/tmp/unary-result-alias-e56f907d5e23476c/uniffi/unary_result_alias/unary_result_alias.kt:496:75: error: unresolved reference: Pointer
    fun ffi_uniffi_unary_result_alias_rust_future_complete_void(`handle`: Pointer,_uniffi_out_err: RustCallStatus,
                                                                          ^
test uniffi_foreign_language_testcase_test_kts ... FAILED
```

But perhaps I have not setup Kotlin correctly... just installed it via brew.

```sh
$ kotlinc -version
info: kotlinc-jvm 1.9.22 (JRE 21.0.1)
```

maybe I'm using a too new version or something? I might need to add something to PATH in my `~/.zshrc`...

jplatte:
You may need something like this, I always have to set it before running Kotlin tests (on arch linux):

```sh
export CLASSPATH=/usr/share/java/jna.jar
```

examples/app/ios/xc-universal-binary.sh
```diff
@@ -34,7 +34,7 @@ if [[ -n "${SDK_DIR:-}" ]]; then
   # Assume we're in Xcode, which means we're probably cross-compiling.
   # In this case, we need to add an extra library search path for build scripts and proc-macros,
   # which run on the host instead of the target.
-  # (macOS Big Sur does not have linkable libraries in /usr/lib/.)
+  # (macOS Big Sure does not have linkable libraries in /usr/lib/.)
```
bendk, line 37: This one's not a typo.

bendk (changes requested):
Thanks for doing this!  It's appropriate that I got assigned this one, I think most of those typos are mine :)

I found one change that was not correct, everything else looks great.

examples/app/ios/xc-universal-binary.sh
Sajjon: well spotted! Darn I thought I had vetted them all! Rectified!

weedle2/tests/defs/dom.webidl
```diff
@@ -128,11 +128,11 @@ DocumentType includes ChildNode;
 Element includes ChildNode;
 CharacterData includes ChildNode;
 
-interface mixin Slotable {
+interface mixin Slottable {
```
mhammond, line 131: Can you please tell me a little more about how "typos" in identifiers are determined? eg, I see you had to exclude rondpoint entirely and given this change I'm not sure I like the policy - I'm not sure I want some tool to tell me I can't use the identifiers I want to use in scenarios like this, or in scenarios where I want to use a foreign word for whatever whimsical reason.

(OTOH, it does seem to have done a great job in picking up actual typos in other identifiers, so I'm not trying to say I don't like the value here, I'm just trying to avoid a world where a tool is telling me I can't use what I want to use without editing some other text file I probably forgot even existed)

Sajjon: `Slottable` seems to be the correct term:
https://en.wiktionary.org/wiki/slottable

@mhammond regarding `rondpoint` - it is written in French! https://crates.io/crates/typos-cli works with English, so we get a lot of false negative when using it on French words....

I hope that makes sense and explains why I excluded `rondpoint` ? 

The alternative is to exclude all the French words in `typos` flags in `rondpoint`... but that is not tenable IMO, imagine if another example were added in another language... `_typos.toml` would quickly grow with "noisy" excluded words in other Languages.

mhammond: I understand why you excluded rondpoint and don't dispute `Slottable` is the preferred spelling. What I'm asking though is what if I actually *prefer* "Slotable" for some reason, be it a pun, or due to some other subtle meaning, or what if I choose to use a French word for a similar reason. An interface definition etc isn't a piece of English prose, like comments or docs are.

IOW, I'm trying to work out when a tool is going to fight against the choices I explicitly made.

Sajjon: Right, so if you were to prefer `Slotable` over `Slottable` you would just add it to `extend-identifiers` / `extend-words` in `_typos.toml`. 

And even more granular - inline options is being worked on: https://github.com/crate-ci/typos/issues/316

mhammond: As I mentioned above though, I'm just trying to avoid a world where a tool is telling me I can't use what I want to use without editing some other text file I probably forgot even existed.

mhammond: (and more to the point, for this specific change, it's complaining about something in code we forked - so this specific case should probably be addressed by ignoring the entire weedle2 directory - but it seems a reasonable example of my general concern)

bendk: AFAICT, this PR doesn't add a commit hook, which feels like a good compromise to me.  There's nothing that will stop us from checking in whatever words we want.  However, the config file is there to help out whoever runs the `typos` tool next.  When they do that, they can handle updating the config file.

Sajjon: > (and more to the point, for this specific change, it's complaining about something in code we forked - so this specific case should probably be addressed by ignoring the entire weedle2 directory - but it seems a reasonable example of my general concern)

I've now reverted the changed strings in `weedle2` directory (and ignored in `_typos.toml`)

Sajjon: > AFAICT, this PR doesn't add a commit hook, which feels like a good compromise to me. There's nothing that will stop us from checking in whatever words we want. However, the config file is there to help out whoever runs the `typos` tool next. When they do that, they can handle updating the config file.

Yes exactly, one one person has use the `_typos.toml` so far - me. And if no hook or CI uses it it has no effect. But for anyone interested in running `typos -w` locally, the configuration I've now setup (if you chose to merge PR with the `_typos.toml` included that is) will ensure no future unwanted changes take place.

mhammond: > However, the config file is there to help out whoever runs the typos tool next. When they do that, they can handle updating the config file.

If we stick to occasionally running the typos tool (which I agree has much value!) then it's probably not strictly necessary for them to edit the toml - it doesn't really matter if there are a few false-positives, it seems fine to just eye-ball the output and make a judgement call about what should be fixed and what to ignore. This seems analogous to how I use, eg, spell-checkers in docs I write - it's not important that the tool reports zero errors, so I don't feel the need to add false-positives to the dictionary each time, I just ignore them.

I think I'd be mildly against having CI run this tool - I think it would be bad to have it fail PRs because this tool decided it didn't like something and I had to appease it before a PR was accepted. Obviously having it run in CI and *not* have it fail PRs would just lead to it always being in a failure state, which seems to have little value too.

So if it was not run in CI, and we think it's fine for a human to make a subjective call about what output we care about and what we don't, then do we need that toml at all?

(I'm not trying to block this PR landing because it has some great fixes - I'm just thinking aloud about what place this tool has beyond a great way to occasionally find typos in docs)

Sajjon: Sure! I've removed `_typos.toml`, I will leave a comment outside of this tread in the PR with the values I had there, for posterity (for me or anyone else to copy paste over and run locally - without committing the file again).

Sajjon:
Left out of PR is the `_typos.toml` file I used for these fixes (apart from hyperlink, which I fixed manually):

```toml
[files]
extend-exclude = [
	"examples/rondpoint/", # Code is in French
	"weedle2" # https://github.com/mozilla/uniffi-rs/pull/1912#discussion_r1435347116
]

[default.extend-identifiers]
ot = "ot"

[default.extend-words]
inout = "inout"
Sur = "Sur" # Big Sur

```

bendk (approved):
Looks great to me, let's merge this.

PR #1914, opened 2023-12-24
Improve documentation of macros - relating to `Enum` and `export`

Sajjon:
I believe those statements to be correct - but I really hope I'm wrong!

Merry Christmas!

### DIFF
--- a/docs/manual/src/proc_macro/index.md
+++ b/docs/manual/src/proc_macro/index.md
@@ -35,8 +35,7 @@ uniffi::setup_scaffolding!();
 
 ## The `#[uniffi::export]` attribute
 
-The most important proc-macro is the `export` attribute. It can be used on functions, `impl`
-blocks, and `trait` definitions to make UniFFI aware of them.
+The most important proc-macro is the `export` attribute. It can be used on functions, `impl` blocks, and `trait` definitions to make UniFFI aware of them.
 
 ```rust
 #[uniffi::export]
@@ -91,9 +90,7 @@ trait MyTrait {
 
 ```
 
-
-All owned [builtin types](../udl/builtin_types.md) and user-defined types can be used as arguments
-and return types.
+All owned [builtin types](../udl/builtin_types.md) and user-defined types can be used as arguments and return types.
 
 Arguments and receivers can also be references to these types, for example:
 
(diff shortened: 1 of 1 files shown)


jplatte (commented):
The diff is really hard to read because you re-wrapped a lot of paragraphs at the same time as adding / changing text.

About the actual changes:

- The fact that fields of enum or record types can't be object types directly (w/o `Arc`) is not specific to the proc-macros, that's why it's not documented there
- The fact that you can't have methods on record and enum types directly is also not specific to the proc-macros, hence not specifically called out there

Sajjon:
@jplatte I got s bit keen on fixing incorrect formatting, true made review harder, will refrain from combining changes in future...

Ok but where are those limitations listed? 

I think the proc-macro docs ought at least mention those limitations briefly and link to longer docs.

If I understand it correctly UniFFI first shipped without the macros with UDL file as only binding config, and then later got macro support. Macro feels like the much more convenient way of describing the bindings since it does not lead to a lot of double work. Hence I believe a lot of users will - like me - focus heavily on reading the proc-macro section of the docs. Thus, it would be advantageous to list to very limiting limitations... dont you agree?

mhammond:
I agree with Jonas:
* The docs contain many formatting changes which don't actually change content. While it's subjective, I think some of these changes are for the worse so should probably be reverted.

* While things have evolved from UDL and the docs reflect this, pages like https://mozilla.github.io/uniffi-rs/udl/enumerations.html really are intended to be, and will evolve to be, the canonical source of docs for the type system. This is where those limitations should be documented.

The thing that's probably missing from the docs is a clear explanation that enum and dictionaries are always copied around by value (which hopefully helps to explain why they can't have methods) whereas "objects" are always copied as a pointer (which hopefully helps explain why they can't have data items). But FWIW, we expect enums and dictionaries to grow the ability to hold objects once we land the GC-enabling code for Kotlin, and I can't see why objects can't later grow getters and setters, both of which should help blur the distinction.

jplatte:
Honestly, things being passed by value doesn't seem like a reason not to allow methods IMHO. But that's something that should be discussed in a dedicated issue :)

Sajjon:
I closed this since did not really know how to proceed.

PR #1916, opened 2023-12-27
Update oneshot-uniffi to 0.1.6

bendk:
This brings in the `into_raw` and `from_raw` methods that we may want to use to pass oneshot senders over the FFI.  It still uses my patched version to avoid #1736.


mhammond (approved):
LGTM, but mildly wondering why we didn't fork it inside our repo like we did for weedle

bendk:
> LGTM, but mildly wondering why we didn't fork it inside our repo like we did for weedle

No reason at all, other than I forgot that we did that for weedle.  What do you think about moving the source into here?

bendk:
I'm just going to go ahead and merge this one.  We can vendor in the `uniffi-oneshot` library in another PR if needed.

Issue #1593, opened 2023-06-13
Support for `#[non_exhaustive]` enums


thunderbiscuit:
Our code is using enums in an upstream library that are marked as `#[non_exhaustive]`. We are currently writing wrappers around these to make them work, but I'm wondering if there is interest in eventually supporting them natively in uniffi. Cheers!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-282)
┆Components: UniFFI
┆Epic: Maintenance collector
┆Issue Number: UNIFFI-282
┆Sprint: Backlog

bendk:
I don't think the issue is so much the `non_exhaustive` attribute, but the fact that the enums come from a different crate.  Since the enums are defined in a different crate, we can't derive/define the trait implementations needed to lift/lower them directly.  What I think you would want is some way to define an enum in your own crate, which mirrors the enum in the other crate, derive `uniffi::Enum` on that, and use something like [custom types](https://mozilla.github.io/uniffi-rs/udl/custom_types.html) to have UniFFI convert between the two.

The `non_exhaustive` attribute creates a slight wrinkle there, but AFAICT the bigger issue is that custom types don't support this use case at all.  Is that correct from your POV?  Can you share which enum you're trying to use?  I want to do some work on custom types in the new year, and some concrete use cases would be great.

mhammond:
This simple patch works fine:
```
diff --git a/fixtures/proc-macro/src/lib.rs b/fixtures/proc-macro/src/lib.rs
index d1f299178..54768a835 100644
--- a/fixtures/proc-macro/src/lib.rs
+++ b/fixtures/proc-macro/src/lib.rs
@@ -184,6 +184,7 @@ pub enum MaybeBool {
     Uncertain,
 }
 
+#[non_exhaustive]
 #[repr(u8)]
 #[derive(uniffi::Enum)]
 pub enum ReprU8 {
```

it would probably help is we had a concrete problem to solve here - ideally via a patch to one of our fixtures so we can see exactly what the problem is and have the start of a vehicle to test any possible fixes.

bendk:
> No, for public `enum`s without `non_exhaustive` defining UniFFI types in a downstream crate works perfectly fine, we've done that successfully for quite a while now, and there is not even a `UniffiCustomTypeConverter` implementation necessary.

[...]

Issue #1865, opened 2023-11-26
Make `UniffiTag` consistent for UDL and procmacros


mhammond:
Currently procmacros all have a blanket implementation of `FfiConverter` whereas UDL uses a local "tagged" implementation (ie, `FfiConverter<crate::UniFfiTag>`. This inconsistency causes problems with "external" types - in some cases it is impossible to know whether a given type name is "tagged" or not, and thus impossible to generate working code for these types (eg, #1854) without significant and undesirable work.

I see 3 solutions

# Remove `UniFfiTag` completely.

@bendk and I strongly prefer this option as it would remove a ton of code and complexity. Only problem is that we can't work out how to make this work. #1863 is our attempt, which gets tantalizingly close, but fails with ["custom types"](https://github.com/mozilla/uniffi-rs/blob/24db38ce673cfc5a7d4e83c8f756656aec4c4ff1/examples/custom-types/src/lib.rs#L36-L47) - whether implemented via UDL or [via procmacros](https://github.com/mozilla/uniffi-rs/blob/24db38ce673cfc5a7d4e83c8f756656aec4c4ff1/fixtures/ext-types/proc-macro-lib/src/lib.rs#L165-L186).

In that scenario we can't implement `FfiConverter<Url>` because neither the trait nor the type are defined in our crate. We've tried a few things (eg, trying to use a newtype wrapper etc) but none of them play out.

# Use `UniFfiTag` for procmacro implementations

This is fairly easy. It leaves the complications and duplication in the code but does work. Custom types work fine because the `FfiConverter` implementations are tagged with a local type so don't fall foul of Rust's rules.

The biggest downside here is that it means *all* external types from procmacros must be explicitly/magically imported before they can work (in the exact same way UDL types must today). Eg, when using a type from another crate it will be necessary to:

> uniffi::use_udl_record!(uniffi_sublib, SubLibType);

Before the type can be used. (Obviously we'd change the name of the macro so it's not UDL specific, and we can probably even make it do the actual `use` if we want, but the spelling here isn't the point, it's the fact it's needed at all)

This is unfortunate (and another reason we'd prefer to kill the tag entirely), is a breaking change for external types from procmacros, but has the killer advantage of actually working! Plus it is consistent (ie, the requirements are the same regardless of how the type was implemented)

You can see this in #1864

# Lean into the difference

[...]

PR #1924, opened 2024-01-03
Add support for non-exhaustive enums (#1593)

bendk:
UDL types can now have the `[NonExhaustive]` attribute.  If present, we add a default branch arm to make things compile correctly with non-exhaustive enums.  I didn't want to make this the default behavior, since in the normal case we want missing variants to result in compile errors.

In theory, this also adds the `non_exhaustive` attribute for proc-macros.  However, there's no way to test this because `non_exhaustive` only matters for types defined in remote crates and there's currently no proc-macro support for those.

One slightly silly part of this is the flow of metadata.  It doesn't make much sense as a metadata field, since it doesn't affect bindings generation at all.  However, it needs to be because of the way that data flows from `uniffi_udl` -> `uniffi_meta` -> `uniffi_bindgen` -> `uniffi_macros`.

### DIFF
--- a/docs/manual/src/udl/enumerations.md
+++ b/docs/manual/src/udl/enumerations.md
@@ -18,6 +18,8 @@ enum Animal {
 };
 ```
 
+## Enums with fields
+
 Enumerations with associated data require a different syntax,
 due to the limitations of using WebIDL as the basis for UniFFI's interface language.
 An enum like this in Rust:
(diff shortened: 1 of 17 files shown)


bendk:
@thunderbiscuit does this fix the issue in your codebase?

bendk:
In the issue, I mentioned that our decision from https://github.com/mozilla/uniffi-rs/issues/1865 might affect this.  However, I'm pretty sure this fix will be compatible with any of the options listed there.

mhammond (approved):
LGTM!

PR #1929, opened 2024-01-04
Allow defaults for proc-macro function arguments

bendk:
- Renamed `FieldDefault` to `DefaultValue` and shared it between the field and argument code.
- Updated the function argument metadata to support a default value.
- Added `FnArgAttributeArguments` which is currently exactly the same as `FieldAttributeArguments`.  I figure the repeated code is fine for now and these structs may diverge in the future.
- Changed `expand_export` to return the source item as well.  This allows it to modify the source item and strip away the defaults
- Fixed Python argument checking when there's a default.

### DIFF
--- a/docs/manual/src/proc_macro/index.md
+++ b/docs/manual/src/proc_macro/index.md
@@ -138,6 +138,46 @@ fn do_something(foo: MyFooRef) {
 }
 ```
 
+### Default values
+
+Exported functions/methods can have default values using the `default` argument of the attribute macro that wraps them.
+`default` inputs a comma-separated list of `[name]=[value]` items.
+
+```rust
+#[uniffi::export(default(text = " ", max_splits = None))]
+pub fn split(
+    text: String,
+    sep: String,
+    max_splits: Option<u32>,
+) -> Vec<String> {
+  ...
+}
+
+#[derive(uniffi::Object)]
+pub struct TextSplitter { ... }
+
+#[uniffi::export]
+impl TextSplitter {
+    #[uniffi::constructor(default(ignore_unicode_errors = false))]
+    fn new(ignore_unicode_errors: boolean) -> Self {
+        ...
+    }
+
+    #[uniffi::method(default(text = " ", max_splits = None))]
+    fn split(
+        text: String,
+        sep: String,
+        max_splits: Option<u32>,
+    ) -> Vec<String> {
+      ...
+    }
+}
+```
+
+Supported default values:
+  - String, integer, float, and boolean literals
+  - `None` for Option<T> types
+
 ### Renaming functions, methods and constructors
 
 A single exported function can specify an alternate name to be used by the bindings by specifying a `name` attribute.
@@ -179,8 +219,7 @@ will fail).
 pub struct MyRecord {
     pub field_a: String,
     pub field_b: Option<Arc<MyObject>>,
-    // Fields can have a default value.
-    // Currently, only string, integer, float and boolean literals are supported as defaults.
+    // Fields can have a default values
     #[uniffi(default = "hello")]
     pub greeting: String,
     #[uniffi(default = true)]
--- a/uniffi_macros/src/export.rs
+++ b/uniffi_macros/src/export.rs
@@ -20,6 +20,7 @@ use self::{
     },
 };
 use crate::util::{ident_to_string, mod_path};
+pub use attributes::{DefaultMap, ExportFnArgs, ExportedImplFnArgs};
 pub use callback_interface::ffi_converter_callback_interface_impl;
 
 // TODO(jplatte): Ensure no generics, …
--- a/uniffi_macros/src/export/attributes.rs
+++ b/uniffi_macros/src/export/attributes.rs
@@ -1,12 +1,16 @@
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 
-use crate::util::{either_attribute_arg, kw, parse_comma_separated, UniffiAttributeArgs};
+use crate::{
+    default::DefaultValue,
+    util::{either_attribute_arg, kw, parse_comma_separated, UniffiAttributeArgs},
+};
 
 use proc_macro2::TokenStream;
 use quote::ToTokens;
 use syn::{
+    parenthesized,
     parse::{Parse, ParseStream},
-    Attribute, LitStr, Meta, PathArguments, PathSegment, Token,
+    Attribute, Ident, LitStr, Meta, PathArguments, PathSegment, Token,
 };
 use uniffi_meta::UniffiTraitDiscriminants;
 
@@ -67,10 +71,11 @@ impl UniffiAttributeArgs for ExportTraitArgs {
     }
 }
 
-#[derive(Default)]
+#[derive(Clone, Default)]
 pub struct ExportFnArgs {
     pub(crate) async_runtime: Option<AsyncRuntime>,
     pub(crate) name: Option<String>,
+    pub(crate) defaults: DefaultMap,
 }
 
 impl Parse for ExportFnArgs {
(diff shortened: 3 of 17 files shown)


uniffi_macros/src/export.rs
```diff
@@ -160,3 +167,37 @@ pub fn rewrite_self_type(item: &mut Item) {
         visitor.visit_impl_item_mut(item);
     }
 }
+
+/// Rewrite the source input
+///
+/// This removes any uniffi attributes in the source.
+pub fn rewrite_source_input(mut item: Item) -> Item {
```
bendk, line 174: Maybe this could be merged with https://github.com/mozilla/uniffi-rs/blob/0d58c94cbd2ef63554f3388d03d55984be76bb1f/uniffi_macros/src/lib.rs#L369-L380, but the comment scared me away from doing that.  At a minimum, the `syn::Item` needs to parse correctly before this function could remove the `[uniffi]` attributes.  I'm not sure how that interacts with `rust-analyzer`.

jplatte: Yeah, non-parseable code is exactly why those dummy macros exist. AFAIK rust-analyzer does some fixups to the code it submits to (attribute) proc-macros such that ones that output an empty token stream if the input is un-parseable still emit something meaningful, but I don't think it's possible for that sort of fixup to work for every possible partial input.

bendk: So with this change, if the input is un-parsable we'll output it as-is, which can now include `#[uniffi(default = ..)]` attributes which would fail to compile.

I did a quick test of this by adding some invalid code to an item with default attributes and it seemed okay.  Both `nvim + rust-analyzer` and `cargo build` highlighted the syntax error and didn't complain about the attribute.

bendk: Another test was adding the `#[uniffi(default)]` attributes in an item without the `#[uniffi::export]` attribute surround it it.  That one also seemed okay.  rust-analyzer doesn't complain about the invalid attributes.  `cargo build` fails with `cannot find attribute uniffi in this scope` which isn't super-user friendly but I don't think we can do anything to improve that.

jplatte: > So with this change, if the input is un-parsable we'll output it as-is, which can now include `#[uniffi(default = ..)]` attributes which would fail to compile.

Right, on function arguments there is no alternative like `#[uniffi::default()]` because proc-macro attributes are not supported in that position, AFAIK. I think I was actually planning to add `#[uniffi::default(arg_name = None)]` because of that.

> I did a quick test of this by adding some invalid code to an item with default attributes and it seemed okay. Both `nvim + rust-analyzer` and `cargo build` highlighted the syntax error and didn't complain about the attribute.

The potential problems are around auto-complete and assists not being able to function properly because the IDE doesn't know syntax nodes incomplete input map to. Not about the IDE reporting unexpected errors.

mhammond: Is there anything that needs to change based on this conversation?

mhammond (commented):
The implementation here seems fine, but I'm not sure if the discussion from earlier this year is still ongoing or is asking for further changes?

docs/manual/src/proc_macro/index.md
```diff
@@ -131,6 +131,27 @@ fn do_something(foo: MyFooRef) {
...
+### Default values
+
+Exported functions/methods can have default values using the `[uniffi(default = ...)]` attribute.
+
+```rust
+#[uniffi::export]
+fn split(
+    text: String,
+    #[uniffi(default = " ")]
+    sep=String,
```
jplatte, line 143: Here's a typo by the way

```suggestion
    sep: String,
```

docs/manual/src/proc_macro/index.md
```diff
@@ -131,6 +131,27 @@ fn do_something(foo: MyFooRef) {
...
+    #[uniffi(default = None)]
+    max_splits: Option<u32>,
+} {
```
jplatte, line 146: Oh, and another one 😅 

```suggestion
) {
```

uniffi_macros/src/export.rs
jplatte: If you agree with my initial goal of minimizing errors when syn fails to parse the proc-macro input, yes. In that case the usage should be changed from

```rust
#[uniffi::export]
fn split(
    text: String,
    #[uniffi(default = " ")]
    sep: String,
    #[uniffi(default = None)]
    max_splits: Option<u32>,
) {
    // ...
}

#[uniffi::export]
impl Object {
    fn foo(
        #[uniffi::default(None)]
        param: Option<String>,
    ) -> Bar {
        // ...
    }
}
```

to

```rust
#[uniffi::export(default(sep = " ", max_splits = None))]
fn split(text: String, sep: String, max_splits: Option<u32>) {
    // ...
}

#[uniffi::export]
impl Object {
    #[uniffi::default(param = None)]
    fn foo(param: Option<String>) -> Bar {
        // ...
    }
}
```

bendk: I think I was missing something the first time around, I'll change the syntax to match your proposal.  I personally think it's less intuitive, but having your editor not work is worse.

jplatte: I agree it's a bit less intuitive, but at the same time could be more readable in many cases. Also, I think "having your editor not work" is a bit of an overstatement right? AFAIU this is just about getting rid of an unnecessary / confusing error when the macro can't parse the input.

bendk: Definitely an overstatement, but I was more thinking about your comments about autocomplete.  It feels bad when that stops working for mysterious reasons.

bendk:
I think we need to do some bikeshedding on this one to make sure our attributes are sensible, especially after #1990.

This syntax doesn't feel right to me, because `name` is inside `uniffi::constructor` while `uniffi::default` is on it's own. 

```
#[uniffi::export]
impl Foo {
    #[uniffi::constructor(name = "makeAFoo")]
    #[uniffi::default(bar = 0)]
    fn new(&self, bar: u32) { }
}
```

I'm thinking we should set a policy that all macro arguments for functions live inside the attribute that wraps the function.  For example:

```
#[uniffi::export]
impl Foo {
    #[uniffi::constructor(name = "makeAFoo", default(bar = 0))]
    fn new(&self, bar: u32) { }
}
```

Another option would be to say that each argument needs its own attribute:

```
#[uniffi::export]
impl Foo {
    #[uniffi::constructor]
    #[uniffi::name = "baaaar"]
    #[uniffi::default(bar = 0)]
    fn new(&self, bar: u32) { }
}
```

I think these will be pretty similar in practice, since needing more than 1 argument is rare.  I lean towards the first one because a) it matches the decision from #1990 and b) we don't need to create so many no-op attributes like `constructor` and `method`.

jplatte:
Makes a lot of sense to me. Also means that we have the same amount of attribute nesting between free functions and methods. (should make parsing a tad easier)

bendk:
Just pushed an update for this.  The new syntax is described in the [manual](https://github.com/mozilla/uniffi-rs/blob/20ad204c9ee38519fab64d52856018fcd4a38ebd/docs/manual/src/proc_macro/index.md#default-values).  Iif you misspell an argument name, you'll get errors like [this](https://github.com/mozilla/uniffi-rs/pull/1929/files#diff-e9b6fa659b233381902820a2527ea655e4a4230acd3293cb50e1bd25392a8a34).

Is this ready to merge?

uniffi_macros/src/export/attributes.rs
```diff
@@ -224,25 +236,37 @@ impl ToTokens for AsyncRuntime {
...
-#[derive(Default)]
-pub struct NonExportArgs {
+/// Arguments for function inside an impl block
+///
+/// This stores the parsed arguments for `uniffi::constructor` and `uniffi::method`
+#[derive(Clone, Default)]
+pub struct ExportedImplFnArgs {
     pub(crate) name: Option<String>,
+    pub(crate) defaults: DefaultMap,
 }
```
bendk, line 246: This is almost the same as `ExportFnArgs`, but it doesn't store the async runtime.  I think we can combine these into 1 struct to simplify the code.  Allowing users to specify an async runtime for a single method seems like a nice side benefit as well.  However, let's do that in follow-up PR.

jplatte (approved):
I have only skimmed the code, but tests and docs are looking great!

Issue #1925, opened 2024-01-03
Defaults for optional fields


bendk:
If you have a struct with an optional field, the only valid default is `None`.  We should also allow other values that are valid on the inner type, either wrapped with `Some` or not.  While we're at it, let's also allow `_` placeholders like Rust does.  For example:

```
#[derive(uniffi::Record)]
pub struct HttpClientConfig {
    #[uniffi(default=Some(1_000))]
    connect_timeout: Option<u32>,
    #[uniffi(default=10_000)]
    timeout: Option<u32>,
}
```

jplatte:
We could even support arbitrary expressions by exporting a generated function that instantiates that value. Or in the case of `#[uniffi(default = function())]` where `function` is already exported, generate scaffolding that calls that function to get the default value.

PR #1930, opened 2024-01-04
More macro defaults

bendk:
Allow more default types with proc-macros (#1925)

I implemented everything, except enum variants and also jplatte's suggestion of a function-based default.  I think those two can be done later, especially since they semi-conflict.  When you're parsing it's hard to distinguish a variant ident from a function ident.  I also think function-based defaults might be better for enums anyways, since they would work with associated data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@
   This eliminates panics when the capacity of the vec exceeds `i32::MAX`.
   This can happen with the current Vec implementation when String/Vec sizes approach `i32::MAX` but don't exceed it.
 
+- Proc-macro function/method arguments can now have defaults
+
+- Proc-macro record defaults now support empty vecs and Some values.
+
 ### What's fixed?
  
 - Fixed a memory leak in callback interface handling.
@@ -77,7 +81,6 @@
 - Add support for docstrings via procmacros [#1862](https://github.com/mozilla/uniffi-rs/pull/1862)
   and [in UDL](https://mozilla.github.io/uniffi-rs/udl/docstrings.html)
 - Objects can now be returned from functions/constructors/methods without wrapping them in an `Arc<>`.
-- Proc-macro function/method arguments can now have defaults
 
 [All changes in v0.26.0](https://github.com/mozilla/uniffi-rs/compare/v0.25.3...v0.26.0).
 
--- a/docs/manual/src/proc_macro/index.md
+++ b/docs/manual/src/proc_macro/index.md
@@ -176,7 +176,8 @@ impl TextSplitter {
 
 Supported default values:
   - String, integer, float, and boolean literals
-  - `None` for Option<T> types
+  - `[]` for empty Vecs
+  - `Option<T>` allows either `None` or `Some(T)`
 
 ### Renaming functions, methods and constructors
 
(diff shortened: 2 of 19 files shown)


docs/manual/src/proc_macro/index.md
```diff
@@ -131,6 +131,28 @@ fn do_something(foo: MyFooRef) {
...
+    max_splits: Option<u32>,
+} {
+  ...
+}
+```
+
+Supported default values:
+  - String, integer, float, and boolean literals
+  - `[]` for empty Vecs
+  - `Option<T>` allows either `None` or a value for the inner type
```
jplatte, line 154: Without writing `Some()`? That seems strange to me.

bendk: I was on the fence about that one.  I'll update it to use the `Some()` version.

jplatte: Thanks. I think it makes more sense that way because a builtin Rust feature for argument defaults would also almost definitely work that way.

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@
 - Ability for UDL to use external trait interfaces [#1831](https://github.com/mozilla/uniffi-rs/issues/1831)
 - Add support for docstrings via procmacros [#1862](https://github.com/mozilla/uniffi-rs/pull/1862)
 - Objects can now be returned from functions/constructors/methods without wrapping them in an `Arc<>`.
+- Proc-macro function/method arguments can now have defaults
+- Proc-macros defaults now support an empty vecs and the inner value of an `Option<T>`
```
mhammond, line 33: nit - s/an empty/empty/ - and do we need 2 lines here?

Also, "and the inner value of an `Option<T>`" seems stale? ie, the `Some()` part is now needed.

docs/manual/src/proc_macro/index.md
mhammond, line 154: ditto here - this is slightly stale wrt the "inner" part?

fixtures/proc-macro/src/lib.rs
```diff
@@ -275,4 +273,28 @@ pub fn join(parts: &[String], sep: &str) -> String {
...
+    vec: Vec<bool>,
+    #[uniffi(default=None)]
+    opt_vec: Option<Vec<bool>>,
+    #[uniffi(default = Some(42))]
+    opt_integer: Option<i32>,
+}
+
+/// Also test that defaults work with functions
+#[uniffi::export]
+fn double_with_default(#[uniffi(default = 21)] num: i32) -> i32 {
```
mhammond, line 296: Looking at #1929, this should read `#[uniffi::export(default(num = 21))]`? Indeed, this is already landed. It looks like this PR needs updating?

mhammond (changes requested):
I think this should be rebased from main?

CHANGELOG.md
```diff
@@ -77,7 +81,6 @@
 - Add support for docstrings via procmacros [#1862](https://github.com/mozilla/uniffi-rs/pull/1862)
   and [in UDL](https://mozilla.github.io/uniffi-rs/udl/docstrings.html)
 - Objects can now be returned from functions/constructors/methods without wrapping them in an `Arc<>`.
-- Proc-macro function/method arguments can now have defaults
```
bendk, line 80: I think this got inserted in the wrong place in #1929, so I moved it.

bendk:
Thanks for the review.  This was in serious need of a rebase and the `Option<T>` parts were indeed stale.

The one thing I kept was the two lines in the changelog.  One is for adding default support for functions, the other is for extending the existing support for records.  I updated the language though, hopefully it's more clear now.

mhammond (approved):

PR #1931, opened 2024-01-05
Escape """ in python docstrings

### LINKED ISSUE #1927: Python Docstring can produce invalid code
Docstrings in the python generator can produce invalid code. One scenario I've found is when the comment ends with `"`. However presumably starting with a quote would also produce invalid code.

```rust
/// A comment, example "Test"
```

```python
"""A comment, example "Test""""
```

Python uses triple double quotes around docstrings, and since the rust docstring ends with a double quote the generated python code has four quotes which results in a syntax error.

I believe there are a few ways to resolve this, but I would be curious if there are any preference for which approach.

1. Treat docstirngs that starts/ends with an "invalid" character as multi-line.
```python
"""
A comment, example "Test"
"""
```

2. Insert new a character after the docstring, like ` ` or `.`, `"""A comment, example "Test" """`

Hinton:
Uniffi docstrings can currently produce invalid python code if the rust comment contains double quotes.

To avoid this the python docstring implementation has been adjusted slightly.

* Always use multi-line docstrings.
* Escape `"""` with `\"\"\"`.

The following comment:

```rust
/// This is a docstring with "quotes" in it.
/// It also has a triple quote: """
/// And a even longer quote: """"""
```

Will produce the following python code:

```py
"""
This is a docstring with "quotes" in it.
It also has a triple quote: \"\"\"
And a even longer quote: \"\"\"""
"""
```

I believe this leaves the comments mostly intact which is in my opinion important since it's one of the few parts of the generated code that will be visible to consumer through intellisense. I would have preferred longer sequences of quotes would all be escaped but the code got unwieldy and it should be a rare enough edge case to never actually happen.

Resolves #1927

### DIFF
--- a/fixtures/docstring/tests/bindings/test_docstring.py
+++ b/fixtures/docstring/tests/bindings/test_docstring.py
@@ -10,7 +10,6 @@ from uniffi_docstring import *
 
 # Test function
 assert test.__doc__.strip() == "<docstring-function>"
-print(test_multiline.__doc__.strip())
 assert test_multiline.__doc__.strip() == "<docstring-function-multiline>\n    <second-line>"
 assert test_without_docstring.__doc__ is None
 
--- a/fixtures/docstring/tests/bindings/test_docstring.py
+++ b/fixtures/docstring/tests/bindings/test_docstring.py
@@ -20,7 +20,7 @@ assert EnumTest.__doc__.strip() == "<docstring-enum>"
 # assert EnumTest.ONE.__doc__ == "<docstring-enum-variant>"
 # assert EnumTest.TWO.__doc__ == "<docstring-enum-variant-2>"
 
-assert AssociatedEnumTest.__doc__ .strip()== "<docstring-associated-enum>"
+assert AssociatedEnumTest.__doc__ .strip() == "<docstring-associated-enum>"
 
 # `__doc__` is lost because of how enum templates are generated
 # https://github.com/mozilla/uniffi-rs/blob/eb97592f8c48a7f5cf02a94662b8b7861a6544f3/uniffi_bindgen/src/bindings/python/templates/EnumTemplate.py#L60
--- a/fixtures/docstring/tests/bindings/test_docstring.py
+++ b/fixtures/docstring/tests/bindings/test_docstring.py
@@ -10,6 +10,8 @@ from uniffi_docstring import *
 
 # Test function
 assert test.__doc__.strip() == "<docstring-function>"
+print(test_multiline.__doc__.strip())
+assert test_multiline.__doc__.strip() == "<docstring-function-multiline>\n    <second-line>"
 assert test_without_docstring.__doc__ is None
 
 # Test enums
--- a/fixtures/docstring/tests/bindings/test_docstring.py
+++ b/fixtures/docstring/tests/bindings/test_docstring.py
@@ -10,7 +10,7 @@ from uniffi_docstring import *
 
 # Test function
 assert test.__doc__.strip() == "<docstring-function>"
-assert test_multiline.__doc__.strip() == "<docstring-function-multiline>\n    <second-line>"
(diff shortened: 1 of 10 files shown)


Hinton:
The failing test in https://github.com/mozilla/uniffi-rs/blob/main/fixtures/docstring/tests/bindings/test_docstring.py seems to be caused by the multi line docstrings. A quick solution would be to add `.strip()` to the left side of the asserts, but it feels suboptimal. Would we prefer to change it to string literals, to assert it's indented correctly?

mhammond:
This is looking good to me, thanks!

Yeah, I think the failure is caused by the fact we always generate multi-line docstrings whereas we previously special-cased single-line. I think I'm fine with this change, but I'll also ask @badboy what they think given they were more involved in landing this in the first place and actually use python via uniffi.

I think your patch will handle this, but maybe it's worth adding leading triple-quotes - so we have triple quotes at the start, end and in the middle?

mhammond:
> A quick solution would be to add .strip() to the left side of the asserts, but it feels suboptimal

To elaborate a little more - I believe the existing code would still need to do this if the test actually used multi-line docstrings. I suspect multi-line docstrings will be far more common than single lines, so in some ways forcing that test to become aware of newlines and indents is probably the right thing to do in general.

mhammond:
I think that existing test would fail if one of its hard-coded strings had a \n in it - having a fix for that too would be a bonus tbh.

Hinton:
I explored updating the right side of the asserts but it ended up being quite noisy in my opinion, `"\n    <docstring-function>\n   "`. I personally prefer the `strip` over that.

badboy (approved):
works for me!

fixtures/docstring/tests/bindings/test_docstring.py
```diff
@@ -9,44 +9,45 @@
...
 # Test enums
-assert EnumTest.__doc__ == "<docstring-enum>"
+assert EnumTest.__doc__.strip() == "<docstring-enum>"
 
 # Simple enum variants can't be tested, because `__doc__` is not supported for enums
 # assert EnumTest.ONE.__doc__ == "<docstring-enum-variant>"
 # assert EnumTest.TWO.__doc__ == "<docstring-enum-variant-2>"
 
-assert AssociatedEnumTest.__doc__ == "<docstring-associated-enum>"
+assert AssociatedEnumTest.__doc__ .strip()== "<docstring-associated-enum>"
```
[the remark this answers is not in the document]
Hinton: Nice find, I'm far to used to auto formatting.

fixtures/docstring/tests/bindings/test_docstring.py
```diff
@@ -9,44 +9,45 @@
...
 # Test enums
-assert EnumTest.__doc__ == "<docstring-enum>"
+assert EnumTest.__doc__.strip() == "<docstring-enum>"
 
 # Simple enum variants can't be tested, because `__doc__` is not supported for enums
 # assert EnumTest.ONE.__doc__ == "<docstring-enum-variant>"
 # assert EnumTest.TWO.__doc__ == "<docstring-enum-variant-2>"
 
-assert AssociatedEnumTest.__doc__ == "<docstring-associated-enum>"
+assert AssociatedEnumTest.__doc__ .strip() == "<docstring-associated-enum>"
```
badboy, line 23:
```suggestion
assert AssociatedEnumTest.__doc__.strip() == "<docstring-associated-enum>"
```

badboy: Nope, only the missing space infront of the `=` was fixed. There's still an unnecessary space infront of `.strip`

mhammond: doh - I was looking at the wrong `test_docstring.py` 🙃

Hinton: 🙈

PR #1932, opened 2024-01-06
Adding ADR on UniffiTag handling

bendk:
We've discussed this is various issues, most recently #1865.  I think it's time that we have an ADR and discuss the question directly and by itself.

There's a lot more complex cases, for example structs/enums that contain fields with types from another crate.  I don't think that affects the decision, but please tell me if I'm missing anything here.

### DIFF
--- a/docs/adr/0009-uniffi-tagging.md
+++ b/docs/adr/0009-uniffi-tagging.md
@@ -1,0 +1,166 @@
+# Handling `UniffiTag` and using types from other crates
+
+* Status: proposed
+* Deciders: Uniffi developers
+* Date: 2024-01-05
+
+## Context and Problem Statement
+
+UniFFI's various FFI-related traits `FfiConverter`, `Lift`, `Lower`, `LiftReturn`, `LowerReturn`, etc. all have a generic "tag" parameter.
+The code in `uniffi_macros::setup_scaffolding` generates a zero-sized `crate::UniffiTag` type for use in that parameter in other generated code.
+For example, when lowering a type we use `<#type as Lower<crate::UniffiTag>>::lower`.
+
+The reason for this is the so-called "Orphan rule", which prevents generated code from implemented a remote trait on a remote type.
+If the FFI traits did not have a generic parameter, then it would be impossible to generate code that implements those traits for a type defined in another crate.
+See [ADR-0006 Wrapping types](./0006-wrapping-types.md) for more discussion.
+
+For local types we have a choice: implement only for the local tag (`impl Lift<crate::UniffiTag> for MyType`) or implement it for all tags (`impl<UT> Lift<UT> for MyType)`).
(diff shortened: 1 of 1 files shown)


docs/adr/0009-uniffi-tagging.md
```diff
@@ -0,0 +1,95 @@
...
+
+### [Option 2] Implement FFI traits for all tags for non-remote tags
+
+## Pros and Cons of the Options
+
+### [Option 1] Always implement FFI traits for the local tag only
+
+- Good, because it's more explicit.
+- Good, because it's easier to explain.
+  Telling users "you always have to add `uniffi::use_*!`" to use a type from another crate is easier than saying you sometimes you need it and sometimes you don't.
```
bendk, line 86: This factor seems the most important to me.   I don't want to have to go into the details about the orphan rule, remote types, etc. in order to explain to users if they need `uniffi::use_record!` or not.

bendk:
Pushed an edit to this that:
- Uses `Uuid` and custom type handling as the example.  I think this is closer to a real-world scenario.
- Added an option for removing the tags altogether.

docs/adr/0009-uniffi-tagging.md
```diff
@@ -0,0 +1,166 @@
...
+* Date: 2024-01-05
+
+## Context and Problem Statement
+
+UniFFI's various FFI-related traits `FfiConverter`, `Lift`, `Lower`, `LiftReturn`, `LowerReturn`, etc. all have a generic "tag" parameter.
+The code in `uniffi_macros::setup_scaffolding` generates a zero-sized `crate::UniffiTag` type for use in that parameter in other generated code.
+For example, when lowering a type we use `<#type as Lower<crate::UniffiTag>>::lower`.
+
+The reason for this is the so-called "Orphan rule", which prevents generated code from implemented a remote trait on a remote type.
+If the FFI traits did not have a generic parameter, then it would be impossible to generate an code that implements those traits for a type defined in another crate.
```
jplatte, line 14:
```suggestion
If the FFI traits did not have a generic parameter, then it would be impossible to generate code that implements those traits for a type defined in another crate.
```

jplatte:
I still think that Option 1 would be extremely annoying for larger projects consisting of multiple crates that don't need much or any bindings of independent external libraries. 

Option 3 actually seems very interesting to me but involves a lot of extra work for reducing the boilerplate so I don't know how realistic it is short-term. If somebody wants to look more into boilerplate reduction for that option, I think [`#[serde(remote = "...")]`](https://serde.rs/remote-derive.html) could be an interesting case study.

Overall, especially because I don't have that much time to put into this, I (still) strongly prefer option 2.

mhammond:
We iterated around this and @bendk ended up landing our current "remote types" system, which I think we are all quite happy with!

PR #1936, opened 2024-01-08
Remove uniffi_meta::ErrorMetadata.

mhammond:
This means that we determine whether a name is used as an error by checking if, well, it's used as an error - it's not impacted by how the name is declared. This has been done primarily for future "objects as errors" work where we don't want to force an object to be declared specially to have it be able to be used as an error.

This has a side-effect which should be invisible to most users but is visible in our test suites - some of our test UDL declares an enum with `[Error]` but doesn't actually use that enum as an error. Previously this enum would get error semantics, whereas now it has been necessary to change these fixtures to use the object as an error. This should be invisible to real users because objects declared as being an error are either (a) used as an error or (b) unused, where the actual semantics don't matter.

@bendk I flagged you as reviewer but it can wait until you get the rest of your queue landed etc - it's not at all urgent

### DIFF
--- a/uniffi_bindgen/src/interface/enum_.rs
+++ b/uniffi_bindgen/src/interface/enum_.rs
@@ -94,7 +94,9 @@
 //!
 //! ```
 //! # let ci = uniffi_bindgen::interface::ComponentInterface::from_webidl(r##"
-//! # namespace example {};
+//! # namespace example {
+//! #   [Throws=Example] void func();
+//! # };
 //! # [Error]
 //! # enum Example {
 //! #   "one",
@@ -130,7 +132,9 @@
 //!
 //! ```
 //! # let ci = uniffi_bindgen::interface::ComponentInterface::from_webidl(r##"
-//! # namespace example {};
+//! # namespace example {
+//! #   [Throws=Example] void func();
+//! # };
 //! # [Error]
 //! # interface Example {
 //! #   one();
@@ -496,7 +500,10 @@ mod test {
     #[test]
     fn test_variants() {
         const UDL: &str = r#"
-            namespace test{};
+            namespace test{
+                [Throws=Testing]
+                void func();
+            };
             [Error]
             enum Testing { "one", "two", "three" };
         "#;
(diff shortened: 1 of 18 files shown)


uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -94,7 +94,9 @@
 //!
 //! ```
 //! # let ci = uniffi_bindgen::interface::ComponentInterface::from_webidl(r##"
-//! # namespace example {};
+//! # namespace example {
+//! #   [Throws=Example] void fn();
```
bendk, line 98: NIT: maybe name this `func` or something.  Using a Rust keyword confuses the issue a bit.

uniffi_meta/src/lib.rs
```diff
@@ -308,6 +308,7 @@ pub struct FieldMetadata {
 pub struct EnumMetadata {
     pub module_path: String,
     pub name: String,
+    pub forced_flatness: Option<bool>,
```
bendk, line 311: Not that important, but I wonder if we could avoid adding this field by always writing out the non-flat version of the metadata and ensuring that each variant has 0 fields.

bendk (approved):
I like this a lot.  I had a small question and a nit.  Feel free to go either way with those and merge.

uniffi_meta/src/lib.rs
mhammond: Unfortunately I don't think we can, for the same reason this must be `Option<bool>` and not a plain bool. In some cases where this is `Some(true)` the variant *is not* flat and we still need the variants written. `Some(true)` is used just by foreign bindings and says "even though this isn't flat, use the same capitalization and other conventions etc as though it was"

Let me know if this doesn't make sense or if I'm confused!

bendk: That makes sense to me.  I don't think your confused, but I still wonder if there's a better way to handle it.  That doesn't really concern this PR though and shouldn't block it.

PR #1940, opened 2024-01-12
Removing the extern-rustbuffer feature

bendk:
I believe this was only used when integrating UniFFI in the mozilla-central reposition.  However, now that
https://bugzilla.mozilla.org/show_bug.cgi?id=1872200 is implemented it's not used there either.


mhammond (approved):
:rubber_stamp

PR #1941, opened 2024-01-13
Allow enum discriminant type to be specified

mhammond:
Towards #1792 we should record the type of enum discriminants - python doesn't care, but others do.


bendk (approved):
Very nice, this looks great.

PR #1942, opened 2024-01-14
Remove the Kotlin FFIObject base class.

mhammond:
Each interface now includes the functionality directly in the implementation.

This is towards supporting interfaces as errors; such interfaces will need to derive from `Exception` and Kotlin/Java don't allow multiple base-classes.


uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -1,25 +1,129 @@
+// This template implements an class for working with a Rust struct via a Pointer/Arc<T>
+// As such it implements core operations for working with the Rust `Arc<T>` and the
+// Kotlin Pointer to work with the live Rust struct on the other side of the FFI.
+//
+// There's some subtlety here, because we have to be careful not to operate on a Rust
+// struct after it has been dropped, and because we must expose a public API for freeing
+// the Kotlin wrapper object in lieu of reliable finalizers. The core requirements are:
+//
+//   * Each `FFIObject` instance holds an opaque pointer to the underlying Rust struct.
```
bendk, line 9: Nit: `FFIObject` no longer exists.  Maybe this could just say "instance".

bendk (approved):
Looks good, other than the documentation nit.

This feels slightly risky to me when we're about to release 0.26.  Maybe we should wait to merge until after?

PR #1943, opened 2024-01-16
Update goblin to 0.8

glandium:


glandium:
I don't think these errors are coming from this change.

badboy:
Rekicked the tests, indeed unrelated failure.

bendk (approved):
Looks good to me, let's merge this.

PR #1944, opened 2024-01-17
Generate underyling enum variant literals for swift and kotlin

skhamis:
Part of #1792 

I started with proc macros as trying to get the types in UDL is painful (without trying to hackily do stuff like putting the type into the string, etc). I still would like to write up some documentation on this but wanted to get the PR up for early eyes!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,11 @@
 
 [All changes in v0.26.1](https://github.com/mozilla/uniffi-rs/compare/v0.26.0...v0.26.1).
 
+### What's new?
+
+- Enums created with proc macros can now produce literals for variants in Kotlin and Swift bindings specifically for unsigned types.. See
+[the section on enum proc-macros](https://mozilla.github.io/uniffi-rs/proc_macro/index.html#the-uniffienum-derive) for more information.
+
 ## v0.26.0 (backend crates: v0.26.0) - (_2024-01-23_)
 
 ### What's changed?
--- a/docs/manual/src/proc_macro/index.md
+++ b/docs/manual/src/proc_macro/index.md
@@ -210,9 +210,66 @@ pub enum MyEnum {
 }
 ```
 
+### Variant Discriminants
+
 Variant discriminants are accepted by the macro but how they are used depends on the bindings.
-Most would be likely ignore it in the example above due to the nature of the enum,
-but some expose it for simple "unit" enums.
+
+Currently only `unsigned integer` variant discriminants are passed through to the foreign bindings. For example this enum:
+
+```rust
+#[derive(uniffi::Enum)]
+pub enum MyEnum {
+    Foo = 3,
+    Bar = 4,
+}
+```
+
+would give you in Kotlin & Swift:
+
+```swift
+// kotlin
+enum class MyEnum {   
+    FOO,
+    BAR;
+    companion object
+}
+// swift 
+public enum MyEnum {    
+    case foo
+    case bar
+}
+```
+
+which means you cannot use the platforms helpful methods like `value` or `rawValue` to get the underlying discriminants. Adding a `repr` will allow the type to be defined in the foreign bindings.
+
+For example:
+
+```rust
+// added the repr(u8), also u16 -> u64 supported
+#[repr(u8)]
+#[derive(uniffi::Enum)]
+pub enum MyEnum {
+    Foo = 3,
+    Bar = 4,
+}
+```
+
+will now generate:
+
+```swift
+// kotlin
+enum class MyEnum(val value: UByte) {
+    FOO(3u),
+    BAR(4u);
+    companion object
+}
+
+// swift
+public enum MyEnum : UInt8 {
+    case foo = 3
+    case bar = 4
+}
+```
 
 ## The `uniffi::Object` derive
 
--- a/fixtures/enum-types/Cargo.toml
+++ b/fixtures/enum-types/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "uniffi-fixture-enum-types"
+version = "0.22.0"
+authors = ["Firefox Sync Team <sync-team@mozilla.com>"]
+edition = "2021"
+license = "MPL-2.0"
+publish = false
+
+[lib]
+crate-type = ["lib", "cdylib"]
+name = "enum_types"
+
+[dependencies]
+uniffi = { workspace = true }
+thiserror = "1.0"
+
+[build-dependencies]
+uniffi = { workspace = true, features = ["build"] }
+
+[dev-dependencies]
+uniffi = { workspace = true, features = ["bindgen-tests"] }
--- a/fixtures/enum-types/src/enum_types.udl
+++ b/fixtures/enum-types/src/enum_types.udl
@@ -1,0 +1,7 @@
+namespace enum_types {};
+
+
+enum Animal {
+  "Dog",
+  "Cat"
+};
--- a/fixtures/enum-types/src/lib.rs
+++ b/fixtures/enum-types/src/lib.rs
@@ -1,0 +1,41 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+enum Animal {
+    Dog,
+    Cat,
+}
+
+// Though it has the proc-macro, we drop the variant
+// literals if there is not a repr type defined
+#[derive(uniffi::Enum)]
+pub enum AnimalNoReprInt {
+    Dog = 3,
+    Cat = 4,
+}
+
+#[repr(u8)]
+#[derive(uniffi::Enum)]
+pub enum AnimalUInt {
+    Dog = 3,
+    Cat = 4,
+}
+
+#[repr(u64)]
+#[derive(uniffi::Enum)]
+pub enum AnimalLargeUInt {
+    Dog = 4294967298, // u32::MAX as u64 + 3
+    Cat = 4294967299, // u32::MAX as u64 + 4
+}
+
+// Signed is currently NOT supported but included to ensure we don't break
+// things for current users of the enum proc-macro
+#[repr(i8)]
+#[derive(uniffi::Enum)]
+pub enum AnimalSignedInt {
+    Dog = -3,
+    Cat = -4,
+}
+
+uniffi::include_scaffolding!("enum_types");
--- a/fixtures/enum-types/tests/bindings/test_enum_types.py
+++ b/fixtures/enum-types/tests/bindings/test_enum_types.py
@@ -1,0 +1,18 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import unittest
+from enum_types import *
+
+assert(Animal.DOG.value == 0)
+assert(Animal.CAT.value == 1)
+
+assert(AnimalNoReprInt.DOG.value == 3)
+assert(AnimalNoReprInt.CAT.value == 4)
+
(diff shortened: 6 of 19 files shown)


fixtures/enum-types/src/enum_types.udl
```diff
@@ -0,0 +1,7 @@
+namespace enum_types {};
+
+
+enum Animal {
+  "Dog",
+  "Cat"
+};
```
mhammond, line 7: nit: add trailing \n (here and elsewhere)

fixtures/enum-types/src/lib.rs
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+enum Animal {
+    Dog,
+    Cat,
+}
+
+#[repr(u8)]
```
mhammond, line 10: If we are going to have a whole fixture for this, we might as well exercise more types, so maybe an enum with a values but no repr, an explicit signed repr (eg, i8), and an enum with a, say, usize repr (which you'd probably use an unsigned int in bidnings? )Even if some of them turn out to be problematic and ends up being done in a followup it might be worth having the basics here with a link to followup issues or similar?

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -8,13 +8,24 @@
 {%- if e.is_flat() %}
 
 {%- call kt::docstring(e, 0) %}
+{% match e.variant_discr_type() %}
+{% when None %}
```
mhammond, line 12: re more examples above, here it appears that if an enum has literals for each variant but no discriminant type then we just drop the variant literals. That may or may not be the right thing to do, but explicitly spelling it out in the fixture (and the eventual docs) would help make this clearer.

mhammond (commented):
This looks great in general, thanks! I agree docs are also needed, but it's probably worth fleshing the fixture out a little more so it's clearer what exactly the docs should say.

fixtures/enum-types/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+edition = "2021"
+license = "MPL-2.0"
+publish = false
+
+[lib]
+crate-type = ["lib", "cdylib"]
+name = "enum_types"
+
+[dependencies]
+uniffi = {path = "../../uniffi", version = "0.25" }
```
mhammond, line 14: FYI, #1859 just landed, so these uniffi references should change from `{ path = ..., version = ... }` to `{ workspace = true }` - check the other fixtures for examples.

fixtures/enum-types/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[repr(u64)]
+#[derive(uniffi::Enum)]
+pub enum AnimalLargeUInt {
+    Dog = 123456789,
+    Cat = 987654321,
+}
+
+
+// TODO: Everything is treated as unsigned right now so
```
mhammond, line 34: I think we should uncomment this as otherwise we risk a breaking change - IIUC, this enum probably *does* work today, but without the correct discr being used by bindings. IIUC, this patch will mean the enum doesn't work at for kotlin?

Similarly, maybe add an enum with `#[repr(C, u8)]` (actually, while [the docs](https://doc.rust-lang.org/reference/type-layout.html#reprc-enums-with-fields) say it works, it now generates a warning, so I guess ignore this :)

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -511,6 +512,17 @@ mod filters {
         Ok(as_ct.as_codetype().literal(literal, ci))
     }
 
+    // Get the idiomatic Kotlin rendering of an individual enum variant's discriminant
+    pub fn variant_discr_literal(e: &Enum, index: &usize) -> Result<String, askama::Error> {
+        let literal = e.variant_discr(*index).expect("invalid index");
+        match literal {
+            // Kotlin doesn't convert between signed and unsigned by default
+            // so we'll need to make sure we define the type as unsigned
+            LiteralMetadata::UInt(v, _, _) => Ok(v.to_string() + "u"),
```
mhammond, line 521: [I think that even a signed repr is going to end up with a UInt here](https://github.com/mozilla/uniffi-rs/blob/26a2fef271e78e9be3cfa5385d96e2c130eaf219/uniffi_meta/src/reader.rs#L425) - ie, the type of each value can't really be trused, only the type on the enum's discr can.

fixtures/enum-types/tests/bindings/test_enum_types.py
```diff
@@ -0,0 +1,14 @@
+import unittest
+from enum_types import *
+
+assert(Animal.DOG.value == 0)
+assert(Animal.CAT.value == 1)
+
+assert(AnimalInt.DOG.value == 3)
```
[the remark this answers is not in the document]
skhamis: It won't work for kotlin or swift here since there is no `Repr` (which ty for suggestion, I'll rename this enum so it's more obvious) and those languages don't default to int for the underylings. One option is to default everything to unsigned if not defined but that could have breaking implications for people using the enum proc-macro right now (if not using unsigned literals).

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
skhamis: yes! That's what makes it slightly broken for signed right now on kotlin specifically (swift and python is fine) but not sure if we want to just have kotlin left out (if that's even possible to detect since it goes through the UInt route still).

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -510,6 +511,29 @@ mod filters {
...
+    // Get the idiomatic Kotlin rendering of an individual enum variant's discriminant
+    pub fn variant_discr_literal(e: &Enum, index: &usize) -> Result<String, askama::Error> {
+        let literal = e.variant_discr(*index).expect("invalid index");
+        match literal {
+            // Kotlin doesn't convert between signed and unsigned by default
+            // so we'll need to make sure we define the type as appropriately
+            LiteralMetadata::UInt(v, _, _) => {
+                if let Some(t) = e.variant_discr_type() {
+                    match t {
+                        Type::Int8 | Type::Int16 | Type::Int32 | Type::Int64 => Ok(v.to_string()),
```
skhamis, line 523: @mhammond curious your thoughts on this, I know it's a bit "hacky" until we get full support for signed integers but prevents current users from breaking when using the proc-macro.

CHANGELOG.md
```diff
@@ -18,6 +18,11 @@
   See https://github.com/mozilla/uniffi-rs/pull/1818 for details and how the other bindings were updated.
 - Removed `FfiType::RustFutureContinuationData`.  `RustFutureContinuation` callbacks now take a `u64` handle.
 
+### What's new?
+
+- Enums using proc-macros can now generate the variant literals for the foreign bindings. See
```
mhammond, line 23: I think this needs updating - I think "all" this PR is doing is exposing these variants to Swift and Kotlin?

docs/manual/src/proc_macro/index.md
```diff
@@ -185,9 +185,66 @@ pub enum MyEnum {
 }
 ```
 
+### Variant Discriminants
+
 Variant discriminants are accepted by the macro but how they are used depends on the bindings.
-Most would be likely ignore it in the example above due to the nature of the enum,
-but some expose it for simple "unit" enums.
+
+Currently only `unsigned integer` variant discriminants are passed through to the foreign bindings. For example this enum:
```
mhammond, line 192: nit: s/the foreign bindings/Swift and Kotlin/?

fixtures/enum-types/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+#[derive(uniffi::Enum)]
+pub enum AnimalUInt {
+    Dog = 3,
+    Cat = 4,
+}
+
+#[repr(u64)]
+#[derive(uniffi::Enum)]
+pub enum AnimalLargeUInt {
+    Dog = 123456789,
```
mhammond, line 28: Might as well make this number greater than an i32 can hold?

fixtures/enum-types/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+    Cat = 987654321,
+}
+
+// Signed is currently NOT supported but included to ensure we don't break
```
mhammond, line 32: as above, something like "by all bindings" or similar?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -510,6 +511,29 @@ mod filters {
...
+                        Type::UInt8 | Type::UInt16 | Type::UInt32 | Type::UInt64 => {
+                            Ok(v.to_string() + "u")
+                        }
+                        _ => unreachable!("only unsigned is supported at this time"),
+                    }
+                } else {
+                    unreachable!("enum hasn't defined a repr")
+                }
+            }
+            _ => unreachable!("expected an UInt!"),
```
mhammond, line 533: If you made [this struct](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/backend/filters.rs#L16) pub and imported it, you could maybe change these to `return askama::Error::Custom(Box::new(UniFFIError::new("mesage...")))`

(this and the one unreachable above - the first unreachable here is fine but maybe with a different message as I wrote above)

fixtures/enum-types/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+// things for current users of the enum proc-macro
+#[repr(i8)]
+#[derive(uniffi::Enum)]
+pub enum AnimalSignedInt {
```
mhammond, line 36: I think all tests should reference this. I think for Python it will work correctly. For kotlin and swift it should represent how it *does not* work - ie, somehow that `.value` doesn't exist for this enum.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
mhammond: I don't quite understand this - it looks like this is going to write the signed value to the generated kotlin as unsigned - ie, that in the Kotlin test, when we add the test for `AnimalSignedInt` we'd find it written as though it was unsigned. If that's true, that's not really what I would expect. If it's false I don't quite understand what it would be doing here.

uniffi_bindgen/src/bindings/swift/gen_swift/mod.rs
```diff
@@ -607,6 +608,15 @@ pub mod filters {
         Ok(oracle().find(&as_type.as_type()).literal(literal))
     }
 
+    // Get the idiomatic Swift rendering of an individual enum variant's discriminant
+    pub fn variant_discr_literal(e: &Enum, index: &usize) -> Result<String, askama::Error> {
+        let literal = e.variant_discr(*index).expect("invalid index");
+        match literal {
```
mhammond, line 614: Looking at this I don't understand how Swift rejects signed reprs?

uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -235,6 +235,10 @@ impl Enum {
         Ok(Literal::new_uint(this))
     }
 
+    pub fn variant_discr_type(&self) -> Option<Type> {
+        self.discr_type.clone()
```
mhammond, line 239: I think would work if you make this return `&Option<Type>` and avoid the clone, because the lifetime of the result should be the same as the `e` you are calling it on.

mhammond (commented):
Looking great!

mhammond (approved):
Thanks Sammy!

PR #1948, opened 2024-01-22
Reject [uniffi::export()] arguments used in invalid contexts.

mhammond:
Allows for better UX because attributes which don't make sense in a particular context are rejected rather than ingored.

Replaces ExportAttributeArguments with multiple structs, one per export context (ie, a struct, fn, trait, etc)

Also removes the undocumented `uniffi::export(constructor)` attribute on an impl block, just use `uniffi::constructor` on each fn.

### DIFF
--- a/fixtures/uitests/tests/ui/export_attrs.stderr
+++ b/fixtures/uitests/tests/ui/export_attrs.stderr
@@ -1,0 +1,25 @@
+error: uniffi::export on a struct must supply a builtin trait name. Did you mean `#[derive(uniffi::Object)]`?
+ --> tests/ui/export_attrs.rs:5:1
+  |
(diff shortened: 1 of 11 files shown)


jplatte (approved):
Have only skimmed the changes but got a very positive impression from that.

fixtures/uitests/tests/ui/export_attrs.stderr
```diff
@@ -0,0 +1,25 @@
+error: uniffi::export on a struct must supply a builtin trait name. Did you mean `#[derive(uniffi::Object)]`?
```
bendk, line 1: Nice!

bendk (approved):
Looks great to me, I think this will definitely help out some users.

PR #1951, opened 2024-01-22
Update min supported JNA version in docs

jmartinesp:
Using JNA's Cleaner API in https://github.com/mozilla/uniffi-rs/pull/1869, which has been merged to main, needs at least JNA `v5.12.0` as we discovered later [here](https://github.com/mozilla/uniffi-rs/pull/1869#issuecomment-1902237867).

When I asked in the UniFFI Matrix room @badboy said the docs should be updated, so this should take care of that.


bendk (approved):
Looks good, thanks for the patch!

PR #1952, opened 2024-01-22
Adding the `TypeId` FFI trait

bendk:
The metadata code now uses this trait to get the `TYPE_ID_METADATA` rather than Lift/Lower.  This feels cleaner, since it was often not clear which trait to use before. This removes the hacky handling of type id metadata for callback interfaces that I added in #1950.

Let's not merge this until after 0.26.0 is released.  It feels like too big of a change at this point in the process.


mhammond (approved):
This does seem cleaner 👍

PR #1959, opened 2024-01-24
Destroy argument buffer after invoking a foreign callback

Lipt0nas:
The argument buffer being passed to the foreign callback is never destroyed, leading to a memory leak every time it's invoked


bendk (approved):
Wow, that was a bad one.  Thanks for the fix.

PR #1960, opened 2024-01-24
Destroy argument buffer after invoking a foreign callback

bendk:
I guess the good thing about making a 0.26.1 release is that we can get this fix in.


mhammond (approved):
A CHANGELOG couldn't hurt here

bendk:
I was thinking the same thing :).  I have a PR for that for main, and I'll also include it in the PR for 0.26.1.

PR #1963, opened 2024-01-24
Allow interfaces to be errors.

mhammond:


mhammond:
The changes here are all in the foreign bindings, so looking for input from kotlin, swift and python. @saks ruby shouldn't break but doesn't get error support.

uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -105,13 +105,21 @@
...
 {%- let interface_docstring = obj.docstring() %}
+{%- let is_error = ci.is_name_used_as_error(name) %}
+{%- let ffi_converter_name = obj|ffi_converter_name %}
 
 {%- include "Interface.kt" %}
 
 {%- call kt::docstring(obj, 0) %}
+{%- let super_init = "" %}
+{% if (is_error) %}
+{%- let super_init = ": super()" %}
```
bendk, line 116: I'm not sure this does anything because of scoping rules and a quick check in the generated kotlin seems to bear that out.  Maybe it's not needed or maybe the above `let` could change to be a ternary expression.

uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -298,6 +307,17 @@ open class {{ impl_class_name }}: Disposable, AutoCloseable, {{ interface_name }
             {{ impl_class_name }}({% call kt::to_ffi_call(cons) %})
         {% endfor %}
     }
+    {% else if is_error %}
+    companion object ErrorHandler : UniffiRustCallStatusErrorHandler<{{ impl_class_name }}> {
+        override fun lift(error_buf: RustBuffer.ByValue): {{ impl_class_name }} {
+            // Due to some mismatches in the ffi converter mechanisms, errors are a RustBuffer.
```
bendk, line 313: I wonder if we could get rid of that restriction at this point -- that can be another follow up.

uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py
```diff
@@ -15,13 +15,7 @@ def {{ func.name()|fn_name }}({%- call py::arg_list_decl(func) -%}):
...
         lambda val: None,
         {% endmatch %}
-        # Error FFI converter
-        {%- match func.throws_type() %}
-        {%- when Some(e) %}
-        {{ e|ffi_converter_name }},
-        {%- when None %}
-        None,
-        {%- endmatch %}
+        {% call py::error_ffi_converter(func) %}
```
bendk, line 18: Does there need to be a similar change for async methods?

bendk (approved):
This is great.  I left a couple comments, but mostly they're just things we can follow up on after merging this.

I mainly looked at the Python code, the only question I have there is error handling in async methods.

It's great how little needs to change to support this.

uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py
mhammond: I don't think so - the test has a top-level `async fn aoops() -> Result<(), Arc<ErrorInterface>>`, but I might have missed something.

fixtures/error-types/src/error_types.udl
```diff
@@ -0,0 +1,31 @@
...
+
+    [Throws=ErrorInterface, Name="fallible_new"]
+    constructor();
+
+    [Throws=ErrorInterface]
+    void oops();
+};
+
+[Traits=(Debug, Display)]
+interface ErrorInterface {
```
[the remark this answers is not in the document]
mhammond: It does :( But for an interesting reason - the generated impl of the trait doesn't actually implement the trait bounds - ie

> the trait `std::error::Error` is not implemented for `UniFFICallbackHandlerErrorTrait`

uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
mhammond: thanks - apparently not needed - probably was in an earlier confused iteration!

uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py
bendk: Ahh, that one is already calling `error_ffi_converter`.  Seems good to me.

mhammond:
In the absense of any other reviews I'm going to land this in the next few days/

PR #1966, opened 2024-01-26
Use iterator in generated Kotlin to maintain Android API level 22 compatibility

richardhuaaa:
The `forEach` here was not introduced until API level 24, this small tweak helps to maintain compatibility.

Tested via `cargo test` as well as in our own project (https://github.com/xmtp/libxmtp).

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/SequenceTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/SequenceTemplate.kt
@@ -16,7 +16,7 @@ public object {{ ffi_converter_name }}: FfiConverterRustBuffer<List<{{ inner_typ
 
     override fun write(value: List<{{ inner_type_name }}>, buf: ByteBuffer) {
         buf.putInt(value.size)
-        value.forEach {
+        value.iterator().forEach {
             {{ inner_type|write_fn }}(it, buf)
         }
     }


mhammond:
This does highlight an omission in our docs. We do talk about [JNA](https://github.com/mozilla/uniffi-rs/blob/main/docs/manual/src/kotlin/gradle.md#jna-dependency) but never mention the min Android API we target. Indeed, I didn't think we explicitly targeted the Android API at all!

We *do* also mention a min kotlin version - what's the relationship between all these things? Assuming we should target an API version, is there a sane way for us to enforce this? IIUC, the min JNA and kotlin versions are set because trying to use anything less simply fails to build - how might we ensure future patches don't accidentally cause us to change the API version?

jmartinesp:
> We do also mention a min kotlin version - what's the relationship between all these things?

So if I'm not mistaken, right now:

- The minimum SDK version supported is `minSdkVersion = 24` (with this change it would be `22`).
- The compile SDK version should be at least `compileSdkVersion = 33` to be able to build the library with the `android` or `android_cleaner` values set to true.
- The target SDK (`targetSdkVersion`) can be set to anything > `minSdkVersion` if I'm not mistaken, as this just states which API level the app will use in runtime.

> how might we ensure future patches don't accidentally cause us to change the API version?

The only thing I can think of is running the same unit tests using an actual Android project/library module, even if it's just a skeleton of it, with these versions set up. So either duplicating the tests or somehow playing with the source sets so those are also included in this android module.

mhammond (approved):
Sorry this slipped.

PR #1971, opened 2024-01-30
Add more errors when #[uniffi] attributes are used incorrectly.

mhammond:
I added some more explicit errors when uniffi attributes are used incorrectly.

I started out seeing if I could do the same for `#[uniffi::constructor]`, but I don't think I can. I updated the comments for this attribute - I *think* the old comment was stale, but I'm really not sure how accurate my new comment is - after updating the uitest I realized I really don't understand things well at all.

@jplatte do you mind checking that the comments I removed were indeed stale, and that the comments I added are at least an improvement? Any insights to make the comments more accurate would obviously be welcome too.

### DIFF
--- a/fixtures/uitests/tests/ui/export_attrs.rs
+++ b/fixtures/uitests/tests/ui/export_attrs.rs
@@ -1,6 +1,6 @@
 fn main() {} /* empty main required by `trybuild` */
 
-#[uniffi::constructor] // <--- should be an error!
+#[uniffi::constructor] // <--- would ideally be an error.
 // Someone might try to 'export' a struct instead of deriving it.
 #[uniffi::export]
 struct S {}
@@ -16,5 +16,23 @@ impl Object {}
 
 #[uniffi::export(with_foreign)]
 fn foreign() {}
+
+#[derive(uniffi::Record)]
+// Records have explicit `#[uniffi()]` handling.
+#[uniffi(flat_error)]
+pub struct One {}
+
+#[derive(uniffi::Record)]
+pub struct Two {
+    #[uniffi(flat_error)]
+    inner: i32,
+}
+
+#[derive(thiserror::Error, uniffi::Error, Debug)]
+pub enum Error {
+    #[error("Oops")]
+    #[uniffi(flat_error)]
+    Oops,
+}
 
 uniffi_macros::setup_scaffolding!();
--- a/uniffi_macros/src/lib.rs
+++ b/uniffi_macros/src/lib.rs
@@ -5,9 +5,6 @@
 #![warn(rust_2018_idioms, unused_qualifications)]
 
 //! Macros for `uniffi`.
-//!
-//! Currently this is just for easily generating integration tests, but maybe
-//! we'll put some other code-annotation helper macros in here at some point.
 
 use camino::Utf8Path;
 use proc_macro::TokenStream;
@@ -369,14 +366,16 @@ pub fn generate_and_include_scaffolding(udl_file: TokenStream) -> TokenStream {
     }.into()
 }
 
-/// A dummy macro that does nothing.
+/// An attribute for constructors.
+///
+/// Constructors are in `impl` blocks which have a `#[uniffi::export]` attribute,
 ///
 /// This exists so `#[uniffi::export]` can emit its input verbatim without
-/// causing unexpected errors, plus some extra code in case everything is okay.
+/// causing unexpected errors in the entire exported block.
+/// This happens very often when the proc-macro is run on an incomplete
+/// input by rust-analyzer while the developer is typing.
 ///
-/// It is important for `#[uniffi::export]` to not raise unexpected errors if it
-/// fails to parse the input as this happens very often when the proc-macro is
-/// run on an incomplete input by rust-analyzer while the developer is typing.
+/// So much better to do nothing here then let the impl block find the attribute.
 #[proc_macro_attribute]
 pub fn constructor(_attrs: TokenStream, input: TokenStream) -> TokenStream {
     input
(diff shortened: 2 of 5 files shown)


fixtures/uitests/tests/ui/export_attrs.rs
```diff
@@ -9,6 +9,9 @@ struct S {}
 struct S2 {}
 
 #[derive(uniffi::Object)]
+// #[uniffi(flat_error)]
+// -> error: cannot find attribute `uniffi` in this scope
+// (Other derives can use them tho?)
```
jplatte, line 14: The difference is in how these derives are defined in `uniffi_macros`:

```rust
#[proc_macro_derive(Record, attributes(uniffi))]
#[proc_macro_derive(Enum)]
#[proc_macro_derive(Object)]
```

uniffi_macros/src/lib.rs
```diff
@@ -369,14 +369,12 @@ pub fn generate_and_include_scaffolding(udl_file: TokenStream) -> TokenStream {
...
-/// A dummy macro that does nothing.
-///
-/// This exists so `#[uniffi::export]` can emit its input verbatim without
-/// causing unexpected errors, plus some extra code in case everything is okay.
-///
-/// It is important for `#[uniffi::export]` to not raise unexpected errors if it
-/// fails to parse the input as this happens very often when the proc-macro is
-/// run on an incomplete input by rust-analyzer while the developer is typing.
+/// An attribute for constructors.
+/// Constructors are in `impl` blocks which have a `#[uniffi::export]` attribute.
```
jplatte, line 373:
```suggestion
/// An attribute for constructors.
///
/// Constructors are in `impl` blocks which have a `#[uniffi::export]` attribute.
```

uniffi_macros/src/lib.rs
```diff
@@ -84,6 +81,9 @@ pub fn setup_scaffolding(tokens: TokenStream) -> TokenStream {
         .into()
 }
 
+/// It is important for `#[uniffi::export]` to not raise unexpected errors if it
+/// fails to parse the input as this happens very often when the proc-macro is
+/// run on an incomplete input by rust-analyzer while the developer is typing.
```
jplatte, line 86: Not really clear to me what "unexpected" error would be.

uniffi_macros/src/util.rs
```diff
@@ -151,11 +151,11 @@ pub fn parse_comma_separated<T: UniffiAttributeArgs>(input: ParseStream<'_>) ->
...
 
 #[derive(Default)]
-pub struct ArgumentNotAllowedHere;
+struct ArgumentNotAllowedHere;
 
 impl Parse for ArgumentNotAllowedHere {
-    fn parse(input: ParseStream<'_>) -> syn::Result<Self> {
-        parse_comma_separated(input)
+    fn parse(_input: ParseStream<'_>) -> syn::Result<Self> {
+        unreachable!();
```
jplatte, line 158: Huh?

uniffi_macros/src/lib.rs
mhammond: yeah, this comment was moved and I wasn't sure either. I didn't check blame but assumed you wrote that. Maybe I should just delete that entire comment?

uniffi_macros/src/util.rs
mhammond: It is unreachable in practice, so I figured it makes sense to call it out as such. I only noticed that because when I was debugging I got a little confused and though this might have been implicated. I would have been less confused with `unreachable()`. I'm fine putting this back though if you think it has value.

uniffi_macros/src/lib.rs
jplatte: Ah, I didn't check the diff carefully enough. Actually the comment makes way more sense to me when placed on `#[uniffi::constructor]`. But I agree I worded it badly. The idea behind `#[uniffi::constructor]` being `#[uniffi::constructor]` is indeed that spurious "attribute not known" / "not supported" errors are avoided when the export macro fails to parse its input and outputs it verbatim as a fallback; then I think it would be confusing for there to be an error about the constructor attribute (in addition to the parse error[s]).

uniffi_macros/src/lib.rs
```diff
@@ -369,14 +369,12 @@ pub fn generate_and_include_scaffolding(udl_file: TokenStream) -> TokenStream {
...
-/// This exists so `#[uniffi::export]` can emit its input verbatim without
-/// causing unexpected errors, plus some extra code in case everything is okay.
-///
-/// It is important for `#[uniffi::export]` to not raise unexpected errors if it
-/// fails to parse the input as this happens very often when the proc-macro is
-/// run on an incomplete input by rust-analyzer while the developer is typing.
+/// An attribute for constructors.
+/// Constructors are in `impl` blocks which have a `#[uniffi::export]` attribute.
+/// Because there's no `derive` for the block, it's impossible to have
+/// `#[uniffi(constructor)]`.
```
jplatte, line 375: This is not true actually. It could be `#[uniffi(constructor)]`, but then the macro output of `#[uniffi::export]` would be invalid when it fails to parse the input and re-emits verbatim it for IDE compatibility.

uniffi_macros/src/lib.rs
```diff
@@ -369,14 +369,12 @@ pub fn generate_and_include_scaffolding(udl_file: TokenStream) -> TokenStream {
...
+/// This means we can't give errors in many cases where it's used incorrectly, so
+/// it's just a noop in those cases.
```
jplatte, line 377: Probably not *that* many, we could put some code here that parses an optional list of extra attributes plus maybe a few tokens afterwards and checks if those tokens are the start of a `fn` or a different item – could error out if the input is a `const` / `static` / type definition rather than a function definition, and maybe even if it's a method (has a `self` argument).

uniffi_macros/src/lib.rs
mhammond: Thanks - that was all very helpful.

jplatte (approved):

PR #1975, opened 2024-02-01
Swift open classes

stefanceriu:
With https://github.com/mozilla/uniffi-rs/pull/1918 in limbo and based on [this comment](https://github.com/mozilla/uniffi-rs/pull/1918#issuecomment-1872161231) I went ahead and extracted out only the bits that open up generated swift classes. 
This shouldn't be controversial and will allow us to mock client side.

Previously done for Kotlin here https://github.com/mozilla/uniffi-rs/pull/1815


mhammond (approved):
This is very consistent with Kotlin and serves the same purpose 👍

PR #1977, opened 2024-02-02
fix: Prevent Vec capacity from triggering panics in RustBuffer

### LINKED ISSUE #1976: Vec capacity can unnecessarily cause Rust panics when creating a `RustBuffer`.
We've run into an issue where passing a large enough `bytes` sequence into Rust will cause a panic in `RustBuffer::from_vec` due to the `Vec`'s capacity exceeding `i32::MAX`. This is despite the length of the `Vec` being less than `i32::MAX`. Rust makes no guarantees about how the capacity of a `Vec` grows, but in our case, it has cut the effective max length by about half. It seems like we should try to shrink the size of the `Vec` before calling `RustBuffer::from_vec` as to prevent what should be valid data from panicking.

MarquessV:
This is a proposal to help fix #1976. We use `Vec.shrink_to` to impose a ceiling on vector capacity that is in line with the max `RustBuffer::from_vec` can handle.

### DIFF
--- a/uniffi_core/src/ffi/rustbuffer.rs
+++ b/uniffi_core/src/ffi/rustbuffer.rs
@@ -230,7 +230,7 @@ pub fn uniffi_rustbuffer_from_bytes(
         let bytes = bytes.as_slice();
         let mut bytes = bytes.to_vec();
         bytes.shrink_to(i32::MAX as usize);
-        Ok(RustBuffer::from_vec(bytes.to_vec()))
+        Ok(RustBuffer::from_vec(bytes))
     })
 }
 
--- a/uniffi_core/src/ffi/rustbuffer.rs
+++ b/uniffi_core/src/ffi/rustbuffer.rs
@@ -228,6 +228,8 @@ pub fn uniffi_rustbuffer_from_bytes(
 ) -> RustBuffer {
     rust_call(call_status, || {
         let bytes = bytes.as_slice();
+        let mut bytes = bytes.to_vec();
+        bytes.shrink_to(i32::MAX as usize);
         Ok(RustBuffer::from_vec(bytes.to_vec()))
     })
 }
@@ -254,7 +256,8 @@ pub fn uniffi_rustbuffer_free(buf: RustBuffer, call_status: &mut RustCallStatus)
 ///
 /// The second argument must be the minimum number of *additional* bytes to reserve
 /// capacity for in the buffer; it is likely to reserve additional capacity in practice
-/// due to amortized growth strategy of Rust vectors.
+/// due to amortized growth strategy of Rust vectors. If the additional capacity  exceeds
+/// i32::MAX, the buffer will be reallocated to have exactly i32::MAX capacity.
 ///
 /// # Safety
 /// The first argument *must* be a uniquely-owned `RustBuffer` previously obtained from a call
@@ -271,6 +274,7 @@ pub fn uniffi_rustbuffer_reserve(
             .expect("additional buffer length negative or overflowed");
         let mut v = buf.destroy_into_vec();
         v.reserve(additional);
+        v.shrink_to(i32::MAX as usize);
         Ok(RustBuffer::from_vec(v))
     })
 }
--- a/uniffi_core/src/ffi_converter_traits.rs
+++ b/uniffi_core/src/ffi_converter_traits.rs
@@ -247,6 +247,7 @@ pub unsafe trait Lower<UT>: Sized {
     fn lower_into_rust_buffer(obj: Self) -> RustBuffer {
         let mut buf = ::std::vec::Vec::new();
         Self::write(obj, &mut buf);
+        buf.shrink_to(i32::MAX as usize);
         RustBuffer::from_vec(buf)
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### What's fixed?
  
 - Fixed a memory leak in callback interface handling.
+- Prevent `Vec<u8>` capacity from unecessarily triggering panics in `RustBuffer::from_vec`
 
 ### ⚠️ Breaking Changes for external bindings authors ⚠️
 


uniffi_core/src/ffi/rustbuffer.rs
```diff
@@ -228,6 +228,8 @@ pub fn uniffi_rustbuffer_from_bytes(
 ) -> RustBuffer {
     rust_call(call_status, || {
         let bytes = bytes.as_slice();
+        let mut bytes = bytes.to_vec();
+        bytes.shrink_to(i32::MAX as usize);
         Ok(RustBuffer::from_vec(bytes.to_vec()))
```
jplatte, line 233: This introduces at least one extra unnecessary copy because to_vec is now called twice. Did you actually see slice::to_vec producing Vec's with a capacity larger than i32::MAX? Or did you just change all places where RustBuffer::from_vec is called?

uniffi_core/src/ffi/rustbuffer.rs
```diff
@@ -271,6 +274,7 @@ pub fn uniffi_rustbuffer_reserve(
             .expect("additional buffer length negative or overflowed");
         let mut v = buf.destroy_into_vec();
         v.reserve(additional);
+        v.shrink_to(i32::MAX as usize);
         Ok(RustBuffer::from_vec(v))
```
jplatte, line 278: The same effect could be achieved more efficiently using `v.reserve(std::cmp::min(additional, i32::MAX - v.len()));`.

uniffi_core/src/ffi_converter_traits.rs
```diff
@@ -247,6 +247,7 @@ pub unsafe trait Lower<UT>: Sized {
     fn lower_into_rust_buffer(obj: Self) -> RustBuffer {
         let mut buf = ::std::vec::Vec::new();
         Self::write(obj, &mut buf);
+        buf.shrink_to(i32::MAX as usize);
```
MarquessV, line 250: This line is what fixed our issue downstream. I patched the other places `RustBuffer::from_vec` was being called in case that was wanted, but happy to remove them if not.

uniffi_core/src/ffi/rustbuffer.rs
MarquessV: Oops, yeah, no need for that extra `to_vec()`, thanks.

I have observed this issue happening in `Lower::lower_rust_into_vec` specifically. The patches in `rustbuffer` are only in the spirit of preventing a different code path from triggering the issue, but I'm happy to remove these if they are too heavy-handed.

uniffi_core/src/ffi/rustbuffer.rs
MarquessV: My understanding is that [reserve](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reserve) will reserve at least `additional` more elements, but it _could_ reserve more, which would put it in danger of exceeding the max capacity. `shrink_to` is a no-op if the current capacity is less than the requested minimum, so I think the cost of `shrink_to` is fairly light for `Vec`s that wouldn't exceed the capacity anyway.

bendk: How about `reserve_exact`?

uniffi_core/src/ffi_converter_traits.rs
bendk: I agree that it would be good to avoid an extra copy here.  What about changing the `write()` implementations to call `reserve_exact` before pushing to the Vec?  Unfortunately, I think that means implementing our own growth algorithm, but something simple should work okay.

bendk (changes requested):
Thanks for pointing out the issue, but I would like to try to fix it without the potential of `shrink_to` performing an extra copy.

In addition to the `reserve_exact` strategy I laid out, it's also possible to just change the `RustBuffer` fields.  For one thing, I don't think it would be hard at all to change them to a `u32`.  The `i32` was because of historical issues with signed ints and Kotlin, but I don't think that's an issue anymore.  For that matter, it wouldn't be out of the question to change it to a `u64`.  I don't think the extra 8 bytes is an issue.  The one thing I don't think we should use is a `usize`.  That can be hard to deal with for some foreign bindings and we've made the decision that 32-bit machines are not worth optimizing for.

MarquessV:
> Thanks for pointing out the issue, but I would like to try to fix it without the potential of `shrink_to` performing an extra copy.
> 
> In addition to the `reserve_exact` strategy I laid out, it's also possible to just change the `RustBuffer` fields. For one thing, I don't think it would be hard at all to change them to a `u32`. The `i32` was because of historical issues with signed ints and Kotlin, but I don't think that's an issue anymore. For that matter, it wouldn't be out of the question to change it to a `u64`. I don't think the extra 8 bytes is an issue. The one thing I don't think we should use is a `usize`. That can be hard to deal with for some foreign bindings and we've made the decision that 32-bit machines are not worth optimizing for.

No problem! I would also prefer a fix that changes the `RustBuffer` to `u64`. I wasn't comfortable making that change due to the documented constraints, but if those don't apply anymore, that feels like the best way forward.

bendk:
#1978 is my attempt at  increasing the field sizes to `u64`

mhammond:
> without the potential of shrink_to performing an extra copy.

I don't understand the extra copy shrink_to implies, and why it doesn't work here - isn't the problem the extra capacity rather than the actual length?

It sounds like it might be easy to reproduce in a test too?

(I agree using 64 bits isn't an issue, but I'm confused why this needs it)

bendk:
> > without the potential of shrink_to performing an extra copy.
> 
> I don't understand the extra copy shrink_to implies, and why it doesn't work here - isn't the problem the extra capacity rather than the actual length?

My understanding is that if the capacity overflows `i32::MAX`, but not the length, then the `shrink_to()` call will allocate a new vec with less capacity and copy the data into it.

mhammond:
> My understanding is that if the capacity overflows `i32::MAX`, but not the length, then the `shrink_to()` call will allocate a new vec with less capacity and copy the data into it.

I thought it might be something like that but I can't find any docs to back that up. Eg, https://doc.rust-lang.org/std/vec/struct.Vec.html#method.shrink_to makes no such claim, just "If the current capacity is less than the lower limit, this is a no-op."

But then stackoverflow hints we should really be looking at https://doc.rust-lang.org/std/alloc/trait.Allocator.html#method.shrink, and it gets a lot closer to implying the shink might reallocate.

So yeah, I agree your other patches makes the most sense if this is true - I do with it was spelled out a little better though.

bendk:
> So yeah, I agree your other patches makes the most sense if this is true - I do wish it was spelled out a little better though.

Me too.  The one that convinced me was in [from_raw_parts](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts): `capacity needs to be the capacity that the pointer was allocated with.` 

The simplest solution from our POV would be to clamp the value of `RustBuffer.capacity`, but that's not allowed.  Based on that I assume that the can't do the same thing internally to the Vec.

mhammond:
This is stale now that #1976 has landed, right?

bendk:
#1976 fixes the bug this in a different way.  I think we can close this one now.

PR #1978, opened 2024-02-02
Change RustBuffer length/capacity to `u64` (#1976)

### LINKED ISSUE #1976: Vec capacity can unnecessarily cause Rust panics when creating a `RustBuffer`.

bendk:
This fixes #1976, which tracks Rust panics with values that overflow the `RustBuffer` fields.  Note that `capacity` can overflow even if the rust buffer size is < `i32::MAX` because of the way vectors grow.  As discussed in https://github.com/mozilla/uniffi-rs/pull/1977 one of the simplest ways to fix this is to just make the fields bigger.

In theory, this should be a pretty safe change.  I'm a bit worried that I missed some detail though -- especially with Kotlin since JNA always feels like it's full of footguns.

I'd love to know how others feel about this.  How risky does it feel to others?  Does anyone have performance concerns?

Note: we still write 32-bit lengths to the RustBuffer itself.  Should we increase that to 64-bits as well?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@
 
 - Procmacros support tuple-enums.
 
+- `RustBuffer` was changed to use `u64` fields.
+  This eliminates panics when the capacity of the vec exceeds `i32::MAX`.
+  This can happen with the current Vec implementation when String/Vec sizes approach `i32::MAX` but don't exceed it.
+
 ### What's fixed?
  
 - Fixed a memory leak in callback interface handling.
@@ -38,6 +42,7 @@
   types including:
   - Rust futures (replacing `FfiType::RustFutureHandle` which was removed)
   - Rust future continuation data (Replacing `FfiType::RustFutureContinuationData` which was moved).
+- `RustBuffer.len` and `RustBuffer.capacity` are now `u64` rather than `i32`.
 
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.26.1...HEAD).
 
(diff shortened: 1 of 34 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,13 @@
 
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.26.0...HEAD).
 
+### What's changed?
+
+- The `rust_future_continuation_callback_set` FFI function was removed.  `rust_future_poll` now
+  inputs the callback pointer.  External bindings authors will need to update their code.
+- `RustBuffer` was changed to use `u64` fields.  This eliminates potential panics with using
```
mhammond, line 15: Maybe make this a little more precise - IIUC, it's "when the capacity of the vec exceeds what can be stored in an `i32`" (whereas the "approach" language is talking more about the current impl of vecs).

(all IIUC of course!)

mhammond (approved):
This all seems straight-forward, and while I'm surprised there doesn't seem an option for avoid this when we never except the length to exceed 32 bits, I don't see what harm this causes.

bendk:
> This all seems straight-forward, and while I'm surprised there doesn't seem an option for avoid this when we never except the length to exceed 32 bits, I don't see what harm this causes.

I think the one other option would be to implement the vec growth algorithm ourselves by calling `reserve_exact()` before pushing elements.  I prefer this one slightly, but I could go for that option as well.

mhammond:
> I think the one other option would be to implement the vec growth algorithm ourselves by calling `reserve_exact()` before pushing elements. I prefer this one slightly, but I could go for that option as well.

I don't really have much of an opinion here, but that other option seems like it might come back to bite us - eg, for vecs we don't create.

bendk:
Talked with Mark about this one and we couldn't see a good reason not to merge this one for 0.27.0.

PR #1980, opened 2024-02-06
Try removing build dependency in procmacro example

mgeisler:
Since there is no `build.rs` file, I don't think the build dependency has any effect.


bendk (approved):
Looks good to me, thanks for the contribution!

PR #1981, opened 2024-02-06
Adding support for Rust async trait methods

bendk:
This was pretty easy, it mostly meant not hard coding that we don't support async methods and which order to apply `#[uniffi::export]` vs `#[async_trait::async_trait]`.

The next step is foreign-implemented traits.

### DIFF
--- a/docs/manual/src/futures.md
+++ b/docs/manual/src/futures.md
@@ -45,3 +45,29 @@ This code uses `asyncio` to drive the future to completion, while our exposed fu
 In Rust `Future` terminology this means the foreign bindings supply the "executor" - think event-loop, or async runtime. In this example it's `asyncio`. There's no requirement for a Rust event loop.
 
 There are [some great API docs](https://docs.rs/uniffi_core/latest/uniffi_core/ffi/rustfuture/index.html) on the implementation that are well worth a read.
+
+## Exporting async trait methods
+
+UniFFI is compatible with the [async-trait](https://crates.io/crates/async-trait) crate and this can
+be used to export trait interfaces over the FFI.
+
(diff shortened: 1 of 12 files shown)


docs/manual/src/futures.md
```diff
@@ -45,3 +45,29 @@ This code uses `asyncio` to drive the future to completion, while our exposed fu
...
+
+```idl
+[Trait]
+interface SayAfterTrait {
+    [Async]
+    string say_after(u16 ms, string who);
+};
+```
+
+When using proc-macros, make sure to put `#[uniffi::export]` outside the `#[async_trait]` attribute:
```
mhammond, line 65: Is it strictly necessary for it to be "**outside** the `#[async_trait]` attribute"? I thought it would work if those 2 lines were swapped?

mhammond (approved):
I love how tiny the actual change is! Also, this needs a CHANGELOG entry.

docs/manual/src/futures.md
bendk: I'm not sure exactly why, but it fails because of the lifetimes `async_trait` adds:

```
    |                  - lifetime `'async_trait` is missing in item created through this procedural macro
332 | pub trait SayAfterTrait: Send + Sync {
333 |     async fn say_after(&self, ms: u16, who: String) -> String;
    |     ^^^^^ undeclared lifetime
```

shaiguelman:
Do you have an estimate of when the next release will be? My team is strongly dependent on this change

bendk:
No estimate, but you can try to use the git commit in to get the feature.  Other teams have had success with that.

Issue #1970, opened 2024-01-30
Python doesn't generate return type of async functions


giarc3:
If I export a function with a signature of
```
#[uniffi::export]
fn double(num: u32) -> u32
```
then it generates Python with a function signature of
```
def double(num: "int") -> "int":
```
---
But if I change the Rust to an `async` function, it generates
```
def double(num: "int"):
```
with no return type specified. This leads to Python documentation saying the return type is `Coroutine[Any, Any, Any]` and warnings whenever I try to access fields on the returned value.

mhammond:
Sadly we don't have many tests here so I'm not easily able to verify what the correct thing to generate is - I *think* it should generate `-> "Awaitable[int]"` - is that correct? Can you verify by hand-editing the generated .py that does what you expect?

mhammond:
(alternatively, I think we could arrange to generate an `async` fn there, which, IIUC, would be happy with `-> "int"`)

giarc3:
I believe that adding `async` and indicating the return type as normal is the correct thing to do here. I hand-edited the generated code and confirmed that it matched my expectation when calling the function

mhammond:
I thought changing to async will also require the final return to be "return await" or similar?

giarc3:
You're correct, that's also necessary. The return statement it generated for me is a call to `_uniffi_rust_call_async`, so it should change to `return await _uniffi_rust_call_async`

giarc3:
Completed by https://github.com/mozilla/uniffi-rs/pull/1984

PR #1983, opened 2024-02-06
Add type hints to Python dict

giarc3:
This adds type hints to Python dicts.

I couldn't find official documentation indicating what Python versions uniffi supports, so I wrote it to support 3.8+ as that's the oldest version not EOL.

In 3.8, the correct way to do this would be `typing.Dict[X, Y]`. Then in 3.9, support was added for `dict[X, Y]` and the original way was deprecated. By importing `from __future__ import annotations`, we can support the current way on 3.8 as well.


mhammond:
Fantastic, thanks - and for the tests! Have you seen #1970 by any chance? ;)

mhammond (approved):

mhammond:
> Have you seen #1970 by any chance?

lol, of course you have, sorry about that!

PR #1984, opened 2024-02-06
Add Python return type for async functions

giarc3:
As discussed in #1970, uniffi currently doesn't put a return type on generated async functions in Python. 

This is my attempt at correcting the function signatures it generates. Because the diff is a bit bad, I'll summarize the changes:
- For async functions, match on if they have a return type
  a. If they do, add `async` to the front of the function, add the `-> ___` return type, and add `await` to the return statement.
  b. If they don't, still add `async` to the front of the function and `await` to the return.
- Remove the `match meth.return_type()` cases from the body of the function, as it's now being checked outside of the function

The changes are repeated in `macros.py` and `TopLevelFunctionTemplate.py`

`b.` technically could remain unchanged, but it seemed more correct to make the generated functions match each other.

Note that I'm by no means a Python expert, so please let me know if there's a more ergonomic/preferred way to accomplish all this.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py
@@ -1,20 +1,18 @@
 {%- if func.is_async() %}
 
-def {{ func.name()|fn_name }}({%- call py::arg_list_decl(func) -%}):
+{%- match func.return_type() -%}
+{%- when Some with (return_type) %}
+
+async def {{ func.name()|fn_name }}({%- call py::arg_list_decl(func) -%}) -> "{{ return_type|type_name }}":
     {%- call py::docstring(func, 4) %}
     {%- call py::setup_args(func) %}
-    return _uniffi_rust_call_async(
+    return await _uniffi_rust_call_async(
         _UniffiLib.{{ func.ffi_func().name() }}({% call py::arg_list_lowered(func) %}),
         _UniffiLib.{{func.ffi_rust_future_poll(ci) }},
         _UniffiLib.{{func.ffi_rust_future_complete(ci) }},
         _UniffiLib.{{func.ffi_rust_future_free(ci) }},
         # lift function
-        {%- match func.return_type() %}
-        {%- when Some(return_type) %}
         {{ return_type|lift_fn }},
-        {%- when None %}
-        lambda val: None,
-        {% endmatch %}
         # Error FFI converter
         {%- match func.throws_type() %}
         {%- when Some(e) %}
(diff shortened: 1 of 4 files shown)


mhammond:
Thanks! This is a larger change than I was anticipating - which isn't necessarily a problem, just needs a closer look. I'll look in more detail over the next day or 2.

uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py
```diff
@@ -1,20 +1,39 @@
 {%- if func.is_async() %}
 
-def {{ func.name()|fn_name }}({%- call py::arg_list_decl(func) -%}):
+{%- match func.return_type() -%}
+{%- when Some with (return_type) %}
+
+async def {{ func.name()|fn_name }}({%- call py::arg_list_decl(func) -%}) -> "{{ return_type|type_name }}":
```
mhammond, line 6: It looks to me like you've moved the check of `return_type` outside of the function just so this first line can be different, with the cost being more duplicated code between each branch - I'm not sure that's the correct trade-off - I think an extra match block just for this line might be better?

The Python docs also imply that the "returns None" case can also be annotated with `-> None` whereas this just declines to annotate that case.

mhammond (changes requested):
Thanks! Please let me know if I missed something here...

uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py
giarc3: > just so this first line can be different

I brought it outside for the first line of the function as well as the `# lift function` cases inside of the function body, essentially reducing duplication of the `{%- match func.return_type() -%}` and cases lines, but increasing duplication of other lines. So I could change it to two match statements (one outside for the first line of the function, one inside for the lift function), or keep them combined at the top. I'm good with it either way, just wanted to confirm with you first which you'd prefer.

> "returns None" case can also be annotated with `-> None`

I think this is a good idea, which I'd extend to the non-async case as well. I was concerned at first that we'd be making explicit something that was implicit in Rust, but I think the Rust compiler requires a return type in a strict enough way that it's not a problem. Thanks for the suggestion!

mhammond: yeah - while this is very subjective, I think "number of duplicated lines" is a reasonable yardstick and my impression is duplicating the control-flow but not the actual template would be smaller than the opposite.

giarc3:
@mhammond I pushed a change to add the `None` and reduce the code duplication as you indicated. 

In doing testing of this against a local project, I happened upon the Protocol classes getting generated, and hit the extent of my knowledge. As I didn't change `Protocol.py`, the functions in the protocol classes remain without `async` and without the return type (if present). Do you know if this is an issue? I would have expected the lack of `async` to cause a problem, but I haven't run into anything.

mhammond (approved):
I think the lack of `async` in protocol should be OK as `async` is just sugar - the prototype code still returns a coroutine. The lack of annotations there isn't ideal, but I don't see how it hurts. The fact our CI is green and this has tests means I think we should land this - if it does turn out we broke something then we'll fix it and add more tests :)

Thanks!

PR #1989, opened 2024-02-09
Use typing.get_type_hints() in python type tests

mhammond:
edit: As @badboy points out, we probably do want strings - sorry about that. It turns out `typing.get_type_hints()` does the magic we need to make the tests sane and able to check against actual types.

### DIFF
--- a/fixtures/futures/tests/bindings/test_futures.py
+++ b/fixtures/futures/tests/bindings/test_futures.py
@@ -2,6 +2,7 @@ from futures import *
 import unittest
 from datetime import datetime
 import asyncio
+import typing
 
 def now():
     return datetime.now()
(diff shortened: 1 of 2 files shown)


mhammond:
(which is really quite ironic - Python's typing "solution" allows it own set of simple accidental type problems :)

badboy:
If types are used in type places they need to exist before use. Linters and external type-checkers will complain.
If declared as strings they are still type-checked, but can reference later-declared types.

mhammond:
Right, but all the python docs use types for the builtin types - eg, `int` and `None` rather than `"int"` and `"None"`, and main is failing with the dict being `{"ms": "'int'", "return": "'bool'"}`.

I guess I need to expand the tests to use named types to see what happens there - but what should we prefer for a type we know is in scope (eg, builtins, including `list[]` etc)?

uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py
```diff
@@ -2,7 +2,7 @@
 
 {%- match func.return_type() -%}
 {%- when Some with (return_type) %}
-async def {{ func.name()|fn_name }}({%- call py::arg_list_decl(func) -%}) -> "{{ return_type|type_name }}":
+async def {{ func.name()|fn_name }}({%- call py::arg_list_decl(func) -%}) -> {{ return_type|type_name }}:
```
mhammond, line 5: at the very least, this is now inconsistent with non-async functions

bendk (approved):
Looks great, thanks for fixing the tests.

PR #1990, opened 2024-02-10
Allow procmacro exported items to be renamed.

### LINKED ISSUE #1753: Allow specifying primary constructor
Use case:

I have an existing Rust library (or a library I wish to have an idiomatic Rust interface). It has a struct 

```rust
#[derive(uniffi::Object)]
pub struct MyLibraryKey {
    bytes: [u8; 512]
}
#[uniffi:export]
impl MyLibraryKey {
    #[uniffi::constructor]
    pub fn new(bytes: [u8; 512]) -> Self {
        Self {
            bytes
        }
    }
}
```

Currently the `primary_constructor` is inferred from it being named `new`. Foreign languages don't support fixed size arrays, but I'd like to keep my ergonomic Rust API surface so I do something like:

```rust
impl MyLibraryKey {
    pub fn new(bytes: [u8; 512]) -> Self {
        Self {
            bytes
        }
    }
}
#[uniffi::export]
impl MyLibrary {
    #[uniffi::constructor]
    pub fn new_ffi(bytes: Vec<u8>) -> Result<Self, String> {
        Ok(Self {
            bytes: bytes.try_into().map_err(|_| "wrong byte length")?
        })
    }
}
```

Expectation:
I want to be able to tell `uniffi` that `new_ffi` is the primary constructor so downstream languages use it in their native constructors (ie `MyLibraryKey(b"deadbeef")`), but Rust keeps its ergonomic interface as well.

Potential Solution:
Something like `#[uniffi::constructor(is_primary=true)]` would work, but doing that in a way that doesn't break current usage that expects `new` to be primary may be tricky. Putting `#[cfg(feature = "ffi")]` and `#[cfg(not(feature = "ffi"))]` around each `impl` would be a workaround but as things get more complex could get extremely unwieldy.

mhammond:
Fixes #1753

While #1753 talks just about the constructor, I noticed that we already had the capability of using different names for exported items but just lacked a way to express that.

### DIFF
--- a/docs/manual/src/proc_macro/index.md
+++ b/docs/manual/src/proc_macro/index.md
@@ -138,6 +138,31 @@ fn do_something(foo: MyFooRef) {
 }
 ```
 
+### Renaming functions, methods and constructors
+
+A single exported function can specify an alternate name to be used by the bindings by specifying a `name` attribute.
+
+```rust
+#[uniffi::export(name = "something")]
+fn do_something() {
+}
+```
+will be exposed to foreign bindings as a namespace function `something()`
+
+You can also rename constructors and methods:
+```rust
+#[uniffi::export]
+impl Something {
+    // Set this as the default constructor by naming it `new`
+    #[uniffi::constructor(name = "new")]
+    fn make_new() -> Arc<Self> { ... }
+
+    // Expose this as `obj.something()`
+    #[uniffi::method(name = "something")]
+    fn do_something(&self) { }
+}
+```
+
 ## The `uniffi::Record` derive
 
 The `Record` derive macro exposes a `struct` with named fields over FFI. All types that are
--- a/fixtures/proc-macro/src/lib.rs
+++ b/fixtures/proc-macro/src/lib.rs
@@ -304,4 +304,27 @@ pub fn join(parts: &[String], sep: &str) -> String {
     parts.join(sep)
 }
 
+// Custom names
+#[derive(uniffi::Object)]
+pub struct Renamed;
+
+// `renamed_new` becomes the default constructor because it's named `new`
+#[uniffi::export]
+impl Renamed {
+    #[uniffi::constructor(name = "new")]
+    fn renamed_new() -> Arc<Self> {
+        Arc::new(Self)
+    }
+
+    #[uniffi::method(name = "func")]
+    fn renamed_func(&self) -> bool {
+        true
+    }
+}
+
+#[uniffi::export(name = "rename_test")]
+fn renamed_rename_test() -> bool {
+    true
+}
+
 uniffi::include_scaffolding!("proc-macro");
--- a/uniffi_macros/src/export/attributes.rs
+++ b/uniffi_macros/src/export/attributes.rs
@@ -70,6 +70,7 @@ impl UniffiAttributeArgs for ExportTraitArgs {
 #[derive(Default)]
 pub struct ExportFnArgs {
     pub(crate) async_runtime: Option<AsyncRuntime>,
+    pub(crate) name: Option<String>,
 }
 
 impl Parse for ExportFnArgs {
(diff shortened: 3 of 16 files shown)


docs/manual/src/proc_macro/index.md
```diff
@@ -138,6 +138,29 @@ fn do_something(foo: MyFooRef) {
 }
 ```
 
+### Renaming exported items
+
+A single exported item can specify an alternate name to be used by the binding by specifying a `name` attribute.
+
+```rust
+#[uniffi::export(name=something)]
```
jplatte, line 146:
```suggestion
#[uniffi::export(name = something)]
```

docs/manual/src/proc_macro/index.md
```diff
@@ -138,6 +138,29 @@ fn do_something(foo: MyFooRef) {
...
+```
+will be exposed to foreign bindings as a namespace function `something()`
+
+You can use this to specify a default contructor by renaming it to `new`:
+```rust
+#[uniffi::export(name=new)]
+impl Something {
+    #[uniffi::constructor]
+    fn make_new() -> Arc<Self> { ... }
+}
```
jplatte, line 158: Why would the c'tor name not be part of `constructor`s arguments?

docs/manual/src/proc_macro/index.md
jplatte: Also I wonder if a bare word should be used here, or a string. If it's a bare word, using a keyword in that position will likely result in it being highlighted, but other than that I can't think of any practical consequences.

uniffi_macros/src/export/attributes.rs
```diff
@@ -86,6 +89,15 @@ impl UniffiAttributeArgs for ExportFnArgs {
             let _: Token![=] = input.parse()?;
             Ok(Self {
                 async_runtime: Some(input.parse()?),
+                name: None,
+            })
+        } else if lookahead.peek(kw::name) {
+            let _: kw::name = input.parse()?;
+            let _: Token![=] = input.parse()?;
+            let name = Some(ident_to_string(&input.parse::<Ident>()?));
```
jplatte, line 97: `.parse::<Ident>()` will forbid Rust keywords in the input, which seems wrong, I could see one primary use for this feature being when you wanna use a Rust keyword as the foreign language name. `Ident::parse_any` gets around that.

fixtures/proc-macro/src/lib.rs
```diff
@@ -304,4 +304,29 @@ pub fn join(parts: &[String], sep: &str) -> String {
...
+        Arc::new(Self)
+    }
+}
+
+#[uniffi::export(name=func)]
+impl Renamed {
+    fn renamed_func(&self) -> bool {
+        true
+    }
+}
```
jplatte, line 325: How about another dummy attribute like constructor (just to have a place to attach method attributes) instead of using one-item impls?


```suggestion
#[uniffi::export]
impl Renamed {
    #[uniffi::method(name = func)]
    fn renamed_func(&self) -> bool {
        true
    }
}
```

docs/manual/src/proc_macro/index.md
mhammond: I was going back and forward here. My initial thought was that an ident is closer to the intent and would help avoid obviously incorrect values (eg, with a space), but the more I look at it, the more I think it probably should just be a string - having an ident there makes my brain pause as I scan over it, wondering where this `something` came from. Making it a string makes it clearer we are just assigning a literal value and as you mention, causes less surprising syntax highlighting..

uniffi_macros/src/export/attributes.rs
mhammond: As above, I'll change to a string, but I think we [already support Rust keywords](https://github.com/mozilla/uniffi-rs/tree/main/fixtures/keywords/rust/src) as foreign language names.

jplatte: Yeah, supporting it elsewhere is why it should not cause a parse error here either. But moot point anyways, as you noted.

uniffi_macros/src/export/attributes.rs
```diff
@@ -226,6 +303,14 @@ impl ExportedImplFnAttributes {
                     }
                     this.constructor = true;
                 }
+                "method" => {
+                    if this.constructor {
```
mhammond, line 307: do we really need this - maybe this can just be left unchecked and this becomes a noop?

docs/manual/src/proc_macro/index.md
```diff
@@ -138,6 +138,31 @@ fn do_something(foo: MyFooRef) {
 }
 ```
 
+### Renaming functions, methods and constructors
+
+A single exported function can specify an alternate name to be used by the bindings by specifying a `name` attribute.
+
+```rust
+#[uniffi::export(name = "something")]
```
Sajjon, line 146: This is nice! Any plans to support renaming of `args`? That would be really fantastic! In the realm of Swift, we typically use prepositions a lot. so `from, `to`, `by` etc is the preferred label, e.g. [see Swift standard library `stride:from:to:by`](https://developer.apple.com/documentation/swift/stride(from:to:by:)). But since Rust does not have the notion of seperated "labels" (externally shown) and argname (internal to the body of the function) like Swift does, as Rust devs it is strange to use `from`, `to`, `by` as arg names.

So it would be amazing of `uniffi::export(name` could support renames of args!

Bikeshedding syntax:
```rust
#[uniffi::export(fn_name = "stride", arg_names = "from,to,by")]
fn do_stride(start: usize, end: usize, stride_length: usize) {
}
```

Would that be interesting to add support for?

Would it be possible?

jplatte (commented):
Haven't had a look in a while but I guess we can merge and improve later, right? If you want a proper second review though, just say so and I'll try to find some time for it.

jplatte (approved):
Review above was supposed to be an approval.

PR #1992, opened 2024-02-12
Refactor FFI definition code

bendk:
This was not quite a pure refactor, because some of the code generation for FFI definitions changed.  AFAICT, it should be equivalent far all current FFI definitions though.

Added the `FfiDefinition` enum, which can hold any type of FFI definition (struct, function, or callback function).  Added `ComponentInterface::ffi_definitions` to iterate over all of the definitions.  The reason for this is that FFI definitions can have complicated dependencies and some languages require that the dependencies come first in source order.  By defining a single function to output all definitons, we can ensure that they come out in the right order.

Factored out the Rust FFI return types calculation into its own function.  I want to use it in some other places.

Updated the codegen for FFI definitions:
  - Kotlin (most of these changes are to workaround JNA issues):
    - Make FFI structs open. 
    - If a FFI struct contains an FFI callback field, make that field nullable.
    - Add default field values.
    - Added `UniffiByValue` for passing structs by value
    - Added the `setValue` utility method
  - Swift: add an IF guard for FFI definitions.  That way we can introduce shared FFI structs without getting a name conflict.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
@@ -340,6 +340,39 @@ impl KotlinCodeOracle {
         }
     }
 
+    /// FFI type name to use inside structs
+    ///
+    /// The main requirement here is that all types must have default values or else the struct
+    /// won't work in some JNA contexts.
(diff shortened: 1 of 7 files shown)


mhammond (approved):
👍

PR #1999, opened 2024-02-27
Simplify path in bindings tests

mgeisler:
Joining “.” with anything ought to be the same as using anything directly.


mhammond:
Does the current behaviour cause a problem for you, or is this just "cleaner"?

mgeisler:
> Does the current behaviour cause a problem for you, or is this just "cleaner"?

It's just cleaner, I saw the code and wondered why it's there — my conclusion is that it's unnecessary so I removed it to avoid having others wonder the same.

mhammond (approved):
Fair enough - I can't argue with that, I just wanted to make sure I wasn't missing anything - thanks!

PR #2001, opened 2024-02-29
Prevent Swift objects' `deinit` from trying to free a nil pointer when subclassed for mocking

stefanceriu:
It seems that #1918 and implicitly #1975 missed one important aspect when opening up classes for mocking:
- the current deinit implementation shouldn't be called for mocked/subclassed objects because the internal `pointer` is nil in those cases and doing so will result in a `Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value`
- deinit is also not overridable in Swift so this is not something that can be worked around downstream

This PR attempts to fix the problem by introducing a guard to check that `pointer` is not nil before trying to free it.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift
@@ -63,6 +63,10 @@ open class {{ impl_class_name }}:
     {%- endmatch %}
 
     deinit {
+        guard let pointer = pointer else {
+            return
+        }
+
         try! rustCall { {{ obj.ffi_object_free().name() }}(pointer, $0) }
     }
 


mhammond:
Conceptually this seems correct, but the swift compiler doesn't seem to like it :)

stefanceriu:
Oh interesting, let me try something..

stefanceriu:
Okay, it's happy now. While `guard let pointer else` is completely valid it's a rather new syntactic sugar addition to Swift, with the original being `guard let pointer = pointer else`. I assume that the tests are using an older Swift version but the original syntax is still completely acceptable so I don't think we need any other changes.

mhammond (approved):
Looks great, thanks!

stefanceriu:
Yey, thank you too!

Issue #1078, opened 2021-10-25
Add support for enum/error fields from tuples


bendk:
Currently, enum/error with fields require the variant to use named fields.  Take the example from the book:

Rust: 

```
enum IpAddr {
  V4 {q1: u8, q2: u8, q3: u8, q4: u8},
  V6 {addr: String},
}
```

UDL:

```
[Enum]
interface IpAddr {
  V4(u8 q1, u8 q2, u8 q3, u8 q4);
  V6(string addr);
};
```

But sometimes the tuple style is much more convenient and feels more Rust-like.  It would be nice to support that too.  What if we added support for these via a UDL annotation?:

Rust: 

```
enum IpAddr {
  V4(u8, u8, u8, u8),
  V6(String),
}
```

UDL:

```
[Enum]
[TupleFields]
interface IpAddr {
  V4(u8 q1, u8 q2, u8 q3, u8 q4);
  V6(string addr);
};
```

Here's how I'm thinking `[TupleFields]` could work:

* It can be applied to either a single variant, or the enum/error as a whole.  Annotating the whole enum/error is like annotating each variant.
* For the scaffolding code, if a variant has `[TupleFields]` applied, then we use the index of the parameter rather than the name to get the field value (`ipaddr.0` instead of `ipaddr.q1` in the above example)
* For the bindings code, we still use named fields.
* The `[StructFields]` annotation would undo `[TupleFields]`.  This is useful for when the entire enum/error is annotated.  We could also consider switching which one of these is the default.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/UNIFFI-103)
┆Issue Number: UNIFFI-103

mhammond:
> For the bindings code, we still use named fields.

I'm not quite convinced about that part yet.

PR #2004, opened 2024-03-03
procmacros support tuple-enums

mhammond:
These are represented in the CI as enum variants with nameless fields.

Part of #1078, although no support for UDL yet.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/templates/EnumTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/EnumTemplate.py
@@ -24,6 +24,28 @@ class {{ type_name }}:
     class {{ variant.name()|enum_variant_py }}:
         {%- call py::docstring(variant, 8) %}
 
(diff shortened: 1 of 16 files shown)


uniffi_bindgen/src/bindings/python/templates/EnumTemplate.py
```diff
@@ -24,20 +24,42 @@ def __init__(self):
     class {{ variant.name()|enum_variant_py }}:
         {%- call py::docstring(variant, 8) %}
 
+    {%-  if variant.has_nameless_fields() %}
+        def __init__(self, *values):
+            if len(values) != {{ variant.fields().len() }}:
+                raise TypeError(f"Expected a tuple of len {{ variant.fields().len() }}, found len {len(values)}")
+        {%- for field in variant.fields() %}
+            if not isinstance(values[{{ loop.index0 }}], {{ field|type_name }}):
```
cschramm, line 32: This can lead to invalid code, see #2445

PR #2009, opened 2024-03-05
Check lowering after coercing default value

badboy:


bendk (approved):
The change seems right to me.

Issue #2017, opened 2024-03-07
Foreign-implemented async traits


bendk:
We currently allow Rust code to implement async trait methods, but foreign implementations are not supported.  We should extend support to allow for foreign code.

I think this is a key feature for full async support.  It allows Rust code to define an async method that depends on a foreign async method.  This allows users to use async code without running a Rust async runtime, you can effectively piggyback on the foreign async runtime.

bendk:
This one was implemented in https://github.com/mozilla/uniffi-rs/pull/2018

PR #2002, opened 2024-02-29
Rename kotlin uniffiRustFutureContinuationCallback to uniffiRustFutureContinuationCallbackImpl

stan-stately:
Fixes https://github.com/mozilla/uniffi-rs/issues/1995

I'm not exactly sure why but certain java runtimes can't distinguish between `uniffiRustFutureContinuationCallback` and `UniffiRustFutureContinuationCallback`. the issue manifests in various ways but I was getting `java.lang.NoClassDefFoundError`:

```
java.lang.NoClassDefFoundError: uniffi/<crate>/uniffiRustFutureContinuationCallback (wrong name: uniffi/<crate>/UniffiRustFutureContinuationCallback)
```

I patched in this version of uniffi and verified that it's fixed

PR #2018, opened 2024-03-08
Foreign-implemented async trait methods (#2017)

bendk:
These methods input a completion function that they call when the async function is complete. They return a ForeignFuture struct, which represents the foreign task object and is used to drop/cancel futures. Like Rust, dropping and cancelling are coupled together into one operation.

- Added `RustCallStatus` as an `FfiType` variant, since I wanted to use it in the `ForeignFutureResult` structs.  In theory, we could define `RustCallStatus` a `FfiType::Struct`, but I didn't want to introduce that change in this PR.
- Fixed the result mapping code to work with async functions.  Before we were executing the mapping call, then awaiting the result, but we need to do that in the opposite order (`foo.into().await` vs `foo.await.into()`). Also, specify the return type generics for `rust_future_new` so that the Rust can figure out the into() generics.
- Updated the `futures` fixture to test this.  Added the `async-api-client` to show how it could work.


bendk:
> @bendk I'm very excited for this work. I'm trying it out right now and I'm getting an exception finding FutureFree...
> `java.lang.ClassNotFoundException: Didn't find class "uniffi.spokeffi.UniffiForeignFutureFree"`
> I see `UniffiForeignFutureFree` in the generated Kotlin.

Weird.  I'm not sure what's happening there, can you put together minimal way to reproduce this?  I can take a closer look once I have a branch or something to use.

jasl8r:
Okay I put together a super simple example with all the Android shell here: https://github.com/jasl8r/foreignfree
The `ExampleInstrumentedTest` will show the failure.

bendk:
I couldn't figure out how to run that test, but the error message seems very similar to https://github.com/mozilla/uniffi-rs/pull/2002.  I made same rename as that PR, can you see if it fixes your issue?

jasl8r:
Thanks @bendk, that is definitely the problem, I slapped in a similar fix on the generated code and it does resolve the problem.

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
```diff
@@ -10,25 +10,31 @@ fileprivate struct {{ trait_impl }} {
...
-            } catch {
-                uniffiCallStatus.pointee.code = CALL_UNEXPECTED_ERROR
-                uniffiCallStatus.pointee.errorBuf = {{ Type::String.borrow()|lower_fn }}("Callback handle map error: \(error)")
-                return
+            let makeCall = {
+                () {% if meth.is_async() %}async {% endif %}throws -> {% match meth.return_type() %}{% when Some(t) %}{{ t|type_name }}{% when None %}(){% endmatch %} in
+                guard let uniffiObj = try? {{ ffi_converter_name }}.handleMap.get(handle: uniffiHandle) else {
+                    throw UniffiInternalError.unexpectedStaleHandle
+                }
+                // var uniffiObj: {{ type_name }}
```
mhammond, line 24: should this be deleted?

uniffi_core/src/ffi/foreignfuture.rs
```diff
@@ -0,0 +1,241 @@
...
+//!
+//! The foreign side should:
+//!   * Input a [ForeignFutureCallback] and a `u64` handle in their scaffolding function.
+//!     This is the sender, converted to a raw pointer, and an extern "C" function that sends the result.
+//!   * Return a [ForeignFuture], which represents the foreign task object corresponding to the async function.
+//!   * Call the [ForeignFutureCallback] when the async function completes with:
+//!     * The `u64` handle initially passed in
+//!     * The `ForeignFutureResult` for the call
+//!   * Wait for the [ForeignFutureHandle::free] function to be called to free the task object.
+//!     If this is called before the task completes, then the task should be cancelled.
```
mhammond, line 18: "should be" isn't clear here - "will be"?

mhammond (approved):
This really is excellent - it's very thorough, well documented and tested, and generally quite amazing :)

uniffi_bindgen/src/bindings/python/templates/Async.py
```diff
@@ -5,6 +5,30 @@
 # Stores futures for _uniffi_continuation_callback
 _UniffiContinuationHandleMap = _UniffiHandleMap()
 
+UNIFFI_GLOBAL_EVENT_LOOP = None
```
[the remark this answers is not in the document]
bendk: Yes it does and yes it's unfortunate.  It only needs to be done in for a fairly specific case though, so maybe it'll be okay.

PR #2019, opened 2024-03-08
Make `uniffi_build` optional in `uniffi_macros`

mgeisler:
As far as I can see, the dependency on uniffi_build is not needed for the macros used by regular clients of UniFFI — it is only used to create a convenience macro for the UI tests.

I named the new feature “trybuild” after the description of the macro. Please let me know if you have preferences for a better name.

The overall goal here is to minimize the number of crates I need to vendor to use UniFFI — I’m hoping to get away with checking in the generated target language sources, so I’m currently focusing on the dependencies in the rest of the code base.

### DIFF
--- a/fixtures/uitests/Cargo.toml
+++ b/fixtures/uitests/Cargo.toml
@@ -11,7 +11,7 @@ name = "uniffi_uitests"
 
 [dependencies]
 uniffi = { workspace = true }
-uniffi_macros = {path = "../../uniffi_macros"}
+uniffi_macros = { path = "../../uniffi_macros", features = ["trybuild"] }
 thiserror = "1.0"
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


mhammond:
This seems fine to me once the conflicts and CI failures are dealt with.

mgeisler:
> This seems fine to me once the conflicts and CI failures are dealt with.

Thanks a lot!

I've actually been struggling to get `cargo test` to run cleanly locally (in general, not just for this PR), so I've been relying on the CI tests to catch mistakes.

Here, the problem seems to be a Serde dependency which doens't enable the `derive` feature (despite deriving `Serialize` and `Deserialize` a lot). I've fixed it now in the commit and updated the commit message — please let me know if you would prefer me to split the commit into two.

mhammond (approved):

PR #2024, opened 2024-03-11
Allow enum discriminants to be signed.

mhammond:
Also now generates an error rather then silently ignoring when a discriminant value is an expression or anything else we can't handle.

### DIFF
--- a/fixtures/enum-types/tests/bindings/test_enum_types.kts
+++ b/fixtures/enum-types/tests/bindings/test_enum_types.kts
@@ -10,3 +10,6 @@ assert(AnimalUInt.CAT.value == 4.toUByte())
 
 assert(AnimalLargeUInt.DOG.value == 4294967298.toULong())
 assert(AnimalLargeUInt.CAT.value == 4294967299.toULong())
+
+// could check `value == (-3).toByte()` but that's ugly :)
+assert(AnimalSignedInt.DOG.value + 3 == 0)
--- a/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
@@ -545,29 +545,33 @@ mod filters {
         Ok(as_ct.as_codetype().literal(literal, ci))
     }
 
+    // Get the idiomatic Kotlin rendering of an integer.
+    fn int_literal(t: &Option<Type>, base10: String) -> Result<String, askama::Error> {
+        if let Some(t) = t {
+            match t {
+                Type::Int8 | Type::Int16 | Type::Int32 | Type::Int64 => Ok(base10),
+                Type::UInt8 | Type::UInt16 | Type::UInt32 | Type::UInt64 => Ok(base10 + "u"),
+                _ => Err(askama::Error::Custom(Box::new(UniFFIError::new(
+                    "Only ints are supported.".to_string(),
+                )))),
+            }
+        } else {
+            Err(askama::Error::Custom(Box::new(UniFFIError::new(
+                "Enum hasn't defined a repr".to_string(),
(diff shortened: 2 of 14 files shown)


uniffi_bindgen/src/interface/enum_.rs
```diff
@@ -221,18 +221,26 @@ impl Enum {
...
+                    next,
+                    if (next as i64) < 0 {
+                        Literal::new_int(next as i64)
+                    } else {
+                        Literal::new_uint(next)
+                    },
+                ),
+                Some(Literal::UInt(v, _, _)) => (v, Literal::new_uint(v)),
+                // in-practice, Literal::Int == a negative number.
+                Some(Literal::Int(v, _, _)) => (v as u64, Literal::new_int(v)),
```
skhamis, line 238: Not sure the need for converting to u64 here?

skhamis (commented):
Nice! Glad we got support for signed quickly! This looks great, just two small questions but otherwise LGTM!

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -545,29 +545,33 @@ mod filters {
         Ok(as_ct.as_codetype().literal(literal, ci))
     }
 
+    // Get the idiomatic Kotlin rendering of an integer.
+    fn int_literal(t: &Option<Type>, base10: String) -> Result<String, askama::Error> {
+        if let Some(t) = t {
+            match t {
+                Type::Int8 | Type::Int16 | Type::Int32 | Type::Int64 => Ok(base10),
+                Type::UInt8 | Type::UInt16 | Type::UInt32 | Type::UInt64 => Ok(base10 + "u"),
```
bendk, line 553: Is it worth double-checking that `base10` is really base-10 digits?  I don't think it's currently possible, but things might get confusing if it started with something like `0x` .

fixtures/enum-types/tests/bindings/test_enum_types.kts
```diff
@@ -10,3 +10,6 @@ assert(AnimalUInt.CAT.value == 4.toUByte())
 
 assert(AnimalLargeUInt.DOG.value == 4294967298.toULong())
 assert(AnimalLargeUInt.CAT.value == 4294967299.toULong())
+
+// markh can't find a better way to check for -3! :(
+assert(AnimalSignedInt.DOG.value + 3 == 0)
```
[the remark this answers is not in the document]
bendk: I half-remember having to do things like `(-3).toShort()` to get this to work.

bendk (approved):
Looks good to me, the code changes were not bad at all.

uniffi_bindgen/src/interface/enum_.rs
mhammond: Every other arm creates a tuple of `(u64, Literal)` and Rust refuses to silently cast an `i64` to a `u64`, hence this explicit cast is needed.

fixtures/enum-types/tests/bindings/test_enum_types.kts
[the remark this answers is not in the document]
mhammond: ah yeah, thanks - in this case it's `toByte()`. That's still ugly though, so I left it as-is but updated the comment.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
mhammond: That comes from https://docs.rs/syn/latest/syn/struct.LitInt.html#method.base10_digits, which is why I stuck with the name. I haven't experimented, but the docs for that seem clear.

PR #2029, opened 2024-03-13
Remove some unnecessary trailing whitespace

cowlicks:
Noticed these while inspecting the generated files


bendk (approved):
Looks great, thanks!

PR #2032, opened 2024-03-15
`Error` creates Kotlin inheritance loop

mgeisler:
If a Rust enum is named `Error`, it becomes `Exception` in the generated Kotlin. The problem is that this class inherits from `Exception`, which I think it supposed to be a built-in Kotlin type:

```kotlin
sealed class Exception : Exception() {
    class Foo() : Exception() {
        override val message
            get() = ""
    }

    companion object ErrorHandler : UniffiRustCallStatusErrorHandler<Exception> {
        override fun lift(error_buf: RustBuffer.ByValue): Exception = FfiConverterTypeError.lift(error_buf)
    }
}
```

If I use `FooError` instead on the Rust side, the code above changes to `class FooException : Exception`.

### DIFF
--- a/fixtures/keywords/kotlin/src/lib.rs
+++ b/fixtures/keywords/kotlin/src/lib.rs
@@ -56,4 +56,14 @@ pub enum fun {
     class { object: u8 },
 }
 
+// `FooError` is turned into `FooException`, so this enum will end up
+// being named just `Exception`. Without special care, this will clash
+// with `kotlin.Exception` class.
+#[allow(non_camel_case_types)]
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("class?")]
+    class { object: u8 },
+}
+
 uniffi::include_scaffolding!("keywords");
--- a/fixtures/keywords/kotlin/src/keywords.udl
+++ b/fixtures/keywords/kotlin/src/keywords.udl
@@ -40,3 +40,8 @@ enum class {
 interface fun {
    class(u8 object);
 };
+
+[Error]
+interface Error {
+   class(u8 object);
+};
--- a/uniffi_bindgen/src/bindings/kotlin/templates/Async.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/Async.kt
@@ -12,7 +12,7 @@ internal object uniffiRustFutureContinuationCallbackImpl: UniffiRustFutureContin
     }
 }
 
-internal suspend fun<T, F, E: Exception> uniffiRustCallAsync(
+internal suspend fun<T, F, E: kotlin.Exception> uniffiRustCallAsync(
     rustFuture: Long,
     pollFunc: (Long, UniffiRustFutureContinuationCallback, Long) -> Unit,
     completeFunc: (Long, UniffiRustCallStatus) -> F,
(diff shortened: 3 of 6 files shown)


mgeisler:
This PR is just to reproduce what looks like an error to me — I hope the tests will show an error when trying to compile this.

mhammond:
I just put up https://github.com/mozilla/uniffi-rs/pull/2033 which side-steps a similarly sounding problem - I wonder if using `kotlin.Exception` or similar (ie, the fully qualified name) can fix this?

mgeisler:
Oh, that looks great! That will fix out https://github.com/awslabs/mls-rs/issues/98.

I was just googling how to quality the `Exception` name... now that I know about the `kotlin` top-level namespace (or package?), this should be easy! I'll amend the PR to fix this as well!

mhammond:
I think you undid the change you need? ie, the PR now shows the only file changed as being the test.

mgeisler:
Yes... I wanted the CI to fail to verify that the change fixes it. Now the [CI passed](https://app.circleci.com/pipelines/github/mozilla/uniffi-rs/4287/workflows/58e6af8a-58f7-4629-8e14-0b26f248e893), so I guess I need to look more at this :smile:

mgeisler:
I haven't set up Kotlin on my machine yet, so I haven't been able to test this locally. I will look more into this next week.

fixtures/keywords/kotlin/src/lib.rs
```diff
@@ -56,4 +56,13 @@ pub enum fun {
     class { object: u8 },
 }
 
+// `FooError` is turned into `FooException`, so this enum will end up
+// being named just `Exception`. Without special care, this will clash
+// with `kotlin.Exception` class.
+#[derive(Debug, thiserror::Error)]
```
mhammond, line 62: I think this is just missing the uniffi derive?

mgeisler: Oh, thanks! I've now tried [with the derive](https://app.circleci.com/pipelines/github/mozilla/uniffi-rs/4289/workflows/7c0d86e1-8069-44f8-a428-45799ae727f4) and with [changes to the UDL file](https://app.circleci.com/pipelines/github/mozilla/uniffi-rs/4290/workflows/339d1fd6-814e-41fe-8c40-59ef4315a365) and both work.

If I understand correct, it should be okay to mix proc-macros and the UDL file, so I would have expected both of them to fail?

mhammond:
huh, that's odd.

mgeisler:
> huh, that's odd.

Yeah... I don't know why I could not reproduce the error in the CI tests. I've now gotten myself a `kotlinc` to test locally and I can easily reproduce the error when with the [`Error` enum here](https://github.com/mgeisler/mls-rs/blob/main/mls-rs-uniffi/src/lib.rs).

This also allowed me to find a few more spots where `Exception` should be turned into `kotlin.Exception`. Now that I know where to grab a bare-bones Kotlin install, I'll setup CI in the mls-rs repository so I can ensure things keep compiling cleanly.

mhammond (approved):
It's a shame we can't repro it, but this is also making type references consistent with something I landed recently, so I see no harm in taking this as is - thanks!

mgeisler:
Thanks! Yeah, I would have loved to reproduce it here, but there's something small that I'm overlooking :smile: 

At least there will soon be more CI testing via the mls-rs project.

PR #2036, opened 2024-03-18
fix: require Android API Level 34 for SystemCleaner

### LINKED ISSUE #2035: Instantiating Cleaner causes crash on Android 13
Related to element-hq/element-x-android#2346

When launching an Element X version that relies on uniffi-rs v0.26.0 or higher on a Pixel 4a running GrapheneOS based on Android 13, the app crashes with the following error:
```
java.lang.AssertionError: Erroneously trying to create Cleaner in zygote
```
The full crash log is available here: [crash.log](https://github.com/mozilla/uniffi-rs/files/14635727/crash.log)

Based on the AOSP commit history and GrapheneOS's commit history, this AssertionError was removed starting with Android 14 (API level 34), so it seems that using `AndroidSystemCleaner` is not supported on all Android OSes with API level 33, especially GrapheneOS.
#1996 updated the min API level for the new Cleaner to 33, it might need to be raised to 34.

SamyDjemai:
Fixes #2035

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/ObjectCleanerHelperAndroid.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/ObjectCleanerHelperAndroid.kt
@@ -2,14 +2,15 @@
 {{- self.add_import("androidx.annotation.RequiresApi") }}
 
 private fun UniffiCleaner.Companion.create(): UniffiCleaner =
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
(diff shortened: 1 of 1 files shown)


mhammond:
I don't see any problem here - @julian-baumann, does this look OK to you?

julian-baumann:
> I don't see any problem here - @julian-baumann, does this look OK to you?

That looks fine to me, though I should mention I don't have much experience with Android development at all.

mhammond (approved):
Thanks for the update - let's take it and see what happens :)

PR #2040, opened 2024-03-19
Blocking task queue without specialized code

bendk:
When discussing  #1837, @mhammond asked me a great question:  should  blocking task queues be UniFFI's responsibility?  This is my attempt to see what would happen if we decided it wasn't.

The code here implements more-or-less the same feature as #1837, but without changing any UniFFI code.  Instead it leverages async callback methods to implement it.

Benefit of this approach:
  - No new code needed
  - No new features to commit to supporting
  - Somewhat more flexible.  With #1837, Python users need to use `futures.executor.ThreadPoolExecutor`.  With this approach, they can use whatever technology they want.

Downsides of this approach:
 - Tasks need to be `'static` and cannot use references from their local scope.  [This makes the code somewhat awkward](https://github.com/bendk/uniffi-rs/blob/blocking-task-queue-without-specialized-code/examples/async-api-client/src/api_client.rs#L48-L52).  Here's [similar code](https://github.com/bendk/uniffi-rs/blob/blocking-task-queue/fixtures/futures/src/lib.rs#L395) in the other PR, note that it could stay basically the same even it used `&self` or any other reference.
 - It requires users to write [boilerplate code](https://github.com/bendk/uniffi-rs/blob/blocking-task-queue-without-specialized-code/examples/async-api-client/src/tasks.rs).  Maybe something like this could be rolled into uniffi or a support library.
 - It requires locking a Mutex and allocating an Arc.

So, should we try to implement #1837 or not?  My feeling is that it's not clear, so for `0.27` maybe we should just merge this one and let that one hang out for a while.

### DIFF
--- a/examples/async-api-client/README.md
+++ b/examples/async-api-client/README.md
@@ -1,0 +1,28 @@
+This crate is a toy build an async API client, with some parts implemented in Rust and some parts
+implemented in the foreign language.  Each side makes async calls across the FFI.
+
+The motivation is to show how to build an async-based Rust library, using a foreign async executor to drive the futures.
+Note that the Rust code does not start any threads of its own, nor does it use startup an async runtime like tokio.
+Instead, it awaits async calls to the foreign code and the foreign executor manages the threads.
+
+There are two basic ways the Rust code in this crate awaits the foreign code:
+
+## API calls
+
+API calls are the simple case.
+Rust awaits an HTTP call to the foreign side, then uses `serde` to parse the JSON into a structured response.
+As long as the Rust code is "non-blocking" this system should work fine.
+Note: there is not a strict definition for "non-blocking", but typically it means not performing IO and not executing a long-running CPU operation.
+
+## Blocking tasks
+
+The more difficult case is a blocking Rust call.
+The example from this crate is reading the API credentials from disk.
+The `tasks.rs` module and the foreign implementations of the `TaskRunner` interface are an experiment to show how this can be accomplished using async callback methods.
+
+The code works, but is a bit clunky.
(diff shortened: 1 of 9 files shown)


Sajjon:
The link to the "boilerplate code" seems broken, the commit does not exist?

bendk:
> The link to the "boilerplate code" seems broken, the commit does not exist?

I think it should work now.

mhammond (approved):
> Benefit of this approach:
> No new code needed
> No new features to commit to supporting
> Somewhat more flexible.

IMO, that makes this a bit of a no-brainer.

> With
https://github.com/mozilla/uniffi-rs/pull/1837, Python users need to use futures.executor.ThreadPoolExecutor. With this approach, they can use whatever technology they want.

TBH, this is my main concern about #1837 - I'm really not sure we are making correct assumptions there, which is best highlighted by this statement and example.

We don't really have a great actual use-case for #1837 yet, and I think it's fine that the first app with a real use-case needs to do some hacky things like that example - it means we can then learn from actual experience and that can help inform whether the approach we are taking there makes sense. Further, there also seems a slight risk that the correct approach might not be a one-size-fits-all - different bindings might have different requirements - ie, the approach in that PR might be perfect for (say) Kotlin, but not at all correct for (say) golang, which might mean consumers end up needing to avoid that mechanism anyway.

So I'm in favor of landing this and keeping #1837 on the back-burner until we get more signals indicating it's a good approach.

Maybe add a new README to the example explaining how it's an experimental, aspirational approach to blocking task queues to help inform how uniffi might better support it natively, with a link to that PR?

bendk:
Yeah, that seems right to me.  I just added a README.  I'll merge this one and remove the `0.27` label from #1837.

Issue #2041, opened 2024-03-19
`Lift` / `LiftRef` error with bad compilation error for large enums


Sajjon:
Follow-up [on thread on Matrix](https://matrix.to/#/!tsNDDGHvNGVNmvOOzi:mozilla.org/$pRoFZYH9at1LGsmrliYwSsKBj9JardJ513vvp3w9WgQ?via=mozilla.org&via=matrix.org&via=element.io) with @mhammond 

This fairly large (107 cases) error enum (see [source here]()https://github.com/radixdlt/sargon/blob/develop/src/core/error/common_error.rs#L10):

```rust
#[repr(u32)]
#[derive(Clone, Debug, ThisError, PartialEq, uniffi::Error)]
#[uniffi(flat_error)]
pub enum CommonError2 {
 // 106 cases omitted
    #[error("String not hex '{bad_value}'")]
    StringNotHex { bad_value: String } = 10011,
}
```

results in UniFFI unable to compile when I add a `#[uniffi::export]`-ed function taking the error enum as argument:
```rust
#[uniffi::export]
pub fn error_message_from_error(error: &CommonError2) -> String {
    format!("{}", error)
}
```

I suspect this method triggers the `LiftRef` to be used, which has not been used before. **To be clear, I'm already using the error enum, it has been UniFFI converted to Swift and Kotlin and I have tested using bindgen methods returning (throwing) the error, and that works...**. But UniFFI gets sad when I write Rust code requiring LIFT (and I suspect LOWER).

The compilation error is not very informative, it does not say what has failed, or at least I don't understand it...:

```sh
error[E0277]: the trait bound `common_error::CommonError2: LiftRef<UniFfiTag>` is not satisfied
   --> src/core/error/common_error.rs:789:1
    |
789 | #[uniffi::export]
    | ^^^^^^^^^^^^^^^^^ the trait `LiftRef<UniFfiTag>` is not implemented for `common_error::CommonError2`
    |
    = help: the following other types implement trait `LiftRef<UT>`:
              <bool as LiftRef<UT>>
              <i8 as LiftRef<UT>>
              <i16 as LiftRef<UT>>
              <i32 as LiftRef<UT>>
              <i64 as LiftRef<UT>>
              <u8 as LiftRef<UT>>
              <u16 as LiftRef<UT>>
              <u32 as LiftRef<UT>>
            and 215 others
[...]

PR #2042, opened 2024-03-19
Increase BUF_SIZE to 16384 | Allowing large errors 

### LINKED ISSUE #1968: Long docstrings result in build assertion failure.
I'm working with a [`#[derive(uniffi::Error)]` enum that's quite large](https://github.com/liveview-native/phoenix-channels-client/blob/416a3e52606a70775fd4cf0d349c6af7b0fba974/src/ffi/io/error.rs#L4-L221).

It was mentioned in the matrix chat that these docstrings end up in the staticlib binary and bloat the size so it's not ideal to have a docstring this large.

Anyway, an extra long docstrings result in:
```
error[E0080]: evaluation of constant value failed
   --> /Users/sebastian.imlay/.cargo/registry/src/index.crates.io-6f17d22bba15001f/uniffi_core-0.26.1/src/metadata.rs:210:9
    |
210 |         assert!(self.size + string.len() + 1 < BUF_SIZE);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'assertion failed: self.size + string.len() + 1 < BUF_SIZE', /Users/sebastian.imlay/.cargo/registry/src/index.crates.io-6f17d22bba15001f/uniffi_core-0.26.1/src/metadata.rs:210:9
    |
note: inside `MetadataBuffer::concat_long_str`
   --> /Users/sebastian.imlay/.cargo/registry/src/index.crates.io-6f17d22bba15001f/uniffi_core-0.26.1/src/metadata.rs:210:9
    |
210 |         assert!(self.size + string.len() + 1 < BUF_SIZE);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Reproduction steps:
`cargo new --lib uniffi-docstring-issue && cd uniffi-docstring-issue && cargo add uniffi`

Put a lot of docstrings in `src/lib.rs`. In my case:
```rust

uniffi::setup_scaffolding!("too_much_docstring");

/// [std::io::Error] and [std::io::ErrorKind], but with `uniffi` support.
#[derive(Clone, Debug, uniffi::Error)]
pub enum IoError {
    /// An entity was not found, often a file.
    NotFound,
    /// The operation lacked the necessary privileges to complete.
    PermissionDenied,
    /// The connection was refused by the remote server.
    ConnectionRefused,
    /// The connection was reset by the remote server.
    ConnectionReset,
    /// The remote host is not reachable.
    HostUnreachable,
    /// The network containing the remote host is not reachable.
    NetworkUnreachable,
    /// The connection was aborted (terminated) by the remote server.
    ConnectionAborted,
    /// The network operation failed because it was not connected yet.
    NotConnected,
    /// A socket address could not be bound because the address is already in
    /// use elsewhere.
    AddrInUse,
    /// A nonexistent interface was requested or the requested address was not
    /// local.
    AddrNotAvailable,
    /// The system's networking is down.
    NetworkDown,
    /// The operation failed because a pipe was closed.
    BrokenPipe,
    /// An entity already exists, often a file.
    AlreadyExists,
    /// The operation needs to block to complete, but the blocking operation was
    /// requested to not occur.
    WouldBlock,
    /// A filesystem object is, unexpectedly, not a directory.
    ///
    /// For example, a filesystem path was specified where one of the intermediate directory
    /// components was, in fact, a plain file.
    NotADirectory,
    /// The filesystem object is, unexpectedly, a directory.
    ///
    /// A directory was specified when a non-directory was expected.
    IsADirectory,
    /// A non-empty directory was specified where an empty directory was expected.
    DirectoryNotEmpty,
    /// The filesystem or storage medium is read-only, but a write operation was attempted.
    ReadOnlyFilesystem,
    /// Loop in the filesystem or IO subsystem; often, too many levels of symbolic links.
    ///
    /// There was a loop (or excessively long chain) resolving a filesystem object
    /// or file IO object.
    ///
    /// On Unix this is usually the result of a symbolic link loop; or, of exceeding the
    /// system-specific limit on the depth of symlink traversal.
    FilesystemLoop,
    /// Stale network file handle.
    ///
    /// With some network filesystems, notably NFS, an open file (or directory) can be invalidated
    /// by problems with the network or server.
    StaleNetworkFileHandle,
    /// A parameter was incorrect.
    InvalidInput,
    /// Data not valid for the operation were encountered.
    ///
    /// Unlike [`InvalidInput`], this typically means that the operation
    /// parameters were valid, however the error was caused by malformed
    /// input data.
    ///
    /// For example, a function that reads a file into a string will error with
    /// `InvalidData` if the file's contents are not valid UTF-8.
    ///
    /// [`InvalidInput`]: ErrorKind::InvalidInput
    InvalidData,
    /// The I/O operation's timeout expired, causing it to be canceled.
    TimedOut,
    /// An error returned when an operation could not be completed because a
    /// call to [`write`] returned [`Ok(0)`].
    ///
    /// This typically means that an operation could only succeed if it wrote a
    /// particular number of bytes but only a smaller number of bytes could be
    /// written.
    ///
    /// [`write`]: std::io::Write::write
    /// [`Ok(0)`]: Ok
    WriteZero,
    /// The underlying storage (typically, a filesystem) is full.
    ///
    /// This does not include out of quota errors.
    StorageFull,
    /// Seek on unseekable file.
    ///
    /// Seeking was attempted on an open file handle which is not suitable for seeking - for
    /// example, on Unix, a named pipe opened with `File::open`.
    NotSeekable,
    /// Filesystem quota was exceeded.
    FilesystemQuotaExceeded,
    /// File larger than allowed or supported.
    ///
    /// This might arise from a hard limit of the underlying filesystem or file access API, or from
    /// an administratively imposed resource limitation.  Simple disk full, and out of quota, have
    /// their own errors.
    FileTooLarge,
    /// Resource is busy.
    ResourceBusy,
    /// Executable file is busy.
    ///
    /// An attempt was made to write to a file which is also in use as a running program.  (Not all
    /// operating systems detect this situation.)
    ExecutableFileBusy,
    /// Deadlock (avoided).
    ///
    /// A file locking operation would result in deadlock.  This situation is typically detected, if
    /// at all, on a best-effort basis.
    Deadlock,
    /// Cross-device or cross-filesystem (hard) link or rename.
    CrossesDevices,
    /// Too many (hard) links to the same filesystem object.
    ///
    /// The filesystem does not support making so many hardlinks to the same file.
    TooManyLinks,
    /// A filename was invalid.
    ///
    /// This error can also cause if it exceeded the filename length limit.
    InvalidFilename,
    /// Program argument list too long.
    ///
    /// When trying to run an external program, a system or process limit on the size of the
    /// arguments would have been exceeded.
    ArgumentListTooLong,
    /// This operation was interrupted.
    ///
    /// Interrupted operations can typically be retried.
    Interrupted,

    /// This operation is unsupported on this platform.
    ///
    /// This means that the operation can never succeed.
    Unsupported,

    // ErrorKinds which are primarily categorisations for OS error
    // codes should be added above.
    //
    /// An error returned when an operation could not be completed because an
    /// "end of file" was reached prematurely.
    ///
    /// This typically means that an operation could only succeed if it read a
    /// particular number of bytes but only a smaller number of bytes could be
    /// read.
    UnexpectedEof,

    /// An operation could not be completed, because it failed
    /// to allocate enough memory.
    OutOfMemory,

    // "Unusual" error kinds which do not correspond simply to (sets
    // of) OS error codes, should be added just above this comment.
    // `Other` and `Uncategorized` should remain at the end:
    //
    /// A custom error that does not fall under any other I/O error kind.
    ///
    /// This can be used to construct your own [`IoError`]s that do not match any
    /// [`ErrorKind`].
    ///
    /// This [`ErrorKind`] is not used by the standard library.
    ///
    /// Errors from the standard library that do not fall under any of the I/O
    /// error kinds cannot be `match`ed on, and will only match a wildcard (`_`) pattern.
    /// New [`ErrorKind`]s might be added in the future for some of those.
    Other,

    /// Any I/O error from the standard library that's not part of this list.
    ///
    /// Errors that are `Uncategorized` now may move to a different or a new
    /// [`ErrorKind`] variant in the future. It is not recommended to match
    /// an error against `Uncategorized`; use a wildcard match (`_`) instead.

    Uncategorized,
}

```

Sajjon:
Increase `BUF_SIZE` to `16,384` (from `4,096`), allowing large errors (enums) to be used. 

Add fixture which unit tests fails even for `BUF_SIZE` of `8,198`, but works with 16k.

Fixes #1968 and #2041 (see those for details and rationale)


Sajjon:
I can also create a related fixture with fewer associated values, but more cases, maybe called "huge error", with cirka 1000 cases, but not so many large associated values, and see if 16k is enough for that, if you want?

fixtures/large-error/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "large-error"
+version = "0.26.1"
+authors = ["Alexander Cyon <alex.cyon@gmail.com>"]
```
Sajjon, line 4: Let me know if you want this to say `["Firefox Sync Team <sync-team@mozilla.com>"]` instead....

fixtures/large-error/src/lib.rs
```diff
@@ -0,0 +1,701 @@
...
+
+#[derive(Clone, Debug, Record)]
+pub struct ErrorWithContext {
+    pub error: LargeError,
+    pub context: String,
+}
+
+#[uniffi::export]
+pub fn error_message_from_error(error: &LargeError) -> String {
+    format!("{}", error)
```
Sajjon, line 652: this and `error_code_from_error` fails to compile without the `BUF_SIZE` increment.

uniffi_core/src/metadata.rs
```diff
@@ -80,7 +80,7 @@ pub mod codes {
     pub const LIT_NULL: u8 = 4;
 }
 
-const BUF_SIZE: usize = 4096;
+const BUF_SIZE: usize = 16384;
```
Sajjon, line 83: Let me know if you wanna use `8196` instead, and then I will need to decrease the amount of variants in the large error enum.

fixtures/large-error/src/large_error.udl
```diff
@@ -0,0 +1 @@
+namespace large_error {};
```
mhammond, line 1: nano-nit - please add a trailing newline here.

uniffi_core/src/metadata.rs
mhammond: Do you mind adding a semi-terse comment here linking to the 2 issues you linked in the new fixture?

mhammond (changes requested):
This seems fine to me - just a couple of nits, but I see no reason to not take this.

fixtures/large-error/Cargo.toml
mhammond: I'm fine with giving you the credit. Note however that if you choose it can be omitted entirely.

uniffi_core/src/metadata.rs
Sajjon: Fixed, check if wording is ok :)

mhammond (approved):
Thanks!

PR #2044, opened 2024-03-21
Update toml dependency to latest version

mgeisler:
I am in the process of vendoring UniFFI in AOSP, which means that I have to make it play nice with the existing crates found here:

  https://cs.android.com/android/platform/superproject/+/main:external/rust/crates/

Android tries to only vendor each crate in a single version, so it would be helpful for me if the toml dependency could be updated.

This updates the “Requires” version to 1.70. I believe this is okay since this version is below 1.74 which is what mozilla-central currently requires.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,7 @@ commands:
   # Our minimum supported rust version is specified here.
   prepare-rust-min-version:
     steps:
-      - run: rustup override set 1.66.0
+      - run: rustup override set 1.70.0
       - run: rustup update
   build-api-docs:
     steps:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -567,6 +567,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a26ae43d7bcc3b814de94796a5e736d4029efb0ee900c12e2d54c993ad1a1e07"
 
 [[package]]
+name = "equivalent"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5443807d6dff69373d433ab9ef5378ad8df50ca6298caf15de6e52e24aaf54d5"
+
+[[package]]
 name = "errno"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


badboy (changes requested):
Unfortunately this causes problems for mozilla-central, so we need to hold off on this for at least the next release.

mgeisler:
> Unfortunately this causes problems for mozilla-central, so we need to hold off on this for at least the next release.

Ah, thanks for looking! I tried to read rust-versions.md and it points to [`util.py`](https://searchfox.org/mozilla-central/source/python/mozboot/mozboot/util.py#14) as the definition of the version mozilla-central supports. Since this is 1.74, I was hoping it would be okay to bump up the version used in CI here?

There's probably something I'm missing?

mhammond:
I think the problem is toml - ie, https://searchfox.org/mozilla-central/rev/f63ca2952da98e0817bdae0ddf1314281a497106/Cargo.lock#5755-5762 - bumping that version here would probably mean m-c ends up with 2 versions of that crate which we try hard to avoid.

badboy:
Yup, sorry, should have been more precise. We also try to avoid dupe crates in tree.
I'm already trying to get current UniFFI in. I don't mind fixing up the other toml uses, but prefer doing that after we get in what we have now to keep the required changes I have to do smaller.

mgeisler:
Thanks both for explaining! So it seems we're in the same boat :sweat_smile: Just with different trees of crates to update.

On my side in Android, I'll just carry the small patch here to make uniffi_macros build with a recent toml dependency. If/when the dependency is updated here and in m-c, then I'll notice on the next import and drop the patch.

We can close the PR here if you like since you're obviously already on top of this.

badboy:
Let's leave it open, I hope we can cut a uniffi release next week, and after that we can land this.

badboy:
With the release out we can now make progress on this.
It's relatively simple to pull the new version into mozilla-central, we can patch over the few uses of the old crate (and only UniFFI needs to change the API call really).
However this does pull in a couple new crates that we will need to audit (using [`cargo-vet`](https://github.com/mozilla/cargo-vet/)) (uniffi does not [yet] use cargo-vet, but Glean and mozilla-central do).

```
5 unvetted dependencies:
  serde_spanned:0.6.5 missing ["safe-to-deploy"]
  toml:0.8.12 missing ["safe-to-deploy"]
  toml_datetime:0.6.5 missing ["safe-to-deploy"]
  toml_edit:0.22.9 missing ["safe-to-deploy"]
  winnow:0.6.5 missing ["safe-to-deploy"]
```

We pull in other trusted audits from Google (via https://github.com/google/rust-crate-audits).
Would you be able to do those audits? That would speed things along.

mgeisler:
> We pull in other trusted audits from Google (via https://github.com/google/rust-crate-audits).
> Would you be able to do those audits? That would speed things along.

Yes, I like to help here! It would be useful for me to become more familiar with `cargo vet`. I'll put it on my TODO and try to come back to it in a few days.

mgeisler:
I looked a bit at the `cargo vet` reviews today, but didn't immediately find a place for me to put them so that they end up in https://github.com/google/rust-crate-audits. I'll have to look some more!

mhammond:
m-c is still on 0.5. Is it worth landing a semver range? Let's us close this and works when the inevitable update does happen...

mgeisler:
Yes, happy to close this!

PR #2045, opened 2024-03-23
Swift: Make it possible to mark value types as `Sendable` with `sendable_value_types` config

Sajjon:
Make it possible to mark Swift value types as [`Sendable`](https://developer.apple.com/documentation/swift/sendable) - which is needed for Swift 6.

* Add new config for Swift: `sendable_value_types` (false by default)
* which if true (`sendable_value_types: true`) will mark:
  - (`[uniffi::Record]`) Swift `struct`s which does not contain object references as `Sendable`
  - (`[uniffi::Enum]`) Swift `enums`s which does not contain object references as `Sendable`

This will make it possible for Swift devs to [skip having to add `@unchecked Sendable` all over the place](https://github.com/search?q=repo%3Aradixdlt%2Fsargon%20%40unchecked%20Sendable&type=code)

I was unable to add a meaningful fixture though, since it is not possible to "unit test" this. The only way to "test" it is to ensure that compilation fails... which would be possible if:
* I could [enable Strict Concurrency check](https://www.swift.org/documentation/concurrency/)...
*  ... AND change warnings => errors. 
* ...[here is a failed attempt to do so](https://github.com/Sajjon/uniffi-rs/pull/1/files) - the problem is I can't seem to be able to pass `strict-concurrency=complete` to `swiftc` for bindgen test.

# Alternatives to consider
AND/OR we can add a different set of (Swift only) config flags: `frozen_struct` and `frozen_enum` (or simply `frozen_value_types`) - because types marked `@frozen` [automatically becomes `Sendable`](https://developer.apple.com/documentation/swift/sendable#Sendable-Structures-and-Enumerations), but I think it valuable to not have to mark a value type as `@frozen` just to get `Sendable`. I can do another PR for the `@frozen` annotation if you'd like?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@
 
 - Proc-macro record defaults now support empty vecs and Some values.
 
+- Swift: Records and Enums without object references can now be made `Sendable` Swift,
+  by opting in to new Configuration `experimental_sendable_value_types` in `uniffi.toml`.
+
 ### What's fixed?
  
 - Fixed a memory leak in callback interface handling.
--- a/docs/manual/src/swift/configuration.md
+++ b/docs/manual/src/swift/configuration.md
@@ -4,17 +4,20 @@ The generated Swift module can be configured using a `uniffi.toml` configuration
 
 ## Available options
 
-| Configuration name | Default  | Description |
-| ------------------ | -------  |------------ |
-| `cdylib_name`      | `uniffi_{namespace}`[^1] | The name of the compiled Rust library containing the FFI implementation (not needed when using `generate --library`). |
-| `module_name`      | `{namespace}`[^1] | The name of the Swift module containing the high-level foreign-language bindings. |
-| `ffi_module_name`  | `{module_name}FFI` | The name of the lower-level C module containing the FFI declarations. |
-| `ffi_module_filename` | `{ffi_module_name}` | The filename stem for the lower-level C module containing the FFI declarations. |
-| `generate_module_map` | `true` | Whether to generate a `.modulemap` file for the lower-level C module with FFI declarations. |
-| `omit_argument_labels` | `false` | Whether to omit argument labels in Swift function definitions. |
-| `generate_immutable_records` | `false` | Whether to generate records with immutable fields (`let` instead of `var`). |
-| `custom_types`      | | A map which controls how custom types are exposed to Swift. See the [custom types section of the manual](../udl/custom_types.md#custom-types-in-the-bindings-code)|
+The configurations prefixed with `experimental_` should be regarded as unstable and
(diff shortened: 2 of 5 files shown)


bendk:
Very interesting.  I want to understand `Sendable` in Swift more before I can really comment on this one.

My first question is what's the difference between it and Rust's `Send` trait?  On the Rust side, all types are required to be `Send`, so why can't we unconditionally mark everything send?

Second question: what's the interaction between `@frozen` and `Sendable`?  Why does marking value types `@frozen` automatically make them `Sendable`?

Sajjon:
Unconditionally marking all value types as Sendable is better than not doing it - I just through letting it be opt it would make it more likely to get this PR in...

I'm not convinced adding it unconditionally for reference types (Swift classes) is a good idea though, maybe that at least should be opt in. But I did not add impl of Sendable for reference types since implementation is trickier - and since _personally_ I dont need it. 

Regarding `@frozen` this text in the [original Swift Evolution proposal (SE-0302)](https://github.com/apple/swift-evolution/blob/main/proposals/0302-concurrent-value-and-concurrent-closures.md#implicit-structenum-conformance-to-sendable) of `Sendable` explains it quite good:

> Public non-frozen structs and enums do not get an implicit conformance, because doing so would present a problem for API resilience: the implicit conformance to Sendable would become part of the contract with clients of the API, even if it was not intended to be

mhammond:
> However, this makes a lot of sense to me. I certainly feel more comfortable approving this if it's behind a config value.

We could even go a little further and name the config value with an `experimental_` or similar prefix, which would give us license to change either the name, type or semantics of that config pref over the next few releases?

Sajjon:
Cool! I will rename flag with _experimental_ prefix and update changelog in a couple of hours

Sajjon:
@bendk @mhammond let me know if you want me to change the wording in the markdown files to something else. Otherwise, ready from my point of view.

bendk (approved):
Looks great, thanks for updating the PR.

LBeernaertProton:
Do you folks think this could also be extended to uniffi objects?

mhammond:
> Do you folks think this could also be extended to uniffi objects?

Sure - but see the conversation above, it's not clear to some of us if it makes sense to be unconditional etc - we'd welcome input into everything about this - ie, whether it should eventually be the default, whether it should be opt-in/opt-out per "broad type" (ie, enum, object), per "single type" (ie, Foo, Bar), etc.

LBeernaertProton:
Noted. As far as I'm concerned, I'm okay with this being the default for all uniffi objects as well, since the underlying Rust objects are already `Sync+Send`. But perhaps it would be better to leave this as opt in at the moment.

PR #2047, opened 2024-03-24
Fix GC/Cleaner intermittent test failures when run locally.

mhammond:
coveralls fails locally for me more often than not. This change looks saner and has been reliable for me.


bendk (approved):
Yes please!  It'll be great to have the CI reliably green again.

Issue #1726, opened 2023-08-30
Tokio runtimes and async


bendk:
I've been thinking about the current [async_runtime](https://github.com/mozilla/uniffi-rs/blob/main/fixtures/futures/src/lib.rs#L193-L198) support and wondering how useful it is:

- If all you want to do is use tokio types and not manage any threads, it seems better to use versions of those types that are executor-agnostic like `async-mutex` and `async-timer` 
- If you want tokio to manage a threadpool, then async_compat isn't a great solution for this since it doesn't allow you to customize the tokio Runtime.

What if instead of using `async-compat`, we allowed users to specify a function that returns a type that derefs to `tokio::Runtime` and let them build their own runtime?  Something like:

```
use once_cell::sync::Lazy;
use tokio::runtime::{Builder,Runtime};

fn main_runtime() -> &'static Lazy<Runtime> {
    static RUNTIME: Lazy<Runtime> = Lazy::new(|| {
         Builder::new_multi_thread()
             .worker_threads(4)
             .thread_name("my-custom-name")
             .thread_stack_size(3 * 1024 * 1024)
             .build()
             .unwrap()
    });
    &RUNTIME
}

#[uniffi::export(tokio_runtime = main_runtime)]
pub async fn use_shared_resource(options: SharedResourceOptions) -> Result<(), AsyncError> {
    ...
}
```

Then inside the scaffolding function, we do what `async-compat` does and [enter the runtime before polling the future](https://github.com/smol-rs/async-compat/blob/master/src/lib.rs#L300-L307).

I'm not really sure how this should work.  @jplatte @Hywan are you currently using UniFFI async with tokio?  If so, how does it work?

Hywan:
I agree that the current `async_runtime` implementation is limited but it has the merit to work.

What I don't feel clear with your proposal is: how is it supposed to work? It's focusing on defining a `Runtime`, which is a specific type of `tokio` in this case. Each async lib comes with its runtime definition. I believe the correct abstraction level is `Future`, as `async-compat` does (I'm not saying we should stick with it, but I reckon the approach is correct —though too strict/limiteed for now).

We would need a way to express we want to wrap the outgoing `Future` inside another `Future` that can be build in some way. Something like:

```rust
#[uniffi::export(async_wraps_with = future_wrapper)]
pub async fn foo(…) -> … { }

fn future_wrapper<F, T>(future: F) -> F
where F: Future<Output = T>>
{
[...]

PR #2052, opened 2024-03-27
Combine ExportedImplFnArgs and ExportFnArgs

bendk:
These were basically the same, so using 1 struct simplifies the code.

The one difference was that ExportFnArgs allowed an async runtime to be specified.  Now that the 2 are merged, it's possible to specify an async runtime for constructors/methods.  If present that runtime will override the one specified in the impl attribute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ## [[UnreleasedUniFFIVersion]] (backend crates: [[UnreleasedBackendVersion]]) - (_[[ReleaseDate]]_)
 
+### What's changed
+
+- The async runtime can be specified for constructors/methods, this will override the runtime specified at the impl block level.
+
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ## [[UnreleasedUniFFIVersion]] (backend crates: [[UnreleasedBackendVersion]]) - (_[[ReleaseDate]]_)
 
+### What's changed
+
+- The async runtime can be specified for constructors/methods, this will override the runtime specified at the impl block level.
+
```
bendk, line 12: This was the only documentation change since we don't currently document `async_runtime` in the manual, and I'm even sure if we want to continue supporting it (see #1726)

mhammond (approved):

PR #2056, opened 2024-03-30
RustFuture: fix memory leak

QuentinPerez:
Currently, `rust_future_free` doesn't properly release memory, leading to a memory leak. This happens because `get_arc` increases the reference count of RustFuture without decreasing it afterward. By using `consume_handle` we do not increase it anymore before giving the ownership to `ffi_free`.

I have tested this fix in my project and confirmed that it resolves the memory leak. Let me know if I missed something.


mhammond:
closing and reopening to try and convince CI to run.

bendk (approved):
This change looks correct to me and all the tests are passing.  I think it should go into 0.27.1.

I'm not sure what's up with the Circle CI error, but it seems like the tests are passing and they also pass when I run them locally.

PR #2059, opened 2024-04-01
Freshen up the docs for the bindings.

mhammond:
The existing binding docs were getting out of date. I've tried to update them to make then more generally applicable.

### DIFF
--- a/docs/manual/src/internals/rendering_foreign_bindings.md
+++ b/docs/manual/src/internals/rendering_foreign_bindings.md
@@ -25,16 +25,25 @@ Bidings also need to do alot of work to make language identifiers etc work corre
 
 ## Breaking down a Rust function called by Python.
 
-Let's take a quick look at where [Python generates a top-level public function](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py#L37-L40+)
+Let's take a look at where [Python generates a top-level public function](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py#L37-L40+).
 
-It's making, eg `def this_func(this_arg=0) -> None:` - let's break it down:
+This will generate code like the following:
 
-`def {{ func.name()|fn_name }}({%- call py::arg_list_decl(func) -%}) -> None:`
+```
+def this_func(this_arg=0) -> None:
+```
 
-The Askama language does funky things with `{ }` blocks and here we are getting string output into the generated code. The first is the `this_func`:
+Let's break the template down:
 
-`{{ func.name()|fn_name }}`: [Here is `func.name()`](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/interface/function.rs#L72) - you can see all the other metadata about functions there too.
-The string result from that function goes through an Askama "filter" concept to "pipe" the output to [`fn_name`, this rust function](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/python/gen_python/mod.rs#L567) - which ends up just handing the fact it might be a Python keyword but otherwise returns the same value.
+```
+def {{ func.name()|fn_name }}({%- call py::arg_list_decl(func) -%}) -> None:
+```
+
+The Askama language uses double-curly braces (`{ }`) to interpolate blocks of code into the string output.
+
+`{{ func.name()|fn_name }}` becomes `this_func`: [It calls the `name` method on a `Function` object](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/interface/function.rs#L72) (you can see all the other metadata about functions there too).
+Askama uses a "filter" concept: Functions that take the value left of the pipe operator (`|`) to produce a new value.
+The "filter" used in the above template is called`fn_name` and is [defined in the Python bindings generator](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/python/gen_python/mod.rs#L567) - which ends up just handing the fact it might be a Python keyword but otherwise returns the same value.
 
 `{%- call py::arg_list_decl(func) -%}`: Calling an Askama macro, passing the `func` object linked above. It knows how to turn the function arguments into valid Python code.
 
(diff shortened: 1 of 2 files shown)


docs/manual/src/internals/rendering_foreign_bindings.md
```diff
@@ -2,84 +2,69 @@
...
+
+Eg, here's where [Python uses `ci` to iterate over the types](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/python/templates/Types.py#L3)
+
+The templates create foreign-native types for everything from ffi-native types (int/etc) to functions, dictionaries etc. The implementation of these generated types might call your your Rust implemented FFI, as described below.
+
+Bidings also need to do alot of work to make language identifiers etc work correctly - eg, turn `this_func(this_arg: ThisType)` into `thisFunc(...)`
+
+## Breaking down a Rust function called by Python.
+
+Let's take a quick look at where [Python generates a top-level public function](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py#L37-L40+)
```
badboy, line 28:
```suggestion
Let's take a look at where [Python generates a top-level public function](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/python/templates/TopLevelFunctionTemplate.py#L37-L40+).
```

docs/manual/src/internals/rendering_foreign_bindings.md
```diff
@@ -2,84 +2,69 @@
...
+
+It's making, eg `def this_func(this_arg=0) -> None:` - let's break it down:
```
badboy, line 30: I would probably spell this out a bit more:
````suggestion
This will generate code like the following:

```
def this_func(this_arg=0) -> None:
```

Let's break the template down:
````

File: docs/manual/src/internals/rendering_foreign_bindings.md [hunk 3]
```diff
@@ -2,84 +2,69 @@
...
+
+`def {{ func.name()|fn_name }}({%- call py::arg_list_decl(func) -%}) -> None:`
```
badboy, line 32: Not sure how it's rendered, but might be better as a code block instead of inline:

````suggestion
```
def {{ func.name()|fn_name }}({%- call py::arg_list_decl(func) -%}) -> None:
```
````

File: docs/manual/src/internals/rendering_foreign_bindings.md [hunk 4]
```diff
@@ -2,84 +2,69 @@
...
+
+The Askama language does funky things with `{ }` blocks and here we are getting string output into the generated code. The first is the `this_func`:
```
badboy, line 34:
```suggestion
The Askama language uses double-curly braces (`{ }`) to interpolate blocks of code into the string output.
```

docs/manual/src/internals/rendering_foreign_bindings.md
```diff
@@ -2,84 +2,69 @@
...
+
+`{{ func.name()|fn_name }}`: [Here is `func.name()`](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/interface/function.rs#L72) - you can see all the other metadata about functions there too.
```
badboy, line 36:
```suggestion
`{{ func.name()|fn_name }}` becomes `this_func`: [It calls the `name` method on a `Function` object](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/interface/function.rs#L72) (you can see all the other metadata about functions there too).
```

docs/manual/src/internals/rendering_foreign_bindings.md
```diff
@@ -2,84 +2,69 @@
...
+The string result from that function goes through an Askama "filter" concept to "pipe" the output to [`fn_name`, this rust function](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/python/gen_python/mod.rs#L567) - which ends up just handing the fact it might be a Python keyword but otherwise returns the same value.
```
badboy, line 37:
```suggestion
Askama uses a "filter" concept: Functions that take the value left of the pipe operator (`|`) to produce a new value.
The "filter" used in the above template is called`fn_name` and is [defined in the Python bindings generator](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/python/gen_python/mod.rs#L567) - which ends up just handing the fact it might be a Python keyword but otherwise returns the same value.
```

docs/manual/src/internals/rendering_foreign_bindings.md
```diff
@@ -2,84 +2,69 @@
...
+Skipping a few lines ahead in that template, we call the FFI function `{% call py::to_ffi_call(func) %}` - which ultimately
+end up a call to an `extern "C"` FFI function you generated named something like `uniffi_some_name_fn_func_this_func(...)`
+
+The bindings also need to do lots of administrivia - eg, calling initialization functions, importing dependencies, etc
+
+### Implementing the FFI.
+
+[All types](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_meta/src/types.rs#L62) must implement an FFI converter.
+
+The FfiConverter is [described here](./lifting_and_lowering.md)
```
badboy, line 50:
```suggestion
The FfiConverter is described in the [Lifting, Lowering and Serialization](./lifting_and_lowering.md) chapter.
```

I personally dislike "here" as a click target; I will always come around and rewrite your links :P

docs/manual/src/internals/rendering_foreign_bindings.md
```diff
@@ -2,84 +2,69 @@
...
+but tl;dr - this means different things for "native" types (ints etc), but otherwise there's a lot of `RustBuffer`!
```
badboy, line 51:
```suggestion
Note that this means different things for "native" types (`int`, etc), but otherwise there's a lot of `RustBuffer`!
```

docs/manual/src/internals/rendering_foreign_bindings.md
```diff
@@ -2,84 +2,69 @@
...
+
+eg, [the Swift `Bool`](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/swift/templates/BooleanHelper.swift#L1C39-L1C51) vs [Swift record/struct support](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift#L38)
```
badboy, line 53: These newlines turn into large gaps in the rendered doc, I think it's better to keep those sentences connected.
```suggestion
eg, [the Swift `Bool`](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/swift/templates/BooleanHelper.swift#L1C39-L1C51) vs [Swift record/struct support](https://github.com/mozilla/uniffi-rs/blob/884f7865f3367c494e9165e21c1255018577db01/uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift#L38)
```

docs/manual/src/internals/rendering_foreign_bindings.md
```diff
@@ -2,84 +2,69 @@
...
+pub extern "C" fn uniffi_some_name_fn_func_this_func(
+    arg: i32,
+    call_status: &mut ::uniffi::RustCallStatus,
+) -> i32 {
+```
+
+The bindings need to use the metadata to create the correct args to make these calls using the FFI converter implementations.
+
+There will be a number of memory/lifetime/etc "adminstrative" FFI functions that will also be used by the generated implementation.
+
```
badboy, line 70:
```suggestion
```

badboy (approved):
Content-wise this is definitely already an improvement.
I left some stylistic comments, but would like to see this land soon.

mhammond:
Fantastic and well considered changed, thanks @badboy!

PR #2062, opened 2024-04-03
0.27.1

mhammond:
I'm not quite sure I'm doing this right!


QuentinPerez:
Do you think that this change should be part of the release https://github.com/mozilla/uniffi-rs/pull/2056 ?

mhammond:
I guess it probably should, yeah.

bendk (approved):
Looks great, :rocket:

PR #2068, opened 2024-04-10
Add test for custom_newtype Python error

giarc3:
This PR is expected to fail in its current state as a way to demonstrate a possible error in generated Python when using custom newtypes.

Please feel free to close or edit as needed.


mhammond:
Thanks, covered by #2075

PR #2069, opened 2024-04-12
improves the "Describing The Interface" page for clarity

rafaelbeckel:
The proc-macro method is now mentioned upfront, instead of as a bottom note. 

This provides a clearer context for new users.


mhammond (approved):
Thanks

PR #2072, opened 2024-04-15
Rename the 'ext-types/guid' fixture to 'ext-types/custom-types'

mhammond:
I always get a little confused about where exactly our tests are for "custom types" before I re-remember they are hidden inside a crate called 'guid'. I've got some custom type tests and fixes stacked up, and thought I'd fix this first, which will make future patches easier to read.


bendk (approved):
Makes sense to me.

PR #2074, opened 2024-04-15
Ensure custom types are treated by bindings as "types"

mhammond:
This ensures things like the case of the name is correct and they are adjusted when keywords etc.


bendk (approved):
Nice fix, looks good to me.

PR #2075, opened 2024-04-15
Python: Fix custom types generating invalid code when there are forward references

### LINKED ISSUE #2067: Python type aliases don't use forward references
In generated Python, our type aliases are sorted alphabetically and placed towards the end of the file. If there are nested aliases, this fails to compile when they're in the wrong order (`"Foo" is not defined`).
```python
# Type alias
Bar = Foo
...
# Type alias
Foo = str
```
For function type hints, we use quotation marks to allow forward references (described [here](https://peps.python.org/pep-0484/#forward-references)). I tried this for the type aliases, but got lots of warnings `Variable not allowed in type expression` and unknown types when they're used in functions.

I also tried changing the type aliases to 
```python
# Type alias
Bar: typing.TypeAlias = "Foo"
...
# Type alias
Foo: typing.TypeAlias = "str"
```
this worked well, but was only added in Python version 3.10 (is the minimum supported version for uniffi documented somewhere?).

My temporary solution is just to avoid nested aliases, but I have some complex types that would benefit from another solution to this problem.

mhammond:
Also includes many other custom-type tests, all of which work, and clarifies in the docs about which types can be used with custom types.

Fixes #2067


uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -298,6 +298,31 @@ impl<'a> TypeRenderer<'a> {
...
+                    .iter()
+                    .position(|x: &(&str, &Type)| *name == x.1.as_codetype().type_label())
+                {
+                    // This 'name' appears as a builtin, so we must insert our type first.
+                    Some(pos) => ordered.insert(pos, (name, builtin)),
+                    // Otherwise at the end.
+                    None => ordered.push((name, builtin)),
+                }
+            }
+        }
```
bendk, line 320: I think this could maybe fail if there were multiple levels of builtins, eg:

```
A { builtin: C},
B { }
C { builtin: B }
```

would get sorted to 

```
C { builtin: B }
A { builtin: C},
B { }
```

That seems pretty pathological for custom types and I think it's probably fine to ignore.  How about adding a note mentioning this?  That way we don't repurpose the algo for a situation where it would matter.

bendk (approved):
Looks good, love all the tests.  Just had one doc suggestion about the ordering algorithm.

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
mhammond: Yeah, thanks!

Issue #2016, opened 2024-03-07
Version our documentation


bendk:
[below is from @mhammond who took over the description to track the current status]

Now that #2077 is in-place I had a bit of a look at `mike` - it's conceptually a fairly simple tool - it manages a small amount of metadata about versions in a JSON file in the root of gh-pages, and it wraps the `mkDocs` build and publish flows to only updates a single directory of docs rather than the entire site. A subtle side-effect of this is that `mike` and the entire docs flow can *never* recreate the entire site, but instead needs to do updates of the existing gh-pages branch.

Another side-effect of this is that it can't really use the `gh-pages/deploy` action - we will need to deploy the old-school way executing `mike deploy` manually.

So my current plan is something like:
- [ ] Update our existing docs deploy flow to use `mkDocs deploy` in preparation for switching to mike.
- [ ] Update the mkDocs templates to include mike versioning (but this doesn't actually get rendered until versions are discovered at runtime via that json file in the root mentioned above)
- [ ] Manually edit the root of the docs site so it has the [same redirecting index.html as jj](https://github.com/martinvonz/jj/blob/gh-pages/index.html)
- [ ] In a throw-away branch, back-port #2077 and a couple of other tweaks to the 0.27 branch and generate docs for 0.27.
- [ ] Manually `mike deploy` 0.27 mkDocs generated docs into a directory names `0.27` with an alias of `latest` (an alias is just a mike-managed symlink)
- [ ] Write a little Python script to replace all existing `*.html` files on the site to be a sym-link to a document of the same name in `./current`

At this stage we should have a functioning site - albeit with only 1 version enabled. All previously existing .html pages should redirect to "current", which itself is currently symlink to 0.27

Then:

- [ ] Manually `mike deploy` the current `main` docs to a version called `unreleased`
- [ ] Update current CI actions to also `mike deploy` the current `main` docs to a version called `unreleased`
- [ ] Remove all old js, assets, etc.
- [ ] Document the new version release process to include the docs (part of which will be to use mike to point `current` to the new version)
- [ ] Work on automatically deploying new versions on new releases so that no extra manual steps are required when cutting new versions.

At this stage we should have a function site with 2 versions available to select from, `0.27` and `unreleased`, with `latest` being a symlink to `0.27`
After the next major release we will have 3 versions - `0.27`, `0.28` and `unreleased`, which `latest` being a symlink to 0.28.

Thoughts welcome!

mhammond:
[...]

PR #2077, opened 2024-04-22
Move doc generation to mkdocs.

mhammond:
The generated "content" (ie, .html layout, #refs etc) is the same as generated by mdbook - all old links should be unchanged.

The other theme etc generated stuff is obviously very different; mkdocs using the 'material' theme.

This is a step toward #2016, although no actual versioning happens here. Note also that I haven't yet tested the CI deploy code fully (although I have manually checked the commands listed should all work correctly in the docker image listed) - so there's a chance some followups will be necessary if I screwed this up.

Rendered version: https://mhammond.github.io/uniffi-rs


bendk (approved):
The docs look good to me, I'd say they're overall an improvement.  The one thing I like about the old docs is that the TOC is just a static list rather than dynamically expanded with the disclosure arrows.  Is it possible to do that with mdbook?

That's pretty nit-picky though.  If this makes versioning easier then I'm all for the change.

bendk:
I don't feel strongly at all about the styling.  If more people care, then maybe we could do majority vote.  If not, I think you should get the say since you're doing the work.

Sajjon:
Looks great! 

Is it possible to get more colorful code highlighting?

Screenshot 2024-04-23 at 20 15 10

Would be so much easier to read with more code highlight :)

mhammond:
> Is it possible to get more colorful code highlighting?

I made a small attempt here - see https://github.com/mozilla/uniffi-rs/blob/b0c963710cead64cddad96c71c96ce341ab9a82c/mkdocs.yml#L18-L21 - but I'm erring on the side of just landing this without it, which might help others help work out why it's not working :)

PR #2078, opened 2024-04-22
Small simplifications to the library mode implementation.

mhammond:
This moves all of the BindingsConfig mutations directly into `generate_external_bindings()`, the idea being that this makes all the operations easier to rationalize about and to make a future simplifications possible.

Also removes the crate-name from `Source` and exposes it as a function on ComponentInterface instead.

### DIFF
--- a/uniffi_bindgen/src/library_mode.rs
+++ b/uniffi_bindgen/src/library_mode.rs
@@ -68,13 +68,27 @@ pub fn generate_external_bindings<T: BindingGenerator>(
     let cdylib_name = calc_cdylib_name(library_path);
     binding_generator.check_library_path(library_path, cdylib_name)?;
 
-    let mut sources = find_sources(
-        binding_generator,
-        &cargo_metadata,
-        library_path,
-        cdylib_name,
-        config_file_override,
-    )?;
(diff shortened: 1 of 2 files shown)


uniffi_bindgen/src/library_mode.rs
```diff
@@ -68,13 +68,27 @@ pub fn generate_external_bindings<T: BindingGenerator>(
...
+                config.update_from_cdylib_name(cdylib_name);
+            }
+            config.update_from_ci(&ci);
+            Ok(Source {
+                config,
+                ci,
+                package,
+            })
+        })
+        .collect::<Result<Vec<_>>>()?;
```
bendk, line 90: This step is also needed for non-external bindings, but I guess that's fine since `generate_bindings` just calls `generate_external_bindings` anyway.  What's stopping us from merging those 2 functions together?

bendk (approved):
Looks good to me.

uniffi_bindgen/src/library_mode.rs
mhammond: Nothing really - this is just a small step to make future steps easier to rationalize about.

PR #2083, opened 2024-04-27
Factor out FFI converter operations

bendk:
Put these all in one module in preparation of updating how this code works. I'm hoping to rework custom type handling and also drop the `UniFfiTag` param. By factoring these into their own module and away from the specific usages, future changes like that should be easier to understand.

I think there are no functional changes here. In a couple places I replaced `FfiConverterArc<Self>` with `SomeFfiTrait<Arc<Self>>`, but those should be equivalent.


mhammond (approved):
makes sense to me 👍

Issue #2079, opened 2024-04-22
Error type named `Error` emits invalid Swift code


arg0d:
This might be a duplicate issue, but I didn't manage to find another issue like this.

For generated Swift code, `Error` is ambiguous with "system" `Error` type.

Version: v0.27.1 (I'm guessing older versions too

`Error` is a very natural type name when starting a new library, IMHO the generators should be fixed to support `Error` as error type name. We've already had 2 different teams working on separate projects step onto this same issue while onboarding their projects to uniffi :sweat_smile: 

There is a very simillar issue in C# bindings generators, https://github.com/NordSecurity/uniffi-bindgen-cs/issues/76.

```diff
diff --git a/fixtures/coverall/src/lib.rs b/fixtures/coverall/src/lib.rs
index f22ab5bc4..9d557ac72 100644
--- a/fixtures/coverall/src/lib.rs
+++ b/fixtures/coverall/src/lib.rs
@@ -23,6 +23,12 @@ pub enum CoverallError {
     TooManyHoles,
 }
 
+#[derive(Debug, thiserror::Error, uniffi::Error)]
+pub enum Error {
+    #[error("out of storage")]
+    DiskProblem {},
+}
+
 #[derive(Debug, thiserror::Error)]
 pub enum CoverallFlatError {
     #[error("Too many variants: {num}")]
```

```
/mounted_workdir/target/tmp/uniffi-fixture-coverall-af68e6307d32e36f/coverall.swift:2778:1: error: inheritance from non-protocol type 'Error'
extension ComplexError: Error { }
^
/mounted_workdir/target/tmp/uniffi-fixture-coverall-af68e6307d32e36f/coverall.swift:2844:1: error: inheritance from non-protocol type 'Error'
extension ComplexMacroError: Error { }
^
/mounted_workdir/target/tmp/uniffi-fixture-coverall-af68e6307d32e36f/coverall.swift:2892:1: error: inheritance from non-protocol type 'Error'
extension CoverallError: Error { }
^
/mounted_workdir/target/tmp/uniffi-fixture-coverall-af68e6307d32e36f/coverall.swift:2940:1: error: inheritance from non-protocol type 'Error'
extension CoverallFlatError: Error { }
^
/mounted_workdir/target/tmp/uniffi-fixture-coverall-af68e6307d32e36f/coverall.swift:2988:1: error: inheritance from non-protocol type 'Error'
extension CoverallFlatMacroError: Error { }
^
[...]

PR #2084, opened 2024-04-28
Swift: Enable error types to be named `Error`

Sajjon:
Solves https://github.com/mozilla/uniffi-rs/issues/2079

This PR makes it possible to name a type `Error` at let it be exported into Swift. It is is a quite simple PR which adds `Swift.` disambiguation.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
@@ -53,18 +53,19 @@ fileprivate extension RustCallStatus {
 }
 
 private func rustCall<T>(_ callback: (UnsafeMutablePointer<RustCallStatus>) -> T) throws -> T {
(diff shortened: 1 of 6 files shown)


uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
```diff
@@ -53,18 +53,19 @@ fileprivate extension RustCallStatus {
 }
 
 private func rustCall<T>(_ callback: (UnsafeMutablePointer<RustCallStatus>) -> T) throws -> T {
-    try makeRustCall(callback, errorHandler: nil)
+    let neverThrow: ((RustBuffer) throws -> Never)? = nil
```
Sajjon, line 56: Must do this dance so that Swift compiler knows error type, just passing a nil literal does not work.

bendk (approved):
Looks great, thanks for the contribution.

Issue #2000, opened 2024-02-28
Unable to conditionally enable uniffi::constructor macro


stan-stately:
I have a simple class that im trying to expose with proc macros like below:
```
#[derive(uniffi::Object)]
pub struct Client {
    api_endpoint: String,
}

#[uniffi::export]
impl Client {
    #[uniffi::constructor]
    pub fn new(api_endpoint: String) -> Self {
        Self { api_endpoint }
    }
}
```

my problem is that i only want uniffi scaffolding if my target is compiling for mobile - not wasm. In the past I have used conditional compilation for this with a lot of success. But when i try the following:
```
#[cfg_attr(not(target_family = "wasm"), derive(uniffi::Object))]
pub struct Client {
    api_endpoint: String,
}

#[cfg_attr(not(target_family = "wasm"), uniffi::export)]
impl Client {
    #[cfg_attr(not(target_family = "wasm"), uniffi::constructor)]
    pub fn new(api_endpoint: String) -> Self {
        Self { api_endpoint }
    }
}
```

I get the error ` associated functions are not currently supported`:
```
error: associated functions are not currently supported
  --> <redacted>/lib.rs:32:9
   |
32 |     pub fn new(api_endpoint: String) -> Self {
```

For now I'm just putting a `#[cfg(not(target_family = "wasm"))]` on the whole block as a workaround, but i think this could be a bug in one of the macros

jplatte:
Yeah, it is infeasible for uniffi to include its own parser + interpreter for cfg expressions such that it could know whether a cfg_attr should apply or not. I've wanted to solve this with a `uniffi` fake-cfg, but I don't think that's implemented yet.

rafaelbeckel:
I came here to report this bug and I found this issue had already been opened.

The conditional works for `wasm_bindgen` macros but not for `uniffi`.

I expected this to work similarly in both APIs, even with cfg aliases:

```rust
#[cfg_attr(wasm, wasm_bindgen)]
#[cfg_attr(not(wasm), uniffi::export)]
impl SomeObject {
    #[cfg_attr(wasm, wasm_bindgen(constructor))] // works as expected, compiles
    #[cfg_attr(not(wasm), uniffi::constructor)] // this issue
    pub fn new() {
        ...
    }
}
```

The error string comes from `uniffi_macros/src/export/scaffolding.rs`, line 72.

rafaelbeckel:
A workaround is possible with the `cfg_if` crate:

```rust
[...]

PR #2086, opened 2024-05-01
WIP - Solves #2000 to conditionally enable uniffi::constructor macro 

rafaelbeckel:
The main discussion happened at https://github.com/mozilla/uniffi-rs/issues/2000

This doesn't break any current test case. WIP because it should include new test cases covering some of the examples mentioned in the original discussion, but I'm still figuring out where they should go. I'll do that later this week.

Reviews are welcome.


jplatte (commented):
This doesn't do what you were talking about in #2000. This would make `#[hello::world::uniffi::constructor]` work, but not `#[cfg_attr(something, uniffi::constructor)]`.

rafaelbeckel:
I see; thanks for pointing it out. I went at it with a wrong assumption. I'll get the tests working, they would have caught it earlier.

bendk:
Have you checked out my comment in the issue?  What do you think about the suggested approach?  It's probably a bit more involved, but I can try to refactor some things so that it's not too bad.

rafaelbeckel:
Closing in favor of https://github.com/mozilla/uniffi-rs/pull/2113

Issue #1896, opened 2023-12-12
Create single module for shared functionality


bendk:
We should look into creating a single module on the foreign side that holds all shared functionality: RustBuffer, Future handling code, etc.  There would be a couple advantages to this:

- Smaller bindings
- Simplifies compatibility issues with external types.  With our current system, we create a separate `RustBuffer` class for each module, which makes using an external type weird since it each module uses a different `RustBuffer`.

alexkirsz:
We would love to have that as well!

In our setup, instead of running uniffi in a single step at the very end of the compilation process, we run it directly on the intermediate rlibs of each uniffi library. The current issues we've seen with this approach are:

1. uniffi will always re-generate bindings for all upstream uniffi libraries at every step; and
2. each module has its own RustBuffer/Primitive/Future handling code.

For 1., we only extract the generated module of the current uniffi library and ignore the rest.

For 2., it's unclear to us at this point if these duplicate implementations increase the final binary size, or if the foreign language's compiler and/or linker allow for deduplication. We will eventually need to investigate this, as we've found that splitting a single uniffi library into two different libraries with the same surface API increased our binary size significantly.

PR #2087, opened 2024-05-01
New system for remote / custom / external types

bendk:
As discussed in #1865, this is how I think we should update our code for remote / custom / external types.

The doc changes are the first commit so that we can agree on how things should work from the user's perspective. 

Changes:

- Make remote types a first-class feature.
- Make UDL generate blanket ffi trait impls for all UniFfiTags. Remove the `use_udl_*` since we don't need them anymore.
- Add `use_remote_type!` to handle the one case where we do need to forward ffi traits impls to another crate's tag.
- Use a macro to define custom types.  This way we have more freedom to change how things are implemented behind the scenes.
- Update the UDL external syntax so that `[Extern]` can work with any type.

Benefits are:
 - UDL and proc-macros will be consistent in their handling of UniFfiTag.
 - First-class remote types would enable things like using anyhow::Error in interfaces.
 - The custom type macro is easier to use then the current code.  It's also easier for us to make changes behind the scenes.
 - External types get a little less hacky.
 - 3rd party crates can provide built-in UniFFI support and implement the ffi traits for all their consumers.


jplatte:
In my mind the biggest benefit is actually that using types from a crate with builtin UniFFI support doesn't requure importing the trait impls at all with this, regardless of UDL vs. proc-macros. Maybe add that to the list?

bendk:
> In my mind the biggest benefit is actually that using types from a crate with builtin UniFFI support doesn't require 'importing' the trait impls at all with this, regardless of UDL vs. proc-macros. Maybe add that to the list?

Added and I agree that it's a benefit in many cases, but I think crates should think carefully before they do that since they lock users in to their choices.  I think this is especially true if they use a custom type conversion.  For example, should UUIDs be a string or a struct of 4 U32s?  Should URLs be a string or an interface?

mhammond:
That description is just a huge list of positives with not many draw-backs, so what's not to love :) These types have all been one hack on another... 

I'm twice confused though :)
> doesn't require 'importing' the trait impls at all with this

That sounds great too, but I don't know what kind of imports you mean here - what will it allow users to avoid doing?

> Added and I agree that it's a benefit in many cases, but I think crates should think carefully before they do 

And I don't get what you mean by this - how/why does what a crate choose?

bendk:
> I'm twice confused though :)
> 
> > doesn't require 'importing' the trait impls at all with this
> 
> That sounds great too, but I don't know what kind of imports you mean here - what will it allow users to avoid doing?

If we were in a future where a 3rd-party crate defined FFI traits for their types, then those traits impls would apply to all downstream crates automatically.  For some of the other solutions described in #1865, the downstream crates would have call `uniffi::use_remove_type!` to get the impls.

> 
> > Added and I agree that it's a benefit in many cases, but I think crates should think carefully before they do
> 
> And I don't get what you mean by this - how/why does what a crate choose?

Great question, I think it's something we have to figure out as we go forward.  But to take a couple examples:

- Suppose fairybridge (a [UniFFI-enabled async HTTP bridge](https://github.com/mozilla/application-services/pull/6018)).  If that ended up being wildly successful and other crates wanted to use it as a standalone library, then fairybridge should define blanket FFI trait impls for its types (otherwise users would need to call `uniffi::use_remote_type!` for each type they wanted to use).
- I don't think it would make sense for `uuid` to do the same for the `Uuid` type, since different consumers might want to pass uuids across the FFI in different ways.  Maybe it would make sense behind a feature flag, but I don't think there should always be a blanket impl.

mhammond:
> * then fairybridge should define blanket FFI trait impls for its types

Thanks, but I'm still mildly confused - in the description, you mention "Make UDL generate blanket ffi trait impls for all UniFfiTags" and IIUC, that's what procmacros do today, right?

So I assumed from that description that we'd always and unconditionally generate blanket impls. It's still not clear to me from the above how crates would opt-out of this. proc-macro users can't opt out of this now, right?

> Add use_remote_type! to handle the one case where we do need to forward ffi traits impls to another crate's tag.

Maybe this is part of my confusion - what exactly is the one case?

bendk:
> > * then fairybridge should define blanket FFI trait impls for its types
> 
> Thanks, but I'm still mildly confused - in the description, you mention "Make UDL generate blanket ffi trait impls for all UniFfiTags" and IIUC, that's what procmacros do today, right?
> 
> So I assumed from that description that we'd always and unconditionally generate blanket impls. It's still not clear to me from the above how crates would opt-out of this. proc-macro users can't opt out of this now, right?

Yes, that's 100% correct.  The only way to "opt-out" is to not use `uniffi::derive` on your crates.

I'm just saying that if you're the author of a crate like `url` and your users are asking you to add UniFFI support, you should probably think twice before just wrapping everything with `uniffi::derive`, since that forces all downstream crates to use your impls.

> > Add use_remote_type! to handle the one case where we do need to forward ffi traits impls to another crate's tag.
> 
> Maybe this is part of my confusion - what exactly is the one case?

This is the current case with `url` .  If one crate implements the FFI traits for `Url`, either through `custom_type!` or `remote_type!`, then other crates will need to call `use_remote_type!`.

bendk:
Just pushed the implementation of this.  I split it up into different commits for readability, but I'll probably squash them together when we merge.

All tests are passing and I'm really liking the new syntax and the concepts behind it, I'd love to know what others think.  As a follow-up I'd like to to enable methods on remote interfaces, but even without them it allows for using `anyhow::Error` which is pretty cool.  If we merge this plus #2093 then we can use `anyhow::Error` directly as the error type, no wrapping by the user needed.

uniffi_meta/src/reader.rs
```diff
@@ -339,6 +341,7 @@ impl<'a> MetadataReader<'a> {
         Ok(ObjectMetadata {
             module_path: self.read_string()?,
             name: self.read_string()?,
+            remote: false, // only used when generating scaffolding from UDL
```
bendk, line 344: These comments are a bit sad.  At some point, I think we should separate the metadata that is used to generate the scaffolding from the metadata used to generate the bindings.

Sajjon:
@bendk I've been testing your source branch quite heavily and it works really well, bravo, well done! I'm using (at time of writing, the last commit - 4b65c3796bc3743404e1ee2a7b4ab51e24589aa6 of your source branch)

**I perfectly understand this is in flux** :) so I'm happy and impressed that it works so well given that it has not been merged yet! 

I found a bug in Kotlin which you might be interested in. It is in my small "Lemon Merringue Pie" repo, and this PR (source branch) contains the bug - bug if that use case is intended to be supported, but I don't see why not:
https://github.com/Sajjon/lemon-merringue-pie/pull/2

(Swift works like a charm though)

Sajjon:
We probably want my PR into Bens fork: https://github.com/bendk/uniffi-rs/pull/3 - without it it is not possible to have a single error type defined in crate `X` but used in crate `Y` (and `Y2`, `Y3`, ...) - which I think is a quite natural setup.

Let me know if I should create a standalone PR into this target repo instead, "standalone" as in "probably can be based on main branch".

bendk:
> We probably want my PR into Bens fork: [bendk#3](https://github.com/bendk/uniffi-rs/pull/3) - without it it is not possible to have a single error type defined in crate `X` but used in crate `Y` (and `Y2`, `Y3`, ...) - which I think is a quite natural setup.
> 
> Let me know if I should create a standalone PR into this target repo instead, "standalone" as in "probably can be based on main branch".

Thanks for that one, just merged it in.

Sajjon:
(Using Bens fork, commit: https://github.com/bendk/uniffi-rs/commit/f4fae828ac496e412fce70ab2f02f09ef19932c4 )

Context: [Asked for help on Matrix](https://matrix.to/#/!tsNDDGHvNGVNmvOOzi:mozilla.org/$MK9c468Qi04Yj3p-YFASfY-byaJMl9XpsXkC7BMah5M?via=mozilla.org&via=matrix.org&via=element.io) about compilation errors in Swift bindings when multi-crates are used:
For Multi-crate Cargo Workspace setup : Rust crates that are dependent on each other each with their own UniFFI setup (`uniffi.toml` files and call to `uniffi::include_scaffolding` / `::setup_scaffolding) today (as this PR stands) will create dependencies between the generated (by bindgen) Swift files too.

Given `CrateX` being dependent on UniFFI code in `CrateY`, this will generate broken Swift code:
`CrateX.swift` will contain swift code calling `lift` / `lower` etc o types in `CrateY.swift` - but those functions are **private**, so this does not compile. I'm unsure about status of Kotlin (and other language bindings), I have not progress that far yet.... 

I think I have an OK idea on how to solve this - maybe there is some detail I miss which makes my naive idea not break, but here goes nothing :
(this assumes on single call to generate (Swift) bindings - aka library mode)
1. All generated Converters will be placed in *separate files*, in lets call it `generated/` folder (as specified by `--out-dir`), and **only be generated once**, i.e. the contents of p`template/StringHelper.swift`](https://github.com/bendk/uniffi-rs/blob/custom-type-rework/uniffi_bindgen/src/bindings/swift/templates/StringHelper.swift) will be put `generated/` folder
2. The actual type - `fileprivate struct FfiConverterString: FfiConverter` - gets a changed visibility to `internal struct` instead of `fileprivate`. 
3. All protocols, types, functions and methods in [`RustBufferTemplate`](https://github.com/bendk/uniffi-rs/blob/custom-type-rework/uniffi_bindgen/src/bindings/swift/templates/RustBufferTemplate.swift) (and similar tempates) get changed visibility to `internal` too
4. Bindgen code for a single crate only contain the author written UniFFI code in one file per crate, places in `generated/` "next to" the UniFFI generated above mentioned converters.

The result is many - smaller - Swift files with the converters, all with internal visibility so that these Swift files can see each all symbols **within this Swift target**, and `CrateX.swift` and `CrateY.swift` will also see each others `lift` and `lower` methods.

I think that will work. And I think most UniFFI consumers build scripts will "just work" (at least if they used `cp` / `mv` with `*.swift` ) or easily fixed. 

I started doing this myself but I will take me days, but probably done by @bendk in a couple of hours tops!

WDYT? I think this will make this PR work with Multi-crate solution!

[Here is commit showcasing the broken bindgen in my small Workspace demo project LemonMerringuePie](https://github.com/Sajjon/lemon-merringue-pie/commit/a633a371c0d44d4a35cad6d0b0f49d0ff7190e49) 

Screenshot of problem:
Screenshot 2024-05-24 at 08 51 16

(Using Bens fork, commit: https://github.com/bendk/uniffi-rs/commit/f4fae828ac496e412fce70ab2f02f09ef19932c4 )

bendk:
I made #2150 to try to land the custom type changes before the remote type changes.  The remote type changes are blocked by #2130.  I think we're pretty close to consensus there, but whatever we decide will require some code changes afterwards.

Sorry for all the noise with the force-pushes -- I messed up by git branch names.

CyonAlexRDX:
@bendk @mhammond did you see my comment https://github.com/mozilla/uniffi-rs/pull/2087#issuecomment-2128711332 above? any plans on fixing support for workspace setup with UniFFI? I think my plan above is pretty straightforward to implement, and should fix the issue no?

bendk:
> @bendk @mhammond did you see my comment [#2087 (comment)](https://github.com/mozilla/uniffi-rs/pull/2087#issuecomment-2128711332) above? any plans on fixing support for workspace setup with UniFFI? I think my plan above is pretty straightforward to implement, and should fix the issue no?

I think it makes sense, although I haven't thought it all the way through.  It seems like it should be a separate issue though, I'd love to isolate the changes and discuss them with as many Swift developers as we can.  I also think some of those ideas could apply to other languages (see #1301, #1896).  Can you create an issue for it?

bendk:
Closing this in favor of #2303.

PR #2088, opened 2024-05-01
Add more functionality to get_custom_type_aliases sorting

giarc3:
I tried to consume the change from #2075 today and noticed that it didn't cover my use case, where a type alias is used as the key for a HashMap which also is an alias. I added a test showing this and made a change to the sorting algorithm to fix this. 

The current algorithm for ordering type aliases is: "if the typename I'm inserting into the sorted list _is the type_ for something already in the list, put this item in front of it".

My proposed new algorithm is: "if the typename I'm inserting into the sorted list is _included in the nested types_ for something already in the list, put this item in front of it".

### DIFF
--- a/fixtures/ext-types/custom-types/src/lib.rs
+++ b/fixtures/ext-types/custom-types/src/lib.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 // A trivial guid, declared as `[Custom]` in the UDL.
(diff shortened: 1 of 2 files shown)


mhammond (approved):
thanks - I feel like we are missing something simpler here that would also fix the problem in the comment too, but I still don't see it :)

PR #2090, opened 2024-05-02
Refactor the derive macro code

bendk:
This is prep work for implementing the system for remote / custom / external types as described in #2087.  The generated code should stay exactly the same, this just refactors how it's generated.

Consolidated the `derive_*_for_udl` macros into a single `udl_derive` macro and added the `DeriveOptions` struct.  I think that we can update UDL-generation to create blanket FFI trait impls by just tweaking which flags we set.  We can also re-use the same pattern to implement remote types (which will have `generate_metadata: true, local_tag: true`).

Changed the specific derive code to use `DeriveOptions` plus an item struct.  For example  RecordItem, which represents the parsed DeriveInput for the record. This will make future changes easier since we only need to modify one of these structs, not update all the function signatures.

Changed the UDL derives to parse attributes exactly like regular derives.  For example, we now wrap enums with `#[uniffi(flat_error)]` rather than calling `derive_error_for_udl!(flat_error)`.  This is prep work for remote types -- in that case I think we should tell the user to define the item exactly like it was a regular derive and not special case the attributes.

Removed the `#[uniffi(non_exhaustive)]` attribute and parse the real `#[non_exhaustive]` attribute instead.  This is easy to do now with the new system and also will be good for remote types.

Made EnumItem also parse error attributes.  I think this is simpler than creating a separate EnumItem vs ErrorItem, at least for now.

### DIFF
--- a/fixtures/uitests/tests/ui/fieldless_errors_used_in_callbacks_cant_have_fields.stderr
+++ b/fixtures/uitests/tests/ui/fieldless_errors_used_in_callbacks_cant_have_fields.stderr
@@ -1,10 +1,7 @@
 error[E0533]: expected value, found struct variant `Self::DivisionByZero`
  --> $OUT_DIR[uniffi_uitests]/errors.uniffi.rs
   |
-  | / #[::uniffi::derive_error_for_udl(
-  | |     flat_error,
-  | |     with_try_read,
-  | | )]
-  | |__^ not a value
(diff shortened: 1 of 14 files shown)


jplatte (approved):
Only skimmed, but didn't notice anything wrong. The underlying idea seems very good!

mhammond (approved):
thanks!

PR #2093, opened 2024-05-03
Adding the `LowerError` trait

bendk:
This specifies how types are lowered when they are the `E` side of `Result<T, E>`.  The main reason for the trait is that it lets us specialize the handling for interfaces by auto-wrapping the value in an Arc.

This makes using interfaces as errors much more ergonomic.

Also, replace the `Error` bound on the `E` types with `Display + Debug`. I just found out that `anyhow::Error` doesn't actually implement `Error` for some obscure reason.

### DIFF
--- a/fixtures/error-types/src/lib.rs
+++ b/fixtures/error-types/src/lib.rs
@@ -27,13 +27,22 @@ impl From<anyhow::Error> for ErrorInterface {
     }
 }
 
-// must do explicit conversion...
+// Test an interface as the error type
 fn oops() -> Result<(), Arc<ErrorInterface>> {
+    // must do explicit conversion to convert anyhow::Error into ErrorInterface
     Err(Arc::new(
         anyhow::Error::msg("oops")
             .context("because uniffi told me so")
             .into(),
     ))
+}
+
+// Like `oops`, but let UniFFI handle wrapping the interface with an arc
+fn oops_nowrap() -> Result<(), ErrorInterface> {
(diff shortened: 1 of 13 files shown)


fixtures/error-types/src/lib.rs
```diff
@@ -28,12 +28,10 @@ impl From<anyhow::Error> for ErrorInterface {
 }
 
 // must do explicit conversion...
-fn oops() -> Result<(), Arc<ErrorInterface>> {
-    Err(Arc::new(
-        anyhow::Error::msg("oops")
-            .context("because uniffi told me so")
-            .into(),
-    ))
+fn oops() -> Result<(), ErrorInterface> {
```
mhammond, line 31: I assume the old code still works, so maybe keep a test with an explicit `Arc<Struct>`? There's also a trait example below that does keep `Arc<>`, but should it also test a non-arc version?

Maybe also update https://mozilla.github.io/uniffi-rs/udl/errors.html to remove the `Arc<>` from the example, and maybe a changelog entry?

mhammond (approved):

fixtures/error-types/src/lib.rs
bendk: > I assume the old code still works, so maybe keep a test with an explicit Arc<Struct>?

That makes sense, I'll update the test code.

> There's also a trait example below that does keep Arc<>, but should it also test a non-arc version?

This isn't currently possible because you need the Arc to handle the `dyn Trait` part.

PR #2094, opened 2024-05-05
Update the BindingGenerator trait and remove BindingsConfig.

mhammond:
This is a breaking change for BindingGenerators, and follows up on other breaking changes made for this release (#2078). Between them, it is intended to offer a better framework for binding generators for future versions and break unintentional coupling between uniffi_bindgen and the builtin bindings.

This patch updates the `BindingGenerator` trait to give binding generators more control over the binding generation process and to simplify the interactions between the generator and `uniffi_bindgen`.

The trait `BindingsConfig` has been removed and replaced with a new method on `BindingGenerator` which passes the generator the entire list of all `ComponentInterface` and `Config` objects to be used in the generation, which the generator can modify as necessary. The binding generator is also passed the entire list of items to generate rather than called once per item - this gives the generator more flexibility in how the items are generated.

A new `Component` struct has been introduced which holds all necessary information for a single crate/namespace, including the `ComponentInterface` and `Config`. These structs are passed to the `BindingGenerator`

Calling everyone with an interest in external binding generators - cc @heinrich5991 @SalvatoreT @arg0d @skeet70 @Lipt0nas. All feedback welcome, including where I went too far, or where I didn't go far enough - eg

* I was thinking a new `GenerationSettings` struct which held things like `out_dir`, `try_format_code`, `cdylib` etc to make the function argument lists smaller and to make future additions less of a breaking change (eg, a new option would not change signatures, but instead just add a new item to a struct). etc.
* I'm still not that happy with how each config is created, but don't have any ideas how to fix that. Trying to remove `Config` as a type param and use `toml::Value` directly doesn't seem particularly viable. I'd welcome ideas here.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/mod.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/mod.rs
@@ -2,10 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use crate::{BindingGenerator, ComponentInterface};
+use crate::{BindingGenerator, Component, GenerationSettings};
 use anyhow::Result;
 use camino::{Utf8Path, Utf8PathBuf};
 use fs_err as fs;
+use std::collections::HashMap;
 use std::process::Command;
 
 mod gen_kotlin;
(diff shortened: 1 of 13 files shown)


heinrich5991:
I don't maintain external binding generators.

arg0d (approved):
Looks good, this is only a "mechanical" breaking change, shouldn't cause much difficulties migrating to this new vesrion.

But I wonder what would be some real use cases for accessing other component interfaces when generating code. I remember I wanted to access other component interfaces for some purpose when working with C# generator, but in the end it wasn't required. The only use case I can think of right now, is the ability to index external types from another component interface. For that purpose, it would be more useful to have `components: HashMap` instead of `components: Vec`. But like I said, this is not based on any present requirements.

mhammond:
> But I wonder what would be some real use cases for accessing other component interfaces when generating code. 

Yeah, for now it's all going to be about external types - kotlin uses that to work out what "package" name to use to refer to external types in other crates.

uniffi_bindgen/src/lib.rs
```diff
@@ -110,53 +110,49 @@ pub mod scaffolding;
...
     /// - `out_dir`: The path to where the binding generator should write the output bindings
     fn write_bindings(
         &self,
-        ci: &ComponentInterface,
-        config: &Self::Config,
+        components: &[Component<Self::Config>],
         out_dir: &Utf8Path,
         try_format_code: bool,
     ) -> Result<()>;
+}
```
bendk, line 148: Would it be simpler to just have a single function that does the work of all of the `BindingGenerator` methods?  This means we don't need the `Config` type anymore, that data would just be a local variables in that method.  I think this would actually simplify the generator code for the builtin bindings.

bendk (commented):
This looks good to me, it's certainly an improvement.  I think it maybe could be pushed further though.  I love the idea of a `GeneratorSetttings` struct, I also posted some thoughts on simplifying the trait methods.

uniffi_bindgen/src/lib.rs
mhammond: So a single function that takes mutable references, then does both the updating of the configs *and* writing the bindings? I did think about that and left it as it was because I kinda thought the separation of mutability and the generation seemed nice - it's not that I thought it was a bad idea, I just left it alone because I could see a reason to split them. But yeah, I am inclined to agree that maybe that's even better.

So to be clear, you'd suggest having that function take a vec of roughly (`CI`, `toml::Value`), then convert to Config objects inside that one method? I think this function would need to return the `Config` objects though - eg, [this swift code](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/bindings/swift/test.rs#L140) need to know what `Config` is.

(Maybe I'll just take smaller steps - stick with `Config` for now then see what people think once the "new world" is a little clearer)

skeet70 (approved):
This looks fine to me, but even though I'm close to having Java binding generation working, I don't think I understand what's going on well enough to know if this is an improvement or not. For what I'm doing this mostly seems different, not necessarily good or bad.

Lipt0nas:
Looks good, accessing other component interfaces should be useful for languages where import order matters

I am very interested in the `GenerationSettings` idea, would reduce a lot of friction when adding additional configuration options, and would work especially well for opt-in features like formatting

uniffi_bindgen/src/bindings/kotlin/mod.rs
```diff
@@ -26,33 +27,66 @@ impl BindingGenerator for KotlinBindingGenerator {
...
-        fs::create_dir_all(&kt_file)?;
-        kt_file.push(format!("{}.kt", ci.namespace()));
-        fs::write(&kt_file, generate_bindings(config, ci)?)?;
-        if try_format_code {
-            if let Err(e) = Command::new("ktlint").arg("-F").arg(&kt_file).output() {
-                println!(
-                    "Warning: Unable to auto-format {} using ktlint: {e:?}",
-                    kt_file.file_name().unwrap(),
-                );
+        // I think this is just trying to insist on a cdylib being able to be obtained from the lib_path?
```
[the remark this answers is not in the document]
mhammond: Thanks for that - I think even better is to just have the CLI enforce that! My new version will do that.

uniffi_bindgen/src/lib.rs
mhammond: Assuming you meant what I thought you meant, another slight problem here is that we filter the CIs - ie, the list of CIs we pass to `update_component_configs` might have more items then what we pass to `write_bindings`.

mhammond:
Thanks for all the feedback - I incorporated much of it in the most recent version, including the new `GenerationSettings`.

I'm fairly happy with this so think it's ready for review, but still obviously open to all feedback.

bendk (approved):
I like this a lot.  `GenerationSettings` seems like a nice step forward and in general this feels simpler.

mhammond:
I meant to reply to:

> The only use case I can think of right now, is the ability to index external types from another component interface. For that purpose, it would be more useful to have `components: HashMap` instead of `components: Vec`.

All the bindings do iterate over it - only kotlin builds its own map. Some bindings might need a map keyed by something over the crate name. And this is more convenient for the bindgen :) So I left the arrays. Welcome suggestions for further feedback in followups though.

Thanks everyone.

PR #2098, opened 2024-05-08
Perform Swift checksum only once

matiasbzurovski:
## Background
Every time a Rust method is called from Swift via UniFFI, the library internally checks that it has been initialized and that the checksum for each exported function is the expected one. To do so, it uses the following computed property:
```Swift
// Use a global variables to perform the versioning checks. Swift ensures that
// the code inside is only computed once.
private var initializationResult: InitializationResult {
   // ..
}
```

## Problem
The comment on the top of the property isn't correct, since Swift will actually compute this property **every time** it is accessed (which is every time a method exported by UniFFI is used). Given a checksum is performed for each of the exported methods, this could add some significant delay on medium sized libraries that perform multiple sequent Rust calls.

## Solution
Initialization should only be performed once, while the checksum for the functions will always have the same result during an app session. Therefore, it makes sense to respect its comment and compute the `InitializationResult` only once.

When testing with the added fixture, the following results where found when simulating 1k/10k method calls:

Screenshot 2024-05-08 at 19 23 12

## Credits
Big thanks to @Sajjon for his help 🙂


fixtures/swift-perform-checksum-once/tests/bindings/test_perform_checksum_once.swift
```diff
@@ -0,0 +1,16 @@
...
+_ = fn0(arg: entity)
+
+// So we measure time between second and 10,000th calls
+_ = fn1(arg: entity)
+let startTime = CFAbsoluteTimeGetCurrent()
+for _ in 0..<10000 {
+  _ = fn2(arg: entity)
+}
+
+let elapsedTime = CFAbsoluteTimeGetCurrent() - startTime
```
mhammond, line 15: The PR seems fine but I'm skeptical of this test - every single time a test tries to assert something about wall time it ends up failing in some overloaded CI environment. TBH it's not clear a fixture is really needed here, although maybe having it under "benchmarks" and arranging for it not to be run by CI would make sense - WDYT?

CyonAlexRDX: It was me suggesting a timed test -- but I had the same concerns as you @mhammond the moment I reviewed this PR by my dear colleague Mathias, it was a bad suggestion by me! But still it was good to measure, we got the results, a 220 x speedup :D 

Converting the fixture to a benchmark makes sense.

CyonAlexRDX: @mhammond we would love to get this PR merged ASAP - and @matiasbzurovski is just now learning Rust, so would take some more time for him to convert the fixture to a benchmark. We might be able to do it later though.

So I suggest we update the PR with just the actual *fix*, and a benchmark can be added later.

WDYT?

mhammond: SGTM

matiasbzurovski: Removed the fixture and left the fix only.

mhammond (approved):

Issue #2089, opened 2024-05-01
Kotlin bug: enum with associated value being `Arc<Object>` generates non compiling Kotlin code (UniFFI 0.27.1)


Sajjon:
Using `0.27.1` (tried `0.27.0` first which failed then found https://github.com/mozilla/uniffi-rs/commit/870362669c4f1ab398e892f0fb17eb5c5e6216da#diff-3e8bc41a800aa4cb18ce6afac785374f59c534c8f918638582d9d930f14ce994 and thought 0.27.1 would help... it did not.)

This non compiling Kotlin code:

```kotlin
    @Suppress("UNNECESSARY_SAFE_CALL") // codegen is much simpler if we unconditionally emit safe calls here
    override fun destroy() {
        when(this) {
            is ProfileFileContents.PlaintextProfile -> {
                
    Disposable.destroy(
        this.``)
                
            }
            is ProfileFileContents.EncryptedProfile -> {// Nothing to destroy
            }
            is ProfileFileContents.NotProfile -> {// Nothing to destroy
            }
        }.let { /* this makes the `when` an expression, which ensures it is exhaustive */ }
    }
```

by this Rust code:

```rust
/// Result of analyzing a file (bytes), containing either a Profile
/// which we were able to successfully JSON deserialize from the bytes,
/// or EncryptedProfile for which wallets will continue prompting the
/// user for an encryption password and then call JSON deserialize
/// of `EncryptedProfileSnapshot` using [`Profile::new_from_encryption_bytes`](Profile::new_from_encryption_bytes)
/// or if we failed to parse as Profile and `EncryptedProfileSnapshot`
/// then `NotProfile` is used, indicating that the bytes is not at all
/// a Profile.
#[derive(Debug, PartialEq, Eq, uniffi::Enum)]
#[allow(clippy::enum_variant_names)]
pub(crate) enum ProfileFileContents {
    /// The JSON deserialized Profile from some bytes.
    PlaintextProfile(Arc<RefProfile>),

    /// We successfully JSON deserialized the bytes into
    /// `EncryptedProfileSnapshot`, the wallets should proceed
    /// with asking the user for the decryption password.
    EncryptedProfile,

    /// The bytes is neither a valid `Profile` nor `EncryptedProfile`,
    /// it is either a corrupt file or it is not at all a Profile file,
    /// contrary to the users beliefs (or the user accidentally selected
    /// a random file...)
    NotProfile,
}
```
[...]

PR #2100, opened 2024-05-09
Bugfix: Enum with associated value generates non compiling Kotlin code

Vinnstah:
As reported in issue #2089 there is a bug generating non-compiling Kotlin code. This happens when exporting a Enum with an associated value that is a `Arc<Object>`. This led to non-compiling Kotlin code:

```
    Disposable.destroy(
        this.``)
```

I recreated this by adding a Enum in fixtures and got the same compile-time error:

```
#[derive(uniffi::Enum)]
pub(crate) enum AnimalAssociatedType {
    Dog(Arc<AnimalObject>),
    Cat,
}
```

This issue was caused by the `destroy_fields` macro. Previously there was no checks if the `field.name()` was empty leading to the macro creating empty methods in cases where some fields had a name and some didn't.

I added a check for `field.name().is_empty()` and in such cases no `destroy` method will be generated by the macro.

No additional tests were added since this was a compile time error that was mitigated by these changes.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/macros.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/macros.kt
@@ -134,10 +134,15 @@ v{{- field_num -}}
 
 // Macro for destroying fields
 {%- macro destroy_fields(member) %}
-    Disposable.destroy(
     {%- for field in member.fields() %}
-        this.{{ field.name()|var_name }}{%- if !loop.last %}, {% endif -%}
-    {% endfor -%})
+        {%- if field.name().is_empty() %} 
+            // No Label
+        {% else %}
+            Disposable.destroy(
+            this.{{ field.name()|var_name }}{%- if !loop.last %}, {% endif -%}
+            )
+        {% endif %}
+    {% endfor -%}
 {%- endmacro -%}
 
 {%- macro docstring_value(maybe_docstring, indent_spaces) %}
--- a/uniffi_bindgen/src/bindings/kotlin/templates/macros.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/macros.kt
@@ -132,16 +132,10 @@ v{{- field_num -}}
 {%- endif -%}
 {%- endmacro %}
 
-// Macro for destroying fields
+ // Macro for destroying fields
 {%- macro destroy_fields(member) %}
     {%- for field in member.fields() %}
-            Disposable.destroy(
-        {%- if field.name().is_empty() %} 
-            this.{%- call field_name(field, loop.index) -%}
-        {% else %}
-            this.{{ field.name()|var_name }}{%- if !loop.last %}, {% endif -%}
-        {% endif %}
-            )
+        Disposable.destroy(this.{%- call field_name(field, loop.index) -%})
     {% endfor -%}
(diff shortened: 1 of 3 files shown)


uniffi_bindgen/src/bindings/kotlin/templates/macros.kt
```diff
@@ -134,10 +134,15 @@ v{{- field_num -}}
...
 {%- macro destroy_fields(member) %}
-    Disposable.destroy(
     {%- for field in member.fields() %}
-        this.{{ field.name()|var_name }}{%- if !loop.last %}, {% endif -%}
-    {% endfor -%})
+        {%- if field.name().is_empty() %} 
+            // No Label
+        {% else %}
+            Disposable.destroy(
+            this.{{ field.name()|var_name }}{%- if !loop.last %}, {% endif -%}
```
bendk, line 142: I don't think you can skip destroying the field in the above case.  Instead, you can use the `field_name` macro which works in both cases.  See [the enum template](https://github.com/mhammond/uniffi-rs/blob/a98523b664064ca80bb1a53a8425a191b01ed9ad/uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt#L117) for an example.  Since you're already in the macro template, you don't need the `kt::` prefix.

bendk (changes requested):
Thanks for posting this and especially for updating the tests for it.

I think there's an issue with the implementation though.  `destroy` needs to be called to free the pointer reference owned by the Kotlin side of the FFI.

uniffi_bindgen/src/bindings/kotlin/templates/macros.kt
Vinnstah: Thank you for your input @bendk. Since I haven't used Kotlin myself I wasn't fully aware and made a incorrect assumption.

Would it be correct to assume that if a Enum `Sealed Class` contain a variant with a field, all variants for that specific Enum (SC) should generate a `destroy` function? For simple Enums (without associated data) no `destroy` is needed to clear up pointer references because they're generated as `enum class` in Kotlin?

I updated my implementation but just want to understand if I'm following the correct assumptions:
```
// Macro for destroying fields
{%- macro destroy_fields(member) %}
    {%- for field in member.fields() %}
            Disposable.destroy(
        {%- if field.name().is_empty() %} 
            this.{%- call field_name(field, loop.index) -%}
        {% else %}
            this.{{ field.name()|var_name }}{%- if !loop.last %}, {% endif -%}
        {% endif %}
            )
    {% endfor -%}
{%- endmacro -%}
```

bendk: That's correct.  I think you can make it even simpler though, since `field_name` works for both cases.:

```
 // Macro for destroying fields
{%- macro destroy_fields(member) %}
    {%- for field in member.fields() %}
            Disposable.destroy(this.{%- call field_name(field, loop.index) -%})
    {% endfor -%}
{%- endmacro -%}
```

Vinnstah: That worked like a charm, updated it in my latest push.

bendk (approved):
Looks great, let's merge this.  Thanks for the contribution.

PR #2102, opened 2024-05-10
Changes for the `0.27.2`

bendk:
I'm hoping to get these changes released so that Firefox can consume them.  Does this make sense for a 0.27.2 release?  Are there any other changes that should go in there?


mhammond (approved):
huh, thanks for the effort of backporting these changes.

bendk:
That fix is going to be hard to get in to the 0.27.x branch.  It's based on top of #2075, which itself is based on top of #2074 and #2072.  I think that one should wait until the 0.28.0 release, which we could also do fairly soon.

PR #2109, opened 2024-05-14
Prefix newly added globals in Python with underscores

heinrich5991:
Followup of #1599.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/templates/wrapper.py
+++ b/uniffi_bindgen/src/bindings/python/templates/wrapper.py
@@ -75,9 +75,6 @@ __all__ = [
     {%- for c in ci.callback_interface_definitions() %}
     "{{ c.name()|class_name }}",
     {%- endfor %}
-    {%- if ci.has_async_fns() %}
-    "uniffi_set_event_loop",
-    {%- endif %}
 ]
 
 {% import "macros.py" as py %}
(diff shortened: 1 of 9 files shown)


uniffi_bindgen/src/bindings/python/templates/wrapper.py
```diff
@@ -75,9 +75,6 @@
     {%- for c in ci.callback_interface_definitions() %}
     "{{ c.name()|class_name }}",
     {%- endfor %}
-    {%- if ci.has_async_fns() %}
```
mhammond, line 78: I don't think we want this change - this function is documented as being exported by the Python bindings.

mhammond (changes requested):
Looks fine except for ` uniffi_set_event_loop` (and maybe I just misunderstood!)

uniffi_bindgen/src/bindings/python/templates/wrapper.py
heinrich5991: It's still exported. `__all__` just determines what you get when you do `from library import *`. You can still get it using `import library; library.uniffi_set_event_loop` or `from library import uniffi_set_event_loop`.

mhammond: I understand that - but I'm not sure this change will meet the expectations of people reading our docs.

mhammond: (to be clear, I'm not saying this change is wrong, I'm just not sure it's correct. @badboy @bendk do either of you have thoughts)

heinrich5991: Added such a note.

mhammond (approved):
Thanks!

PR #2113, opened 2024-05-16
add support for cfg_attr-like attributes

saks:


uniffi_macros/src/export/attributes.rs
```diff
@@ -249,57 +250,85 @@ impl ExportedImplFnAttributes {
...
-                    if this.constructor {
-                        return Err(syn::Error::new_spanned(
-                            attr,
-                            "duplicate constructor attribute",
-                        ));
+            let path = attr.path();
+
+            if is_uniffi_path(path) {
+                this.process_path(path, attr, &attr.meta)?;
+            } else if matches!(attr.meta, Meta::List(_)) {
```
bendk, line 257: Maybe this should also check that `attr.meta.path` is `cfg_attr`?  It seems safer to me to specifically target that attribute rather than all list-like attributes.

rafaelbeckel: I think because `uniffi::constructor` and `uniffi::method` are just markers, it's safe to ignore external APIs and assume that the user's intention, in the context of **uniffi** itself, is to mark the current method as one of the two variants, regardless of external conditions.

We don't need to match special cases like `cfg_attr`. It doesn't really matter if we match `uniffi::constructor`, `cfg_attr(whatever, uniffi::constructor)` or `foo(bar(baz(uniffi::constructor)))` from some `foobar` crate because the external wrappers are not part of our API, and users are free to mess with them as they need.

This general-purpose solution keeps our current API intact, so there's no need to document special cases or create new syntax for conditional parsing.

saks: I tend to agree with @rafaelbeckel, but if anybody feels really strong about limiting the implementation to only work with `cfg_attr`, I'm happy to make a change. After all, my main goal is to get it working at least for `cfg_attr`.

bendk: I don't feel strongly about this one.  It feels safer to me, but I'm struggling to come up with an example where it would matter one way or another.

bendk (approved):
The code looks correct to me, @rafaelbeckel likes it.  Can we merge this one?

jplatte:
I'd still much prefer the main attribute be constrained to be exactly `cfg_attr`, and the first argument be constrained to exactly `uniffi`.

IMHO it is more important that accidentally doing the wrong thing is prevented (i.e. a user writes `#[cfg_attr(actually_meaningful_cfg_that_only_applies_sometimes, uniffi::method(name = "conditionally_different_name"))]`, it appears to work, _later_ they discover that the rename is actually always applied), than supporting the obvious thing (`#[cfg_attr(feature = "uniffi", uniffi::attribute(args))]`). And even if the obvious thing is changed to work later using an implementation approach like what @bendk suggested in the issue where cfg's are actually properly evaluated, there could still be value in the `uniffi` pseudo-cfg being supported (since it is shorter to write / easier to read).

rafaelbeckel:
I have no strong opinion about this, and restricting it to `cfg_attr` does the job for me just fine.

rafaelbeckel (approved):
Looks good to me. Limiting it to `cfg_attr` solves the original issue just fine. I can't think of other use cases. If there is one, someone else will open another issue in the future.

mhammond:
Can we please add a section at the end of [here (I guess?)](https://github.com/mozilla/uniffi-rs/blob/main/docs/manual/src/proc_macro/index.md) and maybe even a changelog? :)

saks:
thank you @mhammond, should have done it earlier. Please see if update to docs is making sense. Copywriting is not necessarily my strength :)

rafaelbeckel (approved):

mhammond:
Thanks!

PR #2116, opened 2024-05-18
Add localizedDescription support to swift errors.

### LINKED ISSUE #2110: Swift errors don't provide `localizedDescription`
With the patch below, coveralls fails because `e.localizedDescription` is `The operation couldn’t be completed. (coverall.ComplexError error 0.)`. We are hitting this in our production code because our swift app uses `localizedDescription` when reporting errors to our external error reporting frameworks etc.

I'm not sure what the right thing to do here is - a naive solution will not, by any reasonable definition, be "localized". However, it's fairly clear to me that the default message is quite useless - particularly when the error is an interface like the other part of the patch - you can't really obtain *any* details about the error (whereas in the enum case, the "error number" shown is apparently the zero-based index of the enum variant)

```
diff --git a/fixtures/coverall/tests/bindings/test_coverall.swift b/fixtures/coverall/tests/bindings/test_coverall.swift
index 91e95342cd...419b45851e 100644
--- a/fixtures/coverall/tests/bindings/test_coverall.swift
+++ b/fixtures/coverall/tests/bindings/test_coverall.swift
@@ -144,7 +144,10 @@
         } else {
             fatalError("wrong error variant: \(e)")
         }
-        assert(String(describing: e) == "OsError(code: 10, extendedCode: 20)", "Unexpected ComplexError.OsError description: \(e)")
+        let expectedString = "OsError(code: 10, extendedCode: 20)";
+        assert(String(describing: e) == expectedString)
+        print("Actual localized", e.localizedDescription)
+        assert(e.localizedDescription == expectedString)
     }
 
     do {
diff --git a/fixtures/error-types/tests/bindings/test.swift b/fixtures/error-types/tests/bindings/test.swift
index cc409c4ad0...1e13b2b021 100644
--- a/fixtures/error-types/tests/bindings/test.swift
+++ b/fixtures/error-types/tests/bindings/test.swift
@@ -20,7 +20,9 @@
 }
 
 let e = getError(message: "the error")
+print("LD:", e.localizedDescription)
+assert(e.localizedDescription == "??")
 assert(String(describing: e) == "the error")
 assert(String(reflecting: e) == "ErrorInterface { e: the error }")
 assert(Error.self is Swift.Error.Type)
 assert(Error.self != Swift.Error.self)

```

mhammond:
Also some extra error string semantics tests for kotlin.

Fixes #2110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 - Objects error types can now be as `Result<>` error type without wrapping them in `Arc<>`.
 
+- Swift errors now provide `localizedDescription` ([#2116](https://github.com/mozilla/uniffi-rs/pull/2116))
+
 ### What's fixed?
 - Custom Type names are now treated as type names by all bindings. This means if they will work if they happen to be
   keywords in the language. There's a very small risk of this being a breaking change if you used a type name which
--- a/fixtures/error-types/tests/bindings/test.kts
+++ b/fixtures/error-types/tests/bindings/test.kts
@@ -13,6 +13,12 @@ try {
     assert(e.chain().size == 2)
     assert(e.link(0U) == "because uniffi told me so")
 }
+try {
+    oops()
+    throw RuntimeException("Should have failed")
+} catch (e: kotlin.Exception) {
+    assert(e.toString() == "because uniffi told me so\n\nCaused by:\n    oops")
+}
 
 try {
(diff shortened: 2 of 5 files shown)


linabutler (approved):
Nice and simple, I love it! Great that `String(reflecting:)` works like a charm.

Thanks, @mhammond!

PR #2118, opened 2024-05-20
Allow errors to be tuple enums.

mhammond:
Follows on from #2004 which allowed tuple enums in plain enums, it's now possible to use tuple enums for errors.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
@@ -11,6 +11,7 @@ class {{ type_name }}(Exception):
 _UniffiTemp{{ type_name }} = {{ type_name }}
 
 class {{ type_name }}:  # type: ignore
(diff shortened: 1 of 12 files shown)


uniffi_bindgen/src/bindings/python/templates/ErrorTemplate.py
```diff
@@ -23,6 +24,21 @@ def __repr__(self):
     class {{ variant_type_name }}(_UniffiTemp{{ type_name }}):
         {%- call py::docstring(variant, 8) %}
 
+    {%-     if variant.has_nameless_fields() %}
+        def __init__(self, *values):
+            if len(values) != {{ variant.fields().len() }}:
+                raise TypeError(f"Expected a tuple of len {{ variant.fields().len() }}, found len {len(values)}")
```
bendk, line 30: Maybe this should say `Expected N arguments` instead, since `values` is packed from the arguments.

bendk (approved):
Looks great to me, just one suggestion for the error message.

PR #2119, opened 2024-05-20
Fix proc_macro defined errors when they aren't used as errors.

### LINKED ISSUE #2108: RustBuffer problem with nested Errors
Hi!

I'm having a problem I quite can't solve even after investigating.

I have the kinda following structure

crate 1, which exposes a couple of flat error enum:

```rust
#[derive(Debug, thiserror::Error, uniffi::Error)]
#[uniffi(flat_error)]
pub enum ForeignError {
	#[error("Error A: {0}")]
	ErrorA(#[from] ...),
	#[error("Error B: {0}")]
	ErrorB(#[from] ...),
}

#[derive(Debug, thiserror::Error, uniffi::Error)]
#[uniffi(flat_error)]
pub enum OtherForeignError {
	#[error("Error A: {0}")]
	ErrorA(#[from] ...),
	#[error("Error B: {0}")]
	ErrorB(#[from] ...),
}
```

crate 2, which consumes those 2 errors in a normal error enum:

```rust
#[derive(Debug, thiserror::Error, uniffi::Error)]
pub enum FfiError {
	#[error(transparent)]
	ForeignError(#[from] crate1::ForeignError),
	#[error(transparent)]
	OtherError(#[from] crate1::OtherForeignError),
}

type FfiResult<T> = Result<T, FfiError>;
```

On both Kotlin and Swift bindings I get those errors whenever `FfiError` is passed around: 

Kotlin: `java.lang.RuntimeException: junk remaining in buffer after lifting, something is very wrong!!` 
Swift: `UniffiInternalError.incompleteData "The buffer still has data after lifting its containing value"`

Any idea what causes this? This used to work before I upgraded to 0.27.x, but I can't poinpoint exactly which commit causes this regression.

mhammond:
When a `#[derive(uniffi::Error)]` didn't appear in any function signatures, the FfiConverter for the item was written to assume an error, but the bindings FfiConverter treated the item as a plain-old Enum. This caused runtime errors when trying to unpack the rustbuffers due to the disagreement about the buffer format.

As part of fixing this, the `Option<bool>` to try and represent the flatness of Enums/Errors was replaced with an `EnumFlatness` enum to make these states clearer and so the ComponentInterface can better understand the layout.

Fixes #2108


uniffi_meta/src/lib.rs
```diff
@@ -309,11 +309,33 @@ pub struct FieldMetadata {
...
+#[repr(u8)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Checksum)]
+pub enum EnumFlatness {
+    // It's an error and variants are exposed.
+    ErrorWithVariants,
+    // It's an error but variants are not exposed to the error enum
+    ErrorFlat,
+    // It's a regular enum, so flatness is based on the actual variants.
+    EnumValues,
+}
```
bendk, line 321: This is way nicer than before, but I'd love to do some more bikeshedding on the names/variants.  How do you feel about:
  -  `EnumStyle` instead of `EnumFlatness`.  I think this is now capturing more than just the flatness.
  - `Enum` instead of `EnumValues`.  To me, "Values" indicates that there are associated values, but I think this would be the variant for a non-error enum with no associated values.
  - `Error { is_flat: bool }` instead of the 2 error variants?  I'm not totally sure about this suggestion, but I think I like it.

bendk (commented):
Looks good to me, the enum definitely seems nicer than tracking an `is_flat` boolean around everywhere.  My only suggestion is maybe we can pick some different names for that enum.

uniffi_meta/src/lib.rs
mhammond: I think I quite like that suggestion in general, but I'll need to ponder more, because:

> I think this is now capturing more than just the flatness.

currently is really is just about the flatness.

jmartinesp:
Just FYI I was testing if these changes could fix an issue we're seeing in our project when using Kotlin with nested errors and I saw the nested errors are added to the bindings imports and usages in other crates as `FooError` but the generated code in the original crate is for `FooException`.

An example:

```kotlin
//  Bindings for original crate
sealed class QrCodeLoginException(message: String): kotlin.Exception(message) { ... }

// Bindings for crate that depends on that one
import uniffi.matrix_sdk.QrCodeLoginError // Error

class LoginException(
        val `error`: QrCodeLoginError // Error
        ) : QrLoginException() {
        override val message
            get() = "error=${ `error` }"
    }
```

uniffi_meta/src/lib.rs
bendk: I was thinking about [this block](https://github.com/mozilla/uniffi-rs/pull/2119/files#diff-554b47625fdbd9d813c73ea0e8feb461e1bf447d5496682b6cefb30c33b92d80R818) where we can also use the value to check if it's used as an error or not.

mhammond: Yeah, I was thinking a little more about this and you are 100% correct

mhammond:
> ```kotlin
> import uniffi.matrix_sdk.QrCodeLoginError // Error
> ```

Huh, that surprises me a little - is `QrCodeLoginError` defined by procmacro or by UDL? I'll try and repro that on my branch.

jmartinesp:
> > ```kotlin
> > import uniffi.matrix_sdk.QrCodeLoginError // Error
> > ```
> 
> Huh, that surprises me a little - is `QrCodeLoginError` defined by procmacro or by UDL? I'll try and repro that on my branch.

By [procmacro](https://github.com/matrix-org/matrix-rust-sdk/blob/poljar/qr-login/crates/matrix-sdk/src/authentication/qrcode/mod.rs#L42). The code that produces the broken bindings is in that same branch.

mhammond:
> Just FYI I was testing if these changes could fix an issue we're seeing in our project when using Kotlin with nested errors and I saw the nested errors are added to the bindings imports and usages in other crates as `FooError` but the generated code in the original crate is for `FooException`.

I fairly sure this actually a different problem and wasn't caused by this patch - could you please open a new issue?

mhammond:
I mean your initial report was actually 2 problems :) I think we might as well fix them separately.

uniffi_meta/src/lib.rs
mhammond: That worked out really well, thanks. I bike-shedded even more and ended up with `EnumShape` but I think it looks good!

bendk (approved):
Love the new names, this seems ready to merge to me.

PR #2120, opened 2024-05-21
Remove unused and incorrect defaults in Kotlin and Swift.

mhammond:
Both bindings supply a default for package_name/cdylib which is different to the default actually and unconditionally applied.

(I found this commit from a while ago and dusted it off - I think less confusion is good.)

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
@@ -99,20 +99,19 @@ pub struct CustomTypeConfig {
 }
 
 impl Config {
+    // We insist someone has already configured us - any defaults we supply would be wrong.
     pub fn package_name(&self) -> String {
-        if let Some(package_name) = &self.package_name {
-            package_name.clone()
-        } else {
-            "uniffi".into()
-        }
+        self.package_name
+            .as_ref()
+            .expect("package name should have been set in update_component_configs")
+            .clone()
     }
 
     pub fn cdylib_name(&self) -> String {
-        if let Some(cdylib_name) = &self.cdylib_name {
-            cdylib_name.clone()
-        } else {
-            "uniffi".into()
-        }
+        self.cdylib_name
+            .as_ref()
+            .expect("cdylib name should have been set in update_component_configs")
+            .clone()
     }
 
     /// Whether to generate immutable records (`val` instead of `var`)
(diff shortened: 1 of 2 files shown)


uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -99,20 +99,19 @@ pub struct CustomTypeConfig {
 }
 
 impl Config {
+    // We insist someone has already configured us - any defaults we supply would be wrong.
```
bendk, line 102: Slightly confused by this one.  Sometimes we do use the default package name and it's okay right?

I'm guessing that this function is only called in some specific circumstances where we do need an explicit package name, like maybe for external types?  If so, how about renaming the function to reflect that?

mhammond: > Slightly confused by this one. Sometimes we do use the default package name and it's okay right?

No - the problem is that there are 2 places a default is applied - eg, for kotlin, [here](https://github.com/mozilla/uniffi-rs/blob/962de986f02f726954b1dec2f5014d8f99828439/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs#L106) and by the binding generator [here](https://github.com/mozilla/uniffi-rs/blob/962de986f02f726954b1dec2f5014d8f99828439/uniffi_bindgen/src/bindings/kotlin/mod.rs#L38) - and as you can see, two different default values exist.

So we do use a default package name, but never use *this* default package name.

> I'm guessing that this function is only called in some specific circumstances where we do need an explicit package name,

The function is used by the templates - it's just never used when `self.package_name` remains None.

bendk: I'm starting to understand this one.  I think the `expect` message say something like `package name should have been set in update_component_configs`.  The way it currently reads, it seems like the issue was that the user didn't set it in the `config.toml`, but IIUC the issue is actually in the code.

bendk (approved):
Code's good. I have a nitpick on the error message but I'll let you make the call there.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
mhammond: > The way it currently reads, it seems like the issue was that the user didn't set it in the config.toml, but IIUC the issue is actually in the code.

Yeah, that's a very good point, I'll update the message.

PR #2132, opened 2024-05-27
Introduce `mike` to the doc generation process.

mhammond:
This causes the version selector to be rendered into the pages, however they will not be shown until we actually deploy version information.

Includes a script to help generate redirects for all our "old" pages.

On the path to #2016. Actual deployment using this system to follow once these are landed.

### DIFF
--- a/docs/manual/src/README.md
+++ b/docs/manual/src/README.md
@@ -1,0 +1,61 @@
+# The docs for the "manual".
+
+The manual is what we publish to https://mozilla.github.io/uniffi-rs/ and
+is generated from the content in this directory.
+
+This README is not published as part of the docs - but almost everything
+else in this directory is. This document describes how that publishing
+works.
+
+* This repo uses `mkdocs` - but **we must never deploy using that tool**.
+  Deploying using mkdocs will publish into the root of the site, **damaging
+  the versioning system we have in place.**
+
+* We use a [`mike`](https://github.com/jimporter/mike) to manage
+  the `mkdocs` build process and the deployment - it deploys to a versioned
+  (eg, `./0.27`) directory and manages aliases (eg, `latest`) on the site.
+
+* There are various HTML redirects and symlinks keeping things together. These
+  are maintained by hand. See also any docs related scripts in `./tools`.
+
+To put it another way: we never deploy the entire site, but instead use `mike`
+to deploy just a single version.
+
+## How to deploy a specific version.
+
+While the main branch is published as a version called `next` by CI, the
+steps to publish a specific version must be performed manually for now.
+
+* Ensure you have the release branch checked out.
+
+* Prepare your Python environment by executing:
+> pip install -r tools/requirements_docs.txt
+
+* From the root of your checkout, execute:
+> mike deploy 0.XX latest --update-aliases --push
+
+If you leave out `--push` your local `gh-pages` branch will be changed but it will not
+be pushed, which can be helpful if you want to check the result. You can later push
+the branch normally.
+
+`--update-aliases` is only strictly necessary if this is the first time a
+new version has been pushed and you want it to be the new "latest" version.
+
+## Layout of the published site
+The layout of the published pages is:
+
+* `./versions.json` is a file maintained by `mike` and used in the generated
+  html as a version selector.
+
+* `./next` is the directory where `mike` publishes the docs on currrent git `main`.
+
+* `./0.27`, `./0.28` etc are directories where `mike` publishes released versions.
+
+* `./latest` is a symlink managed by `mike` which points to the most recently released version.
+
+The other content is managed manually by the UniFFI maintainers by directly editing the `gh-pages` branch:
+
+* Other `*.html` files are the entry-points to our docs which existed before we versioned the docs.
+  They have been replaced with redirects to `0.27` - ie, `./Motivation.html` is content which redirects to `0.27/Motivation.html`.
+
+* `./index.html` is a little shim that redirects to `./latest`.
--- a/tools/make_old_docs_symlinks.py
+++ b/tools/make_old_docs_symlinks.py
@@ -1,0 +1,84 @@
+# Make redirects for our old html files relative to the root to version specific content.
+# See also ../docs/manual/src/README.md
+#
+# This is a "migration" script, so not run regularly and not by CI,
+#
+# We used to publish the entire site in the "root" - ie, we had/have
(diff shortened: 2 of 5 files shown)


docs/manual/src/README.md
```diff
@@ -0,0 +1,57 @@
...
+The manual is what we publish to https://mozilla.github.io/uniffi-rs/ and
+is generated from the content in this directory.
+
+This README is not published as part of the docs - but almost everything
+else in this directory is. This document describes how that publishing
+works.
+
+* This repo uses `mkdocs` - but **we must never deploy using that tool**.
+  Deploying using mkdocs will publish into the root of the site, **damaging
+ the versioning system we have in place.**
```
badboy, line 12:
```suggestion
  the versioning system we have in place.**
```

docs/manual/src/README.md
```diff
@@ -0,0 +1,57 @@
...
+  the `mkdocs` build process and the deployment - it deploys to a versioned
+  (eg, `./0.27`) directory and manages aliases (eg, `latest`) on the site.
+
+* There are various HTML redirects and symlinks keeping things together. These
+  are maintained by hand. See also any docs related scripts in `./tools`.
+
+To put it another way: we never deploy the entire site, but instead use `mike`
+to deploy just a single version.
+
+## How to deploy
```
badboy, line 24: I think  those should be slightly restructured so it's easier to follow.

docs/manual/src/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+* Prepare your Python environment by executing `pip install -r tools/requirements_docs.txt`
```
badboy, line 26: If this is not inline code but a block it might be easier to copy (double click) vs needing to carefully select the text.

docs/manual/src/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+* Although done by CI, we publish `main` as a version called `next` - so you 
+  could deploy this with `mike deploy next --push`
```
badboy, line 29: I don't understand this. Is this done automatically or is it not?
If it is, it shouldn't be in this section of the instructions. Let's focus on the manual steps that _need_ to be done.

docs/manual/src/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+* To publish a version after release, from that branch, execute `mike deploy 0.XX latest --update-aliases --push`
```
badboy, line 31: Why is this a manual step btw? Why is this not done on CI?

docs/manual/src/README.md
```diff
@@ -0,0 +1,57 @@
...
+The layout of the published pages is:
+
+* ./versions.json is a file maintained by `mike` and used in the generated
+  html as a version selector.
+
+* ./next is the directory where `mike` publishes the docs on currrent git `main`.
+
+* ./0.27, ./0.28 etc are directories where `mike` publishes released versions.
+
+* ./latest is a symlink managed by `mike` which points to the most recently released version.
```
badboy, line 50:
```suggestion
* `./versions.json` is a file maintained by `mike` and used in the generated
  html as a version selector.

* `./next` is the directory where `mike` publishes the docs on currrent git `main`.

* `./0.27`, `./0.28` etc are directories where `mike` publishes released versions.

* `./latest` is a symlink managed by `mike` which points to the most recently released version.
```

docs/manual/src/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+This README is not published as part of the docs - but almost everything
+else in this directory is. This document describes how that publishing
+works.
+
+* This repo uses `mkdocs` - but **we must never deploy using that tool**.
+  Deploying using mkdocs will publish into the root of the site, **damaging
+ the versioning system we have in place.**
+
+* We use a tool, [`mike`](https://github.com/jimporter/mike), which manages
```
badboy, line 14:
```suggestion
* We use a [`mike`](https://github.com/jimporter/mike) to manage
```

docs/manual/src/README.md
```diff
@@ -0,0 +1,57 @@
...
+* ./versions.json is a file maintained by `mike` and used in the generated
+  html as a version selector.
+
+* ./next is the directory where `mike` publishes the docs on currrent git `main`.
+
+* ./0.27, ./0.28 etc are directories where `mike` publishes released versions.
+
+* ./latest is a symlink managed by `mike` which points to the most recently released version.
+
+The other content is managed maually by directly editing the `gh-pages` branch:
```
badboy, line 52: I do not understand this sentence. Who edits the other content?

tools/make_old_docs_symlinks.py
```diff
@@ -0,0 +1,84 @@
+# Make redirects for our old html files relative to the root to version specific content.
+# See also ../docs/manual/src/README.md
+#
+# This is a "migration" script, so not run regularly and not by CI,
```
badboy, line 4: Does that mean this can go away once we did the full migration?

badboy (changes requested):

docs/manual/src/README.md
mhammond: Because I don't want to block on that for landing this. I'd love it if you helped do this after landing.

docs/manual/src/README.md
mhammond: Yes, pushing docs for the main branch is on CI, pushing docs for a just-released version is not. But sure, I'll remove this.

docs/manual/src/README.md
mhammond: Us. Anything in the root of the directory or otherwise not published by mike in a versioned directory is maintained by the uniffi maintainers manually.

tools/make_old_docs_symlinks.py
mhammond: As per the comment, it might make sense to keep it around so the top-level html redirects can be bumped to a later version. If we think we never ever want to move them from 0.27 it could be removed. In fact it could be removed now, because I've already run it, but am yet to push it to gh-pages.

docs/manual/src/README.md
mhammond: I tried to make that clearer.

docs/manual/src/README.md
badboy: Got it! Just wanted to make sure there's nothing fundamentally blocking this. Happy to help in a followup.

badboy (approved):

PR #2136, opened 2024-05-29
Change log level of methods name logged in Scaffolding

### LINKED ISSUE #1702: Use trace level logging in generated code
Is it possible to not have one function log every time its called?

I have a very simple function 

```rust
#[uniffi::export]
pub fn pod_restart_count(pod: Pod) -> i32 {
    pod.total_restart_count()
}
```

Its called a lot of times and drowns out all the other logs, would it be possible to opt out of logging on a per function basis?
Screenshot 2023-08-10 at 6 45 24 PM

Vinnstah:
Fixes #1702 

Change the log level to trace for method names logged by UniFFI internals when FFI calls into Rust. 

Augmented existing logging tests in Python, asserting that the changed level works as intended.

Let me know if you want me to put the tests in a **new** fixture instead of recycling an existing one.

### DIFF
--- a/fixtures/regressions/logging-callback-interface/Cargo.toml
+++ b/fixtures/regressions/logging-callback-interface/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "uniffi-fixture-logging-callback-interface"
 edition = "2021"
+authors = ["Firefox Sync Team <sync-team@mozilla.com>", "Viktor Jansson <v.l.jansson@gmail.com>"]
 version = "0.22.0"
 license = "MPL-2.0"
 publish = false
--- a/fixtures/regressions/logging-callback-interface/src/lib.rs
+++ b/fixtures/regressions/logging-callback-interface/src/lib.rs
@@ -6,7 +6,7 @@ use std::sync::{Once, RwLock};
 
 // Logger trait that the foreign code implements
 pub trait Logger: Sync + Send {
-    fn log_message(&self, message: String);
+    fn log_message(&self, message: String, level: String);
 }
 
 // Logger struct that implements the `log::Log` trait.
@@ -21,7 +21,7 @@ impl log::Log for RustLogger {
 
     fn log(&self, record: &log::Record<'_>) {
         if let Some(foreign_logger) = &*self.0.read().unwrap() {
-            foreign_logger.log_message(record.args().to_string());
+            foreign_logger.log_message(record.args().to_string(), record.level().to_string());
         }
     }
 
@@ -32,7 +32,7 @@ fn init() {
     static ONCE: Once = Once::new();
     ONCE.call_once(|| {
         log::set_logger(&RUST_LOGGER).expect("Error in install_logger()");
-        log::set_max_level(log::LevelFilter::Debug);
+        log::set_max_level(log::LevelFilter::Trace);
     });
 }
 
@@ -41,8 +41,20 @@ pub fn install_logger(logger: Box<dyn Logger>) {
     *RUST_LOGGER.0.write().unwrap() = Some(logger);
 }
 
-pub fn log_something() {
-    log::warn!("something");
+pub fn log_foo_at_warn() {
+    log::warn!("foo");
+}
+
+pub fn log_bar_at_info() {
+    log::info!("bar");
+}
+
+pub fn set_log_level_to_debug() {
+    log::set_max_level(log::LevelFilter::Debug)
+}
+
+pub fn log_buzz_at_error() {
+    log::error!("buzz");
 }
 
 uniffi::include_scaffolding!("test");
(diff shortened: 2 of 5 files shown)


fixtures/regressions/logging-callback-interface/Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "uniffi-fixture-logging-callback-interface"
 edition = "2021"
+authors = ["Firefox Sync Team <sync-team@mozilla.com>", "Viktor Jansson <v.l.jansson@gmail.com>"]
```
Vinnstah, line 4: Please let me know if you want me to revert this change. I git blamed and saw that Ben has contributed previously, let me know if you want me to add him as well (I think Mark is a member of the sync team?).

fixtures/regressions/logging-callback-interface/src/lib.rs
```diff
@@ -6,7 +6,7 @@ use std::sync::{Once, RwLock};
 
 // Logger trait that the foreign code implements
 pub trait Logger: Sync + Send {
-    fn log_message(&self, message: String);
+    fn log_message(&self, message: String, level: String);
```
Vinnstah, line 9: String is more clear than `u8`, and seems like overkill introducing a UniFFI exported Enum.

uniffi_macros/src/export/scaffolding.rs
Vinnstah, line 1: The actual fix of #1702

fixtures/regressions/logging-callback-interface/src/test.udl
Vinnstah, line 1: Will make sense when you read `test.py`...

Vinnstah (commented):
Left some clarifying comments.

Sajjon (approved):
Nice, I've been waiting for this fix actually, it will be most useful for me in [Sargon](https://github.com/radixdlt/sargon/pull/131), where I've recently worked with logging.

fixtures/regressions/logging-callback-interface/Cargo.toml
mhammond: Ben's also in the sync team - we'd also be up for removing this line entirely, but until then I think this is fine.

uniffi_macros/src/export/scaffolding.rs
```diff
@@ -285,7 +285,7 @@ pub(super) fn gen_ffi_function(
             #[doc(hidden)]
             #[no_mangle]
             pub extern "C" fn #ffi_ident(#(#param_names: #param_types,)*) -> ::uniffi::Handle {
-                ::uniffi::deps::log::debug!(#name);
+                ::uniffi::deps::log::trace!(#name);
```
mhammond, line 288: I guess I don't really understand why most of the above changes - ie, I'm not sure how exactly the fixture changes really ensures these 2 lines actually work. Or to put it yet another way, I think I'd r+ this with just these 2 lines.

mhammond: I'm not quite sure what you are saying there, but my point is that unless I'm still missing something, I think the test code here is testing the log crate rather than the uniffi crate?

Sajjon: The changed code in the test **piggybacks** on an existing test, but it does **indeed** test uniffi crate.

It tests the changed UniFFI source code. Without the changed source code the test would fail. 

The test ensures the level at which UniFFI source code logs name of methods is trace.

mhammond: > Without the changed source code the test would fail.

I hope it's clear that I do infact understand that.

Sajjon: @Vinnstah I would benefit from this PR getting landed :D 

Sounds like Mark would prefer if the changes in the test be reverted, maybe you can revert and this might get merged :)

mhammond: To be clear, I'm only weakly against tests for this - IMO the value is lower than the cost. But I'm open to be convinced otherwise, and it's difficult to argue too strongly *against* new tests really :)

mhammond:
Hi, is there any interest in removing some of the test code here? If I don't hear back soon, I'll put up the 2-line PR that actually fixes this and close this, but I'd like to offer the opportunity to have it completed here.

Vinnstah:
Hi Mark,
Apologies for the delay. I'll fix the tests and push a new commit asap.

praveenperera:
thank you for this! @Vinnstah !

bendk:
I didn't realize this issue has been sitting in the queue for so long.  Does https://github.com/mozilla/uniffi-rs/pull/2211 handle your needs or is there still work to be done here?

bendk:
I should have pointed you to https://github.com/mozilla/uniffi-rs/pull/2275.  As far as I know, there are no more log calls in the current code.  I don't think we even have a dependency on the crate.  Those calls were replaced by an internal logging mechanism, that's disabled by default.

praveenperera:
@bendk I'm using master now and its working, no more noisy logs thank you!

mhammond:
We've recently removed all `log` calls from the generated scaffolding.

PR #2139, opened 2024-05-30
Deploy the docs site using mike

mhammond:
I'm not 100% sure this actually works, but reading the circle docs and looking at jj it seems like it should.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,10 +122,16 @@ jobs:
     steps:
       - checkout
       - install-mkdocs
-      - run: python -m mkdocs build
-      - gh-pages/deploy:
-          build-dir: target/rendered-docs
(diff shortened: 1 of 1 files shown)


badboy (approved):

mhammond:
Still not clear if it *actually* works and will not be until a docs change is pushed. CI for the job shows:

```
INFO    -  Cleaning site directory
INFO    -  Building documentation to directory: /home/circleci/project/target/rendered-docs
INFO    -  The following pages exist in the docs directory, but are not included in the "nav" configuration:
  - index.md
WARNING -  Doc file 'tutorial/foreign_language_bindings.md' contains a link '../Overview.md', but the target 'Overview.md' is not found among documentation files.
INFO    -  Documentation built in 0.77 seconds
warning: nothing changed in commit
  To create a commit anyway, retry with --allow-empty
```
which is what I expected. I guess I should just put a PR up to fix that warning and see what happens :)

Note that the gh-pages branch is correctly deployed though, so the versioning is all in place.

PR #2150, opened 2024-06-12
New custom type system

bendk:
**PR notes**: I split this out out from #2087 so that we can discuss the custom type changes separate from the remote type changes.  I think the new syntax is pretty uncontroversial, but this does require a breaking change so I wanted to make room to discuss that separate from the main conversation.  On the topic of breaking changes: I tried but failed on avoiding them.  Maybe it's possible with some extra work, but I don't think it's worth it at this point.  How does that sound to others?

---

Implemented a new custom type system described in the new docs.  This system is easier to use and also allows us to do some extra tricks behind the scene.  For example, I hope to add a flag that will toggle between implementing `FfiConverter` for the local tag vs a blanket impl. It's not possible to do that with the current system and adding support would be awkward.

I wanted to keep backwards compatibility for a bit, but couldn't figure out a good way to do it.  The main issue is that for the old system, if a custom type is declared in the UDL then we generate the `FfiConverter` implementation, while the new system expects the user to call `custom_type!` to create the impl.  I tried to get things working by creating a blanket impl of `FfiConverter` for types that implemented `UniffiCustomTypeConverter`, but ran into issues -- the first blocker I found was that there's no way to generate `TYPE_ID_META` since we don't know the name of the custom type.

Removed the nested-module-import fixture.  The custom type code will no longer test it once we remove the old code, since it's not using `UniffiCustomTypeConverter`.  I couldn't think of a way to make it work again.


bendk:
@mhammond I just added an upgrading guide for users.  I'm not sure how to hook it in to the new doc system though.  Does this mean updating the `gh-pages` branch after we merge this?  Maybe you could walk me through it sometime.

mhammond (commented):
oops, I had these pending but I think they are still worthwhile

mhammond:
> I'm not sure how to hook it in to the new doc system though. Does this mean updating the gh-pages branch after we merge this? Maybe you could walk me through it sometime.

I'm not quite sure what you are asking - you need to list it in https://github.com/mozilla/uniffi-rs/blob/main/mkdocs.yml, but what's less clear is how this new file (which I think could maybe be renamed to just "upgrading.md"?) will function. Off-hand, I'd say our release process should be tweaked to say (a) remove old entries from this file and (b) have a quick check over the changelog for new items that should be added.

Or did you have something else in mind?

docs/manual/src/udl/custom_types.md
```diff
@@ -1,72 +1,95 @@
 # Custom types
 
-Custom types allow you to extend the UniFFI type system to support types from your Rust crate or 3rd
-party libraries. This works by converting to and from some other UniFFI type to move data across the
-FFI. You must provide a `UniffiCustomTypeConverter` implementation to convert the types.
+Custom types allow you to extend the UniFFI type system by converting to and from some other UniFFI
+type to move data across the FFI.
```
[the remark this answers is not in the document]
bendk: Great idea.  I added an overview/motivation for custom types here and thought it looked good so I didn't move the `custom_newtype` example.  WDYT?

bendk:
> > I'm not sure how to hook it in to the new doc system though. Does this mean updating the gh-pages branch after we merge this? Maybe you could walk me through it sometime.
> 
> I'm not quite sure what you are asking - you need to list it in https://github.com/mozilla/uniffi-rs/blob/main/mkdocs.yml, ...

I was just looking for that YAML file, thanks!

jeddai (commented):
I really like this adjustment — I feel like it works better with the switch to macros for the bindings for other types

mhammond (commented):
I've spent more time trying to understand the bigger picture here than looking at the implementation details of this - which got me back to looking more at #2087 so see how all these pieces fit together. I think I'm getting my head around it and I like it!

Here's a possibly controversial take on the naming though:

I never really liked the old names very much (I think I might have introduced them),
and sadly I think I like the new ones even less - but I see an opportunity:

So here's the old world:
```
impl UniffiCustomTypeConverter for MyType {
    type Builtin = UniFFIBuiltinType;

    fn into_custom(val: Self::Builtin) -> uniffi::Result<Self> {
        ...
    }

    fn from_custom(obj: Self) -> Self::Builtin {
        ...
    }
}
```
What I like about this is the liberal use of the term "builtin" and the fact that the types are explicit.
It's clear what `Self` refers to and it's very clear what's the "builtin" vs the "custom".

Here's the new world:
```
// Expose `http::HeaderMap` to Uniffi.
uniffi::custom_type!(HeaderMap, Vec<HttpHeader>, {
    from_custom: |obj| { ... }
    try_into_custom: |val| { ... }
}
```
What I don't like about this is that the term "builtin" is no longer used and there are no types in the signatures.
Instead, just the term "custom" appears, and without much experience with UniFFI it might not be obvious what is meant by the term (eg, it might be easy to convince yourself that the term "custom" actually is referring to the builtin if that's the only term actually used)

When I read the other docs and the implementation though, I see:

```
// Defining `From<Handle> for i64` also gives us `Into<i64> for Handle`
impl From<Handle> for i64 { ... }
impl TryFrom<i64> for Handle { ... }
```

Which makes me think - why not change the new model to:

```
// Expose `http::HeaderMap` to Uniffi.
uniffi::custom_type!(HeaderMap, Vec<HttpHeader>, {
    into_builtin: |obj| { ... }
    try_from_builtin: |val| { ... }
}
```

The improvements here IMO are:
* The types become that little more obvious - what the "builtin" means seems more obvious.
* It better matches how Rust names the equivalent functionality, so is more likely to be familiar to the reader.
* It better matches the actual implementation and makes the docs that bit easier to read.

eg, the docs can say something like "into_builtin defaults to `Into<Builtin>` and "try_from_builtin" defaults to `TryFrom<Builtin>`.
Indeed, they can even encourage you to implement those builtin traits instead of manually specifying the "overrides" (which they kinda do in your patch, but having this consistency makes that story stronger)

It could be argued it makes the breakage that little bit more difficult to mechanically fix,
but it seems like it is actually a much better fit for how it actually works and seems far more intuitive for new UniFFI users.
It's an opportunity we will probably not get again.

(I'll also admit this came to me just last night, so I haven't thought it through in great detail, but though it worth throwing out there anyway!)

bendk:
> Instead, just the term "custom" appears, and without much experience with UniFFI it might not be obvious what is meant by the term (eg, it might be easy to convince yourself that the term "custom" actually is referring to the builtin if that's the only term actually used)

This is a very good point.  I'd love to get a very specific and easy to understand name here.

The reason I didn't go with "builtin" is that type doesn't actually need to be a builtin type. You could convert the custom type into a user-defined record/enum/interface.  Would this actually happen in practice?  I wasn't really sure.

Maybe this would be a good opportunity to introduce a term here.  There were a lot of times when I was writing docs that I wanted a word for a type that had the FFI traits implemented, but couldn't find a good one.  Sometimes I wanted to call it a "UniFFI type", but it's not so clear what that means from the words alone.  I usually ended up saying something like "a type that can be used in the exported UniFFI API", which was pretty awkward.  I think it could be good to pick a term and add a section in the manual that defined it.  "UniFFI type" or "UniFFI-enabled type" are the best names that I could come up with.  WDYT?

> ..the docs can say something like "into_builtin defaults to Into<Builtin> and "try_from_builtin" defaults to TryFrom<Builtin>

True, but isn't this also true for the current code?  You could say "`from_custom` defaults to `From<Custom>`".  I think the main issue is that "custom" isn't a great name.

mhammond:
> The reason I didn't go with "builtin" is that type doesn't actually need to be a builtin type. You could convert the custom type into a user-defined record/enum/interface. Would this actually happen in practice? I wasn't really sure.

This is the http-header example right? The "custom type" is a remote type and the "builtin" type is a record?

I agree "builtin" isn't a great name, but I think it's important the *concept* have a name.

> I think the main issue is that "custom" isn't a great name.

My take is that:
* It's ideal that the concept clearly includes 2 different terms - one for the new type being exposed to the object model (what we call the "custom type") and the other being the "uniffi compatible type used to represent that custom type" (ie, what we call the builtin).
* The old model wasn't great, but the fact it used 2 distinct terms meant less confusion was likely. Even if we came up with a better name for "custom", I think the fact only one name is used is a sticking point - users may be confused about which of the 2 types involved it refers to.

mhammond:
or to put my last comment another way - I think the term "builtin" is more problematic than "custom", because it implies records are can not be used - but having 2 distinct "not ideal" terms is somewhat better than just using a single "ideal" term.

bendk:
Using two terms for the two different types makes a lot of sense.  What if we leaned into that more and used a syntax like this?

```
uniffi::custom_type!(
   custom: HeaderMap,
   builtin: Vec<HttpHeader>,
   into_builtin: |obj| { ... },
   try_from_builtin: |val| { ... },
);
```

- Maybe replacing "builtin" with some other term that we come up with.
   - We could call it `uniffi` or `uniffi_type`, but those names don't feel right to me.  "UniFFI" feels too overloaded to be meaningful to me.
   - What about `converted`?  That would be a horrible name by itself, but if you think about "custom" and "converted" together it makes some sense.  The docs could say explain it as you're converting your custom type into a UniFFI compatible type.
- Maybe brackets instead of parens?   I think Rust lets you use either interchangeably so it's just a question of which one we use in our docs/examples.

mhammond:
> uniffi::custom_type!(
>    custom: HeaderMap,
>    builtin: Vec<HttpHeader>,
>    into_builtin: |obj| { ... },
>    try_from_builtin: |val| { ... },
> );

What I *do* like about this is what I liked about my idea above - ie, I think this is basically as good as simply
```
uniffi::custom_type!(HeaderMap, Vec<HttpHeader>
    into_builtin: |obj| { ... },
    try_from_builtin: |val| { ... },
);
```
because it does use the 2 terms. In both cases though, the sticking point remains more "builtin" than "custom". Another bad option might be "existing"?

bendk:
I've been thinking about this one off and on for a bit and "existing" seems like the best option to me, should we just go with that one?

mhammond:
> should we just go with that one?

That sgtm, at least as the next straw-man!

docs/manual/src/Upgrading.md
```diff
@@ -0,0 +1,35 @@
...
+
+```
+uniffi::custom_type!(MyType, UniFFIBuiltinType, {
+    try_from_existing: |val| { ... },
+    into_existing: |obj| { ... },
+})
+```
+
+The custom_type macro is more flexible than the old system.  For example, the `try_from_existing` and
+`into_existing` can be omitted in many cases, and will use the `TryInto` and `From` traits by default.
```
mhammond, line 34: Reading the docs below, it looks like we might also use `Into` and `TryFrom` depending on the direction/context. If so, could you please tweak this? In particular, this sentence almost seems like the reverse of what's sensible (ie, that we have a `try_from` method, which uses `From` if omitted, and we have an `into` method which, if omitted, uses `TryInto` appears wrong and inconsistent).

(If I'm not wrong about that, the changes make here could make the text quite vague, which I think is OK if the alternative makes things appear more restrictive than it is)

docs/manual/src/Upgrading.md
```diff
@@ -0,0 +1,35 @@
...
+        ...
+    }
+}
+```
+
+After:
+
+```
+uniffi::custom_type!(MyType, UniFFIBuiltinType, {
+    try_from_existing: |val| { ... },
```
mhammond, line 28: really nit-picking here, but I wonder if this line should say, eg, `try_from_existing: |val| { Ok(...) },` to highlight the fact this is a `Result<>`?

docs/manual/src/udl/custom_types.md
```diff
@@ -1,72 +1,107 @@
...
+These types can be passed across the FFI as the primitive type, which can be more efficient than passing a struct.
+The foreign bindings will treat these types as the primitive type, for example `Guid` could appear a string to the foreign code.
+
+Custom types can also be customized on the foreign side.  For example, a URL could be:
+* Represented by the `url::Url` type in Rust
+* Passed across the FFI as a string
+* Represented by the `java.net.URL` type in Kotlin
+
+Lastly, the `Guid(String)` example shows another benefit of custom types: they can be used to
+support Rust types that are not currently supported by UniFFI like tuple-style structs.
```
mhammond, line 16: tuple structs are supported by procmacros.

docs/manual/src/udl/custom_types.md
```diff
@@ -1,72 +1,107 @@
...
+/// `SerializableStruct` objects will be passed across the FFI the same way `String` values are.
+uniffi::custom_type!(SerializableStruct, String);
 ```
 
-In this trivial example, the simplest way to expose this is with a macro.
+By default:
+
+ - Values passed to the foreign code will be converted using `<SerializableStruct as Into<String>>` before being lowered as a `String`.
+ - Values passed to the Rust code will be converted using `<String as TryInto<SerializableStruct>>` after lifted as a `String`.
+ - The `TryInto::Error` type can be anything that implements `Into<anyhow::Error>`.
```
mhammond, line 37: My brain hurts - following on from the last comment - have we got these names correct? We have `into_existing` which doesn't return a result, but here talking about "TryInto" (which does) seems confusing and backwards.

I'm worried my last suggestion which gave us these new names actually made things more complicated?

mhammond (commented):
This implementation and motivation is great, thanks for your patience while we catch up with your thinking.

I'm now a bit worried that my naming suggestions of before have actually made things worse. Maybe some small tweaks to the docs are all that's necessary, but I do think it's still worth thinking about before landing - that naming is almost the *only* part of this which will be visible by users so it's worth getting right, and I'm really sorry if I'm just making it worse or more confusing.  Once I get over that I think this is ready to go.

docs/manual/src/Upgrading.md
bendk: Thinking about this some more, I think we should spell out the exact trait and always use the into version.  So this should say "For example, the `try_from_existing` and `into_existing` can be omitted in many cases, and will use the `TryInto<Custom>` and `Into<Existing>` traits by default."  The reason for this is that Into is auto-derived from From, but not the other way around.

Maybe we should rename `try_from_existing` to `try_into_custom` to match this?

docs/manual/src/udl/custom_types.md
bendk: I think I just got this backwards and I'll reverse them.

bendk:
Please nit-pick, I agree that the names are very important here.  I just pushed some fixes/clarifications to the docs.

I'm currently 50/50 on renaming `try_from_existing` to `try_into_custom`.  What's your feeling on this?

docs/manual/src/Upgrading.md
```diff
@@ -0,0 +1,37 @@
...
+    fn from_custom(obj: Self) -> Self::Builtin {
+        ...
+    }
+}
+```
+
+After:
+
+```
+uniffi::custom_type!(MyType, UniFFIBuiltinType, {
```
mhammond, line 27: If we are moving to the term "existing" then this example should maybe be change - eg, line 12 could say, eg, `type Builtin = AnExistingUniFFIType;`  and this line could use the same name - it avoids using the name "builtin" when describing the new world.

docs/manual/src/Upgrading.md
```diff
@@ -0,0 +1,37 @@
...
+    try_from_existing: |val| { Ok(...) },
+    into_existing: |obj| { ... },
+})
+```
+
+The custom_type macro is more flexible than the old system.  For example, the `try_from_existing`
+and `into_existing` can be omitted in many cases.  If `try_from_existing` is omitted, then the
```
mhammond, line 34: nit - trailing "the" on this line should be removed.

docs/manual/src/Upgrading.md
```diff
@@ -0,0 +1,37 @@
...
+`TryFrom<UniFFIBuiltinType>` will be used instead.  Likewise, if `into_existing` is omitted, then
+`From<MyType>` will be used instead. See the [Custom Types](./udl/custom_types.md) for details and
+other features of the new macro.
```
mhammond, line 37: Thanks for updating these docs. As above though, if we stick with the term "existing" we should try and remove `UniFFIBuiltinType` entirely from these docs. Maybe even replace `MyType` with `NewCustomType` or something to fully reinforce this? I think that's largely independent of how we name the 2 methods - IIUC, we want to avoid the term "builtin" regardless, right?

So the very first part of the docs is perfect - ie, "If `try_from_existing` is omitted, then `TryFrom<AnExistingUniFFIType>` will be used" seems completely logical.

However, the second part is a little less so. "Likewise, if `into_existing` is omitted, then
`From<NewCustomType>` will be used instead." doesn't have quite the same symmetry. Should this then be `from_custom`?

That would leave us with:
"""
uniffi::custom_type!(NewCustomType, AnExistingUniFFIType, {
    try_from_existing: |val| { Ok(...) },
    from_custom: |obj| { ... },
})

The custom_type macro is more flexible than the old system. For example, `try_from_existing` and `from_custom` can be omitted in many cases. If `try_from_existing` is omitted, then `TryFrom<AnExistingUniFFIType>` will be used instead. Likewise, if `from_custom` is omitted, then `From<NewCustomType>` will be used instead.
"""

which seems to better capture reality and makes it much clearer about what the user is expected to implement to make this work, even if it is missing the symmetry of from/info.

WDYT?

docs/manual/src/udl/custom_types.md
```diff
@@ -1,72 +1,105 @@
...
+use some_mod::SerializableStruct;
+
+/// `SerializableStruct` objects will be passed across the FFI the same way `String` values are.
+uniffi::custom_type!(SerializableStruct, String);
 ```
 
-In this trivial example, the simplest way to expose this is with a macro.
+By default:
+
+ - Values passed to the foreign code will be converted using `<SerializableStruct as TryInto<String>>` then lowered as a `String`.
```
mhammond, line 32: If we did the above, then these docs would need some minor tweaking and would also be a good opportunity to reinforce the normal Rust symmetry about Rust's Into/From.

mhammond (commented):
Think of these comments more as brainstorming than a concrete proposal :)

docs/manual/src/Upgrading.md
jplatte: Nitpick: Should consistently use `UniFfi` in Rust type namespace.

uniffi_macros/src/custom.rs
```diff
@@ -2,96 +2,263 @@
...
+                ..Self::default()
+            })
+        } else if lookahead.peek(kw::try_from_existing) {
+            input.parse::<kw::try_from_existing>()?;
+            input.parse::<Token![:]>()?;
+            Ok(Self {
+                try_from_existing: Some(input.parse()?),
+                ..Self::default()
+            })
+        // Temporarily disabled until we land #2807 and get the new remote types system
```
badboy, line 91:
```suggestion
        // FIXME(#2807): Temporarily disabled until we land #2807 and get the new remote types system
```

makes it easy to grep for it.

docs/manual/src/Upgrading.md
bendk: One more option that I like is using the names `try_from_existing` and `from_custom`, since Rust encourages users to implement `From` and also using the less-specific description from the 3rd option.

docs/adr/0009-remote-types-interfaces-and-proc-macros.md
```diff
@@ -160,14 +160,14 @@ pub enum LogLevel {
 
 /// Define a custom type conversion from `log::Level` to the above type.
```
gruberb, line 161:
```
/// Define a type conversion from `log::Level` to the custom [`LogLevel`] enum.
```

docs/adr/0009-remote-types-interfaces-and-proc-macros.md
```diff
@@ -160,14 +160,14 @@ pub enum LogLevel {
...
 uniffi::custom_type!(log::Level, LogLevel, {
-  from_custom: |l| match l {
+  into_existing: |l| match l {
```
gruberb, line 163: The word `existing` doesn't sound right to me. Everything is existing, really. They way I see it, is that we have `builtin` types which come from Rust, we have `custom` types which we create, and we have `external` types from other crates.

docs/adr/0009-remote-types-interfaces-and-proc-macros.md
```diff
@@ -160,14 +160,14 @@ pub enum LogLevel {
...
     log::Level::Error => LogLevel::Error,
     log::Level::Warn => LogLevel::Warn,
     log::Level::Info => LogLevel::Info,
     log::Level::Debug => LogLevel::Debug,
     log::Level::Trace => LogLevel::Trace,
   },
-  try_into_custom: |l| Ok(match l ({
+  try_from_existing: |l| Ok(match l ({
```
gruberb, line 170: Same, would rather see `try_into_custom`.

docs/manual/src/udl/custom_types.md
```diff
@@ -1,72 +1,105 @@
...
+```rust
+uniffi::custom_type!(SerializableStruct, String, {
+    into_existing: |s| s.serialize(),
+    try_from_existing: |s| s.deserialize(),
+});
 ```
+
+### custom_newtype!
+
+The `custom_newtype!` can trivially handle newtypes that wrap a UniFFI type.
```
gruberb, line 51: I would not put "trivially" in any sort of documentation. Maybe:

> The `custom_newtype!` macro is able to handle custom Rust structs which wrap a UniFFI type by default.

I don't want to go there yet, but the `custom_type!` macro name could be a bit more telling. It's not at all obvious what it does.

examples/custom-types/src/lib.rs
```diff
@@ -1,98 +1,70 @@
 use url::Url;
 
 // A custom guid defined via a proc-macro (ie, not referenced in the UDL)
-// By far the easiest way to define custom types.
 pub struct ExampleCustomType(String);
+
+// custom_newtype! is the easiest way to define custom types.
```
gruberb, line 6: Also, `custom_newtype!` is not obvious what it does from the name. But I think this one is used throughout the codebase.

gruberb (commented):
Just a few thoughts. I think it's a bit confusing reading `existing`, `custom_newtype` and `custom_type`.

I would like to define overall naming patterns somewhere. 

We have:
* Built-in types from the Rust world
* User created types via `struct` and `enum`
* UniFFI types
* Types from external crates

How about:
* `primitive_type`
* `custom_type`
* `bridge_type`
* `external_type`

There is a lot of context that I am properly missing. But might be worth defining the terms we use first, and then adjust the conventions in the codebase.

docs/manual/src/udl/custom_types.md
bendk: Good point.  I adjusted the language slightly, but hopefully captured your overall point here.  Feel free to re-open if there are more changes here though.

bendk:
It's great to hear a take coming from someone who hasn't been so deep in the weeds on this one.  I'm going to try to rephrase it slightly, please tell me if I'm missing anything:

1) There should be an overview of the UniFFI type system that describes all of the kinds of types in broad strokes.

2) One obvious split is types that are defined by UniFFI (`u32`, `String`, `HashMap`, etc.) vs ones defined outside of UniFFI via UDL/proc-macros.  This split should inform the terms we use.  My preferred terms are:
  - `builtin types` for the first category.  I like this better than `primitive`, since `HashMap` doesn't feel like a primitive type to me.
  - `user-defined types` for the second category.
  - Stop using the term `custom types` since `custom type` could easily be interpreted as `user-defined type`.

3) Use a term like `bridge type` for the feature we're describing here to more clearly distinguish it from other user-defined types.  `bridge type` would work for me, but I like `adapter type` or `converter type` a bit better.  Are there other terms like this that we can consider?

4) `custom_newtype` is confusing.  You've convinced me on this and made we want to step back and reconsider this feature altogether.  It seems to me that what would be really nice is first-class support for newtypes.  You could have Rust code like this:

```
#[derive(uniffi::newtype)]
pub struct Guid(String);
```

On the Rust side, this would work exactly like a custom_newtype today.  On the foreign side, we could define a `Guid` newtype rather than passing a raw string around.  One potential drawback is that users can currently define their own custom type converters for custom newtypes on the foreign side, for example maybe they could convert them into some standard Kotlin `Guid` type.  I think this is tractable though.

At this point, I think we're talking about more than one PR worth of work.  Maybe we could split this up into multiple ones:
- This PR is where we pick the names/syntax for bridge types/adapter types/whatever we end up calling them.  It will implement the `custom_newtype` macro, but we'll plan to rework this before the next release.
- A PR to add the type overview manual page
- A PR that reworks newtypes
- A PR that reworks converting types in the foreign bindings.

jplatte:
> Use a term like `bridge type` for the feature we're describing here to more clearly distinguish it from other user-defined types. `bridge type` would work for me, but I like `adapter type` or `converter type` a bit better. Are there other terms like this that we can consider?

Idk whether this question was directed at @gruberb or everybody, but personally I really like `bridge(d) type`! :)

bendk:
If others like bridge type, that works for me.

bendk:
We discussed this today, got some more clarity and came up with some decisions.  I misunderstood part of Bastian's point.  He's fine with "custom type" as a name, the issue was with the other type involved, which he wants to name "bridge type".

The new plan is to:
  - Keep "custom types" as the name for this feature
  - Call the types involved "custom type" and "bridge type"
  - Refer to "bridge type" in the macro.  Use `into_bridge_type` and `try_from_bridge_type` as the attribute names.

bendk:
Updated the language as describe above.  I think this is maybe finally ready to land, tell me what you think.

docs/manual/src/udl/custom_types.md
```diff
@@ -1,72 +1,110 @@
...
 
 ```rust
-pub struct Handle(i64);
+
+// Some complex struct that can be serialized/deserialized to a string.
+use some_mod::SerializableStruct;
+
+// When passing `SerializableStruct` objects to the foreign side, they will
+// be converted to a `String`, then `String` will be used as the bridge type
+// to pass the value across the FFI. Conversly, when objects are passed to
```
mhammond, line 28: typo: Conversely

also: maybe explicitly note here that this example assumes the relevant Into/TryInto implementations already exist? (it's implied, but maybe worth spelling out)

docs/manual/src/udl/custom_types.md
```diff
@@ -1,72 +1,110 @@
...
+Default conversions to the bridge type:
+
+- Values passed to the foreign code will be converted using `Into<String>` then lowered as a `String` value.
+- Values passed to the Rust code will lifted as a `String` then converted using `<String as TryInto<SerializableStruct>>`.
+- The `TryInto::Error` type can be anything that implements `Into<anyhow::Error>`.
+- `TryFrom<String>` and `From<SerializableStruct>` will also work, using the blanket impl from the core library. 
+
+### custom_type! with manual conversions
 
+You can also manually specify the conversions by passing extra params to the macro:
```
mhammond, line 45: like above, maybe add to the end here ", for cases where the trait implementations do not already exist, or they aren't desirable in this context for some reason"?

mhammond (approved):
This is great, thanks so much for your patience shepherding this through! I'd still love to get more eyes on this, so maybe give it a week or so for them to magically materialize :)

jhugman:
I'm really happy to see this work happen. Thank you so much Ben!

Having read the discussion about names, I totally agree with you about `custom` and `builtin`, and see the difficulties in picking a ~~new color of bikeshed~~ more descriptive name.

My suggestion would be to keep the `custom type` naming, but lean into the lifting and lowering metaphors, specifically use a `lowered` type instead of `builtin`:

```rs
uniffi::custom_type!(HeaderMap, Vec<HttpHeader>
    into_lowered: |obj| { ... },
    try_from_lowered: |val| { ... },
);
```

alternates, using verbs:
```rs
uniffi::custom_type!(HeaderMap, Vec<HttpHeader>
    lower: |obj| { ... },
    try_lift: |val| { ... },
);
```

My hunch would be that if you're at the stage of adding new types to cross the FFI, then you've already found the lifting and lowering concepts already prominent in the docs.

bendk:
> My suggestion would be to keep the `custom type` naming, but lean into the lifting and lowering metaphors, specifically use a `lowered` type instead of `builtin`:

This is a pretty interesting idea.  Conceptually, it means that values would get lowered/lifted several times (`Guid` is lowered to `String` is lowered to `RustBuffer`).  That adds some complexity to the concept of lifting/lowering, but then you don't need to introduce a new concept of "converting into/from the existing type".  I remember you suggesting something like this way at the beginning.

My initial feeling is that I like it, what do others think?

bendk:
I like the suggestion of going with `lower` and `lift` because IMO it lets is paint a better picture in the description.  I updated the documentation to lean into this, tell me what you think.

bendk:
We discussed this yesterday and there's general agreement to lean into the lower/lift terminology to describe this.  I updated the PR to use `lower`/`try_lift` instead of `into_bridge_type`/`try_from_bridge_type`.

I'd love to get this merged.  I think @mhammond has some suggestions for the new doc page, does anyone else have suggestions/comments?

bendk:
Merged some great updates to the docs from Mark and updated the changelog.  Once this passes CI I'm going to hit the merge button.

Issue #2151, opened 2024-06-13
Support for projects not using `cargo`


emarteca:
Currently, the `library_mode` `generate_bindings` calls `cargo` to retrieve metadata about the project for which bindings are being generated: e.g. `generate_bindings` starts with [this line](https://github.com/mozilla/uniffi-rs/blob/cd38ccea8236df7d93aff336c325a3a8e524af5d/uniffi_bindgen/src/library_mode.rs#L45):
```
let cargo_metadata = MetadataCommand::new()
        .exec()
        .context("error running cargo metadata")?;
```

This means `uniffi-rs` does not work for projects that use a different build system, where `cargo` is not available and/or where the project is missing a `Cargo.toml` file.

It would be nice if `uniffi-rs` was generalized to not be reliant on `cargo`.

mhammond:
Hopefully fixed by #2175

PR #2152, opened 2024-06-13
Removing cargo use

emarteca:
This removes the use of `cargo` in `library_mode`, in order to support systems that do not have `cargo` available. It also removes the requirement that the project for which bindings are being generated have a `Cargo.toml` in the root. 

This also removes reading UDL files in `library_mode`'s `find_components`; it was not necessary for our test cases so far, and this code had more complex use of the data extracted from `cargo metadata`. 

This is a first attempt at addressing #2151; it is a bit hacky, particularly in the `find_components` and the addition of command-line arguments to pass in the package root and crate name (which was metadata previously extracted with `cargo metadata`).

Would appreciate any thoughts on how to improve this implementation of `cargo`-free `library_mode`. Thanks!

### DIFF
--- a/uniffi/src/cli.rs
+++ b/uniffi/src/cli.rs
@@ -108,6 +108,12 @@ enum Commands {
 
         /// Path to the UDL file, or cdylib if `library-mode` is specified
         source: Utf8PathBuf,
+
+        #[clap(long)]
+        package_name: Option<String>,
+        
+        #[clap(long)]
+        crate_root: Option<Utf8PathBuf>,
     },
 
     /// Generate Rust scaffolding code
@@ -138,6 +144,8 @@ fn gen_library_mode(
     cfo: Option<&camino::Utf8Path>,
     out_dir: &camino::Utf8Path,
     fmt: bool,
+    package_name: Option<String>,
+    crate_root: Option<&camino::Utf8Path>,
 ) -> anyhow::Result<()> {
     use uniffi_bindgen::library_mode::generate_bindings;
     for language in languages {
@@ -158,6 +166,8 @@ fn gen_library_mode(
                 cfo,
                 out_dir,
                 fmt,
+                package_name.clone(),
+                crate_root,
             )?
             .len(),
             TargetLanguage::Python => generate_bindings(
@@ -167,6 +177,8 @@ fn gen_library_mode(
                 cfo,
                 out_dir,
                 fmt,
+                package_name.clone(),
+                crate_root,
             )?
             .len(),
             TargetLanguage::Ruby => generate_bindings(
@@ -176,6 +188,8 @@ fn gen_library_mode(
                 cfo,
                 out_dir,
                 fmt,
+                package_name.clone(),
+                crate_root,
             )?
             .len(),
             TargetLanguage::Swift => generate_bindings(
@@ -185,6 +199,8 @@ fn gen_library_mode(
                 cfo,
                 out_dir,
                 fmt,
+                package_name.clone(),
+                crate_root,
             )?
             .len(),
         };
@@ -257,6 +273,8 @@ pub fn run_main() -> anyhow::Result<()> {
             source,
             crate_name,
             library_mode,
+            package_name,
+            crate_root,
         } => {
             if library_mode {
                 if lib_file.is_some() {
@@ -273,6 +291,8 @@ pub fn run_main() -> anyhow::Result<()> {
                     config.as_deref(),
                     &out_dir,
                     !no_format,
+                    package_name,
+                    crate_root.as_deref(),
                 )?;
             } else {
                 gen_bindings(
--- a/uniffi_bindgen/src/bindings/kotlin/test.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/test.rs
@@ -42,6 +42,8 @@ pub fn run_script(
         None,
         &out_dir,
         false,
+        None,
+        None,
     )?;
     let jar_file = build_jar(crate_name, &out_dir, options)?;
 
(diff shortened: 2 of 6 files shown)


mgeisler:
Thanks Ellen for putting this up! For more context, the need for using UniFFI without Cargo has come up in the work I've been doing in Android and in Google in general. There, we either use [Soong](https://source.android.com/docs/setup/build) or [Bazel](https://bazelbuild.github.io/rules_rust/defs.html).

This means that we ideally need `uniffi-bindgen` to be able to do it's work using only a compiled `.so` file and possibly the original Rust sources.

mhammond:
Thanks for the contribution!

Library mode is mainly needed when there are multiple UniFFI'd crates in a single library and we use the cargo metadata to help resolve some of the types between these crates. We certainly don't need all the metadata cargo gives us, but we do need what we actually use. In general I think we'd be perfectly fine with, say, a feature flag which allowed you to avoid even trying to use cargo, but we'd certainly insist all our existing use-cases continue to work by default.

I only had a fairly quick look, so please correct me where I'm wrong.

This patch seems to assume there's only actually one such crate and simply removes most of the stuff necessary when there is more than 1. If you really only have 1 UniFFI crate then we can probably come up with something, but I doubt that thing will be able to be just removing the code that allows more than 1 crate. If you do actually have more than one crate (or ever see yourselves in that situation), then I guess we'll need some way of expressing that metadata instead of relying on cargo, but in that scenario I don't really understand how this patch is on a path to actually doing that.

(The good news for us is that the stuff we need from cargo metadata actually has great tests, in particular the `external_types` fixture. If you were planning on getting our CI green before having a serious discussion about this and I just jumped in a little early, then please let us know and we can wait while you work out a plan that works for us both)

mhammond:
> I think the first step is identifying what this is. Based on the changes, it seems like package name / crate root is all we use. Is that right?

I think that's correct, yeah (but I was reluctant to quantify what we use under the assumption I'd forget something :). And I suspect the crate root is only used to find the full path to the UDL files - so there might be an opportunity to rethink that (eg, in this kind of world then you must also supply a map of UDL names to paths, or simply error out if UDL is used in this context?)

> maybe users count point to a TOML file with all the info.

Yeah - you could imagine a refactor which has this "dependency" code just work with a `toml::Value` - the default code could create an in-memory toml representation from cargo metadata, but environments using things other than Cargo could create a little utility script that creates an *actual* TOML file that's then provided to UniFFI, and the most extreme use-case of all would be to just hand-edit a TOML.

emarteca:
Thanks for the feedback, and the added context about when library mode is used! That makes sense. 

1) I'll make the cargo removal a flag that defaults to false, to maintain your current functionality, and make sure that the current tests all pass.

2) My understanding is also that the only information we need from the cargo metadata is the crate root and the package name. Passing in the package name directly as an argument will skip the existence check that's now in `find_package_by_crate_name`, but I think that is fine for our optional feature. 

The TOML input option for multiple crates makes sense to me -- what do you think @mgeisler ?

mhammond:
> My understanding is also that the only information we need from the cargo metadata is the crate root and the package name. 

Note that IIRC, we need this for *every* crate/package in the final library which uses UniFFI features.

The term "library mode" is a bit unfortunate because it conflates two things:
* Use the library to find all uniffi symbols created by procmacros.
* Use the library to identify all the crates/packages involved so things work when types are shared between them.

It almost sounds like you care only about the first and not the second - a solution which only cared about that might look different from one that involved both - eg, solving just that first probably *could* work with just command-line args and some tweaks, whereas a general solution for both would struggle via only the cmd-line and is where toml makes more sense.

mhammond:
> The term "library mode" is a bit unfortunate because it conflates two things:
>     * Use the library to find all uniffi symbols created by procmacros.

That's not actually correct - the cmdline already [does this](https://github.com/mozilla/uniffi-rs/blob/cd38ccea8236df7d93aff336c325a3a8e524af5d/uniffi_bindgen/src/lib.rs#L203) without involving "library mode" at all - so I'm guessing you do need the "multi crate" scenario.

emarteca:
The TOML file specifying a list of `(crate root, package name)` works for us -- I'll push an implementation of this tomorrow. Thank you!

mjpauly:
Oh wow this is really cool. UniFFI feels like the dream for integrating cross-platform Rust code into apps. Having it work with other polyglot build tools would be awesome.

I made an example [UniFFI+Bazel repo](https://github.com/mjpauly/uniffi-bazel-example/tree/main) to test things with, based on `examples/arithmetic-procmacro`. The [patches](https://github.com/mjpauly/uniffi-bazel-example/tree/main/patches) to UniFFI are just enough to get this working without Cargo or UDL. It even works for exporting UniFFI'd types from a dependent crate. ([subcrate](https://github.com/mjpauly/uniffi-bazel-example/tree/main/arithmetic/subcrate), [export](https://github.com/mjpauly/uniffi-bazel-example/blob/main/arithmetic/lib.rs#L43-L50), [use](https://github.com/mjpauly/uniffi-bazel-example/blob/main/test_arithmetic.py#L19))

It seems that procmacros can't do everything UDL can, but a simple no-UDL/no-Cargo mode could be a decent bare-bones starting point.

mhammond:
That's really cool!

> I made an example [UniFFI+Bazel repo](https://github.com/mjpauly/uniffi-bazel-example/tree/main) to test things with, based on examples/arithmetic-procmacro

If you are keen on taking this further, try adding a few different crates to your example - eg next to "arithmetic" add a few crates which depend on each other, ending up in some final library. That adds complexities which I'd love your feedback on :)

> It seems that procmacros can't do everything UDL can

It's very close. macros are far more convenient.

mhammond:
Hopefully obsoleted by #2175

Issue #2134, opened 2024-05-29
#[doc = ...] fails to parse with uniffi attribute macros


notmgsk:
This does not work
```
#[doc = std::concat!("A", "has a", "B")]
#[derive(uniffi::Enum)]
pub enum A {
    B,
}
```

but this does (without the derive)
```
#[doc = std::concat!("C", "has a", "D")]
pub enum C {
    D,
}
```

BatmanAoD:
This works in 0.26, but is broken in 0.27.

badboy:
Note: "does not work" is usually not very helpful. If you can it's always helpful to include the full error message you get.

badboy:
The test I add in #2154 also fails on 0.26. 

Doc string parsing first appeared in 0.26 here:

https://github.com/mozilla/uniffi-rs/blob/cd38ccea8236df7d93aff336c325a3a8e524af5d/uniffi_macros/src/util.rs#L295-L301

So yeah, that never worked since we added that.

PR #2154, opened 2024-06-17
Regression test for doc annotations parsing failure

badboy:
Regression test for #2134 -- let's see if we can also fix that?


mhammond:
Thanks - but I'm not sure your intent - I'm happy to approve it, but we don't want it landed, right?

badboy:
nope! I actually wanted this in draft mode, oops. figured it would be good to have a regression test for the reported bug.

PR #2166, opened 2024-06-21
Make `RustBuffer::from_raw_parts` comment more precise

mgeisler:
The method is used from outside of this module, so the comment seems unnecessary: the method should either be private or it should be public.

### DIFF
--- a/uniffi_core/src/ffi/rustbuffer.rs
+++ b/uniffi_core/src/ffi/rustbuffer.rs
@@ -81,7 +81,7 @@ impl RustBuffer {
     /// # Safety
     ///
     /// You must ensure that the raw parts uphold the documented invariants of this class.
-    pub unsafe fn from_raw_parts(data: *mut u8, len: u64, capacity: u64) -> Self {
+    pub(crate) unsafe fn from_raw_parts(data: *mut u8, len: u64, capacity: u64) -> Self {
         Self {
             capacity,
             len,
--- a/uniffi_core/src/ffi/rustbuffer.rs
+++ b/uniffi_core/src/ffi/rustbuffer.rs
@@ -78,9 +78,6 @@ impl RustBuffer {
 
     /// Creates a `RustBuffer` from its constituent fields.
     ///
-    /// This is intended mainly as an internal convenience function and should not
-    /// be used outside of this module.
-    ///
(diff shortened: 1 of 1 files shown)


mhammond:
I wonder if we can make it private? I'm on a bit of a mission to reduce the api surface as much as possible before we move towards a 1.0, just so we don't accidentally force ourselves to have a change be "breaking" when it isn't in practice (which IIUC, mislabeling a private method as public risks)

mgeisler:
Yes, making it private would be great! :smile: 

Cc @badboy, a small note from my import review.

badboy:
> I wonder if we can make it private? I'm on a bit of a mission to reduce the api surface as much as possible before we move towards a 1.0, just so we don't accidentally force ourselves to have a change be "breaking" when it isn't in practice (which IIUC, mislabeling a private method as public risks)

```
error[E0624]: associated function `from_raw_parts` is private
  --> uniffi_core/src/ffi/ffidefault.rs:56:24
   |
56 |         unsafe { Self::from_raw_parts(std::ptr::null_mut(), 0, 0) }
   |                        ^^^^^^^^^^^^^^ private associated function
   |
  ::: uniffi_core/src/ffi/rustbuffer.rs:87:5
   |
87 |     unsafe fn from_raw_parts(data: *mut u8, len: u64, capacity: u64) -> Self {
   |     ------------------------------------------------------------------------ private associated function defined here

error[E0624]: associated function `from_raw_parts` is private
   --> uniffi_core/src/ffi/ffiserialize.rs:213:37
    |
213 |         unsafe { crate::RustBuffer::from_raw_parts(data, len, capacity) }
    |                                     ^^^^^^^^^^^^^^ private associated function
    |
   ::: uniffi_core/src/ffi/rustbuffer.rs:87:5
    |
87  |     unsafe fn from_raw_parts(data: *mut u8, len: u64, capacity: u64) -> Self {
    |     ------------------------------------------------------------------------ private associated function defined here
```

well, I guess we didn't follow our own advice :D

badboy:
It can at the very leat be `pub(crate)`

mgeisler:
> It can at the very leat be `pub(crate)`

Done!

badboy (approved):

PR #2167, opened 2024-06-21
Make panic section more precise for `RustBuffer::len`

mgeisler:
The panic condition seems to be a mismatch between the foregin code and the size of `usize` on the platform.


mgeisler:
Cc @badboy, a small note from my import review.

mhammond (approved):

PR #2170, opened 2024-06-24
Clarify pointer alignment for `Handle`

### LINKED ISSUE #2165: Rust pointers can have both even and odd addresses
The documentation seems to imply that pointers always have an even address (aligned to 2 bytes):
 
https://github.com/mozilla/uniffi-rs/blob/cd38ccea8236df7d93aff336c325a3a8e524af5d/uniffi_core/src/ffi/handle.rs#L14-L15

However, this is [not necessarily true](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=81edc92b4aba101a52601805ed7a1e94):

```rust
    let a: [u8; 3] = [10, 20, 30];
    dbg!(&a);
    dbg!(&a[0] as *const u8);
    dbg!(&a[1] as *const u8);
    dbg!(&a[2] as *const u8);
```

gives me

```
[src/main.rs:4:5] &a[0] as *const u8 = 0x00007ffd556a28f5
[src/main.rs:5:5] &a[1] as *const u8 = 0x00007ffd556a28f6
[src/main.rs:6:5] &a[2] as *const u8 = 0x00007ffd556a28f7
```

which has both odd and even pointers. So if I run this safe code, I will have violated this invariant:

```
Handle::from_pointer(&a[2])
```

When I tried the same with `Box::into_raw`, I did indeed only end up with pointers to even addresses. I don't know if I was just lucky or if there is some underlying reason why the comment is true specifically for leaked `Box` values?

Cc @badboy.

mgeisler:
Fixes #2165.

### DIFF
--- a/uniffi_core/src/ffi/handle.rs
+++ b/uniffi_core/src/ffi/handle.rs
@@ -12,7 +12,7 @@
 /// For all currently supported architectures and hopefully any ones we add in the future:
 /// * 0 is an invalid value.
 /// * The lowest bit will always be set for foreign handles and never set for Rust ones (since the
-///   leaked pointer will be aligned).
+///   leaked pointer will be aligned to `size_of::<Arc<T>>()` == `size_of::<*const T>()`).
 ///
 /// Rust handles are mainly managed is through the [crate::HandleAlloc] trait.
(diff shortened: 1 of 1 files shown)


bendk (approved):
Looks good to me, let's merge this one.

PR #2172, opened 2024-06-28
Use Data(bytesNoCopy:size:deallocator:) for conversion between RustBuffer and Data

Akazm:
This avoids unnecessary copies between FFI and Swift.

Tests are apparently entirely unaffected (i.e. no changes to their results) by this change. `deallocator` is set to `.none` since the `FfiConverterRustBuffer` extension function `lift(_:)` ensures deallocation on it's own.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/RustBufferTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/RustBufferTemplate.swift
@@ -37,9 +37,11 @@ fileprivate extension ForeignBytes {
 
 fileprivate extension Data {
     init(rustBuffer: RustBuffer) {
-        // TODO: This copies the buffer. Can we read directly from a
-        // Rust buffer?
-        self.init(bytes: rustBuffer.data!, count: Int(rustBuffer.len))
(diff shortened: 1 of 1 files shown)


mhammond (commented):
This seems fine to me, any other thoughts?

badboy:
Looks like we use `Data` in two places:

https://github.com/mozilla/uniffi-rs/blob/bfa98ed377629de931587cf412692f4388ca9471/uniffi_bindgen/src/bindings/swift/templates/ObjectTemplate.swift#L174

https://github.com/mozilla/uniffi-rs/blob/bfa98ed377629de931587cf412692f4388ca9471/uniffi_bindgen/src/bindings/swift/templates/RustBufferTemplate.swift#L173

Both times `Data` is created adhoc and the reader passed on as a reference.
The Rust buffer is created sometime before those functions are called and is then deallocated shortly after.
It therefore should be safe to have `Data` not copy out the data, but just reference it.

badboy (approved):

Akazm:
Oh, didn't notice the other invocation! Should I open another PR for this?

PR #2173, opened 2024-06-29
`RustCallStatus::error_buf` is now `ManuallyDrop` and not `MaybeUninit`.

### LINKED ISSUE #2168: `RustCallStatus::error_buf` should be `ManuallyDrop`, not `MaybeUninit`?
In this code

https://github.com/mozilla/uniffi-rs/blob/cd38ccea8236df7d93aff336c325a3a8e524af5d/uniffi_core/src/ffi/rustcalls.rs#L46-L55

it seems that `ManuallyDrop` should be used instead of `MaybeUninit`?

This line could be made more precise:

> `MaybeUninit` requires unsafe code, since we are preventing rust from dropping the value.

Not drop values is not _unsafe_ — leaking memory is considered safe behavior in Rust, which is why things like `Box::leak` is a safe function.

Cc @badboy, another thing that came up in my import review.

mhammond:
`error_buf` is owned by the foreign side. Because we assume it has been initialized we use `ManuallyDrop` instead of `MaybeUninit`, which clarifies ownership and avoids unsafe code.

Fixes #2168

### DIFF
--- a/uniffi_core/src/ffi/rustcalls.rs
+++ b/uniffi_core/src/ffi/rustcalls.rs
@@ -12,7 +12,7 @@
 //!      exception
 
 use crate::{FfiDefault, Lower, RustBuffer, UniFfiTag};
-use std::mem::MaybeUninit;
+use std::mem::ManuallyDrop;
 use std::panic;
 
 /// Represents the success/error of a rust call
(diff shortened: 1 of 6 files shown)


badboy (approved):

uniffi_core/src/ffi/rustcalls.rs
```diff
@@ -42,52 +42,42 @@ use std::panic;
 #[repr(C)]
 pub struct RustCallStatus {
     pub code: RustCallStatusCode,
-    // code is signed because unsigned types are experimental in Kotlin
```
bendk, line 45: Not completely related, but how about a comment saying that code is initialized by the foreign side to RustCallStatus::Success?

uniffi_core/src/ffi/rustcalls.rs
```diff
@@ -42,52 +42,42 @@ use std::panic;
...
-    //     avoiding the drop is a small optimization.
-    //   - If consumers pass in invalid data, then we should avoid trying to drop it.  In
-    //     particular, we don't want to try to free any data the consumer has allocated.
-    //
-    // `MaybeUninit` requires unsafe code, since we are preventing rust from dropping the value.
-    // To use this safely we need to make sure that no code paths set this twice, since that will
-    // leak the first `RustBuffer`.
+    pub error_buf: ManuallyDrop<RustBuffer>,
+    // error_buf is owned by the foreign side.
+    // - Whatever we are passed, we must never free.
```
bendk, line 47: I think the foreign side always passes in an empty buffer, this is important since otherwise we would leak it whenever we write to the field.

uniffi_core/src/ffi/rustcalls.rs
```diff
@@ -171,11 +161,7 @@ where
         // Callback returned an Err.
         Ok(Err(buf)) => {
             out_status.code = RustCallStatusCode::Error;
-            unsafe {
-                // Unsafe because we're setting the `MaybeUninit` value, see above for safety
-                // invariants.
-                out_status.error_buf.as_mut_ptr().write(buf);
-            }
+            *out_status.error_buf = buf;
```
bendk, line 164: If we set this field multiple times, then we'll leak any previous values.  Can you add a comment saying that we won't leak because a) it's only set in this function and b) there's no branch where it gets set twice.

bendk (approved):
This all looks right to me, I just had some docstring suggestions.

uniffi_core/src/ffi/rustcalls.rs
mhammond: There's already a comment few lines up that says exactly this, so I wont bother repeating it.

mgeisler:
Thanks a lot, @mhammond! Cc @maurer.

PR #2175, opened 2024-07-03
Add a cargo_metadata feature so using cargo_metadata can be avoided i…

mhammond:
…n some cases

This is a draft early cut to help with #2151 - this is a first-step but I don't think what I have here is ideal - happy to take any and all feedback.

cc @mgeisler @emarteca

### DIFF
--- a/uniffi_bindgen/src/library_mode.rs
+++ b/uniffi_bindgen/src/library_mode.rs
@@ -20,8 +20,7 @@ use crate::{
     GenerationSettings, Result,
 };
 use anyhow::{bail, Context};
-use camino::Utf8Path;
-use cargo_metadata::{MetadataCommand, Package};
+use camino::{Utf8Path, Utf8PathBuf};
 use std::{collections::HashMap, fs};
 use uniffi_meta::{
     create_metadata_groups, fixup_external_type, group_metadata, Metadata, MetadataGroup,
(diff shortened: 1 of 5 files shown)


uniffi_bindgen/src/library_mode.rs
```diff
@@ -138,39 +142,73 @@ fn find_components(
...
+                    )
+                })
+            })
+            .collect();
+        Self { libs }
+    }
+}
+
+// We should think this struct through a little.
+// * package_name: used only for tests for swift - should we
```
mhammond, line 193: See https://github.com/mozilla/uniffi-rs/commit/19b639d4b40af963ff30e9caf9b5768d092b0730 - getting rid of `package_name` is actually quite easy and I think it's an improvement - it does leave this struct with 1 element though!

mhammond:
I went all the way here, I think it's much clearer what's going on and allows for things to work with no metadata at all in no-udl cases.

mhammond:
This cleans up enough that I'm putting it up for review now - we can work out how to define this map externally when not using cargo_metadata later, but this is a good first step regardless of what we do there.

uniffi_bindgen/src/library_mode.rs
```diff
@@ -42,24 +41,25 @@ pub fn generate_bindings<T: BindingGenerator + ?Sized>(
...
+    // Or can we also read it from `config_file_override` - what are the semantics of that?
+    #[cfg(feature = "cargo_metadata")]
+    let path_map = {
+        let cargo_metadata = cargo_metadata::MetadataCommand::new()
+            .exec()
+            .context("error running cargo metadata")?;
+        CratePathMap::from(cargo_metadata)
+    };
+    #[cfg(not(feature = "cargo_metadata"))]
+    let path_map = Default::default();
```
bendk, line 54: How about extracting this to a `CratePathMap` method and defining that differently based on the feature flag?  It's hard to follow code where individual statements are behind feature flags.

bendk (approved):
Looks great to me.  I like how it makes it much more clear exactly how we're using cargo metadata.

uniffi_bindgen/src/library_mode.rs
```diff
@@ -138,39 +124,68 @@ fn find_components(
...
+
+// all this to get a few paths!?
+#[cfg(feature = "cargo_metadata")]
+impl From<cargo_metadata::Metadata> for CratePathMap {
+    fn from(metadata: cargo_metadata::Metadata) -> Self {
+        let paths: HashMap<String, Utf8PathBuf> = metadata
+            .packages
+            .iter()
+            .flat_map(|p| {
+                p.targets.iter().filter(|t| t.is_lib()).filter_map(|t| {
```
cschramm, line 175: This seemingly harmless filter does not seem correct. I just investigated why my code generation broke with uniffi 0.28.1 and the answer is that the crate just does not get added to the map due to this anymore.

It has `crate-type = ["staticlib", "cdylib"]` and the `cargo_metadata` crate might not do what you expect (as its documentation suggests differently): `kind` and `crate_types` both are `["staticlib", "cdylib"]` and `is_lib` only returns true if `"lib"` is contained there. From what the `cargo_metadata` interface suggests, the target is of no type (none of the provided `is_*` methods returns true and `kind` does not have any of the values mentioned in its documentation).

PR #2176, opened 2024-07-03
feat: add option through .udl to rename objects and functions i

gruberb:
* [DISCO-2879](https://mozilla-hub.atlassian.net/jira/software/c/projects/DISCO/boards/301?selectedIssue=DISCO-2879)

* Closes https://github.com/mozilla/uniffi-rs/issues/1426

---

We want to allow renaming of functions and variables. This PR adds the ability to add the following to a `uniffi.toml` file:

```toml
[bindings.kotlin.rename]
OtherError = "AnotherError"

[bindings.python.rename]
OtherError = "AnotherError"

[bindings.ruby.rename]
UnknownError = "UnspecificError"

[bindings.swift.rename]
OtherError = "AnotherError"
```

Which will rename whatever `enum`, `function` or `struct`, `constant` or other `parameter` matches the name, to the name in the double quotes to the right. 

For this, we extend the `filter` functions in the templates for each language, and pass down the `config` variable. From there, we read the additional `rename` parameter, and check for a match. 

#### Example: Python `class_name`:

```python
pub fn class_name(nm: &str, config: &Config) -> Result<String, askama::Error> {
    if let Some(overwrite_class_name) = config.rename.get(nm) {
        return Ok(PythonCodeOracle.class_name(overwrite_class_name));
    }
    Ok(PythonCodeOracle.class_name(nm))
}
```

And this will be picked up in the template:
```python
{%- let variant_type_name = variant.name()|class_name(config) -%}
```

### ToDo

- [x] Python
- [x] Swift
- [x] Kotlin
- [x] Ruby

### Testing
- [x] Python
- [x] Swift
- [x] Kotlin
- [x] Ruby

## Open

I have to figure out if that's currently possible for Ruby. Would need a second pair of eyes, as this is my first PR in this repository.


bendk (changes requested):
Code looks great, thanks for doing this!

I have some reservations about the changes to `var_name`, mostly because that filter applies to a very wide range of things and should probably be split up.  The other issue is that I think member fields should be specified as `[type_name].[field]` rather than just `[field]`.

Another thing that could be good to rename would be methods, which I think should have a similar syntax.

I made https://github.com/mozilla/uniffi-rs/issues/2184 after seeing how many hoops you needed to go through in order to use the `Config` to rename things.  I also think that if you went with my `[type_name].[field]` suggestion it might be even more work.  Maybe if we implemented that one first, this would all become easier.

What do you think of only targeting type names and function names as a first pass?  I think that would address the majority of the use-cases, since the main issue is name collisions with top-level items.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -569,31 +584,54 @@ mod filters {
...
+        }
         Ok(KotlinCodeOracle.fn_name(nm))
     }
 
     /// Get the idiomatic Kotlin rendering of a variable name.
-    pub fn var_name(nm: &str) -> Result<String, askama::Error> {
+    pub fn var_name(nm: &str, config: &Config) -> Result<String, askama::Error> {
+        if let Some(overwrite_var_name) = config.rename.get(nm) {
+            return Ok(KotlinCodeOracle.var_name(overwrite_var_name));
+        }
```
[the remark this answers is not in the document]
gruberb: Oh, through all the debugging I totally forgot my note on changing this to `[type_name].[field_name]`. Thanks for reminding me! 

So we put a lid on this for now and focus on a cleaner structure first (based on your opened issue) or should I still adjust to `[type_name].[field_name]`? 

It was very confusing for me what `var_name` changes. So yes, totally agree.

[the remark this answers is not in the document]
gruberb: I will push a change to adopt `[type_name].[field_name]`. Just to double check since I don't have a gut feeling yet about how people use this feature: We expect `type_name` to be the `type` of the Rust representation, or the associated type of the target language?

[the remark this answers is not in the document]
bendk: I think the Rust representation is best.  I think we've used that in other cases when we want to identify things.  Also, it makes cut+paste easier when you want to rename something in multiple languages.

gruberb:
Updated the PR, needs some review again. I am not quite sure how we directly map Rust types to the foreign type bindings. Like: https://github.com/mozilla/uniffi-rs/pull/2176/files#diff-522f0ef157bc0a32453bafd790a9058a35cbb8a721df58a9b51b1ec6b74db5caR543-R554

I think both an `enum` and a `struct` can end up as `type_name` for example in Kotlin. So I check both. I think this will be tackled with this issue (https://github.com/mozilla/uniffi-rs/issues/2184) in the future.

docs/manual/src/bindings.md
```diff
@@ -13,6 +13,13 @@ Each of the bindings reads a file `uniffi.toml` in the root of a crate which sup
 various options which influence how the bindings are generated. Default options will be used
 if this file is missing.
 
+You can rename functions and structures via `[bindings.{language}.rename]` in the `uniffi.toml`. Options include: `enums`, `structs` and `functions`. For example:
+
+```toml
+[bindings.python.rename]
+enums.OtherError = "AnotherError"
+```
```
bendk, line 21: I think I prefer the previous system where all the names were in one map, rather than splitting up by type.  There's no chance of name collision -- if there were, they're also be a collision in the Rust code.  Also, I think there could be some confusion since UniFFI has the concept of interfaces, which can be structs or enums.

uniffi_bindgen/src/bindings/python/templates/EnumTemplate.py
```diff
@@ -136,16 +136,16 @@ def check_lower(value):
     def write(value, buf):
         {%- for variant in e.variants() %}
         {%- if e.is_flat() %}
-        if value == {{ type_name }}.{{ variant.name()|enum_variant_py }}:
+        if value == {{ type_name }}.{{ variant.name()|enum_variant_py(config) }}:
             buf.write_i32({{ loop.index }})
         {%- else %}
-        if value.is_{{ variant.name()|var_name }}():
+        if value.is_{{ variant.name()|var_name}}():
```
bendk, line 142: Nano-nit: keep the space before the `}}` here and in some other locations.  That's generally how we format things and also will keep the commit size down.

bendk (changes requested):
This is looking really good, I just had one comment on the config syntax.  I don't have strong feelings though, please push back on that if I'm missing some of your reasoning for the change.

docs/manual/src/bindings.md
gruberb: I agree on both sides... For now, it is too confusing because I think the generating of the `FFI` template uses structs and enums as type names. I liked the structured approach, it makes it very specific what will be changed. 

I think for now it is ok to remove it again. After we implmeneted your new issue with the more structured approach, we might want to revisit this imo.

uniffi_bindgen/src/bindings/python/templates/EnumTemplate.py
gruberb: Ups, yeah, oversight through Search+Replace.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -523,7 +533,12 @@ mod filters {
     pub(super) fn type_name(
         as_ct: &impl AsCodeType,
         ci: &ComponentInterface,
+        config: &Config,
     ) -> Result<String, askama::Error> {
+        if let Some(overwrite_type_name) = config.rename.get(&as_ct.as_codetype().type_label(ci)) {
+            return Ok(overwrite_type_name.to_owned());
+        }
```
bendk, line 540: If you moved the renaming to happen in KotlinCodeOracle then you wouldn't need to do anything special here, since `type_label` ends up calling `KotlinCodeOracle::class_name`.

This code makes me realize you'd need to change `type_label` to input a `Config`, but I think that should be fine.

bendk (changes requested):
I'm currently thinking that the best place to do the renames is in the code oracles themselves, but I'm not totally sure.  What do you think about that?

mhammond:
Is there an actual use-case for this now? If not, it is starting look like experimenting with #2184 before landing this might make our future-selves less unhappy with our present-selves :) Another great thing about that is we could experiment with just one binding to understand how viable it is.

gruberb:
Closing this, since this PR makes this work a lot easier:
https://github.com/mozilla/uniffi-rs/pull/2191

PR #2177, opened 2024-07-04
Use fully-qualified names to the stdlib's `core` library

### LINKED ISSUE #2127: Updating from 0.25.3 to 0.26.x or 0.27.x fails for interface
Env:

nightly-aarch64-apple-darwin - Up to date : 1.80.0-nightly (867900499 2024-05-23)
uniffi 0.27.2

The following works

```
/// Example
namespace example {
  /// Hello Rust
  string hello_world();
};

/// Example Enum
enum ExampleEnum {
  "v1",
  "v2",
};


/// Glucose
dictionary Glucose {
    f32 raw;
    timestamp timestamp;
    timestamp local_timestamp;
};
```

Adding the following fails:
```
// lib.rs
pub struct TodoList;

impl TodoList {
    fn new() -> Self {
        TodoList {
        }
    }
}

// udl file
/// TodoList
interface TodoList {
    constructor();
};
```

The error is:
```
error[E0433]: failed to resolve: could not find `marker` in `core`
  --> <hidden>/cross_platform/target/debug/build/example-904b99f23226f7e3/out/example.uniffi.rs:69:1
   |
69 | #[::uniffi::derive_object_for_udl]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ could not find `marker` in `core`

   |
   = note: this error originates in the attribute macro `::uniffi::derive_object_for_udl` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this module
  --> src/lib.rs:1:1
   |
1  + use std::marker;
   |
```

This is the content of example.uniffi.rs:69:1
```
#[::uniffi::derive_object_for_udl]
struct r#TodoList { }
#[::uniffi::export_for_udl]
impl r#TodoList {
    #[uniffi::constructor]
    pub fn r#new(
    ) -> std::sync::Arc<r#TodoList>
    {
        unreachable!()
    }
}
```

mhammond:
Fixes #2127

### DIFF
--- a/uniffi_core/src/ffi/ffiserialize.rs
+++ b/uniffi_core/src/ffi/ffiserialize.rs
@@ -81,7 +81,7 @@ pub trait FfiSerialize: Sized {
     fn write(buf: &mut &mut [FfiBufferElement], value: Self) {
         Self::put(buf, value);
         // Lifetime dance taken from `bytes::BufMut`
-        let (_, new_buf) = core::mem::take(buf).split_at_mut(Self::SIZE);
+        let (_, new_buf) = ::core::mem::take(buf).split_at_mut(Self::SIZE);
         *buf = new_buf;
     }
 }
--- a/uniffi_meta/src/types.rs
+++ b/uniffi_meta/src/types.rs
@@ -146,7 +146,7 @@ impl Type {
 }
 
 // A trait so various things can turn into a type.
-pub trait AsType: core::fmt::Debug {
+pub trait AsType: ::core::fmt::Debug {
     fn as_type(&self) -> Type;
 }
 


jplatte (approved):

jplatte:
By the way, you could add a fixture that uses [`#![no_implicit_prelude]`](https://doc.rust-lang.org/reference/names/preludes.html#the-no_implicit_prelude-attribute) to ensure that the generated code does not rely on the standard library prelude in any way.

badboy (approved):

mhammond:
> By the way, you could add a fixture that uses [`#![no_implicit_prelude]`](https://doc.rust-lang.org/reference/names/preludes.html#the-no_implicit_prelude-attribute) to ensure that the generated code does not rely on the standard library prelude in any way.

Thanks - I had a quick look at this but struggled to see how it would work in practice.

I see how that would help with code generated by macros, but in these 2 specific cases they were just normal items in `uniffi_core` and `uniffi_meta`. Adding that prelude to these crates means things like `Result`, `Into` etc all fail - they require the full path - so I'm not sure that's practical.

But even if we decided to do this just for macros, the fixture would need to be fairly comprehensive to ensure all macros are exercised in all ways - so I was thinking maybe we could add it to the `coveralls` and `proc_macros` fixtures, but then those fixtures would all need significant modification in the same way (ie, every `Result`, `None` etc would be touched.)

Is there something I'm missing here?

PR #2178, opened 2024-07-05
feat: implement module path config for python

### LINKED ISSUE #1182: Fixup some exteranl type issues
While rebasing #1167, I noticed some issues:

  - Python doesn't support the `external_packages` config value.
  - The Python`ext-types` tests only pass if you  build the crates for the external types before running the tests.  Kotlin works, but only because it lazily loads the library.  If we ever needed to call a function from the UDL I think it would fail too.  The issue is that we are building the `.so` files in the `target/debug/deps` directory in this case, but we're expecting them `target/debug/` directory.

dignifiedquire:
Closes #1182

### DIFF
--- a/uniffi_bindgen/src/bindings/python/mod.rs
+++ b/uniffi_bindgen/src/bindings/python/mod.rs
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-use std::process::Command;
+use std::{collections::HashMap, process::Command};
 
 use anyhow::Result;
 use fs_err as fs;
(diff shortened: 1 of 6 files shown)


mhammond:
Thanks! But can we please get a test for this? Maybe specifying it for, eg, fixtures/ext-types/sub-lib?

dignifiedquire:
> Thanks! But can we please get a test for this? Maybe specifying it for, eg, fixtures/ext-types/sub-lib?

sure, will add to that fixture

uniffi_bindgen/src/bindings/python/mod.rs
```diff
@@ -33,13 +33,33 @@ impl crate::BindingGenerator for PythonBindingGenerator {
...
+                .iter()
+                .map(|c| (c.ci.crate_name().to_string(), c.config.module_name())),
+        );
+        for c in components {
+            for (ext_crate, ext_package) in &packages {
+                if ext_crate != c.ci.crate_name()
+                    && !c.config.external_packages.contains_key(ext_crate)
+                {
+                    c.config
+                        .external_packages
```
mhammond, line 58: We don't have good tests for this, but I think this will break [this external type](https://github.com/mozilla/uniffi-rs/blob/c5da61baf98d421f2da521b1ab085f576695178d/uniffi_bindgen/src/bindings/python/gen_python/mod.rs#L139-L148) support when the module name is overridden.

For my interest, what's your use-case for using a different module name than the namespace, and are you using multiple crates?

dignifiedquire: > For my interest, what's your use-case for using a different module name than the namespace, and are you using multiple crates?

No just a single crate, but the issue that I was running into is that the name of my ffi crate can't be the same as the dependency that I am creating the bindings for, which I ran into when starting to use proc-macros, as uniffi is confused about multiple dylibs of the same name.

mhammond: Can you just change the namespace name? It doesn't have to be the crate name.

mhammond:
Feel free to reopen this if you'd like to pursue it!

dignifiedquire:
sorry, yeah I solved this in a different way and forgot about the PR

PR #2180, opened 2024-07-05
Create fixture proc-macro-no-implicit-prelude

jplatte:
This is what I was talking about in #2177. I think you weren't missing anything @mhammond, it's a bunch of boilerplate so I put it in a new fixture instead of `proc-macro`. It's a copy of that one though.

Note that there is still the language prelude that is _not_ disabled by this attribute (which contains primitive types such as `u8` and I think `str` as well, plus lots of macros like `#[align]`, `#[allow]`, `#[automatically_derived]`, ...). Because some of these things can nevertheless be shadowed, this test doesn't catch all potential name confusions within macro code. It it pretty good at it though, as you can see in the first commit 😄 (the few `::std::primitive` insertions were drive-by improvements and not steered by error messages)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1785,6 +1785,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "uniffi-fixture-proc-macro-no-implicit-prelude"
+version = "0.22.0"
+dependencies = [
+ "uniffi",
+]
+
+[[package]]
 name = "uniffi-fixture-regression-callbacks-omit-labels"
 version = "0.22.0"
 dependencies = [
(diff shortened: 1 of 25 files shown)


jplatte:
Ah, I think rust-analyzer wasn't showing me errors from UDL scaffolding generation, or something?

Happy for anybody else to take over this PR and finish it 😅 
I don't really feel like putting in more time into it, today at least.

If you want I can also post the first commit as a separate PR, that should pass CI as-is.

mhammond:
Thanks @jplatte, that's useful and helpful! I opened https://github.com/mozilla/uniffi-rs/pull/2181 because the failing tests all seem quite shallow - I'm happy for either you to merge that into yours, or for me to squash them up and land them - but you did most of the work :)

mhammond (approved):

PR #2182, opened 2024-07-08
Kotlin: use `Enum.entries` instead of `Enum.values()`

jmartinesp:
Since Kotlin `1.9`, the `.entries` property is generated at build time for enum classes. Unlike `Enum.values()`, this property doesn't instantiate a new array with every enum case every time it's used, instead a single List instance is reused.

This should improve performance a tiny bit for Kotlin bindings. However, note this will only be supported since Kotlin 1.9, I don't know if there is a min version the project wants to support.

### DIFF
--- a/docs/manual/src/kotlin/configuration.md
+++ b/docs/manual/src/kotlin/configuration.md
@@ -4,16 +4,16 @@ The generated Kotlin modules can be configured using a `uniffi.toml` configurati
 
 ## Available options
 
-| Configuration name           | Default  | Description |
-|------------------------------| -------  |------------ |
-| `package_name`               |  `uniffi` | The Kotlin package name - ie, the value used in the `package` statement at the top of generated files. |
+| Configuration name           | Default                  | Description |
+|------------------------------|--------------------------|------------ |
+| `package_name`               | `uniffi`                 | The Kotlin package name - ie, the value used in the `package` statement at the top of generated files. |
 | `cdylib_name`                | `uniffi_{namespace}`[^1] | The name of the compiled Rust library containing the FFI implementation (not needed when using `generate --library`). |
-| `generate_immutable_records` | `false` | Whether to generate records with immutable fields (`val` instead of `var`). |
-| `custom_types`               | | A map which controls how custom types are exposed to Kotlin. See the [custom types section of the manual](../udl/custom_types.md#custom-types-in-the-bindings-code)|
-| `external_packages`          | | A map of packages to be used for the specified external crates. The key is the Rust crate name, the value is the Kotlin package which will be used referring to types in that crate. See the [external types section of the manual](../udl/ext_types_external.md#kotlin)
-| `android`                    | `false` | Used to toggle on Android specific optimizations
-| `android_cleaner`            | `android` | Use the [`android.system.SystemCleaner`](https://developer.android.com/reference/android/system/SystemCleaner) instead of [`java.lang.ref.Cleaner`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/ref/Cleaner.html). Fallback in both instances is the one shipped with JNA.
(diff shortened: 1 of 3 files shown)


jmartinesp:
The CI seems to fail because it uses Kotlin 1.8.20 or lower 😅

jmartinesp:
I added an `use_enum_entries` config parameter to enable this change on Kotlin  >= `1.9.0` while keeping backwards compatibility with previous versions of the language.

docs/manual/src/kotlin/configuration.md
```diff
@@ -4,15 +4,16 @@ The generated Kotlin modules can be configured using a `uniffi.toml` configurati
...
+| `custom_types`               | | A map which controls how custom types are exposed to Kotlin. See the [custom types section of the manual](../udl/custom_types.md#custom-types-in-the-bindings-code)|
+| `external_packages`          | | A map of packages to be used for the specified external crates. The key is the Rust crate name, the value is the Kotlin package which will be used referring to types in that crate. See the [external types section of the manual](../udl/ext_types_external.md#kotlin)
+| `android`                    | `false` | Used to toggle on Android specific optimizations
+| `android_cleaner`            | `android` | Use the [`android.system.SystemCleaner`](https://developer.android.com/reference/android/system/SystemCleaner) instead of [`java.lang.ref.Cleaner`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/ref/Cleaner.html). Fallback in both instances is the one shipped with JNA.
+| `use_enum_entries`           | `false` | Uses `EnumClass.entries` instead of `EnumClass.values()` in the bindings, which is more efficient as it reuses the same list instance every time. Note this is only available since Kotlin 1.9.0
```
bendk, line 16: I like the idea of using a config value for this a lot.  My one concern is that we'll end up with tons of these entries that people will need to fiddle with.  What do you think of adding something like `kotlin_target_version`?  The `use_enum_entries` method can check if it's >= `1.9.0` and this same logic could be re-used for using other language features.

jplatte: A simple `>=` will only work if this is stored as a tuple though, not if it's stored as a string.

jmartinesp: That sounds like a sensible idea. Let me check if I can get something working real quick.

jmartinesp: https://github.com/mozilla/uniffi-rs/pull/2182/commits/486fa87acb6f7853bb58fa8ad85bbd72c97a9e49

bendk: This looks great, my one last concern is the use of `cargo_metadata` for the parsing.  We're working towards making that optional, so I'd rather we did the parsing ourselves.

However, this is getting really far away from your original concerns though.  Do you want to try to tackle this or should we merge it as-is and I can try to handle that one as a follow-up?

jmartinesp: I'd rather merge it as is, if possible. I didn't realise you were trying to get rid of cargo metadata. Maybe I should have used [semver](https://docs.rs/semver/latest/semver/) instead?

bendk: Sounds good, let's merge this and figure out how to remove it later.   I'm kind of thinking that we can just parse the versions by hand for now and avoid the extra dependency, but we can discuss in a separate PR.

bendk: Oops, I see there's still a clippy error.  I can merge once you fix that one.

jmartinesp: I hope https://github.com/mozilla/uniffi-rs/pull/2182/commits/1dfd0a5abe67a86e2bd64900b26ca3535d97938a will fix it 🤞 .

bendk (approved):
Looks like that worked, thanks for the contribution.

PR #2185, opened 2024-07-12
Only invoke cargo metadata for desired targets (fixes #2183)

### LINKED ISSUE #2183: Only invoke cargo metadata for desired targets
(Based on a similar issue: https://github.com/bbqsrc/cargo-ndk/issues/83)

Unlike most other cargo tools, cargo metadata counter-intuitively defaults to including all targets. This means that when running with `--library`, `uniffi` ends up processing (and thus downloading dependencies for) all targets, many of which are likely to be unsupported by the `cdylib` that the user pointed the CLI at. Specifically, the `MetadataCommand` invocation here is broader than it needs to be:
https://github.com/mozilla/uniffi-rs/blob/main/uniffi_bindgen/src/library_mode.rs#L45-L47

If the user passes a `--crate`, `uniffi` would ideally skip the `MetadataCommand` entirely, as the `--help` message implies it does. Without `--crate`, perhaps `uniffi` could either attempt to infer the target of the `cdylib` (maybe a bad idea) or accept a new parameter for `--target` that scopes down the command.

mhammond:


bendk (approved):
LGTM.

PR #2190, opened 2024-07-17
Don't panic for try_lift errors

bendk:
This avoids crashing for consumers that set panic=abort (Firefox).

- Updated the Err type for `rust_call`.  The closure can now return an internal error.  The only way to signal an internal error before was to panic.
- Reworked the `LowerReturn` trait and the future code to work with the new system.
- Added a type for argument lift errors, this makes the rustfuture signatures nicer.

### DIFF
--- a/uniffi_core/src/ffi/rustcalls.rs
+++ b/uniffi_core/src/ffi/rustcalls.rs
@@ -111,6 +111,29 @@ impl TryFrom<i8> for RustCallStatusCode {
     }
 }
 
+/// Error type for Rust scaffolding calls
+///
+/// This enum represents the fact that there are two ways for a scaffolding call to fail:
+/// - Expected errors (the Rust function returned an `Err` value).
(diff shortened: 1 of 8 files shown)


mhammond (approved):

pierre-wehbe:
@mhammond Would this allow flat_errors to be lifted by any chance?

`Can't lift flat errors`
`rustPanic("Can\'t lift flat errors")`

mhammond:
> @mhammond Would this allow flat_errors to be lifted by any chance?

I doubt it - that's largely impossible IIUC - consider that we allow a flat enum to have any type - eg
```
enum Foo {
  Variant(SomeOtherType),
}
```

Where `SomeOtherType` can be literally anything that works in Rust - it doesn't need to be a Uniffi type. This works because it's flat - we just lower the variant discriminator without `SomeOtherType` at all. This makes it almost impossible to lift - lifting implies somehow we are able to create `Variant(SomeOtherType)` without a value for `SomeOtherType`.

Issue #2184, opened 2024-07-11
Add a conversion pass before rendering the bindings templates


bendk:
I think it would simplify and improve the code to add pass [before constructing the template object](https://github.com/mozilla/uniffi-rs/blob/f4ca11f2ebc665efa8f959eab9b04e7e856f3688/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs#L125) that we use to render the foreign bindings.

This pass would convert `ComponentInterface` and `Config` into something simpler from the POV of the template code.   For example, something like this:

```
struct BindingsInterface {
   functions: Vec<Function>,
   classes: Vec<Class>,
   records: Vec<Records>,
   enums: Vec<Enum>,
   ...
}

struct Function {
   // Name of the function for the bindings (after applying the `fn_name` filter and any other transformations needed) 
   name: String,
   is_async: bool,
   args: Vec<NameAndType>,
   return_type: Option<Type>,
   ...
}

struct NameAndType {
   // Name of the field/argument, etc.
   name: String,
   // Concrete type name (again, after applying `type_name` and any other transformations)
   type_name,
   // Name of the FFI converter struct
   ffi_converter: String,
   ...
}

...many more structs, hopefully I painted an okay picture of how they would look
```
The template code could then render this more-or-less directly, without filters or extra processing (`fun {{ function.name }}` rather than `fun {{ function.name|fn_name }}`).

I think there would be quite a few benefits of doing this:
  - Less need to pass around values in the template code.  For example, we currently always need to pass around both the `Config` and `ComponentInterface` [when we want to render a type name](https://github.com/mozilla/uniffi-rs/blob/dee12a5271c44fc540bd11e1f6c9a316ea1aded6/uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt#L21).  In [#2176](https://github.com/mozilla/uniffi-rs/pull/2176/), Bastian is adding the ability to rename things and this is going to require passing the `Config` in more places.
  - Dealing with options is easier in Rust than in the templates.  For example, [this template code](https://github.com/mozilla/uniffi-rs/blob/f4ca11f2ebc665efa8f959eab9b04e7e856f3688/uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceImpl.kt#L51) is pretty messy and would be simpler in Rust.  I think this may extend to other matching/conditional as well.
  - I think it would remove the need for a lot of our macro code, which is hard to follow.
[...]

PR #2191, opened 2024-07-19
Replace templating filters for Python

gruberb:
This PR aims to replace the majority of template filters being used in Python, and adds a new `PythonInterfaceAdapter`, which takes the `ComponentInterface`, and adds the filter changes previously applied in the template, directly on a Rust object, which can be passed on to the template.

Partly addresses: https://github.com/mozilla/uniffi-rs/issues/2184

## Python Filters this PR removes

- [x] `fn_name`
- [ ] `class_name` (one instance left, special case)
- [ ]  `var_name` (one instance left, special case)
- [x]  `enum_variant_py`


gruberb:
@bendk @mhammond This is still super WIP, and the goal is to do one language at a time. But would like to get an early design feedback so I don't stray too far from the path.

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -298,10 +299,13 @@ pub struct PythonWrapper<'a> {
     type_imports: BTreeSet<ImportRequirement>,
 }
 impl<'a> PythonWrapper<'a> {
-    pub fn new(config: Config, ci: &'a ComponentInterface) -> Self {
+    pub fn new(config: Config, ci: &'a mut ComponentInterface) -> Self {
         let type_renderer = TypeRenderer::new(&config, ci);
         let type_helper_code = type_renderer.render().unwrap();
         let type_imports = type_renderer.imports.into_inner();
+
+        ci.apply_naming_conventions(PythonCodeOracle::default());
```
mhammond, line 307: Thanks for having a poke at this! Another approach (which may or may not work :) is some technique for "walking/visiting" a CI - possibly even creating a new CI as we go rather than mutating a single one. That might allow us to avoid having the interface itself be aware of any new traits etc. If you squint it ends up looking a lot like our ["external type converter"](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_meta/src/group.rs#L88)

mhammond (commented):
Regardless of the implementation, this PR certainly demonstrates the benefits the binding templates would see, which is great!

uniffi_bindgen/src/bindings/python/templates/ExternalTemplate.py
```diff
@@ -3,7 +3,7 @@
 # External type {{name}} is in namespace "{{namespace}}", crate {{module_path}}
 {%- let ffi_converter_name = "_UniffiConverterType{}"|format(name) %}
 {{ self.add_import_of(module, ffi_converter_name) }}
-{{ self.add_import_of(module, name|class_name) }} {#- import the type alias itself -#}
```
gruberb, line 6: Note: Check if this is actually correct.

uniffi_bindgen/src/lib.rs
```diff
@@ -157,6 +158,49 @@ pub trait BindingGenerator: Sized {
...
+    /// If we support callback interfaces, the protocol name is the object name, and the class name is derived from that.
+    /// Otherwise, the class name is the object name and the protocol name is derived from that.
+    ///
+    /// This split determines what types `FfiConverter.lower()` inputs.  If we support callback
+    /// interfaces, `lower` must lower anything that implements the protocol.  If not, then lower
+    /// only lowers the concrete class.
+    fn object_names(&self, obj: &Object) -> (String, String);
+}
+
+pub trait Renameable {
```
gruberb, line 198: Trait might not be needed, implement `rename` directly for desired structs.

uniffi_bindgen/src/lib.rs
```diff
@@ -157,6 +158,49 @@ pub trait BindingGenerator: Sized {
     ) -> Result<()>;
 }
 
+pub trait CodeOracle {
```
mhammond, line 161: Part of the motivation for my previous comment is that I really don't like trying to have a `CodeOracle` outside the bindings - it implies there will be a single trait that will capture the requirements of all internal and external binding languages, both current and future, which is something I'm skeptical of. Something that could "walk" the CI might mean the Oracles can remain in the bindings (or indeed, they can implement what they need any way they see fit).

(We saw something similar in the early days - trying to share too much code between the bindings ended up meaning we were forever working around impedance mismatches and avoiding this had significant benefits overall - eg, the `AsCodeType` trait was in a central place, but that turned out to be a mistake)

gruberb: Thanks! This makes a lot of sense! I continued with the work because I wanted it to pass first, so to be sure the Rust functionality can cover the base cases for the Python templating. I am at a point now where I can rework it, and take inspiration from your mentioned "[external type converter"](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_meta/src/group.rs#L88). 

So for this change, I before passing the generated `CI` down to the `PythonWrapper` (for example), we can walk through each field of the `ci`, adapt the names, and create a specific `python_ci`. Is this what you had in mind?

mhammond: I was thinking code in each of the bindings can explicitly do that itself. I'm actually a little vague because it's not immediately clear what will work! That "external type converter" isn't a great fit - I more meant very vaguely how it visits all parts of the metadata rather than the concrete implementation - for one thing it's walking a different data structure.

uniffi_bindgen/src/interface/conventions.rs
```diff
@@ -0,0 +1,127 @@
...
+use crate::{CodeOracle, ComponentInterface, Renameable};
+use crate::interface::{FfiDefinition, Record};
+
+impl ComponentInterface {
+    pub fn apply_naming_conventions<O: CodeOracle>(&mut self, oracle: O) {
+
+        for (_, t) in &mut self.types.type_definitions {
+
+            if t.name().is_some() {
+                t.rename(oracle.var_name(&t.name().unwrap()));
```
bendk, line 12: I think this should be either `class_name` since it's usually something like a record/enum/interface name.

At first I thought it should be a match statement that handled different variants differently, but AFAICT it's always a class name.

uniffi_bindgen/src/interface/conventions.rs
```diff
@@ -0,0 +1,127 @@
...
+            }
+        }
+
+        let mut known: BTreeSet<Type> = BTreeSet::new();
+
+        for t in &mut self.types.all_known_types.iter() {
+            dbg!("type in all_known {:#?}", t.clone());
+            let mut ty = t.clone();
+            if t.name().is_some() {
+                ty.rename(oracle.external_types_name(&t.name().unwrap()));
```
bendk, line 23: I think this should also be `class_name`.  In general, `types.all_known_types` should be handled like `types.type_definitions`

uniffi_bindgen/src/interface/conventions.rs
```diff
@@ -0,0 +1,127 @@
...
+
+            for cons in &mut object_item.constructors {
+                if !cons.is_primary_constructor() {
+                    cons.rename(oracle.fn_name(cons.name()));
+                }
+
+            }
+        }
+
+        for field in self.ffi_definitions() {
```
bendk, line 110: If possible, I think this shouldn't be renamed since they're part of the FFI and not user-facing.

bendk (commented):
Thanks for working through this one, it's a large change and it's not totally clear how it should work.  I left a couple of suggested changes in apply_naming_conventions, but for the most part that looks right to me.  I think the main work remaining is to a) fix any issues there and b) figure out where that in the codebase that code should live.

gruberb:
>  I think the main work remaining is to a) fix any issues there and b) figure out where that in the codebase that code should live.

Agree. After working on this quite a while, I sense that we might want to tackle this "filter-by-filter". So we could focus on removing `class_name` filter and apply the changes to the local structure. I could completely replace the `class_name`, and `var_name` is close to be gone as well. 

For "where should this code live". It is also now very clear to create a language specific structure which holds the updated structure, and pass it on to the template. So it's 
```
UDL|Macro -> ComponentInterface -> {Language}InterfaceAdapter* -> Template
``` 

The `struct PythonInterfaceAdapter` could hold all the CI + adapted changes. Which then we pass on to the templates.

uniffi_bindgen/src/interface/function.rs
```diff
@@ -73,6 +73,10 @@ impl Function {
         &self.name
     }
 
+    pub fn rename(&mut self, new_name: String) {
+        self.name = new_name;
```
bendk, line 77:
```suggestion
        // Note: Don't recalculate the checksum.  In order to have consistent checksums, we use the Rust names as input.
        self.name = new_name;
```

uniffi_meta/src/types.rs
```diff
@@ -143,6 +144,41 @@ impl Type {
...
+                    None
+                }
+            }
+            _ => None,
+        }
+    }
+
+    // Currently we just change the `import_name`. Different languages have slightly different
+    // naming patterns. We don't change the name of the external type, because the original
+    // names are used to find them in the `ComponentInterface`.
```
bendk, line 176: This is a bit surprising to me.  I would think that we need to update the names for the types as well.  Maybe this is currently working because both Rust and Python use UpperCamelCase.  What if you updated the `class_name` function so that prepended `Py` to all classes or something like that to test that the type names would line up if the language bindings changed them.

My guess is that you would need to translate the `name` field for all types and if you did that, then you wouldn't need to add the `Type::External::import_name` field.

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -437,6 +441,156 @@ impl PythonCodeOracle {
...
+        let mut new_records: BTreeMap<String, Record> = BTreeMap::new();
+
+        for (key, record_item) in ci.records.iter_mut() {
+            let mut record = record_item.clone();
+
+            // We just want to prefix reserved keywords for the name, without modifying it
+            record.rename(self.class_name(record_item.name()));
+
+            for field in &mut record.fields {
+                field.rename(self.var_name(field.name()));
```
bendk, line 456: If I'm right about needing to rename type names, then I think you also would want to have a `field.rename_type()` method that called here.  This also applies to enum fields and function/method/constructor arguments.

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -437,6 +441,156 @@ impl PythonCodeOracle {
...
+            }
+
+            new_records.insert(key.to_string(), record);
```
bendk, line 459: Maybe key should be transformed as well?  I think this goes with the suggestion to transform the `Type` instances.

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -437,6 +441,156 @@ impl PythonCodeOracle {
     }
 }
 
+impl VisitMut for PythonCodeOracle {
+    fn visit_record(&self, ci: &mut ComponentInterface) {
+        // Conversions for RecordTemplate.py
+        let mut new_records: BTreeMap<String, Record> = BTreeMap::new();
+
+        for (key, record_item) in ci.records.iter_mut() {
```
bendk, line 449: What do you think about moving loops like this into the `ComponentInterface` code?  I'm picturing something like this:

```rust
impl ComponentInterface {
    fn visit_mut<V: VisitMut>(&mut self, visitor: V) {
        for record_item in self.records.values_mut() {
            visitor.visit_record(self, record_item);
            for field in &mut record.fields {
                visitor.visit_field(self, record_item, field);
            }
        }
        self.fix_keys_after_rename(self.records);

        // ...more loops here
    }
}
```

The benefit is that we only have to write the tree-walking code once rather than duplicate it for each language.  Also, you wouldn't need to change the CI field visibility.  It's hard to strike the right balance on shared code in `ComponentInterface` and language-specific code, but this feels right to me.  Please to push back on this suggestion if you don't like it.  You know the details of this process better than me.

bendk (changes requested):
This is looking really good.  I had some regular-sized suggestions and also the suggestion to rename type names, which may blow up into a lot of work.  If it does, maybe we could defer that work.  I think that means not touching the `class_name` filter and only focusing on the `var_name`, `fn_name`, and `enum_variant_py` filters.

uniffi_meta/src/types.rs
```diff
@@ -143,6 +143,58 @@ impl Type {
...
+    fn rename(&mut self, new_name: String) {
+        match self {
+            Type::Object { name, .. } => *name = new_name,
+            Type::Record { name, .. } => *name = new_name,
+            Type::Enum { name, .. } => *name = new_name,
+            Type::External { name, .. } => *name = new_name,
+            Type::Custom { name, .. } => *name = new_name,
+            Type::Optional { inner_type } | Type::Sequence { inner_type } => {
+                inner_type.rename(new_name);
+            }
```
bendk, line 168: Do you need this or does `rename_recursive` handle it?

bendk (approved):
This looks really great to me.  I left a couple of suggestions, but nothing is big.  Feel free to fix them however you want or skip them altogether.  I don't think I need to re-review.

uniffi_bindgen/src/interface/conventions.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Each suppoerted language implements the [`VisitMut`] Trait and is able
+    /// to alter the functions, enums etc. to its own naming conventions.
+    pub fn visit_mut<V: VisitMut>(&mut self, visitor: &V) {
+        for type_ in self.types.type_definitions.values_mut() {
+            visitor.visit_type(type_);
+        }
+
+        let mut all_known_types_altered: BTreeSet<Type> = BTreeSet::new();
+
+        for type_ in self.types.all_known_types.iter() {
```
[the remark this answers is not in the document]
gruberb: The only way to alter a `BTreeSet` is to take out the element (via `self.types.all_known_types.get()`), mutate it, and insert it back. Since we are modifying all elements, it is more efficient to just create a new `BTreeSet` all together. AFAIK, there is no `iter_mut` implementation for this structure. But maybe I am missing something!

gruberb:
> I haven't had a look in detail at this, but it seems the filter functions haven't actually been removed. Is there a reason we can't remove them?

Two are removed (`fn_name` and `enum_variant_name`). Two are still there (`class_name` and `var_name`). They treat one special case each: We were not sure if adding a `is_name` attribute to the general type, or if we should add the filter directly in the template. So we left this out for future reasoning about, and I left a comment where it is currently being used.

The `class_name` filter is also being used in one spot, to rename the `callback_interface` function. I want to tackle this in a separate PR, because this little renaming has a huge impact on everything else.

uniffi_bindgen/src/interface/conventions.rs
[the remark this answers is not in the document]
bendk: Nope, I was missing that this was a BTreeSet and not a BTreeMap (even though I typed out the type name :man_facepalming: )

PR #2195, opened 2024-07-25
Provide more external type information to `FfiType::RustBuffer`

skeet70:
All the currently supported official languages support some form of aliasing so nothing output changes for them. For languages without aliasing (ie Java), this allows the fully qualified type name to be generated in place where necessary.

The `uniffi_testing` switch from `cargo build` to `cargo test --no-run` includes dev dependencies in the message tree, which is needed for external binding authors where the uniffi fixtures are included as dev dependencies (vs in the same workspace as they are here). Making `no_deps` during component discovery is for the same problem, deps are needed when discovering external `cdylibs` to test with.

[`uniffi-bindgen-java`](https://github.com/IronCoreLabs/uniffi-bindgen-java) does test these changes by its test suite running at all, this fixes a regression (from our POV) introduced in [this commit](https://github.com/mozilla/uniffi-rs/commit/4803eb4d6ff32fd51456be084b4797080860d24a). If there are tests I can write to cover both use cases (#2183 and ours), I can do it given direction.

### DIFF
--- a/uniffi_bindgen/src/interface/ffi.rs
+++ b/uniffi_bindgen/src/interface/ffi.rs
@@ -150,7 +150,7 @@ impl From<&Type> for FfiType {
             } => FfiType::RustBuffer(Some(ExternalMetadata {
                 name: name.clone(),
                 module_path: module_path.clone(),
-                namespace: namespace.clone()
+                namespace: namespace.clone(),
             })),
             Type::Custom { builtin, .. } => FfiType::from(builtin.as_ref()),
         }
@@ -161,7 +161,7 @@ impl From<&Type> for FfiType {
 pub struct ExternalMetadata {
     pub name: String,
     pub module_path: String,
-    pub namespace: String
+    pub namespace: String,
 }
 
 // Needed for rust scaffolding askama template
--- a/uniffi_bindgen/src/interface/ffi.rs
+++ b/uniffi_bindgen/src/interface/ffi.rs
@@ -43,7 +43,7 @@ pub enum FfiType {
     /// or pass it to someone that will.
     /// If the inner option is Some, it is the name of the external type. The bindings may need
     /// to use this name to import the correct RustBuffer for that type.
-    RustBuffer(Option<String>),
+    RustBuffer(Option<ExternalMetadata>),
     /// A borrowed reference to some raw bytes owned by foreign language code.
     /// The provider of this reference must keep it alive for the duration of the receiving call.
     ForeignBytes,
@@ -144,11 +144,24 @@ impl From<&Type> for FfiType {
             Type::External {
                 name,
                 kind: ExternalKind::DataClass,
+                module_path,
+                namespace,
                 ..
-            } => FfiType::RustBuffer(Some(name.clone())),
+            } => FfiType::RustBuffer(Some(ExternalMetadata {
+                name: name.clone(),
+                module_path: module_path.clone(),
+                namespace: namespace.clone()
+            })),
             Type::Custom { builtin, .. } => FfiType::from(builtin.as_ref()),
         }
     }
+}
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct ExternalMetadata {
+    pub name: String,
+    pub module_path: String,
+    pub namespace: String
 }
 
 // Needed for rust scaffolding askama template
--- a/uniffi_testing/src/lib.rs
+++ b/uniffi_testing/src/lib.rs
@@ -159,7 +159,8 @@ fn get_cargo_metadata() -> Metadata {
 
 fn get_cargo_build_messages() -> Vec<Message> {
     let mut child = Command::new(env!("CARGO"))
-        .arg("build")
+        .arg("test")
+        .arg("--no-run")
         .arg("--message-format=json")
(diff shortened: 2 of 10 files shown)


uniffi_testing/src/lib.rs
```diff
@@ -159,7 +159,8 @@ fn get_cargo_metadata() -> Metadata {
 
 fn get_cargo_build_messages() -> Vec<Message> {
     let mut child = Command::new(env!("CARGO"))
-        .arg("build")
+        .arg("test")
+        .arg("--no-run")
```
bendk, line 163: Could this be `build --profile=dev`?  That seems to capture the intent better to me.

uniffi_bindgen/src/interface/ffi.rs
```diff
@@ -144,13 +144,26 @@ impl From<&Type> for FfiType {
...
         }
     }
 }
 
+#[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct ExternalMetadata {
+    pub name: String,
+    pub module_path: String,
+    pub namespace: String,
+}
```
bendk, line 165: This makes a lot of sense to me, but I know that @mhammond has some plans for external types.  Would this change conflict with them?

Hopefully at some point we can consolidate `module_path` and `namespace`, but it makes sense to me to have 2 fields for now since `Type::External` also have 2 fields.

mhammond (commented):
It might make sense to split this so the 2 issues being dealt with here can be considered independently.

Regarding `no_deps`, I'm not sure this command-line flag is the way forward. As you will note in library_mode, the intent is that cargo metadata support can be disabled by a feature, so even at face value, this option existing in that scenario seems odd. But more generally, the longer term idea is that people who disable that feature (because they don't use cargo) probably still need the capability of supplying the map so the crate source can be located. ie, we might be better doing [what this comment](https://github.com/mozilla/uniffi-rs/blob/b97973dbbaa628db763dfb016bea5a1273b7f371/uniffi_bindgen/src/library_mode.rs#L145) says.

If that capability existed, then maybe our test code could itself just invoke cargo metadata however it wants, do the simple parsing and pass the resulting map in.

I must however admit that I'm mildly concerned this `no_deps()` might cause other breakage - the fact it doesn't break uniffi itself is one thing, but if it ends up causing breakage for other build environments we have a bit of a problem and might need to rethink it. In other words, maybe the issue we are having here might just be because it's a non-trivial build environment and the problem will actually turn out to be common. I guess I need to at least see what happens in our own build environment!

bendk (commented):
This looks really good to me.  The only reason I'm not approving now is that I want to give Mark some time to comment.

bendk:
> If that capability existed, then maybe our test code could itself just invoke cargo metadata however it wants, do the simple parsing and pass the resulting map in.

Maybe the test code could even build the map itself and avoid the cargo metadata call.  That could lead to some speedups.

uniffi_bindgen/src/interface/ffi.rs
mhammond: I do want to kill `Type::External` and replace it with the actual type but that's my problem :)

I don't get why we need both `namespace` and `module_path` though? If the name is to the Rust ffi function `namespace` might be problematic - [that's not known by the udl/scaffolding](https://github.com/mozilla/uniffi-rs/blob/b97973dbbaa628db763dfb016bea5a1273b7f371/uniffi_udl/src/finder.rs#L159), so gets [fixed for the binding generation](https://github.com/mozilla/uniffi-rs/blob/b97973dbbaa628db763dfb016bea5a1273b7f371/uniffi_meta/src/group.rs#L240) - meaning rust code generated from udl might have a different value than the bindings get.

`ExternalFfiMetadata` or similar is probably a better name so it's not confused as a `uniffi_meta` struct.

Not that you should add it now, but I wonder if you will ever strike a `RustArcPtr` ever needing a fq-name too?

skeet70: The `namespace` and `module_path` are needed to [generate the qualified type path for `RustBuffer` in Java](https://github.com/IronCoreLabs/uniffi-bindgen-java/blob/main/src/gen_java/mod.rs#L181-L189). The `module_path` is used to get the crate to look up in `config.external_packages`, and the `namespace` is needed for the default if nothing is  found.

uniffi_testing/src/lib.rs
skeet70: It was already `build` (without `--release`), which should've defaulted to `--profile=dev` right? `cdylibs` of [uniffi-bindgen-java](https://github.com/IronCoreLabs/uniffi-bindgen-java/blob/main/Cargo.toml#L33) weren't found on creation of the `TestHelper` as it was. `cargo build --profile=test` might work, but that feels a little less straightforward than `test --no-run` to me personally.

uniffi_bindgen/src/interface/ffi.rs
mhammond: Right, same as us :) But yeah, that `namespace` probably isn't going to work in many cases - in the case it does (eg, "single crate with udl") namespace is probably module path anyway. But yeah, I'm totally fine with carrying that all around, I just long for a time when it's less muddy.

mhammond (approved):
Thanks!

PR #2198, opened 2024-07-26
wip: poc of trait + language interface

gruberb:
This is just a POC to discuss how it might look like when we convert the ComponentInterface into a trait. Implemented for Python in this branch. Not final version, some things are not working, but the overall design would look like this.

Note that We still have a `ComponentInterface` struct, because we just converted Python to the trait impl here. The other languages still rely on the old structure.

https://github.com/mozilla/uniffi-rs/issues/2184


uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -108,6 +111,56 @@ static KEYWORDS: Lazy<HashSet<String>> = Lazy::new(|| {
...
+            .ok_or_else(|| lookup_error!("record {name} not found"))
+    }
+
+    /// Get an Object definition by name
+    pub fn get_object_definition_filter<'a>(&self, name: &str) -> askama::Result<&'a Object> {
+        self.get_object_definition(name)
+            .ok_or_else(|| lookup_error!("object {name} not found"))
+    }
+}
+impl From<ComponentInterface> for PythonComponentInterface {
```
gruberb, line 149: This might be where we apply naming conventions etc. So we just iterate once over all the structures and apply the changes.

PR #2199, opened 2024-07-26
Allow UDL to avoid the `[Rust=...]` attribute by using a plain-old ty…

mhammond:
…pedef

This is a WIP and needs a little cleaning up, but thought I'd throw it out there. It allows people to stop using the `[Rust=...]` attribute in UDL.

This doesn't seem *that* interesting now, but it will really make sense once Ben's other work lands and we could take this further and use it with `[External]` (to, eg, drop the entire `ExternalInterface` attribute) and with `[Remote]`.

### DIFF
--- a/docs/manual/src/udl/ext_types.md
+++ b/docs/manual/src/udl/ext_types.md
@@ -14,8 +14,7 @@ giving more detail.
 ## Types from procmacros in this crate.
 
 If your crate has types defined via `#[uniffi::export]` etc you can make them available
-to the UDL file in your own crate via a `typedef` with a `[Rust=]` attribute. Eg, your Rust
-might have:
+to the UDL file in your own crate via a `typedef` describing the concrete type.
 
 ```rust
 #[derive(uniffi::Record)]
@@ -26,8 +25,7 @@ pub struct One {
 you can use it in your UDL:
 
 ```idl
-[Rust="record"]
-typedef extern One;
+typedef record One;
 
 namespace app {
     // use the procmacro type.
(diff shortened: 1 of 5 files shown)


docs/manual/src/udl/ext_types.md
```diff
@@ -37,6 +35,17 @@ namespace app {
...
-*  "enum", "trait", "callback", "trait_with_foreign"
-* For records, either "record" or "dictionary"
+* "enum", "trait", "callback", "trait_with_foreign"
+* For records, either "record", "dictionary" or "struct"
 * For objects, either "object" or "interface"
+
+eg:
+```
+typedef enum MyEnum;
+typedef interace MyObject;
```
badboy, line 45:
```suggestion
typedef interface MyObject;
```

docs/manual/src/udl/ext_types.md
```diff
@@ -37,6 +35,17 @@ namespace app {
...
+```
+etc.
+
+Note that in 0.28 and prior, we also supported this capability with a `[Rust=]` attribute.
+However, this new model is simpler and more consistent with future UDL plans so
+is recommended.
```
badboy, line 51: We can probably just call the `[Rust=]` part deprecated.

badboy (commented):
I am all here for the simplification!

uniffi_udl/src/finder.rs
```diff
@@ -129,43 +129,89 @@ impl TypeFinder for weedle::TypedefDefinition<'_> {
...
+                        "dictionary" | "record" | "struct" => Type::Record {
+                            module_path,
+                            name,
+                        },
+                        "enum" => Type::Enum {
+                            module_path,
+                            name,
+                        },
+                        "custom" => panic!("don't know builtin"),
+                        "interface" | "impl" => Type::Object {
```
mhammond, line 160: I should call out that I added a couple of names compared to the Rust attribute. In particular, `struct` as an alias for `record` above, and `impl` as an alias for `interface` here. My thinking was that these names more closely match what a Rust programmer is likely to think of - ie, a `derive(uniffi::Record)` lives on a `struct` defn while `uniffi::export` lives on an `impl` block.

On the negative side, `impl` is also used for traits.

I personally quite like these names, but don't feel too strongly about it, so will happily concede to someone who strongly dislikes them!

mhammond:
Note that I snuck in a change to move a previously misplaced CHANGELOG entry here.

PR #2200, opened 2024-07-29
Fix so cargo_metadata isn't a dependency without the cargo_metadata feature

mhammond:
The `uniffi_test` crate unconditionally depends on `cargo_metadata`, and this will be difficult to change. This means that because `uniffi_test` is an unconditional dependency of `uniffi_bindgen`, the `cargo_metadata` feature doesn't actually prevent `uniffi_bindgen` from depending on `cargo_metadata`.

The top level `uniffi` create already has a feature `bindgen-tests`. This patch introduces a feature of the same name to `uniffi_bindgen`. Enabling that feature also enables the metadata feature. This means that if you do not enable either of these features you should be able to avoid `cargo_metadata`.

This patch also renames the new `cargo_metadata` feature to be `cargo-metadata` to be more consistent with our existing features and Rust feature naming in general.

This is on the path to trying to untangle the requirements for #2195, so I'd love feedback from @skeet70 - note that this doesn't actually fix the issue found there, but it's a step in the right direction. My intention is to push the actual execution of `cargo metadata` out to both the cli and to `uniffi_testing`, and this is a step in that direction.

### DIFF
--- a/fixtures/benchmarks/Cargo.toml
+++ b/fixtures/benchmarks/Cargo.toml
@@ -20,7 +20,7 @@ criterion = "0.5.1"
 uniffi = { workspace = true, features = ["build"] }
 
 [dev-dependencies]
-uniffi_bindgen = {path = "../../uniffi_bindgen"}
+uniffi_bindgen = {path = "../../uniffi_bindgen", features = ["bindgen-tests"]}
 
 [[bench]]
 name = "benchmarks"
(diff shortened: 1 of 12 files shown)


skeet70 (approved):
Yeah, I think this makes sense as prerequisite of moving where the metadata run (or lack of it) happens.

badboy (approved):

PR #2201, opened 2024-07-31
Move cargo_metadata execution out to the edges (ie, to the cli)

### LINKED ISSUE #2183: Only invoke cargo metadata for desired targets

mhammond:
This continues the cargo_metadata feature work but making the execution of cargo_metadata the resonsibility of the uniffi_bindgen callers rather that executing it implicitly. This means the CLI, which in-turn means the top-level uniffi crate also gets a `cargo-metadata` feature.

This reverts what we did to fix #2183 - by making `no_deps` the default, it means we will be unable to support reuse of UniFFI components, because it means we only support all components being in the same workspace. While this is a common use-case, it's not the only use-case we want to support. So grabbing all dependencies from cargo_metadata is again the default, but there's a new command-line option to avoid it.

It also replaces some of #2195.

@skeet70, does this make sense to you? Please feel free to bikeshed the names, I'm not particularly happy with them - and anything else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
 
 ### What's new?
 
+- A new bindgen command line option `--metadata-no-deps` is available to avoid processing
+  cargo_metadata for all dependencies.
+
 - Objects error types can now be as `Result<>` error type without wrapping them in `Arc<>`.
 
 - Swift errors now provide `localizedDescription` ([#2116](https://github.com/mozilla/uniffi-rs/pull/2116))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1475,6 +1475,7 @@ version = "0.28.0"
 dependencies = [
  "anyhow",
  "camino",
+ "cargo_metadata",
  "clap",
  "trybuild",
  "uniffi_bindgen",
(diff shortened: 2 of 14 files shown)


bendk (approved):
These changes look good to me.  I agree that it seems like there could be some better names, but I don't have any suggestions.

skeet70 (approved):
This makes sense to me. To get the `From<Metadata>` impl for the provided `CrateConfigSupplier` it does mean I need my bindgen crate on the same version of `cargo_metadata` as `uniffi_bindgen`. Obviously I could get around that by implementing my own supplier though.

I confirmed that it did work through `uniffi-bindgen-java`. I can't think of better names, they made some sort of sense to me.

PR #2202, opened 2024-07-31
Adding changelog entries for merges since 0.28.0

bendk:
I'm hoping to do a 0.28.1 release and these would be good to have in it.

I also considered adding an entry for the work that @mgeisler has been doing, but I couldn't think of a good wording for that.  @mgeisler do you have any suggestions for changelog entry text?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@
 
 ## [[UnreleasedUniFFIVersion]] (backend crates: [[UnreleasedBackendVersion]]) - (_[[ReleaseDate]]_)
 
+### What's new?
+
+- Lift errors will not cause an abort when `panic=abort` is set.
+- Added the `cargo_metadata` feature, which is on by default.  In some cases, this can be disabled
+  for better compatibility with projects that don't use cargo.
+
+### What's changed?
+- Kotlin will use the more efficient Enum.entries property instead of Enum.values() when possible
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
 
 ## [[UnreleasedUniFFIVersion]] (backend crates: [[UnreleasedBackendVersion]]) - (_[[ReleaseDate]]_)
 
+### What's new?
+
+- Lift errors will not cause an abort when `panic=abort` is set.
+- Added the `cargo_metadata` feature, which is on by default.  In some cases, this can be disabled
+  for better compatibilyt with projects that don't use cargo.
```
jmartinesp, line 13:
```suggestion
  for better compatibility with projects that don't use cargo.
```

Also maybe mention https://github.com/mozilla/uniffi-rs/pull/2182 ? Something like "In Kotlin bindings the more efficient `Enum.entries` property will be used instead of `Enum.values()` when possible". Sorry, I forgot to add this myself in the PR.

mhammond (approved):

mhammond:
> I'm hoping to do a 0.28.1 release and these would be good to have in it.

A 0.29 might not hurt before we land the breaking changes queued up too (I'm thinking the custom type changes) - I also just noticed I screwed up some changelog entries which are listed as being in 0.28 but are not! I'll try and remember to put a PR up to fix that.

PR #2204, opened 2024-08-06
Record metadata for a struct implementing a trait.

### LINKED ISSUE #2196: Tell the bindings about traits implemented by objects.
In #2169 there's:

```
#[uniffi::export]
impl AutoComplete for Bip39AutoComplete {
```

and we should support it! It could mean
* Any time (some times?) a trait is used, eg in a function signature the object could be used directly, which isn't possible today (you'd need a Rust wrapper to convert it, which is odd)
* The bindings could generate inheritance (which IIUC is what #2169 is asking for)

Generating the metadata for this is actually quite easy and sane (I've a wip ;), and we could land this without bindings support, which would be the hard part to do "correctly" - but that could come later and it's not strictly necessary for all bindings to support it.

wdyt?

mhammond:
Record metadata for a struct implementing a trait.

```
#[uniffi::export]
impl MyTrait for MyObject { ... }
```

Previously worked as it ignored `MyTrait`. This adds new metadata to record it,
allowing foreign bindings to implement things like inheritance.

Includes Python generating an inheritance chain to reflect this relationship.

This will not generate correct code if a struct declares more than 1 trait,
and there's some undesirable re-wrapping when traits from these objects gets
passed back and forward, but seems to work surprisingly well.

On the path to #2196.

### DIFF
--- a/fixtures/coverall/tests/bindings/test_coverall.py
+++ b/fixtures/coverall/tests/bindings/test_coverall.py
@@ -441,9 +441,33 @@ class TraitsTest(unittest.TestCase):
         py_node.set_parent(None)
         traits[0].set_parent(None)
 
+    def test_struct_traits(self):
+        # A struct, but also a trait.
+        self.assertTrue(issubclass(Node, NodeTrait))
+        node = Node("node")
+        # We start off using methods declared on the struct. `repr()` is our magic `Debug` support.
+        self.assertTrue(repr(node).startswith("Node { name: Some(\"node\"), parent: Mutex { "))
+        # We can also use methods declared on the trait.
+        self.assertTrue(node.get_parent() is not None)
+
+        # Unfortunately, this trait object gets re-wrapped as it's passed back to Rust, because
+        # we can't easily know it's already the correct type.
+        expected_desc = "Some(Node { name: Some(\"via node\"), parent: Mutex { ";
+        self.assertTrue(node.describe_parent().startswith(expected_desc))
+
+        # But if we round-trip the original rust-implemented parent, we get a wrapped object.
+        # Not clear this can be fixed: We'd need FFI to "cast" our struct into the trait - but how can this work?
+        # Can't magically turn an `Arc<T>` into an `Arc<dyn X>`?)
+        node.set_parent(node.get_parent())
+        # ie, this now acts like PyNode below rather than the "raw" parent.
+        # actual: Some(UniFFICallbackHandlerNodeTrait { handle: 18 })
+        #self.assertEqual(node.describe_parent(), expected_desc)
+
+        node.set_parent(PyNode())
+        self.assertTrue(node.describe_parent().startswith("Some(UniFFICallbackHandlerNodeTrait { handle: "))
+
     def test_round_tripping(self):
         rust_getters = make_rust_getters();
-        coveralls = Coveralls("test_round_tripping")
         # Check that these don't cause use-after-free bugs
         test_round_trip_through_rust(rust_getters)
 
--- a/uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
@@ -25,7 +25,7 @@
 {%- if ci.is_name_used_as_error(name) %}
 class {{ impl_name }}(Exception):
 {%- else %}
-class {{ impl_name }}:
+class {{ impl_name }}({% for t in obj.trait_impls() %}{{ t.trait_name }},{% endfor %}):
 {%- endif %}
     {%- call py::docstring(obj, 4) %}
     _pointer: ctypes.c_void_p
--- a/uniffi_meta/src/metadata.rs
+++ b/uniffi_meta/src/metadata.rs
@@ -23,7 +23,7 @@ pub mod codes {
     pub const UNIFFI_TRAIT: u8 = 11;
     pub const TRAIT_INTERFACE: u8 = 12;
     pub const CALLBACK_TRAIT_INTERFACE: u8 = 13;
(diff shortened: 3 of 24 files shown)


mhammond:
Bindings all turned out to be fairly easy - which in some ways is "bad" because it highlights an inefficiency here, but good because it seems fine and useful.

mhammond:
The other languages actually aren't working :) So just Python for now.

fixtures/coverall/tests/bindings/test_coverall.py
```diff
@@ -441,9 +441,28 @@ def test_path(self):
...
         traits[0].set_parent(None)
 
+    def test_struct_traits(self):
+        # A struct, but also a trait.
+        node = Node("node")
+        # parent is a "raw" Rust object
+        self.assertEqual(node.describe(), "Node: name=Some(\"node\"), parent=true")
+        # Unfortunately, this trait object gets re-wrapped as it's passed back to Rust, because
+        # we can't easily know it's alreqdy the correct type.
+        # Nasty to rely on the repr of the inner mutex, but it neatly demonstrates the issue.
```
bendk, line 451: - Maybe also point people towards the `proc-macro` fixture so that they can see the normal usage without all the gotchas that you need to mention here.
- This re-wrapping also happens for regular trait objects right?  Do you think the handle changes I've been hoping to introduce do could address both cases?

uniffi_meta/src/metadata.rs
```diff
@@ -23,7 +23,7 @@ pub mod codes {
     pub const UNIFFI_TRAIT: u8 = 11;
     pub const TRAIT_INTERFACE: u8 = 12;
     pub const CALLBACK_TRAIT_INTERFACE: u8 = 13;
-    //pub const UNKNOWN: u8 = 255;
+    pub const AS_TRAIT: u8 = 14;
```
bendk, line 26: `AS_TRAIT` doesn't communicate what this is doing to me, I keep thinking of `AsRef<T>` and expecting some sort of casting to be happening.  That's essentially what happens when you use the object as a trait, but here we're just tracking the list.  What about something like `TRAIT_IMPLS`?

uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
```diff
@@ -8,7 +8,7 @@
 {% if ci.is_name_used_as_error(name) %}
 class {{ impl_name }}(Exception):
 {%- else %}
-class {{ impl_name }}:
+class {{ impl_name }}({% for t in obj.as_trait_impls() %}{{ t.trait_name|class_name }},{% endfor %}):
```
bendk, line 11: :exploding_head: it's amazing that this is all we need to support this feature.

bendk (changes requested):
Looks great and I'm pretty amazed how easy it was.  I just noticed a couple minor things.

I think this is a breaking change since the metadata format is going to change right?  If so, please bump `UNIFFI_CONTRACT_VERSION` and let's wait to merge until I can get a `0.28.2` release out.

fixtures/coverall/tests/bindings/test_coverall.py
mhammond: > This re-wrapping also happens for regular trait objects right? Do you think the handle changes I've been hoping to introduce do could address both cases?

It does, yes. It will be mildly more complicated to fix here, only because we have a handle/pointer to a struct that we need to convert to a handle/pointer to a dyn trait. I think we'd also need an FFI function something like `fn to_trait(ob: Arc<StructName>) -> Arc<dyn TraitName>`, which is just a trivial `Arc::new(*ob as dyn TraitName)` or similar, then some other magic in the bindings that knows to use this at the right time.

Certainly fixing the existing times we do this wrapping would be a pre-req though!

mhammond:
> I think this is a breaking change since the metadata format is going to change right? If so, please bump UNIFFI_CONTRACT_VERSION and let's wait to merge until I can get a 0.28.2 release out.

I'm a little unclear on this - I don't think it changes any existing metadata item formats, but does add a new one - does that qualify as needing this? I'd have though we expect consumers to ignore what they don't understand?

uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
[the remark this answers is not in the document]
mhammond: huh, I forgot about that work - we probably can, yeah.

bendk:
> > I think this is a breaking change since the metadata format is going to change right? If so, please bump UNIFFI_CONTRACT_VERSION and let's wait to merge until I can get a 0.28.2 release out.
> 
> I'm a little unclear on this - I don't think it changes any existing metadata item formats, but does add a new one - does that qualify as needing this? I'd have though we expect consumers to ignore what they don't understand?

I'm not clear either, but I think [this line will fail](https://github.com/mozilla/uniffi-rs/blob/8caac8e67c5782959d0c1b945ebfe489720c06c9/uniffi_bindgen/src/macro_metadata/extract.rs#L183) and that will bubble up to make the entire metadata parsing step fail.  I could definitely be wrong about that, but I think that's how it works.

I agree that it makes more sense to ignore things you don't understand.  `Metadata::read` should probably be changed to return `Result<Option<Self>>` .

mhammond:
I see, thanks for the explanation!

uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py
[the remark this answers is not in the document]
mhammond: done! Also opened #2270 to try and avoid making that mistake again :)

bendk (approved):
Thanks for updating this, the changes look great.

PR #2205, opened 2024-08-06
Expose the find_components function

bendk:
The generate_bindings docs mentioned wanting to expose a function that just finds ComponentInterface and config tables for each component and leaves the rest to the external bindings generator.  This is exactly what I want to use for uniffi-bindgen-gecko-js.

The `find_components` function was pretty much exactly what we wanted. I made some minor changes to it and exposed it as a pub function.

### DIFF
--- a/uniffi_bindgen/src/library_mode.rs
+++ b/uniffi_bindgen/src/library_mode.rs
@@ -22,17 +22,18 @@ use crate::{
 use anyhow::bail;
 use camino::Utf8Path;
 use std::{collections::HashMap, fs};
+use toml::value::Table as TomlTable;
 use uniffi_meta::{
     create_metadata_groups, fixup_external_type, group_metadata, Metadata, MetadataGroup,
 };
 
 /// Generate foreign bindings
 ///
+/// This replicates the current process used for generating the builtin bindings.
+/// External bindings authors should consider using [find_components], which provides a simpler
+/// interface and allows for more flexibility in how the external bindings are generated.
+///
 /// Returns the list of sources used to generate the bindings, in no particular order.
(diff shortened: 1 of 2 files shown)


uniffi_bindgen/src/library_mode.rs
```diff
@@ -128,11 +137,26 @@ fn find_components(
...
+                .get_toml(ci.crate_name())?
+                .unwrap_or_default();
+            Ok(Component { ci, config })
         })
         .collect()
 }
 
+/// Run [find_components] using cargo-metadata to find UDL and TOML files
+#[cfg(feature = "cargo-metadata")]
+pub fn find_components_cargo(library_path: &Utf8Path) -> Result<Vec<Component<TomlTable>>> {
```
mhammond, line 150: Is there any way to push this out to the caller? I'm thinking of the --no-deps issue and how a user with that problem would avoid it here?

bendk: I can think of a couple options, what do you think is best?
 - Adding add a `no_deps` flag
 - Not exposing this for now and forcing users to use the `find_components` function.  If we get requests to support it, we can ask those users what API they want.

mhammond: I was thinking that this entire function just move to the caller (which is the gecko bindings?) - so yeah, only exposing `find_components` and leave the job of running cargo-metadata to the callers, like it is now in the cli and tests.

bendk: Sounds good, just pushed an update that removes `find_components_cargo`.

mhammond (approved):

PR #2207, opened 2024-08-08
Remove panic logging

### LINKED ISSUE #2203: Undefined features `log_panics` and `log_backtraces`
We have code feature-gated in `uniffi_core` on features that are not defined. It's therefore impossible to activate them (through cargo).

Here:

https://github.com/mozilla/uniffi-rs/blob/106f0b2d8a4680fd47585de0e9986c1261992926/uniffi_core/src/panichook.rs#L2

and here:

https://github.com/mozilla/uniffi-rs/blob/106f0b2d8a4680fd47585de0e9986c1261992926/uniffi_core/src/panichook.rs#L24

noticed that when compiling with Rust 1.80 thanks to [`check-cfg`](https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html#checked-cfg-names-and-values)

badboy:
It was migrated from ffi-support back in 2021,
but never actually worked because the used cargo features are undefined.

Fixes #2203


mhammond (approved):
I think killing this is fine

mhammond:
(well, after the rustfmt fixes :)

PR #2222, opened 2024-08-23
Switch from askama to rinja

GuillaumeGomez:
Recently, me and another `askama` maintainer forked the project into `rinja`. For more details about the why, I wrote a blog post [here](https://blog.guillaume-gomez.fr/articles/2024-07-31+docs.rs+switching+jinja+template+framework+from+tera+to+rinja). But in short: `rinja` is ahead of `askama` now in a lot of areas so I think going forward, your project will benefit more from it.

If you're not interested about this switch, don't hesitate to close the PR.

If you want more information about `rinja`, don't hesitate to ask!


uniffi_bindgen/src/bindings/swift/templates/macros.swift
```diff
@@ -157,15 +157,15 @@ v{{- field_num -}}
     {%- endfor %}
 {%- endmacro %}
 
-{%- macro async(func) %}
+{%- macro is_async(func) %}
```
GuillaumeGomez, line 160: The renaming for this macro is because `rinja` doesn't allow to name item from Rust keywords. Same for `try` below.

mhammond:
The last release of askama was over 18 months ago, so this might make sense. The changes you list between the forks all seems improvements to me and rustdocs moving to it is a good sign. So I guess I'm somewhere between +0 and +1 for this. Does anyone else have thoughts?

GuillaumeGomez:
It's also what we use in docs.rs. Anyway, you summed up the situation pretty nicely, nothing else to add. Don't hesitate if you have questions.

badboy:
I'm not opposed to switching over; I just want to call out that this requires us to properly [vet](https://mozilla.github.io/cargo-vet/) this, so we can import it into mozilla-central. As this is a fork, it shouldn't be too much work though. And right now uniffi is the only one with a dependency on askama, so we won't have both askama and rinja around on import.

gruberb:
After reading the blog post and the reasoning behind it, makes sense to me! I don't think we were limited by the template engine, but it's also not too bad and agree on following rustdocs. So 👍 all around :)

bendk:
I'm +1 on this.  I've wanted support for `..` patterns and calling methods for a while now.

uniffi_bindgen/src/bindings/ruby/templates/ErrorTemplate.rb
```diff
@@ -56,7 +56,7 @@ def to_s
 {%- for e in ci.enum_definitions() %}
 {% if ci.is_name_used_as_error(e.name()) %}
 {%- let typ=ci.get_type(e.name()).unwrap() %}
-{%- let canonical_type_name = canonical_name(typ.borrow()).borrow()|class_name_rb %}
+{%- let canonical_type_name = self::canonical_name(typ.borrow()).borrow()|class_name_rb %}
```
mhammond, line 59: huh, this will be the first use of `self::` in our templates - I didn't know that was a thing :) I'm sure we are doing quite a few things the hard way...

mhammond (commented):
This all looks great to me. As @badboy mentioned we probably do need one of us to agree to vetting it first (which is really a high-level diff against askama) and checking all the other new sub-deps are fine version wise. But I'm soon to take off for a month so can't commit to that before I return. I hope it lands before then though, thanks!

CHANGELOG.md
```diff
@@ -434,7 +434,7 @@ Significant patches to UniFFI's builtin bindings which you will need to port inc
 
 [All changes in v0.19.2](https://github.com/mozilla/uniffi-rs/compare/v0.19.1...v0.19.2).
 
-- Fixed sccache issue with the `askama.toml` config file.
```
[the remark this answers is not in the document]
GuillaumeGomez: Oups, too powerful `sed`. Fixing it.

uniffi_bindgen/src/bindings/ruby/templates/ErrorTemplate.rb
GuillaumeGomez: There were some weird cases in askama, like how they handle methods. Now we're closer to rust in this regard, making it much simpler (imo) to do. Docs are [here](https://rinja.readthedocs.io/) but if anything is unclear or missing, please don't hesitate to tell me (or opening an issue on the project) so we can fix it.

GuillaumeGomez:
Reverted changes in the CHANGELOG.

mhammond:
Thanks for those changes, but do you mind *adding* an entry for this?

GuillaumeGomez:
No problem!

GuillaumeGomez:
Done!

uniffi_bindgen/src/bindings/ruby/templates/ErrorTemplate.rb
bendk: I think the tradeoff is requiring `self::` for static methods, but not requiring `self.` for regular methods which I will definitely take ([example](https://github.com/mozilla/uniffi-rs/blob/74c065a6ddb98320a1abe1155fd52db5677ca6ea/uniffi_bindgen/src/bindings/kotlin/templates/Types.kt#L136-L142))

GuillaumeGomez: `self::` is to tell `rinja` that it's not a method actually. If you want a static method, it's either `Self::` or `TypeName::`. But in short, if it's a path (ie it contains `::`), `rinja` will not consider it a method.

GuillaumeGomez:
Fixed merge conflict.

mhammond:
It looks like the only new dep here is [once_map](https://github.com/a1phyr/once_map) which looks OK and I'd be fine approving. I did strike a strange issue when moz-central insisted on pulling 0.4.20 even though rinja is fine with [0.4.18](https://github.com/rinja-rs/rinja/blob/441756b712ae6a15c4db5851e45c70c19e00c4f1/rinja_derive/Cargo.toml#L42) - 0.4.28 wants a new hashbrown where .18 does not. I'm assuming that's a "me not working the tools correctly" problem...

This also will require a `rustc-hash` dupe - this needs 2.0 whereas everyone else in tree is on 1.? I don't *think* that will be a problem - @badboy @bendk wdyt?

badboy:
Any idea what the rustc-hash changes for 2.0 are? Maybe we can re-export that as we do for other things in m-c. But otherwise I don't think a single duplicated one like that will be problematic

Kijewski:
rustc-hash v2 has a new algorithm compared to v1: <https://github.com/rust-lang/rustc-hash/blob/master/CHANGELOG.md>.

mhammond:
> rustc-hash v2 has a new algorithm compared to v1

A quick look shows that it's really the use of `FxBuildHasher` which would avoid trivially using either version.

mhammond:
Just got confirmation the rustc-hash version isn't going to be a problem for this, so I'm fine with landing it once updated!

uniffi_bindgen/src/bindings/python/gen_python/mod.rs
```diff
@@ -569,71 +569,85 @@ pub mod filters {
     use super::*;
     pub use crate::backend::filters::*;
 
-    pub(super) fn type_name(as_ct: &impl AsCodeType) -> Result<String, askama::Error> {
+    pub(super) fn type_name(as_ct: &impl AsCodeType) -> Result<String, rinja::Error> {
         Ok(as_ct.as_codetype().type_label())
     }
 
-    pub(super) fn ffi_converter_name(as_ct: &impl AsCodeType) -> Result<String, askama::Error> {
+    //TODO: Remove. Currently just being used by EnumTemplate.py to
```
mhammond, line 576: looks like a merge error, these additions should be removed.

mhammond (changes requested):
looks like c4963eccd00bd24df9db8d1787d6625242237bce broke this, with python getting 
```
198 | #[derive(Template)]
    |          ^^^^^^^^
    |          |
    |          expected `&String`, found `String`
    |          expected due to the type of this binding
```
I'll see if I can spot it.

mhammond:
Might actually be 1cb0ca0131fba6ec738b9861da46abf3a9d53a86 which broke it

mhammond:
Looks like it might be `type_name` - https://github.com/mozilla/uniffi-rs/blob/a8da0375b66f5b433d79035976771d010151691b/uniffi_bindgen/src/bindings/python/templates/ObjectTemplate.py#L18-L22 - but that doesn't really make sense.

GuillaumeGomez:
Fixed my mess. :)

Lesson of the day: don't fix merge conflicts right before going to sleep.

mhammond (approved):
Thanks!

uniffi_bindgen/Cargo.toml
```diff
@@ -18,7 +18,7 @@ ffi-trace = ["uniffi_testing?/ffi-trace"]
 
 [dependencies]
 anyhow = "1"
-askama = { version = "0.12", default-features = false, features = ["config"] }
+rinja = { version = "0.3", default-features = false, features = ["config"] }
```
cschramm, line 21: I think this actually requires 0.3.4. With older versions I run into

```
   --> uniffi_bindgen-0.29.0/src/bindings/python/gen_python/mod.rs:186:10
    |
186 | #[derive(Template)]
    |          ^^^^^^^^
    |          |
    |          expected `&String`, found `String`
    |          expected due to the type of this binding
    |
    = note: this error originates in the derive macro `Template` (in Nightly builds, run with -Z macro-backtrace for more info)
```

badboy: Can you file this as a bug?

Issue #2224, opened 2024-08-29
Ability to completely remove logging code (likely with a feature flag)


timboudreau:
### The Problem

I am using uniffi with some code that does DSP - called from C, Objective C and Swift, and using the [os_log crate](https://github.com/steven-joruk/oslog) to integrate the `log` crate, which uniffi's generated code uses, with Mac OS's system-level logging.

The code uniffi generates logs *every inbound call into Rust* - due to the [lines visible in this diff](https://github.com/mozilla/uniffi-rs/compare/main...timboudreau:uniffi-rs:remove-logging-v0.26.1)

That might seem reasonable, assuming logging is zero-cost when not enabled (it is not always, and it is not with `os_log`), and assuming calls across the FFI boundary are *infrequent* (not a safe assumption at all).

Imagine you have some legitimate, useful `debug`-level logging you need to see happening in your Rust code you are calling.

Now imagine the FFI wrapper code you generated is also doing `debug`-level logging ... of a call that is happening 768,000 times per second (8 frequency bands x 96,000 samples per second).

Good luck reading that log.

Not to mention, there is measurable impact on performance of doing *anything* "near-zero cost" ... if you do it 768,000 times a second.

### The Solution

It seems to me that unasked for, *unremovable without forking* logging is a bug, not a feature.  I can totally understand wanting logging of FFI calls *sometimes*.  The logging code should, IMHO, be off-by-default but enablable with a feature flag.

Or, if that is likely to break existing code that expects it, at the very least, able to be completely excised with a feature flag.

badboy:
These are [about to become `trace` level](https://github.com/mozilla/uniffi-rs/pull/2136), so at least some of the issues you report become less of a problem.
As a workaround right now you could add a filter. We do that [over in Glean as well](https://github.com/mozilla/glean/blob/main/glean-core/src/lib.rs#L788-L789). This should even avoid the overhead of calling into the underlying log implementation, because filtering is done before that.

mhammond:
While @badboy has some great workarounds, I don't see any reason we wouldn't take a patch for a feature here. That might also dovetail into a future where people start asking how to use `tracing` or similar instead of `log`

mhammond:
I think we've removed it entirely!

PR #2211, opened 2024-08-09
2210: Tracing printouts for FFI calls

bendk:
Added the `uniffi:ffi-trace` feature.  If enabled, UniFFI will printout tracing-level logs for FFI calls.  I hooked up the futures code to this, in the future we could also log more stuff.  Lowering/lifting/cloning arcs is the first example that comes to mind.

To see tracing, use the `--features=ffi-trace` when running a tests, for example:
  - `cd fixtures/futures/src`
  - `cargo test --features=ffi-trace`

Here's an example of the tracing printouts when running tests from the futures fixtures:

```
rust_future_new: Handle(94402427620656)
rust_future_poll: Handle(94402427620656)
RustFuture::wake called
rust_future_poll: Handle(94402427620656)
RustFuture::poll is ready (canceled: false)
rust_future_complete: Handle(94402427620656)
rust_future_free: Handle(94402427620656)
```

PR #2227, opened 2024-08-30
Make log crate dependency and logging an optional feature

timboudreau:
Addresses #2224 - even disabled logging can have overhead depending with some backends, and in very high frequency calls, generates unacceptable overhead and log-spam.

Those tests that can run without `kotlinc` installed pass.

I cannot verify that it is problem-free until I backport it to `0.26.1` as using `0.28.1` with my code fails at build time, with or without the patch, as follows:
```
thread 'main' panicked at /Users/tim/.cargo/git/checkouts/uniffi-rs-cc388e82c08ac837/ddc23a3/uniffi/src/lib.rs:30:21:
called `Result::unwrap()` on an `Err` value: Failed to extract data from archive member `curve.curve.e27b66861dd0975a-cgu.0.rcgu.o`

Caused by:
    Constructor return type must be Arc<Self>
```

Note this patch **entirely** removes the `log` dependency - both call logging and panic logging.  If that seems like overkill, it could probably be split into two flags, one to remove logging entirely, and one just to remove FFI-call logging.

### DIFF
--- a/uniffi/Cargo.toml
+++ b/uniffi/Cargo.toml
@@ -37,6 +37,8 @@ build = [ "dep:uniffi_build" ]
 bindgen = ["dep:uniffi_bindgen"]
 cargo-metadata = ["dep:cargo_metadata", "uniffi_bindgen?/cargo-metadata"]
 
+log = ["uniffi_macros/log", "uniffi_core/log"]
+
 # Support for `uniffi_bindgen_main()`. Enable this feature for your
 # `uniffi-bindgen` binaries.
 cli = [ "bindgen", "dep:clap", "dep:camino" ]
--- a/uniffi_core/Cargo.toml
+++ b/uniffi_core/Cargo.toml
@@ -16,7 +16,7 @@ readme = "../README.md"
 anyhow = "1"
 async-compat = { version = "0.2.1", optional = true }
 bytes = "1.3"
-log = "0.4"
+log = { version = "0.4", optional = true }
 once_cell = "1.10.0"
 # Regular dependencies
 paste = "1.0"
@@ -24,6 +24,8 @@ static_assertions = "1.1.0"
 
 [features]
 default = []
+
+log = ["dep:log"]
 
 # Enable support for Tokio's futures.
 # This must still be opted into on a per-function basis using `#[uniffi::export(async_runtime = "tokio")]`.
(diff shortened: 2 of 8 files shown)


bendk:
Completely unrelated to this patch, but I believe your build issue is caused by mixing UniFFI versions.  That [error message](https://github.com/mozilla/uniffi-rs/blob/760660b3f1d70e54363836bf371a85fa5d7913d4/uniffi_meta/src/reader.rs#L262) was in 0.26.x and 0.27.x, but removed in [0.28.x](https://github.com/mozilla/uniffi-rs/blob/2296746b335ff27f49a669a49e4fbbfca9ef9214/uniffi_meta/src/reader.rs#L262).  My guess is that you're using 0.26.x to generate your bindings.

bendk:
This feels similar in spirit to my PR to add the `ffi-trace` feature https://github.com/mozilla/uniffi-rs/pull/2211.  What do others think about putting these logging statements behind that feature flag?

In general, my feeling is that these kind of logging statements are only really useful for debugging failures when writing scaffolding/bindings code.  There's not a lot of point in logging them when a library is being used in some consumer application.

timboudreau:
Would appreciate it if any obstacles to merging this could be brought up, and if not, it could be merged.

I wound up ripping uniffi out of the codebase I created this patch to solve problems in - the marshalling overhead was significant compared with calling the same code via the C ABI, and removing it cut 5 minutes off my build times (parts that touched uniffi get built multiple times with and without macro generated metadata used to generate C/C++/Swift code).

I'm happy to follow up and get this patch to the finish line, if there are any objections or suggested improvements, though - I will probably use uniffi again in the future.

badboy:
> Would appreciate it if any obstacles to merging this could be brought up, and if not, it could be merged.

There was a weekend and a holiday in between. Sometimes things just take a bit of extra time. I'll see that we can get this forward this week.

badboy:
@timboudreau after talking with the others yesterday we decided to go ahead. Could you add a changelog entry?
Just add a new `###  What's new?` heading under here: https://github.com/mozilla/uniffi-rs/blob/0bddc32a2d1c725ca3b75bfe8e44f0354f30d55a/CHANGELOG.md#L9

Then add a line a la "- Added a new default-disabled feature `log` that enables some internal debugg logging" (or similar if you find a better phrasing)

bendk:
I think  #2275 covers this one and I'm going to close.  Please re-open if you have use-cases that it doesn't meet.

PR #2230, opened 2024-09-02
Test callback lifetime

arg0d:
Added test to ensure callback references are counted correctly. We are just now discovering that callback references are counted incorrect in v0.25.0 for Kotlin, C#, and Go.


bendk (approved):
Looks like a nice test to me, let's merge this one.

PR #2235, opened 2024-09-09
Generate a single modulemap in library mode

bendk:
I don't believe multiple modulemaps are useful at all.  At Mozilla, we end up deleting all the generated module maps and hand-writing one of our own.  Once this lands, I think we will be able to delete that code and use the single generated modulemap.


uniffi_bindgen/src/bindings/swift/test.rs
```diff
@@ -153,25 +151,9 @@ impl GeneratedSources {
...
-                    "{}",
-                    modulemap_glob
-                        .into_iter()
-                        .map(|path| Ok(read_to_string(path)?))
-                        .collect::<Result<Vec<String>>>()?
-                        .join("\n")
-                )?;
-                path
-            }
+            n => bail!("{n} modulemap files found in {out_dir}"),
```
bendk, line 156: I'm not a Swift expert, but my hope is that if the tests pass after this change then the single-modulemap should also work for real-world consumption.

uniffi_bindgen/src/lib.rs
```diff
@@ -124,11 +124,26 @@ use uniffi_meta::Type;
...
 pub struct GenerationSettings {
     pub out_dir: Utf8PathBuf,
     pub try_format_code: bool,
     pub cdylib: Option<String>,
+    pub mode: GenerationMode,
+}
+
+#[derive(Debug)]
+pub enum GenerationMode {
+    // Generating code for a single component
```
gruberb, line 137: Can we change these to docstrings?

gruberb (commented):
I tested it locally and it works (at least to generate the swift-docs, now we can remove the hard coded `modulemap` and use the generated one with this PR).

Two things:
* I would love to see an example command in the docs (how to use it)
* Using docstrings instead of comments

bendk:
I changed the comments to docstrings and added some more documentation.

I don't think this documentation explains things completely adequately, but hopefully it helps.  I think we may need a larger documentation overhaul to make this make sense.  The Xcode section could definitely use a refresh to describe how things work with proc-macros and library mode, but I'm not sure what's needed there.

gruberb (approved):

PR #2240, opened 2024-09-19
Fix Python error name regression

bendk:
I hit this one when trying to upgrade application-services to 0.28.1.  It's kind of surprising that this is the first we're hearing of it, but I guess the Python bindings users are a minority.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/gen_python/error.rs
+++ b/uniffi_bindgen/src/bindings/python/gen_python/error.rs
@@ -22,7 +22,7 @@ impl CodeType for ErrorCodeType {
     }
 
     fn canonical_name(&self) -> String {
-        format!("Type{}", self.id)
+        format!("Type{}", self.type_label())
     }
 
     fn literal(&self, _literal: &Literal) -> String {
--- a/uniffi_bindgen/src/interface/visit_mut.rs
+++ b/uniffi_bindgen/src/interface/visit_mut.rs
@@ -131,6 +131,15 @@ impl ComponentInterface {
                 }
             }
         }
+
+        self.errors = self
+            .errors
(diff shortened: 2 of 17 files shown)


uniffi_bindgen/src/interface/visit_mut.rs
```diff
@@ -131,6 +131,14 @@ impl ComponentInterface {
...
             }
         }
+
+        self.errors = self.errors
+            .drain()
+            .map(|mut name| {
+                visitor.visit_error_name(&mut name);
+                name
+            })
+            .collect()
```
bendk, line 141: This is the basic issue, this is a somewhat hacky field that stores error type names when we see them in "[Throws=]" UDL attributes.  I think that proc-macros might have handled this correctly, but I'm not sure.

uniffi_bindgen/src/bindings/python/gen_python/error.rs
```diff
@@ -22,7 +22,7 @@ impl CodeType for ErrorCodeType {
     }
 
     fn canonical_name(&self) -> String {
-        format!("Type{}", self.id)
+        format!("Type{}", self.type_label())
```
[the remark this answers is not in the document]
bendk: Maybe I only needed to modify `ErrorCodeType::canonical_name`, but it seemed safer to change all of the class-like types to use the converted name.

gruberb (approved):
👍 LGTM

uniffi_bindgen/src/interface/visit_mut.rs
gruberb: I like `drain` in combination with `collect`. I think I could have used something similar here as well!

PR #2242, opened 2024-09-20
Re-export library mode items in uniffi

bendk:
I want to use this in a CLI tool that I'm building.  The uniffi crate currently re-exports `generate_bindings`, but not `library_mode::generate_bindings`.


gruberb (approved):
👍 LGTM

PR #2246, opened 2024-09-25
Fix for Mach-O symbol extraction

bendk:
The `n_sect` field is 1-based with `0` meaning `NO_SECT` meaning the symbol is not present in any section in the library.  This change fixes the index mismatch.

Reference: https://github.com/aidansteele/osx-abi-macho-file-format-reference#nlist_64, 

The code was mostly working before because of the way we calculate the offset (https://github.com/mozilla/uniffi-rs/blob/2e3b59cf6dbcb46983028b08afc6e0ee51183f2e/uniffi_bindgen/src/macro_metadata/extract.rs#L106).  Because of this, even if we were off by one when picking the section, things would still work as long as the two sections involved lined up correctly.

Thanks to 0c0w3 for pointing this out and sending me the patch.

### DIFF
--- a/uniffi_bindgen/src/macro_metadata/extract.rs
+++ b/uniffi_bindgen/src/macro_metadata/extract.rs
@@ -99,8 +99,13 @@ pub fn extract_from_macho(macho: MachO<'_>, file_data: &[u8]) -> anyhow::Result<
         //   - Has type=N_SECT (it's regular data as opposed to something like
         //     "undefined" or "indirect")
         //   - Has a metadata symbol name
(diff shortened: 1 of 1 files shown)


bendk:
I tested this by printing out the symbol name and the extracted metadata.  Before, we would get index errors for some of the symbols, now we successfully extract metadata and it lines up with the symbol name.  For example `_UNIFFI_META_NAMESPACE_SYNC` was extracted as

```
Namespace(NamespaceMetadata { crate_name: "sync15", name: "sync15" })
```

0c0w3 (approved):
Looks familiar :-)

badboy (approved):

PR #2248, opened 2024-09-27
uniffi-bindgen-swift

bendk:
This is a `uniffi-bindgen` variant dedicated to Swift code, with options specialized for Swift.  I'm currently thinking that we should consider doing this for all languages.  In addition to allowing specialized options, it also makes the in-tree bindings less special-cased compared to external bindings.

Reverted https://github.com/mozilla/uniffi-rs/pull/2235, which was an attempt to do something similar.  I like this approach better because it supports more features, like xcframework compatible module maps.  It's also a non-breaking change, so we can release it in a `0.28.2`.


uniffi_bindgen/src/bindings/swift/mod.rs
```diff
@@ -66,15 +67,10 @@ impl BindingGenerator for SwiftBindingGenerator {
 
     fn update_component_configs(
         &self,
-        settings: &GenerationSettings,
+        _settings: &GenerationSettings,
```
gruberb, line 70: I personally would rather see changing the trait to `Option<&GenerationSettings>` instead.

uniffi/src/cli/swift.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// outside of this workspace.
+    /// This can be used in environments when all types are in the namespace and fetching
+    /// all sub-dependencies causes obscure platform specific problems.
+    #[clap(long)]
+    metadata_no_deps: bool,
+}
+
+#[derive(Debug, Args)]
+#[group(required = true, multiple = true)]
+struct Kinds {
```
gruberb, line 40: Would it be helpful to have a `default`? Like "building `.Xcframework`" could be the default. So we can set the `bool` values accordingly.

docs/manual/src/swift/configuration.md
```diff
@@ -9,11 +9,11 @@ more likely to change than other configurations.
...
 | `module_name`                       | `{namespace}`[^1]        | The name of the Swift module containing the high-level foreign-language bindings.                                                                                  |
 | `ffi_module_name`                   | `{module_name}FFI`       | The name of the lower-level C module containing the FFI declarations.                                                                                              |
 | `ffi_module_filename`               | `{ffi_module_name}`      | The filename stem for the lower-level C module containing the FFI declarations.                                                                                    |
-| `generate_module_map`               | `true`                   | Whether to generate a `.modulemap` file for the lower-level C module with FFI declarations.  Not valid when using `--library`, in library-mode the modulemap is always generated.
+| `generate_module_map`               | `true`                   | Whether to generate a `.modulemap` file for the lower-level C module with FFI declarations. (ignored by `uniffi-bindgen-swift`)                            |
```
gruberb, line 16: Can we add an example command for both a Swift and a `.Xcframework`?

gruberb (approved):
Just a couple of comments. I really like the Swift focused options, I think this makes it very clear what we can (and have to do) in each scenario!

docs/manual/src/swift/configuration.md
bendk: I added a bunch to `docs/manual/src/swift/uniffi-bindgen-swift.md`.

uniffi/src/cli/swift.rs
bendk: A default sounds good, but I'm not sure what default you're suggesting.  Is it `--xcframework` is the default and `--no-xcframework` would disable it?  Or is it that generating one these kinds of sources is the default?

uniffi_bindgen/src/bindings/swift/mod.rs
bendk: This is implemented for all the bindings though.  Some bindings, like Python, currently need the `GenerationSettings` to be passed.

gruberb: Yeah, we would need to change a lot of files.

uniffi/src/cli/swift.rs
gruberb: I would say we could add the default to always be `xcframework`, which I think will be most of the time the case. We need Swift specific settings at the moment just for the documentation. So we could shorten the command when building for `xcframework` and don't have to pass any parameters.

bendk: After thinking about this a bit more, I'm not so sure.  `--no-xcframework` seems a bit awkward as a flag.  It's harder to describe since it's a negative, and it also sort-of implies that if you don't have it then you're building an xcframework somehow.  Something like `--not-for-xcframework` is more precise and easier to describe, but much harder to type out.

How about we leave it as-is for now, we can always change the flag later.

mhammond:
> I'm currently thinking that we should consider doing this for all languages

I think later we should try and push the implementation down to the bindings. If we treat ourselves more like external bindings, instead of top-level `uniffi::uniffi_bindgen_main` variants, we require an explicit dep on, say, `uniffi_bindgen::swift::uniffi_bindgen_main` and a `uniffi-bindgen-swift` binary?

ie, let `uniffi_bindgen::swift`, `uniffi_bindgen::python` etc all use clap and have their own cli driver. `uniffi::cli` becomes helper functions and b/w compat - but the "new" way to generate python bindings is `uniffi-bindgen-python`.

This means `uniffi_bindgen` takes the dep for `clap` - but in practice, the dep is already there. And later, if we move the bindings to their own crate, (ie,`uniffi_bindgen_swift::uniffi_bindgen_main`) - `uniffi_bindgen` then no longer needs clap and becomes more of an "internal" crate.

WDYT? Does that make any sense for gecko @bendk?

mhammond:
> Gecko is very similar since we write our own CLI there. I definitely would like it if uniffi-bindgen-python and uniffi-bindgen-gecko-js were on similar footing.

Yeah exactly - which implies `uniffi::` can't know about python etc - and the only way I can see for that to work is to have no cli in `uniffi`?

bendk:
> Yeah exactly - which implies `uniffi::` can't know about python etc - and the only way I can see for that to work is to have no cli in `uniffi`?

Good point.  You've convinced me that we should move the clap code into `uniffi_bindgen`.

PR #2249, opened 2024-10-01
Hide internal classes generated docs

bendk:
Some of these could be made internal, but most could not.  The reason is that UniFFI code needs to import them from other modules in order to support external types.  This change mostly adds docstrings/annotations to hide them from the generated docs.


badboy:
image

PR against the wrong branch.
Looks better against `main`: https://github.com/mozilla/uniffi-rs/compare/main...bendk:uniffi-rs:push-tnsvklnxuxkw

bendk:
Thanks for catching that, it looks much better now.

gruberb (approved):

PR #2265, opened 2024-10-11
Potential approach to support external errors

Hinton:
This is a bit of a stab in the dark attempt at supporting external errors in uniffi.

It currently seems to behave well for swift but before I continue diving into the uniffi internals I would appreciate some guidance if this is the appropriate way to address this as the internals of uniffi is a large mystery to me.

Remaining tasks:
- Support kotlin
- Fix any broken tests.
- Write new tests


mhammond:
I'm mildly surprised we don't support this, but it looks like we do not! A very quick look implies this seems sane for Swift as it's what we did for other uses for external types - however, we'd obviously want tests here. I'd suggest maybe reusing the ext-types fixture - eg, there's a `GuidError` in the guid crate which you might be able to have a top-level function throw.

Hinton:
I wrote up a test case for the proc-macro which works as expected. Unfortunately I ran into a few issues with udl. With the latest commit I get the following error message:

```
error[E0277]: the trait bound `GuidError: LowerError<UniFfiTag>` is not satisfied
   --> /target/debug/build/uniffi-fixture-ext-types-dac13a3e9f364522/out/ext-types-lib.uniffi.rs:253:1
    |
253 | #[::uniffi::export_for_udl]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | |
    | the trait `LowerError<UniFfiTag>` is not implemented for `GuidError`, which is required by `Result<(), GuidError>: LowerReturn<UniFfiTag>`
    | in this procedural macro expansion
    |
   ::: /uniffi_macros/src/lib.rs:203:1
    |
203 | pub fn export_for_udl(attrs: TokenStream, input: TokenStream) -> TokenStream {
    | ---------------------------------------------------------------------------- in this expansion of `#[::uniffi::export_for_udl]`
    |
    = help: the trait `LowerError<ext_types_custom::UniFfiTag>` is implemented for `GuidError`
    = help: for that trait implementation, expected `ext_types_custom::UniFfiTag`, found `UniFfiTag`
    = note: required for `Result<(), GuidError>` to implement `LowerReturn<UniFfiTag>
```

My suspicion from debugging the expanded udl file is that it's missing a `ffi_converter_forward` for `GuidError`. Unfortunately I'm a bit outside my knowledge depth, but I suspect we need to wire up some logic in `uniffi_udl/src/attributes.rs` to ensure external errors are appropriately imported?

---

I also experimented slightly with Kotlin which in some ways seems to also be quite straightforward as we _should_ only need to add an import to the Exception. Unfortunately my initial attempt of adding a `add_import` function call inside `uniffi_bindgen/src/bindings/kotlin/templates/macros.kt` failed with ```no method named `add_import` found for reference `&KotlinWrapper<'a>` in the current scope```.

Curious if you have any quick thoughts on a better approach.

mhammond:
> My suspicion from debugging the expanded udl file is that it's missing a `ffi_converter_forward` for `GuidError`

Yeah - I suspect you need `::uniffi::ffi_converter_forward!(GuidError, ext_types_custom::UniFfiTag, crate::UniFfiTag);` where it's imported.

> `uniffi_bindgen/src/bindings/kotlin/templates/macros.kt` failed with `` no method named `add_import` found for reference `&KotlinWrapper<'a>` in the current scope ``.

The templating system is tricky, but that should work if you move the call out of macros.kt

A quick look though shows you seem to be on the right track!

Hinton:
Alright, solving udl for swift turned into a bigger hassle than I expected. It seems to work but it's mostly based on guess work.

I implemented a custom forwarder since we only want the `LiftError` implementation, which also necessitated a custom UDL, `ExternalError`. Does this seem reasonable, or am I completely off the track here.

uniffi_meta/src/types.rs
```diff
@@ -57,6 +57,7 @@ pub enum ExternalKind {
     Trait,
     // Either a record or enum
     DataClass,
+    Error,
```
mhammond, line 60: I don't think this is going to work - because errors can be enums of interfaces, I'd think that just `ExternalKind::DataClass` and `ExternalKind::Interfaces` would be OK. Why did you find it necessary to add this new discriminant?

Hinton: I needed a way to differentiate errors from other types in order to provide the appropriate forward macro, https://github.com/mozilla/uniffi-rs/pull/2265/files#diff-6cbadee39d35e495122970278e12fc509f5243d4e3279a911112f69e798fe247R11-R12.

mhammond: It's possible you might be able to use `ci.is_name_used_as_error()`, eg, https://github.com/mozilla/uniffi-rs/blob/1c8dd50de3413c03b7526aa9cab739049b64d331/uniffi_bindgen/src/scaffolding/templates/scaffolding_template.rs#L19

Hinton: 👍 , that seems to work. I pushed an update to this effect.

mhammond:
I think this is mostly working now in #2390 - no kotlin yet, but the key parts are now all in place for that to work once someone works out how to refactor things.

Hinton:
Great thanks @mhammond. We're still blocked by kotlin but having half the functionality in is better than nothing.

Hoping I can get some time to pick this back up unless someone gets it done before me.

PR #2267, opened 2024-10-11
Python: Fix class hierarchy generated for traits. Fixes #2264.

### LINKED ISSUE #2264: Python bindings generate invalid class hierarchy for foreign traits
Best demonstrated with this patch:
```
--- a/fixtures/proc-macro/tests/bindings/test_proc_macro.py
+++ b/fixtures/proc-macro/tests/bindings/test_proc_macro.py
@@ -25,11 +25,13 @@
 assert(rename_test())
 
 trait_impl = obj.get_trait(None)
+assert isinstance(trait_impl, Trait)
 assert trait_impl.concat_strings("foo", "bar") == "foobar"
 assert obj.get_trait(trait_impl).concat_strings("foo", "bar") == "foobar"
 assert concat_strings_by_ref(trait_impl, "foo", "bar") == "foobar"
 
 trait_impl2 = obj.get_trait_with_foreign(None)
+assert isinstance(trait_impl2, TraitWithForeign)
 assert trait_impl2.name() == "RustTraitImpl"
 assert obj.get_trait_with_foreign(trait_impl2).name() == "RustTraitImpl"
```

which causes the second addition to fail: `TypeError: Instance and class checks can only be used with @runtime_checkable protocols`

To understand the problem, first consider how traits without `with-foreign` are generated:
```
class TraitProtocol(typing.Protocol):
    ...

class Trait:
    _pointer: ctypes.c_void_p
    ...
```

which seems correct - `Trait` (the name of the object) *does not* derive from `typing.Protocol` - protocols are used only by type checkers rather than at runtime.

However, consider foreign traits:
```
class TraitWithForeign(typing.Protocol):
    ...

class TraitWithForeignImpl:
    _pointer: ctypes.c_void_p

```

here, `TraitWithForeign` (the name of the object) *does* derive from `typing.Protocol` - hence the `is_instance` check fails.

While it would be possible to allow this to work by opting-in to runtime checks for these protocols, I think we should do the "correct" thing and arrange to generate a class hierarchy that does what the typing docs recommend.

Part of explaining why we do this is apparently [this comment](https://github.com/mozilla/uniffi-rs/blob/0b05ee1ab599445388956de1135cc80cfdc929fb/uniffi_bindgen/src/bindings/python/gen_python/mod.rs#L432-L434) - ISTM that it might be fine to relax this a little, and just support `typing.Protocol` in all cases.

mhammond:


bendk (approved):
Looks good.  Does this affect `lower()` like you mentioned in that issue?  I wrote that comment, but I can't remember exactly what it means.

mhammond:
> Does this affect `lower()` like you mentioned in that issue? I wrote that comment, but I can't remember exactly what it means.

I don't think it does - this adds a new class which doesn't derive from the protocol, but the other concrete classes still exist and are still referenced in the lower calls - it's just that the abstract class they derive from has changed.

PR #2270, opened 2024-10-14
Remove Python var_name and class_name filters

mhammond:
A followup to where I accidentally used these filters. It turns out the bindings were slightly inconsistent in how variant names were used. This fixes that and as a result, enums now have 2 `is_*` methods - the `is_VARIANT()` we thought and commented that we wrote, and the `is_variant()` we actually wrote.


bendk (approved):
Looks good, both the new code and the workaround for the confusion in the methods.

PR #2271, opened 2024-10-14
Tutorial tweaks making procmacro docs a little clearer

mhammond:
Some improvements I came up with after re-reading the tutorial from the start.


docs/manual/src/udl/namespace.md
```diff
@@ -1,5 +1,7 @@
 # Namespace
 
+Every crate has a a UniFFI "namespace". This is the name exposed to the foreign bindings, typically as a module.
```
badboy, line 3:
```suggestion
Every crate has a UniFFI "namespace". This is the name exposed to the foreign bindings, typically as a module.
```

docs/manual/src/tutorial/udl_file.md
```diff
@@ -1,10 +1,19 @@
 # Describing the interface
 
 There are two ways of describing your interface:
-1 - with a UDL file (a type of IDL, Interface Definition Language);
-2 - with [Rust procmacros](../proc_macro/index.md), similarly to how `wasm-bindgen` work. This avoids repeating your interface definitions in a separate file. Unfortunately, our docs aren't quite as good for that yet though.
 
-## The UDL File
+## Procmacros
+
+[Rust procmacros](../proc_macro/index.md) can describe your interface - you'd define the function as:
```
badboy, line 7: Nit-pick, but we use `proc macro` in other places. 

```suggestion
## Proc macros

[Rust proc macros](../proc_macro/index.md) can describe your interface - you'd define the function as:
```

But I guess either is fine, so feel free to ignore.

docs/manual/src/udl/namespace.md
```diff
@@ -8,7 +10,10 @@ namespace math {
 };
 ```
 
-It serves multiple purposes:
-- It identifies the name of the generated Rust scaffolding file `<namespace>.uniffi.rs`.
-- It identifies the package name of the generated foreign-language bindings (e.g. `uniffi.<namespace>` in Kotlin)
-- It also contains all [top-level *functions*](./functions.md) that get exposed to foreign-language bindings.
+which might be used as `from math import exp`/`import math`/`import omg.wtf.math.exp` etc.
+
+Proc-macros use the crate name as the namespace by default, but it can be specified with
```
badboy, line 15: heh, and here with a dash. We should at least be consistent.

badboy (approved):

docs/manual/src/udl/namespace.md
mhammond: yeah, thanks!

PR #2280, opened 2024-10-22
Update template pattern matching to leverage rinja.

mhammond:
Mostly using `{ .. }`, `if let`, and `when Some with (x)` -> `when Some(x)`

I did this really just to play with some of the features and pattern matching always made me grumpy.

On the way I made some errors in the templates, and the error messages from rinja are *vastly improved* over askama - not only having very precise details about the error but often suggestions for how to fix (eg, with mismatched tags) - so thanks @GuillaumeGomez!


bendk (approved):
Nice!  The `if let` changes are my favorite.

Issue #2285, opened 2024-10-24
Error: Failed to extract data from archive member 


jhugman:
I'm trying to use `library_mode` to `generate_bindings` from a reasonably large crate, the [matrix-sdk-ffi bindings](https://github.com/matrix-org/matrix-rust-sdk/tree/main/bindings/matrix-sdk-ffi).

Also relevant: I've bumped `uniffi` to version `0.28.2`.

Using `cargo ndk`, I'm building the staticlib version for `aarc64-linux-android`, in release mode only.

```sh
  cargo ndk \
  --manifest-path ./bindings/matrix-sdk-ffi/Cargo.toml \
  --target arm64-v8a \
  --platform 21 --no-strip -- \
  build --release
```

Then using `print-repr` to look at the metadata.

```sh
cargo run \
  --manifest-path uniffi-bindgen/Cargo.toml \
  print-repr \
  target/aarch64-linux-android/release/libmatrix_sdk_ffi.a 
```

Output is:
```
    Finished `dev` profile [unoptimized] target(s) in 0.13s
     Running `target/debug/uniffi-bindgen print-repr target/aarch64-linux-android/release/libmatrix_sdk_ffi.a`
Failed to extract data from archive member `matrix_sdk_ffi.matrix_sdk_ffi.bf8c76dbb4572677-cgu.07.rcgu.o`
```

A lightly instrumented uniffi which prints in `extract_item` that a symbol has been recognized, but is an unrecognized metadata code:

```
Metadata symbol UNIFFI_META_MATRIX_SDK_FFI_INTERFACE_ENCRYPTION
thread 'main' panicked at uniffi_meta/src/metadata.rs:84:18:
XXXX UNKNOWN ITEM 88 XXXX
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7cf61ebde7b22796c69757901dd346d0fe70bd97/library/std/src/panicking.rs:647:5
   1: core::panicking::panic_fmt
             at /rustc/7cf61ebde7b22796c69757901dd346d0fe70bd97/library/core/src/panicking.rs:72:14
   2: uniffi_meta::metadata::codes::metadata_as_str
             at ./uniffi_meta/src/metadata.rs:84:18
   3: uniffi_meta::reader::MetadataReader::read_metadata
             at ./uniffi_meta/src/reader.rs:39:33
   4: uniffi_meta::reader::read_metadata
             at ./uniffi_meta/src/reader.rs:10:5
[...]

PR #2286, opened 2024-10-24
Fix metadata extraction for large ELF files (#2285)

bendk:
Symbol table entries store a 16-bit index for their section header.  If the index overflows 16-bits then it's stored in a separate section instead.

References:

- https://refspecs.linuxbase.org/elf/gabi4+/ch4.symtab.html
- https://docs.oracle.com/cd/E19683-01/817-3677/chapter6-94076/index.html


mhammond (approved):
awesome! changelog?

jhugman:
Amazing. Thanks for the fast turn around.

Issue #2279, opened 2024-10-17
[Swift 6] Compilation issues when migrating to Swift 6


trevoranderson:
Similar to https://github.com/mozilla/uniffi-rs/issues/2274 but adding a few more from my project. This is all tested by setting `SWIFT_STRICT_CONCURRENCY = complete` in the project's build settings in XCode. Most of the issues are on the class that Swift passes into Rust to get notified about certain events. I believe it's called `Foreign traits` in the docs.

# Foreign Trait Swift 6 concurrency issues
```rust
private enum UniffiCallbackInterfaceRustEventHandler {
    // Create the VTable using a series of closures.
    // Swift automatically converts these into C callback functions.
    static var vtable: UniffiVTableCallbackInterfaceRustEventHandler = .init(
```
gives the error
```
Static property 'vtable' is not concurrency-safe because it is nonisolated global shared mutable state; this is an error in the Swift 6 language mode
```
another location
```rust
public struct FfiConverterTypeRustEventHandler: FfiConverter {
    fileprivate static var handleMap = UniffiHandleMap<RustEventHandler>()
```
gives the error
```
Static property 'handleMap' is not concurrency-safe because it is nonisolated global shared mutable state; this is an error in the Swift 6 language mode
```
# Initialization Result Swift 6 concurrency issues
This last one is perhaps more important because it seems like it will affect all Rust+Swift users. The comment seems to disagree with the compiler. (This one might be fixable by switching `var` to `let`)

```rust
// Use a global variable to perform the versioning checks. Swift ensures that
// the code inside is only computed once.
private var initializationResult: InitializationResult = {
```
produces error
```
Var 'initializationResult' is not concurrency-safe because it is nonisolated global shared mutable state; this is an error in the Swift 6 language mode
```

# Fixing
In all cases, the trivial fix is to mark the relevant locations as `nonisolated(unsafe) var`, which should immediately solve the warnings and allow users to migrate to Swift6 while a more "Swifty" solution is designed. The `InitializationResult` one looks like just a one-liner bug

CyonAlexRDX:
Relevant: https://github.com/mozilla/uniffi-rs/issues/2046

@trevoranderson I agree that it would be good with the `nonisolated(unsafe)` **as a temporary hack**. 

[...]

PR #2290, opened 2024-10-30
render global variable as a constant

heckj:
more correctly handles swift6 data race concerns

fixes the specific issue in #2279, but may not be all of the changes needed or desired for swift6 compatibility


Sajjon (approved):
Ah it was never mutated, good!

mhammond (approved):

PR #2292, opened 2024-10-31
fix: move all dependencies to workspace dependencies

gruberb:
Started with one and ended up with moving every dependency to a `workspace.dependency`. I think it's easier to manage version upgrades and get a view on which crates the whole project is relying on. 

But also, not sure what should be a workspace dependency and what shouldn't. We can argue just dependencies in > 1 workspace member should be moved to be a `workspace.dependency`. Happy to discuss, or drop it all together.


mhammond:
There were some interesting comments in #1859. I also previously struck that vendoring with workspace deps didn't quite work as I expected, so it's probably worth doing a test vendor into m-c to see how it all plays out.

PR #2294, opened 2024-10-31
Fix some more Swift concurrency issues

bendk:
Converted some more vars to lets.  Swift still complains about async calls.  I believe the next step would be to make all UniFFI objects conform to `Sendable`.

See #2279.


mhammond (approved):

bendk:
These seem safe enough to me, I'm going to go ahead and merge them.

PR #2296, opened 2024-11-01
Move the bindings into a new uniffi_bindings crate.

mhammond:
The idea is to move the templates and special-casing of the builtin bindings out from uniffi_bindgen.

Eventually, we'd drop all bindings generation from the top-level `uniffi` crate - you depend and use the specific bindings crate, just like consumers of 3rd-party languages are forced to do now. It helps keep some degree of separation between the bindings and the bindgen crate.

OTOH - it's yet another crate to release etc. Is it actually worth it?  WDYT?


mhammond:
Assuming the split makes sense - maybe "uniffi_bindgen" is the correct name for the bindings, and it's the "build a component interface" part that should get a new name?

PR #2297, opened 2024-11-01
Kotlin classes can inherit from traits.

mhammond:
Builds on #2204 which landed the metadata and Python support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@
 
 - Fixed bug in metadata extraction with large ELF files.
 
+### What's new?
+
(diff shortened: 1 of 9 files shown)


bendk (approved):
Nice, this looks great.  I'm not sure why that keywords fixture is failing, maybe something needs to be quoted?

mhammond:
turned out to be a few minor issues with callback interfaces. Then I discovered the trait and struct for that fixture both had the same method with different signatures 🙃 That's an interesting edge-case for the future, but I just avoided doing that for now. Fixed all that and a couple of other tweaks but all of which were sane and small enough that I didn't think it needed a new review.

PR #2308, opened 2024-11-12
CLI to print/diff generated sources

bendk:
This allows you to save the generated sources for some fixture/example, make some changes to the foreign bindings code, then print out a diff. I've been using this pattern when writing code, let's make it easier to use.

This came out of talking with @gruberb about ways to make it easier to know how changes to the code are affecting the bindings.


gruberb (approved):
What a great idea, I like it :)

bendk:
I think the `peek` and `diff` commands from #2333 is a better version of this, let's close this one.

Issue #2310, opened 2024-11-13
Swift: invalid redeclaration of 'UniffiForeignFutureTask' 


QuentinPerez:
👋 Hey there!

Looks like there’s a compilation issue with Swift when defining more than one async_trait. To help track it down, a small commit was made to reproduce the problem [here](https://github.com/QuentinPerez/uniffi-rs/commit/3bd09bcd94fb17ad187d948ffd346a5101e2652a). You can see it in action by running cargo test -p uniffi-fixture-multiple-definitions.

It turns out that defining this code only once seems to fix things:

```swift
// Protocol for tasks that handle foreign futures.
//
// Defining a protocol allows all tasks to be stored in the same handle map, 
// which isn’t doable with the task object itself because of its generic parameters.
protocol UniffiForeignFutureTask {
    func cancel()
}

extension Task: UniffiForeignFutureTask {}
```

mhammond:
Fixed by #2311

PR #2311, opened 2024-11-13
Fix(swift): invalid redeclaration of 'UniffiForeignFutureTask'

QuentinPerez:
see #2310


bendk (commented):
Looks good, but can you also include the test fixture you ran?  That way we can make sure the fix sticks.  Maybe it could be named something like `futures-multi-crate` and it can be the place we test async issues when multiple crates are involved.

BTW, I think there's another issue here.  What if 1 crate defines multiple async traits?  That can be for another PR though.

QuentinPerez:
closed by https://github.com/mozilla/uniffi-rs/pull/2341

bendk:
Oops, I thought I recognized that code.  I guess we should have preferred this one since the new PR didn't have tests either.  In any case, thanks for the code -- even if it wasn't merged.

QuentinPerez:
My bad, I had not time to add the tests, I was planning to do it during my christmas holidays 🙂

PR #2316, opened 2024-11-15
Clean up filters and error handling in the bindings.

mhammond:
* Using anyhow in templates works fine, so start doing that for literals, replacing a number of panics with errors.

* Filters can't return anyhow, so add better helpers for converting &str and anyhow::Error to rinja::Error.

* Other shared filters weren't used consistently and add no value so are removed.


bendk (approved):
Looks great to me.

I started to suggest maybe defining some traits to make conversion more ergonomic (`.map_rinja_err()` instead of ` .map_err(|e| to_rinja_error(&e))`).  However, I'm not sure if it's worth it and we can add them later if we want.

mhammond:
> I started to suggest maybe defining some traits to make conversion more ergonomic

I kinda thought about that but didn't even try it because I thought it was doomed to fail seeing I was trying to convert between 2 types not in this crate (ie, `anyhow::Error` -> `rinja::Error`) - but agree it would be a nice improvement if it works.

Issue #1988, opened 2024-02-08
`uniffi::custom_newtype!` in crate `a` gives `Lift<UniFfiTag>` is not implemented for `Newtype` in crate `b`


mgeisler:
Hello!

I've been poking around at UniFFI for a few days and I really like what I see: the ability to automatically generate high-level APIs is amazing.

However, I think I've found a small bug. I have two crates, `a` and `b`. The `a` crate defines a newtype which I hope to use from `b`:

_a/src/lib.rs_:

```rust
uniffi::setup_scaffolding!();

uniffi::custom_newtype!(Newtype, u16);

#[derive(Debug)]
pub struct Newtype(u16);
```

_b/src/lib.rs_:

```rust
uniffi::setup_scaffolding!();

#[uniffi::export]
pub fn use_newtype(newtype: a::Newtype) {
    println!("{newtype:?}");
}
```

This fails with

```
error[E0277]: the trait bound `Newtype: Lift<UniFfiTag>` is not satisfied
 --> src/lib.rs:3:1
  |
3 | #[uniffi::export]
  | ^^^^^^^^^^^^^^^^^ the trait `Lift<UniFfiTag>` is not implemented for `Newtype`
  |
  = help: the trait `Lift<a::UniFfiTag>` is implemented for `Newtype`
  = help: for that trait implementation, expected `a::UniFfiTag`, found `UniFfiTag`
  = note: this error originates in the attribute macro `uniffi::export` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `Newtype: Lift<UniFfiTag>` is not satisfied
 --> src/lib.rs:4:29
  |
4 | pub fn use_newtype(newtype: a::Newtype) {
  |                             ^^^^^^^^^^ the trait `Lift<UniFfiTag>` is not implemented for `Newtype`
  |
  = help: the trait `Lift<a::UniFfiTag>` is implemented for `Newtype`
  = help: for that trait implementation, expected `a::UniFfiTag`, found `UniFfiTag`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `b` (lib) due to 2 previous errors
```

It took me a while to spot the difference! :smile: The `Newtype` implements `Lift<a::UniFfiTag>` but not `Lift<b::UniFfiTag>`.

I'm not understanding the details of how the `UniFfiTag` struct is supposed to work, but I hope the above is more clear to someone here.

[...]

PR #2317, opened 2024-11-15
Common system for remote type handling (#1865)

bendk:
Remote types present an issue because we can't implement FFI traits like `Lower` and `Lift` on them directly.  The workaround is to only implement the FFI trait for the current crate.  If another crate wants to use that impl, we need to implement it again by forwarding to the initial implementation.

Historically, UDL and proc-macros have approached this differently which has lead to a lot of extra complexity and confusion.  This change aligns them on a common system:  By default, we implement traits for all crates (a "blanket impl").  However, users can opt-in to only implementing it for the current crate ("a local impl`)

See the doc changes for a description of the new system.

This required a bunch of changes:

- UDL now creates a blanket impl by default. The [Remote] attribute can be used to switch to a local impl.
- Proc-macros get the `#[remote]` attribute macro, which allows users to create a local impl.
- Custom types now create a blanket impl by default.  The `remote` parameter can be used to create a local impl.
- Added the `remote_type!` macro to which handles the case where you want to use a remote type and another crate has already created a local impl.  This creates a local impl for your crate by forwarding to the other impl.  Removed the `use_udl_*` macros, which were a kludge for doing the same thing.

Added a motivating example where `anyhow::Error` is used as an error type.  Changed the error type bound to `Display + Debug` rather than `Error`. For some reason `anyhow::Error` doesn't actually implement error.

One feature that we've lost in defining callback/trait interfaces using traits defined in a remote crate.  I think this is okay because I doubt anyone is using that feature.


CHANGELOG.md
```diff
@@ -13,6 +13,13 @@
   We did this to help fix some edge-cases with custom types wrapping types from other crates (eg, Url).
   See https://mozilla.github.io/uniffi-rs/next/Upgrading.html for help upgrading and https://mozilla.github.io/uniffi-rs/next/udl/custom_types.html for details.
 
+- Handling of remote and external types has changed significantly:
+    - UDL users need to add the `[Remote]` attribute for remote types
+    - The `use_udl_*` macros are no longer needed and have been removed.
+    - To share remote type implementations between crates, use the `use_remote_type` macro.
+    - The UDL syntax for external types in the UDL has been changed.
+    - See https://mozilla.github.io/uniffi-rs/udl/remote_ext_types.html for details
```
mhammond, line 21: This might not work - eg, https://mozilla.github.io/uniffi-rs/udl/ext_types.html -> https://mozilla.github.io/uniffi-rs/0.27/udl/ext_types.html

(there's another copy of this link below)

mhammond (approved):
This is fantastic. It cleans up so much of the messy stuff between UDL and procmacros (eg, fixes #1988?). There are some docs/changelog/upgrading changes we can make before we release, but that's polish.

CHANGELOG.md
bendk: I'm going to add `0.29` to the path, since that's seems like the correct URL as we go forward.  However, that one is also not going to work until we make the 0.29 release.  Is there a better way to do this?

bendk:
> There are some docs/changelog/upgrading changes we can make before we release, but that's polish.
Sounds good to me, let's merge now and follow-up on this later.

Issue #2305, opened 2024-11-09
Remove localized error implementation for objects?


bendk:
The `error-types` fixture is currently failing for me because of a SIGBUS.  I'm pretty sure I'm running into this issue: https://github.com/swiftlang/swift/issues/55963.  I think Fedora recently upgraded to swift `6.0.1`, it was working with the previous version.

Maybe we should disable this feature for objects used as errors?  It seems like somewhat of a niche feature and I'm not the only one who's experienced this bug.  I'm not really sure though, maybe it's more of a me problem and I should just skip that test.

mhammond:
I'm fairly sure I saw firefox-ios using this (or maybe I'm mistaken and it's difficult to test :(

#2319 allows you to turn it off, maybe we could default it off in the fixtures?

PR #2319, opened 2024-11-18
Add a configuration to allow removing `LocalizedError` conformation

crazytonyli:
In Swift, all generated error types conform to `LocalizedError`. The implementation uses a compiler generated string, which is not an appropriate `LocalizedError` implementation.

This PR adds a new config to "uniffi.toml": `error_types_conform_to_localized_error`. When not set, its config value is `true`, which means existing code is not affected by this change.

I looked into adding unit tests, but couldn't find how. Happy to add some though, if anyone can point me to the right direction.

### DIFF
--- a/docs/manual/src/swift/configuration.md
+++ b/docs/manual/src/swift/configuration.md
@@ -13,11 +13,12 @@ more likely to change than other configurations.
 | `module_name`                       | `{namespace}`[^1]        | The name of the Swift module containing the high-level foreign-language bindings.                                                                                  |
 | `ffi_module_name`                   | `{module_name}FFI`       | The name of the lower-level C module containing the FFI declarations.                                                                                              |
 | `ffi_module_filename`               | `{ffi_module_name}`      | The filename stem for the lower-level C module containing the FFI declarations.                                                                                    |
-| `generate_module_map`               | `true`                   | Whether to generate a `.modulemap` file for the lower-level C module with FFI declarations. (ignored by `uniffi-bindgen-swift`)                            | 
+| `generate_module_map`               | `true`                   | Whether to generate a `.modulemap` file for the lower-level C module with FFI declarations. (ignored by `uniffi-bindgen-swift`)                            |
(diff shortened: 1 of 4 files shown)


mhammond:
Thanks! Can you please explain why you'd want to opt-out of this? It should get a mention at https://github.com/mozilla/uniffi-rs/blob/main/docs/manual/src/swift/configuration.md, so maybe briefly explaining the rationale there too would be good.

> error_types_conform_to_localized_error. When not set, its config value is true, which means existing code is not affected by this change.

Could we maybe negate this - use a name where the default being `false` covers existing code?

crazytonyli:
> Can you please explain why you'd want to opt-out of this?

A couple of reasons:
1. The `errorDescription` function is supposed to return ["a specialized error that provides localized messages describing the error and why it occurred."](https://developer.apple.com/documentation/foundation/localizederror). The implemented used here `String(reflecting:)` does not match that API design.

```swift
$ swift repl
Welcome to Apple Swift version 6.0.2 (swiftlang-6.0.2.1.2 clang-1600.0.26.4).
Type :help for assistance.
  1> enum Foo { case bar }
  2> String(reflecting: Foo.bar)
$R0: String = "__lldb_expr_1.Foo.bar"
```

2. Applications that use uniffi-rs may want to add their own `LocalizedError` implementation, because it's a generic type that can be used to display error messages directly to end users. They can't do that if the generated error types already conforms to `LocalizedError`.

crazytonyli:
To be honest, I don't think uniffi-rs should add `LocalizedError` implementations at all. Because it simply can do the _localization_ part properly.

If you have to implement it, Rust `error.to_string()` might be a better choice than Swift `String(reflecting:)`

mhammond:
heh, just re-found #2305. I think we added it because of some assumptions in our consumers, but I'm sure they just want a string to log etc. So yeah, maybe we should look at removing it.

crazytonyli:
Hi @mhammond , any update on this PR?

mhammond (approved):
I dug back a bit and I don't think removing it entirely is reasonable, so this is great, thanks!

mhammond:
Sorry, I mean to talk to som

PR #2325, opened 2024-11-21
uniffi::use_remote_type uses 'crate::name' form to specify types.

mhammond:


bendk (approved):
Wow, that was easy.  I like this new syntax.

PR #2329, opened 2024-11-25
Make all extern C functions public

jhugman:
This PR verifies that the `extern "C"` functions are all `pub`.

In the case of `FfiFunctions` defined by UDL, and callback functions, this required a change.

### DIFF
--- a/uniffi_macros/src/export/callback_interface.rs
+++ b/uniffi_macros/src/export/callback_interface.rs
@@ -52,7 +52,7 @@ pub(super) fn trait_impl(
         let lift_return_type = ffiops::lift_return_type(&sig.return_ty);
         if !sig.is_async {
             quote! {
-                #ident: extern "C" fn(
+                pub #ident: extern "C" fn(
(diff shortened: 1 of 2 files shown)


mhammond:
Hi James. Why is this change necessary?

bendk (approved):
I'm for this one.  To me it's weird to have export a private function as a global symbol -- why should you be able to call it from C, but not Rust?

That said, I'd be okay going the other way as well and making them all non-pub.  The one thing I really like is aligning the UDL and proc-macro behavior.

jhugman:
> Hi James. Why is this change necessary?

Hi Mark!

I'm writing a binding generator. Part of the bindings are in Rust, so need the C ABI to be publicly visible from Rust.

As this is already visible to languages other than Rust (via the ABI), opening this up as part of the Rust API seems pretty low risk. Ben says it better than me, above:

> it's weird to have export a private function as a global symbol -- why should you be able to call it from C, but not Rust?

PR #2333, opened 2024-11-27
Bindings render pipeline

bendk:
**Reviewers**: I like these changes a lot, but I want to explore one more approach for this.  In the new approach, the `PythonIr` and template code will stay essentially the same, but how we handle the pipeline that transforms the metadata into that `PythonIr` will be much different.  If you want to review, you probably want to just focus on the Python changes, skipping `convert_ir.rs`.

---

Rework bindings rendering to work more like a compiler pipeline:

* Start with the metadata
* Transform the metadata into an intermediate representation.  This stage currently goes through the `ComponentInterface`, but if we decide to adopt this approach then we can rework it to be a more straightforward transformation.
* Transform the IR into a language-specific IR.  Walk the tree of IR nodes using the `VisitMut` trait and update the names, format the docstring, add language-specific data, etc.
* Render the language-specific IR into code.

The first commit documents this in some more detail.

One nice feature of this system is that you can inspect the data at any stage of the pipeline, run it through diffs, etc.   I think this will really help both new and experienced devs to understand what's going on during bindings generation.  I documented my plans for this tool, but haven't implemented it yet.

There's a ton of changes here, but I think it's worth it.


bendk:
@lougeniaC64 flagging you for review since you said you'd take a look at the new docs.  Note that these docs are for a proposed new system.  I'd love your take on both the docs themselves and if the overall process makes sense.

uniffi_bindgen/src/bindings/python/gen_python/visit_mut.rs
```diff
@@ -0,0 +1,569 @@
...
+
+/// Implements `VisitMut` to specialize the BindingsIr for Python
+pub struct BindingsIrVisitor {
+    pub config: Config,
+    // The following fields are populated as we walk the tree for VisitMut
+    pub imports: BTreeSet<String>,
+    pub protocols: Vec<Protocol>,
+    pub runtimes: Runtimes,
+    pub exports: Vec<String>,
+}
```
bendk, line 67: This module essentially replaces the `CodeOracle` system (as well as a couple others).  Is it better to have all of the code in one place or try to split it out over multiple modules?  I think I like it all in one place, but I could be convinced to try to split it up.

One thing I definitely like over the old system is that there's a lot less indirection involved: `BindingsIrVisitor` sets data, then the filters read it and that's basically the whole store.

uniffi_bindgen/src/bindings/python/gen_python/visit_mut.rs
```diff
@@ -0,0 +1,569 @@
...
+                | FfiTypeKind::Int64
+                | FfiTypeKind::Handle => "0".to_string(),
+                FfiTypeKind::Float32 | FfiTypeKind::Float64 => "0.0".to_string(),
+                FfiTypeKind::RustArcPtr(_) => "ctypes.c_void_p()".to_string(),
+                FfiTypeKind::RustBuffer(meta) => match meta {
+                    None => "_UniffiRustBuffer.default()".to_string(),
+                    Some(meta) => {
+                        let module_name = self.config.module_for_namespace(&meta.namespace);
+                        format!("{module_name}._UniffiRustBuffer.default()")
+                    }
```
bendk, line 535: This is another thing I like about the new system, we always have access to the `Config` which makes referencing the external rust buffer more straightforward.

Before we didn't want to do this because it means that `ffi_type_name()` and any other function that depended on it would need to input the config, which was deemed too annoying.

bendk:
I got the `peek` and `diff` working.  I've been having fun playing around with them by calling `diff-save`, making some changes to an example crate, then calling `diff [metadata|ir|python-ir]` to see how those changes affect the pipeline.

bendk:
> Wuih, big change! I won't be able to review this this year though.

Yeah it's a big one alright, there's no rush on getting it merged.

bendk:
I've been thinking about what this mean for external bindings.  I think they have 2 choices:

- Buy in to this concept: implement `VisitMut` to specialize the bindings like Python does in this PR, maybe add the `peek` and `diff` commands to their CLI.
- Don't buy in: switch to using `BindingsIr` instead of `ComponentInterface`, but keep everything else the same.  This limits the scope to minor changes, like using `&rec.name` instead of `rec.name()` and that's about it.

jhugman:
This is a meta-comment on this PR.

I haven't had the bandwidth to dig into this.

I very much like the use of the mermaid diagrams you've already done; could I request that you add another one that shows the current state, so we can get an idea of a migration path?

On the whole, I very much like the compiler metaphor: moving more logic out of the templates, having a principled approach for imports and once per type and once per type of type would be super helpful, especially for newer bindgens.

I'd also applaud having an escape hatch: I agree that there absolutely must be at least one upgrade path that has minimal breaking changes (e.g. type aliasing `BindingsIR` as `ComponentInterface` and deprecating the `name()` method rather than removing it). I'm happy to switch to a new way, but don't make me do it immediately :)

bendk:
After talking with @mhammond and @gruberb about this one, I decided to try a slightly different approach.  Rather than using the `VisitMut` trait and the `LanguageData` field, models this as more of a regular conversion from `BindingsIr` to `PythonBindingsIr`. 

I like this approach because it seems more straight forward.  Before the conversion was handled by a combination of mutating fields, filling in the `lang_data` field (technically also a mutation, but it felt much different than the other ones), and also collecting new items in the `BindingsIrVisitor` fields.  It worked, but it seemed a bit strange.

The main downside of this new approach is we need to duplicate some structs.  However, I don't really mind it for a couple reasons:
  - The Python structs are significantly different than the ones in `interface::ir`.  The fact that some of it is duplicated doesn't seem that bad to me.
  - I used a macro to implement traits like `AsType`.  Duplicating all of this boilerplate impl code was the thing I really wanted to avoid but the macros solve this well.

The second main change was making `Callable` a type and introducing the `AsCallable` trait, which works kind-of like `AsType`.  This new system works better with the macros and also makes converting slightly easier, since the `map_callable` method replaces a bunch of duplicated code from `map_function`/`map_method`/`map_constructor`.

I'm still liking the `peek` and `diff` commands.  I had a few errors when re-implementing all of the code and `peek` came in very handy to fix them.

I'm liking this one better, but I'd love feedback from others on it.  Here's the [old changes for comparison](https://github.com/mozilla/uniffi-rs/compare/main...d3ac76262937644c324fa783bb74f52c4934efaf).

bendk:
> I very much like the use of the mermaid diagrams you've already done; could I request that you add another one that shows the current state, so we can get an idea of a migration path?

I think this is a great idea, but I'm not sure how to diagram the current state.  That's one of the things I'm hoping to improve -- the new system has a more cohesive story to me than the old one.
super helpful, especially for newer bindgens.
 
> I'd also applaud having an escape hatch: I agree that there absolutely must be at least one upgrade path that has minimal breaking changes (e.g. type aliasing `BindingsIR` as `ComponentInterface` and deprecating the `name()` method rather than removing it). I'm happy to switch to a new way, but don't make me do it immediately :)

I didn't touch `ComponentInterface` code, this new code lives on top of it.  I think we can keep `ComponentInterface` around for a couple versions and so that external bindings authors have a while before they need to switch.

docs/manual/src/internals/bindings_render_pipeline.md
```diff
@@ -0,0 +1,307 @@
...
+## Proc-macro Metadata Generation
+
+When a UniFFI proc-macro wraps a Rust item, it records metadata about the item it's wrapping.
+For example, when `#[uniffi::export]` wraps a function, it records the function name, arguments, return type, etc.
+This metadata is then serialized into a buffer and stored as an exported symbol in the Rust library.
+The generated code will look something like this:
+
+```rust
+// Medata buffer for the `arithmetic::add()` function
+const UNIFFI_META_CONST_ARITHMETIC_FUNC_ADD: ::uniffi::MetadataBuffer = ::uniffi::MetadataBuffer::from_code(
```
lougeniaC64, line 36: nit: it might be helpful to also show the add function that generated this metadata

docs/manual/src/internals/bindings_render_pipeline.md
```diff
@@ -0,0 +1,307 @@
...
+    ),
+    checksum: Some(
+        39863,
+    ),
+    docstring: None,
+},
+```
+
+
+## Metadata from the UDL file
```
lougeniaC64, line 110: nit: Since there aren't details about this process (which I'm assuming is because pro-macros are favored over UDLs), you could remove this section, generalize the heading for the proc-macro metadata section, and add this as a note there.

docs/manual/src/internals/bindings_render_pipeline.md
```diff
@@ -0,0 +1,307 @@
...
+        name: "uniffi_arithmeticpm_fn_func_add",
+        ...
+    },
+    ...
+}
+```
+
+## Specializing the BindingsIr
+
+This phase transforms the general `BindingsIr` into a language-specific one.  Each bindings generate
```
lougeniaC64, line 176: I think, "Each bindings generate . . . " should probably be, "Each binding generated . . . "

docs/manual/src/internals/bindings_render_pipeline.md
```diff
@@ -0,0 +1,307 @@
...
+defines it's own version of Bindings Ir types like `Record` and `Field`.  These types mostly mirror
+the ones from the `ir` module, but there will be several differences:
+
+* Language-specific data is added, like concrete type names and the names of the `FfiConverter` classes that lift/lower types
+* New top-level items will be added to the language-specific `BindingsIr` struct, for example a list of import statements.
+* Fields like `name` and `docstring` will transformed to the preferred style for the language.
```
lougeniaC64, line 182: " . . . will *_be_* transformed . . ."

lougeniaC64:
> @lougeniaC64 flagging you for review since you said you'd take a look at the new docs. Note that these docs are for a proposed new system. I'd love your take on both the docs themselves and if the overall process makes sense.

Sorry for the delay. I left a few nits but overall the docs look good! I know you talked about maybe breaking out the bindings generation phases so processing was more transparent to uniffi devs, but wouldn't `peek` and `diff` be sufficient?

bendk:
I didn't mean to close this one quite yet, but I think it's okay anyways.  I'm going to pushing a new version of this soon.

PR #2337, opened 2024-11-28
Ignore clippy::empty_line_after_doc_comments in generated code

atezet:
Suggestion to ignore this rule, assuming the empty line and doc comments are there for a reason. If not, I happily remove the empty line or convert the doc comment to a normal comment

### DIFF
--- a/uniffi_bindgen/src/scaffolding/templates/UdlMetadata.rs
+++ b/uniffi_bindgen/src/scaffolding/templates/UdlMetadata.rs
@@ -1,4 +1,4 @@
-
+#![allow(clippy::empty_line_after_doc_comments)]
 /// Export info about the UDL while used to create us
 /// See `uniffi_bindgen::macro_metadata` for how this is used.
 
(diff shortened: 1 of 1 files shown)


jplatte:
I don't think it makes sense for that doc comment to be in that place if it's meant to apply to the const (which it currently does).

I'm having a hard time even understanding what it's trying to say 😅

atezet:
Yeah, I was too. Should I just place the doc comment above line 9? Or make it a normal comment? Or both?

mhammond:
I think line 2 could read something like "/// Export info about this UDL file", line 3 can stay the same, and lines 4-5 can just be removed?

mhammond (approved):
Thanks

jplatte:
But... It's still a doc comment on the metadata buffer constant. Is that really intended? Is this not supposed to be a regular comment, or doc comment a generated module file?

mhammond:
Yeah, I noticed that just after I merged it. We shouldn't be generating doc comments anywhere I don't think.

atezet:
Sorry about that. Do you want me to open another PR? I don't know if there are any other files that should be checked for doc comments being generated?

mhammond:
Another PR for this specific issue would be great, and if you spot any in the .rs files in the same directory there's a really good chance they should be changed too - but it's not a big deal, so as much or as little as you are willing to do would be great!

atezet:
Opened !2338

PR #2338, opened 2024-11-29
Change doc-comments in templates to comments

atezet:


uniffi_bindgen/src/scaffolding/templates/UdlMetadata.rs
```diff
@@ -1,6 +1,6 @@
```
jplatte, line 1:
```suggestion
```

jplatte (approved):

atezet:
Hehe, I was contemplating to remove that line. It is the only template file containing doc-comments

PR #2341, opened 2024-12-05
Declare `UniffiForeignFutureTask` as fileprivate

crazytonyli:
Like many other types in the generated binding files, `UniffiForeignFutureTask` should be declared as `fileprivate` or `private`. It's currently causing conflicts when multiple libraries have public async functions.


bendk (approved):
Looks good, thanks for the fix!

Issue #2340, opened 2024-12-03
`MethodTooLargeException` runtime error when executing Android tests for our Kotlin artifact


Sajjon:
I recently started getting `MethodTooLargeException` errors, e.g.

```
jdk.internal.org.objectweb.asm.MethodTooLargeException: Method too large: jdk/proxy3/$Proxy28.<clinit> ()V
	at java.base/jdk.internal.org.objectweb.asm.MethodWriter.computeMethodInfoSize(MethodWriter.java:2118)
	at java.base/jdk.internal.org.objectweb.asm.ClassWriter.toByteArray(ClassWriter.java:527)
	at java.base/java.lang.reflect.ProxyGenerator.generateClassFile(ProxyGenerator.java:506)
	at java.base/java.lang.reflect.ProxyGenerator.generateProxyClass(ProxyGenerator.java:178)
	at java.base/java.lang.reflect.Proxy$ProxyBuilder.defineProxyClass(Proxy.java:558)
	at java.base/java.lang.reflect.Proxy$ProxyBuilder.build(Proxy.java:670)
	at java.base/java.lang.reflect.Proxy.lambda$getProxyConstructor$0(Proxy.java:429)
	at java.base/jdk.internal.loader.AbstractClassLoaderValue$Memoizer.get(AbstractClassLoaderValue.java:329)
	at java.base/jdk.internal.loader.AbstractClassLoaderValue.computeIfAbsent(AbstractClassLoaderValue.java:205)
	at java.base/java.lang.reflect.Proxy.getProxyConstructor(Proxy.java:427)
	at java.base/java.lang.reflect.Proxy.newProxyInstance(Proxy.java:1037)
	at com.sun.jna.Native.load(Native.java:624)
	at com.sun.jna.Native.load(Native.java:596)
	at com.radixdlt.sargon.UniffiLib$Companion$INSTANCE$2.invoke(sargon.kt:59715)
	at com.radixdlt.sargon.UniffiLib$Companion$INSTANCE$2.invoke(sargon.kt:1213)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
	at com.radixdlt.sargon.UniffiLib$Companion.getINSTANCE$sargon_android_debug(sargon.kt:1213)
	at com.radixdlt.sargon.SargonKt.newAccessControllerAddressSampleMainnet(sargon.kt:50684)
[...]

PR #2344, opened 2024-12-06
Fix JNA bug `MethodTooLargeException` by splitting big `UniffiLib` interface

Sajjon:
This PR solves #2340 Kotlin bug which is caused by a limit in JNA causing Kotlin generated code to throw `MethodTooLargeException` at runtime.

> [!NOTE]
> The name `MethodTooLargeException` is a red herring! This is super confusing since it is, in fact, not at all
> any method being too large, it is an **interface** being too large, specifically the `UniffiLib`.

> [!TIP]
> Apart from all unit tests passing locally:
> I've asserted that this change works. See [failing Kotlin (JVM) tests in Sargon CI before this fix](https://github.com/radixdlt/sargon/actions/runs/12163143241/job/33921636969) (many `jdk.internal.org.objectweb.asm.MethodTooLargeException` all over the place) and [see successful CI when I use my UniFFI fork with this fix](https://github.com/radixdlt/sargon/actions/runs/12198795606/job/34031302782)

We at Radix got hit with this bug because apparently [`Sargon`](https://github.com/radixdlt/sargon) is the worlds biggest UniFFI crate with Kotlin binding :). The reason why Mozilla (or any other company) has not been hit with this JNA bug is probably because you use a multi-crate setup - `Sargon` does not (yet).

This PR pushes the limit of methods/functions in a UniFFI Kotlin project by a factor of ~2x. The solution is actually quite simple: 
* I've moved all `uniffi_***_uniffi_checksum_func_***` functions and the check contract version function out from `interface UniffiLib` and put them into a new `interface IntegrityCheckingUniffiLib `
* We note that this is an unintrusive change, since when `UniffiLib`'s singleton `INSTANCE` is loaded the checksums is just called once. And not referenced from anywhere else.
* I've modified the `internal val INSTANCE: UniffiLib` lazy init, to load the library with `loadIndirect` first as `IntegrityCheckingUniffiLib ` and then call `uniffiCheckContractApiVersion` and then `uniffiCheckApiChecksums` and then load the library again with `loadIndirect` as `UniffiLib`.
* I've tried calling `loadIndirect` just once, and casting it, but that does not work since JNA load creates a proxy instance which cannot be cast later on, however...
* ... the cost of calling `loadIndirect` is `~50ms` on JVM macOS Sonoma | Mac Studio M1 Max (slower on slow Android devices ofc). Since this is a one time cost (well two times now) and typically still fast, this seems acceptable.

## Changes expressed in generated Kotlin Code
```diff
+ internal interface IntegrityCheckingUniffiLib: Library {
+     fun ffi_uniffi_contract_version(): FfiFunction
+
+     fun uniffi_sargon_uniffi_checksum_func_new_foobar1(): Short
+     fun uniffi_sargon_uniffi_checksum_func_new_foobar2(): Short
+ }

internal interface UniffiLib: Library {
    companion object {
        internal val INSTANCE: UniffiLib by lazy {
+            val componentName = "{{ ci.namespace() }}" // PR Comment: e.g. "sargon"
+            loadIndirect<IntegrityCheckingUniffiLib>(componentName)
+                .also { lib: IntegrityCheckingUniffiLib ->
+                    uniffiCheckContractApiVersion(lib)
+                    uniffiCheckApiChecksums(lib)
+                }
-            loadIndirect<UniffiLib>("sargon")
+            loadIndirect<UniffiLib>(componentName)
-                .also { lib: UniffiLib ->
-                   uniffiCheckContractApiVersion(lib)
-                   uniffiCheckApiChecksums(lib)
-                }
+ /* COND */     .also { lib: UniffiLib -> // PR Comment: only included if `initialization_fns()` isnt empty
+ /* COND */          // initialization functions 
+ /* COND */      }
        }
    }

-   fun ffi_uniffi_contract_version(): FfiFunction

-   fun uniffi_sargon_uniffi_checksum_func_new_foobar1(): Short
-   fun uniffi_sargon_uniffi_checksum_func_new_foobar2(): Short
}

- private fun uniffiCheckContractApiVersion(lib: UniffiLib) {
+ private fun uniffiCheckContractApiVersion(lib: IntegrityCheckingUniffiLib) {

- private fun uniffiCheckApiChecksums(lib: UniffiLib) {
+ private fun uniffiCheckApiChecksums(lib: IntegrityCheckingUniffiLib) {
```

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
@@ -78,6 +78,9 @@ fun {{ func.name() }}(
 // into `IntegrityCheckingUniffiLib` and these methods are called only once,
 // when the library is loaded.
 internal interface IntegrityCheckingUniffiLib : Library {
+    // Integrity check functions only
+    {# newline below wanted #}
+
 {%- call decl_kotlin_functions(ci.iter_ffi_function_integrity_checks()) %}
 }
 
@@ -112,6 +115,11 @@ internal interface UniffiLib : Library {
             // to trigger this issue, the performance impact is negligible, running on
             // a macOS M1 machine the `loadIndirect` call takes ~50ms.
             loadIndirect<UniffiLib>(componentName)
+            {%- if !self.initialization_fns().is_empty() -%}
+                {#-
+                // We only include the `.also` block if there are initialization functions
+                // otherwise we get linting errors saying `lib` is unused.
+                -#}
                 .also { lib: UniffiLib ->
                     // No need to check the contract version and checksums, since 
                     // we already did that with `IntegrityCheckingUniffiLib` above.
@@ -119,6 +127,9 @@ internal interface UniffiLib : Library {
                     {{ fn }}(lib)
                     {% endfor -%}
                 }
+            {%- endif -%}
+
+            // Loading of library with integrity check done.
         }
         {% if ci.contains_object_types() %}
         // The Cleaner for the whole library
@@ -127,6 +138,10 @@ internal interface UniffiLib : Library {
         }
         {%- endif %}
     }
+
+    // FFI functions
+    {# newline below before call decl_kotlin_functions is needed #}
+
     {%- call decl_kotlin_functions(ci.iter_ffi_function_definitions_excluding_integrity_checks()) %}
 }
 
--- a/uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
@@ -55,20 +55,70 @@ internal open class {{ ffi_struct.name()|ffi_struct_name }}(
 {%- endmatch %}
 {%- endfor %}
 
+
+{%- macro decl_kotlin_functions(func_list) -%}
+{% for func in func_list -%}
+fun {{ func.name() }}(
+    {%- call kt::arg_list_ffi_decl(func) %}
+): {% match func.return_type() %}{% when Some with (return_type) %}{{ return_type.borrow()|ffi_type_name_by_value }}{% when None %}Unit{% endmatch %}
+{% endfor %}
+{%- endmacro %}
+
+// For large crates we prevent `MethodTooLargeException` (see #2340)
+// N.B. the name of the extension is very misleading, since it is 
+// rather `InterfaceTooLargeException`, caused by too many methods 
+// in the interface for large crates.
+//
+// By splitting the otherwise huge interface into two parts
+// * UniffiLib 
+// * IntegrityCheckingUniffiLib (this)
+// we allow for ~2x as many methods in the UniffiLib interface.
+// 
+// The `ffi_uniffi_contract_version` method and all checksum methods are put 
+// into `IntegrityCheckingUniffiLib` and these methods are called only once,
+// when the library is loaded.
+internal interface IntegrityCheckingUniffiLib : Library {
+{%- call decl_kotlin_functions(ci.iter_ffi_function_integrity_checks()) %}
+}
+
 // A JNA Library to expose the extern-C FFI definitions.
 // This is an implementation detail which will be called internally by the public API.
-
 internal interface UniffiLib : Library {
     companion object {
         internal val INSTANCE: UniffiLib by lazy {
-            loadIndirect<UniffiLib>(componentName = "{{ ci.namespace() }}")
-            .also { lib: UniffiLib ->
-                uniffiCheckContractApiVersion(lib)
-                uniffiCheckApiChecksums(lib)
-                {% for fn in self.initialization_fns() -%}
-                {{ fn }}(lib)
-                {% endfor -%}
-            }
+            val componentName = "{{ ci.namespace() }}"
+            // For large crates we prevent `MethodTooLargeException` (see #2340)
+            // N.B. the name of the extension is very misleading, since it is 
+            // rather `InterfaceTooLargeException`, caused by too many methods 
+            // in the interface for large crates.
+            //
+            // By splitting the otherwise huge interface into two parts
+            // * UniffiLib (this)
+            // * IntegrityCheckingUniffiLib
+            // And all checksum methods are put into `IntegrityCheckingUniffiLib`
+            // we allow for ~2x as many methods in the UniffiLib interface.
+            // 
+            // Thus we first load the library with `loadIndirect` as `IntegrityCheckingUniffiLib`
+            // so that we can call `uniffiCheckApiChecksums`...
+            loadIndirect<IntegrityCheckingUniffiLib>(componentName)
+                .also { lib: IntegrityCheckingUniffiLib ->
+                    uniffiCheckContractApiVersion(lib)
+                    uniffiCheckApiChecksums(lib)
+                }
+            // ... and then we load the library as `UniffiLib`
+            // N.B. we cannot use `loadIndirect` once and then try to cast it to `UniffiLib`
+            // => results in `java.lang.ClassCastException: com.sun.proxy.$Proxy cannot be cast to ...`
+            // error. So we must call `loadIndirect` twice. For crates large enough
+            // to trigger this issue, the performance impact is negligible, running on
+            // a macOS M1 machine the `loadIndirect` call takes ~50ms.
+            loadIndirect<UniffiLib>(componentName)
+                .also { lib: UniffiLib ->
+                    // No need to check the contract version and checksums, since 
+                    // we already did that with `IntegrityCheckingUniffiLib` above.
+                    {% for fn in self.initialization_fns() -%}
+                    {{ fn }}(lib)
+                    {% endfor -%}
+                }
         }
         {% if ci.contains_object_types() %}
         // The Cleaner for the whole library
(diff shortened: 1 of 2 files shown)


uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
```diff
@@ -55,20 +55,65 @@ internal open class {{ ffi_struct.name()|ffi_struct_name }}(
...
+            // * UniffiLib (this)
+            // * UniffiLibChecksums
+            // And all checksum methods are put into `UniffiLibChecksums`
+            // we allow for ~2x as many methods in the UniffiLib interface.
+            // 
+            // Thus we first load the library with `loadIndirect` as `UniffiLibChecksums`
+            // so that we can call `uniffiCheckApiChecksums`...
+            loadIndirect<UniffiLibChecksums>(componentName)
+                .also { lib: UniffiLibChecksums ->
+                    uniffiCheckApiChecksums(lib)
```
Sajjon, line 102: Let me know if I should adjust `interface UniffiLibChecksums` to contain the `ffi_uniffi_contract_version` function so that we can call `uniffiCheckContractApiVersion(lib)` inside this `.also` block, which would be closer to what we do on `main` branch (call `uniffiCheckContractApiVersion` **before** `uniffiCheckApiChecksums`).

Alternatively I can change to this:

```kotlin
val lib =  loadIndirect<UniffiLib>(componentName)
                .also { lib: UniffiLib ->
                   uniffiCheckContractApiVersion(lib)
                }

// check checksums
loadIndirect<UniffiLibChecksums>(componentName)
                .also { lib: UniffiLibChecksums ->
                    uniffiCheckApiChecksums(lib)

lib // return lib
```

to ensure `uniffiCheckContractApiVersion` is called before `uniffiCheckApiChecksums`?

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -657,12 +657,33 @@ impl ComponentInterface {
...
+        include_checksums: bool,
+    ) -> impl Iterator<Item = FfiFunction> + '_ {
+        let iterator = self
+            .iter_user_ffi_function_definitions()
             .cloned()
             .chain(self.iter_rust_buffer_ffi_function_definitions())
             .chain(self.iter_futures_ffi_function_definitions())
-            .chain(self.iter_checksum_ffi_functions())
-            .chain([self.ffi_uniffi_contract_version()])
+            .chain([self.ffi_uniffi_contract_version()]);
```
Sajjon, line 678: this change puts `ffi_uniffi_contract_version ` **before** in `iter_checksum_ffi_functions`, but it should not have any implications is code. Since it just affects the order in where all generated source code places the `uniffiCheckContractApiVersion` and `checksum` functions. Which does not incur any logic change.

uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
```diff
@@ -55,20 +55,65 @@ internal open class {{ ffi_struct.name()|ffi_struct_name }}(
 {%- endmatch %}
 {%- endfor %}
 
+
+{%- macro decl_kotlin_functions(func_list) -%}
```
Sajjon, line 59: Called from `UniffiLib` (as usual) but this time **without the checksum functions** - for Kotlin only. Other language bindings are not affected.

And also called from new `interface UniffiLibChecksums`, **with checksum functions only**

uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
```diff
@@ -55,20 +55,65 @@ internal open class {{ ffi_struct.name()|ffi_struct_name }}(
...
+{% for func in func_list -%}
+fun {{ func.name() }}(
+    {%- call kt::arg_list_ffi_decl(func) %}
+): {% match func.return_type() %}{% when Some with (return_type) %}{{ return_type.borrow()|ffi_type_name_by_value }}{% when None %}Unit{% endmatch %}
+{% endfor %}
+{%- endmacro %}
+
+// For large crates we prevent `MethodTooLargeException` (see #2340)
```
Sajjon, line 67: let me know if you want me to make those comments not be part of emitted Kotlin code, but "local to" the template (Rinja comments), personally I like having them in the generated kotlin code. Let me know!

uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
```diff
@@ -55,20 +55,65 @@ internal open class {{ ffi_struct.name()|ffi_struct_name }}(
...
+            // rather `InterfaceTooLargeException`, caused by too many methods 
+            // in the interface for large crates.
+            //
+            // By splitting the otherwise huge interface into two parts
+            // * UniffiLib (this)
+            // * UniffiLibChecksums
+            // And all checksum methods are put into `UniffiLibChecksums`
+            // we allow for ~2x as many methods in the UniffiLib interface.
+            // 
+            // Thus we first load the library with `loadIndirect` as `UniffiLibChecksums`
```
Sajjon, line 98: let me know if you want me to make those comments not be part of emitted Kotlin code, but "local to" the template (Rinja comments), personally I like having them in the generated kotlin code. Let me know!

micbakos-rdx:
We run some tests with @Sajjon measuring the time difference it takes to resolve the `INSTANCE` on our [wallet](https://github.com/radixdlt/babylon-wallet-android)

We didn't notice any relative performance hit. We built sargon-uniffi crate in debug mode and exported two android `aar` libraries. One without the fix, but with lower amount of methods (so as not to hit the  `MethodTooLargeException`) and one with the fix.

The resulting android app was also built in debug mode and run on android emulator and in a relatively low spec Samsung device (Samsung A53). Keep in mind that these measurements are always inflated in debug mode.

Emulator: ~1ms of max difference (sometimes the instance was resolved quicker with the fix). (65.8ms ~ 66.8ms)  
Samsung A53:  ~12ms of max difference (sometimes the instance was resolved quicker with the fix). (247ms ~ 259ms)

Although we were not able to minimise the noise affected by other factors (especially with the real device), we didn't observe any noticeable performance hit.

Sajjon:
@bendk let me know if you need something else to get this merged! :)

uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
bendk: I prefer `2`, since a contract version change could potentially affect the checksum function signature.  However, if this makes the `interface` code more complicated, it's not worth it.

bendk (changes requested):
Thanks for doing this.  Hopefully this is enough to avoid the issue although I guess if your library was 2x as big then you'd run into it again.  I'm flagging this as requesting changes, just because there was the question of how to handle the library version check.  If you think the current way is the cleanest, just say so and I'm happy to take it as-is.

I'm really wondering if all these checksum checks are worth the complexity, but I don't think we're going to make a decision on that soon.  In the meantime, this seems like a good workaround.

uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
Sajjon: I will give it a go

Sajjon: @bendk OK I've renamed to new lib `IntegrityCheckingUniffiLib` (instead of `UniffiLibChecksums`) since it now also performs the contract version check. I think "integrity check" is OK terminology to describe the aggregation of "contract version check" and "checksum checks".

Sajjon: @bendk I've ensured to not get lint errors ([which I got at first](https://app.circleci.com/pipelines/github/mozilla/uniffi-rs/5037/workflows/da1d88a7-0392-45bb-8187-9410b1328e5c/jobs/21347)) if `lib: UniFFILib` is not used - which can be the case now if `initialization_fns()` is empty. 

I've updated to only conditionally include the `.also` after loading lib, only if `initialization_fns()` is not empty.

I think that is the solution. Let me know if you are happy with it.

Sajjon:
@bendk 
> Hopefully this is enough to avoid the issue although I guess if your library was 2x as big then you'd run into it again.
Yes if we were to 2x our lib we would be hit by this again, that is true. But I expect us to grow our API with 30-50% only (i.e. less than 100%) in the coming months, and then actually decrease the UniFFI exported API "surface" as we are able to use more complex flows in Rust, thus reducing need for some methods and records :)

But we can note that it is possible to accommodate even bigger libraries if we were to split both `IntegrityCheckingUniffiLib` and `UniffiLib` lib in **many** interfaces each. This will be easy for  `IntegrityCheckingUniffiLib`. It is sliiightly tricker for `UniffiLib` since we would need to know which "batch" of `UniffiLibBatchX` we would call to for our FFI functions/methods, e.g.:

```kotlin
fun gradient(ln: Line): kotlin.Double {
    FfiConverterDouble.lift(
        uniffiRustCall { _status ->
            UniffiLibBatch1.INSTANCE.uniffi_uniffi_geometry_fn_func_gradient(
                FfiConverterTypeLine.lower(ln), _status
            )
        }
    )
}
```
Should it be `UniffiLibBatch1.INSTANCE` or `UniffiLibBatch2.INSTANCE` ? We would need to keep track of that. Not hard, just some more changes... Maybe would would be able to know it deterministically through a hash of the signature and then modulus and map to BatchX or something like that. Or maybe OK to keep a `Map` in `Rinja` of method/function -> Batch number.
```

mhammond:
@skeet70 do you have any thoughts here? uniffi_bindgen-java will have the same problem I assume?

skeet70:
Ah that's unfortunate, along with the fix introducing more initial delay and if needed again doesn't scale well. I don't have a better solution though. You're right @mhammond that the Java bindings are likely affected by the same bug, I'll make a ticket. All classes/interfaces are split out into individual files there, but if the problem is a single interface getting too large, it'll still be an issue.

bendk:
> We can note that it is possible to accommodate even bigger libraries if we were to split both IntegrityCheckingUniffiLib and UniffiLib lib in many interfaces each. This will be easy for IntegrityCheckingUniffiLib. It is sliiightly tricker for UniffiLib since we would need to know which "batch" of UniffiLibBatchX we would call to for our FFI functions/methods, e.g:...

I think #2333 might help with this if we ever needed it (especially after I make some changes suggested by @mhammond).  The general idea is to add a transformation pass where we convert the generalized interface into the Kotlin-specific interface.  Part of that could be to take the FFI functions and split them up into different batches and also associate the batch number with each FFI function reference.

bendk (approved):
The changes look great to me, I'd be good with merging.  Are there any remaining concerns?

Sajjon:
> Are there any remaining concerns?

Good to merge from my perspective, what say @mhammond  ?

PR #2347, opened 2024-12-09
Cleanup docs

mhammond:
WIP


mhammond:
I ended up splitting many docs originally authored by @jplatte, so explicitly asking for a signoff on my butchering.

docs/manual/src/proc_macro/functions.md
```diff
@@ -0,0 +1,88 @@
...
+Arguments and receivers can also be references to these types, for example:
+
+```rust
+// Input data types as references
+#[uniffi::export]
+fn process_data(a: &MyRecord, b: &MyEnum, c: &Option<MyRecord>) {
+    ...
+}
+
+## Default values
```
jplatte, line 24: Before this (and after the general section above), I think it would be good to have a reference to the part of `interfaces.md` that tells you that `#[uniffi::export]` on an `impl` block must be for an interface type.

docs/manual/src/types/interfaces.md
```diff
@@ -0,0 +1,381 @@
+# Interfaces
+
+Interfaces can have constructors and have methods. In Rust, they are represented as `impl` blocks. In the Kotlin or Swift world, they are a class, so they are often known as "Objects"
```
jplatte, line 3: I'd love if the canonical term for this was "object types" in the future. Maybe not this PR. (I think "interfaces" evokes generics too much)

docs/manual/src/types/structs.md
```diff
@@ -1,54 +1,32 @@
-# Structs/Dictionaries
+# Records
```
jplatte, line 1: Yay! Maybe rename the file accordingly?

jplatte (approved):
I haven't read over everything but what I've seen, I like.

docs/manual/src/types/structs.md
```diff
@@ -141,3 +108,5 @@ data class TodoEntry (
 ```
 
 This works for Swift and Python targets too.
+
+XXX - what does this even mean? :) Do proc-macros support it?
```
joe-p, line 112: This is also unclear to me. I would assume this is the default behavior of `Option`?

docs/manual/src/udl/docstrings.md
```diff
@@ -3,8 +3,12 @@
...
 transformations. What you see in UDL is what you get in generated bindings. The only change made to
 UDL comments are the comment syntax specific to each language. Docstrings can be used for most
-declarations in UDL file. Docstrings are parsed as AST nodes, so incorrectly placed docstrings will
-generate parse errors. Docstrings in UDL are comments prefixed with `///`.
+declarations in UDL file
+
+Docstrings in UDL are comments prefixed with `///`.
+
+Docstrings are parsed as AST nodes, so incorrectly placed docstrings will
+generate parse errors
```
joe-p, line 11:
```suggestion
Docstrings are parsed as AST nodes, so incorrectly placed docstrings will generate parse errors
```

docs/manual/src/udl/index.md
```diff
@@ -1,6 +1,9 @@
 # The UDL file
 
-This file defines which functions, methods and types are exposed to the foreign-language bindings.
+A UDL file allows you to define an interface externally from your Rust code.
+It defines which functions, methods and types are exposed to the foreign-language bindings.
```
joe-p, line 4: Perhaps a line or two here about the relationship of the UDL file itself to binding generators. Ie it *can* be used as input to binding generators but typically `--library` is used.

joe-p (commented):
Left a decent amount of comments. Some are nits and some might seem obvious but I am new to both UniFFI and Rust and future readers might be as well. 

The one significant piece of feedback I have is that I think it would be easier to consume and maintain the docs if `udl/` and `proc_procmacro/` were removed and the information just directly integrated on the respective `types/` page. This puts all the conceptually stuff in the same place as how implementation documentation which makes it a bit easier to navigate and reduces duplication of information.

docs/manual/src/index.md
```diff
@@ -16,13 +16,6 @@ This process can take place either during the build process or be manually initi
 
 ## Supported languages
 
-- Kotlin
-- Swift
-- Python
-- Ruby
+UniFFI comes with full support for Kotlin, Swift and Python, plus partial legacy support for Ruby.
```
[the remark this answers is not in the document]
bendk: I don't disagree with this take.  The Ruby bindings were contributed early on and we happily took them.  But as time went on, it's clear that we can't maintain them.

docs/manual/src/proc_macro/docstrings.md
```diff
@@ -0,0 +1,14 @@
...
+For example:
+```rust
+/// This is the docstring for MyObject
+#[derive(uniffi::Object)]
+pub struct MyObject {}
+```
+
+Will cause Python, Swift and Kotlin to all generate a wrapper for `MyObject` with appropriate docstrings for that language.
+
+You can see examples of how they are rendered in the [UDL docstrings documentation](../udl/docstrings.md)
```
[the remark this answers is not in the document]
bendk: I'm torn on this one.  I think "inlining" the UDL and proc-macro sections is nice when they're short, but if they got too long and went into too much specifics then they could be a distraction from the rest of the documents.

Where are these pages linked from?  In general it seems like there's a generic `types/foo.md` page that links to the `udl/foo.md` and `proc-macro/foo.md` pages.  If so, then it's hard to get to the specific pages without going through the general "why" / "how" sections.   But, that's not true for docstrings.

I'll just re-iterate that I'm torn.  Maybe we could inline the UDL/proc-macro sections in general, but use UDL/proc-macro specific pages when the text gets long.  I'd also fine with the current layout, it's really hard to know what's best.

mkdocs.yml
```diff
@@ -38,22 +38,37 @@ nav:
...
+    - ./udl/records.md
+    - ./udl/docstrings.md
+  - 'Proc macros':
+    - ./proc_macro/index.md
+    - ./proc_macro/enumerations.md
+    - ./proc_macro/errors.md
+    - ./proc_macro/functions.md
+    - ./proc_macro/interfaces.md
+    - ./proc_macro/records.md
+    - ./proc_macro/docstrings.md
```
bendk, line 71: Maybe we don't all of these TOC entries.  Like @joe-p says, if I wanted to learn about records and I was using proc-macros, I'm still probably better off starting an `types/records.md`.

Instead, there could just one top-level proc-macro TOC entry, which would contain the overview on proc-macros and a link to the docstrings info.

bendk (approved):
I'm going to mark this approve, since I think this is a big win overall and we should continue to iterate on the docs.  If you want to take my suggestions go for it, but we can also just discuss them in a future PR instead.

(Also, I want to see this in book form to really understand the effect of the changes).

docs/manual/src/types/interfaces.md
bendk: I feel this as well, but then the literal part of me wonders: aren't records and enums also "objects"?  Maybe it doesn't matter if "object" invokes the right feeling in users.

I'll also throw out "class" as another term that might work.

docs/manual/src/proc_macro/interfaces.md
```diff
@@ -0,0 +1,122 @@
+# The `uniffi::Object` derive
+
+The `Object` derive registers a [UniFFI interface](../types/interfaces.md).
```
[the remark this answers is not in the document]
mhammond: The `proc_macro` docs shouldn't be describing types, but a back-link to ../types makes sense.

docs/manual/src/proc_macro/interfaces.md
```diff
@@ -0,0 +1,122 @@
...
+    // Returning objects is also supported, either as `Self` or `Arc<Self>`
+    fn method_b(self: Arc<Self>) {
+        // ...
+    }
+}
+```
+
+See [more about constructors here](./functions.md)
+
+# Traits
```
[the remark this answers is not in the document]
mhammond: As above, that belongs in ../types

docs/manual/src/types/structs.md
bendk: I think proc-macros support this.  As to the "Swift and Python targets too", the only meaning I can figure out is that it's saying just because there's a Kotlin example, don't assume that it's Kotlin-only.   That doesn't seem like something we really need to say though.

docs/manual/src/specifying.md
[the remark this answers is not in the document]
mhammond: The tutorial hopefully serves some of that purpose, but I agree we should more here but shouldn't block for new docs.

docs/manual/src/types/enumerations.md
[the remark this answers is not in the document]
mhammond: I agree entirely - getting these enum docs merged only`types/enumerations.md` exists would be perfect. But this PR is just shuffling the existing docs around into something more coherent so we can move towards that, one type at a time :)

docs/manual/src/types/interfaces.md
```diff
@@ -0,0 +1,381 @@
...
+Along with a default constructor, an interface can have named constructors, implemented as static functions.
+
+Constructors:
+* may omit or include the outer `Arc<>` - eg, we could have written `fn new() -> Arc<Self>`
+* can return a `Result<>`
+* can be async, although foreign language constraints means support for async primary constructors is patchy.
+
+## Destructors
+
+The foreign bindings will typically generate destructors, but regardless of the foreign semantics, they always hold an `Arc<>` to the Rust object, so these destructors will only drop their reference and may not drop the Rust object.
```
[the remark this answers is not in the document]
mhammond: It's correct, but I think it will be obvious to anyone who understands a Rust `Arc<>` and I value short docs ;)

docs/manual/src/udl/index.md
mhammond: would make a great followup PR!

mhammond (commented):
Thanks for the great comments! A few do reflect we need better docs in general, but this PR is more about shuffling stuff around rather than adding too much new. Hopefully it will make it easier to find the correct place for new/better docs in the future.

docs/manual/src/proc_macro/docstrings.md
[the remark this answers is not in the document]
mhammond: It already is that for custom types :) I do see the `types` version being the only version of the docs in many cases, and I actually agree docstrings might be a good candidate.

mkdocs.yml
mhammond: That's where I started too, but I think this ends up working better in the rendered versions given they are default collapsed. I agree there's work needed here though!

docs/manual/src/proc_macro/functions.md
mhammond: I think this is what you meant!

docs/manual/src/proc_macro/docstrings.md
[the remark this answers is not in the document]
joe-p: Another potential solution for longer docs is to put more detail in the docstings themselves (with docstrings being the one we can't do that for...). This would mean these markdown docs have the most important/common information on usage and then link to docs.rs for more detailed information on specific objects. This also has the added benefit of giving instant access  to this information within IDEs. Although now that I think about it this only applies to proc macros and not UDL, so perhaps not a great solution.

PR #2348, opened 2024-12-09
[WIP] Better bindings tests

bendk:
I'm sharing this one early, before I get too deep in it.  I only implement the first few basic tests and only in Python, but hopefully it paints a fairly clear picture of what I'm going for. WDYT?

-------

This comes from my experiences with #2333 and uniffi-bindgen-gecko-js. For those, I needed to run a lot of tests while (re)implementing a bindings generator.  I love how the current tests cover basically all UniFFI features, but I think there a few things that can be improved. This commit adds a suite of tests specifically aimed at the bindings.

The new tests target specific features rather than trying to mimic real-world examples.  Before, I spent a lot of time trying to figure out what a test failure meant.  I'm hoping that when one of these tests fail, it's obvious what's not working.  It also means that the new test suites don't require writing so much foreign code.

They also have a better external bindings story.  I'm thinking we can direct external bindings authors to copy everything the `bindings-tests/fixtures` sub-directories and write their own version of `bindings-tests/tests/tests.rs`. They get to use the macro now, since it's doesn't hard code the mapping from file extension to test runner. Copying the code feels a bit weird, but I like it better than having to publish all these test fixture crates and I think it will work okay in practice.
I still want to keep the current examples/fixtures. They're nice because they showcase real-world use-cases and test that those use-cases work like we expect them to. If this gets merged, we can rework those crates to focus on that.  Maybe we rework them to only test one bindings language per example/fixture.


bendk:
I didn't mean to close this one quite yet, but I think it's okay anyways.  I'm going to pushing a new version of this soon.

Issue #2343, opened 2024-12-05
Using async traint crashes at "Foreign pointer not set"


crazytonyli:
I ran into a crash when having a crate using an async trait defined in another crate. Here is the crash site.

https://github.com/mozilla/uniffi-rs/blob/2c003b16d1e70e1914b5d8ceb517eef3676cd187/uniffi_core/src/ffi/foreigncallbacks.rs#L31-L33

To reproduce this crash:
1. Get the project from https://github.com/crazytonyli/uniffi-foreign-async-crash
2. Run `./build.sh && swift test --package-path swift`. You can see the full backtrace from Xcode, too.

I'm using Xcode 16.1 on macOS 15.

mhammond:
Thanks for the report. One of us will try and repro this, but if you'd like to help things along, the most impactful thing you could do would be to submit a patch which demonstrates it in our fixtures - this is the first thing anyone would need to do regardless so that any fix has suitable tests, and having this done beforehand would make it that much more tractable for one of us to look into the actual problem.

crazytonyli:
I can have a look at that if I find some time. Meanwhile, to whoever is looking at this issue, please feel free to use the code in the project if needed.

bendk:
I believe https://github.com/mozilla/uniffi-rs/pull/2367 fixes this one

PR #2349, opened 2024-12-10
Add test cases to reproduce issue #2343

crazytonyli:
I don't want to add failing test cases to the repo. To avoid blocking development and/or CI, the new test cases passes when running `cargo test`, but fails when adding a specific env: `ISSUE_2342=1 cargo test`.

Here is an example to verify the tests locally:

```
// Pass
cargo test --package uniffi-fixture-ext-types-sub-lib

// Panic
ISSUE_2343=1 cargo test --package uniffi-fixture-ext-types-sub-lib
```

From the crash log:

```
thread 'thread '<unnamed><unnamed>' panicked at .../uniffi_core/src/ffi/foreigncallbacks.rs:32:18' panicked at :
.../uniffi_core/src/ffi/foreigncallbacks.rs:Foreign pointer not set.  This is likely a uniffi bug.32
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1816,6 +1816,7 @@ name = "uniffi-fixture-ext-types-lib-one"
 version = "0.22.0"
 dependencies = [
  "anyhow",
+ "async-trait",
  "bytes",
  "uniffi",
 ]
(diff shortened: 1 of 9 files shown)


bendk:
I think https://github.com/mozilla/uniffi-rs/pull/2367 covers this one and we can close it.  Thanks again for reproing this one with a nice test.

PR #2353, opened 2024-12-13
Async internal docs

bendk:
I've been meaning to document the async internals for a while now, here's my first draft.

### DIFF
--- a/docs/manual/src/internals/async.md
+++ b/docs/manual/src/internals/async.md
@@ -1,0 +1,161 @@
+# UniFFI Async internals
+
+## What runtime does async code run on?
+
+The fundamental issue here is that UniFFI can't rely on a Rust async runtime.
+We don't want to force library authors into a particular runtime.
+Furthermore, we want to allow library authors to not use a runtime at all and instead to integrate with the runtime of the application using the library.
+
+To accomplish this, UniFFI piggybacks off the runtime from the foreign bindings.
+The generated Rust code schedules work using callbacks provided by the foreign bindings.
+
+## Rust Async functions
+
+Rust async functions are implemplemented by wrapping the
+[Future](https://doc.rust-lang.org/std/future/trait.Future.html) into a `uniffi::RustFuture` struct
+and providing scaffolding functions so that the foreign bindings can drive the future to completion.
+
+
+```mermaid
+sequenceDiagram
+  participant User as User code
+  participant Foreign as Foreign Language Bindings
+  participant Rust as Rust Scaffolding
+
+  User->>+Foreign: call async function
+  Foreign->>+Rust: call scaffolding function
+  Rust-->>-Foreign: return RustFuture handle
+
+  loop until RustFuture is ready
+    Foreign->>Rust: call RustFuture poll fn
+    Rust->>Foreign: call future callback
+  end
+
+  Foreign->>+Rust: call RustFuture complete fn
+  Rust-->>-Foreign: return result
+  Foreign->>Rust: call RustFuture free fn
+  Foreign-->>-User: return from async function
+```
+
+For each async function, UniFFI generates 4 scaffolding functions:
+* A scaffolding function that returns a `RustFuture` handle
+* `rust_future_poll` to poll the future
+* `rust_future_complete` to receive the completed result of a future
+* `rust_future_free` to destroy a future freeing any underlying memory
+* Technically, the function names are more like `uniffi_[module_prefix]_rust_future_poll`, but this document uses above names make for shorthand.
+
+The UniFFI generated code for an async function performs these steps:
+
+1. Call the Rust scaffolding function, receiving a `RustFuture` handle
+1. Call the `rust_future_poll` function for the future until the future is ready.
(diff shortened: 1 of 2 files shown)


mhammond (approved):

bendk:
Thanks for the quick review, I just pushed one more section that describes an async call at a high-level.  @gruberb Does that section makes sense?  I was trying to describe this in our meeting 2 weeks ago, but I think the diagram does a better job.

docs/manual/src/internals/async.md
```diff
@@ -0,0 +1,159 @@
...
+   * If the future is pending, then the generated code registers a waker that will call the callback function with `RUST_FUTURE_MAYBE_READY`.
+     When the generated foreign code sees this, it calls poll again, starting the loop over.
+   * If the future is ready, then the callback function is immediately called with `RUST_FUTURE_READY` and we move to the next step.
+1. Call `rust_future_complete`, receiving the return value of the future
+1. Call `rust_future_free` (ideally in a `finally` block)
+1. Return the result from `rust_future_complete`
+
+### Why not have `rust_future_poll` return a boolean?
+
+As described above, `rust_future_poll` inputs a callback and an opaque pointer.
```
gruberb, line 61: Can you explain (or link) to the definition of an "opaque pointer"?

docs/manual/src/internals/async.md
```diff
@@ -0,0 +1,159 @@
...
+If the future is ready, then `rust_future_poll` immediately invokes the callback, passing it the pointer back and `RUST_FUTURE_READY`
+A more straightforward API would be to return a boolean if the future is ready, why not do that?
+
+The issue is that it doesn't work well with the way foreign code's lifetime management.
+The opaque pointer is usually either something like an `Arc` or `Box` that holds data about the async call.
+When the foreign code passes the pointer to `rust_future_poll`, it temporarily leaks a reference (either an Arc refcount or the Box itself).
```
gruberb, line 67: What does it mean to "leak a reference"?

docs/manual/src/internals/async.md
```diff
@@ -0,0 +1,159 @@
...
+Then when callback is called, it reconstructs the leaked reference using the pointer, undoing the leak and restoring balance to the universe.
```
gruberb, line 68: "Undoing the leak"?

docs/manual/src/internals/async.md
```diff
@@ -0,0 +1,159 @@
...
+
+If `rust_future_poll` returned `true` to indicate that the future is ready, then the foreign code is in an awkward situation.
+It wants to use the async call data that the pointer was referring to, but it just sent a leaked pointer to Rust.
+It would have to hold on to a copy of the raw pointer, use that to restore the reference, and assume the callback is never going to be called.
```
gruberb, line 72: Raw pointer, opaque pointer, restore the reference.... There are a lot of terms. Could we have an index of terms and their definitions somewhere? The definition doesn't have to come from us!

docs/manual/src/internals/async.md
```diff
@@ -0,0 +1,159 @@
...
+  AsyncRust->>WorkerQueue: Schedule task
+  Note over AsyncRust: await oneshot channel
+  WorkerQueue->>SyncRust: Run task (in background thread)
+  SyncRust->>AsyncRust: Send result via oneshot channel
+  AsyncRust-->>-Foreign: Return API result
+```
+
+* The async Rust code runs in a [Future::poll](https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll) method, inside the foreign event loop.  This code never blocks.
+* The sync Rust code runs inside a worker queue where it's okay to block.
+* This all combines together to create an async Rust call that's driven by the foreign event loop.
```
gruberb, line 149: Don't know where to put it: But I would love to also talk here about "data". Where is it? Showing the functions and who is calling what is excellent. **And** I would love to see some form of diagram which explains where does the data live and when does it getting transferred through the FFI etc. If that makes sense?

gruberb (approved):
That was a great read, thank you! What I am a bit struggling with are the terms. But it's already highly valuable to have this

docs/manual/src/internals/async.md
bendk: I love this idea, but I feel like that's out-of-scope for the async docs.  The best we currently have is this doc: https://mozilla.github.io/uniffi-rs/latest/internals/object_references.html. Maybe we could update that soon to also talk about foreign objects, future handles, etc.

docs/manual/src/internals/async.md
bendk: I updated the diagram to do a bit more of this.  Initially, I had a very detailed diagram that showed all the FFI calls, but it got extremely busy so I decided to show more of a zoomed out view.

bendk:
That was great feedback, thanks!  I updated the doc in a few places to address the concerns.  I think there's still more work to be done, but I'm not sure how to do it at this point.  I'm definitely open to suggestions.  WDYT?

gruberb:
>  I'm definitely open to suggestions. WDYT?

I think it's in a really great state! The diagrams and explanations helped a lot!! We can always iterate!

PR #2355, opened 2024-12-16
Stop supporting `typedef extern`, `[ExternalInterface=...]` in UDL etc.

mhammond:
This consolidates how external types are expressed in UDL. Instead of

```
[External="crate_name"]
typedef extern MyEnum

[ExternalInterface="crate_name"]
typedef extern MyInterface

```
you would use:
```
[External="crate_name"]
typedef enum MyEnum

[External="crate_name"]
typedef interface MyInterface
```

See the docs and upgrading notes in this commit for more.

It also takes the opportunity to use Type::External far less often in uniffi_udl, instead using the actual type of the object.

### DIFF
--- a/docs/manual/src/Upgrading.md
+++ b/docs/manual/src/Upgrading.md
@@ -1,4 +1,10 @@
-# v0.28.x -> v0.29.x
+# Upgrading v0.28.x -> v0.29.x
+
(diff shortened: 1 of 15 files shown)


mhammond:
gah, checksums have been broken for external types :(

gruberb (approved):

PR #2363, opened 2024-12-22
Swift classes can inherit from traits

### LINKED ISSUE #2169: Structs which `#[uniffi::export]` the impl of a trait should have that reflected in the bindings.
I have the following trait

```rust
use crate::impl_default_for;

pub trait AutoComplete: Send + Sync + std::fmt::Debug + 'static {
    fn autocomplete(&self, word: String) -> Vec<String>;
}

#[derive(Debug, Copy, Clone, uniffi::Object)]
pub struct Bip39AutoComplete;


#[uniffi::export]
impl Bip39AutoComplete {
    #[uniffi::constructor]
    pub fn new() -> Self {
        Self
    }
}

#[uniffi::export]
impl AutoComplete for Bip39AutoComplete {
    fn autocomplete(&self, word: String) -> Vec<String> {
        // todo: implement
        vec![]
    }
}
```

And this SwiftUI component 

```swift
struct AutocompleteField<AutoCompleter: AutoComplete>: View {
    let autocompleter: AutoCompleter
    @Binding var text: String

   ...
```

But I can't automatically pass in `Bip39AutoComplete` into the SwiftUI component. To get it to work I have to do the following on the swift side.

```swift
extension Bip39AutoComplete: AutoComplete {}
```

It's not a big deal, just one line, but I was wondering if it could be added to the already generated code as below:

```diff
- open class Bip39AutoComplete: Bip39AutoCompleteProtocol {
+ open class Bip39AutoComplete: Bip39AutoCompleteProtocol, AutoComplete {
```

mhammond:
Builds on #2196, #2204 and #2297. Closes #2169.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/gen_swift/mod.rs
+++ b/uniffi_bindgen/src/bindings/swift/gen_swift/mod.rs
@@ -272,6 +272,27 @@ impl Config {
     }
 }
 
+// Given a trait, work out what the protocol name we generate for it.
+// This differs based on whether the trait supports foreign impls (ie,
+// whether is has a "callback interface".
+fn trait_protocol_name(ci: &ComponentInterface, name: &str) -> Result<String> {
+    let (obj_name, has_callback_interface) = match ci.get_object_definition(name) {
+        Some(obj) => (obj.name(), obj.has_callback_interface()),
+        None => (
+            ci.get_callback_interface_definition(name)
+                .ok_or_else(|| anyhow::anyhow!("no interface {}", name))?
+                .name(),
+            true,
+        ),
+    };
+    let class_name = SwiftCodeOracle.class_name(obj_name);
+    if has_callback_interface {
+        Ok(class_name)
+    } else {
+        Ok(format!("{class_name}Protocol"))
+    }
+}
+
 /// Generate UniFFI component bindings for Swift, as strings in memory.
 pub fn generate_bindings(config: &Config, ci: &ComponentInterface) -> Result<Bindings> {
     let header = BridgingHeader::new(config, ci)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,8 +33,9 @@ We have [detailed upgrade notes](https://mozilla.github.io/uniffi-rs/next/Upgrad
 [Detailed upgrade notes](https://mozilla.github.io/uniffi-rs/next/Upgrading.html)
 ### What's new?
 
-- Kotlin: Proc-macros exporting an `impl Trait for Struct` block now has a class inheritance
-  hierarcy to reflect that. [#2297](https://github.com/mozilla/uniffi-rs/pull/2297)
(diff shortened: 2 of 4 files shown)


uniffi_bindgen/src/bindings/swift/gen_swift/mod.rs
```diff
@@ -272,6 +272,27 @@ impl Config {
...
+                .ok_or_else(|| anyhow::anyhow!("no interface {}", name))?
+                .name(),
+            true,
+        ),
+    };
+    let class_name = SwiftCodeOracle.class_name(obj_name);
+    if has_callback_interface {
+        Ok(class_name)
+    } else {
+        Ok(format!("{class_name}Protocol"))
```
bendk, line 292: Is there a good reason why this has "Protocol" appended?  IIUC, this is a Rust trait without callback interface support, so Swift can't implement it.  In that case, it seems like protocol should get `{class_name}` which feels like a nicer name, and the class that implements the protocol should get something like`{class_name}Impl`.  I think that's how the other languages handle it.

bendk (approved):
Looks great to me.

I had a comment about the protocol name, but that's not really in-scope for this issue.  If you agree and wanted to make that change too, that's good with me.  If you don't agree or want to do it in a separate PR that's good too.

uniffi_bindgen/src/bindings/swift/gen_swift/mod.rs
bendk: > Is this also your understanding - ie, that this patch doesn't change the class hierarchy for existing types
Yes agreed.

> ..and that doing so is something we'd need to be very careful about?

Maybe I have a slight difference of opinion here.  I think it could be okay to make the change, since most end-users aren't going to see a difference.    If they store a `Foo` and only use it to call methods from the protocol, it shouldn't matter if `Foo` is a concrete type that implements the protocol or the protocol itself.

I agree we should be careful, but maybe I'd remove the "very" part.

bendk: Also, I think I got confused and believed that we had already made the change for other languages.  If Kotlin works this way, then I think Swift should too.  If we decide to change them, let's change both at the same time.

mhammond:
> Also, I think I got confused and believed that we had already made the change for other languages. If Kotlin works this way, then I think Swift should too. If we decide to change them, let's change both at the same time.

I remember looking into this when doing this for kotlin - the names spelt wrong, but I couldn't work out how to make them good.

When UniFFI generates a plain-old trait, it only emits a protocol - there's no concrete implementation. So our `NodeTrait` generates `public protocol NodeTrait : AnyObject {}`.

But when we need a local impl of the trait (callback/foreign) we'd generate `public protocol NodeTraitProtocol : AnyObject {}` and `open class NodeTrait` which holds the pointer).

I don't see how to change those names?

praveenperera:
awesome thanks @mhammond !

PR #2367, opened 2024-12-24
Initialize VTables for external types (#2343)

bendk:
Fixed a tricky case from Swift:  Because of it's C lineage, it doesn't have a way to run initialization code when the module is imported so instead it runs a one-time initialization function the first time a FFI call is run.  This system failed when external types were used before the external module ever made an FFI call.  I believe this is Swift only, but it might affect other C-like external bindings as well.

@crazytonyli thanks for the great test case for this, I shrunk it down to a mimimal test and added it to the existing fixture.  When I rebase your test suite on top of this and run it with `ISSUE_2343=1 cargo test --package uniffi-fixture-ext-types-sub-lib` it passes.   I don't think we need the all of the tests, but tell me if you think the other ones are still worth adding.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/wrapper.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/wrapper.swift
@@ -62,7 +62,9 @@ private let initializationResult: InitializationResult = {
     return InitializationResult.ok
 }()
 
-private func uniffiEnsureInitialized() {
+// Make the ensure init function public so that other modules which have external type references to
+// our types can call it.
+public func {{ ensure_init_fn_name }}() {
     switch initializationResult {
     case .ok:
         break
--- a/fixtures/ext-types/lib/src/lib.rs
+++ b/fixtures/ext-types/lib/src/lib.rs
@@ -181,6 +181,11 @@ fn get_uniffi_one_trait(t: Option<Arc<dyn UniffiOneTrait>>) -> Option<Arc<dyn Un
     t
 }
 
+#[uniffi::export]
+fn invoke_uniffi_one_trait(t: Arc<dyn UniffiOneTrait>) -> String {
+    t.hello()
+}
+
 fn get_uniffi_one_proc_macro_type(t: UniffiOneProcMacroType) -> UniffiOneProcMacroType {
     t
 }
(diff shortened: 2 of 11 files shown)


mhammond (approved):
This looks great! I actually started looking at this last night and didn't get this far, so thanks!

Maybe change the commit message to explicitly say "Fixes #2343"?

mhammond:
does it need Python and Kotlin too?

crazytonyli:
The tests look good to me. I agree with @mhammond that it might be good to cover other languages. _I think_ the same issue happens on Kotlin, too.

bendk:
I didn't think Python or Kotlin needed this, but when I wrote the tests I realized that we did have the same bug in Kotlin.  Added some test cases and updated the kotlin code to deal with this.

uniffi_bindgen/src/bindings/swift/templates/wrapper.swift
```diff
@@ -62,7 +62,9 @@ private let initializationResult: InitializationResult = {
     return InitializationResult.ok
 }()
 
-private func uniffiEnsureInitialized() {
+// Make the ensure init function public so that other modules which have external type references to
+// our types can call it.
+public func {{ ensure_init_fn_name }}() {
```
cschramm, line 67: Is this actually necessary? According to https://mozilla.github.io/uniffi-rs/next/types/remote_ext_types.html#foreign-bindings, swift files for multiple crates are expected to be compiled in a single module, so this should be visible anyway (and otherwise, it would have to get imported somehow). Also, flipping it back to `private` does not break your tests.

bendk: Maybe not, feel free to open a PR that changes this back to private.  The swift docs say "Private access restricts the use of an entity to the enclosing declaration", so is the enclosing declaration the module?  That feels a bit weird to me, but maybe that's how swift works.

cschramm: :thinking: I think you're right, yes. On top-level the surrounding declaration should be the file, so that `private` and `fileprivate` have the same effect. I realized that the test does not actually seem to cover external initialization functions. However, this should still be `internal` instead of `public`.

PR #2369, opened 2024-12-26
Remove a panic in add_type_definition and add error contexts so resul…

mhammond:
…ting errors make sense.

With this patch, output from errors looks something like:
```
  Caused by:
      0: adding function Function { name: "get_maybe_url", module_path: "uniffi_ext_types_lib", is_async: false, arguments: [Argument { name: "url", type_: Optional { inner_type: Custom { module_path: "custom_types", name: "Url", builtin: None } }, by_ref: false, optional: false, default: None }], return_type: Some(Optional { inner_type: Custom { module_path: "custom_types", name: "Url", builtin: None } }), ffi_func: FfiFunction { name: "uniffi_uniffi_ext_types_lib_fn_func_get_maybe_url", is_async: false, arguments: [], return_type: None, has_rust_call_status_arg: true, is_object_free_function: false }, docstring: None, throws: None, checksum_fn_name: "uniffi_uniffi_ext_types_lib_checksum_func_get_maybe_url", checksum: None }
      1: adding type Optional { inner_type: Custom { module_path: "custom_types", name: "Url", builtin: None } }
      2: adding optional type Custom { module_path: "custom_types", name: "Url", builtin: None }
      3: Custom { module_path: "custom_types", name: "Url", builtin: None } must have a resolved builtin

```

### DIFF
--- a/uniffi_bindgen/src/interface/universe.rs
+++ b/uniffi_bindgen/src/interface/universe.rs
@@ -6,7 +6,7 @@
 //! which can be used by the bindings generator code to determine what type-related helper
 //! functions to emit for a given component.
 //!
-use anyhow::Result;
+use anyhow::{Context, Result};
 use std::{collections::hash_map::Entry, collections::BTreeSet, collections::HashMap};
 
 pub use uniffi_meta::{AsType, ExternalKind, NamespaceMetadata, ObjectImpl, Type, TypeIterator};
@@ -45,9 +45,12 @@ impl TypeUniverse {
     fn add_type_definition(&mut self, name: &str, type_: &Type) -> Result<()> {
         match self.type_definitions.entry(name.to_string()) {
             Entry::Occupied(o) => {
-                // all conflicts have been resolved by now in udl.
-                // I doubt procmacros could cause this?
-                assert_eq!(type_, o.get());
+                // mismatched types are bad.
+                let cur = o.get();
+                anyhow::ensure!(
+                    type_ == cur,
+                    "conflicting types:\ncur: {cur:?}\nnew: {type_:?}",
+                );
                 Ok(())
             }
             Entry::Vacant(e) => {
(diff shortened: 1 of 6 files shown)


uniffi_bindgen/src/interface/universe.rs
```diff
@@ -111,7 +118,8 @@ impl TypeUniverse {
     /// Add many [`Type`]s...
     pub fn add_known_types(&mut self, types: TypeIterator<'_>) -> Result<()> {
         for t in types {
-            self.add_known_type(t)?
+            self.add_known_type(t)
+                .with_context(|| format!("adding type {t:?}"))?
```
bendk, line 122: Are these getting overridden by the `parsing udl file...` context?  That what it seems like from the trybuild tests, but I'm not sure.

bendk (approved):
LGTM, I think the context message handling may be could be improved but I'm not sure.  Either way this feels like an improvement to me.

uniffi_bindgen/src/interface/universe.rs
mhammond: I don't think so:

mhammond: oops:

It's gone from:
> error: Failed to generate name: Invalid trait name: PartialEq
to:
> error: Failed to generate name: parsing udl <snip>: Invalid trait name: PartialEq

 which is redundant here, but I don't think is overriding anything.

Interresting note re messages - this has:

```
.add_known_types(defn.iter_types())
.with_context(|| format!("adding enum {defn:?}"))?;
```
which imo is nicer in the code, but worse in the errors. For some reason I don't like using the word "error" there:
```
.add_known_types(defn.iter_types())
.with_context(|| format!("error adding enum {defn:?}"))?;
```
but I'm torn

Issue #2025, opened 2024-03-11
Custom Types from external crate do not generate correct binding function signature when used in exports


LBeernaertProton:
Tested with uniffi v0.26.1.

When using custom types declared in another crate, the generated uniffi binding code does not generate the correct function signature for the bindings. This may be potentially related to #1988. I have verified this happens with the Swift binding generator, but could also affect other languages 

### Example
Crate A:
```
pub struct MyType(u64);

uniffi::custom_new_type(MyType,u64);

#[uniffi::export]
pub fn crate_a_test(t:MyType); 
```

Crate B:
```
use crate_a::MyType;

// Due to #1988
uniffi::ffi_converter_forward(crate_a::MyType, crate_a::UniFfiTtag, crate::UniFiiTag);

#[uniffi::export]
pub fn crate_b_test(t:MyType); 
```

Crate A will generate the following function signature:
```
uniffi_crate_a_fn_crate_a_test(uin64 t);
```

Where as Crate B will generate:
```
uniffi_crate_b_fn_crate_b_test(RustBuffer t);
```

I'm happy to assist in resolving this issue if someone can provide me some pointers on where this is handled in the code.

mhammond:
This is a little unfortunate, but IIUC is just a side-effect of external types. All generated callers to these should do the right thing. What problems are you seeing because of this?

LBeernaertProton:
The main issue I am having is that, I can not export uniffi functions that use external new types such as the example above that wrap scalar/basic types such as integers. We are currently working around this by just exposing the wrapped type to the bindings and internally converting it to the new type.

That being said, custom new types that are more "complex" such as `String` work without issues since these always translate down to a binding function that needs a `RustBuffer`.

mhammond:
> That being said, custom new types that are more "complex" such as String work without issues

Ah, yeah - I can reproduce this, thanks.

mhammond:
https://github.com/mozilla/uniffi-rs/commit/a30a67f52f2d733fd23db71632be000a3a01bb54 is a repro, but I suspect this will be very difficult to fix?

LBeernaertProton:
I am afraid that at this point in time I can not answer that question 😅

mhammond:
lol, sorry, it wasn't a question for you - it was more a statement of my understanding of the situation, with a glimmer of questioning that @bendk might see something I missed :)

mhammond:
[...]

PR #2371, opened 2024-12-30
External custom types in UDL must use the [Custom] attribute

mhammond:
Previously they were declared with the `External` attribute, but this mean UniFFI could not know the underlying type, leading to issues like #2025.

This PR does not however fix any such issues, it instead allows UniFFI metadata to carry enough information to fix in the future. This is primarily being done now to include this breaking change in with all other breaking changes we are making in the next release.

### DIFF
--- a/docs/manual/src/Upgrading.md
+++ b/docs/manual/src/Upgrading.md
@@ -3,13 +3,12 @@
 We've made a number of breaking changes in this release, particularly
 to:
 
-* Custom types (both UDL and proc-macros impacted)
-* External Types (UDL impacted)
+* Custom types: UniffiCustomTypeConverter has been removed.
+* External types: `extern` has been removed; you must describe the type.
 
 ## Custom types
 
-Custom types are now implemented using a macro rather than implementing the `UniffiCustomTypeConverter` trait,
-addressing some edge-cases with custom types wrapping types from other crates (eg, Url).
+Custom types still implemented via the `UniffiCustomTypeConverter` trait must move to proc-macros.
 
 Before:
 
@@ -29,7 +28,7 @@ impl UniffiCustomTypeConverter for NewCustomType {
 
 After:
 
-```
+```rust
 uniffi::custom_type!(NewCustomType, BridgeType, {
     try_lift: |val| { Ok(...) },
     lower: |obj| { ... },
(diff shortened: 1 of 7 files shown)


docs/manual/src/Upgrading.md
```diff
@@ -61,6 +60,20 @@ Edge-cases broken include:
...
+```
+[External="crate_name"]
+typedef extern Url;
+```
+
+But now you must use:
+```
+[Custom="crate_name"]
+typedef string Url; // replace `string` with any appropriate type.
+```
```
bendk, line 75: I haven't looked over everything, but I think the main question is how this should be spelled out in the UDL.  The main alternative in my mind is something like:

```
[External="crate_name"]
typedef custom Url;
```

The upside is that this looks like other external type definitions, the downside is that it doesn't define the bridged type.  Maybe that's okay though.  I think we can get that info from the metadata when `--library` is used.

The other case is when `--library` isn't used.  In that case, maybe users can specify the bridged type with another attribute:

```
[External="crate_name", BridgeType="string"]
typedef custom Url;
```

This feels worse than the proposed syntax, but IMO bindgen without library mode is outdated and we should start deprecating it and/or making it somewhat of a second-class citizen.

mhammond: > The upside is that this looks like other external type definitions, the downside is that it doesn't define the bridged type. Maybe that's okay though. I think we can get that info from the metadata when --library is used.

I don't think this is true - I actually started thinking it was, and started making the builtin elt of Custom an Option<Type> - the problem is that this falls apart in the Rust scaffolding - it needs to know the type - even for external types - to know how to lower etc it.

>This feels worse than the proposed syntax, but IMO bindgen without library mode is outdated

Agree entirely - I really would prefer to have found a way to not specify the type, but failed (a few times actually :)

mhammond: > the problem is that this falls apart in the Rust scaffolding

oops, I should have clarified - we work around that in the rust scaffolding today by assuming a `RustBuffer`, but that falls apart in #2025.

bendk: I tried for a while to make this work.  I made some things work, but I hit a fundamental roadblock: if the user is not using library mode, then we need to generate the bindings directly from the UDL.  In that case, there needs to be a way in the UDL to specify the builtin-type, since that's going to determine the FFI type.  This proposal seems as good as any other to me.

bendk (approved):

PR #2374, opened 2024-12-31
Fix docstring tests for Python 3.13

bendk:


mhammond (approved):
That's a bit annoying, but this seem great!

PR #2375, opened 2024-12-31
Fix typo: generatd -> generated

brollb:


bendk (approved):
Nice catch, let's merge this.

PR #2376, opened 2024-12-31
Refactor group type converter

mhammond:
This is a refactor of the group.rs converter/walker - it tries to split the "walking" from the "converting". It should have no functional changes, but should make it easier to perform different kinds of walking in the future.

### DIFF
--- a/uniffi_meta/src/group.rs
+++ b/uniffi_meta/src/group.rs
@@ -47,9 +47,8 @@ pub fn group_metadata(group_map: &mut MetadataGroupMap, items: Vec<Metadata>) ->
             continue;
         }
 
-        let crate_name = calc_crate_name(item.module_path()).to_owned(); // XXX - kill clone?
-
-        let item = fixup_external_type(item, group_map);
+        let crate_name = calc_crate_name(item.module_path()).to_owned();
+        let item = convert_external_metadata_item(item, group_map);
         let group = match group_map.get_mut(&crate_name) {
             Some(ns) => ns,
             None => bail!("Unknown namespace for {item:?} ({crate_name})"),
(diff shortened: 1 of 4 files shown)


uniffi_meta/src/group.rs
```diff
@@ -177,121 +251,35 @@ impl<'a> ExternalTypeConverter<'a> {
...
-{
-    let is_external = |module_path: &str| calc_crate_name(module_path) != owner_module_path;
-    match ty {
-        // Convert `ty` if it's external
-        Type::Enum { module_path, name } | Type::Record { module_path, name }
-            if is_external(&module_path) =>
-        {
-            Type::External {
-                namespace: crate_to_namespace(&module_path),
+        // first, we reeurse
```
bendk, line 254: NIT: recurse.

bendk (approved):
Looks right to me, I'm for merging this one.

I do have some random thoughts about visitors and converters though, since I've been writing a bunch of that code lately.  I read a comment a while ago that mentioned that visitors can always be transformed into an iterator which would avoid having to use inversion of control.  I wonder what this would look like if it was modeled like that.  Rather than the `TypeConverter`  trait you'd have something like `Metadata::iter_types_mut` which would walk metadata item, generate `&mut Type`s.  Then the type conversion would be iterating over those items and mutating them -- maybe changing the variant.  Would that be an improvement?  I'm not really sure and I'm not sure it's even worth the time to experiment with, I just wanted to throw that that idea and see what you thought about it.

mhammond:
That's an interesting idea and something we should experiment with for some of the other visitor work we are doing, and is something we could do here later for sure!

PR #2378, opened 2025-01-01
Don't add builtins and composite types as named types

mhammond:
For apparently historical reasons, we added, eg, `Type::U8` as a named type with name `u8`. This doesn't make sense for various reasons (these names are never actually treated as names, and if they were, they'd never be correct for all bindings).

This is just a cleanup to help simplify our ComponentInterface. It did however uncover two bugs:

* Type::iter_types neglected to include the CustomType builtin
* Type::get_name() returned None for a CallbackInterface even though they are named.

### DIFF
--- a/uniffi_bindgen/src/interface/universe.rs
+++ b/uniffi_bindgen/src/interface/universe.rs
@@ -71,46 +71,13 @@ impl TypeUniverse {
         if !self.all_known_types.contains(type_) {
             self.all_known_types.insert(type_.to_owned());
         }
-        match type_ {
-            Type::UInt8 => self.add_type_definition("u8", type_)?,
-            Type::Int8 => self.add_type_definition("i8", type_)?,
-            Type::UInt16 => self.add_type_definition("u16", type_)?,
-            Type::Int16 => self.add_type_definition("i16", type_)?,
-            Type::UInt32 => self.add_type_definition("i32", type_)?,
-            Type::Int32 => self.add_type_definition("u32", type_)?,
-            Type::UInt64 => self.add_type_definition("u64", type_)?,
-            Type::Int64 => self.add_type_definition("i64", type_)?,
-            Type::Float32 => self.add_type_definition("f32", type_)?,
-            Type::Float64 => self.add_type_definition("f64", type_)?,
-            Type::Boolean => self.add_type_definition("bool", type_)?,
-            Type::String => self.add_type_definition("string", type_)?,
-            Type::Bytes => self.add_type_definition("bytes", type_)?,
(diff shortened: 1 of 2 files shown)


uniffi_bindgen/src/interface/universe.rs
```diff
@@ -71,46 +71,13 @@ impl TypeUniverse {
...
-                key_type,
-                value_type,
-            } => {
-                self.add_known_type(key_type)
-                    .with_context(|| format!("adding map key type {key_type:?}"))?;
-                self.add_known_type(value_type)
-                    .with_context(|| format!("adding map value type {value_type:?}"))?;
-            }
+        // all sub-types, but we want to skip this type as we just added it above.
+        for sub in type_.iter_types().skip(1) {
```
bendk, line 75: This relies on the type being the first thing generated from the iterator which could fail.  What about removing the `skip` and the `add_known_type` call above?  If that doesn't work, maybe this could be a `filter` with an equality check rather than a skip.

mhammond: > skip and the add_known_type call above?

As per the comment above, that would mean many extra clones which are currently avoided. I'm not sure that matters though - wdyt?

mhammond: > In theory, this could be a breaking change for external bindings authors, is it worth adding a CHANGELOG entry?

TBH I don't see how - IIUC, that theoretical scenario would be that they ask for a known type named "u8" just so they can get back a `Type::U8` - wouldn't they just use the type directly? Or is there something else I'm missing here?

bendk: Ahh, I understand now.  I don't think it matters much but what about splitting out `iter_nested_types()` from `iter_types`?  That avoids the clone and seems less fragile than the `skip(1)`.

mhammond:
I tried to fix renaming of callback interfaces as a kind of ride-along here assuming it was a simple oversight, but that broke in unexpected ways, so I reverted that part of this patch and opened #2381.

uniffi_bindgen/src/interface/universe.rs
bendk: > > In theory, this could be a breaking change for external bindings authors, is it worth adding a CHANGELOG entry?
> 
> TBH I don't see how - IIUC, that theoretical scenario would be that they ask for a known type named "u8" just so they can get back a `Type::U8` - wouldn't they just use the type directly? Or is there something else I'm missing here?

I think that would have to be the chain of events.  I'm fine not adding anything since it seems pretty unlikely, I just wanted to ask the question.

Issue #2382, opened 2025-01-02
Complete Swift packaging example


kriswuollett:
Add an example command to show how a complete Swift package can be build and used as a local path dependency. Helps with usability/education issue hinted at in #2362.

kriswuollett:
What I'm expecting to be able to do can be seen in this script: https://gist.github.com/kriswuollett/ec360f943410e09d7fb38e83ae822f09.

Basically given a package `whole` that depends on package `part_a` and `part_b` (and some types in `part_a` depend on `part_b`, I'd have the code for `whole` as:

```rust
pub use part_a;
pub use part_b;
```

Then I run:

```shell
export RUST_PACKAGE="whole"
export RUST_PACKAGE_DIR="crates/$RUST_PACKAGE"
export XCFRAMEWORK_MODULE_NAME="WholeFFI"

./uniffi-bindgen-swift.sh
```

It yields a file structure like:

```
swift-package
├── WholeFFI.xcframework
│   ├── Info.plist
│   ├── ios-arm64
│   │   ├── Headers
│   │   │   ├── part_aFFI.h
│   │   │   ├── part_bFFI.h
│   │   │   └── module.modulemap
│   │   └── libwhole.a
│   ├── ios-arm64_x86_64-simulator
│   │   ├── Headers
│   │   │   ├── part_aFFI.h
│   │   │   ├── part_bFFI.h
│   │   │   └── module.modulemap
│   │   └── libwhole.a
│   └── macos-arm64_x86_64
│       ├── Headers
│       │   ├── part_aFFI.h
│       │   ├── part_bFFI.h
│       │   └── module.modulemap
│       └── libwhole.a
├── Package.swift
└── Sources
    ├── PartAFFI
    │   └── PartAFFI.swift
    └── PartBFFI
        └── PartBFFI.swift
```

And `Package.swift` file:

```swift
// swift-tools-version:6.0
import PackageDescription

let package = Package(
    name: "WholeFFI",
    platforms: [.iOS(.v18), .macOS(.v15)],
    products: [
        .library(
            name: "PartAFFI", targets: ["PartAFFI"]
        ),
        .library(
            name: "PartBFFI", targets: ["PartBFFI"]
        )
    ],
    targets: [
        .target(
            name: "PartAFFI",
            dependencies: [
                .target(name: "WholeFFI"),
                .target(name: "PartBFFI")
            ]
        ),
        .target(
            name: "PartBFFI",
            dependencies: [
[...]

PR #2383, opened 2025-01-02
wip: example binary to build swift package (#2382)

kriswuollett:
Closes: #2382


kriswuollett:
An alternative to a command written in Rust would just be a Makefile. But in either case, for it to be realistically used and maintained, it would depend on Swift 6 fixes and binary published in cargo landing.

PR #2385, opened 2025-01-03
CustomTypes are no longer turned into ExternalTypes

### LINKED ISSUE #2025: Custom Types from external crate do not generate correct binding function signature when used in exports

mhammond:
The bindings all work when the Type::Custom has a different module_path and external bindings will need to do the same.

Fixes #2025

(There's no urgency at all with this, just wanted to get it up as I was happy to see how close things are getting to killing `Type::External`!)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@ or use UDL with types from more than one crate.
 
 - Added the `FfiType::MutReference` variant.
 
+- `Type::Custom` is no longer turned in to a `Type::External` for the bindings. Binding authors
+  must check the type is local themselves before deciding to treat it as a local or external type.
+  We expect `Type::External` will be removed soon and all types will be treated in this way.
+
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.28.3...HEAD).
 
 ## v0.28.3 (backend crates: v0.28.3) - (_2024-11-08_)
--- a/fixtures/ext-types/custom-types/src/lib.rs
+++ b/fixtures/ext-types/custom-types/src/lib.rs
@@ -61,6 +61,15 @@ pub fn get_ouid(ouid: Option<Ouid>) -> Ouid {
     ouid.unwrap_or_else(|| Ouid("Ouid".to_string()))
 }
 
+// A custom-type wrapping a simple ffitype which is also consumed as an external type (#2025)
+pub struct HandleU8(pub u8);
+uniffi::custom_newtype!(HandleU8, u8);
+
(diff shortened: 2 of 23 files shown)


CHANGELOG.md
```diff
@@ -50,6 +50,10 @@ or use UDL with types from more than one crate.
 
 - Added the `FfiType::MutReference` variant.
 
+- `Type::Custom` is no longer turned in to a `Type::External` for the bindings. Binding authors
+  must check the type is local themselves before deciding to treat it as a local or external type.
+  We expect `Type::External` will be removed soon and all types will be treated in this way.
```
bendk, line 55: An alternative to this would be to lean in to `Type::External` rather than away from it:
 - Remove the `module_path` from each type variant -- types are assumed to be from the current module
 - `Type::External` contains a `module_path` field and an `inner` field that can wraps another variant.

I like your approach better though.  The code changes seem simpler.  Also, maybe we can handle the internal/external distinction at another level.  For example, in my bindings pipeline PR, we could add a `TypeDefinition::External` variant.  I'm not sure exactly why, but that feels better to me.

bendk (approved):
Seems like a good direction to me.  I had some thoughts about other approaches that I left in a comment but TLDR I like this change.

CHANGELOG.md
mhammond: I agree with this and think having the bindings themselves make the distinction is a good one - it's much like what this patch already does - `Type::Custom` isn't represented by `Type::External`, but there *is* still external type templates for used for it. You will probably not be too surprised to hear I've got another PR on top of this to kill `Type::External` entirely, but that still uses the same approach - the external templates still exist.

PR #2387, opened 2025-01-03
Remove Type::External

mhammond:
There are 2 commits here for ease of review - the first stops using it, the second removes it entirely. It probably makes sense to land each commit individually rather than squashing.

Assuming this is approved and lands, I intend updating the changelog to point at a few commits to help external binding authors understand better what was changed.


uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -742,16 +738,19 @@ mod filters {
...
-                kind: ExternalKind::DataClass,
-                name,
-                ..
-            }) => {
-                // Need to convert the RustBuffer from our package to the RustBuffer of the external package
-                let suffix = KotlinCodeOracle.class_name(ci, &name);
-                format!("{call}.let {{ RustBuffer{suffix}.create(it.capacity.toULong(), it.len.toULong(), it.data) }}")
+            Some(return_type) => {
+                match return_type {
+                    Type::Enum { name, .. } | Type::Record { name, .. }
```
bendk, line 741: Maybe this could be better future-proofed by matching on `return_type.ffi_type()`.  The extra conversion can kick in if that matches `FfiType::RustBuffer(Some(metadata))`.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -317,17 +317,22 @@ impl ComponentInterface {
...
     }
 
-    /// Get details about all `Type::Custom` types
-    pub fn iter_custom_types(&self) -> impl Iterator<Item = (&String, &Type)> {
-        self.types.iter_known_types().filter_map(|t| match t {
-            Type::Custom { name, builtin, .. } => Some((name, &**builtin)),
-            _ => None,
-        })
+    /// Iterate over all known local types in the interface.
+    pub fn iter_types(&self) -> impl Iterator<Item = &Type> {
```
bendk, line 321: Maybe this should change to `iter_local_types` too.  That means we need to update all the code that calls it, but that doesn't seem terrible to me.  The behavior is changing, so it's a good opportunity for us to double check that the code still makes sense.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -375,16 +375,8 @@ impl ComponentInterface {
...
-                t,
-                Type::Object { .. }
-                    | Type::External {
-                        kind: ExternalKind::Interface,
-                        ..
-                    }
-            )
-        })
+        self.iter_types_in_item(item)
+            .any(|t| matches!(t, Type::Object { .. }))
```
bendk, line 379: Is this still broken with external records?  I'm not sure, but maybe the comment above can be removed.

bendk (approved):
Very nice!  I love all the deleted code.  I left a couple minor comments, I'll let you decide how to resolve them.

uniffi_bindgen/src/interface/mod.rs
mhammond: I was torn about this, so now you mentioned it it flipped me back over the edge :)

PR #2389, opened 2025-01-06
Callables now iterate their "throws" type.

mhammond:
Was playing with external errors and stumbled on this.

I tried adding `iter_types` to `Callable`, but got a bit stuck with lifetimes, so here we are


bendk (approved):
LGTM.

PR #2390, opened 2025-01-07
Get external errors working for Swift and Python

mhammond:
Almost - Swift+Python work. Kotlin doesn't work (#2392), however it works around failing simply by declining to render any functions with external errors, which seems reasonable for now.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
@@ -576,6 +576,15 @@ impl<T: AsType> AsCodeType for T {
     }
 }
 
+// A work around for #2392 - we can't handle functions with external errors.
+fn can_render_callable(callable: &dyn Callable, ci: &ComponentInterface) -> bool {
+    // can't handle external errors.
+    callable
+        .throws_type()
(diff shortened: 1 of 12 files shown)


uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -576,6 +576,15 @@ impl<T: AsType> AsCodeType for T {
     }
 }
 
+// A work around for #2392 - we can't handle functions with external errors.
+fn can_render_callable(callable: &dyn Callable, ci: &ComponentInterface) -> bool {
```
mhammond, line 580: @bendk this is what I was thinking re our discussions around tests etc - at some point it seems reasonable for binding generators to simply decline to render things they can't handle - eg, I could imagine a new bindgen might start by declining to implement functions which try and use anything other than primitive types as they are starting out etc.

bendk (approved):
LGTM, hopefully we can get kotlin working somewhat soon, do you think that's a requirement for the release?

mhammond:
> do you think that's a requirement for the release?

I don't think so, no. I'm actually becoming convinced it's really not at all trivial to fix given how our "companion" objects and external rust buffers are setup (but I could be very wrong about that)

PR #2395, opened 2025-01-10
Fix optional features in `uniffi` Cargo.toml

yukibtc:
Fix the following issue: "the package `<package>` depends on `uniffi`, with features: `uniffi_bindgen` but `uniffi` does not have these features."


mhammond (approved):
Thanks!

PR #2397, opened 2025-01-11
Remove uniffi::generate_component_scaffolding

mhammond:
It's a direct re-export of uniffi_bindgen::generate_component_scaffolding and behind the 'bindgen' feature. For this reason it's not actually used in build scripts - they use 'uniffi::generate_scaffolding' which is correctly behind the 'build' feature.

We can't commit the "bindgen" feature to generating scaffolding, and it's kinda confusing, so let's kill it.


uniffi_build/src/lib.rs
```diff
@@ -39,9 +39,6 @@ pub fn generate_scaffolding_for_crate(
     // The UNIFFI_TESTS_DISABLE_EXTENSIONS variable disables some bindings, but it is evaluated
     // at *build* time, so we need to rebuild when it changes.
     println!("cargo:rerun-if-env-changed=UNIFFI_TESTS_DISABLE_EXTENSIONS");
-    // Why don't we just depend on uniffi-bindgen and call the public functions?
```
mhammond, line 42: this comment was explaining why we do *not* do exactly what we do :)

badboy (approved):

PR #2399, opened 2025-01-13
Fixup lifetimes in Callable/ResultType

mhammond:
Changing `return_type` and `throws_type` to be references is more consistent and allows the trait to implement `iter_types`.

I came across this recently and thought I'd have a poke and it turned out to be easier than I thought.

### DIFF
--- a/uniffi_bindgen/src/interface/object.rs
+++ b/uniffi_bindgen/src/interface/object.rs
@@ -395,6 +395,8 @@ pub struct Constructor {
     // Force a checksum value, or we'll fallback to the trait.
     #[checksum_ignore]
     pub(super) checksum: Option<u16>,
+    // to help with lifetimes elsewhere...
+    pub(super) self_type: Type,
 }
 
 impl Constructor {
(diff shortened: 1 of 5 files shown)


uniffi_bindgen/src/interface/object.rs
```diff
@@ -494,6 +492,7 @@ impl From<uniffi_meta::ConstructorMetadata> for Constructor {
             throws: meta.throws.map(Into::into),
             checksum_fn_name,
             checksum: meta.checksum,
+            self_type,
```
bendk, line 495: I like this change and I think we could take it even further.  If all type-like structs contain a field like `type_`, `self_type` , `ty`, etc.  Then `AsType` can be reduced to `AsRef<Type>`.  Similarly, we could define a `Callable` struct and then change the `Callable` trait to `AsRef<Callable>`.

I'm just mentioning this as food for thought, I don't think we need to do that in this PR.

bendk (approved):
Seems like a nice improvement to me, let's do it.

PR #2403, opened 2025-01-15
Add config option to omit the checking of library checksums.

### LINKED ISSUE #2255: Make checksums optional
We run the uniffi bindgen with Bazel, and our rules ensure there can be no version mismatch between the Rust and foreign language bindings. This makes checksums purely redundant.

We would love to have the option to disable checksums entirely.

mhammond:
This code started showing up in profiles, so my motivation to get this in our next release went up.

Kinda hate "omit" (it seems a somewhat obscure word?) so open to all bikesheds. Started with "skip_checksums", but then saw "omit" in swift so reluctantly went for consistency there.

Doesn't always avoid generating the checksum functions, but does avoid calling them.

Fixes #2255

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,10 @@ or use UDL with types from more than one crate.
 
 - Proc-macros now allow Enums to hold objects (#1372)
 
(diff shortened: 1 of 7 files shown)


bendk (approved):
LGTM.

PR #2405, opened 2025-01-17
Remove unused deps

yukibtc:
Remove `bincode` and `bytes` deps


bendk (approved):
Thanks for noticing this one, seems like a no-brainer.

PR #2406, opened 2025-01-17
core: reduce monomorphization of the futures code

yukibtc:
These changes will reduce the monomorphization of the futures code, resulting in a reduction of the binary size.

In my case, the binaries for android targets have become 100/200kb smaller.

### DIFF
--- a/uniffi_core/src/ffi/rustfuture/future.rs
+++ b/uniffi_core/src/ffi/rustfuture/future.rs
@@ -76,6 +76,7 @@
 //! [`RawWaker`]: https://doc.rust-lang.org/std/task/struct.RawWaker.html
 
 use std::{
+    future::Future,
     marker::PhantomData,
     ops::Deref,
     panic,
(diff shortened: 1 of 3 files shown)


bendk:
Thanks for the patch it's a really interesting idea.  I think we should try to think through the trade-offs here a bit more before landing this one.

I hope to put together a bunchmark on async call overhead.  This is my main question: how much overhead does boxing the futures add per-call?  The other half of the trade-off is the binary size, can you share some details about your project so we can understand that better?  Like, how many async functions are you exporting?

Once we know both of those, I think we'll be in a much better position to judge this one.  Maybe the boxing overhead will be small enough that we can just hit the merge button.  Maybe it's a real trade-off and we should add a config option for this.

yukibtc:
> I hope to put together a bunchmark on async call overhead. This is my main question: how much overhead does boxing the futures add per-call?

I'll try to add some benchmark soon.

> The other half of the trade-off is the binary size, can you share some details about your project so we can understand that better? Like, how many async functions are you exporting?

This is the project: https://github.com/rust-nostr/nostr/tree/d030b776f521fc2bb032f9490bab89e3bfbb5108/bindings/nostr-sdk-ffi

I tried to count the `async fn` with a script and the result is `117` functions.

This is the result of `cargo bloat --release -n 1000` of my project compiled for `x86_64-unknown-linux-gnu` target using `uniffi` b46ab54ce0513dd4a9292e097e17e7ca28827795:

[cargo-bloat-uniffi-b46ab54.txt](https://github.com/user-attachments/files/18502531/cargo-bloat-uniffi-b46ab54.txt)

While this is the result using the patch (`uniffi` a6c2f678b43c15df6969ba5f50aa76523e3f3f79):

[cargo-bloat-uniffi-a6c2f67.txt](https://github.com/user-attachments/files/18502533/cargo-bloat-uniffi-a6c2f67.txt)

yukibtc:
> I'll try to add some benchmark soon.

I tried to take a look at the current benches but I'm not sure how the new ones should be added

bendk:
Thanks for those details, it's very interesting to go through.  I'll try to put together an async benchmark in the next week or two.

bendk:
> Thanks for those details, it's very interesting to go through. I'll try to put together an async benchmark in the next week or two.

Well, one month later and those benchmarks still aren't up.  I still really want to put them together, but I've had to focus on other priorities.  OTOH, you have solid numbers that this is an improvement for your use case.  I'm leaning towards merge, but I'd like to get the opinion of some other UniFFI devs.  @badboy @mhammond what do you think?

bendk:
@yukibtc can you update the PR.  I think https://github.com/mozilla/uniffi-rs/pull/2418 indroduced some conflicts, but I don't think they're hard to fix.

mhammond:
I think I lean towards taking them simply because there's a known problem fixed by it, and only a lack of certainty without real evidence about side-effects. However, that lack of certainty means that I don't think it should be in a 0.29 version (or at least not the next one, but instead it should have time to bake on main)

yukibtc:
Rebased and fixed the conflicts

bendk (approved):
Thanks for the code, let's merge this one.

PR #2418, opened 2025-01-27
Relax `Send` bound for `Future` for wasm32 targets.

### LINKED ISSUE #2417: RustFuture does not compile for the `wasm32-unknown-unknown` target
The [`jplatte/future-wasm` branch](https://github.com/mozilla/uniffi-rs/tree/jplatte/futures-wasm) contains a fixture designed to compile under wasm32.

```sh
error: future cannot be sent between threads safely
  --> fixtures/futures-wasm/src/lib.rs:9:1
   |
9  | #[uniffi::export]
   | ^^^^^^^^^^^^^^^^^ future created by async block is not `Send`
   |
   = help: the trait `Send` is not implemented for `(dyn FnMut() + 'static)`, which is required by `{async block@fixtures/futures-wasm/src/lib.rs:9:1: 9:18}: Send`
note: future is not `Send` as it awaits another future which is not `Send`
  --> fixtures/futures-wasm/src/lib.rs:11:5
   |
11 |     TimeoutFuture::new(2_000).await;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^ await occurs here on type `TimeoutFuture`, which is not `Send`
note: required by a bound in `rust_future_new`
  --> /Users/jhugman/workspaces/mozilla/uniffi-rs/uniffi_core/src/ffi/rustfuture/mod.rs:47:52
   |
41 | pub fn rust_future_new<F, T, UT>(future: F, tag: UT) -> Handle
   |        --------------- required by a bound in this function
...
47 |     F: Future<Output = Result<T, LiftArgsError>> + Send + 'static,
   |                                                    ^^^^ required by this bound in `rust_future_new`
   = note: this error originates in the attribute macro `uniffi::export` (in Nightly builds, run with -Z macro-backtrace for more info)```
```

`wasm-bindgen` issue: ["Why are futures generated by `wasm-bindgen` not `Send`?"](https://github.com/rustwasm/wasm-bindgen/issues/2833) is closed with: "at this time I don't know of anything that can be done"

This issue is to remove `Send` as a requirement for Futures for `wasm32` targets.

jhugman:
Fixes #2417.

This PR introduces a marker trait for the Futures that uniffi supports.

This is a small refactor which tidies up 7 places where it was used:

```rust
pub trait FutureResult<T, E>: Future<Output = Result<T, E>> {}

#[cfg(not(target_arch = "wasm32"))]
impl<T, F, E> FutureResult<T, E> for F where F: Future<Output = Result<T, E>> + Send + 'static {}
```

It also adds a second `impl`, only active during `wasm32`:

```rust
// `Promise`s cannot be sent to or from WebWorkers, but `JsValue` is not `Send`.
#[cfg(target_arch = "wasm32")]
impl<T, F, E> FutureResult<T, E> for F where F: Future<Output = Result<T, E>> + 'static {}
```

I am not sure:
- naming: should it be `RustFutureResult` or `UniffiFutureResult` or … something else?
- how to introduce wasm32 into the existing CI to verify that this has fixed the original issue.

### DIFF
--- a/uniffi_core/src/ffi/rustfuture/future.rs
+++ b/uniffi_core/src/ffi/rustfuture/future.rs
@@ -76,7 +76,6 @@
 //! [`RawWaker`]: https://doc.rust-lang.org/std/task/struct.RawWaker.html
 
 use std::{
-    future::Future,
     marker::PhantomData,
     ops::Deref,
     panic,
@@ -85,14 +84,14 @@ use std::{
     task::{Context, Poll, Wake},
 };
 
-use super::{RustFutureContinuationCallback, RustFuturePoll, Scheduler};
+use super::{FutureResult, RustFutureContinuationCallback, RustFuturePoll, Scheduler};
 use crate::{rust_call_with_out_status, FfiDefault, LiftArgsError, LowerReturn, RustCallStatus};
 
 /// Wraps the actual future we're polling
 struct WrappedFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: Future<Output = Result<T, LiftArgsError>> + Send + 'static,
+    F: FutureResult<T, LiftArgsError>,
     T: LowerReturn<UT> + Send + 'static,
     UT: Send + 'static,
 {
@@ -106,7 +105,7 @@ where
 impl<F, T, UT> WrappedFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: Future<Output = Result<T, LiftArgsError>> + Send + 'static,
+    F: FutureResult<T, LiftArgsError>,
     T: LowerReturn<UT> + Send + 'static,
     UT: Send + 'static,
 {
@@ -186,7 +185,7 @@ where
 unsafe impl<F, T, UT> Send for WrappedFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: Future<Output = Result<T, LiftArgsError>> + Send + 'static,
+    F: FutureResult<T, LiftArgsError>,
     T: LowerReturn<UT> + Send + 'static,
     UT: Send + 'static,
 {
@@ -196,7 +195,7 @@ where
 pub(super) struct RustFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: Future<Output = Result<T, LiftArgsError>> + Send + 'static,
+    F: FutureResult<T, LiftArgsError>,
     T: LowerReturn<UT> + Send + 'static,
     UT: Send + 'static,
 {
@@ -212,7 +211,7 @@ where
 impl<F, T, UT> RustFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: Future<Output = Result<T, LiftArgsError>> + Send + 'static,
+    F: FutureResult<T, LiftArgsError>,
     T: LowerReturn<UT> + Send + 'static,
     UT: Send + 'static,
 {
@@ -267,7 +266,7 @@ where
 impl<F, T, UT> Wake for RustFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: Future<Output = Result<T, LiftArgsError>> + Send + 'static,
+    F: FutureResult<T, LiftArgsError>,
     T: LowerReturn<UT> + Send + 'static,
     UT: Send + 'static,
 {
@@ -302,7 +301,7 @@ pub trait RustFutureFfi<ReturnType>: Send + Sync {
 impl<F, T, UT> RustFutureFfi<T::ReturnType> for RustFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: Future<Output = Result<T, LiftArgsError>> + Send + 'static,
+    F: FutureResult<T, LiftArgsError>,
     T: LowerReturn<UT> + Send + 'static,
     UT: Send + 'static,
 {
--- a/uniffi_core/src/ffi/rustfuture/future.rs
+++ b/uniffi_core/src/ffi/rustfuture/future.rs
@@ -84,14 +84,14 @@ use std::{
     task::{Context, Poll, Wake},
 };
 
-use super::{FutureResult, RustFutureContinuationCallback, RustFuturePoll, Scheduler};
+use super::{RustFutureContinuationCallback, RustFuturePoll, Scheduler, UniffiCompatibleFuture};
 use crate::{rust_call_with_out_status, FfiDefault, LiftArgsError, LowerReturn, RustCallStatus};
 
 /// Wraps the actual future we're polling
 struct WrappedFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: FutureResult<T, LiftArgsError>,
+    F: UniffiCompatibleFuture<Result<T, LiftArgsError>> + 'static,
     T: LowerReturn<UT> + Send + 'static,
     UT: Send + 'static,
 {
@@ -105,7 +105,7 @@ where
 impl<F, T, UT> WrappedFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: FutureResult<T, LiftArgsError>,
+    F: UniffiCompatibleFuture<Result<T, LiftArgsError>> + 'static,
     T: LowerReturn<UT> + Send + 'static,
     UT: Send + 'static,
 {
@@ -185,7 +185,7 @@ where
 unsafe impl<F, T, UT> Send for WrappedFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: FutureResult<T, LiftArgsError>,
+    F: UniffiCompatibleFuture<Result<T, LiftArgsError>> + 'static,
     T: LowerReturn<UT> + Send + 'static,
     UT: Send + 'static,
 {
@@ -195,7 +195,7 @@ where
 pub(super) struct RustFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: FutureResult<T, LiftArgsError>,
+    F: UniffiCompatibleFuture<Result<T, LiftArgsError>> + 'static,
     T: LowerReturn<UT> + Send + 'static,
     UT: Send + 'static,
 {
@@ -211,7 +211,7 @@ where
 impl<F, T, UT> RustFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: FutureResult<T, LiftArgsError>,
+    F: UniffiCompatibleFuture<Result<T, LiftArgsError>> + 'static,
     T: LowerReturn<UT> + Send + 'static,
     UT: Send + 'static,
 {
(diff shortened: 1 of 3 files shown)


uniffi_core/src/ffi/rustfuture/future.rs
```diff
@@ -106,7 +105,7 @@ where
 impl<F, T, UT> WrappedFuture<F, T, UT>
 where
     // See rust_future_new for an explanation of these trait bounds
-    F: Future<Output = Result<T, LiftArgsError>> + Send + 'static,
+    F: FutureResult<T, LiftArgsError>,
```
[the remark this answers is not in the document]
bendk: Something like that could actually work with my naming suggestion of `UniffiFutureBounds`.  Maybe that could replace the `Send` bound here and everything else could stay the same.

bendk (approved):
The code changes look good to me.  I think we just need to figure out how to name/document things.

bendk (changes requested):
Oops, didn't mean to hit the approve button so fast.  Even though, I do think the code is fine.

uniffi_core/src/ffi/rustfuture/mod.rs
```diff
@@ -30,6 +30,17 @@ pub enum RustFuturePoll {
 /// to continue progress on the future.
 pub type RustFutureContinuationCallback = extern "C" fn(callback_data: u64, RustFuturePoll);
 
+/// This marker trait allows us to put different bounds on the `Future`s we support, based on
+/// `#[cfg()]` configuration.
+pub trait FutureResult<T, E>: Future<Output = Result<T, E>> + 'static {}
```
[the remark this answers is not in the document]
jhugman: I had `UniffiCompatibleFuture` as well, as `FutureCompat` and `UniffiFutureCompat`.

I ended up at `FutureResult<T, E>` as it is most descriptive: it is at its base `Future<Result<T, E>>`.

```rust
pub trait UniffiFutureResult<T, E>: Future<Output = Result<T, E>> {}
```

I _don't_ like the `FutureResult<T,E>` because it's exposed in the public API, and feels like a good name for useful `struct`, like people should be using it but being frustrated that they can't because of the orphan rule, so I'm going to go with `UniffiCompatibleFuture`.

mhammond:
I'd love to see Jonas's thoughts here too.

uniffi_core/src/ffi/rustfuture/mod.rs
```diff
@@ -30,6 +30,32 @@ pub enum RustFuturePoll {
...
+///
+/// WASM32 is a single threaded environment. However, in a browser there do exist
+/// [`WebWorker`](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers)s which do
+/// not share memory or event-loop with the main browser context.
+///
+/// The most common source of asynchrony,
+/// [`JsPromise`](https://github.com/rustwasm/wasm-bindgen/blob/main/crates/futures/src/lib.rs)
+/// is not `Send` because:
+///
+/// 1. `T`` and `E`` are both `JsValue`
```
jplatte, line 50:
```suggestion
/// 1. `T` and `E` are both `JsValue`
```

uniffi_core/src/ffi/rustfuture/mod.rs
```diff
@@ -30,6 +30,32 @@ pub enum RustFuturePoll {
...
+/// 2. `JsValue` may contain `JsFunction`s, either as a function themselves or an object containing functions.
+/// 3. Functions cannot be serialized and sent to `WebWorker`s.
+///
+/// The polling and waking happens in the binding language, i.e. the main browser context, i.e. the same thread
```
jplatte, line 54: "polling and waking happens in the binding language, i.e. the main browser context" - I don't know much about web workers, what prevents users from calling UniFFI stuff from within one?

uniffi_core/src/ffi/rustfuture/future.rs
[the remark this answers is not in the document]
jplatte: Agree about putting the `'static` into the supertraits.

uniffi_core/src/ffi/rustfuture/mod.rs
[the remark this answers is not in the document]
jplatte: Why would this trait be generic at all? I'd make it a `trait UniffiFuture` that exactly mirrors `Future` (output not necessarily being `Result<T, E>`), except with the required supertraits.

uniffi_core/src/ffi/rustfuture/mod.rs
```diff
@@ -30,6 +30,32 @@ pub enum RustFuturePoll {
...
 
+/// This marker trait allows us to put different bounds on the `Future`s we support, based on
+/// `#[cfg()]` configuration.
+pub trait UniffiCompatibleFuture<T, E>: Future<Output = Result<T, E>> {}
+
+#[cfg(not(target_arch = "wasm32"))]
+impl<T, F, E> UniffiCompatibleFuture<T, E> for F where F: Future<Output = Result<T, E>> + Send {}
+
+/// # Safety:
+///
```
bendk, line 41: Nit: How about mentioning the `Send` bound at the top?  The comment is long enough that readers might not see the actual type alias.  Maybe something like "Future's on WASM32 are not `Send` because it's a single threaded environment".

uniffi_core/src/ffi/rustfuture/mod.rs
```diff
@@ -30,6 +30,32 @@ pub enum RustFuturePoll {
...
+/// WASM32 is a single threaded environment. However, in a browser there do exist
+/// [`WebWorker`](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers)s which do
+/// not share memory or event-loop with the main browser context.
+///
+/// The most common source of asynchrony,
+/// [`JsPromise`](https://github.com/rustwasm/wasm-bindgen/blob/main/crates/futures/src/lib.rs)
+/// is not `Send` because:
```
bendk, line 48: This seems like the most important point here.  We're just following the rules that wasm-bindgen already set, I'd love that to be emphasized.  Maybe something like "UniFFI is following the path of the most common source of asynchrony, ..."

uniffi_core/src/ffi/rustfuture/mod.rs
bendk: Good question, can you poll a future in a web worker?  I think that's still safe, since the model seems to be that you send messages across threads rather than share the futures.  Maybe it's more accurate to say "polling and waking happens in the same thread the future was created in".

uniffi_core/src/ffi/rustfuture/future.rs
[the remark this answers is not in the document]
bendk: `'static` is currently present in this code.  The current version has it staying here rather than moving it out.  That said, maybe it does make sense to move it out.  `'static` is a bound just like `Send`, so why not have them together?  I don't really have strong feelings about this either way though.

uniffi_core/src/ffi/rustfuture/mod.rs
```diff
@@ -30,6 +30,32 @@ pub enum RustFuturePoll {
 /// to continue progress on the future.
 pub type RustFutureContinuationCallback = extern "C" fn(callback_data: u64, RustFuturePoll);
 
+/// This marker trait allows us to put different bounds on the `Future`s we support, based on
+/// `#[cfg()]` configuration.
+pub trait UniffiCompatibleFuture<T, E>: Future<Output = Result<T, E>> {}
```
bendk, line 35: There's a couple things here that feel a bit weird: The `T` and `E` params are moving from being part of the associated type to the generic parameters, which kind of implies that you could have a type that implements `UniffiCompatibleFuture` in multiple ways.  Also, users can technically implement this trait on their types, even if they don't have the right bounds.

The more I think about it, the more I like having this trait just express the bounds, not the Future output.

```
#[cfg(not(target_arch = "wasm32"))]
pub trait UniffiCompatibleFuture: Send + 'static {}

// or pub trait UniffiCompatibleFuture: Future + Send + 'static {}


#[cfg(target_arch = "wasm32")]
pub trait UniffiCompatibleFuture: 'static {}
```


... and the bounds in `future.rs` would look like: `F: Future<Output = Result<T, LiftArgsError>> + UniffiCompatibleFuture`

But again, I don't care that much either way since this is just for internal use.  I'd be okay with it staying as-is if we added `#[doc(hidden)]` and a note telling users that this is an internal detail, subject to change, and don't implement it yourself.

bendk (changes requested):
I think we can get this in the release.  The fact that wasm-bindgen does something similar gives me a lot of confidence that this is a good path.

I still had some questions on the trait definition.  Like I said in the comments, I'm okay with any direction as long as we make it clear that this is an internal trait that we can change anytime.  That way we're not really locking ourselves into any decision.

uniffi_core/src/ffi/rustfuture/future.rs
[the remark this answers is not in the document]
jhugman: I've left the `'static` in method type bounds, mainly because there are other arguments that are `'static`. It seemed clearer that way.

uniffi_core/src/ffi/rustfuture/mod.rs
jhugman: I have enumerated the various combinations, which are either possible, or impossible for different reasons.

uniffi_core/src/ffi/rustfuture/mod.rs
```diff
@@ -30,6 +30,55 @@ pub enum RustFuturePoll {
...
+/// exist [`WebWorker`][webworker]s which do not share memory or event-loop
+/// with the main browser context.
+///
+/// Communication between contexts is only possible by message passing,
+/// using a small number of ['transferable' object types][transferable].
+///
+/// The most common source of asynchrony in Rust compiled to WASM is
+/// [wasm-bindgen's `JsFuture`][jsfuture]. It is not `Send` because:
+///
+/// 1. `T`` and `E` are both `JsValue`
```
jplatte, line 56:
```suggestion
/// 1. `T` and `E` are both `JsValue`
```

uniffi_core/src/ffi/rustfuture/mod.rs
```diff
@@ -30,6 +30,55 @@ pub enum RustFuturePoll {
...
+///    an object containing functions.
+/// 3. Functions cannot be [serialized and sent][transferable] to `WebWorker`s.
+///
+/// Implementors of binding generators should be able to enumerate the
+/// combinations of Rust or JS communicating across different contexts (here
+/// using: <->), and in the same context (+) to account for why it is safe
+/// for UniFFI to support `Future`s that are not `Send`:
+///
+/// 1. JS + Rust in the same contexts: polling and waking happens in the same
+///    thread, no Send is needed.
```
jplatte, line 67:
```suggestion
///    thread, no `Send` is needed.
```

jplatte (approved):

uniffi_core/src/ffi/rustfuture/mod.rs
bendk: I'm still very slightly worried that users will try to implement this trait themselves.  Can you add the `#[doc(hidden)]` and/or a note in the docstring that says don't try to implement this yourself?

bendk (changes requested):
This is looking great.  Can you make the docstring suggestion around `UniffiCompatibleFuture` and Jonas's suggestions?  I can merge after that.

uniffi_core/src/ffi/rustfuture/mod.rs
jhugman: Damn you VSCode!

uniffi_core/src/ffi/rustfuture/mod.rs
jhugman: Drat, I didn't see the bit about `#[doc(hidden)]` first time.

jhugman:
Great, thanks @jplatte and @bendk !

PR #2420, opened 2025-01-28
prealloc size of Lower::lower_into_rust_buffer

oriongonza:


uniffi_core/src/ffi_converter_traits.rs
```diff
@@ -245,7 +245,7 @@ pub unsafe trait Lower<UT>: Sized {
 
     /// Convenience method
     fn lower_into_rust_buffer(obj: Self) -> RustBuffer {
-        let mut buf = ::std::vec::Vec::new();
+        let mut buf = ::std::vec::Vec::with_capacity(size_of::<Self>());
```
bendk, line 248: This would be a great optimization, but is `size_of<Self>` the right value?  Our serialization code and Rust's type layout are very different.  Still, maybe `size_of<Self>` is a good estimate -- I'm not really sure.

Can you do some tests?  Like create a set of Records and Enums, some with few fields/variants, some with lots of them, then compare `size_of::<Self>` against `lower_into_rust_buffer(self).len()`.  I'd love to see the results of that.

bendk (changes requested):
Seems like a good idea, but I'd love some more data before we try to make this optimization.

mhammond:
Thanks for the patch - let us know if you want to address the comments above and we can reopen it.

PR #2428, opened 2025-02-04
Fixup comments and whitespace in generated Python code.

mhammond:
This is purely aesthetic and has no functional changes.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/templates/ExternalTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/ExternalTemplate.py
@@ -1,20 +1,20 @@
 {%- let namespace = ci.namespace_for_type(type_)? %}
-{%- let module = python_config.module_for_namespace(namespace) -%}
+{%- let module = python_config.module_for_namespace(namespace) %}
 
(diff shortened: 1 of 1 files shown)


uniffi_bindgen/src/bindings/python/templates/ExternalTemplate.py
```diff
@@ -1,20 +1,20 @@
...
+{{- self.add_import_of(module, name) }} {#- import the type alias itself -#}
 
-# External type {{ name }} is an error.
 {%- if ci.is_name_used_as_error(name) %}
+# External type {{ name }} is an error.
 {%-     match type_ %}
 {%-         when Type::Object { .. } %}
 {%-             let err_ffi_converter_name = "{}__as_error"|format(ffi_converter_name) %}
-{{ self.add_import_of(module, err_ffi_converter_name) }}
+{{- self.add_import_of(module, err_ffi_converter_name) }}
```
mhammond, line 14: mildly surprised this works, but it does!

badboy: Jinja whitespacing remains a mystery to me.

badboy (approved):

PR #2435, opened 2025-02-11
Remove clap color feature

mhammond:
This is a WIP, but I expect something like this might be necessary to use uniffi-bindgen in m-c to avoid the many deps required by the clap `color` feature. An alternative would be to duplicate the cli parsing in m-c and avoid that dependency there.

@badboy @bendk, thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 - Bindings support `lift` and `lower` for CustomTypes in uniffi.toml to match the docs ([#2438](https://github.com/mozilla/uniffi-rs/issues/2438))
 
+### What's changed?
+
+- The uniffi-bindgen CLI support no longer brings in the `clap/color` feature, reducing dependencies ([#2435](https://github.com/mozilla/uniffi-rs/pull/2435))
+
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.29.0...HEAD).
 
 ## v0.29.0 (backend crates: v0.29.0) - (_2025-02-06_)
--- a/uniffi/Cargo.toml
+++ b/uniffi/Cargo.toml
@@ -21,7 +21,8 @@ uniffi_macros = { path = "../uniffi_macros", version = "=0.29.0" }
 anyhow = "1"
 camino = { version = "1.0.8", optional = true }
 cargo_metadata = { version = "0.15", optional = true }
-clap = { version = "4", features = ["cargo", "std", "derive"], optional = true }
+# avoid 'clap/color' due to dependency hell.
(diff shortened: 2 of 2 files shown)


mhammond:
Looking a little more: `clap` is already in m-c due to gecko-driver and 1 other util, which bring it in without defaults. Bringing the color feature brings in way too many "terminal" related modules and duplicates the massive "windows" crate. Even if we could minimize some of that, the supply-chain hassles just aren't worth it. So I think we do need this kind of feature.

Re the feature itself, instead of a feature `cli-color` I changed it to `cli-minimal` and `cli enables "clap/color", "clap/suggestions" etc. They don't bring in extra deps, but it's less code to build on the hosts and seems cleaner. wdyt?

I tried to test this with our builtin binary, but can't work out why, and can't work out why this patch actually works, as you'll see. Any clues? A version of this did work as expected on m-c, but I'll retest this there.

badboy:
In Glean we keep the `cli` feature disabled to fully avoid the clap dependency. It's just way to big for the 3 flags we need.

Should we just keep the color part disabled fully?
Or should we switch away from clap? It is a useful library, but it is heavy and I'm not sure it carries its weight here.

> I tried to test this with our builtin binary, but can't work out why, and can't work out why this patch actually works, as you'll see. Any clues? A version of this did work as expected on m-c, but I'll retest this there.

I'm not sure I understand what this means. What did you test and what did/didn't work?

mhammond:
> I'm not sure I understand what this means. What did you test and what did/didn't work?

Sorry for being vague - in the original patch I removed the `cli` feature entirely from our test bindgen binary and it still magically worked, and I didn't understand why.

But that's moot now, I just removed `color` entirely, which I'm also fine with!

mhammond:
> But that's moot now, I just removed `color` entirely, which I'm also fine with!

Even with this patch though, when I run `cargo run --bin uniffi-bindgen -- help` I still see output with bold and underline, which I'm not expecting.

mhammond:
> Even with this patch though, when I run `cargo run --bin uniffi-bindgen -- help`

I don't understand this. On top of this patch, remove the `cli` feature [here](https://github.com/mozilla/uniffi-rs/blob/84619f51d73f67187c8cf4a535db776cdee540ef/examples/app/uniffi-bindgen-cli/Cargo.toml#L12) - eg [this commit is green](https://github.com/mozilla/uniffi-rs/pull/2435/commits/46530521c203a75654257851ed93fed0cf8ee155) - the above command works in CI and has bold/underline locally. But it's clearly broken.

`cargo run -p uniffi-bindgen-cli -- help` however fails as expected in that case, and correctly does not have the bold/underline otherwise. So I'm quite confused by the differences between those 2 commands and why I was nearly able to land something broken 🤷 😓

bendk (approved):
Just discussed this on the UniFFI call and it seems great.

PR #2437, opened 2025-02-12
Hide `init(unsafeFromRawPointer)` from docs

jkmassel:
This method’s comments indicates it shouldn’t be public API, so let’s exclude it from documentation generation.


mhammond (approved):
looks correct to me, thanks!

PR #2442, opened 2025-02-16
Bindings support `lift` and `lower` for CustomTypes in uniffi.toml to…

### LINKED ISSUE #2438: Documentation for uniffi.toml bindings
Hello,

In the latest version 0.29.0, I followed the changelog:
[Uniffi Changelog](https://github.com/mozilla/uniffi-rs/blob/main/CHANGELOG.md) and performed the necessary migration.

The changelog link to:
[Custom Types Documentation](https://mozilla.github.io/uniffi-rs/next/types/custom_types.html), specifically the section **"Custom types in the bindings code"** at the bottom.

The documentation provides the following example:

```toml
[bindings.kotlin.custom_types.Url]
# Name of the type in the Kotlin code
type_name = "URL"
# Classes that need to be imported
imports = ["java.net.URL"]
# Expression to convert the built-in type to the custom type. `{}` will be replaced with the input value.
lift = "URL({})"
# Expression to convert the custom type to the built-in type. `{}` will be replaced with the URL value.
lower = "{}.toString()"
```

However, after renaming my existing bindings to use `lift/lower` instead of `into_custom/from_custom`, it fails with the following error:

```
Caused by:
    missing field into_custom for key custom_types.Url
error: Failed to run uniffi_bindgen: Command execution failed
```

But it works correctly if I keep using:

```toml
[bindings.kotlin.custom_types.Url]
type_name = "URL"
imports = ["java.net.URL"]
into_custom = "URL({})"
from_custom = "{}.toString()"
```

I'm not sure if this is a mix-up in the documentation, an outdated template, or if I've misunderstood something?
I like the `lift/lower` renaming and it make sense since that's the naming used in the custom_type! macro.

Thanks

mhammond:
… match the docs.

Old names `into_custom` and `from_custom` are still supported. Fixes #2438


bendk (approved):
:rocket:

PR #2443, opened 2025-02-16
Error types in the swift binding conforms to the Error protocol

crazytonyli:
When the `omit_localized_error_conformance` configuration (added in https://github.com/mozilla/uniffi-rs/pull/2319) is turned on, the `Error` conformance was also removed because it was previously implied by the `LocalizedError`.

This PR makes sure error types in Swift bindings always conform to `Swift.Error`, regardless of whether they conform to `LocalizedError`.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
@@ -1,5 +1,5 @@
 {%- call swift::docstring(e, 0) %}
-public enum {{ type_name }} {
+public enum {{ type_name }}: Swift.Error {
 
     {% if e.is_flat() %}
     {% for variant in e.variants() %}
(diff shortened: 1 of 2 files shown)


mhammond:
Wish I could workout why CI is refusing to run!?

crazytonyli:
If I read the CI error log correctly, the failure is in the `uniffi-fixture-futures` package. However, the command `UNIFFI_TESTS_DISABLE_EXTENSIONS=kt,kts,rb,py cargo test --package uniffi-fixture-futures` passes on my mac.

mhammond:
I believe that's an intermittent but unrelated problem. I'll re-trigger it

mhammond:
Just to be a little paranoid because of our ongoing failure to run swift 6 in CI, are you able to verify swift6 also works fine?

crazytonyli:
I have tested this change on my iOS project which is on Swift 6.

crazytonyli:
I should have done the same test in my original PR, though. Not sure how I missed the issue.

mhammond (approved):
Thanks - even though this specific change isn't behind the `omit_local_error_conformance` flag and it is intended to only help those who *do* use that flag, it seems fine to me and CI appears to agree - thanks!

PR #2447, opened 2025-02-21
fix: `#[no_mangle]` error in Rust 2024 Edition

martinmose:
Hey,

I know, I know - Rust 2024 just dropped, and I’m not trying to rush anything. But after upgrading to the Rust 2024 edition, I ran into the following build error:

```rust
error: unsafe attribute used without unsafe
  --> /Users/martinmose/project/target/debug/build/project_shared-0c37e55cf0c6dc4f/out/project_shared.uniffi.rs:96:3
   |
96 | #[no_mangle]
   |   ^^^^^^^^^ usage of unsafe attribute
   |
help: wrap the attribute in `unsafe(...)`
   |
96 | #[unsafe(no_mangle)]
   |   +++++++         +
```

I found this [Unsafe Attributes](https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-attributes.html)

This seems like a simple fix - just wrapping `#[no_mangle]` in `#[unsafe(...)]`. But maybe there is something I’m unaware of?

It looks like it fixes my upgraded Rust project at least 😅

### DIFF
--- a/uniffi_bindgen/src/scaffolding/templates/Checksums.rs
+++ b/uniffi_bindgen/src/scaffolding/templates/Checksums.rs
@@ -1,5 +1,5 @@
 {%- for (name, checksum) in ci.iter_checksums() %}
-#[no_mangle]
+#[unsafe(no_mangle)]
 #[doc(hidden)]
 pub extern "C" fn r#{{ name }}() -> u16 {
     {{ checksum }}
--- a/uniffi_bindgen/src/scaffolding/templates/ReexportUniFFIScaffolding.rs
+++ b/uniffi_bindgen/src/scaffolding/templates/ReexportUniFFIScaffolding.rs
@@ -20,7 +20,7 @@ pub const fn uniffi_reexport_hack() {}
 macro_rules! uniffi_reexport_scaffolding {
     () => {
         #[doc(hidden)]
-        #[no_mangle]
+        #[unsafe(no_mangle)]
         pub extern "C" fn {{ ci.namespace() }}_uniffi_reexport_hack() {
             $crate::uniffi_reexport_hack()
         }
--- a/uniffi_bindgen/src/scaffolding/templates/UdlMetadata.rs
+++ b/uniffi_bindgen/src/scaffolding/templates/UdlMetadata.rs
@@ -9,5 +9,5 @@ const {{ const_udl_var }}: ::uniffi::MetadataBuffer = ::uniffi::MetadataBuffer::
     .concat_str("{{ udl_base_name }}");
 
 #[doc(hidden)]
-#[no_mangle]
+#[unsafe(no_mangle)]
 pub static {{ static_udl_var }}: [u8; {{ const_udl_var }}.size] = {{ const_udl_var }}.into_array();


bendk:
Looks like this doesn't work on older editions though, based on the CI.

Maybe we could add an `#[allow]` to prevent the error?  If not, I guess we'd need some sort of config option where you specify which edition you're targetting.

badboy:
We can't `allow` it, as `#[unsafe(...)]` is not accepted.
This was introduced in Rust 1.82, so unless we upgrade to that as our minimum Rust we can't support it that way.
And you can't allow it on edition 2024, because it's not a lint, but a hard compiler error.

So _if_ we want to support projects using edition 2024 with our generated code we have no other way but to introduce a flag for that so we can generate either the one or the other form.

badboy:
[m-c just updated to 1.82](https://bugzilla.mozilla.org/show_bug.cgi?id=1945020#c23)

mhammond:
> [m-c just updated to 1.82](https://bugzilla.mozilla.org/show_bug.cgi?id=1945020#c23)

But IIUC still has an [MSR of 1.76](https://searchfox.org/mozilla-central/source/python/mozboot/mozboot/util.py#16)

mhammond:
> [Nope](https://github.com/mozilla/gecko-dev/blob/405d0b26a98dec92d8009b38a8ad1c1609475f1e/python/mozboot/mozboot/util.py#L16) (searchfox only updates once a day, so it doesn't have the new change yet)

Oh, fantastic, thanks!

mhammond:
Does every `#[no_mangle]` need this?

badboy:
> Does every `#[no_mangle]` need this?

For now only those in generated code, because the generated code is compiled as part of the user's project, which might be running under edition 2024 mode.
UniFFI crates themselves are not yet using edition 2024, so they won't run into problems for now

mhammond:
I experimented a bit and there are 2 other `#[no_mangle]` references in `uniffi_bindgen/src/scaffolding/templates` which seems to need the same treatment. If you don't mind updating this PR for that, we can now merge this (or if you prefer, I'm happy to roll it up, but I'd rather you get credit for it :)

mhammond:
(there are many in `uniffi_macros` too, and I don't understand why they are OK without the new attribute! However, both the "coveralls" and "proc-macro" fixtures work with 2024 and just the templates updated. That also requires moving our "normal" Rust version to 1.85 - I'll put a PR up for all of that soon)

badboy:
> (there are many in `uniffi_macros` too, and I don't understand why they are OK without the new attribute! However, both the "coveralls" and "proc-macro" fixtures work with 2024 and just the templates updated. 

because of macro hygiene. `uniffi_macros` uses edition 2021 and so everything its macros generate is tagged with edition 2021, even when the crate where the macro is called otherwise uses a different edition.

martinmose:
> I experimented a bit and there are 2 other `#[no_mangle]` references in `uniffi_bindgen/src/scaffolding/templates` which seems to need the same treatment. If you don't mind updating this PR for that, we can now merge this (or if you prefer, I'm happy to roll it up, but I'd rather you get credit for it :)

Nice! It should be updated now.

mhammond:
Could you please also rebase against current main?

martinmose:
> Could you please also rebase against current main?

Yes ofc... 🤦‍♂️  sry

mhammond (approved):
Thanks!

PR #2450, opened 2025-02-21
Fix Swift Sendable conformance for traits.

mhammond:
This takes the same approach as #2318 and #2415, but here we need the protocol to conform to Sendable.

I can't explicitly flag @crazytonyli or @martinmose for review, but I'd love your thoughts.

without this patch:

```
% UNIFFI_TEST_SWIFT_VERSION=6 cargo test -p uniffi-fixture-ext-types-proc-macro
...
error: emit-module command failed with exit code 1 (use -v to see invocation)
/Users/skip/src/moz/uniffi-rs/target/tmp/uniffi_ext_types_proc_macro_lib-74d4b2cfff69a2a/imported_types_lib.swift:617:16: error: stored property 'maybeTrait' of 'Sendable'-conforming struct 'ObjectsType' has non-sendable type '(any UniffiOneTrait)?'
 615 | 
 616 | public struct ObjectsType {
 617 |     public var maybeTrait: UniffiOneTrait?
     |                `- error: stored property 'maybeTrait' of 'Sendable'-conforming struct 'ObjectsType' has non-sendable type '(any UniffiOneTrait)?'
 618 |     public var maybeInterface: UniffiOneInterface?
 619 | 

/Users/skip/src/moz/uniffi-rs/target/tmp/uniffi_ext_types_proc_macro_lib-74d4b2cfff69a2a/uniffi_one_ns.swift:777:17: note: protocol 'UniffiOneTrait' does not conform to the 'Sendable' protocol
 775 | 
 776 | 
 777 | public protocol UniffiOneTrait: AnyObject {
     |                 `- note: protocol 'UniffiOneTrait' does not conform to the 'Sendable' protocol
 778 |     
 779 |     func hello()  -> String
test uniffi_foreign_language_testcase_test_imported_types_swift ... FAILED

```

(huh - but *with* this patch, the above command causes the swift test to hang. Without `UNIFFI_TEST_SWIFT_VERSION=6` the test passes without warning. So I'll leave this as a draft while I try and work that out, but welcome any clues.


mhammond:
There's obviously something I don't understand here. [Our docker version of swift](https://github.com/mozilla/uniffi-rs/blob/c82945bdf04f947215f44609d70d1ec346c662b1/docker/Dockerfile-build#L41) says `6.0.3`, my local `swift --version` reports it is `swift-driver version: 1.115.1 Apple Swift version 6.0.3 (swiftlang-6.0.3.1.10 clang-1600.0.30.1)` - but locally I just see these CI errors as warnings, not errors.

I'm also pushing a fix to the CI warnings here, but I'm not at all confident it will continue to work with swift 5. I'm confused :(

martinmose:
I totally missed #2415 - sorry! Feel free to tag me next time.

I’ve tested #2318 (with my local Uniffi build) against my Swift 6 project, and everything worked fine.
I also tried upgrading to `0.29.0` and ran into some strict concurrency issues.

Back to this PR - it might actually help with what I saw in `0.29.0`.

Regarding Swift 5 support, Sendable was officially added in [Swift 5.7](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0302-concurrent-value-and-concurrent-closures.md), so this shouldn’t break compatibility.

That is weird... Could it be something like strict concurrency checking is enabled differently between environments? Can you check: [SWIFT_STRICT_CONCURRENCY](https://developer.apple.com/documentation/xcode/build-settings-reference#Strict-Concurrency-Checking)?

I've reviewed the code, and it looks good to me!

martinmose (approved):

mhammond:
Test hang is #2451. No idea about CI (`SWIFT_STRICT_CONCURRENCY` isn't set for me nor in CI) or what was going on there, but some test fixes got things green. Futures remain broken in swift 6, but that's for another pr...

bendk (approved):
Code changes look good.  I think we need a changelog/upgrading entry that explains the new requirements now that protocols implement `Sendable`.

PR #2452, opened 2025-02-23
Remove redundant and/or incorrect swift compiler version checks

mhammond:
In https://github.com/mozilla/uniffi-rs/pull/2450#issuecomment-2675889831

> Regarding Swift 5 support, Sendable was officially added in [Swift 5.7](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0302-concurrent-value-and-concurrent-closures.md), so this shouldn’t break compatibility.

Which got me thinking about the checks in this PR - if anything, it seems like they should be `swift(>=5.7)` checks - but we're [already unconditionally using `Sendable`](https://github.com/mozilla/uniffi-rs/blob/c82945bdf04f947215f44609d70d1ec346c662b1/uniffi_bindgen/src/bindings/swift/templates/HandleMap.swift#L1)

We have.a few `swift(>=5.8)` checks - which I guess means we implicitly support a min swift version of 5.7?

I think they were added by @crazytonyli, cc @martinmose

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
@@ -22,9 +22,7 @@ public enum {{ type_name }} : {{ variant_discr_type|type_name }} {
 }
 {% endmatch %}
 
-#if compiler(>=6)
 extension {{ type_name }}: Sendable {}
-#endif
 
 #if swift(>=5.8)
 @_documentation(visibility: private)
--- a/uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/RecordTemplate.swift
@@ -15,9 +15,7 @@ public struct {{ type_name }} {
     }
 }
 
-#if compiler(>=6)
 extension {{ type_name }}: Sendable {}
-#endif
 
 {% if !contains_object_references %}
 extension {{ type_name }}: Equatable, Hashable {


mhammond:
so I guess I learned there's quite a distinction between "sendable exists" vs "sendable conformance" :)

mhammond:
@Sajjon do you have any idea how all this works?

crazytonyli:
If I remember correctly, the check was there to make CI passes.

Sajjon (approved):
@mhammond 

> but we're already unconditionally using Sendable

Yes this is a key part of this PR, since we were already unconditionally (as in no `#if Swift >`) checks, then indeed it makes no sense to have those checks elsewhere. Or rather I should say that I agree that it should be symmetrical at least. 

Sendable was introduced in and could have been used since Swift 5.5. That was 3,5 years ago. In Swift-ecosystem metrics; that's **very** long time ago. So consumers of UniFFI should most definitely have - as also it seems with the unconditional conformance for classes - a >= Swift 5.5 version code base.

However, the Swift version is only half of the piece of the puzzle in the Apple ecosystem regarding availability (and availability _checks_); the other half is the OS version at runtime. 

[`Sendable`](https://developer.apple.com/documentation/swift/sendable) is supported by iOS 8/macOS 10.10, which is also ancient, those are from 2014 (so called "wwdc2014" os versions).

As of January 2025, [68% of all iOS user were on iOS 18. 19% on iOS 17. Only 13% on pre-iOS-17](https://developer.apple.com/support/app-store/) so it feels very unlikely any user is on iOS 7.

So yes I agree with dropping availability checks for Sendable 👍🏻

mhammond:
> So yes I agree with dropping availability checks for Sendable

This PR does that, but it fails, and I'd like to understand why.

Sajjon:
> > So yes I agree with dropping availability checks for Sendable
> 
> 
> 
> This PR does that, but it fails, and I'd like to understand why.

Here is the error:
> error: stored property 'timeIntervalSecFlt' of 'Sendable' - conforming struct 'CustomTypesDemo' has non-sendable type 'TimeIntervalSecFlt' (aka 'Date')

Ok so we need to import Foundation with `@preconcurrency import Foundation` since it seems `Date` is not Sendable.

I can push a commit to this PR if you want, maybe tomorrow I will have the time

mhammond:
My question is more about why the removal of the compiler guard checks causes this.

ie:
> Ok so we need to import Foundation with @preconcurrency import Foundation since it seems Date is not Sendable.

I don't understand why that's not necessary on `main` today, but becomes necessary with this very simple PR.

Sajjon:
> Ok so we need to import Foundation with `@preconcurrency import Foundation` since it seems `Date` is not Sendable.
> 
image 

Date **is** Sendable, so I suspect this CI is running a quite old Swift..?

mhammond:
It appears to be using Swift 6.0.3 - https://github.com/mozilla/uniffi-rs/blob/f1c900cf2d28bb6773740e2effba6004fd993132/docker/Dockerfile-build#L41-L48

Sajjon:
Ahh it is an Ubuntu runner!

Ok... hmm then we need to know id that date is indeed `Foundation.Date` or if it is a type from some package, swift has "recently" made an effort to open source foundation and include support for Linux. Foundation is not supported for Linux. 

So I need to dig deeper.

Sajjon:
Any specific reason why you don't use a macOS runner for Swift - ensuring it works on Linux, is that it? 

Or maybe CircleCI does not have macOS runners?

martinmose:
> @mhammond
> 
> > but we're already unconditionally using Sendable
> 
> Yes this is a key part of this PR, since we were already unconditionally (as in no `#if Swift >`) checks, then indeed it makes no sense to have those checks elsewhere. Or rather I should say that I agree that it should be symmetrical at least.
> 
> Sendable was introduced in and could have been used since Swift 5.5. That was 3,5 years ago. In Swift-ecosystem metrics; that's **very** long time ago. So consumers of UniFFI should most definitely have - as also it seems with the unconditional conformance for classes - a >= Swift 5.5 version code base.
> 
> However, the Swift version is only half of the piece of the puzzle in the Apple ecosystem regarding availability (and availability _checks_); the other half is the OS version at runtime.
> 
> [`Sendable`](https://developer.apple.com/documentation/swift/sendable) is supported by iOS 8/macOS 10.10, which is also ancient, those are from 2014 (so called "wwdc2014" os versions).
> 
> As of January 2025, [68% of all iOS user were on iOS 18. 19% on iOS 17. Only 13% on pre-iOS-17](https://developer.apple.com/support/app-store/) so it feels very unlikely any user is on iOS 7.
> 
> So yes I agree with dropping availability checks for Sendable 👍🏻

@Sajjon 
Sorry - just to clarify, I believe Sendable was introduced in Swift 5.7, not 5.5, according to [SE-0302](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0302-concurrent-value-and-concurrent-closures.md), which explicitly states it was **implemented in Swift 5.7**. I just wanted to double-check this so we’re all on the same page moving forward? :)

martinmose:
> > So yes I agree with dropping availability checks for Sendable
> 
> This PR does that, but it fails, and I'd like to understand why.

@mhammond 

Looking at the CI output:
```
"swiftc" "-emit-module" "-module-name" "custom_types" "-o" "libtestmod_custom_types.so" "-emit-library" "-swift-version" "5" ...
```

The CI is using Swift 5, but the Sendable protocol was only introduced in Swift 5.7. By dropping the `#if compiler(>=6)` check, we're trying to use Sendable unconditionally in Swift 5, which doesn't recognize this protocol.

The error message we're seeing:
```
/home/circleci/project/target/tmp/custom_types-e851b673b3e54171/custom_types.swift:496:16: error: stored property 'timeIntervalSecFlt' of 'Sendable'-conforming struct 'CustomTypesDemo' has non-sendable type 'TimeIntervalSecFlt' (aka 'Date')
    public var timeIntervalSecFlt: TimeIntervalSecFlt
```

This appears to be a symptom of the compiler struggling with the Sendable protocol - it's trying to validate something it doesn't fully support in Swift 5. The error message itself may be misleading since Swift 5 shouldn't even recognize Sendable.

Maybe I'm off, but I think the `-swift-version 5` flag is a clue in the right direction.

martinmose:
And I only updated the Docker image to Swift `6.0.3`. And tbh, I wasn’t sure how the tests were run. Sorry!

martinmose:
Just a follow up on one of the errors:
```
/home/circleci/project/target/tmp/custom_types-e851b673b3e54171/custom_types.swift:492:16: error: stored property 'url' of 'Sendable'-conforming struct 'CustomTypesDemo' has non-sendable type 'Url' (aka 'URL')
    public var url: Url
```

Maybe it is a "real" error. I found:
https://forums.swift.org/t/swift-foundation-now-available/73530

So it could be that there are some issues with _"older"_ Swift Foundation on Linux. Like URL doesn't conform to Sendable. So it would be a good idea to ensure it always is compiled with Swift 6.

But now I'm just guessing - this is not something I know super much about...

Sajjon:
I would be able to help if it was not a Linux runner. I've never done any Swift development on Linux. So we need to know howcome we even can use Foundation typed on Linux, since AFAIK Foundation is not available Linux.

So are we using the Swift package in UniFFI?

https://github.com/swiftlang/swift-corelibs-foundation

?

mhammond:
As above, let's leave this (and I guess the fact async is broken on swift 6) to people who are actually impacted.

PR #2454, opened 2025-02-24
Add missing semicolons in UDL interface declarations

regexident:
A couple of documentation snippets are missing the trailing semicolons, one of which I ended up copy-pasting, resulting in quite a bit of confusion as to why my `.udl` file suddenly kept failing to parse.


badboy (approved):
Thank you!

Issue #2153, opened 2024-06-14
Workspace broken | UniFFI generates invalid Swift & Python bindings


Sajjon:
(@bendk [asked me to create an issue](https://github.com/mozilla/uniffi-rs/pull/2087#issuecomment-2166132975) )

Following up on [my comment in 2087](https://github.com/mozilla/uniffi-rs/pull/2087#issuecomment-2128711332), using UniFFI with a Workspace works well in the **land of Rust** - especially with the new system for remote / custom / external types laid out in #2087.  

However, when one generates Swift bindings, no cigar, the Swift code in each file tries to reference `(file)private` symbols such as `lift`/lower` on types in the other file - which are not visible. The Kotlin bindings, however, does seem to work. The Kotlin Bindgen test in my small UniFFI Workspace demo project [*Lemon Meringue Pie*](https://github.com/Sajjon/lemon-merringue-pie/blob/main/crates/lemon_meringue_pie/tests/bindings/pie_baking.kts#L11) passes. I have not tried doing an actual Kotlin bindgen generation besides the bindgen test.

The Python bindings seems broken, for [the small Python Bindgen tests in *Lemon Meringue Pie*](https://github.com/Sajjon/lemon-merringue-pie/blob/main/crates/lemon_meringue_pie/tests/bindings/pie_baking.py), I get error:
```
    from .common import Uuid
ImportError: attempted relative import with no known parent package
```

We cannot simply change the access modifier to `internal` in the generated Swift code since **each generated Swift file** contains the same types, such as `fileprivate protocol FfiConverterRustBuffer`.

In this issue I lay out a straightforward idea on how to fix the broken Swift bindings specifically. I do not know enough about Python to lay out a plan for it, but perhaps the say idea is viable also for Python?

## Idea of solution for Swift bindings
My idea is to:
1. All generated Converters will be placed in *separate files*, in lets call it `generated/` folder (as specified by `--out-dir`), and **only be generated once**, i.e. the contents of [`template/StringHelper.swift`](https://github.com/bendk/uniffi-rs/blob/custom-type-rework/uniffi_bindgen/src/bindings/swift/templates/StringHelper.swift) will be put `Generated/ScaffoldingUniFFI` folder
2. The actual type - `fileprivate struct FfiConverterString: FfiConverter` - gets a changed visibility to `internal struct` instead of `fileprivate`. 
[...]

PR #2456, opened 2025-02-26
Fix compiling issue in Swift bindings for async throw functions

crazytonyli:
`<ErrorType>.lift` Swift functions are private API and are referenced in `errorHandler: {{ e|ffi_error_converter_name }}.lift`. That would cause compiling issues in Swift bindings of a crate that uses external error types in async functions.

The first commit reproduces the issue, and the second commit fixes the problem.

### DIFF
--- a/fixtures/ext-types/lib/src/lib.rs
+++ b/fixtures/ext-types/lib/src/lib.rs
@@ -200,6 +200,11 @@ fn throw_uniffi_one_error() -> Result<(), UniffiOneError> {
     Err(UniffiOneError::Oops("oh no".to_string()))
 }
 
+#[uniffi::export]
+async fn throw_uniffi_one_error_async() -> Result<(), UniffiOneError> {
+    Err(UniffiOneError::Oops("oh no - async".to_string()))
+}
+
 // external interface errors don't quite work.
 #[uniffi::export]
 fn throw_uniffi_one_error_interface() -> Result<(), UniffiOneErrorInterface> {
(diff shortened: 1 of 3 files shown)


crazytonyli:
BTW, this PR is _potentially_ related to https://github.com/mozilla/uniffi-rs/issues/2153. I'm not entirely sure because I haven't read the entire discussion there.

bendk:
This looks good to me, thanks for the patch!

Can you merge this together into 1 commit and re-push?  I think 1 commit is better to merge, since it works better with git-bisect.  Also, I'm hoping that a re-push will unbreak the circle CI pipeline.

crazytonyli:
@bendk Done. All changes are in one commit now.

bendk:
We're really close here:
- Can you rebase one more time?  We just merged a fix for the `rustup` breakage.  Hopefully that will make CI run.
- When I run this with Swift 6, I get errors from calling an async function outside of an async context.  I updated the test to use a task here: https://github.com/mozilla/uniffi-rs/pull/2466/commits/78cdd16bb902e2797321534076008b71988f6d74 how does that look to you?  Feel free to merge it or make some other changes, I'm not tied to that code.

crazytonyli:
@bendk I have added your changes to this PR. It's probably an "issue" of Swift Linux, because I can run the test just fine without Task on my mac.

bendk (approved):
Still not sure what's going on with CI, but this change looks good to me.

PR #2463, opened 2025-03-03
Fix Circle CI rustup update

bendk:
`rustup update` will download the latest version, however we want to get the version specified in `rust-toolchain.toml`.  We're currently running into an issue with this -- my guess is that somehow the worker downloaded Rust 1.84.1 or 1.85.0 so `rustup update` isn't going to download 1.84.0.


badboy (approved):
Yeah, this broke Rust ecosystem wide, breaking change in latest rustup: https://blog.rust-lang.org/2025/03/02/Rustup-1.28.0.html#whats-new-in-rustup-1280

PR #2468, opened 2025-03-04
Update KMP bindings project link

paxbun:
The original UniFFI KMP bindings project is forked to another repo.

@benkuly


bendk (approved):
Thanks for the update!

PR #2472, opened 2025-03-06
Update default Rust version to 1.85

mhammond:


mhammond:
eek - so obviously the 2024 changes aren't going to work on our MSR - do you think it's OK to just work out how to exclude those 2 fixtures in that case, or should we do something else?

Maybe the fixture changes (ie, this PR) isn't actually worthwhile?

mhammond:
> Maybe the fixture changes (ie, this PR) isn't actually worthwhile?

Landing this after msr is 1.85 seems the simplest thing...

mhammond:
Sorry for the noise, we might as well get the 1.85 changes in and ignore 2024 for now.

badboy (approved):

PR #2474, opened 2025-03-10
Remove use of the `paste` crate

### LINKED ISSUE #2473: Dependency on unmaintained crate "paste"
This crate depends on "paste" which is no longer maintained and the repository is archived. https://rustsec.org/advisories/RUSTSEC-2024-0436

Would it be possible to remove all usages of the paste! macro and replace with some alternative?

badboy:
It was only used in a single place, and that can be handled by just passing the identifiers.

Closes #2473


mhammond (approved):
looks great, but looks like it needs a `cargo fmt`

PR #2477, opened 2025-03-11
uniffi_bindgen: Add support for specifying extra link frameworks in generated modulemap file

anayw2001:
When including another static library in the form of an xcframework (C++/C), the resulting Rust library also takes a dependency on the frameworks that are linked as part of that xcframework. This change allows specifying the frameworks so that this contract is upheld.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/gen_swift/mod.rs
+++ b/uniffi_bindgen/src/bindings/swift/gen_swift/mod.rs
@@ -172,6 +172,8 @@ pub struct Config {
     omit_localized_error_conformance: Option<bool>,
     #[serde(default)]
     custom_types: HashMap<String, CustomTypeConfig>,
+    #[serde(default)]
+    link_frameworks: Vec<String>,
 }
 
 #[derive(Debug, Default, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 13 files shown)


uniffi_bindgen/src/bindings/swift/gen_swift/mod.rs
```diff
@@ -172,6 +172,7 @@ pub struct Config {
     omit_localized_error_conformance: Option<bool>,
     #[serde(default)]
     custom_types: HashMap<String, CustomTypeConfig>,
+    link_frameworks: Vec<String>,
```
bendk, line 175: I think CI is failing because you need `#[serde(default)]` above this.

bendk (changes requested):
The code looks good, other than that config error.

I don't really understand how modulemaps work, maybe someone with more Swift knowledge can check this.  @crazytonyli @kriswuollett what do you think?

uniffi_bindgen/src/bindings/swift/gen_swift/mod.rs
anayw2001: Thanks! Done!

mhammond:
FWIW, I think we'd probably want docs [here](https://github.com/mozilla/uniffi-rs/blob/main/docs/manual/src/swift/configuration.md) and a quick note in the CHANGELOG

crazytonyli:
I don't see any issue with this change. However, I don't know what the use case is. Plus, we have the option to generate the modulemap file ourselves (`generate_module_map = false`). So, we can already add `link` declarations or whatever to modulemap files without this PR. I guess it'd be easier for users to let the library do it.

BTW, you can find modulemap documentation here: https://clang.llvm.org/docs/Modules.html#link-declaration.

bendk:
[This stackoverflow](https://stackoverflow.com/questions/58235025/are-link-framework-lines-necessary-in-a-modulemap-file) seems to say that it's a niche feature, but sometimes useful.

bendk (approved):
Thanks for adding those docs/changelog.  I think might be ready to merge, but I'll give some more time for other Swift devs to chime in.

anayw2001:
@crazytonyli We use `cargo-swift` which will by default, generate the modulemap from the crate using uniffi. I'm not sure if we can get it to use a custom modulemap file.

anayw2001:
Hi folks, could we merge this PR given it's been approved?

bendk:
> Hi folks, could we merge this PR given it's been approved?
Yup, I meant to on Monday but it slipped my mind.

PR #2485, opened 2025-03-19
Drop unused paste dependency

anayw2001:


bendk:
Good catch and thanks for the PR!

bendk (approved):

Issue #2357, opened 2024-12-17
Serialize a UniFFI object to JSON from the foreign platform code


chris13524:
Consumers of my Rust code, in Kotlin and Swift, want to serialize one of the `#[uniffi::Record]` objects to a JSON string. Is this possible currently?

I've tried creating a generic helper function to serialize any serializable object. However it looks like `#[uniffi::export]` doesn't support functions with generic arguments.

In the docs it looks like UniFFI supports [exposing functions from standard Rust crates](https://mozilla.github.io/uniffi-rs/next/types/interfaces.html#exposing-methods-from-standard-rust-traits), however `serde::Serialize` is not supported here.

mhammond:
As you noted, functions with generic args don't work. While records can't have methods, you could have a namespace function which does the serialize - eg `fn serialize_my_record(record: &MyRecord) -> String`

Re `Encodable`, #1905 sketches out a proposal which seems workable, but I don't think anyone has started working on that at this stage.

chris13524:
> you could have a namespace function which does the serialize - eg fn serialize_my_record(record: &MyRecord) -> String

Got it, so I need to make many functions, one for each type they want to serialize.

Issue #1905, opened 2023-12-19
How do I automatically implement protocols/extensions/conformance in Swift?


boehs:
I have many structs and enums that need to be `Codable`, among others. Right now, what I do is I have mirror types named `FFI*` and funcs for conversion:

```
// shared.swift

// Note that we don't yet support `indirect` for enums.
// See https://github.com/mozilla/uniffi-rs/issues/396 for further discussion.
public enum FfiReviewUpvoteStatus {
    case upvote
    case neutral
    case downvote
}

// otherCode.swift

enum ReviewUpvoteStatus: Hashable, Codable {
    case upvote, downvote, neutral
    
    init(ffi: FfiReviewUpvoteStatus) {
        self = switch ffi {
            case .upvote:
                Self.upvote
            case .downvote:
                Self.downvote
            case .neutral:
                Self.neutral
        }
    }
}
```

Implementing these manually inside an extension is forbidingly difficult, but when they are there on definition they automatically conform. One potential thing I considered is just finding and replacing the output swift but come on ;) Maybe some sort of macro  like

```rust
#[derive(Debug, uniffi::Enum)]
#[uniffi::implements("Codable")]
pub enum FFIReviewUpvoteStatus {
    Upvote,
    Neutral,
    Downvote,
}
```

akin to serde?

pronebird:
I upvote this. This should work in most cases and we don't need any validation, just a way to inject a list of conformances to append to the swift file. Seems like a low hanging fruit.

mhammond:
I don't quite understand where an implementation of `Codable` would come from - or if it was external, how it would actually work!

Regardless though, it would probably be relatively easy to allow uniffi.toml  to have a map of type_name => protocol_list and have the binding generator blindly apply that to types with matching names. I'd be happy to help someone get started on a PR here.

boehs:
> I don't quite understand where an implementation of Codable would come from - or if it was external, how it would actually work!

[`Codable`](https://developer.apple.com/documentation/swift/codable) is part of the swift standard library. It can be [automatically](https://forums.swift.org/t/how-does-automatic-codable-compliance-work-for-the-sake-of-extending-it/25933) synthesized in many circumstances, though I am not convinced that we, in rust, can know when this automatic synthesis is broken, plus it may not be desirable to have this automatically applied.

[...]

PR #2487, opened 2025-03-21
Macro support for custom autogenerated Swift protocols

Iskander508:
A draft implementation for generation of Swift declarations for protocols supporting automatic implementation like  `Codable` or `CaseIterable`

This should provide a fix for https://github.com/mozilla/uniffi-rs/issues/1905 and https://github.com/mozilla/uniffi-rs/issues/2357

I'm not sure about the naming and the whole ergonomy of the new annotation. Feel free to propose changes.

TODOs:
- [ ] decide whether we even want something like this
- [ ] decide on the annotation syntax (considering readability, ergonomy, future extensibility, ...)
- [ ] currently supported only for Enum and Record. Maybe it makes sense also for Error? For the other types I cannot think of a use-case.
- [ ] extend tests
- [ ] add docs

### DIFF
--- a/fixtures/metadata/src/tests.rs
+++ b/fixtures/metadata/src/tests.rs
@@ -205,6 +205,7 @@ mod test_metadata {
                     },
                 ],
                 docstring: None,
+                swift_protocols: vec![],
             },
         );
     }
--- a/fixtures/metadata/src/tests.rs
+++ b/fixtures/metadata/src/tests.rs
@@ -241,6 +241,7 @@ mod test_metadata {
                 ],
                 non_exhaustive: false,
                 docstring: None,
+                swift_protocols: vec![],
             },
         );
     }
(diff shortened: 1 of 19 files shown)


Iskander508:
I have created a separate MR to generate at least the `CaseIterable` protocol conformance for simple enums: https://github.com/mozilla/uniffi-rs/pull/2489

mhammond:
Thanks for the contribution! We are trying to avoid having bindings-specific attributes in the core, mainly because it doesn't scale, even for languages in this repo, but particularly for language bindings outside of it. What would be better is a mechanism to specify this in `uniffi.toml`, that way only the swift bindings would need to be touched.

Iskander508:
Thanks @mhammond After looking at the current structure and seeing how these information must be passed via metadata inside the built libraries, I agree. Please have a look at the CaseIterable PR: https://github.com/mozilla/uniffi-rs/pull/2489
That one should be OK in regards of the overall architecture. I will study a bit the `Codable` protocol. Maybe we could do something similar with that.

mhammond:
Yeah, that's great - we can control optional behaviour with the toml when necessary. I'll close this one for now, thanks!

PR #2488, opened 2025-03-22
Cleanup template 'include' mechanisms, removing all `include_once` me…

mhammond:
…chanisms.

Must closer to how JS works, which seems cleaner.


bendk (approved):
LGTM.  I added this, but I don't think it's pulling it's weight.

PR #2489, opened 2025-03-22
Generate swift CaseIterable for simple enums and errors

Iskander508:
Identifying enums that conform to the `CaseIterable` protocol is relatively simple, so we probably can autogenerate the swift protocol conformance code without any risks.

Potentially we could add a new config toggle specifically for this protocol.


mhammond:
> so we probably can autogenerate the swift protocol conformance code without any risks.

I'm interested to know why this wouldn't need a config option whereas, say, `Codable` does? This isn't to say I think it should, but rather I'm trying to work out where we would draw the line?

Iskander508:
> I'm interested to know why this wouldn't need a config option whereas, say, `Codable` does? This isn't to say I think it should, but rather I'm trying to work out where we would draw the line?

Yes, why not. I have added a config flag similar to the one about `LocalizedError`. Also I have added autogeneration for simple errors, since there it is a similar case.

uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
```diff
@@ -116,3 +116,7 @@ extension {{ type_name }}: Foundation.LocalizedError {
     }
 }
 {% endif %}
+
+{% if !config.omit_case_iterable_conformance() && !e.is_flat() && !e.contains_variant_fields() %}
+extension {{ type_name }}: CaseIterable {}
+{% endif %}
```
mhammond, line 122: nano-nit, please fix this eol

mhammond (commented):
Thanks, this is great - I like that it's enabled by default rather than opt-in as it seems useful and something we should maybe do in #2490?

I'm mildly confused by the `contains_variant_fields` rather than object_references checks though?

Iskander508:
> I'm mildly confused by the `contains_variant_fields` rather than object_references checks though?

The `CaseIterable` protocol can only be autogenerated for enums that don't contain additional fields (even primitive fields).
Few examples:
```swift
public enum ConnectionEventEnum {
    case dummy
    case connected(deviceAddress: String)
    case disconnected(deviceAddress: String)
}
extension ConnectionEventEnum: CaseIterable {} // ERROR, this will not compile, because some variants contain fields

public enum Animal {
    case dog
    case cat
}
extension Animal: CaseIterable {} // OK here, all variants are simple (no additional fields)
```

uniffi_bindgen/src/bindings/swift/templates/ErrorTemplate.swift
Iskander508: It should be OK now

mhammond (changes requested):
Given the discussion in #2490, we want to default this to off, right? cc @crazytonyli . And any chance of a changelog note for both PRs? :)

Iskander508:
I have changed the config setting to off-by-default and renamed to `generate_case_iterable_conformance` (with default `false`).
And also added points to changelog. Please adjust as needed.

mhammond (approved):
Sorry for the delay, thank you for your patience!

PR #2495, opened 2025-03-31
Remove `uniffi_bindgen::backend` module.

mhammond:
Towards splitting bindings into their own crates. If we are happy to accept binding generation panicing on invalid literals we can remove some low-value code.

A breaking change :( I'll keep it draft until something else forces one :)


bendk (commented):
I kind of liked the `to_askama_error` function.  It's not that big of a deal, but I'd rather return an error than panic if possible.  OTOH, I do think the `backend` module feels a bit weird.  What if this was moved to something like a `util` or `error_util` module?  That feels a bit more light-weight.

Stepping out -- I figured that when bindings are split out into their own crates, they would still have a dependency on `uniffi_bindgen` and `uniffi_bindgen` would still be a place for shared code between all the bindings.  Did you have something else in mind?

mhammond:
> and `uniffi_bindgen` would still be a place for shared code between all the bindings. Did you have something else in mind?

I'm really not sure TBH - I guess I don't really see much use for that - that was the intend of the backend module, and as we can see, it really wasn't useful in practice.

bendk:
> > and `uniffi_bindgen` would still be a place for shared code between all the bindings. Did you have something else in mind?
> 
> I'm really not sure TBH - I guess I don't really see much use for that - that was the intend of the backend module, and as we can see, it really wasn't useful in practice.

I'm for ditching the `backend module`, but in my mind they would still have a `uniffi_bindgen` dependency for the `ComponentInterface` or the `pipeline` module.

PR #2496, opened 2025-04-03
Add uniffi-bindgen runner executable to the crate

Le-Maz:
This pull request adds the bindgen executable to the uniffi crate. Thanks to it you would be able to install uniffi-bindgen using `cargo install uniffi --features cli` instead of adding glue code to projects that use uniffi, which is currently the recommended way.


Le-Maz:
I think that having a global executable would be a reasonable default. It is the way that other tools, like for example wasm-bindgen use. It would still be possible for a package to choose to do it how it's done now, or use another way of depending on the tool, like with docker or nix.

mhammond:
As Ben said, this is what we used to do. In practice though, it was impossible to reliably ever use the global version - all you need are 2 projects which use different versions of UniFFI and you are guaranteed it will be wrong for one of them. I'd go as far as to say that once we moved to the current model the number of version mismatches we saw fell to almost zero - ie, of all the "version mismatch" protection and possibilities, it was this which was the most effective by far.



> It would still be possible for a package to choose to do it how it's done now, or use another way of depending on the tool, like with docker or nix.

Wouldn't it also be possible for you to install your own global tool? ie, if we have a choice between "make the default a footgun, but people who shoot themselves can jump through additional hoops to avoid it" vs "make the default safe, but people who want additional convenience and accept the footgun risk can jump through additional hoops", the choice seems easy?

Le-Maz:
Okay then, maybe having a global executable is a bad idea when working on more than one project if they are on incompatible versions, however in the guide there is a suggestion that I find misleading, that with cargo nightly, the binary doesn't have to rest in the project:

> Ideally you would then run the uniffi-bindgen binary from the uniffi crate to generate your bindings, but if not on [Cargo nightly](https://doc.rust-lang.org/cargo/reference/unstable.html#artifact-dependencies), you need to create a binary in your project that does the same thing.

And even with cargo nightly, you can't add an artifact dependency if the binary is not here. Maybe you can include the binary, but keep not recommending a global installation of it.

mhammond (changes requested):
Fair enough - I don't see how this can hurt :) Can you please add a license header?

Le-Maz:
Sorry for the wait, I didn't notice there was a new reply.

uniffi/Cargo.toml
```diff
@@ -53,12 +53,15 @@ tokio = ["uniffi_core/tokio"]
 # This is needed for the gecko-js bindings.
 scaffolding-ffi-buffer-fns = ["uniffi_core/scaffolding-ffi-buffer-fns", "uniffi_macros/scaffolding-ffi-buffer-fns"]
 
-# Support for WebAssembly targets in a single-threaded environment.
```
mhammond: thanks for the update, but I think we want this comment to remain.

mhammond (changes requested):

uniffi/Cargo.toml
Le-Maz: Of course, sorry for that 😅

mhammond (approved):
Thanks!

PR #2497, opened 2025-04-04
Relax `Send` bound for objects containing `Future`s for wasm32 targets

jhugman:
This is a followup to #2417, #2418.

This PR provides a [test](https://github.com/mozilla/uniffi-rs/blob/96a3a8de28cd84f233cac5f767f0f7273d23d522/fixtures/wasm-arc-futures/src/lib.rs) which was failing to compile, and the fix.

### DIFF
--- a/docs/manual/src/wasm/configuration.md
+++ b/docs/manual/src/wasm/configuration.md
@@ -1,0 +1,60 @@
+# WASM
+
+WASM defines a portable binary format for virtual machines running untrusted code in browsers and various other scenarios. Depending on the scenario, the host language may vary.
+
+As such, `uniffi-rs` does not come with bindings generators "for WASM". This is the job for external uniffi-bindgen generators. e.g. [Typescript][ubrn], [Kotlin][gobley].
+
+However, the generated scaffolding needs to be configured for that WASM32 target.
+
+[ubrn]: https://github.com/jhugman/uniffi-bindgen-react-native
+[gobley]: https://github.com/gobley/gobley
+
+# Configuration
+
+Configuring the Rust scaffolding ready for WASM binding generators is done by opting into features.
+
+e.g.
+
+```toml
+[dependendencies]
+uniffi = { version = "0.29.2", features = ["wasm-unstable-single-threaded"]}
+```
+
+## Features
+
+### `wasm-unstable-single-threaded`
+
+At time of writing, the state of threading in WASM and Rust support is still in flux.
+
+Much of the ecosystem is built around [gloo][gloo], [`web-sys`][websys] and [`wasm-rayon`][rayon].
+
+Enabling the `wasm-unstable-single-threaded` feature opts out of the `Send` and `Sync` checks when building for `wasm32` target architectures.
+
+This feature only affects the `wasm32` architecture.
+
+Hint: running `clippy` on client code built for `wasm32` may trigger a clippy [warning when constructing `Arc`s][clippy/arc].
+
+The lint's remedy is either to:
+- make the object to `Send` and `Sync`, or
+- change from an `Arc` to an `Rc`.
+
+As we are using this feature:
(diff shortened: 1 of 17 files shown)


bendk (changes requested):
IIUC, this is just an extension of #2418 and we're not committing to anything fundamentally new.  Am I missing anything?  If not, it looks good to me, I'd just prefer it implemented without the macro.

bendk (approved):
Looks great to me then.  @mhammond did you want to review this one?

uniffi_macros/src/object.rs
```diff
@@ -121,6 +121,7 @@ fn interface_impl(object: &ObjectItem, options: &DeriveOptions) -> TokenStream {
         // if they are not, but unfortunately it fails with an unactionably obscure error message.
         // By asserting the requirement explicitly, we help Rust produce a more scrutable error message
         // and thus help the user debug why the requirement isn't being met.
+        #[cfg(not(target_arch = "wasm32"))]
```
mhammond, line 124: Removing this check seems to be removing it for far more things that just objects holding `Future`s?

mhammond (commented):
It seems like this might need docs, both for future users, but also just for me to better understand exactly what promises we are no longer making in that world and to better understand the implications. What happens when wasm supports threads, which IIUC, is progressing?

jhugman:
As discussed with @mhammond and @bendk: this introduces a new feature `wasm-unstable-single-threaded`, with documentation.

mhammond (approved):
Thank you!

docs/manual/src/wasm/configuration.md
```diff
@@ -0,0 +1,58 @@
...
+
+### `wasm-unstable-single-threaded`
+
+At time of writing, the state of threading in WASM and Rust support is still in flux.
+
+Much of the ecosystem is built around [gloo][gloo], [`web-sys`][websys] and [`wasm-rayon`][rayon].
+
+Enabling the `wasm-unstable-single-threaded` feature opts out of the `Send` and `Sync` checks when building for `wasm32` target architectures.
+
+It may be safely enabled when building for other architectures.
```
bendk, line 31: Nit: Can you make "safely" a bit more precise?  How about something like "This feature only affects the `wasm32` architecture".

fixtures/wasm-arc-futures/README.md
```diff
@@ -0,0 +1,39 @@
...
+$ cargo test
+```
+
+It is possible to filter by test names, like `cargo test -- swift` to only run
+Swift's tests.
+
+## Run the examples
+
+At the time of writing, each `examples/*` directory has a `Makefile`. They are
+mostly designed for Unix-ish systems, sorry for that.
```
bendk, line 19: I'm confused by this one, since I don't see a `Makefile` being checked in.

What do you think about keeping these tests in your repo?  I don't see other bindings generators getting any use out of it and I'd rather not add it to the list of crates we need to publish (in fact, I'm hoping to shrink that list very soon).

bendk (changes requested):
Looks great, I love the docs.  My only concern is the new test fixture crate.  I'm currently thinking it's better for it to live in your repo, but I can be convinced otherwise pretty easily.

fixtures/wasm-arc-futures/README.md
jhugman: Oh ha, that was some copy pasta. I can definitely review and re-word this, if you still want the test.

jhugman:
> My only concern is the new test fixture crate. I'm currently thinking it's better for it to live in your repo, but I can be convinced otherwise pretty easily.

I have a test ready to land at https://github.com/jhugman/uniffi-bindgen-react-native/pull/241. 

Relying on that test only would put the feature in danger of accidental breakage, that wouldn't get detected until downstream WASM bindgen crates tried to upgrade uniffi.

PR #2500, opened 2025-04-13
Support extracting metadata from COFF files

paxbun:
COFF is the object file format used by Visual C++ and MinGW. Though these two use different archive file formats (`.lib` and `.a`), both of them use COFF as entries in archives. This PR adds support for extracting metadata from COFF files.

### DIFF
--- a/uniffi_bindgen/src/macro_metadata/extract.rs
+++ b/uniffi_bindgen/src/macro_metadata/extract.rs
@@ -9,7 +9,7 @@ use goblin::{
     archive::Archive,
     elf::Elf,
     mach::{segment::Section, symbols, Mach, MachO, SingleArch},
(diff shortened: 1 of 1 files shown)


paxbun:
Additional context:

1. We are making KMP bindings for UniFFI, and we have a user who wants to use UniFFI with MinGW Kotlin Native on Windows, which needs a static library (https://github.com/gobley/gobley/discussions/105).

2. The changes are tested with our bindgen, but it seems the fixtures use shared libraries; what would be a better way to add unit tests for this?

image

bendk (approved):

I don't feel like we need to go out of our way to unit test these.  The code is fairly straightforward and if it works for you that seems good enough to me.

PR #2504, opened 2025-04-17
Async naming updates and doc fixes

bendk:
After implementing foreign futures these in JS, I wanted to make some updates.  The FFI is staying the same, but some names have been changed.

`UniffiForeignFutureFree` is now named `UniffiForeignFutureDroppedCallback` and `UniffiForeignFuture` is now `UniffiForeignFutureDroppedCallbackStruct`. This reflects the fact that we don't really need to use these to manage the future, the real reason they're here nowadays is to support cancellation. For languages like JS, where we can't implement cancellation, we can just ignore the param.

Updated some of the code that deals with this to treat it like the out parameter it is, rather than a return value.  I think that was leftover from previous implementations.

There's also one change for RustFuture: `RustFuturePoll::MaybeReady` is now `RustFuturePoll::Wake`.  "MaybeReady" was always a confusing/ambiguous name.  "Wake" seems reasonable, since it's what the foreign bindings should do and reflects how the Rust futures code works.

Updated some outdated docstrings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,15 @@
 
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.29.1...HEAD).
 
+### ⚠️ Breaking Changes for external bindings authors ⚠️
+
+* Some async-related names have changed.  Bindings authors may need to update their code to reflect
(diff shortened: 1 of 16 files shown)


mhammond (approved):
technically a breaking change, so we'll land after decide to force one.

bendk:
`0.29.3` is out so it's a good time to merge some breaking changes.

PR #2505, opened 2025-04-17
Adding detailed documentation for foreign futures

bendk:
I'm thinking of this as a companion piece to the existing docs.  That one describes the flow at a high level, this one describes the low-level FFI details.

I also started a glossary section.  I find I'm using words in these docs like "lift" and "lower" that are useful, but I don't want to fully explain.  I'm thinking that we can link to the glossary instead.


docs/manual/src/glossary.md
```diff
@@ -0,0 +1,26 @@
...
+This is an `extern "C"` API that helps the foreign bindings make Rust calls.
+
+## Bindings
+
+UniFFI-generated code in a foreign language.
+This implements the exported Rust API by calling in to the Rust scaffolding.
+
+## Lower
+
+The process of converting a high-level Rust/foreign language type into a lower-level FFI type as
```
mhammond, line 17: At first read I thought this was just from the scaffolding perspective. Maybe: "The process of converting a high-level Rust type (when used in the context of the Scaffolding) or foreign language type (when used in the context of the Bindings) ..."? Ditto for lower

(but if you think that's worse, I'm fine with your version - we can improve it later if needed)

mkdocs.yml
```diff
@@ -98,7 +98,8 @@ nav:
   - ./internals/crates.md
   - ./internals/lifting_and_lowering.md
   - ./internals/object_references.md
-  - ./internals/async.md
```
mhammond, line 101: huh, so I guess I'm a little confused - you want to delete the existing async.md? It's not removed in this patch though, and that seems to cover things (including a cute mermaid diagram) that's not covered in the new docs.

mhammond (approved):

mkdocs.yml
bendk: I'm not deleting, just renaming it to async overview to differentiate it from async FFI details.

mhammond: oops, sorry, I worked that out but must have left that stray comment!

bendk:
Those suggestions were great.  I updated the docs based on that.  Merging now, we can continue to improve these as we go.

PR #2506, opened 2025-04-18
Add builtin headers to modulemap

anayw2001:
The generated Swift header file uses stdbool, stddef and stdint. In certain build environments, the Swift module is unable to be generated due to these headers being missing from the modulemap dependencies.


anayw2001:
Hi folks, could I get a review here?

mhammond (approved):

PR #2510, opened 2025-04-23
Add search feature on documentation website

### LINKED ISSUE #2509: Enable search bar/feature on documentation website
Material for MkDocs has a [built-in plugin for search](https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-search/?h=search#built-in-search-plugin) which works great. You can enable it without any plugin import requirements.

```yaml
plugins:
  - search
```

This would be useful because the docs are fairly extensive and finding what you're looking for is not always easy!

thunderbiscuit:
HI team! This PR fixes #2509.

Here is what it looks like when enabled:

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -32,6 +32,7 @@ extra:
 plugins:
   # otherwise a no-op docs build ends with a diff to sitemap.xml.gz
   - no-sitemap
+  - search
 
 # This lists all the files that become part of the documentation
 nav:


mhammond (approved):
Thanks!

mhammond:
it looks like that commit was deployed by CI, but I'm not seeing the search? I guess maybe we need a manual deploy, but I'm not sure why that would be

thunderbiscuit:
Weird. This is the workflow that deploys right?
```yaml
  Deploy website:
    docker:
      - image: cimg/rust:1.77.1
    resource_class: small
    steps:
      - checkout
      - install-mkdocs
      # It should be possible to do something like https://github.com/jimporter/mike/issues/49
      # So we can automagically deploy on release too!
      - run:
          name: Deploy docs
          # can't use `gh-pages/deploy`, it doesn't handle the versioning.
          command: |
            # circle ci docs and jj both do this git config.
            git config user.name 'uniffi-docs[bot]'
            git config user.email 'uniffi-docs[bot]@users.noreply.github.io'
            mike deploy next --push
```

I've never used the mike library but it looks like it could be the issue here.

Looking at your `tools/requirements_docs.txt` it looks like your mkdocs-material and mike libraries are not super old but could be updated?

Locally I do `mkdocs serve` and the search gets built. It's a div with the `md-search` class, which definitely doesn't exist on the current live website. I was unable to test the `mike serve` command locally, it just tells me I need to set a default but the list of available defaults it empty. Anyway I'd start with updating your mkdocs-material and mike to the latest versions (this should also fix your padding issue between the dropdown group items in the sidebar which was happening on older versions of Material for mkdocs.

Would you like me to open a PR for that (just the bump of the 2 libs)?

mhammond:
> Would you like me to open a PR for that (just the bump of the 2 libs)?

Yeah, sure, thanks!

mhammond:
(I sure didn't remove search when moving to mike, so the fact stuff needs updating to support it isn't that surprising)

Issue #2508, opened 2025-04-22
Python async trait impl generates a missing `XRustBuffer.default()` call in handle_error


skeet70:
In [ironcore-alloy](https://github.com/IronCoreLabs/ironcore-alloy/tree/dependabot/cargo/uniffi-0.29.1) there's a [test](https://github.com/IronCoreLabs/ironcore-alloy/blob/dependabot/cargo/uniffi-0.29.1/python/ironcore-alloy/tests/test_ironcore_alloy.py#L643) that implements a foreign trait in Python. The test fails because when an error is raised in the foreign implementation, the `handle_error` has a `_UniffiRustBufferAlloyHttpClientResponse.default()` call which doesn't exist.

bendk:
Yup, that's a UniFFI bug.  I just put up a PR to fix it.

skeet70:
This was fixed by https://github.com/mozilla/uniffi-rs/commit/9c8471305d62af1746d71a363b4a297b6d0e9d0a

PR #2517, opened 2025-04-25
Python: fix `ffi_default_value` (#2512) backport for release

skeet70:
The change to `ffi_type_label` from
`https://github.com/mozilla/uniffi-rs/pull/2385` also needs to be applied here.

See https://github.com/mozilla/uniffi-rs/issues/2508 for details.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/gen_python/mod.rs
+++ b/uniffi_bindgen/src/bindings/python/gen_python/mod.rs
@@ -424,7 +424,7 @@ impl PythonCodeOracle {
     /// Default values for FFI types
     ///
     /// Used to set a default return value when returning an error
-    fn ffi_default_value(&self, return_type: Option<&FfiType>) -> String {
+    fn ffi_default_value(&self, return_type: Option<&FfiType>, ci: &ComponentInterface) -> String {
         match return_type {
             Some(t) => match t {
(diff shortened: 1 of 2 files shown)


bendk:
Do we need to backport issues individually or can we just take everything in `main`?  I did a quick look through the tickets and these are the only ones I could see that might have issues:
- [Allow uniffi_reexport_scaffolding! macro to work in Rust 2024](https://github.com/mozilla/uniffi-rs/commit/0775272488c4119c5070dc941a097644589a90a6) -- This will bump the MSRV to `1.82.0`, which seems okay to me.  moz-central is already on that one.
- [Migrate to askama 0.13.0](https://github.com/mozilla/uniffi-rs/commit/1564451ee4f3853962c21a4f26de80270ddfd659) -- technically brings in a new crate, but it's really just a version upgrade for rinja.

mhammond:
Yeah, I think we should just drop these backports, wait for James's PR, and just to 0.29.2 from main. wdyt?

skeet70:
Fine with me, I made this one because I didn't want to run through all the changes to see if it was safe to release it all, not knowing the codebase well enough. I only know of this bug as one I explicitly need fixed to release `ironcore-alloy` w/ updated uniffi.

PR #2519, opened 2025-04-25
Upgrade cargo_metadata to 0.19

anayw2001:


mhammond:
We are going to need to be careful going past 0.15 - https://searchfox.org/mozilla-central/rev/d65eb426c0e9f5cade6dfbf5e3f4e1b1903b6d04/Cargo.lock#681

anayw2001:
@mhammond: It looks like the only project using cargo_metadata in that Cargo.lock file is uniffi, minus uniffi-bindgen-gecko-js, which looks to be an internal Mozilla crate as it's not on crates.io, so I think this is fairly safe to do.

bendk:
Yes that seems right to me.  :rocket:

mhammond (approved):

PR #2526, opened 2025-05-07
Build docs.rs documentation with rustdoc "jump to definition" feature enabled

GuillaumeGomez:
I was reading source code on docs.rs and realized that the feature was not enabled, so here it is. :)

Enabling this feature makes your source code pages generate links on items so you can jump to where they're defined.

You can this feature in action in [`syn`](https://docs.rs/syn/latest/src/syn/lib.rs.html#986-1010) for example.

### DIFF
--- a/uniffi/Cargo.toml
+++ b/uniffi/Cargo.toml
@@ -59,3 +59,6 @@ wasm-unstable-single-threaded = [
     "uniffi_core/wasm-unstable-single-threaded",
     "uniffi_macros/wasm-unstable-single-threaded",
 ]
+
+[package.metadata.docs.rs]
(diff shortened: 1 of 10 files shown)


bendk (approved):
Seems good to me.  @badboy @mhammond any reason not to accept this?

badboy (approved):
sounds good to me!

mhammond:
That's cute! Looks like a no-brainer to me (although I do wonder why it's not the default!)

GuillaumeGomez:
Nightly-only feature (for now). Hopefully in a few years it should become stable and enabled by default.

PR #2527, opened 2025-05-07
wasm-unstable-single-threaded: Relax `Send` for `async_trait` for `ForeignFuture`.

### LINKED ISSUE #2511: `wasm-unstable-single-threaded`: Relax `Send` and `Sync` for callback interfaces and traits
Followup to #2417 and #2497.

#2497 introduces a `wasm-unstable-single-threaded` to relax `Send` and `Sync` bounds for Objects passed into and returned from Rust from a single-threaded binding.

This issue covers the case where callback interfaces and traits.

jhugman:
Fixes #2511 

This is a followup for #2418 and #2497.

It does not add anything fundamentally new, just doing the same as the other PRs but for `ForeignFuture`s.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1992,6 +1992,7 @@ dependencies = [
 name = "uniffi-fixtures-wasm-unstable-single-threaded"
 version = "0.29.0"
 dependencies = [
+ "async-trait",
  "uniffi",
 ]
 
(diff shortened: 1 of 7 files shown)


bendk (approved):
Changes look correct to me, let's merge this one.

PR #2534, opened 2025-05-10
Clarify semantics around uniffi_reexport_scaffolding! macro

### LINKED ISSUE #2475: Clarify semantics around `uniffi_reexport_scaffolding!` macro
[Our docs talk about `uniffi_reexport_scaffolding!`](
https://mozilla.github.io/uniffi-rs/latest/tutorial/Rust_scaffolding.html#libraries-that-depend-on-uniffi-components), which links to [rust-lang#50007](https://github.com/rust-lang/rust/issues/50007), which is marked as "fixed". According to #1242, that was fixed in Rust 1.62, and we are way past that.

However, if I patch m-c to remove `uniffi_reexport_scaffolding!` I see undefined symbol errors for all our scaffolding functions. So it appears something remains necessary for symbols in other crates to be exported in the final library.

I bring this up because according to #2459, the code generated by this macro fails to build with Rust 2024 - instead of "fixing" that I was hoping we could just remove the capability entirely, but apparently we can not. Assume that's true, we should try and better document this work-around so we aren't blaming an issue that no longer exists.

mhammond:
Fixes #2475.

### DIFF
--- a/docs/manual/src/tutorial/Rust_scaffolding.md
+++ b/docs/manual/src/tutorial/Rust_scaffolding.md
@@ -39,16 +39,14 @@ If you don't specify a [`namespace`](../types/namespace.md) the crate name is us
 
 Suppose you want to create a shared library that includes one or more
 components using UniFFI. The typical way to achieve this is to create a new
-crate that depends on the component crates.  However, this can run into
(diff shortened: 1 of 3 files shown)


bendk (approved):
The new working looks good to me, thanks for giving this some love.

PR #2536, opened 2025-05-13
General pipeline updates

bendk:
This were the result of me implementing the Python bindings using the new pipeline system.  There were a handful of things I realized I needed/wanted there that didn't come up when I was implementing the JS bindings.

Add `FfiTypeNode`. I didn't originally have this as part of the general pipeline, since it wasn't needed there.  However, virtually all language pipelines are going to want to add it so we may as well do it in shared code.

Determine enum discriminants for the bindings. This uses basically the same algorithm as from `ComponentInterface`. The one difference is that for enums without a specified `repr`, this calculates the smallest integer width needed to store the variants.

I used a new pipeline technique for the enum discriminents: the old/optional `discr` field gets renamed to `meta_discr` so that we can create a new/non-optional `discr` field.

Adding the `Node::has_descendant` method. I've noticed I'm implementing this code again and again, so it feels a good time to generalize.

Several other smaller changes/fixes.

### DIFF
--- a/docs/manual/src/internals/bindings_ir_pipeline.md
+++ b/docs/manual/src/internals/bindings_ir_pipeline.md
@@ -65,6 +65,33 @@ pub struct Node {
 }
 ```
 
+### Field Conversion order
+
+Fields are converted in declaration order.
+This matters when a field is listed twice with one of them using `#[node(from(<name_of_the_other_field>)]`.
+In this case the first field declared will be initialized with the data from the previous pass and the second field will be initialized to the default value.
+
+You can take advantage of this to convert an optional value into a non-optional value.
+For example:
+
+```rust
+#[derive(Node)]
+pub struct SourceNode {
+    name: Option<String>
+}
+
+#[derive(Node)]
+pub struct DestNode {
+    /// This field will be set to `SourceNode::name`
+    #[node(from(name))]
+    name_from_previous_pass: Option<String>
+    /// This field will be a non-optional version of `SourceNode::name`.
+    /// It will be initialized to an empty string, then a pipeline pass will populate it using
+    /// `name_from_previous_pass` combined with logic to handle the `None` case.
+    name: String,
+}
+```
+
 ## Module structure
 
 Each IRs will typically have a module dedicated to them with the following structure:
@@ -89,7 +116,6 @@ pub fn pipeline(): Pipeline<initial::Root, Root> {
         // This will use the logic from the Node conversions section above.
         .convert_ir_pass::<Root>()
         // Add passes to populate the new fields, mutate existing fields, etc.
-        // Each pass
         .pass(foo::pass)
         .pass(bar::pass)
         .pass(baz::pass)
--- a/uniffi_bindgen/src/pipeline/general/ffi_types.rs
+++ b/uniffi_bindgen/src/pipeline/general/ffi_types.rs
@@ -9,7 +9,7 @@ use super::*;
 pub fn pass(module: &mut Module) -> Result<()> {
     let module_name = module.name.clone();
     module.visit_mut(|node: &mut TypeNode| {
-        node.ffi_type = generate_ffi_type(&node.ty, &module_name);
+        node.ffi_type = FfiTypeNode::from(generate_ffi_type(&node.ty, &module_name));
     });
     Ok(())
 }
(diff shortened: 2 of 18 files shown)


uniffi_bindgen/src/pipeline/general/ffi_types.rs
```diff
@@ -58,6 +58,21 @@ fn generate_ffi_type(ty: &Type, current_module_name: &str) -> FfiType {
...
         | Type::Timestamp
         | Type::Duration => FfiType::RustBuffer(None),
-        Type::Custom { builtin, .. } => generate_ffi_type(builtin, current_module_name),
+        Type::Custom {
+            module_name,
+            builtin,
+            ..
+        } => {
+            match generate_ffi_type(builtin, current_module_name) {
+                // Fixup `module_name` for primitive types that lower to `RustBuffer`.
```
mhammond, line 67: this and the module/crate name confusion from the CI makes me sad. Isn't the module name the crate name transformed by the bindings config?

Can the bindings somehow fix this in a later stage? I don't quite get the new module node in the same way.

bendk: It's definitely sad.  The corner case here is something like:

- Crate A defines a custom URL type that lowers to a string
- Crate B uses URL as an external type
- When determining the FFI type, we start by getting the FFI type for the builtin.  In this case the builtin type is string, so the FfiType is `FfiType::RustBuffer(None)`.
- However, since it's an external type used in crate B, we need to change that to `FfiType::RustBuffer(Some("A"))`

The core of this issue is the fact that we're creating multiple RustBuffer types, which means the custom URL type has 2 different ffi types, depending on the module `FfiType::RustBuffer(None)` vs `FfiType::RustBuffer(Some("A"))`.  Part of this is that `FfiTypeRustBuffer(None)` is special cased to mean "the RustBuffer from current module", which doesn't work well with external types.

Maybe it would be slightly cleaner to make the module name non-optional for RustBuffer.  However, that's still pretty confusing.  It means that when we're determining FFI types, we always need to know the "current module" and that module will change once you recurse inside an external type.

I think the real solution to this is having a single RustBuffer type that all modules share.  At this point I'm very confident we can do that with this pipeline code.

docs/manual/src/internals/bindings_ir_pipeline.md
```diff
@@ -65,6 +65,28 @@ pub struct Node {
 }
 ```
 
+### Field Conversion order
+
+Fields are converted in declaration order.
+If a node contains 2 fields that can come from a single source field, the first one declared will "win".
```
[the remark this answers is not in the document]
bendk: Yeah "winning" wasn't the best word choice.  I think it's more clear now, tell me what you think.

docs/manual/src/internals/bindings_ir_pipeline.md
```diff
@@ -65,6 +65,33 @@ pub struct Node {
 }
 ```
 
+### Field Conversion order
+
+Fields are converted in declaration order.
+This matters when a field one uses `#[node(from(<name_of_the_another_field>)]`.
```
mhammond, line 71: This is much clearer, thanks!

Maybe "This matters when a field is listed twice with one of them using `#[node(from(<name_of_the_another_field>)]`"?

(I'm still not quite sure what the semantics would be of the `#[node(from(<name_of_the_another_field>)]` field was listed second and the non-option item was listed first - how does the option get converted to the non-option there?

bendk: That wording sounds good, thanks.

> (I'm still not quite sure what the semantics would be of the #[node(from(<name_of_the_another_field>)] field was listed second and the non-option item was listed first - how does the option get converted to the non-option there?

That would be the same as the general case of converting an optional value into a non-option.  It will always fail -- even if there were no None values (I'm not 100% sure on that, but I'm pretty sure that's how it should work).

docs/manual/src/internals/bindings_ir_pipeline.md
```diff
@@ -65,6 +65,33 @@ pub struct Node {
 }
 ```
 
+### Field Conversion order
+
+Fields are converted in declaration order.
+This matters when a field is listed twice with one of them using `#[node(from(<name_of_the_another_field>)]`.
```
badboy, line 71:
```suggestion
This matters when a field is listed twice with one of them using `#[node(from(<name_of_the_other_field>)]`.
```

docs/manual/src/internals/bindings_ir_pipeline.md
```diff
@@ -65,6 +65,33 @@ pub struct Node {
...
+}
+
+#[derive(Node)]
+pub struct DestNode {
+    /// This field will be set to `SourceNode::foo`
+    #[node(from(foo))]
+    maybe_foo: Option<String>
+    /// This field will be a non-optional version of `SourceNode::foo`.
+    /// It will be initialized to an empty string, then a pipeline pass will populate it using
+    // `maybe_foo` combined with logic to handle the `None` case.
```
badboy, line 90:
```suggestion
    /// `maybe_foo` combined with logic to handle the `None` case.
```

mhammond (approved):

PR #2537, opened 2025-05-13
Reimplement Python bindings using the pipeline code

bendk:
This shows how the new pipeline code affects an existing bindings implementation.  Before I was linking to the Gecko-JS bindings, but these have a bunch of extra complexity.

This is built on top of https://github.com/mozilla/uniffi-rs/pull/2536, which contains the changes to the general pipeline that I realized needed to be made when implementing Python.

I split this up so the diffs are easier to read and maybe the commits can serve as breadcrumbs for implementing other bindings using the pipeline code.  I recommend reading this PR one commit at a time.


uniffi_bindgen/src/bindings/python/pipeline/nodes.rs
```diff
@@ -0,0 +1,659 @@
...
+    pub throws: Option<Type>,
+    pub checksum: Option<u16>,
+    pub docstring: Option<String>,
+}
+
+/// Common data from Function/Method/Constructor
+#[derive(Debug, Clone, Node)]
+pub struct Callable {
+    pub name: String,
+    pub is_async: bool,
```
mhammond, line 123: why do all the types above have a `Callable` but also their own copies of the fields?

bendk: I've gone back and forth one this a lot, I'm not sure what's best.

The `initial -> general` pass is where `Callable` is added.  We need the extra copies to generate the `Callable` data.
In a previous version of the pipeline code, I had another general pass which dropped the unneeded fields after we had created `Callable` from them.  However, it felt to me like the extra pass was adding more complexity than it was worth and it was better to have a policy that passes added fields, but didn't remove them.

For the `general -> python` pass, we could easily drop the fields but I didn't because it felt consistent with the "policy".  However, I'm not loving this policy anymore, it seems kind of silly to keep these fields around in the python pass and also introduces the possibility of errors if a pass changes one of the fields but not the other.

bendk: That makes sense, I removed those fields.  This seems like the best balance.

Note: there's still some duplication here, for fields added in the Python passes.  For example, `Interface.methods` and `Interface.protocol.methods` are the same list.  It's not great, but I can't see a simple way to avoid this.  I think the best way would be adding module like `nodes2.rs`, which doesn't seem great.  I'm thinking the general rule is that it's okay to have duplicates when your pass adds new fields, but the pass after that should drop the duplicates.  In general, I guess we should try to be practical and not try for a one-size-fits-all solution.

mhammond: I don't really fully understand the implications of all this yet, but I agree entirely with your "rule" above and the need to be flexible/pragmatic - thanks!

uniffi_bindgen/src/bindings/python/templates/DurationHelper.py
```diff
@@ -4,7 +4,7 @@
 # There is a loss of precision when converting from Rust durations,
 # which are accurate to the nanosecond,
 # to Python durations, which are only accurate to the microsecond.
-class _UniffiConverterDuration(_UniffiConverterRustBuffer):
+class {{ type_node.ffi_converter_name }}(_UniffiConverterRustBuffer):
```
mhammond, line 7: still struggling to get my head around `ffi_converter_name` needing an asref so needing a filter, but here we can just reference it.

and eg, below with `string_type_node|lower_fn` - why isn't that `string_type_node.lower_fn` if we can use `type_node.ffi_converter_name` here?

bendk: I think you're right about this and we should be able to simplify things.  If possible, maybe we can remove the `AsRef`-related macros and have one less concept that engineers need to understand.

bendk: Just pushed the [change](https://github.com/mozilla/uniffi-rs/compare/4c0312cade5048dc033fcce9a316f508216856fb..3dc2f92f4f56dc5cb245380766e0f78f504f3262) and it seems great.   The code is slightly more verbose, but devs don't need to think about `AsRef` anymore which is seems great.

mhammond: thanks, that's much better IMO - the code is only marginally more verbose but quite a bit clearer - thanks!

PR #2539, opened 2025-05-18
Support default values for HashMaps

### LINKED ISSUE #2538: HashMap fields can't have defaults
I have the following object definition:

```rust
pub struct RequestParams {
    pub message: String,
    pub context: HashMap<String, String>,
}
```

I'd like to make it so that in the generated bindings look something like:

```swift
struct RequestParams {
  let message: String
  let context: [String: String]

  init(message: String, context: [String: String] = [:]) {
    // ...
  }
}

```

I tried doing:

```rust
pub struct RequestParams {
    pub message: String,
    #[uniffi(default = HashMap::new())]
    pub context: HashMap<String, String>
}
```

and a bunch of other stuff, but it seems like it's just not supported (as far as I can tell).

I _could_ probably represent this using:

```rust
pub struct RequestParams {
    pub message: String,
    #[uniffi(default = None)]
    pub context: Option<HashMap<String, String>>
}
```

but that's slightly less correct for our needs.

I'm wondering if this is something that should be supported (similar to a default empty `vec`)?

Thanks!

mhammond:
Like already supported for lists, we should allow default values for maps. Most of the support for this already exists. Supporting it for procmacros is also fairly easy, using `{}` in the macros.

Fixes #2538


fixtures/proc-macro/src/lib.rs
```diff
@@ -378,6 +378,8 @@ pub struct RecordWithDefaults {
     float_var: f64,
     #[uniffi(default=[])]
     vec: Vec<bool>,
+    #[uniffi(default={})]
+    map: HashMap<String, String>,
```
jplatte, line 382: Really not a fan of this because `{}`, while being a valid Rust expression, is never of type `HashMap`. That's why I proposed just `#[uniffi(default)]` as a new similar feature in #2538. Same thing applies to the `Vec` thing above, which I didn't know was a thing.

jplatte: It may be that there's a small misunderstanding here. I was proposing exactly the semantics we already have for `= None` and `= []` and I assume that's the same thing you implemented here for `HashMap`. The function call thing was a side note I had about supporting arbitrary Rust expressions, but I did not propose `#[uniffi(default)]` to invoke `Default::default`.

I guess one downside to this would be that if it's also implemented for `Option` and `Vec`, there's going to be two ways of defaulting them to `None` and the empty vec, respectively. But doesn't seem too bad to me.

bendk: It seems like the only real question is how to spell this out.  I agree that none of the option seem great:
- `#[uniffi(default={})]` -- Looks like an empty struct literal.  Could conflict with future syntax for writing struct literals (i.e. we might want to allow `#[uniffi(default={foo: 100, bar: ""})]`)
- `#[uniffi(default)]`  -- To me this looks the best, but we might want to reserve it for calling `Default::default()`.  Also, requires that we know the type to generate the default.
- `#[uniffi(default=HashMap::new())]`  -- Looks like a function call and quite different than the other literals.  Users could get confused and write something `HashMap::default()` by mistake.
- `#[uniffi(default=empty_map]`  -- I'll throw this one out here too.  This one looks weird compared but the fact that no one could confuse it with Rust syntax might be a plus.

I don't think any option is clearly best.  If we could make `#[uniffi(default)]` work without having to change too bunch of the bindings code, that would be my preference.  We could reserve `#[uniffi(default=default())` for calling `Default::default()`.  If we can't do that, then maybe I like defining a new keyword like `empty_map` for this, to try to avoid confusing users too much.

bendk: I haven't thought this through much, but I could could see us using one form to mean that the foreign bindings generate a default value directly and a second form to mean that the foreign bindings make a call into Rust to get the result of `Default::default()`.  The first version could be used for primitives, empty lists/maps, and structs that contained only those types.  The second version could be for interfaces, structs that contained interfaces, etc..

I'm not really sure about it all, but I'm pretty sure that the first version is more useful -- and I think that's what you're getting at too.  So, for this issue I think we could decide that `#[uniffi(default)]` should generate an empty map, empty list, empty string, false, etc.  If we need the `Default::default()` version, then it can get a more wordy syntax (maybe `#[uniffi(default=default())]`).

mhammond:
OK, this is just UDL, which weedle already supports and probably isn't going to get more generic "default" support.

jplatte (approved):

bendk (approved):
UDL-only makes sense as a first step.  This looks good to me.

PR #2541, opened 2025-05-22
docs: Fixing typos and list indentation

badboy:
Lists need to be separated by a blank line from other paragraphs to be rendered correctly.


mhammond (approved):
> Lists need to be separated by a blank line from other paragraphs to be rendered correctly.

Annoyingly, vscode's "preview" and (IIRC) github itself *does not* have this requirement, so often it's only noticed once rendered by mkdocs etc :(

badboy:
> > Lists need to be separated by a blank line from other paragraphs to be rendered correctly.
> 
> Annoyingly, vscode's "preview" and (IIRC) github itself _does not_ have this requirement, so often it's only noticed once rendered by mkdocs etc :(

Yup. That's how I noticed, I was finally re-reading the pipeline docs on our doc page! :D

PR #2543, opened 2025-05-26
Proc-macros allow default values without needing to specify a literal.

### LINKED ISSUE #2538: HashMap fields can't have defaults

mhammond:
@jplatte @bendk like we spoke about, this seems quite nice to me. It's a breaking change (and I think the changelog is messed up, but we can sort that out)

This makes it possible to say you want the default value for a type rather than specifying a literal - eg, `#[uniffi(default)]` and similarly for defaults args for functions.

Supports named types but it's up to the consumer to ensure they can be created without args.

via and fixes #2538


jplatte (approved):
Nice!

CHANGELOG.md
```diff
@@ -20,6 +20,15 @@
 
 - HashMaps in UDL now support a default value with an empty map ([#2539](https://github.com/mozilla/uniffi-rs/pull/2539)).
 
+XXX - thiss change log is missing 29.2??
```
bendk, line 23: I think I forgot to open a PR after releasing 0.29.2.  Here it is now though.: https://github.com/mozilla/uniffi-rs/pull/2546/

docs/manual/src/proc_macro/functions.md
```diff
@@ -59,11 +63,18 @@ impl TextSplitter {
...
+Supported defaults without a literal:
 
-  - String, integer, float, and boolean literals
-  - `[]` for empty Vecs
-  - `Option<T>` allows either `None` or `Some(T)`
+- All `Option` types default to `None`, Vecs and Maps default to empty items
+- All builtin types have default values (zero, false, empty)
+- Named types (ie, Records, Objects, etc) are supported, but you must ensure the
+  item is able to be created without arguments (eg, has a default constructor
+  with no args, is a record with all fields having defaults, etc)
```
bendk, line 72: Looks right to me, but I think describing 1 type per row would be clearer.  It's less compact, but I think it's worth the space.  Maybe a table would be better:

| Type | Default value | 
|---------|----------------------|
| `Option` | `None` |
| String | Empty string |
| `Vec<T>` | Empty list |
| `HashMap<T>` | Empty map |
| Builtin numeric types | 0 |
| `bool` | false |
| Records | Record with all fields set to their default value (only valid if they all have defaults) |
| Objects | Primary constructor called with 0 arguments |

docs/manual/src/proc_macro/functions.md
```diff
@@ -26,14 +26,15 @@ To export methods of an interface you can use the [`#[uniffi::export]` attribute
 ## Default values
 
 Exported functions/methods can have default values using the `default` argument of the attribute macro that wraps them.
-`default` inputs a comma-separated list of `[name]=[value]` items.
+`default` inputs a comma-separated list of either `[name]` items (where the type's default value should be used)
+or `[name]=[value]` items (where the default value is specified as a literal).
```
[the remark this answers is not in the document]
mhammond: I changed this to "where the type's *foreign* default value should be used" to try and help reinforce this is all on the foreign side.

docs/manual/src/proc_macro/functions.md
mhammond: thanks - that just makes the section below this look even worse, but I just opened #2551 instead of getting distracted by that.

mhammond:
Hey @bendk, I think I'm heading in the right direction here but missing some things. I added `DefaultValue` to pipeline/initial which exactly matches the metadata type. In pipeline/general I expanded the type a little:

```
#[derive(Debug, Clone, Node)]
pub enum DefaultValue {
    Default(TypeNode),
    Literal(LiteralNode),
}
```

So the `Default` variant now has a value, the `TypeNode` - so that rendering the default value knows the type for the default value. What I *have not* done is work out how to wire any of these together.

The end result is that I'm getting an error running the tests:

```
Error: Node conversion error: expected LiteralMetadata, actual DefaultValueMetadata::Literal { 0: LiteralMetadata::Int { 2: Type::Int32, 1: Radix::Decimal, 0: 42i64 } } (path: <root>.fields[6].default.Literal::0.Some::inner)
```

which I think is saying I'm using `LiteralMetadata` where I should be using `DefaultValueMetadata`, but I can't see where - `LiteralMetadata` is now only used directly by Enum's `discr`.

Can you help me work out what I'm yet to do?

uniffi_meta/src/lib.rs
```diff
@@ -272,7 +272,7 @@ pub enum LiteralMetadata {
     EmptySequence,
     EmptyMap,
     None,
-    Some { inner: Box<LiteralMetadata> },
+    Some { inner: Box<DefaultValueMetadata> },
```
bendk, line 275: I think this might be the issue, this type should still be a `LiteralMetadata` right?

bendk:
I found one possibility, tell me if that works or not.

I also think the error message could be improved a lot:

```
Error: Node conversion error: expected LiteralMetadata, actual DefaultValueMetadata::Literal { 0: LiteralMetadata::Int { 2: Type::Int32, 1: Radix::Decimal, 0: 42i64 } } (path: <root>.fields[6].default.Literal::0.Some::inner)
```

a) "actual" and "expected" don't mean much.  Full module paths would be better.
b) It calculates the path to the error, which is really useful, but then it burries in the details.
c) The path string is really hard to read.  Maybe we should expand it

How about something like this:

```
Error converting node from previous IR:

RecordMetadata {
   fields: [
      ..., // 5 previous items
      FieldMetadata {
         default: Some(
            DefaultValueMetadata::Literal(
                Some(
                    <ERROR>
                )
            )
        ),
      },
   ],
}

Prev IR node: uniffi_meta::DefaultValueMetadata
Next IR node: pipeline::initial::LiteralMetadata
```

uniffi_meta/src/lib.rs
mhammond: This was in my initial patch (ie, worked perfectly before the pipeline) and it's `DefaultValueMetadata` more for convenience as it makes other things easier. However, I didn't adjust the copy of `LiteralMetadata` in the Python pipeline accordingly. I've adjusted this which seems to get me a little closer.

mhammond:
> How about something like this:

That looks like an improvement - not clear if it would have helped me though. Regardless, in the new version I'm now seeing:

> Node conversion error: expected DefaultValueNode, actual DefaultValue::Literal { 0: LiteralNode { lit: Literal::Int { 2: TypeNode { is_used_as_error: false, ty: Type::Int32, ffi_type: FfiTypeNode { ty: FfiType::Int32 }, canonical_name: "Int32" }, 1: Radix::Decimal, 0: 21i64 } } } (path: <root>.modules["proc_macro"].functions[3].callable.arguments[0].default)

which makes a little more sense to me, because I haven't arranged anywhere for `DefaultValue` and `DefaultValueNode` to be associated with each other.

mhammond:
In particular, [this type](https://github.com/mozilla/uniffi-rs/blob/62760f4b4f7b1e2d8cda0d23cd6a63405d3af212/uniffi_bindgen/src/pipeline/general/nodes.rs#L156) has no `#[node(wraps)]` whereas `LiteralNode` after it does - but I can't see how to use `wraps` on this. I'm *guessing* it's this magic which makes things work correctly for the other types.

bendk:
> In particular, [this type](https://github.com/mozilla/uniffi-rs/blob/62760f4b4f7b1e2d8cda0d23cd6a63405d3af212/uniffi_bindgen/src/pipeline/general/nodes.rs#L156) has no `#[node(wraps)]` whereas `LiteralNode` after it does - but I can't see how to use `wraps` on this. I'm _guessing_ it's this magic which makes things work correctly for the other types.

Yes that's correct.

`python::Argument::default` stores an `Option<DefaultValueNode>`, but  `general::Argument::default` stores an Option<DefaultValue>` -- so you're getting the error.  If `DefaultValueNode::default` had a `#[node(wraps)]` attribute, then the wrapping logic would kick in and make the conversion work.

Like I mentioned, something about putting the `#[node(wraps)]` on the `DefaultValueNode` makes this feel a bit too magical to me.  Maybe we should just be very explicit and require some sort of attribution on `Argument::default` and every other attribute where this wrapping should take place.  It's extra typing, but maybe it makes it clearer what's going on.

mhammond:
> `If`DefaultValueNode::default`had a`#[node(wraps)]` attribute, then the wrapping logic would kick in and make the conversion work.

That "works", but best I can tell we never get the new node. Which makes sense because I don't understand here is how it would ever be possible to automatically create the new enum at https://github.com/mozilla/uniffi-rs/blob/62760f4b4f7b1e2d8cda0d23cd6a63405d3af212/uniffi_bindgen/src/pipeline/general/nodes.rs#L156 - this `DefaultValue::Default` takes a `TypeNode`, which will need to come from the container (ie, the `Argument` or `Field`)

IOW, I'm expecting that I need to write explicit code which takes an `Argument` or a `Field` and knows how to create the new `DefaultValue` enum with the correct `TypeNode` - that can't possibly happen magically, right?

bendk:
>  this DefaultValue::Default takes a TypeNode, which will need to come from the container (ie, the Argument or Field)
> 
> IOW, I'm expecting that I need to write explicit code which takes an Argument or a Field and knows how to create the new DefaultValue enum with the correct TypeNode - that can't possibly happen magically, right?

No, the code will initialize it to a default/meaningless value.  Then you can write a pass function that fills in the values.  Something like:

```
module.visit_mut(|arg: &mut Argument| {
   arg.visit_mut(|default: DefaultValue| {
      if let DefaultValue::Default(type_node) = default {
          type_node = arg.ty.clone();
      }
   });
});
```

mhammond:
Thanks! This seems to be working, but there are some stylistic issues:

[edit: realized I had some of this in the wrong "pipeline", so original questions don't make sense]

* It's not clear we need both `literal.rs` and `default.rs` - should I just move it all into `default.rs`?
* The extra clones in `default_value_pass` suck - it probably doesn't matter in practice, but is there something I'm missing here?

uniffi_bindgen/src/pipeline/general/default.rs
```diff
@@ -0,0 +1,27 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use super::*;
+
+pub fn pass(module: &mut Module) -> Result<()> {
+    // fields and arguments need `DefaultValue` conversion.
+    module.visit_mut(|arg: &mut Argument| {
+        // eek, this sucks - can't borrow it to clone in the body
+        let ty = arg.ty.clone();
+        arg.visit_mut(|default: &mut DefaultValue| {
```
bendk: In hindsight, I think my suggestion of a second `visit_mut` was silly.  Can you just use `arg.default` directly and avoid the clone?

bendk:
> * It's not clear we need both `literal.rs` and `default.rs` - should I just move it all into `default.rs`?

That makes sense to me.

> * The extra clones in `default_value_pass` suck - it probably doesn't matter in practice, but is there something I'm missing here?

Unfortunately no, the clones are currently needed.  So far I've only needed these for outer loops, like `module.name`.  Doing it for every argument does feel bad.  I think it's avoidable though -- although I agree it probably doesn't matter in practice.

uniffi_bindgen/src/pipeline/general/default.rs
bendk: ...and my feeling is that we should be able to avoid these kinds clones in most cases.  Even if `DefaultValue` wasn't so easy to reach, I think you could use `arg.default.visit_mut()` and borrow `arg.ty` inside the closure.

mhammond: doh! Yes, thanks.

uniffi_bindgen/src/bindings/python/pipeline/default.rs
```diff
@@ -0,0 +1,77 @@
...
+        },
+        // We assume the Literal pass has already run, so `py_lit` already has a value.
+        DefaultValue::Literal(lit) => lit.py_lit.clone(),
+    })
+}
+
+pub fn literals_pass(lit: &mut LiteralNode) -> Result<()> {
+    lit.py_lit = render_literal(&lit.lit)?;
+    Ok(())
+}
```
bendk: What do you think about making a `defaults_and_literals` pass?  The individual passes are really short and if they're combined you could ensure the ordering is correct.

uniffi_macros/src/default.rs
```diff
@@ -32,6 +36,7 @@ impl ToTokens for DefaultValue {
             DefaultValue::None(kw) => kw.to_tokens(tokens),
             DefaultValue::Some { inner, .. } => tokens.extend(quote! { Some(#inner) }),
             DefaultValue::EmptySeq(_) => tokens.extend(quote! { [] }),
+            DefaultValue::Default => unimplemented!(),
```
bendk: I think this could be implemented with `kw::default`.

uniffi_macros/src/export/attributes.rs
```diff
@@ -380,16 +380,29 @@ impl Parse for DefaultMap {
...
 }
 
 impl Parse for DefaultPair {
     fn parse(input: ParseStream<'_>) -> syn::Result<Self> {
-        Ok(Self {
-            name: input.parse()?,
-            eq_token: input.parse()?,
-            value: input.parse()?,
-        })
+        // I'm sure there is a better way here - either want (Ident = Value) or (Ident)
```
bendk, line 388: Maybe, but I'm not sure.  I've struggled with this very issue too and haven't found a good solution.

uniffi_meta/src/lib.rs
bendk: Could `DefaultValueMetadata` be eliminated if there was a `LiteralMetadata::Default` variant?  I don't think this should block the merge, but it could be a good followup if it worked.

mhammond:
> Could DefaultValueMetadata be eliminated if there was a LiteralMetadata::Default variant?

This is actually where I started, but ended up moving to this because it made some other things cleaner, and it seemed kinda wrong to be shoe-horning this as a "literal", because it isn't actually a litteral - it's more the *absence* of a literal. This seems more correct to me even though the cost of another enum. But yeah, maybe we could revisit this later.

uniffi_bindgen/src/bindings/python/pipeline/default.rs
mhammond: done - so I just called it `pass`

Issue #1020, opened 2021-08-10
Invalid Kotlin codegen for an error named "Error"


rfk:
The application-services autofill component defines an error enum named "Error", like this:

```
[Error]
enum Error {
   "OpenDatabaseError", "SqlError", ...
};
```

When we generate the Kotlin bindings from such a definition, we convert the name "Error" into "Exception" and end up trying to do this:

```
// Error Error
sealed class Exception: Exception()  {
    // Each variant is a nested class
    class OpenDatabaseException : Exception()
    class SqlException : Exception()
    ...
}
```

Which doesn't end well. It's confusing, but also it shadows the builtin name `Exception` which breaks our attempts to subclass `Exception` in other parts of the helper code.

A workaround is to not do that, changing the autofill component to use a more descriptive name. But this is a bad user experience, and we should either fix the codegen to work in this case, or error out cleanly.

rfk:
(Previously we would define an exception named `Error`, which also shadows a Java buitlin of the same name, but we don't use that name in our generated code anywhere so it never caused us any issues)

artfuldev:
The way we have currently solved this in our repo is to use a type alias in the rust code and use the type alias name in the udl 

```rust
type CustomError = custom::Error
```

This helped us work with minimal changes to our source.

But I agree this is an issue.

enricozb:
A similar error happens with swift as well, and I also had to use a type alias in `lib.rs` and had to refer to that alias in the UDL to solve this.

ecstatic-morse:
@jhugman It seems like using fully qualified names in the templates for builtin types (e.g. `kotlin.Exception`) would solve this relatively easily. Is there something I'm missing?

I'm hitting the same issue for code like:

```idl
[Enum]
interface ScalarValue {
  String(string value);
  Boolean(bool value);
  // ...
}
```
which ends up like this

```kotlin
sealed class ScalarValue {
    data class String(val `value`: String) : ScalarValue()
    data class Boolean(val `value`: Boolean) : ScalarValue()
}
```

PR #2544, opened 2025-05-26
Explicitly catch kotlin.Exception in async traits

marioradonic:
We had a crash in our Android apps because someone added an `Error` in uniffi that got translated to `Exception` in Kotlin.

This type then overshadowed the `kotlin.Exception` (like in https://github.com/mozilla/uniffi-rs/issues/1020) and our async traits started crashing the app because the exception was not handled in the binding.

The sync version of this method specifies the exception `kotlin.Exception` so I see no reason for async version not to specify it.

I couldn't find any tests that cover this. I could try and write failing tests, but this would take me much longer.

In general, I'd prefer if it was made impossible to export `Error` errors since this has the potential to cause huge confusions.


mhammond (approved):
Ouch, thanks!

PR #2548, opened 2025-05-27
Lifting/lowering documentation updates

bendk:
- Added some diagrams to the `lifting_and_lowering` docs.
- Split out the docs on FFI converter traits, updated them, and added some more details.

### DIFF
--- a/docs/manual/src/internals/ffi_converter_traits.md
+++ b/docs/manual/src/internals/ffi_converter_traits.md
@@ -1,0 +1,67 @@
+## Rust FFI conversion traits
+
+UniFFI leverages a set of FFI converter traits to implement [lifting and lowering](./lifting_and_lowering.md) on the Rust side.
+Each trait handles a single step in the lifting/lowering process (e.g. lifting an argument, lowering a return, etc.).
+We implement these traits for each type used in the exported API then leverage them in the codegen.
+
+For example, `uniffi::Lift` is used to lift values.
+To handle a function like `fn print(msg: String)`, the generated code will use:
+
+* `<String as Lift<crate::UniFfiTag>>::FfiType` when it needs to specify the FFI type (`RustBuffer` for strings).
+* `<String as Lift<crate::UniFfiTag>>::try_lift()` when it needs to lift an argument value.  In this example, this means taking an FFI value (`<String as Lift<crate::UniFfiTag>>::FfiType` AKA `RustBuffer`) and converting it into a Rust `String` for passing to the Rust function.
+
+Using a trait for this is important for proc-macros, which only see Rust tokens and don't know the surrounding context.
+For example, if macros always used `RustBuffer` as the FFI type whenever it sees `String`, then that would fail if users created a type alias like `type MyTypeAlias = String`.
+This may be unusual for `String`, but it's very common for `Result`.
+In general, any reasoning about the tokens is fragile and should be avoided.
+
+### `UniFfiTag` and the orphan rule
+
+One odd part about the above code is that the `Lift` trait has a generic parameter which is always set to `crate::UniFfiTag`.
+In general, all of the FFI converter traits have this parameter (i.e. we generate `Lower<crate::UniFfiTag>`, `LowerReturn<crate::UniFfiTag>`, etc.).
+What's the point of all of this?
+
+The main reason is to work around issues with the Rust orphan rule and types from 3rd-party crates.
+For example, the [custom types](../types/custom_types.md) documentation shows how `url::Url` can be used in an exported API.
+For these types, we normally can't implement `Lift` in the code we generate in the crate since neither `uniffi::Lift` or `url::Url` is local to that crate.
+This same issue applies to all of the FFI converter traits.
+
+To work around this we:
+
+* Add a generic parameter to each trait (`Lift` becomes `Lift<UT>` where "UT" is short for `UniFfiTag`).
+* Define a unit struct in each crate named `UniFfiTag` (the term "tag" is borrowed from the [C++ template pattern](https://www.geeksforgeeks.org/tag-dispatch-in-cpp/)).
+* We use that unit struct as the generic parameter for the trait (e.g. `Lift<crate::UniFfiTag>` is used to lift a value).
+
+Using the local type as a generic parameter means the impl no longer violates the orphan rule.
+For details on this see the [Rust Chalk Book](https://rust-lang.github.io/chalk/book/clauses/coherence.html#the-orphan-rules-in-rustc)
+The TLDR is that generic parameters "count" towards the requirement that there be a local type in the impl.
+
+However, this makes it harder to use this impl from another crate.
+UniFFI handles that in 2 ways:
+
+* The `uniffi` crate generates blanket trait impls for all UniFFI tag params (`impl<UT> Lift<UT> for String`).
+  This allows all crates to use them automatically with their `UniFfiTag` struct.
+* UniFFI defines the `use_remote_type!` macro, which generates an implementation for the local
+  `UniFfiTag` by forwarding to the implementation from another crate's `UniFfiTag`.
+  See the [Remote and external types](../types/remote_ext_types.md#remote-external-types) for example usage.
+  This is also what the `remote` flag of the [custom type macro](../types/custom_types.md) does.
+
+### An incomplete list of FFI traits
+
+UniFFI defines a large number of FFI conversion traits, each one used for a specific purpose.
+This section describes a few them for explanatory purposes.
+See `uniffi_core/src/ffi_converter_traits.rs` for a full and up-to-date list.
+
+* `Lift`: Lift an value
+* `Lower`: Lower a value
+* `LowerReturn`: Lower a return value.
+  * For most types this is equivalent `Lower`, but a specialized impl is created for `Result<T, E>`.
+* `LiftRef`: Lift for a reference type.
+  This is often just `Lift` then a borrow, but a specialized impl is created for `Arc<T>`.
+* `FfiConverter`: General-purpose FFI conversion logic.
+  When `FfiConverter` is defined on a type, all other FFI traits are automatically derived.
+  This is what we implement for user-defined types like records and enums.
+* `FfiConverterArc`: FfiConverter implementation for `Arc<T>`.
+  This is another trait that we use to get around orphan rules.
+  Crates can't directly implement `FfiConverter` on `Arc<T>` for some interface, so they implement `FfiConverterArc` instead.
+  `uniffi` defines a blanket impl `FfiConverter` impl for these types (`impl<T: FfiConverterArc<UT>, UT> FfiConverter<UT> for Arc<T>`).
--- a/docs/manual/src/internals/lifting_and_lowering.md
+++ b/docs/manual/src/internals/lifting_and_lowering.md
@@ -13,11 +13,50 @@ receiving side "***lifts***" that primitive type into its own language-specific
 data type.
 
 Lifting and lowering simple types such as integers is done by directly casting the
-value to and from an appropriate type. For complex types such as optionals and
-records we currently implement lifting and lowering by serializing into a byte
-buffer, but this is an implementation detail that may change in future. (See
-[ADR-0002](https://github.com/mozilla/uniffi-rs/blob/main/docs/adr/0002-serialize-complex-datatypes.md) for the reasoning
-behind this choice.)
+value to and from an appropriate type. Non-trivial types such as Strings, Optionals and
+Records, etc. are lowered to a byte buffer called a `RustBuffer` internally.
+
+For example, a Python `str` is passed to Rust by lowering it to a `RustBuffer`, which is then
+lifted to a Rust `String`
+
+```mermaid
+block-beta
+  columns 2
+  block:PythonLayer
+    columns 1
+    PythonTitle["Python"]
+    PythonStr["str"]
+  end
+  block:RustLayer
+    columns 1
+    RustTitle["Rust"]
+    RustString["String"]
+  end
+  block:FFILayer:2
+    columns 3
+    space:2
+    FfiTitle["FFI"]
+    space
+    RustBuffer["RustBuffer"]
+    space
+  end
+
+  PythonStr --"Lower"--> RustBuffer
+  RustBuffer --"Lift"--> RustString
+
+classDef default stroke:transparent,fill:#fff
+classDef layer stroke:#c0c0c0,fill:#cae9ff
+classDef title stroke:transparent,fill:transparent,color:#f72585
+classDef invisible stroke:transparent,fill:transparent
+class RustLayer layer
+class PythonLayer layer
+class FFILayer layer
+class RustTitle title
+class PythonTitle title
+class FfiTitle title
+```
+
+## Lifting and Lowering when calling functions
 
 As a concrete example, consider this interface for accumulating a list of integers:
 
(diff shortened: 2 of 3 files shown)


docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,67 @@
...
+
+The main reason is to work around issues with the Rust orphan rule and types from 3rd-party crates.
+For example, the [custom types](../types/custom_types.md) documentation shows how `url::Url` can be used in an exported API.
+For these types, we normally can't implement `Lift` in the generated code since neither `uniffi::Lift` or `url::Url` is local to the crate.
+This same issue applies to all of the FFI converter traits.
+
+To work around this we:
+  - Add a generic parameter to each trait (`Lift` becomes `Lift<UT>` where "UT" is short for `UniFfiTag`).
+  - Define a unit struct in each crate named `UniFfiTag` (the term "tag" is borrowed from the [C++ template pattern](https://www.geeksforgeeks.org/tag-dispatch-in-cpp/)).
+  - We use that unit struct as the generic parameter for the trait (e.g. `Lift<crate::UniFfiTag>` is used to lift a value).
```
badboy, line 32: No need for indentation and for correct rendering we need a newline
```suggestion
To work around this we:

- Add a generic parameter to each trait (`Lift` becomes `Lift<UT>` where "UT" is short for `UniFfiTag`).
- Define a unit struct in each crate named `UniFfiTag` (the term "tag" is borrowed from the [C++ template pattern](https://www.geeksforgeeks.org/tag-dispatch-in-cpp/)).
- We use that unit struct as the generic parameter for the trait (e.g. `Lift<crate::UniFfiTag>` is used to lift a value).
```

docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,68 @@
+## Rust FFI conversion traits
+
+UniFFI leverages a set of FFI converter traits to implement [lifting and lowering](./lifting_and_lowering.md) on the Rust side.
+Each trait handles a single step in the lifting/lowering process (e.g. lifting an argument, lowering a return, etc.).
+We implement the trait for each type used in the exported API then use that trait in the codegen.
+
+For example, `uniffi::Lift` is used to lift values.
+To handle a function like `fn print(msg: String)`, the generated code will use:
+
+- `<String as Lift<crate::UniFfiTag>>::FfiType` when it needs to specify the FFI type (`RustBuffer`).
```
mhammond, line 10: maybe "(`RustBuffer` for strings)

docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,68 @@
...
+- `<String as Lift<crate::UniFfiTag>>::try_lift()` when it needs to lift an argument value.
```
mhammond, line 11: maybe end with "(ie, in this example, to take an FFI value of type `<String as Lift<crate::UniFfiTag>>::FfiType` (as above, a`RustBuffer`) and convert it into a Rust string suitable for passing to the Rust function)" or similar?

docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,68 @@
...
+
+Using a trait for this is important for proc-macros, which only see Rust tokens and don't know the surrounding context.
+For example, if macros always used `RustBuffer` as the FFI type whenever it sees "String", then that would fail if users created a type alias like `type String = u8`.
```
mhammond, line 14: `type String = u8` seems like an odd example - would using something like `Guid` or even `Url` be better? It makes `String` seems less unusual.

docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,68 @@
...
+
+### `UniFfiTag` and the orphan rule
+
+One odd part about the above code is that the `Lift` trait has a generic parameter which is always set to `crate::UniFfiTag`.
+In general, all of the FFI converter traits have this parameter (i.e. we generate `Lower<crate::UniFfiTag>`, `LowerReturn<crate::UniFfiTag>`, etc.).
+What's the point of all of this?
+
+The main reason is to work around issues with the Rust orphan rule and types from 3rd-party crates.
+For example, the [custom types](../types/custom_types.md) documentation shows how `url::Url` can be used in an exported API.
+For these types, we normally can't implement `Lift` in the generated code since neither `uniffi::Lift` or `url::Url` is local to the crate.
```
mhammond, line 26: s/generated code/code we generate in the crate/ here, to help keep the focus on that orphan rule?

docs/manual/src/internals/lifting_and_lowering.md
```diff
@@ -13,11 +13,55 @@ receiving side "***lifts***" that primitive type into its own language-specific
 data type.
 
 Lifting and lowering simple types such as integers is done by directly casting the
-value to and from an appropriate type. For complex types such as optionals and
-records we currently implement lifting and lowering by serializing into a byte
-buffer, but this is an implementation detail that may change in future. (See
-[ADR-0002](https://github.com/mozilla/uniffi-rs/blob/main/docs/adr/0002-serialize-complex-datatypes.md) for the reasoning
-behind this choice.)
+value to and from an appropriate type. Complex types such as optionals and
+records are lowered to a byte buffer called a `RustBuffer` internally.
```
mhammond, line 17: Maybe replace "Complex types such as optionals and records are lowered" with something like "Non-trivial types such as Strings, Optionals, Records, etc"? This ties in better with your String example in the ffi traits docs. Not that happy with "non-trivial" either, but it's mildly better than "complex"? Even "primitive" isn't ideal because (a) String is a primitive and it doesn't handle pointers. Whatever you decide :)

docs/manual/src/internals/lifting_and_lowering.md
```diff
@@ -13,11 +13,55 @@ receiving side "***lifts***" that primitive type into its own language-specific
...
+classDef invisible stroke:transparent,fill:transparent
+class RustLayer layer
+class PythonLayer layer
+class FFILayer layer
+class RustTitle title
+class PythonTitle title
+class FfiTitle title
+```
+
+Note: `RustBuffer` is an implementation detail that may change in future. (See
```
mhammond, line 59: I'd go further here and say "ALL of this, including the us of a RustBuffer ..." :)

docs/manual/src/internals/lifting_and_lowering.md
```diff
@@ -27,27 +71,61 @@ namespace example {
...
-7. The foreign language bindings receive the return value and need to ***lift*** it into an
-   appropriate native data type. Since it is a complex data type, it is lifted by deserializing
-   from the returned byte buffer into a language-native list of integers.
+3. The generated Python function passes the lowered arguments to the Rust scaffolding function.
+   This is a `repr(C)` FFI function in Rust library and named `uniffi_fn_add_to_list` in this example.
+4. The Rust scaffolding function ***lifts*** each argument received over the FFI into a native Rust type.
+   Since `item` is a plain integer no conversion is needed.
+5. The Rust scaffolding passes the lifted arguments to the user-defined Rust `add_to_list` function, which then executes normally.
+6. The Rust scaffolding function receives the return value and now needs to ***lower*** the it to pass it back across the FFI.
+   Since this is a complex data type, it is lowered by serializing the values into a byte buffer (`RustBuffer`), which is then returned.
```
mhammond, line 126: Maybe "Since this is a complex data type" can be replaced with "Since this type's `::FfiType` is a RustBuffer`..." (ie, the point here isn't about the complexity of the type) - or even "Since Vec's have their `::FfiType` as a `RustBuffer`"?

docs/manual/src/internals/lifting_and_lowering.md
```diff
@@ -27,27 +71,61 @@ namespace example {
...
+7. The generated Python function receives the return value, and then ***lifts*** it to a native data type.
+   Since it is a complex data type, it is lifted by deserializing from the RustBuffer a language-native list of integers.
```
mhammond, line 128: and kinda ditto here - the point isn't about the complexity of the type, it's about matching the `::FfiType`.

mhammond (approved):
This is a great improvement, thanks. My comments are just what popped into my head when reading and aren't necessarily an improvement, do with them whatever you like!

docs/manual/src/internals/lifting_and_lowering.md
bendk: Yeah, it's always hard to come up with words for this.  But I like "non-trivial".

bendk:
Great improvements, thanks!  My plan is to leave this out out for review for a week or so to let more trickle in.  I'm also hoping to push out one or two more of these.

docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,68 @@
+## Rust FFI conversion traits
+
+UniFFI leverages a set of FFI converter traits to implement [lifting and lowering](./lifting_and_lowering.md) on the Rust side.
+Each trait handles a single step in the lifting/lowering process (e.g. lifting an argument, lowering a return, etc.).
+We implement the trait for each type used in the exported API then use that trait in the codegen.
```
badboy: Above it talks about multiple traits (but never names them), and this sentence now talks about _the trait_ (singular). I feel like that's a bit hard to understand without already knowing the details.
Maybe we can be a bit more specific here.

badboy: Ah, I see the list of traits is further below. I still think this sentence could be worded slightly differently, I'll think about it

docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,68 @@
...
+
+For example, `uniffi::Lift` is used to lift values.
+To handle a function like `fn print(msg: String)`, the generated code will use:
+
+* `<String as Lift<crate::UniFfiTag>>::FfiType` when it needs to specify the FFI type (`RustBuffer` for strings).
+* `<String as Lift<crate::UniFfiTag>>::try_lift()` when it needs to lift an argument value.  In this example, this means taking an FFI value (`<String as Lift<crate::UniFfiTag>>::FfiType` AKA `RustBuffer`) and converting it into a Rust String for passing to the Rust function.
```
badboy:
```suggestion
* `<String as Lift<crate::UniFfiTag>>::try_lift()` when it needs to lift an argument value.  In this example, this means taking an FFI value (`<String as Lift<crate::UniFfiTag>>::FfiType` AKA `RustBuffer`) and converting it into a Rust `String` for passing to the Rust function.
```

docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,68 @@
...
+
+Using a trait for this is important for proc-macros, which only see Rust tokens and don't know the surrounding context.
+For example, if macros always used `RustBuffer` as the FFI type whenever it sees "String", then that would fail if users created a type alias like `type MyTypeAlias = String`.
```
badboy: Let's use the same formatting throughout for types
```suggestion
For example, if macros always used `RustBuffer` as the FFI type whenever it sees `String`, then that would fail if users created a type alias like `type MyTypeAlias = String`.
```

docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,68 @@
...
+
+* Add a generic parameter to each trait (`Lift` becomes `Lift<UT>` where "UT" is short for `UniFfiTag`).
+* Define a unit struct in each crate named `UniFfiTag` (the term "tag" is borrowed from the [C++ template pattern](https://www.geeksforgeeks.org/tag-dispatch-in-cpp/)).
+* We use that unit struct as the generic parameter for the trait (e.g. `Lift<crate::UniFfiTag>` is used to lift a value).
+
+Using the local type as a generic parameter means the impl no longer violates the orphan rule.
+For details on this see the [Rust Chalk Book](https://rust-lang.github.io/chalk/book/clauses/coherence.html#the-orphan-rules-in-rustc)
+The TLDR is that generic parameters "count" towards the requirement that there be a local type in the impl.
+
+In essence, `Lift<crate::UniFfiTag>` is a crate-local version of the `Lift` trait which gets around the orphan rule.
```
badboy: These two sentences seem to be roughly the same thing. IMO the first one can be removed.

docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,68 @@
...
+However, this makes it harder to use this impl from another crate.
+UniFFI handles that in 2 ways:
+
+* The `uniffi` crate generates blanket trait impls (`for <UT> impl Lift<UT> for String`).
```
badboy: This syntax seems confusing to me.

docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,68 @@
...
+### An incomplete list of FFI traits
+
+UniFFI defines a large number of FFI conversion traits, each one used for a specific purpose.
+This section describes a few them for explanatory purposes.
+See `uniffi_core/src/ffi_converter_traits.rs` for a full and up-to-date list.
+
+* `Lift`: Lift an value
+* `Lower`: Lower a value
+* `LowerReturn`: Lower a return value.
+  * For most types this is equivalent `Lower`, but a specialized impl is created for `Return<T, E>`.
```
badboy: Did you mean `Result`?
```suggestion
  * For most types this is equivalent `Lower`, but a specialized impl is created for `Result<T, E>`.
```

docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,68 @@
...
+* `LiftRef`: Lift for a reference type.
+  This is often just `Lift` then a borrow, but a specialized impl is created for `Arc<T>`.
+* `FfiConverter`: General-purpose FFI conversion logic.
+  When `FfiConverter` is defined on a type, all other FFI traits are automatically derived.
+  This is what we implement for user-defined types like records and enums.
+* `FfiConverterArc`: FfiConverter implementation for `Arc<T>`.
+  This is another trait that we use to get around orphan rules.
+  Crate's can't directly define `FfiConverter` on `Arc<T>` for some interface, so they define `FfiConverterArc` instead.
```
badboy:
```suggestion
  Crates can't directly define `FfiConverter` on `Arc<T>` for some interface, so they define `FfiConverterArc` instead.
```

badboy: Define? Or implement?

docs/manual/src/internals/ffi_converter_traits.md
```diff
@@ -0,0 +1,68 @@
...
+  `uniffi` defines a blanket `for T: FfiConverterArc impl FfiConverter<Arc<T>> `
```
badboy: Same syntax again.
As this is not valid Rust syntax as is, should we maybe explain that better?

badboy (commented):
took a bit longer than I planned, but hey, within the week you promised to keep it open!
Content-wise this is a really good explanation. I'm all for landing it now.

I did add a few comments for some style and wording that I think should clear up the remaining vague points.

bendk:
Thanks for the review.  I took your suggestions and tried to reformat the rest of the text.

That syntax was pretty bad, sorry about that.  I always forget exactly how to spell out those blanket impls, which kind of argues for not using them in the documentation.  OTOH, English isn't the best way to write them out either.  I fixed the syntax and added few more words, any suggestions here would be great though.

bendk:
Thanks for all the help here, I think I'm going to merge this one so that I can close out my JIRA ticket.  I'm very happy to continue iterating on these, just open a PR/issue.

PR #2550, opened 2025-05-29
Internal docs for Rust/callback interface calls

bendk:


docs/manual/src/internals/foreign_calls.md
```diff
@@ -0,0 +1,84 @@
...
+
+How can Rust code make a call into the foreign language?
+The [lifting and lowering](./lifting_and_lowering.md) docs describe this at a high-level, this document will explore the details.
+
+This document only describes non-async calls.
+See the [Async FFI](./async-ffi.md) docs for a description of async calls.
+
+## VTables
+
+One major difference from a [Foreign -> Rust call](./rust_calls.md), is that the Rust code does not have access to a library with functions it can call.
```
mhammond, line 11: This only applies to objects and traits, not top-level functions, right? So maybe make that clear?

docs/manual/src/internals/foreign_calls.md
```diff
@@ -0,0 +1,84 @@
...
+Because of this, the foreign code registers a "VTable" with the Rust code.
+This is a `repr(C)` struct where each method is a function pointer corresponding to a method in a callback interface.
```
mhammond, line 13: s/method/entry/ or something?  (ie, I don't really think of a struct having a "method", unless it is a function pointer, which is the point you are making?)

docs/manual/src/internals/foreign_calls.md
```diff
@@ -0,0 +1,84 @@
...
+This is because some foreign languages, like Python, have issues returning C structs like `RustBuffer`.
+Methods that don't return anything still have an out pointer argument, but it's never written to.
+
+The first argument is a `u64` handle rather than a `void *`.
+In both cases however, it's a handle that represents a reference the underlying object.
+These are different for historical reasons only, we hope to eventually use the same type for both cases.
+
+Different bindings have different systems for callback interface handles, but they generally fall into 2 camps:
+* The handle is the key to a hash map that maps `u64` -> callback interface objects.
+* The is a raw pointer to a callback interface object casted to a `u64`.
```
mhammond, line 55: s/casted/cast/?

docs/manual/src/internals/foreign_calls.md
```diff
@@ -0,0 +1,84 @@
...
+  * Nothing is written to `uniffi_out_return`.
+* If the method throws some other exception, then:
+  * The foreign bindings set `uniffi_call_status.code` to `RustCallStatusCode::InternalError`
+  * The foreign bindings should try to to serialize the error message to a `RustBuffer` and write it to `error_buf`.
+    However, bindings may also not write to `error_buf` and the generated Rust code will handle this.
+  * Nothing is written to `uniffi_out_return`.
+  * If the callback method returns a `Result<T, E>` type and the bindings implement `From<uniffi::UnexpectedUniFFICallbackError> for E`,
+    then the generated Rust will convert the error to an `Err(E)` value and return it.
+  * If not, then the Rust bindings will panic.
+  * This is why it's recommended to always have callback methods return `Result` types and implement `From<uniffi::UnexpectedUniFFICallbackError>`.
```
mhammond, line 72: Maybe combine this sentence with the above (ie, it's part of the point above rather than a new point)

mhammond (approved):
Thanks for this! As usual, take these comments for what they cost!

docs/manual/src/internals/foreign_calls.md
mhammond: yeah, sorry, I was confused :) We don't have plain foreign functions (yet? ;) - they are always in an interface. Maybe a sentence like "because foreign bindings are only called via an interface, the foreign references are all managed via a vtable" or similar - that was an unmentioned difference between the 2 docs I noticed.

docs/manual/src/internals/foreign_calls.md
```diff
@@ -0,0 +1,88 @@
...
+pub trait TodoList {
+    fn append(&self, title: String);
+    fn get_items(&self) -> Vec<String>;
+}
+
+/// A VTable like this will be generated:
+#[repr(C)]
+pub struct UniffiTodoListVTable {
+    // One function pointer field for each method
+    add: extern "C" fn(uniffi_handle: u64, title: RustBuffer, uniffi_out_return: &mut (), uniffi_call_status: &mut RustCallStatus),
```
gruberb: Shouldn't this be `append`?

docs/manual/src/internals/foreign_calls.md
```diff
@@ -0,0 +1,88 @@
...
+
+This document only describes non-async calls.
+See the [Async FFI](./async-ffi.md) docs for a description of async calls.
+
+## VTables
+
+Calling into foreign code has 2 major differences from [calling Rust code](./rust_calls.md):
+
+* Foreign code is only called via an interface.
+* The Rust code does not have access to a library with functions it can call.
```
gruberb: I would like to reword this _slightly_. Like what does it mean? Does it mean that "Rust doesn't know the functions it can call on the foreign language side". 

It clashes a bit with the next paragraph, which says that the foreign code registers functions Rust can call?

I think I know what you mean, but I had to read it a few times to get it.

docs/manual/src/internals/foreign_calls.md
```diff
@@ -0,0 +1,88 @@
...
+
+Because of this, the foreign calls all go through a vtable that the foreign code registers with the Rust code.
```
gruberb: What does "registers" mean? Who does the registering? 

Is it: When UniFFI generates the foreign language template, it creates a VTable automatically with the functions Rust can call which are implemented on the foreign language through the build process in UniFFI?

docs/manual/src/internals/foreign_calls.md
```diff
@@ -0,0 +1,88 @@
...
+    fn append(&self, title: String);
+    fn get_items(&self) -> Vec<String>;
+}
+
+/// A VTable like this will be generated:
+#[repr(C)]
+pub struct UniffiTodoListVTable {
+    // One function pointer field for each method
+    add: extern "C" fn(uniffi_handle: u64, title: RustBuffer, uniffi_out_return: &mut (), uniffi_call_status: &mut RustCallStatus),
+    get_items: extern "C" fn(uniffi_handle: u64, uniffi_out_return: &mut RustBuffer, uniffi_call_status: &mut RustCallStatus);
```
gruberb, line 35: Would be nice to add a paragraph why we use `RustBuffer`.

docs/manual/src/internals/foreign_calls.md
```diff
@@ -0,0 +1,88 @@
...
+#[no_mangle]
+pub extern "C" fn uniffi_init_todo_list_vtable(vtable: std::ptr::NonNull<UniffiTodoListVTable>) {...}
+```
+
+
+## VTable methods
+
+VTable methods work similar to a [Rust method call](./rust_calls.md)], with some exceptions.
+
+Instead of returning the value as normal, it's written to an out pointer.
```
gruberb: Super nit, but is it that returning means the foreign side writes the return value in a block of memory (heap or stack) and returns a pointer which points to the place in memory the Rust function can read the value from? Which then, it needs to "lift" and make it a Rust type?

docs/manual/src/internals/foreign_calls.md
```diff
@@ -0,0 +1,88 @@
...
+This is because some foreign languages, like Python, have issues returning C structs like `RustBuffer`.
```
gruberb: What does this mean, having issues returning C structs?

docs/manual/src/internals/foreign_calls.md
```diff
@@ -0,0 +1,88 @@
...
+Methods that don't return anything still have an out pointer argument, but it's never written to.
+
+The first argument is a `u64` handle rather than a `void *`.
```
gruberb: Sorry, but I also don't understand why it would be `void *` in the first place :)

docs/manual/src/internals/foreign_calls.md
bendk: Yes that wasn't very clear.  "Register" here means the foreign code passes the callback to a generate Rust function so that the Rust code knows about it.  I just removed all of the wording, because the concept is shown in the code below and I think it's clearer there.

docs/manual/src/internals/foreign_calls.md
bendk: Not exactly, Rust passes in the pointer rather than having the foreign side return it.  That makes the memory management a bit simpler.  I updated the wording, tell me if it's clearer now.

docs/manual/src/internals/foreign_calls.md
bendk: It means ctypes throws an exception rather than returns the value.  The design choice completely baffles me.  I updated the text and linked to the issue for that.

docs/manual/src/internals/foreign_calls.md
bendk: I was trying to contrast and compare callback interfaces with "Rust interfaces" (i.e. Rust structs where we expose their methods).  I don't think it's that useful though, especially since I'm hoping to make them use the same handle type, so I just removed the section.

bendk:
Thanks for all the help here, I think I'm going to merge this one so that I can close out my JIRA ticket. I'm very happy to continue iterating on these, just open a PR/issue.

Issue #2563, opened 2025-06-06
Errors Produced in `try_lift` are difficult to Debug.


oldwomanjosiah:
Recently I was working on adding support for sending [`Uuid`] across the FFI boundary as a `String`. Originally, I had introduced this conversion using a bare

```rust
use uuid::Uuid;

uniffi::custom_type!(Uuid, String, { remote });
```

This appeared to work on Android and iOS when converting from `ParcelUuid` and `CBUUID` respectively, but would periodically crash with just "invalid length: expected length 32 for simple format, found 4" when running on iOS. This message is coming from the error returned from `Uuid::try_from(String)`, which was being lifted for a field of type `Vec<Uuid>`, and was being passed a short UUID as specified by Bluetooth 4.

I understand that this isn't necessarily this libraries responsibility, but it would be really nice if (when the `ffi-trace` feature is enabled) more context would be attached to the anyhow report to aid in debugging.

I'm thinking:

1. Unconditionally attach type names and source location information to `custom_type!()` invocations.
2. When `ffi-trace` is enabled:
    * Attach type name in try_lift
    * Attach field name in `try_read_field`

oldwomanjosiah:
I would also be interested in contributing to these changes, if they are something that the project would be interested in but aren't a high priority for anyone else atm.

mhammond:
hi, thanks for looking at uniffi! We'd be happy to look at zero (ish?) cost improvements to our diagnostics here.

mhammond:
Let's call this fixed by #2564, but we'd welcome more improvements or concrete problem cases in new PRs/issues.

PR #2564, opened 2025-06-07
Add Context to Custom Type Lifting

oldwomanjosiah:
Relates to #2563 

## Motivation

In many cases it is helpful to use the `custom_type!()` macro to delegate the conversion between domain and FFI-types, but this means that users of the library do not have direct control over the error messages which are produced in cases where the conversion fails. These are currently surfaced by the host language as an "internal error", which can lead to confusion when debugging.

## Proposed Change

- Use `anyhow::Context::with_context` to attach information on the conversion being attempted, and where in the codebase it was defined, when the `try_from` call returns an error.
- Update `LowerReturn` default and override on `Result` to format the full anyhow context chain, instead of just the topmost error.

### DIFF
--- a/uniffi/tests/custom_types.rs
+++ b/uniffi/tests/custom_types.rs
@@ -45,7 +45,7 @@ fn test_string_validation_failure() {
     let result: anyhow::Result<RecordWithValidatedStrings> =
         uniffi::Lift::<UniFfiTag>::try_lift(uniffi::Lower::<UniFfiTag>::lower(base));
 
-    let expected = r#"Lifting custom type `custom_types::ValidatedString` from FFI type `alloc::string::String` failed at uniffi/tests/custom_types.rs:39
+    let expected = r#"Lifting custom type `custom_types::ValidatedString` from FFI type `alloc::string::String` failed at uniffi/tests/custom_types.rs:34
 
 Caused by:
     Value cannot contain 'invalid'"#;
--- a/uniffi_macros/src/custom.rs
+++ b/uniffi_macros/src/custom.rs
@@ -192,9 +192,7 @@ pub(crate) fn expand_custom_type(args: CustomTypeArgs) -> syn::Result<TokenStrea
         Some(convert_closure) => convert_closure.token_tuple(),
         None => (
             quote! { val },
-            quote! {
-                ::core::result::Result::Ok(<#uniffi_type as TryInto<#custom_type>>::try_into(val)?)
-            },
+            quote! { ::core::result::Result::Ok(<#uniffi_type as TryInto<#custom_type>>::try_into(val)?) },
         ),
     };
 
@@ -207,6 +205,18 @@ pub(crate) fn expand_custom_type(args: CustomTypeArgs) -> syn::Result<TokenStrea
                 ::core::any::type_name::<#uniffi_type>(),
                 location.file(),
                 location.line()
+            )
+        }
+    };
+
+    // We allow needless_question_mark here because the error type of `TryInto` only needs conversion
+    // iff it is not a `anyhow::Error` already.
+    let try_lift_expr = quote::quote! {
+        #[allow(clippy::needless_question_mark)]
+        {
+            ::uniffi::deps::anyhow::Context::with_context(
+                (move || -> ::uniffi::Result<#custom_type> { #try_lift_expr })(),
+                #build_context
             )
         }
     };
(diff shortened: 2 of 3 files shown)


uniffi_core/src/ffi_converter_impls.rs
```diff
@@ -486,13 +486,14 @@ where
...
-            }
-        }
+        let crate::LiftArgsError { arg_name, error } = error;
+
+        let error = match error.downcast::<E>() {
+            Ok(user) => RustCallError::Error(<E as LowerError<UT>>::lower_error(user)),
+            Err(error) => crate::LiftArgsError { arg_name, error }.to_internal_error(),
+        };
+
+        Err(error)
```
bendk, line 496: Does this change the behavior or is it just a refactoring?  I kind of prefer the original version, since it keeps all the logic in one place.  What's the motivation to change this?

fixtures/ext-types/custom-types/tests/bindings/test_custom_types.py
```diff
@@ -10,6 +10,13 @@ def run(self, guid):
         self.saw_guid = guid
         return guid
 
+def get_value_arg_issue(value):
+    return f"""Failed to convert arg 'value':
+Lifting custom type `ext_types_custom::Guid` from FFI type `alloc::string::String` failed at fixtures/ext-types/custom-types/src/lib.rs:101
```
bendk, line 15: Is this enough context to understand the error?  Pointing to the `custom_type!` macro is feels like an improvement, but maybe a full backtrace would be even better.  That's probably more expensive though, so maybe there should be a feature flag for it.

I'm mostly just thinking out loud.  If this string feels good to you, then I'm happy to go with it now and and we can continue to iterate on it.

uniffi_macros/src/custom.rs
```diff
@@ -192,10 +192,35 @@ pub(crate) fn expand_custom_type(args: CustomTypeArgs) -> syn::Result<TokenStrea
...
+        }
+    };
+
+    // We allow needless_question_mark here because the error type of `TryInto` only needs conversion
+    // iff it is not a `anyhow::Error` already.
+    let try_lift_expr = quote::quote! {
+        #[allow(clippy::needless_question_mark)]
+        {
+            ::uniffi::deps::anyhow::Context::with_context(
+                (move || -> ::uniffi::Result<#custom_type> { #try_lift_expr })(),
```
bendk, line 218: Do you need a closure for this?  I would have thought `#try_lift_expr` would work by itself here.

uniffi/tests/custom_types.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub name: ValidatedString,
+    pub children: Vec<ValidatedString>,
+}
+
+uniffi::custom_type!(ValidatedString, String);
+
+uniffi::setup_scaffolding!();
+
+#[test]
+fn test_string_validation_failure() {
```
bendk, line 39: Does this test run?  I'm not seeing it included in the module tree, but I might just be confused by the `trybuild` system.

uniffi_macros/src/custom.rs
oldwomanjosiah: A closure is needed here to catch errors propagated with `?` or `return` in the user provided code

uniffi_core/src/ffi_converter_impls.rs
oldwomanjosiah: The behavior here is the same, but I pulled out the formatting out so that we could use the same function for the default impl and this override.

It's not enough code that that feels super necessary, but the duplication tripped me up while I was working on this so I figured I'd centralize.

uniffi/tests/custom_types.rs
oldwomanjosiah: Yeah, rust files in the `<crate root>/tests` directory are automatically included as "integration tests". Originally, I placed this here as it used `cargo-insta` to validate the error message and it felt cleaner to have the snapshots out of the main src tree, but the [dependency managment docs](https://github.com/mozilla/uniffi-rs/blob/main/docs/dependency-management.md) are missing and I didn't want to add it without being able to review the rules there.

I can move these back into `src/lib.rs::tests` if you would prefer.

fixtures/ext-types/custom-types/tests/bindings/test_custom_types.py
oldwomanjosiah: I'm not dead set on this specific message, but I do think that adding in the attempted conversion and source location is enough to catch most issues that a user would run into. I'm not entirely sure when `anyhow` will choose to print out a backtrace, but I do know that it should be propagating it if it is captured by the source error.

I originally had a version that was more explicit about the fact that we had delegated to `try_from`, but it felt both too verbose and also potentially incorrect when the end user provided some custom conversion code.

uniffi/tests/custom_types.rs
oldwomanjosiah: Okay actually, I just attempted to move the test as I mentioned, and all the macros assume that you will be able to import with the `::uniffi` syntax, which requires that uniffi be a crate dependency. This is true for integration tests, but not unit tests in the same crate

uniffi_core/src/ffi_converter_impls.rs
bendk: Oh, I missed that this removed some duplication.  It all makes sense to me now.

bendk (approved):
Thanks for the quick and thoughtful response.  All the changes seem well motivated to me, let's merge this.

Issue #2567, opened 2025-06-09
Trait interfaces can be wrapped multiple times


bendk:
When trait interfaces that have the `with_foreign` attribute are passed across the FFI, we always create a new wrapper layer.  For example:

- The trait interface is Rust is implemented in Rust
- An instance of the trait interface is passed to Python
- When Python code calls methods on the trait, behind the scenes we arrange for a Rust FFI call.  All good so far
- The Python code now passes this trait interface back to Rust.
- Here's where the problems start.  If the Rust code calls trait interface methods, it does so by making an FFI call to Python (via the vtable).  The Python code that handles the FFI call, then goes back and makes an FFI call to Rust.  We've just made a round-trip FFI call, for essentially no purpose.
- If the trait interface was passed back again, then this issue would just compound and method invocations would require 3 trips across the FFI.

To fix this we need to:
  - Make it possible to know which side of the FFI a trait interface handle came from.  For example, by using the lowest bit to signal Rust vs the foreign side.  Maybe we should use 2-3 bits for this, so that we could support https://github.com/mozilla/uniffi-rs/issues/2295.
  - When passing a trait interface handle across the FFI, if we know it came from the other side then clone the handle and pass that back directly.  Don't pass a handle to the callback interface implementation object that you constructed on your side of the FFI.
  - Take this into account when receiving a trait interface handle from across the FFI.  For example, on the Rust side:
     - If the handle came from Rust, then re-construct the `Arc<dyn Trait>` using something like `Arc::from_raw`.
     - If the handle came from the foreign side, then keep the current behavior: create a new instance of our callback interface handler struct that implements `Trait`.

mhammond:
Fixed via #2586

PR #2569, opened 2025-06-10
Use u64 for trait interface handles

bendk:
Use `u64` values for interfaces and trait interfaces rather than `void *`.  Now interfaces, callback interfaces, and trait interfaces all use `u64` handles as their FFI type.

* Trait interfaces with foreign support are simplified.  These effectively combine a callback interface and trait interface.  Before it was weird, since those 2 things had different FFI types.
* We were already casting pointers to `u64` when reading/writing them from RustBuffers.
* This is a first step to solving https://github.com/mozilla/uniffi-rs/issues/2567

Updated / fleshed out the object references documentation.


bendk:
I've tried addressing handles in many ways over the years.  I'm hoping that we can come to consensus on this fairly easily.  I don't think it makes sense to have different FFI types for these things, as outlined in the commit description.  The other option that could make sense would be to make these always pointer-sized by using `size_t` for callback interface handles.  However, that seems like more work for little gain.  Using a single integer width simplifies things and I think taking a small performance hit on 32-bit machines is totally fine.

We could create an ADR for this, but at this point it's not feeling like a real decision to me.  Please push back on this if you do think this choice is non-trivial.

I hope to build on this, probably eventually landing on a system like https://github.com/mozilla/uniffi-rs/pull/1823.  At this point, I think https://github.com/mozilla/uniffi-rs/pull/1808 is probably more changes than is worth it.

uniffi_macros/src/object.rs
```diff
@@ -67,28 +67,26 @@ pub fn expand_object(input: DeriveInput, options: DeriveOptions) -> syn::Result<
...
+            handle: ::std::primitive::u64,
             call_status: &mut ::uniffi::RustCallStatus
-        ) -> *const ::std::ffi::c_void {
-            ::uniffi::deps::trace!("clone: {} ({:?})", #name, ptr);
+        ) -> ::std::primitive::u64 {
+            ::uniffi::deps::trace!("clone: {} ({:?})", #name, handle);
             ::uniffi::rust_call(call_status, || {
-                unsafe { ::std::sync::Arc::increment_strong_count(ptr) };
-                ::std::result::Result::Ok(ptr)
+                unsafe { ::std::sync::Arc::increment_strong_count(handle as *const #ident) };
```
bendk: IIRC, we used to need to do this in two steps: `handle as *() as *const #ident`.  I'm not sure when this changed in Rust, but it doesn't seem necessary anymore.  It's also very possible that I'm misremembering this.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -514,7 +514,6 @@ impl KotlinCodeOracle {
             FfiType::Float32 => "Float".to_string(),
             FfiType::Float64 => "Double".to_string(),
             FfiType::Handle => "Long".to_string(),
```
bendk, line 523: I wasn't sure if we should use `Long` or `ULong` for these, but `Long` seems to work fine and it's a bit ergonomic in some cases.

bendk:
Last comment on this: does anyone know a good way to test this on a 32-bit arch?  It seems like this should be easy, but I couldn't figure out an easy way.

uniffi_macros/src/export/trait_interface.rs
```diff
@@ -51,17 +51,17 @@ pub(super) fn gen_trait_scaffolding(
...
-            ptr: *const ::std::ffi::c_void,
+            handle: ::std::primitive::u64,
             call_status: &mut ::uniffi::RustCallStatus
-        ) -> *const ::std::ffi::c_void {
-            ::uniffi::deps::trace!("clonining trait: {} ({:?})", #trait_name, ptr);
+        ) -> u64 {
+            ::uniffi::deps::trace!("clonining trait: {} ({:?})", #trait_name, handle);
             ::uniffi::rust_call(call_status, || {
-                let ptr = ptr as *mut ::std::sync::Arc<dyn #self_ident>;
+                let ptr = handle as *mut ::std::sync::Arc<dyn #self_ident>;
```
jplatte: I think we should use https://doc.rust-lang.org/std/ptr/fn.with_exposed_provenance.html to be more explicit.

uniffi_macros/src/export/trait_interface.rs
```diff
@@ -51,17 +51,17 @@ pub(super) fn gen_trait_scaffolding(
...
                 let arc: ::std::sync::Arc<_> = unsafe { ::std::clone::Clone::clone(&*ptr) };
                 ::std::result::Result::Ok(
                     ::std::boxed::Box::into_raw(
                         ::std::boxed::Box::new(arc),
-                    ) as *const ::std::ffi::c_void
+                    ) as u64
```
jplatte:
```suggestion
                    ).expose_provenance() as u64
```

Same as with the other cast. I wish there was a nicer way of converting between u32 / u64 as well. Maybe adding an extension trait would actually be helpful 🤔

jplatte (commented):
Int to pointer casts are the root of all evil 🙊

uniffi_macros/src/export/trait_interface.rs
bendk: I considered this, but I wasn't so sure.  The value is coming from across the FFI, which makes it very unclear to me what the provenance should be.  The function docs don't seem to specify how things should work in this case, do you have more details/links?

jplatte: I think the docs on `with_exposed_provenance` are very clear, what exactly are you unsure about? Also, this is entirely a no-op that just clarifies intent / gets us closer off of error-prone `as` conversions. The pointer-to-int cast has exactly the same semantics w.r.t. provenance.

bendk: > The exact provenance that gets picked is not specified. The compiler will do its best to pick the “right” provenance for you (whatever that may be), but currently we cannot provide any guarantees about which provenance the resulting pointer will have – and therefore there is no definite specification for which memory the resulting pointer may access.

Suppose I create 2 objects and pass them both across the FFI.  Then inside this clone function, we call `with_exposed_provenance`.  The pointers for the 2 objects have different provenances.  Which provenance will the compiler pick?  How could the compiler possibly know what's correct?  That's what's unclear to me.

jplatte: The docs unfortunately just say

> The exact provenance that gets picked is not specified. The compiler will do its best to pick the “right” provenance for you (whatever that may be), but currently we cannot provide any guarantees about which provenance the resulting pointer will have – and therefore there is no definite specification for which memory the resulting pointer may access.

But again, my change suggestion is just to provide clarity about working with exposed provenance here - which is the case regardless of whether you add the suggested functions or not.

Further reading in case you had missed the link at the end of that function's docs: https://doc.rust-lang.org/std/ptr/index.html#exposed-provenance

bendk: To be honest, I don't understand the vision outlined in that doc.  It lays out a goal of defining semantics for pointer casts, but no plan on how to achieve that goal.  It feels like switching to those APIs would be a negative.  I kind of understand how `as` works with pointer/int casts.  If `with_exposed_provenance` works differently, then I don't know what the difference is.  If it works exactly the same, then why define new functions?

I also don't feel like it adds clarity to say that we're working with exposed provenance semantics, if the docs don't explain what those semantics are.  Our code casts pointers to bit-patterns, treats the 2 as equivalent, and we shouldn't pretend otherwise, IMO.

I guess I have negative feelings about the exposed provenance initiative in general.  Maybe that's just me though.  I could switch to the APIs if I'm outnumbered on this one.  I would also be interested in moving away from any kind of pointer casts and towards real handles (https://github.com/mozilla/uniffi-rs/pull/1808).

jplatte: > If it works exactly the same, then why define new functions?

Because `as` casts can do a lot of things, so when types change your `as` cast may still compile, but suddenly do something different. That's at least the general idea, here we're unfortunately still doing another `as` cast, but it still shows intent w.r.t. provenance.

Also, there's not really an exposed provenance initiative. There's the strict provenance initiative, and strict provenance does define clear semantics, but I'm pretty sure we can't use it here, and that's kind of the point: We call out that we are _intentionally_ not using that (we can't), rather than just using as casts because that's the traditional solution.

bendk: Ahh, so the benefit of using the functions is to make it clear that aren't using strict provenance.  It's more of a negative claim then a positive one in some sense.  That feels better to me.

uniffi_macros/src/export/trait_interface.rs
```diff
@@ -167,27 +169,30 @@ pub(crate) fn ffi_converter(
             dyn #trait_ident: ::core::marker::Sync, ::core::marker::Send
         );
 
+        // We're going to be casting raw pointers to `u64` values to pass them across the FFI.
+        // Ensure that we're not on some 128-bit machine where this would overflow.
+        ::uniffi::deps::static_assertions::const_assert!(::std::mem::size_of::<*const ()>() <= 8);
```
jplatte, line 169: Nice 😄

uniffi_macros/src/export/trait_interface.rs
```diff
@@ -167,27 +169,30 @@ pub(crate) fn ffi_converter(
...
+
         unsafe #impl_spec {
-            type FfiType = *const ::std::os::raw::c_void;
+            type FfiType = ::std::primitive::u64;
+
 
             fn lower(obj: ::std::sync::Arc<Self>) -> Self::FfiType {
-                ::std::boxed::Box::into_raw(::std::boxed::Box::new(obj)) as *const ::std::os::raw::c_void
+                ::std::boxed::Box::into_raw(::std::boxed::Box::new(obj)) as ::std::primitive::u64
```
jplatte:
```suggestion
                ::std::boxed::Box::into_raw(::std::boxed::Box::new(obj)).expose_provenance() as ::std::primitive::u64
```

uniffi_macros/src/object.rs
```diff
@@ -67,28 +67,32 @@ pub fn expand_object(input: DeriveInput, options: DeriveOptions) -> syn::Result<
...
 
         #[doc(hidden)]
         #[unsafe(no_mangle)]
         pub unsafe extern "C" fn #free_fn_ident(
-            ptr: *const ::std::ffi::c_void,
+            handle: ::std::primitive::u64,
             call_status: &mut ::uniffi::RustCallStatus
         ) {
-            ::uniffi::deps::trace!("free: {} ({:?})", #name, ptr);
+            ::uniffi::deps::trace!("free: {} ({:?})", #name, handle);
```
jplatte: This is now going to be printed as decimal, probably want hex?

jplatte (commented):
Oh, looks like the exposed provenance API is too new. I think per policy we could probably bump MSRV to 1.84, but it seems a bit weird to bump for this. So maybe revert that stuff in a separate commit, and re-apply it at a later point in time?

jplatte (approved):
Was going to approve though.

bendk:
> Oh, looks like the exposed provenance API is too new. I think per policy we could probably bump MSRV to 1.84, but it seems a bit weird to bump for this. So maybe revert that stuff in a separate commit, and re-apply it at a later point in time?

This made me think we could consolidate the logic somewhere, which would make it easy to switch between the two.  Then I realized, we already had that: the `Handle` class.  I refactored the code to use this.  When we want to switch to the exposed provenance APIs, just need to modify the methods there.

This also made me think one more time around about the way we double-wrap our trait objects.  I changed it from `Box<Arc<dyn Trait>>` to `Arc<Arc<dyn Trait>>`.  Here's my thinking:

* `Box<Arc<dyn Trait>>` requires us to create a new box every time we clone a handle.  That feels pretty bad.
* `Arc<Arc<dyn Trait>>` requires us to increment a couple atomics every time we want to use the handle.  That feels bad, but not so bad.

The reason we need to mess with the atomics is there's so cast from `Arc<Arc<dyn Trait>>` to `Arc<dyn trait>` -- at least, I couldn't figure out a way to make it work.  In the absence of that cast, we end up doing something like this:
1. Increment the outer arc in the clone function.
2. Dereference `Arc<Arc<dyn Trait>>` to `&Arc<dyn Trait>` and clone the inner Arc.  This increments the inner counter and decrements the outer counter.
3. Use the inner arc for whatever we need

It kind of weird, it requires a couple extra atomic operations, but it seems better then an extra allocation.

PR #2574, opened 2025-06-19
RustFuture: Clarify Send bound and switch to RawWaker

bendk:
Updated the `unsafe impl `Send` for WrappedFuture` docstring. I believe the new docs describe the situation more accurately.

Moved back to using `RawWaker` because I want to drop the need for WrappedFuture to be `Send`. `WrappedFuture` will not always be `Send` when `wasm-unstable-single-threaded` is enabled. The current waker creation depends on `RustFuture` being `Send` which depends on `WrappedFuture` being `Send` (https://doc.rust-lang.org/std/task/struct.Waker.html#impl-From%3CArc%3CW%3E%3E-for-Waker).

As a side bonus, I think the tracing printouts will improve.

### DIFF
--- a/uniffi_core/src/ffi/rustfuture/future.rs
+++ b/uniffi_core/src/ffi/rustfuture/future.rs
@@ -78,11 +78,10 @@
 use std::{
     future::Future,
     marker::PhantomData,
-    ops::Deref,
     panic,
     pin::Pin,
     sync::{Arc, Mutex},
-    task::{Context, Poll, Wake},
+    task::{Context, Poll, RawWaker, RawWakerVTable, Waker},
 };
 
 use super::{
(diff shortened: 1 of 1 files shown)


uniffi_core/src/ffi/rustfuture/future.rs
```diff
@@ -180,10 +179,16 @@ where
...
-// Rust will not mark it Send by default when T::ReturnType is a raw pointer.  This is promising
-// that we will treat the raw pointer properly, for example by not returning it twice.
+// FIXME: This is required to make the futures code compile, although it's not entirely correct.  
+// If the `wasm-unstable-single-threaded` feature is enabled, then the wrapped future may not be `Send`, and thus `WrappedFuture` is not Send either. 
+// This seems okay for now, since if `wasm-unstable-single-threaded` then futures should stay on one thread.
+// However, we should probably fix this in the long-term.
+//
+// In addition, `WrappedFuture::result` can be a raw pointer which is `!Send`.
+// This `!Send` bound is because Rust can't uphold safety guarantees for raw pointers.
+// However, this code does in fact treat the pointers correctly, for example we only return them across the FFI once.
```
bendk, line 191: @jhugman Does this seem correct to you?

jhugman:
```
// If the `wasm-unstable-single-threaded` feature is enabled, then the wrapped future may not be `Send`, and thus `WrappedFuture` is not Send either.
// This seems okay for now, since if `wasm-unstable-single-threaded` then futures should stay on one thread.
```

This seems correct.

I'm not sure I can comment on the rest of the comment though.

mhammond (approved):

PR #2575, opened 2025-06-19
Fix rust-analyzer by removing dependency on `futures`

bendk:
This was causing rust-analyzer to crash because of https://github.com/rust-lang/rust-analyzer/issues/20038

This dependency was only needed for one test and I don't think that test was really needed.  The simplest solution is to just drop the dependency and test together.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -646,28 +646,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0845fa252299212f0389d64ba26f34fa32cfe41588355f21ed507c59a0f64541"
 
 [[package]]
-name = "futures"
-version = "0.3.29"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "da0290714b38af9b4a7b094b8a37086d1b4e61f2df9122c3cad2577669145335"
-dependencies = [
- "futures-channel",
- "futures-core",
- "futures-executor",
- "futures-io",
- "futures-sink",
- "futures-task",
- "futures-util",
-]
-
-[[package]]
 name = "futures-channel"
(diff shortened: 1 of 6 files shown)


bendk:
This made rust-analyzer work for me for a bit, but then it started failing again.  I'm not sure what's happening here, but this doesn't seem to be the right fix.

mhammond:
> This made rust-analyzer work for me for a bit, but then it started failing again

Damn. I wonder if we should just open a bug for rust-analyzer - one one hand they may find it too large of a repro to be actionable, but OTOH they might actually care and have a look given all they need to do is check out the repo.

ShoyuVanilla:
> > This made rust-analyzer work for me for a bit, but then it started failing again
> 
> Damn. I wonder if we should just open a bug for rust-analyzer - one one hand they may find it too large of a repro to be actionable, but OTOH they might actually care and have a look given all they need to do is check out the repo.

It would be great to have a minimal reproducible example, but even just knowing which file the panic occurs in would help me try to reproduce it.

mhammond:
> It would be great to have a minimal reproducible example, but even just knowing which file the panic occurs in would help me try to reproduce it.

I'm not sure how to determine that, but would be happy to with instructions. What I see is just:
```
thread 'PrimeCaches#3' has overflowed its stack
fatal runtime error: stack overflow
[Error - 2:43:27 AM] Server process exited with signal SIGABRT.
[Info  - 2:43:27 AM] Connection to server got closed. Server will restart.
true
```

I always see this just a few seconds after opening the root directory as a workspace in vscode.

ShoyuVanilla:
Oh, the cache priming triggers the panic. Thanks! I'll try looking into it

PR #2581, opened 2025-06-29
Fix a documentation output filename collision

Le-Maz:
I accidentally introduced a filename collision with my previous change. It turns out, `cargo doc` doesn't like having two binaries in one project with the same name, which caused it to raise an error when I tried to generate the documentation, conflicting `uniffi/uniffi-bindgen.rs` with `examples/app/uniffi-bindgen-cli/uniffi-bindgen.rs`. To fix it, I have marked the binary with `doc = false`.


Le-Maz:
I apologize, I included the merge commits. I'll close this pull request and open a clean one.

PR #2582, opened 2025-06-29
Fix a documantation output filename conflict

Le-Maz:
I accidentally introduced a filename collision with my previous change. It turns out, `cargo doc` doesn't like having two binaries in one project with the same name, which caused it to raise an error when I tried to generate the documentation, conflicting `uniffi/uniffi-bindgen.rs` with `examples/app/uniffi-bindgen-cli/uniffi-bindgen.rs`. To fix it, I have marked the binary with `doc = false`.


mhammond (approved):
Thanks! Longer term we should probably try and kill examples/app/uniffi-bindgen-cli/uniffi-bindgen.rs

PR #2586, opened 2025-07-02
Trait interface handle improvements (#2567)

bendk:
Use the lowest bit of trait interface handles to distinguish handles for Rust-implemented objects vs foreign implemented ones.  Use this when lifting/lowering trait interfaces to avoid the double-wrapping issue described in #2567.

* Added the `uniffi_foreign_handle` method to each trait interface exported with UniFFI.  For foreign-implemented traits, this returns the foreign handle.  For Rust-implement traits, it returns None.
* Added a `clone` field to VTables, we need this to return foreign handles from Rust efficiently.
* As long as we're changing the VTable layout, let's put `free` and `clone` at the start of the vtable rather than the end.  This is how C++ lays things out and it makes it possible to free a vtable handle without knowing the concrete type / full VTable layout.


uniffi_macros/src/export/trait_interface.rs
```diff
@@ -126,18 +127,37 @@ pub(crate) fn ffi_converter(
...
+                    use ::std::clone::Clone;
+                    // Note: handle is for a double-wrapped arc
+                    // https://mozilla.github.io/uniffi-rs/latest/internals/object_references.html
+                    let obj: ::std::sync::Arc<::std::sync::Arc<dyn #trait_ident>> = unsafe {
+                        handle.into_arc()
+                    };
+                    // We have a `Arc<Arc<dyn #trait_ident>>>`, but we need to return an `Arc<dyn #trait_ident>`
+                    // Unfortunately, Rust won't cast the double-arc to a single-arc trait object.
+                    // So we have to clone the inner arc and return that.
+                    ::std::result::Result::Ok((*obj).clone())
```
bendk, line 151: This feels pretty weird and I have a couple ideas on how to fix it:

* Switch to using "wide handles" for trait interfaces.  This would be a struct with 2 `u64` handles.  This way we can represent the trait interface directly without having to create the extra arc.
* Implement something like `https://github.com/mozilla/uniffi-rs/pull/1808` where we manage the handles ourselves. 

I think I prefer the latter, but it's not clear if we want to go that route and I didn't want to block the trait interface fixes on that decision, so this PR only focuses on the trait interface wrapping issue.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 ## [[UnreleasedUniFFIVersion]] (backend crates: [[UnreleasedBackendVersion]]) - (_[[ReleaseDate]]_)
 
+### ⚠️ Breaking Changes ⚠️
+- UDL-based trait interfaces must now be wrapped with the `#[uniffi::trait_interface]` attribute.
```
mhammond, line 10: I think this could be improved, I'll try and be more specific tomorrow :)

CHANGELOG.md
```diff
@@ -22,6 +27,15 @@
   - Lowering/lifting now uses `u64` values
   - The free function inputs a `u64` handle rather than a raw pointer
   - The clone function inputs and returns a `u64` handle rather than a raw pointer
+- Trait / Callback interface changes
+    - Added the `clone` function to the VTable
+    - Reordered VTable fields. `free` and `clone` now the first items and second fields, followed by
```
mhammond: I think you should kill the second point here, then roll it up with the 1 above.

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
```diff
@@ -9,6 +9,22 @@ fileprivate struct {{ trait_impl }} {
...
+                try {{ ffi_converter_name }}.handleMap.remove(handle: uniffiHandle)
+            } catch {
+                print("Uniffi callback interface {{ name }}: handle missing in uniffiFree")
+            }
+        },
+        uniffiClone: { (uniffiHandle: UInt64) -> UInt64 in
+            do {
+                return try {{ ffi_converter_name }}.handleMap.clone(handle: uniffiHandle)
+            } catch {
+                print("Uniffi callback interface {{ name }}: handle missing in uniffiClone")
```
mhammond: yeah - should these just be panics? prints seem unlikely to ever be seen in the wild.

uniffi_core/src/ffi/handle.rs
```diff
@@ -30,6 +30,14 @@ impl Handle {
         self.0 as *const T
     }
 
+    /// Was this handle generated by the foreign side of the FFI?
+    pub fn is_foreign(&self) -> bool {
+        // Foreign handles have the lowest bit set.
+        // We know that Rust handles will never have this set, since the pointers that they're casted from have alignment > 1;
```
mhammond: nano-nit - s/casted/cast/?

uniffi_macros/src/export/trait_interface.rs
mhammond: Yeah, it's already pretty wierd. I still don't quite get it tbh - we *could* use a box, but prefer the double arc as it avoids an allocation, right? Maybe the docs could discuss that tradeoff more? But it's already there, so doesn't block this.

CHANGELOG.md
mhammond: So I think I was mildly confused about what was being asked here. I think maybe something like "Trait interfaces described in UDL must have the Rust implementation of the trait wrapped with the `#[uniffi::trait_interface]` attribute." - optionally with a link to your new docs. It might also be useful to show what the error is if they do not, just for people to try to upgrade without checking the notes, then go looking afterwards?

(It does seem a shame to require this especially given the implementation is to unconditionally return None. I do 1/2 understand why though - I take it you are quite confident there's no alternative here?)

uniffi_macros/src/export/trait_interface.rs
```diff
@@ -204,3 +231,31 @@ pub(crate) fn ffi_converter(
         }
     }
 }
+
+pub fn alter_trait(item: &ItemTrait) -> TokenStream {
```
mhammond, line 235: It might be nice to have a terse comment explaining what it is doing and specifically what "trait" means in this context. It could also be `pub(crate)` right?

bendk:
> (It does seem a shame to require this especially given the implementation is to unconditionally return None. I do 1/2 understand why though - I take it you are quite confident there's no alternative here?)

This seems like the worst part and I've been thinking about alternatives too.  The one that I can think of is to rely on the relatively new trait upcasting feature + `Any`.  Here's a sketch of how that would look: https://github.com/mozilla/uniffi-rs/commit/5f350f2fffe67445147f70cbcc9b4ebc1ab826ea

The main drawback is we need to depend on Rust 1.86.0.  That's currently the "Uses" version for Firefox, so I guess we'd need to convince the build team that we should make that the MSRV sometime before we release `0.30.X` and FF switches to it.

The other difference is that the trait bounds change.  That seems neutral to me, maybe slightly positive.

bendk:
This is a side issue, but if we went with the idea I proposed in my last comment, we could define both `uniffi::TraitInterface` and `uniffi::TraitInterfaceWithForeign` and let the supertrait determine if we add the foreign support or not instead of the `with_foreign` tag:

```
#[uniffi::export]
pub trait MyTrait: uniffi::TraitInterface {
   ...
}

#[uniffi::export] // no (with_foreign) needed anymore
pub trait MyOtherTrait: uniffi::TraitInterfaceWithForeign {
   ...
}
```

mhammond:
I haven't thought much about this, but your initial proposal seems more consistent with what we do already today and is only required for traits in UDL, with no changes for traits in proc-macros, right? The second proposal would be a requirement everywhere and the first time we'd be insisting types implement one of our traits?

bendk:
> I haven't thought much about this, but your initial proposal seems more consistent with what we do already today and is only required for traits in UDL, with no changes for traits in proc-macros, right? 

Correct.

> The second proposal would be a requirement everywhere and the first time we'd be insisting types implement one of our traits?

Also yes, but there's a blanket impl for any type that's `Send + Sync + 'static`.  So, from a user's POV it just means replacing `Send + Sync`  with `uniffi::TraitInterface` as the super-traits.


Neither of these seems that bad to me, I'd be happy either way.

bendk:
I talked some more with Mark and we both thought we should land this as-is without trying to add https://github.com/mozilla/uniffi-rs/commit/5f350f2fffe67445147f70cbcc9b4ebc1ab826ea.  That may or may not be a good idea, but it feels more committal since it requires that all users switch to using our super-trait.  The current status of this just requires UDL users to add a proc-macro attribute, which seems like less of a commitment.

PR #2587, opened 2025-07-02
Use IndexSet for FFI definitions

bendk:
This simplifies the dupe removal and I think it'll make future work easier.  For example, this would make  it easier to factor out common FFI definitions into a shared module.


mhammond (approved):
heh - I've actually got a patch stacked up with *removes* deriving of `Eq` etc by `Enum` and `Record` (and replaces the maps they are stored in with arrays, just like objects are), but I don't see how this would impact that.

PR #2588, opened 2025-07-07
Enum and Record no longer derive `Eq`

mhammond:
This is so they can hold `uniffi_trait` objects, which cannot reasonably implement it. We now store `Enum` and `Record` objects in a `Vec<>` like we already do for `Object`s.


bendk (approved):
Makes sense to me.

PR #2589, opened 2025-07-07
Flatten `UniffiTrait` objects into individual methods in the pipeline

mhammond:
It's becoming apparent that the `Vec<UniffiTrait>` is a pita and this seems in the spirit of pipelines.

### DIFF
--- a/uniffi_bindgen/src/pipeline/general/nodes.rs
+++ b/uniffi_bindgen/src/pipeline/general/nodes.rs
@@ -265,6 +265,7 @@ pub struct Interface {
     pub constructors: Vec<Constructor>,
     pub methods: Vec<Method>,
     pub uniffi_traits: Vec<UniffiTrait>,
+    pub uniffi_trait_methods: UniffiTraitMethods,
     pub trait_impls: Vec<ObjectTraitImpl>,
     pub imp: ObjectImpl,
     pub self_type: TypeNode,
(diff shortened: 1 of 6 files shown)


uniffi_bindgen/src/pipeline/general/nodes.rs
```diff
@@ -590,6 +591,34 @@ pub struct Checksum {
...
+pub struct UniffiTraitMethods {
+    pub debug_fmt: Option<Method>,
+    pub display_fmt: Option<Method>,
+    pub eq_eq: Option<Method>,
+    pub eq_ne: Option<Method>,
+    pub hash_hash: Option<Method>,
+    pub ord_cmp: Option<Method>,
+}
+
+impl UniffiTraitMethods {
```
mhammond, line 605: is there a canonical thing here I'm not doing?

bendk: Maybe the `fill_from` method could be moved to `objects::pass`.  There's something that's nice about having the code in the pass itself and I doubt it will need to be shared.  I don't feel that strongly either way though, however you want to do it is food with me.

mhammond:
> My only other comment is that it's not totally clear to me why the old way was a PITA. Can you point to how this is going to make things easier?

Conceptually, I don't think any consumer actually *wants* to ever iterate over the full set, because each item needs to be handled separately and differently. Eg, knowing whether an object handles "Ord::cmp" shouldn't involve looking at everything - concretely, [this kotlin code](https://github.com/mozilla/uniffi-rs/blob/e33dedd7249b6dc3acdcdad6bd3366a71e8a607f/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt#L116-L120) - it needs to iterate over every trait when it only cares about 1. That loop changes to the much saner [check for just the trait it cares about](https://github.com/mozilla/uniffi-rs/blob/e69f379d845ce9ba55d412f6592020bd518d1577/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt#L117-L119).

[edit: fixed links!]

uniffi_bindgen/src/pipeline/general/nodes.rs
mhammond: > and I doubt it will need to be shared.

That's the cool part - it probably will - by `Enum` and `Record` nodes!

mhammond:
I meant to elaborate on "in the spirit of pipelines" - what I liked about this was that for pipeline bindings, they lost the whole `UniffiTrait` (`Vec<>` or otherwise) entirely. That existed due to metadata considerations, but by the time it got to codegen, that abstration was removed entirely and replaced with something more appropriate. The outcome for Python seemed really satisfying.

PR #2590, opened 2025-07-07
Methods now carry more info about their `self`, towards methods on en…

mhammond:
…ums/records.

In the templates, for `Callable`, `x.takes_self()` is replaced with `x.self_type().is_some()` etc.


uniffi_meta/src/lib.rs
```diff
@@ -182,10 +182,72 @@ impl ConstructorMetadata {
     }
 }
 
+// A "sub-type" which represents all the "self" receivers we support.
+// It would ideally be `Type`
```
bendk, line 186: I agree, making `Method::receiver` be `Type` seems nicer to me, why can't we do that?  The only downside that I can see is that the code would need to handle `receiver` being an invalid type, but worst-case we can just panic.

uniffi_meta/src/lib.rs
```diff
@@ -182,10 +182,72 @@ impl ConstructorMetadata {
...
+    pub fn name(&self) -> &str {
+        match self {
+            Self::Enum { name, .. } => name,
+            Self::Record { name, .. } => name,
+            Self::Object { name, .. } => name,
+        }
+    }
+
+    // a bit smelly, but it is what it is.
+    pub fn update_name(&mut self, new_name: String) {
```
bendk: Is this only used by `uniffi_udl`?  How about making this a function in that crate?

bendk (approved):
The code changes look correct and the overall plan makes sense.  My only question is if we can re-use `Type` instead of adding the `MethodReceiver` struct.  Whatever you decide is good with me there.

uniffi_meta/src/lib.rs
mhammond: There are 2 reasons this is tricky:
* We can't reasonably support an "unnamed" type. I almost thought about `Type` having a sub-enum to handle named types, but
* The other problem is that for `Object` we don't know the `ObjectImpl`, and I'm not sure how we could get it.

PR #2595, opened 2025-07-15
Metadata for Enums and Records carries uniffi-trait info.

mhammond:
This is towards Enums and Records exposing `Debug`, `Eq` etc. None of the bindings support it yet, but the tests show how this is setup for the scaffolding.

(I do have all the bindings supporting it in [here](https://github.com/mhammond/uniffi-rs/tree/all-uniffi-traits) though :)

### DIFF
--- a/uniffi_bindgen/src/pipeline/general/ffi_functions.rs
+++ b/uniffi_bindgen/src/pipeline/general/ffi_functions.rs
@@ -8,15 +8,14 @@ use super::*;
 
 pub fn pass(namespace: &mut Namespace) -> Result<()> {
     let crate_name = namespace.crate_name.clone();
-    let namespace_name = namespace.name.clone();
     let mut ffi_definitions = vec![];
 
     namespace.visit_mut(|callable: &mut Callable| {
         let name = &callable.name;
         let ffi_func_name = match &callable.kind {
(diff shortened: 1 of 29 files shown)


mhammond:
I think this is ready to go - it doesn't include the bindings yet.

uniffi_bindgen/src/pipeline/general/ffi_functions.rs
```diff
@@ -27,13 +26,9 @@ pub fn pass(namespace: &mut Namespace) -> Result<()> {
...
         let receiver_argument = match &callable.kind {
-            CallableKind::Method { interface_name } => Some(FfiArgument {
-                name: "uniffi_ptr".to_string(),
-                ty: FfiType::Handle(HandleKind::Interface {
-                    namespace: namespace_name.clone(),
-                    interface_name: interface_name.clone(),
-                })
-                .into(),
+            CallableKind::Method { self_type } => Some(FfiArgument {
+                name: "self".to_string(),
```
bendk: Can we name this arg `uniffi_self`?  That way we avoid collisions with keywords and user arg names.

bendk (approved):
Looks great to me, I just had one argument name suggestion.

Issue #2584, opened 2025-06-30
Syntax for allowing standard rust traits for Records?


mhammond:
via #2555, it seems reasonable to allow:

```
#[derive(uniffi::Record)]
#[uniffi::export(Display)]
pub struct MyRecord {
    s: String,
    i: i8,
}

impl std::fmt::Display for MyRecord { ... }
```

And we'd generate foreign code to copy the record over the FFI and call the function. Problem is we currently support:
```
#[derive(uniffi::Object)]
#[uniffi::export(Display)]
pub struct MyRecord { ... }
```
and our [support for these traits](https://github.com/mozilla/uniffi-rs/blob/0116e54c5eaa0a52e67878c97a176db646b007e7/uniffi_macros/src/export.rs#L152) doesn't know if it is dealing with a `Record` or an `Object`, so can't get the args correct.

(Enums will not have this problem, `#[uniffi::export(Display)]` on an `enum` block isn't ambiguous)

Anything I'm missing here or general thoughts? cc @jplatte @bendk

bendk:
I like the syntax.  What would be the issue not knowing if it's a record or object?  I took a quick look at the code and it seemed like it would work with either one to me.

mhammond:
The problem is that we need to know whether the FFI function is taking an `Arc<>` or not (ie, whether the FFI function is a "function" or a "method") - in particular, [this code](https://github.com/mozilla/uniffi-rs/blob/0116e54c5eaa0a52e67878c97a176db646b007e7/uniffi_macros/src/export/utrait.rs#L165-L170) isn't correct for data classes. I suspect we will want to call `FnSignature::new_function` in this scenario.

mhammond:
Via #2607 we worked out we don't actually need this.

PR #2596, opened 2025-07-15
Record trait macros

mhammond:
On top of #2595, so look at the last commit. Demonstrates #2584.
tl;dr -  `#[uniffi::export]` on a `struct` might refer to either the object or record. It's a shame we need to know, but fixing that is tricky (in particular, knowing whether `Arc` or `RustBuffer` is the very part). This adds a new macro, so a record exposing builtin traits looks like:

```
#[derive(uniffi::Record, Debug)]
#[uniffi::export_for_record(Debug)]
pub struct TraitRecord { .. }
```

Kinda related, this also opens the door to general methods on data-classes. So later, we might find:

```
#[uniffi::export]
impl Foo { ... }
```

and we are likely to also need to know more about `Foo` - now it must be object. In that case, the `export_for_record` above ends up growing into many things, which seems very ugly, but I can't think of a better story.

@bendk @jplatte any thoughts?


jplatte:
I forget why we have `#[uniffi::export]` on `struct`s for trait impls in the first place, as opposed to `#[uniffi(export())]` (a derive helper attribute). Maybe because it looks nicer? If we changed to a helper attribute, we wouldn't need `export_for_record` _for traits_.. but still would need it for inherent `impl`s so maybe it's not worth the churn?

uniffi_macros/src/export/scaffolding.rs
```diff
@@ -147,22 +147,30 @@ impl ScaffoldingBits {
...
+                    }
+                } else {
+                    quote! { #try_lift(uniffi_self_lowered) }
                 }
             }
-        } else {
-            quote! { #try_lift(uniffi_self_lowered) }
+            MethodReceiverKind::Enum | MethodReceiverKind::Record => {
+                assert!(!is_trait);
+                quote! { #try_lift(uniffi_self_lowered) }
```
bendk: Is this the only time where you need to handle records/enums and objects differently?

If so, maybe we don't need 2 different attributes because [2586](https://github.com/mozilla/uniffi-rs/pull/2586/files#diff-bff922c65ef3007e6164f74d62c32454769b177527c7a53b29dae9075edfbafa) is going  should make it so we only need 1 code path here.

bendk:
I only skimmed the code.  My takeaways are: this is a nice feature and I like how it makes things more general (e.g. replacing `interface_name` with `self_type`).

I'd rather we only have `#[export]` as an attribute, but I'm not sure if that's possible or not.  I think we can make probably make this feature work by leaning on the ffi converter traits and/or cleaning up some things like 2586.  However, even if that's true I'm not sure if it'll be true for all features.

I guess I learn towards not adding `#[export_for_record]` and `#[export_for_enum]`, but it wouldn't take much to convince me otherwise.

uniffi_macros/src/export/scaffolding.rs
mhammond: We still have the few lines above where we need to know whether we are receiving an `Arc<>` or not. I've rebased it.

mhammond:
> I'd rather we only have `#[export]` as an attribute, but I'm not sure if that's possible or not. 

> I think we can make probably make this feature work

The feature actually works great. Enums work perfectly as there is no ambiguity. So long as the scaffolding macros know whether we want an `Arc<>` or the raw type it all comes together perfectly.

The problem is that the macros don't know which to use when they see

```
#[uniffi::export(Debug)]
pub struct Foo { .. }
```
the generated ffi functions need to know whether this is going to be a record or an object.

> I guess I learn towards not adding `#[export_for_record]` and `#[export_for_enum]`, 

As above, `Enum` is fine, because 

```
#[uniffi::export(Debug)]
pub enum Foo { .. }
```
is not ambigious - the generated ffi functions know for sure it's a dataclass.

Relatedly, @jplatte :

>  If we changed to a helper attribute, we wouldn't need `export_for_record` _for traits_..

I'm not that familiar with helper attributes - can you please elaborate here?

jplatte:
So, derive macros (not attribute or function-like macros) allow you to register an attribute name that you want your macro to parse (we already make use of this [here](https://github.com/mozilla/uniffi-rs/blob/a94c2c5d6dc87d6e735d1783cd646bfafd2621fb/uniffi_macros/src/lib.rs#L117), for `#[uniffi(default = ...)]`). This is required for derive macros to parse custom attributes because they can only add to the input, not rewrite it, so they can't remove attributes they have parsed from the input (and you'd get an unknown-attribute error if the attribute wasn't registered).

The difference would be that it would no longer be a separate proc-macro's job to export the trait impls, it would be the job of `derive_object` or `derive_record`, so we'd have the context of which one the user used.

For more detail, I wrote a tutorial-style article about how to use helper attributes (and how to do do error handling) in proc-macros a few years ago: https://blog.turbo.fish/proc-macro-error-handling/

mhammond:
> The difference would be that it would no longer be a separate proc-macro's job to export the trait impls, it would be the job of `derive_object` or `derive_record`, so we'd have the context of which one the user used.

That's very helpful, thanks, so I see what you meant about "only traits" - so let me repeat the problem AIUI:

* If we used a helper attribute to export items from the struct impl, we could use identical syntax for objects or records as the derive macros become responsible for the export and they already know.
* However, if we consider the general case of `impl Something { ... } `, there's no derive macro, so we are stuck with the same problem - the macro needs to know whether we have a datatype or object.

It needs to know when it generates the ffi function. Maybe we can fix that?

Today we generate something like this for `Ord::cmp`, where we take `other`:

```
pub extern "C" fn uniffi_uniffi_trait_methods_fn_method_proctraitmethods_uniffi_trait_ord_cmp(
  uniffi_self_lowered: < ::std::sync::Arc<ProcTraitMethods>as ::uniffi::Lift<crate::UniFfiTag>> ::FfiType,
  other: < <ProcTraitMethods as ::uniffi::LiftRef<crate::UniFfiTag>> ::LiftType as ::uniffi::Lift<crate::UniFfiTag>> ::FfiType,
  call_status: &mut ::uniffi::RustCallStatus,) ->  <i8 as ::uniffi::LowerReturn<crate::UniFfiTag>> ::ReturnType {
...
```

Note the `Arc` around self - this is what I need to remove for dataclasses - it's added [here](https://github.com/mozilla/uniffi-rs/blob/a94c2c5d6dc87d6e735d1783cd646bfafd2621fb/uniffi_macros/src/export/scaffolding.rs#L137-L140). If we can remove this the macros would probably be fine.

Writing this out, note the lack of an `Arc` and the `LiftRef` for `other` - can we use this for `self` somehow? Is it the `Arc<dyn Foo>>` case the reason maybe?

bendk:
> Writing this out, note the lack of an `Arc` and the `LiftRef` for `other` - can we use this for `self` somehow? Is it the `Arc<dyn Foo>>` case the reason maybe?

This seems like the right way to me.  I messed around was able to get the `trait-methods` fixture working with [this code](https://github.com/mozilla/uniffi-rs/commit/a4ec93749ec8b62d415186028566975fb8aeca61).  I'm not sure if there are issues elsewhere, but it seems like it's the right start.

mhammond:
> This seems like the right way to me. I messed around was able to get the `trait-methods` fixture working with [this code](https://github.com/mozilla/uniffi-rs/commit/a4ec93749ec8b62d415186028566975fb8aeca61).

That works perfectly and allows us to side-step this at the cost of a little more complexity in the bindings, which seems the correct tradeoff - thanks!

PR #2599, opened 2025-07-22
Allow default values in an enumeration, improve default value docs.

### LINKED ISSUE #2551: Better document our support for numeric literals (and literals/defaults in general?)
Via #2543, our docs for literals in proc-macros aren't great - "String, integer, float, and boolean literals".

Also, via UDL, numeric literals carry a radix etc - so we've always supported rendering literals as hex or decimal etc in the bindings. Not sure how much of this is supported in proc-macros, but we should find out.

Ultimately it's probably better still for an entire page dedicated to defaults and literals, but small steps to get started.

### LINKED ISSUE #2553: Allow to set default values for fields in an enumeration
Hi, trying to set the default values for some field in an enum, I noticed that it's not currently possible. Would be awesome to have the support for it.

Example:

```rust
#[derive(uniffi::Enum)]
pub enum MyEnum {
    Variant {
        foo: u8,
        #[uniffi(default = None)]
        bar: Option<i32>,
    },
}
```

mhammond:
Fixes #2553, fixes #2551

### DIFF
--- a/uniffi_macros/src/lib.rs
+++ b/uniffi_macros/src/lib.rs
@@ -121,7 +121,7 @@ pub fn derive_record(input: TokenStream) -> TokenStream {
         .into()
 }
 
-#[proc_macro_derive(Enum)]
+#[proc_macro_derive(Enum, attributes(uniffi))]
 pub fn derive_enum(input: TokenStream) -> TokenStream {
     expand_enum(parse_macro_input!(input), DeriveOptions::default())
         .unwrap_or_else(syn::Error::into_compile_error)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,15 @@
 ### What's new?
 
 - Objects can export the `Ord` trait, allowing such objects to be ordered.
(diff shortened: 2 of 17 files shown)


uniffi_macros/src/lib.rs
```diff
@@ -121,7 +121,7 @@ pub fn derive_record(input: TokenStream) -> TokenStream {
         .into()
 }
 
-#[proc_macro_derive(Enum)]
+#[proc_macro_derive(Enum, attributes(uniffi))]
```
gruberb, line 124: Would be nice to add a comment for future devs why this is needed: So that the `default()` on the enum fields is getting passed down to the AST, if I understand correctly?

mhammond: yeah, exactly - so that `[uniffi(...)]` attributes with in that context - `[uniffi(default)]` is the only thing actually used here but others could also work if they make sense. We now only have 1 `#[proc_macro_derive()]` macro which does *not* support `#uniffi` attributes.

Issue #2600, opened 2025-07-22
Use after free (resp. too early drop of Struct containing __m256i)


threema-theo:
Hi,

While using Rust Crypto libraries exposed with UniFFI in an Android App, I've discovered that _some_ structs get dropped too early. 

# Reproducing

I've created a minimal working example in https://github.com/threema-theo/uniffi-drop to allow easier reproduction.

In a nutshell:
1. Write a library of the form
   ```rust
   uniffi::setup_scaffolding!();
   
   use std::arch::x86_64::{__m128i, __m256i, _mm_set1_epi8, _mm256_set1_epi8};
   
   #[derive(uniffi::Object)]
   pub struct MyStruct256(__m256i);
   
   /// This is the problematic struct:
   /// it gets dropped before its end of life...
   #[uniffi::export]
   impl MyStruct256 {
       pub fn debug(&self) {
           dbg!(&self.0);
       }
   
       #[uniffi::constructor]
       pub fn new() -> Self {
           dbg!("New 256");
           Self(unsafe { _mm256_set1_epi8(1) })
       }
   }
   
   impl Drop for MyStruct256 {
       fn drop(&mut self) {
           dbg!("I'm just dropping 256!");
           // Simulate zeroizing?
           self.0 = unsafe { _mm256_set1_epi8(0) }
       }
   }
   ```
   
2. Call the code from an Android Unit Test:
   ```kt
     val struct256 = MyStruct256()
     struct256.debug()
     // but here the problem arises! Why is struct256 dropped?
     struct256.debug()
     struct256.debug()
    ```
3. Run the code (still on the x86_64 computer) and observe the following output:
   ```txt
   [src/lib.rs:18:9] "New 256" = "New 256"
   [src/lib.rs:13:9] &self.0 = __m256i(
       72340172838076673,
       72340172838076673,
       72340172838076673,
       72340172838076673,
   )
   [src/lib.rs:25:9] "I'm just dropping 256!" = "I'm just dropping 256!"   # <--- THIS SHOULD NOT HAPPEN HERE!
   [src/lib.rs:13:9] &self.0 = __m256i(
       0,
       0,
       0,
       0,
   )
   [src/lib.rs:13:9] &self.0 = __m256i(
       0,
       0,
       0,
       0,
   )

   ```     

# Further triage
- These seems to apply especially to `__m256i`, I cannot reproduce it with `__m128i`.
[...]

PR #2602, opened 2025-07-23
Cast object pointers to the correct type in `clone`/`free`

bendk:
Before we were calling `Arc::increment_strong_count` with a `*const ::std::ffi::c_void`, rather than the correct type.  I'm not sure why but this sometimes causes objects to be dropped early (#2600).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ## [[UnreleasedUniFFIVersion]] (backend crates: [[UnreleasedBackendVersion]]) - (_[[ReleaseDate]]_)
 
+- Fixed a bug where objects could be freed to early (https://github.com/mozilla/uniffi-rs/issues/2600)
+
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.29.3...HEAD).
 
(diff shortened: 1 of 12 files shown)


bendk:
This bug causes objects to potentially be dropped early, which seems quite bad.  I'm thinking this merits a `0.29.4` release to fix -- especially since the fix is pretty simple.

mhammond (approved):
Agree re the new release too.

mhammond:
(but do we want a changelog here?)

bendk:
> (but do we want a changelog here?)

Yes good point.  I just added a bullet.

badboy:
Can we maybe add a regression test?
The bug report looked like a simple enough case (though `x86_64` only).
I won't block landing the fix though and I might find some time tomorrow to put the test case into a regression test if you want.

bendk:
> Can we maybe add a regression test? The bug report looked like a simple enough case (though `x86_64` only). I won't block landing the fix though and I might find some time tomorrow to put the test case into a regression test if you want.

This makes sense.  I added a fixture for this.  It's getting hard to find names for these fixtures, so I just named it after the GH issue.  It's x86-64 only, I didn't repro with the other types I tried (`[u64; 4]` and `i256::I256`).

PR #2603, opened 2025-07-24
Support default value for custom type

flisky:
cc @mhammond


fixtures/proc-macro/src/lib.rs
```diff
@@ -379,10 +383,14 @@ pub struct RecordWithDefaults {
     float_var: f64,
     #[uniffi(default=[])]
     vec: Vec<bool>,
+    #[uniffi(default = "Uncertain")]
```
jplatte: I think it makes a lot more sense for this to be

```suggestion
    #[uniffi(default = MaybeBool::Uncertain)]
```

mhammond: The other challenge I see if supporting non-trivial enums, such as tuple enums or those with fields - eg, seeing this grow to support `SomeEnum::SomeVariant { a: 0 }` seems tricky, but not having a story for these enums seems to reduce the utility.

Is the intent here to generally support a single default value when a particular enum is used? IOW, I'm wondering if a better first step might be to just support `#[uniffi(default)]` and for us to grow a way to better be able to reflect what the default value is for enums (which I *could* see learning how to support fields relatively easily and natively etc)

flisky: I thought `#[unffi(default)]` is ready for enum, but it turns out not.

I'm on vacation next week, and have little time on this, so I split the PR and only keep the changes for custom type.

fixtures/proc-macro/src/lib.rs
```diff
@@ -383,6 +387,8 @@ pub struct RecordWithDefaults {
     opt_vec: Option<Vec<bool>>,
     #[uniffi(default = Some(42))]
     opt_integer: Option<i32>,
+    #[uniffi(default = 42)]
```
mhammond, line 391: This is great, thanks! Unfortunately this doesn't correctly support `#[uniffi(default)]` for Python - could you please add a test case for this scenario (along with the test case you added for an explicit value) and look into supporting it for Python? Not supporting it for Python isn't necessarily a show-stopper, but it would make the docs for default values a little unfortunate - and speaking of which, it would also be great to get some docs for this in the manual (the docs could be fairly terse though - I've a PR up which improves the docs at https://github.com/mozilla/uniffi-rs/pull/2599 and I'll incorporate your docs into that). A CHANGELOG entry would also be great, where you can link to this PR and get all the glory :)

Thanks!

flisky: Updated. Please take a look.

PR #2607, opened 2025-07-24
Kill `uniffi_meta::MethodReceiver` etc so `self_type` becomes a bindi…

mhammond:
…ngs-only problem.

This came out of the discussion in #2596, where Ben showed how we can avoid the metadata knowing anything about the self type.

The CI gets a little uglier - you can no longer directly `From<>` a `MethodMetadata` into a `Method`, as you need to supply a type, but it successfully decouples methods from objects.


bendk (approved):
Looks right to me and the code deletions are nice.

PR #2609, opened 2025-07-27
Improve Kotlin documentation by adding more explicit instructions for interop

bes:
I spent a lot of time trying to figure out how to get the interop working. Adding more explicit instructions may save others a lot of blood, sweat, and tears.

### DIFF
--- a/docs/manual/src/kotlin/gradle.md
+++ b/docs/manual/src/kotlin/gradle.md
@@ -69,7 +69,9 @@ It's highly recommend to call `AttachCurrentThread` when spawning a rust thread,
 static VM: once_cell::sync::OnceCell<jni::JavaVM> = once_cell::sync::OnceCell::new();
 
 // need call this function in java/kotlin first
-#[export_name = "Java_com_xxx_xxx"]
+// Before Rust 1.85 use #[export_name = "Java_com_xxx_xxx"]
+// Since Rust 1.85 use #[unsafe(export_name = "Java_com_xxx_xxx")]
+#[unsafe(export_name = "Java_com_xxx_xxx")]
 pub extern "system" fn java_init(
     env: jni::JNIEnv,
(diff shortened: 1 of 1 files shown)


docs/manual/src/kotlin/gradle.md
```diff
@@ -86,6 +88,36 @@ tokio::runtime::Builder::new_multi_thread().on_thread_start(|| {
...
+```kotlin
+// Kotlin
+package my.package
+
+class MainActivity : ComponentActivity() {
+    external fun javaInit()
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        // Loads "libmy_library.so" so that it can be called by "external fun" above
+        System.loadLibrary("my_library")
```
bes, line 115: It might be better to do an init block in a companion ("static init") instead?

bendk: I'm not sure.  It seems like this way gives you a bit more control over when the initialization happens, but I'm not sure if that's useful or not.  Whichever way you think is best seems good to me.

bendk (approved):
This seems like an improvement today, although this whole situation seems unfortunate to me.  I haven't looked into this at all, but it seems like UniFFI should wrap some of this into a more convenient interface.   I'm hoping to explore using JNI instead of JNA later this year, maybe that's a good time.

In the meantime, improved docs are good.  What's you're opinion on static init vs `onCreate`?

bes:
For me the current way works fine, but I'm not a Kotlin expert at this point. Let's keep it as-is; if someone has a good reason to change it, then we can change it!

docs/manual/src/kotlin/gradle.md
bes: Keep

bendk:
Sounds good to me, let's merge this one.

PR #2614, opened 2025-08-04
Rename `Module` to `Namespace` in the pipelines.

mhammond:
As Ben and I discussed, it's a large patch but an improvement worth making before we have any more consumers

### DIFF
--- a/uniffi_bindgen/src/pipeline/initial/from_uniffi_meta.rs
+++ b/uniffi_bindgen/src/pipeline/initial/from_uniffi_meta.rs
@@ -18,16 +18,14 @@ use uniffi_pipeline::Node;
 /// * Call [Self::try_into_initial_ir] to construct a `initial::Root` node.
 #[derive(Default)]
 pub struct UniffiMetaConverter {
-    // Map module_path to `Module::name`
-    //
-    // This is needed because UDL and proc-macro code can generate different module paths.  To
-    // normalize things, we find the `NamespaceMetadata::name` that matches the module path.
-    //
-    // Use BTreeMap for each of these so that things stay consistent, regardless of how the
-    // metadata is ordered.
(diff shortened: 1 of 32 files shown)


uniffi_bindgen/src/pipeline/initial/from_uniffi_meta.rs
```diff
@@ -18,14 +18,10 @@ use uniffi_pipeline::Node;
...
-    //
-    // This is needed because UDL and proc-macro code can generate different module paths.  To
-    // normalize things, we find the `NamespaceMetadata::name` that matches the module path.
-    //
-    // Use BTreeMap for each of these so that things stay consistent, regardless of how the
-    // metadata is ordered.
+    // Use BTreeMap for each of these so that things stay consistent, regardless of how the metadata is ordered.
+    // All metadata has a `module_path`, which isn't the namespace/module name - this map converts them.
     module_path_map: BTreeMap<String, String>,
+    // Everyth8ng below here keyed by namespace/module name, not crate_name/module_path.
```
bendk:
```suggestion
    // Everything below here keyed by namespace/module name, not crate_name/module_path.
```

bendk: Should "module name" be `Module.name`?  "module name" could be confused with `module_path`.

I think we might want to rename the `Module` type at some point.  If we want to support nested modules, then maybe this struct should be called `Package` or `Namespace`.  I'm not sure if you want to tackle that in this PR, just throwing it out there.

bendk (approved):
Looks great to me.  The doc changes are nice and "Namespace" seems like a better term as we discussed.  It's what the existing docs usually use and it's more abstract and less tied to any particular language's terminology.  I just skimmed the code changes, I assume it was all mechanical.

Issue #2430, opened 2025-02-05
Can't declare an object implements a trait which is defined externally.


mhammond:
If you try and do, eg:

```
#[uniffi::export]
impl SomeTrait for SomeObject { ... }
```

It breaks if `SomeTrait` is declared in a different crate from `SomeObject` (ie, if `SomeTrait` is "external")

Demonstrated in #2423

PR #2615, opened 2025-08-04
Remove ObjectTraitImpl::tr_module_name as it is never Some()

mhammond:
It tricks you into thinking it is a path to #2430, but it's not. I've a fix for that on top of this which helps keep things clearer.


bendk (approved):
Nice cleanup, thanks!

PR #2616, opened 2025-08-07
Support interfaces deriving from external traits (fixes #2430)

### LINKED ISSUE #2430: Can't declare an object implements a trait which is defined externally.

mhammond:


uniffi_bindgen/src/pipeline/general/missing_types.rs
```diff
@@ -0,0 +1,49 @@
...
+use std::collections::HashMap;
+
+pub fn pass(root: &mut Root) -> Result<()> {
+    // ObjectTraitImpl carries only a trait name, we turn the trait into a real Type.
+    // find out which module defines which traits.
+    let mut known_traits: HashMap<String, String> = Default::default();
+    root.visit(|m: &Module| {
+        m.visit(|int: &Interface| {
+            if matches!(int.imp, ObjectImpl::Trait | ObjectImpl::CallbackTrait) {
+                known_traits.insert(int.name.clone(), m.name.to_owned());
```
bendk: It seems like this would fail if 2 modules defined traits with the same name.

I think that's okay for now, since we can't really do anything better.  But what do you think about eventually adding a `trait_module` field in the metadata?  I guess this would do what `tr_module_name` was trying to do before.

bendk (approved):
This seems good to me.  I had one concern about the module name, but I'm fine with merging this now and figuring that out later.

uniffi_bindgen/src/pipeline/general/missing_types.rs
mhammond: > It seems like this would fail if 2 modules defined traits with the same name.

yeah, I didn't think about that! So I made it fallible and made the restriction a little more obvious. It made me think too that it should be focused more on this specific issue, so renamed 'missing_types` to `trait_types`.

> But what do you think about eventually adding a trait_module field in the metadata? I guess this would do what tr_module_name was trying to do before.

Yeah, but I don't see a path to capturing that in the macros - certainly not the name, hence killing `tr_module_name`. I tried to use `TYPE_ID_META` but things went sideways due to the trait and I failed to work that out. It would be great if we could capture more metadata though and this could go away.

PR #2618, opened 2025-08-11
Support multiple crates exporting traits with the same name

bendk:
Use `TYPE_ID_META` for this, which contains both the trait and module name.  This meant implement `TYPE_ID_META` for `dyn Trait` so that we can use the same syntax to look it up for both callback and trait interfaces.

I only got this working using the pipeline, which is generally better at tracking external types.  `ComponentInterface` stores a map of object names to external types, so it would need a refactor to get this working.  I'd rather just not support it for those languages for now and work to move them over to the pipeline system.  Note that on Kotlin it will sometimes work and sometimes not, it depends how the metadata is ordered in the final library.


uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -366,26 +366,34 @@ fn object_interface_name(ci: &ComponentInterface, obj: &Object) -> String {
...
-        } else {
-            Ok(format!("{class_name}Interface"))
-        };
+fn trait_interface_name(ci: &ComponentInterface, trait_ty: &Type) -> Result<String> {
+    let namespace = match &trait_ty.module_path() {
+        Some(module_path) => ci.namespace_for_module_path(module_path)?,
+        None => bail!("Invalid trait_type: {trait_ty:?}"),
+    };
+    let Some(ci_look) = ci.find_component_interface(namespace) else {
+        anyhow::bail!("no interface {}", namespace);
```
mhammond: if you have `ci.find_component_interface` take a module path, the existing `find_ci_for` in that file could be replaced with a call to that plus the unwrap(). You don't otherwise use the namespace here so would avoid the lookup. I don't really care either way, but it seems a little cleaner?

mhammond (approved):
this is great, thanks!

PR #2621, opened 2025-08-15
askama 0.13->0.14. Significant as all filter function signatures change.

mhammond:
All filter functions now take at least 2 args; the new 2nd arg is now typically `_: &dyn askama::Values`. The `askama::Values` concept seems good in general.

@badboy @bendk this is going to be a little painful, but I guess we have to take this eventually so uniffi .30 seems like it should be fine? We'll probably need to upgrade nimbis-fml at the same time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,9 +94,9 @@ checksum = "a4668cab20f66d8d020e1fbc0ebe47217433c1b6c8f2040faf858554e394ace6"
 
 [[package]]
 name = "askama"
-version = "0.13.0"
+version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9a4e46abb203e00ef226442d452769233142bbfdd79c3941e84c8e61c4112543"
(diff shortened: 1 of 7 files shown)


badboy:
They increased the MSRV to Rust 1.83.0 in that release: https://github.com/askama-rs/askama/commit/04a4d5b0206378244f397c0f85c4fd1a2a865d13

m-c is at 1.82.0 for now: https://searchfox.org/mozilla-central/rev/67ff378efd69fbfb26d47ee4d9d0d249ab4be1d5/Cargo.toml#62-63

That might change by the time we get an UniFFI update into m-c, but would be a blocker for now.

anayw2001:
Hi folks, any updates on an ETA for this getting merged?

bendk:
I just checked the status in moz-central and it looks like [they're working to bump the MSRV version right now](https://bugzilla.mozilla.org/show_bug.cgi?id=1986626).  I'd expect a fix in the next few weeks, but it's hard to estimate.  My guess is that we'll release `0.30.0` with the current askama version, but maybe `0.30.1` could bump it.

PR #2626, opened 2025-08-24
Python: methods now have typing annotations for return values.

### LINKED ISSUE #2625: Foreign async traits missing return type in python
Given the following trait definition:

```rs
#[async_trait]
pub trait AsyncAdder: Send + Sync {
    async fn add_async(&self, a: u64, b: u64) -> u64;
}
```

The Python protocol is missing the return type:

```rs
class AsyncAdder():
    def add_async(self, a: "int",b: "int"):
        raise NotImplementedError
```

It does, however, still function as expected. Given the following function

```rs
#[uniffi::export]
pub async fn call_async_adder(adder: &dyn AsyncAdder, a: u64, b: u64) -> u64 {
    adder.add_async(a, b).await
}
```

The following python code works:

```py
class PythonAsyncAdder(AsyncAdder):
    async def add_async(self, a: int, b: int) -> int:
        return a + b


async def demo():
    res = await call_async_adder(PythonAsyncAdder(), 1, 2)
    assert res == 3, f"Expected 3, got {res}"
```

This seems to be a problem purely with the type generation. I should also point out the Uniffi-generated `AsycAdderImpl` for rust implementations is typed correctly. 

First observed on 0.28.3. Replicated on 0.29.4 here: https://github.com/joe-p/uniffi-wasm-playground/blob/2955c0e776e9f8c49beea8495afc65423309e934/consumers/python/playground.py?plain=1#L1489

Presumably needs a fix similar to https://github.com/mozilla/uniffi-rs/pull/1984

mhammond:
Fixes #2625

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
   ([#2543](https://github.com/mozilla/uniffi-rs/pull/2543)).
   Custom types too ([#2603](https://github.com/mozilla/uniffi-rs/pull/2603))
 - Kotlin: The `NoPointer` placeholder object used to create fake interface instances has been renamed to `NoHandle`
+- Python: methods now have typing annotations for return values ([#2625](https://github.com/mozilla/uniffi-rs/issues/2625))
 
 ### ⚠️ Breaking Changes for external bindings authors ⚠️
 
--- a/fixtures/coverall/src/lib.rs
+++ b/fixtures/coverall/src/lib.rs
@@ -655,4 +655,9 @@ pub fn validate_html(_source: String) -> Result<(), HTMLError> {
     Err(HTMLError::InvalidHTML)
 }
 
+#[uniffi::export]
+pub async fn async_bool(b: bool) -> bool {
+    b
+}
+
 uniffi::include_scaffolding!("coverall");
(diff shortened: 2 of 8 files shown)


bendk (approved):
Nice!

akhramov:
@mhammond pyright caught a problem for us on this one. The code that gets generated for this signature

```
pub async fn sleep2(duration: Duration) -> Duration {
```

is

```
async def sleep2(duration: Duration) -> Awaitable[Duration]:
```

It should rather be 

```
async def sleep2(duration: Duration) -> Duration:
```

(otherwise we would have to await twice)

mhammond:
does mypy also see that? I was surprised that I seemed to need that extra `Awaitable[]` - maybe I misread something though.

akhramov:
Sure, here's the universal repro.

```py
# bar.py
# generated bindings reside in bindings.py
import bindings
from datetime import timedelta

async def test() -> None:
  time = timedelta(milliseconds = 5)
  foo = await bindings.sleep(time)
  print(foo >= time)
```

```bash
$ mypy --strict bar.py | grep bar.py
bar.py:7: error: Unsupported operand types for <= ("timedelta" and "Awaitable[timedelta]")  [operator]
bar.py:7: note: Maybe you forgot to use "await"?
```

mhammond:
I think I vaguely remember how I confused myself this was necessary :) Would seem ideal if our mypy tests also covered our test code, but that's a bit tricky

Issue #2392, opened 2025-01-09
External Kotlin errors don't work


mhammond:
#2390 added better support for Swift and Python, but making Kotlin work was tricky enough to split out into its own issue.

The problem is, unsurprisingly, RustBuffers and how we lift the errors. We generate a [`UniffiRustCallStatusErrorHandler`](https://github.com/mozilla/uniffi-rs/blob/e6cb4a37cfd03ab053ab11324349feb280df3a22/uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt#L44-L46) for each module which is used to lift errors in that module, but `UniffiRustCallStatusErrorHandler` has a reference to this module's `RustBuffer`. When an external error is involved, `UniffiRustCallStatusErrorHandler` fails to work because the `RustBuffer` is the one from the external module.

So somehow we need to rejig the generic code here such that a reference to the remote RustBuffer can be specified.

PR #2629, opened 2025-08-27
Support throwing external Kotlin error

paxbun:
Fixes: #2392

This PR adds support for throwing external Kotlin errors by generating additional "delegating" error handlers in `ExternalTypeTemplate.kt`. If a function throws an external error, the "delegating" error handler is passed to `uniffiRustCallWithError` or `uniffiRustCallAsync` instead of the original one.


bendk (approved):
Looks good to me, thanks for doing this.

I really hope that when we start using the pipeline code we can simplify things by only defining a single `RustBuffer`, but this seems like a fine workaround in the meantime.

bendk:
I'm not sure what's going on with CI, but the other rust tests task passed so I'm just going to merge this.

PR #2637, opened 2025-09-03
Uprev toml to 0.8.23

anayw2001:


bendk:
I would love to bump the `toml` version, but it's currently at `0.5` because that's matches the version that Firefox uses.  Do you want to bump the version just to keep things up-to-date or does it fix a bug for you?

uniffi_macros/Cargo.toml
```diff
@@ -21,7 +21,7 @@ proc-macro2 = "1.0"
 quote = "1.0"
 serde = { version = "1.0.136", features = ["derive"] }
 syn = { version = "2.0", features = ["full", "visit-mut"] }
-toml = "0.5.9"
+toml = "0.8.23"
```
martinvonz: There's another instance in `uniffi_bindgen/Cargo.toml`

martinvonz:
> I would love to bump the `toml` version, but it's currently at `0.5` because that's matches the version that Firefox uses. 

I don't know anything about `unnffi-rs`, so I apologize for a naive question, but why does it need to match Firefox? Any chance Firefox can also upgrade?

> Do you want to bump the version just to keep things up-to-date or does it fix a bug for you?

We import this crate into Google's monorepo. That repo has a policy that allows at most 3 versions of crates at once. We would like to remove dependencies on toml 0.5 to allow us to import toml 0.9.

bendk:
> > I would love to bump the `toml` version, but it's currently at `0.5` because that's matches the version that Firefox uses.
> 
> I don't know anything about `unnffi-rs`, so I apologize for a naive question, but why does it need to match Firefox? Any chance Firefox can also upgrade?

Most of the UniFFI devs are from Mozilla so it's convenient if they match.

> 
> > Do you want to bump the version just to keep things up-to-date or does it fix a bug for you?
> 
> We import this crate into Google's monorepo. That repo has a policy that allows at most 3 versions of crates at once. We would like to remove dependencies on toml 0.5 to allow us to import toml 0.9.

This seems like a good reason to me to upgrade though.  Let me investigate how hard this would be and I'll get back to you on the best way to get this upgraded.  It should just be a day or two.

mhammond:
out of interest, is going straight to 0.9 going to cause the same problem for you?

[edit: oops, I see above, that's what you are actually trying to do!]

martinvonz:
> out of interest, is going straight to 0.9 going to cause the same problem for you?

Going to 0.8 will be easier because we currently have 0.5, 0.7, and 0.8. Going straight to 0.9 probably means that we will have to replace 0.5 by 0.9 and upgrade uniffi-rs in the same commit. That should be fine, but 0.8 would be easier.

mhammond:
OK thanks - another crate we depend on, https://github.com/martinthomson/ohttp, went directly from 0.5 to 0.9. We might still be able to sort something out though (our "policy" is more like "try very hard to have only 1, but if you can make a case for a dupe it might be fine")

martinvonz:
> our "policy" is more like "try very hard to have only 1, but if you can make a case for a dupe it might be fine"

I don't want to make you do something that's annoying for you for Google's benefit. I think we can deal with an upgrade straight from 0.5 to 0.9. We can presumably get a temporary exemption to allow 4 simultaneous versions if necessary.

mhammond:
thanks - and just to be sure, this is in anticipation of our next release, or do you vendor from a git rev?

martinvonz:
We normally vendor from crates.io, but until it gets there, we can apply a patch locally in our monorepo. So no need to rush a release for us, if that's what you mean. Thanks.

mhammond:
I was thinking more that if we decide we are moving to 0.9 for the next release we could jump to 0.6 in the meantime - sounds like that's not any help in practice though.

martinvonz:
Correct, 0.6 will unfortunately not help us.

mhammond:
`">=0.8, <=0.9"` seems to work and what we need? (didn't check where the *actual* breaking change happened tho, so maybe earlier?)

bendk:
> `">=0.8, <=0.9"` seems to work and what we need? (didn't check where the _actual_ breaking change happened tho, so maybe earlier?)

What do you think about doing that now for the `0.30.0` release?  That will force some changes in moz-central before we can take that upgrade, but we can just make those slowly since there won't be a rush to take `0.30.0`.

mhammond:
> What do you think about doing that now for the `0.30.0` release? That will force some changes in moz-central before we can take that upgrade, but we can just make those slowly since there won't be a rush to take `0.30.0`.

That sounds fine to me, especially given ohttp is going to force us anyway, and the only other dependency in-tree is also owned by moz.

uniffi_macros/Cargo.toml
anayw2001: Done, thanks!

bendk (approved):
Looks great, thanks for the update.

PR #2638, opened 2025-09-04
Avoid setting CARGO_REGISTRIES_CRATES_IO_PROTOCOL for MSR tests as it apparently breaks CI.

mhammond:


mhammond:
looks like this env var was the problem!?

PR #2641, opened 2025-09-06
Fix `is_name_used_as_error` returning `false` for unused external errors

### LINKED ISSUE #2636: Wrong Android error name when using nested external error
I've been trying the latest merged changes that add support for external errors on Android, and I've found that UniFFI is not correctly doing the `*Error` to `*Exception` rename when the external error is being used inside another error type:

```rust
// This error is defined in crate-a

#[derive(uniffi::Error)]
pub enum TestError {
    BadThingsHappened(String),
    ErrorFromCrateB(crate_b::CrateBError),
}

// This error is defined in crate-b
#[derive(uniffi::Error)]
pub enum CrateBError {
    CrateBScrewedUp(u64),
}
```

If I check the generated bindings, I can see that the rename is done correctly in `crate_b.kt`:
```kt
sealed class CrateBException: kotlin.Exception() {
...
}
```

But when it gets imported in `crate_a.kt`, the rename doesn't happen:
```kt
import uniffi.crate_b.CrateBError    // <!--- Compile error here

sealed class TestException: kotlin.Exception() {
    class BadThingsHappened(
        val v1: kotlin.String
        ) : TestException() { ... }
    
    class ErrorFromCrateB(
        val v1: CrateBError    // <!--- Compile error here
        ) : TestException() { ...  }
```

If I manually rename the `CrateBError` to `CrateBException` in the file everything works correctly, so this seems to only be a bug in the template generation. Let me know if you need more info, I have a repro here if it's helpful: https://github.com/dani-garcia/uniffi-android-error-repro

paxbun:
Fixes #2636 by making `ComponentInterface::is_name_used_as_error` return `true` for unused external errors using `all_component_interfaces` introduced in #2612. `self.errors.contains(name) ||` is still required for cases where library mode is not used, which is the reason why 422442381d3a21764eeca0de418655b2914a2d33 has failed.

### Why #2636 occurs

`self.error.insert` happens when the error enum itself is defined in the crate, or it is thrown from a function or a constructor. Since external errors are not defined in the crate, they are not registered through `self.error.insert` when it is not used. The check to determine whether an enum is an error requires an `EnumMetadata`, but `EnumMetadata` for external errors used in a variant of an enum is not being provided to CI.

### DIFF
--- a/fixtures/ext-types/lib/src/lib.rs
+++ b/fixtures/ext-types/lib/src/lib.rs
@@ -233,4 +233,12 @@ pub enum ContainsExternalError {
     ExternalError(NotToThrowError),
 }
 
+#[derive(uniffi::Record)]
+pub struct ContainsExternalError2 {
+    error: NotToThrowError,
+}
+
+#[uniffi::export]
+fn takes_external_error(_error: NotToThrowError) {}
+
 uniffi::include_scaffolding!("ext-types-lib");
--- a/fixtures/ext-types/lib/src/lib.rs
+++ b/fixtures/ext-types/lib/src/lib.rs
@@ -8,7 +8,7 @@ use uniffi_one::{
     UniffiOneEnum, UniffiOneError, UniffiOneErrorInterface, UniffiOneInterface,
     UniffiOneProcMacroType, UniffiOneTLA, UniffiOneTrait, UniffiOneType, UniffiOneUDLTrait,
 };
-use uniffi_sublib::SubLibType;
+use uniffi_sublib::{NotToThrowError, SubLibType};
 use url::Url;
 
 // Remote types require a macro call in the Rust source
@@ -226,6 +226,11 @@ fn get_uniffi_one_udl_trait(
     t: Option<Arc<dyn UniffiOneUDLTrait>>,
 ) -> Option<Arc<dyn UniffiOneUDLTrait>> {
     t
+}
+
+#[derive(uniffi::Error)]
+pub enum ContainsExternalError {
+    ExternalError(NotToThrowError),
 }
 
 uniffi::include_scaffolding!("ext-types-lib");
--- a/uniffi_bindgen/src/interface/mod.rs
+++ b/uniffi_bindgen/src/interface/mod.rs
@@ -1021,6 +1021,10 @@ impl ComponentInterface {
 
     pub fn is_name_used_as_error(&self, name: &str) -> bool {
         self.errors.contains(name)
+            || self
+                .all_component_interfaces
+                .iter()
+                .any(|ci| ci.errors.contains(name))
     }
 
     /// Called by `APIBuilder` impls to add a newly-parsed callback interface definition to the `ComponentInterface`.
--- a/fixtures/ext-types/sub-lib/src/lib.rs
+++ b/fixtures/ext-types/sub-lib/src/lib.rs
@@ -26,4 +26,19 @@ fn get_trait_impl() -> Arc<dyn UniffiOneTrait> {
     Arc::new(OneImpl {})
 }
 
+#[derive(uniffi::Error, Debug)]
+pub enum NotToThrowError {
+    Variant(u64),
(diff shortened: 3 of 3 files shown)


uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1021,6 +1021,10 @@ impl ComponentInterface {
 
     pub fn is_name_used_as_error(&self, name: &str) -> bool {
         self.errors.contains(name)
```
mhammond, line 1023: you could just remove this?

paxbun: Thanks for the review! That part is still needed (as all_component_interfaces is set only in library mode) and removing it breaks some tests; would setting all_component_interfaces in non-library mode be better?

mhammond: right 😅 "library mode" has always seemed strange to me. What does "non library mode" mean with external types? Non library mode having only a clone of itself in `all_component_interfaces` seems fine too, but not worth it here i guess?

(part of the confusion is actually around udl generating rust scaffolding - that's "not library mode" - but also doesn't need `all_component_interfaces`. In terms of bindings, "not library mode" shouldn't be a thing)

paxbun: I haven't had a deep look at the UDL parsing part (as I'm primarily using library mode), but if you find it awkward that the library mode cloning CIs to populate `all_component_interfaces`, how about removing the grouping logic in `library_mode::find_components`? In UDL files, external types to be used are all declared again, and it seems the UDL parsing logic doesn't re-group the metadata at all...

fixtures/ext-types/lib/src/lib.rs
```diff
@@ -228,4 +228,9 @@ fn get_uniffi_one_udl_trait(
     t
 }
 
+#[derive(uniffi::Error)]
+pub enum ContainsExternalError {
+    ExternalError(NotToThrowError),
+}
+
```
paxbun, line 235: @mhammond This is the case that @dani-garcia mentioned; should I add cases like where `NotToThrowError` is used as parameters or record properties for better test coverage?

uniffi_bindgen/src/interface/mod.rs
mhammond: UDL parsing is done for both scaffolding and bindings - ie, is used by library mode too. Further, individual crates can use both udl and macros - so I don't think we can just remove the grouping in that way.

fixtures/ext-types/lib/src/lib.rs
mhammond: > @mhammond This is the case that @dani-garcia mentioned; 

sorry, I've lost some context - mentioned where?

> should I add cases like where `NotToThrowError` is used as parameters or record properties for better test coverage?

I guess that can't hurt, but probably isn't strictly necessary - just a test which fails without your patch but works with it seems fine to me.

mhammond:
Hey @paxbun, we are hoping to cut a new release next week or so, so if you are able to update this with tests it can still make it!

fixtures/ext-types/lib/src/lib.rs
paxbun: I've just seen this comment right now; sorry for the late reply. This is the case that @dani-garcia mentioned in #2641; this is a minimal reproducible error. Without this fix, the code added to the fixture in this PR will fail to compile.

image

Should I add more test cases as I described above? I'll be working on it tonight KST (So it's tomorrow morning ET).

paxbun: Added more test cases as I said above. As you can see in 8717fddcfd99284d71fefcccac68935689207, the bindgen renders `NotToThrowError` instead of `NotToThrowException`, making the build fail. After the fix, the bindgen always renders `NotToThrowException`, and the build succeeds.

PR #2642, opened 2025-09-08
Reorg python templates; builtins get their own directory.

mhammond:
This seems a nice cleanup towards reduced complexity with pipelines.


bendk (approved):
I think this is an improvement.

I also wonder if we could have a single `Builtin.py` file with a big switch statement.  A lot of these modules are tiny and we often want to change them together.  WDYT?

mhammond:
> I also wonder if we could have a single Builtin.py file with a big switch statement. 

That sounds good. OTOH, I still haven't entirely given up on having more nodes specify their own template, which this might work against (but tbh, I don't think that idea of having more nodes with their own template is actually going to work well enough in practice, so 🤷 )

PR #2648, opened 2025-09-10
Adding BindgenLoader

bendk:
I'm opening this based on our discussion in https://github.com/mozilla/uniffi-rs/pull/2640.

@paxbun can you take a look  at the new Swift code and tell me if you think this would work for Gobley?

---

This is a new way for binding generators to load the CIs, Configs, metadata, etc.  I'm hoping that this can replace the `generate_bindings` and `generate_external_bindings` functions, as well some other functions like `library_mode::find_components`.

The goal for the new design is to let bindings generators drive the process instead of a function like `generate_external_bindings`.  The advantage of this is that it's easier to customize and we don't need to keep adding new hook methods.  It also feels simpler overall to me.  See https://github.com/mozilla/uniffi-rs/pull/2640 for a discussion of this.

If we adopt this new system, then I think we can use it to remove a bunch of duplicate code.  We can keep around the old functions for backwards-compatibility, but I think they can just be wrappers around `BindgenLoader`.  Also, we should figure out a nice way to hook this up to the pipeline code.

Made `uniffi-bindgen-swift` use the new system.  This was mostly to test the code and to serve as an example, but a nice side-benefit is that `uniffi-bindgen-swift` now can handle UDL files.  I also added a `uniffi-bindgen-swift` binary file to go with the `uniffi-bindgen` binary.

### DIFF
--- a/uniffi/Cargo.toml
+++ b/uniffi/Cargo.toml
@@ -68,3 +68,9 @@ name = "uniffi-bindgen"
 path = "uniffi-bindgen.rs"
 required-features = ["cli"]
 doc = false
+
+[[bin]]
+name = "uniffi-bindgen-swift"
+path = "uniffi-bindgen-swift.rs"
+required-features = ["cli"]
+doc = false
--- a/uniffi/src/cli/swift.rs
+++ b/uniffi/src/cli/swift.rs
@@ -13,8 +13,9 @@ use uniffi_bindgen::bindings::{generate_swift_bindings, SwiftBindingsOptions};
 struct Cli {
     #[command(flatten)]
     kinds: Kinds,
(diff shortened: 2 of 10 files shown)


paxbun:
@bendk Looks good to me as well! I don't remember the exact situation, but I once thought that I wanted to adjust the timing at which `BindingGenerator::update_component_configs` is invoked, and this would help us achieve it more easily.

bendk:
> This seems good to me. I'm mildly surprised Swift needed so many changes instead of being able to reuse more of the core, but that's a me problem :)

That's mostly because I was using Swift to demonstrate the changes.  I'll do a second pass to update the other parts.

PR #2656, opened 2025-09-16
Uprev to askama 0.14

anayw2001:
Add askama::Values as the second parameter to all filters to line up with what askama expects.


anayw2001:
Just noticed this was a duplicate of https://github.com/mozilla/uniffi-rs/pull/2621. Closing.

PR #2657, opened 2025-09-18
Fix python relative import

flisky:
Python 3.9+ use `from . import module`, and `import .module` throws `SyntaxError`.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/filters.rs
+++ b/uniffi_bindgen/src/bindings/python/filters.rs
@@ -30,3 +30,16 @@ pub fn docstring(docstring: &Option<String>, indent: usize) -> Result<String> {
     let indented = textwrap::indent(&escaped, &indent);
     Ok(format!("\"\"\"\n{indented}\n\"\"\"\n{indent}"))
 }
+
+/// Get the idiomatic Python import statement for a module
+pub fn import_statement(module: &str) -> Result<String> {
+    Ok(if module.starts_with('.') {
+        let Some((from, name)) = module.rsplit_once('.') else {
+            unreachable!()
+        };
+        let from = if from.is_empty() { "." } else { from };
+        format!("from {from} import {name}")
+    } else {
+        format!("import {module}")
+    })
+}
--- a/uniffi_bindgen/src/bindings/python/templates/Module.py
+++ b/uniffi_bindgen/src/bindings/python/templates/Module.py
@@ -31,7 +31,7 @@ import asyncio
 {%- endif %}
 import platform
 {%- for import in imports %}
-import {{ import }}
+{{ import|import_statement }}
 {%- endfor %}
 
 


bendk:
What version of Python are you using?  I'm pretty sure this is the standard since 3.0, or at least early in the 3.x series.

flisky:
It failed at least from python 3.9 and above.  

(And it goes `TypeError: 'type' object is not subscriptable` with`dict[str, str]` on python 3.8)

bendk:
Can you post some code to reproduce the issue?  I'm pretty sure we're successfully using this with python 3.9.  I'm guessing it has to do with the code formatting rather than the import statements themselves.  Maybe 2 lines are getting squashed together without a newline or something.

badboy:
He is right that `import .module` is invalid syntax though.

```python
; python3 --version
Python 3.12.11
; cat a.py
import .b
; python3 a.py
  File "/private/tmp/pytest/a.py", line 1
    import .b
           ^
SyntaxError: invalid syntax
```

* https://docs.python.org/3/reference/import.html#submodules
* https://docs.python.org/3/reference/import.html#package-relative-imports

---

Glean doesn't have any additional imports, so we didn't hit that issue yet.
I don't actually know where `imports` is set though in UniFFI.

bendk:
Oh, I think I was just misreading the description.  So this patch replaces `import .module` with `from . import module`?  That sounds right to me as well.

badboy:
> Oh, I think I was just misreading the description. So this patch replaces `import .module` with `from . import module`? That sounds right to me as well.

Yup. Though I would like to understand where we (or the user?) generates an import name of `.module`

mhammond:
I haven't had a chance to look in detail, but we should be careful here - we have some hackery to try and support either relative or absolute imports, to handle when the intent is for the generated modules to exist in a package or as top-level modules. I don't think we have great tests for this in tree.

Not great docs at https://mozilla.github.io/uniffi-rs/latest/python/configuration.html#external-packages

badboy:
Ah, I guess here? https://github.com/mozilla/uniffi-rs/blob/6d46b3f756dde3213357c477d86771a0fc5da7b4/uniffi_bindgen/src/bindings/python/pipeline/external_types.rs#L43

mhammond:
> Ah, I guess here? 

Yeah - that's the code I was referring to and I don't think we have great tests for that case - our tests pretty much all do something like https://github.com/mozilla/uniffi-rs/blob/6d46b3f756dde3213357c477d86771a0fc5da7b4/fixtures/ext-types/lib/uniffi.toml. I'm out of time, and can't recall if we have *few* tests for the package case or literally none :(

badboy:
> > Ah, I guess here?
> 
> Yeah - that's the code I was referring to and I don't think we have great tests for that case - our tests pretty much all do something like [`6d46b3f`/fixtures/ext-types/lib/uniffi.toml](https://github.com/mozilla/uniffi-rs/blob/6d46b3f756dde3213357c477d86771a0fc5da7b4/fixtures/ext-types/lib/uniffi.toml). I'm out of time, and can't recall if we have _few_ tests for the package case or literally none :(

I can make some time tomorrow to look at this.

mhammond:
cc @joe-p

joe-p:
I have python bindings that use external types

On 0.29.4 everything works fine. The imports are generated as:

```py
from .algokit_http_client import HttpClient
from .algokit_http_client import _UniffiConverterTypeHttpClient
from .algokit_transact_ffi import BoxReference
from .algokit_transact_ffi import OnApplicationComplete
from .algokit_transact_ffi import SignedTransaction
from .algokit_transact_ffi import StateSchema
from .algokit_transact_ffi import Transaction
from .algokit_transact_ffi import _UniffiConverterTypeBoxReference
from .algokit_transact_ffi import _UniffiConverterTypeOnApplicationComplete
from .algokit_transact_ffi import _UniffiConverterTypeSignedTransaction
from .algokit_transact_ffi import _UniffiConverterTypeStateSchema
from .algokit_transact_ffi import _UniffiConverterTypeTransaction
from .algokit_http_client import _UniffiRustBuffer as _UniffiRustBufferHttpClient
from .algokit_transact_ffi import _UniffiRustBuffer as _UniffiRustBufferBoxReference
from .algokit_transact_ffi import _UniffiRustBuffer as _UniffiRustBufferOnApplicationComplete
from .algokit_transact_ffi import _UniffiRustBuffer as _UniffiRustBufferSignedTransaction
from .algokit_transact_ffi import _UniffiRustBuffer as _UniffiRustBufferStateSchema
from .algokit_transact_ffi import _UniffiRustBuffer as _UniffiRustBufferTransaction
```


Switching to main, the imports become

```py
import .algokit_transact_ffi
import .algokit_http_client
```

This then results in the following error on Python 3.13:

```
E     File "/Users/joe/git/algorandfoundation/algokit-core/packages/python/algokit_utils/algokit
_utils/algokit_utils_ffi.py", line 30
E       import .algokit_transact_ffi
E              ^
E   SyntaxError: 
```

Manually changing the generated python binding file to 

```py
from . import algokit_transact_ffi
from . import algokit_http_client
```

Results in the imports working correctly and tests once again passing (except for async, but that seems to be a seperate bug)

Definitely not a **minimal** repro, but If you want to reproduce my exact setup the repo is here: https://github.com/algorandfoundation/algokit-core/tree/spike/utils_uniffi_main

`cargo pkg utils py` to generate the bindings followed by `cd packages/python/algokit_utils && poetry run pytest` to run the tests and see the error.

badboy:
Ok, actually ran the algokit tests now using this patch (instead of just reading the code). The syntax error is gone and the import works (tests fail due to other things, but that's because I don't have a full dev setup for that project).

I'm gonna merge this.

joe-p:
I can confirm this is now working as expected on my end. Thanks!

PR #2658, opened 2025-09-18
Support `Type::Custom{ builtin: Enum|Object, .. }` as throws type

flisky:


uniffi_bindgen/src/bindings/python/templates/CallableBody.py
```diff
@@ -29,7 +29,7 @@
 
 {%- match callable.throws_type.ty %}
 {%- when Some(e) %}
-{%-    match e.ty %}
+{%-    match e.ty.canonical() %}
```
mhammond: I think the right thing here is a new pipeline pass, eg, "error.rs"? It would walk `Callable` and adjust `throws_type` directly. This would also then handle CallbackInterfaceImpl.py

mhammond (changes requested):
This looks great, thanks. A changelog entry would be nice too. cc @bendk to check my thinking re the pipeline?

Issue #2651, opened 2025-09-15
The "Ensure" section of ffi.swift is not stable - changes often on recompile


bes:
I have a build step where I build our Rust library on every build in XCode. Most of the time nothing has changed, but UniFFI still changes / updates the ffi.swift file, but the only thing that changes is the order of the "Ensure" section.

Example:

First compilation

```swift
uniffiEnsureCrateOneInitialized()
uniffiEnsureCrateTwoInitialized()
uniffiEnsureCrateThreeInitialized()
```

Second compilation
```swift
uniffiEnsureCrateTwoInitialized()
uniffiEnsureCrateOneInitialized()
uniffiEnsureCrateThreeInitialized()
```

It would be nice if this was deterministic!

badboy:
I assume that it's this part: https://github.com/mozilla/uniffi-rs/blob/2d1912a57f731f9fc4936d0cc5270c3f416a1044/uniffi_bindgen/src/bindings/swift/templates/wrapper.swift#L60-L62

which gets filled in here: https://github.com/mozilla/uniffi-rs/blob/2d1912a57f731f9fc4936d0cc5270c3f416a1044/uniffi_bindgen/src/bindings/swift/gen_swift/mod.rs#L489-L512

It might be as easy as:

```patch
- init_fns.chain(extern_module_init_fns).collect()
+ let mut fns = init_fns.chain(extern_module_init_fns).collect();
+ fns.sort();
+ return fns;
```

mhammond:
fixed by #2660

PR #2660, opened 2025-09-19
swift: Make initialization fns have a stable ordering (#2651)

bendk:


mhammond (approved):
maybe worth a changelog?

Issue #2659, opened 2025-09-19
async foreign trait returning an external type in python results in TypeError: "_UniffiRustBuffer instance instead of _UniffiRustBuffer"


joe-p:
I have Python tests that work on 0.29.4. 

Switching to [1db0c36](https://github.com/mozilla/uniffi-rs/commit/1db0c3650a81c90d623f1c12f86cb4361e5929af) results in the Python runtime hanging when calling an async foreign trait due to the following exception. Also below is the code

To summarize: `TransactionSigner` is a async foreign trait (*not* external) which returns `SignedTransaction`, which is an external type. Presumably the external type is the issue here, but I don't have the time right now to minimally reproduce. I'm not sure if the trait being foreign and/or async has an impact here, but I would suspect not. I can try to make a minimal reproduction if helpful later this weekend. 

Note: I did stumble upon this in #2627 but that issue was a bit of a mess as it was really multiple problems in one.
```
tests/test_utils.py DEBUGPRINT[64]: test_utils.py:48 (after def get_signer(self, address: str) -
> Tr…)
DEBUGPRINT[62]: test_utils.py:105 (before await composer.gather_signatures())
DEBUGPRINT[66]: test_utils.py:38 (after stxns.append(SignedTransaction(transacti…)
UniFFI: Unhandled exception in trait interface call
Traceback (most recent call last):
  File "/Users/joe/git/algorandfoundation/algokit-core/packages/python/algokit_utils/algokit_uti
ls/algokit_utils_ffi.py", line 1194, in make_call_and_call_callback
    handle_success(await make_call())
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/Users/joe/git/algorandfoundation/algokit-core/packages/python/algokit_utils/algokit_uti
ls/algokit_utils_ffi.py", line 1491, in handle_success
    _UniffiForeignFutureResultRustBuffer(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        _UniffiFfiConverterSequenceTypeSignedTransaction.lower(return_value),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        _UniffiRustCallStatus.default()
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
[...]

PR #2669, opened 2025-09-26
test: reproduce #2659

joe-p:
Minimal reproduction of the issue in #2659 

I want to highlight that the error occurs when adding the definition to the crate. **Not even using the trait in python** results in the error from the issue when calling a trait that was previously working. 

Video of the behavior:

https://github.com/user-attachments/assets/c1954171-f4d1-41f6-9a95-9c8d4840f8b7

### DIFF
--- a/fixtures/futures/tests/bindings/test_futures.py
+++ b/fixtures/futures/tests/bindings/test_futures.py
@@ -154,7 +154,9 @@ class TestFutures(unittest.TestCase):
 
         async def test():
             trait_obj = PyAsyncParser()
+            print("DEBUGPRINT[82]: test_futures.py:159 (before self.assertEqual(await as_string_using_t…)")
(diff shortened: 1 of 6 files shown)


fixtures/futures/tests/bindings/test_futures.py
```diff
@@ -154,7 +154,9 @@ async def try_delay(self, delay_ms):
 
         async def test():
             trait_obj = PyAsyncParser()
+            print("DEBUGPRINT[82]: test_futures.py:159 (before self.assertEqual(await as_string_using_t…)")
             self.assertEqual(await as_string_using_trait(trait_obj, 1, 42), "42")
```
joe-p, line 158: When the new trait with an external return type is added, this starts to fail despite the fact that it was previously working. This means the definition of the trait in rust, even if not used in python, is ultimately what causes this error

PR #2671, opened 2025-09-28
Use root object protocol conformance

jkmassel:
In Swift, extension protocol conformance cannot be made public – it’s tied to the original module (and conformance can’t be re-declared later).

This has implications for cases where types from this library are part of the public API of another library.

For example, in the following setup:

```
+-----------------------------------------------------------+
| Application                                               |
|  (calls public API of FooLibrary)                         |
|                                                           |
|  +---------------------------------------------+          |
|  | FooLibrary (Library A)                      |          |
|  |  - Exports select Rust types                |          |
|  |  - Associated some free-floating functions  |          |
|  |    with types using `public extension`      |          |
|  |   +-------------------------------------+   |          |
|  |   | UniFFIWrapper                           |          |
|  |   +-------------------------------------+   |          |
|  +---------------------------------------------+          |
|                                                           |
+-----------------------------------------------------------+
```

Extension-based Protocol Conformances in `UniFFIWrapper` cannot be made `public` (Swift will say `'public' modifier cannot be used with extensions that declare protocol conformances`). To avoid breaking encapsulation (and thus needing `Application` to import both `FooLibrary` and `UniFFIWrapper` the protocol conformance must be part of the type declaration which _is_ public. 

Trying to take advantage of `Equatable` in `Application` in a file where `import FooLibrary` is present but we haven't imported `UniFFIWrapper`, Swift will emit the following error:

> Cannot use conformance of 'MyObject' to 'Equatable' here; 'UniFFIWrapper' was not imported by this file

for the following code:

```swift
extension MyObject: Comparable {
    public static func < (lhs: FooObject, rhs: FooObject) -> Bool {
        false
    }
}
```

A few things to note:
- This PR moves the conformance listing out of the template into code because the combinatorial complexity makes the template a bit of a mess – it's a lot cleaner to do it this way. The `CustomTypeConfig` might not be the right place for this code to live, but it seemed nicely aligned with a lot of the properties we needed to read.
- This PR removes the conditional `Sendable` conformance used elsewhere, because it's inconsistently applied – in much of the rest of the codebase we're not checking the compiler version so this is now aligned.
- There was previously a note about `Codable` conformance – questioning whether it was applied correctly – this implementation is conservative, only applying it where it's explicitly requested by the configuration but also where there are no class-based members. Class-based members are not themselves `Codable` so it's unlikely they could benefit from automated conformance.
- This PR switches to automatically-derived `Hashable` and `Equatable` by the Swift compiler except in cases where the UniFFI traits `eq` or `hash` are applied – in that case the Rust implementation is used instead.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
@@ -4,7 +4,7 @@
 {%- let uniffi_trait_methods = e.uniffi_trait_methods() %}
 {% match e.variant_discr_type() %}
 {% when None %}
-public enum {{ type_name }} {
+public enum {{ type_name }}: {{ config.conformance_list_for_enum(e, contains_object_references) }} {
     {% for variant in e.variants() %}
     {%- call swift::docstring(variant, 4) %}
     case {{ variant.name()|enum_variant_swift_quoted }}{% if variant.fields().len() > 0 %}(
@@ -12,7 +12,7 @@ public enum {{ type_name }} {
     ){% endif -%}
     {% endfor %}
 {% when Some(variant_discr_type) %}
-public enum {{ type_name }} : {{ variant_discr_type|type_name }} {
+public enum {{ type_name }}: {{ variant_discr_type|type_name }}, {{ config.conformance_list_for_enum(e, contains_object_references) }} {
     {% for variant in e.variants() %}
     {%- call swift::docstring(variant, 4) %}
     case {{ variant.name()|enum_variant_swift_quoted }} = {{ e|variant_discr_literal(loop.index0) }}{% if variant.fields().len() > 0 %}(
@@ -20,10 +20,9 @@ public enum {{ type_name }} : {{ variant_discr_type|type_name }} {
     ){% endif -%}
     {% endfor %}
 {% endmatch %}
+
+{% call swift::uniffi_trait_impls(uniffi_trait_methods) %}
 }
-#if compiler(>=6)
-extension {{ type_name }}: Sendable {}
-#endif
 
 #if swift(>=5.8)
 @_documentation(visibility: private)
@@ -85,21 +84,3 @@ public func {{ ffi_converter_name }}_lift(_ buf: RustBuffer) throws -> {{ type_n
 public func {{ ffi_converter_name }}_lower(_ value: {{ type_name }}) -> RustBuffer {
     return {{ ffi_converter_name }}.lower(value)
 }
-
-{% call swift::uniffi_trait_impls(uniffi_trait_methods) %}
-
-{%- if !contains_object_references %}
-{# We auto-generate `Equatable, Hashable`, but only if we have no objects. We could do better - see #2409 #}
-{% if !(contains_object_references || uniffi_trait_methods.eq_eq.is_some() || uniffi_trait_methods.hash_hash.is_some()) %}
-extension {{ type_name }}: Equatable, Hashable {}
-{% endif %}
-
-{# Not clear if `generate_codable_conformance` actually needs to avoid objects too? #}
-{%-     if config.generate_codable_conformance() %}
-extension {{ type_name }}: Codable {}
-{%-     endif %}
-{% endif %}
-
-{% if config.generate_case_iterable_conformance() && !e.contains_variant_fields() %}
-extension {{ type_name }}: CaseIterable {}
-{% endif %}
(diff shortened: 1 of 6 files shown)


uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
```diff
@@ -85,21 +84,3 @@ public func {{ ffi_converter_name }}_lift(_ buf: RustBuffer) throws -> {{ type_n
 public func {{ ffi_converter_name }}_lower(_ value: {{ type_name }}) -> RustBuffer {
     return {{ ffi_converter_name }}.lower(value)
 }
-
-{% call swift::uniffi_trait_impls(uniffi_trait_methods) %}
-
-{%- if !contains_object_references %}
-{# We auto-generate `Equatable, Hashable`, but only if we have no objects. We could do better - see #2409 #}
```
mhammond, line 92: we should carry this comment forward.

uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
```diff
@@ -4,26 +4,25 @@
...
     case {{ variant.name()|enum_variant_swift_quoted }} = {{ e|variant_discr_literal(loop.index0) }}{% if variant.fields().len() > 0 %}(
         {%- call swift::field_list_decl(variant, variant.has_nameless_fields()) %}
     ){% endif -%}
     {% endfor %}
 {% endmatch %}
+
+{% call swift::uniffi_trait_impls(uniffi_trait_methods) %}
 }
-#if compiler(>=6)
-extension {{ type_name }}: Sendable {}
```
mhammond, line 25: it looks like your CI failures are hitting this as our docker is on an old swift?

jkmassel: Interestingly it's not _quite_ that simple – upgrading to Swift 6 will fix the problem, but not for language reasons, it's because in the standard library that ships with Swift 6, `Date` becomes `Sendable`.

There's a few different ways we could approach this:

1. Update the Docker image to use Swift 6. The downside here is that it effectively raises the minimum Swift version to 6 for this entire project, which might not be desirable – especially because the vtable implementation can't be easily made Swift Concurrency-compatible (though [I'm curious if `consume` can help here] when moving pointers and closures into a `Task`(https://github.com/swiftlang/swift-evolution/blob/main/proposals/0366-move-function.md)).
2. I don't know exactly when `Date` was made `Sendable`, but it's not in `5.2` (the Swift version that used by `docker/cargo-docker.sh` or `5.5` (the version used in CI). `5.10` seems like a good candidate because it was after the `Foundation` unification, but I don't have an easy way to test that locally without access to the build scripts for these containers. This nicely threads the needle between updating to Swift 6 while bringing the project forward a bit from a pretty outdated Swift version.
3. I've been doing some research and found that `Sendable is a marker protocol`, which AFAICT means that the compiler doesn't need to actually generate an implementation for it. AFAICT that means it doesn't _need_ to be part of the public type signature – declaration in an extensions persists into the object metadata emitted by the module either way and preserves the marker.

For backward compatibility, I think option 3 is probably best unless y'all were planning to raise the required Swift version anyway. But even if that's the case, that should probably be its own PR so it can be more easily rolled back if needed.

WDYT?

uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
jkmassel: In https://github.com/mozilla/uniffi-rs/pull/2671/commits/59a8b18471ddaa586c281a6938b88addaec419d1, I hoisted this into `conformance_list_for_record` and `conformance_list_for_error` (it was already in `conformance_list_for_enum`). 

Is that what you had in mind?

uniffi_bindgen/src/bindings/swift/templates/EnumTemplate.swift
jkmassel: Implemented in https://github.com/mozilla/uniffi-rs/pull/2671/commits/501a245d44aeb885cf119a81e88d31342e18fc8a since it seems like the most sensible way forward. That has an issue (seems related to `CustomStringConvertable` or similar) that I'll figure out next.

mhammond: > Update the Docker image to use Swift 6. 

I was under the impression that just updating it to a later 5 would solve some issues - I don't think we are ready to force a move to 6 yet as 6 gives sendable errors with async functions. Most of us are pretty clueless about Swift in general.

I don't fully understand (3) but it does sound like it might be a good option.

jkmassel: [501a245](https://github.com/mozilla/uniffi-rs/pull/2671/commits/501a245d44aeb885cf119a81e88d31342e18fc8a) and [12122a6](https://github.com/mozilla/uniffi-rs/pull/2671/commits/12122a64a39d69f555a3f061396cdd3eaeeadf72) now implement Option 3, and I think it turned out pretty good – AFAICT it's fully backward-compatible.

Issue #2662, opened 2025-09-20
A C function pointer can only be formed from a reference to a 'func' or a literal closure


lifr0m:
When exposing any async function to swift, Xcode gives error on lint and on compilation:

<img width="1121" height="589" alt="Image" src="https://github.com/user-attachments/assets/802e17a9-bbed-4df9-bd77-e579f0e3b119" />

It can be mitigated by replacing this line to:
```
{ handle, pollResult in
    uniffiFutureContinuationCallback(handle: handle, pollResult: pollResult)
}
```

I personally don't know why Xcode complains because generated code seems valid.

Swift version 5.
Xcode version 26.0 (latest).
Rust 1.90.0.
Uniffi version 0.29.4.

mhammond:
fixed in #2674

PR #2674, opened 2025-09-29
Fix: Wrap Swift async poll callback in closure

duszmox:
Fix Xcode’s “A C function pointer can only be formed…” error by wrapping the Swift async poll callback in a literal closure that forwards to uniffiFutureContinuationCallback.

Referenced in #2662
(I was also facing the same problem, and ended up using a terminal patching script for a while before filing this PR)


bendk (approved):
I don't get why this is only an error in some scenarios, but the change makes sense to me.

bendk:
Thanks for the contribution!

PR #2676, opened 2025-09-29
Python: Fix async functions that return external types (#2659)

bendk:
The issue was that we were using the same name for both the local and external types, which lead to the one of them being lost when they were de-duped.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/pipeline/external_types.rs
+++ b/uniffi_bindgen/src/bindings/python/pipeline/external_types.rs
@@ -59,6 +59,18 @@ pub fn pass(namespace: &mut Namespace) -> Result<()> {
             _ => (),
         };
     });
+    // Transform RustBuffer namespaces to concrete Python module names
+    namespace.visit_mut(|ffi_type: &mut FfiType| {
+        if let FfiType::RustBuffer(Some(ref mut namespace)) = ffi_type {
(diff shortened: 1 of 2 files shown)


bendk:
@joe-p This fixes things for me with the test case you provided.  Can you double check that it works for you?

joe-p:
It looks like it uses the incorrect module name by not using the package provided in the `bindings.python.external_packages` config. If I remove the external_packages mapping, however, it works. 

So it works, but needs to be updated to use the proper external package name

bendk:
> it would be interesting to lean more into fully-qualified named instead of the "from foo import" bar syntax - I think we make lots of new symbols in this way and it always bothers me a little :)

Totally agree.  It's tricky because this is happening in the general code and we don't currently have the concept of "importing a FFI struct definition from another crate".  Maybe we should add that, but I'm not sure exactly what the cleanest way would be.  In the back of my head I'm thinking we can avoid this as part of the FFI-1.0 work.

> It looks like it uses the incorrect module name by not using the package provided in the bindings.python.external_packages config. If I remove the external_packages mapping, however, it works.

I think this is fixed, do you have a good way of testing it?  I needed to manually move around files and setup the PYTHONPATH.

joe-p:
Confirmed to be working on my end 👍

PR #2229, opened 2024-09-02
Kotlin: Switch to JNA direct mapping

badboy:
According to https://github.com/java-native-access/jna/blob/master/www/DirectMapping.md direct mapping can improve performance substantially, maybe even that of custom JNI. Given we know all methods it's easy to switch it all over.

Note that I have not run any actual benchmark to see how it helps.

PR #2678, opened 2025-09-30
Kotlin: Pass explicit `T::class.java` to `Native.register`

paxbun:
A follow-up PR for #2229. This is to prevent `Native.register` from throwing `IllegalStateException` on Android like the following:

```
java.lang.IllegalStateException: The SecurityManager implementation on this platform is broken; you must explicitly provide the class to register
```


mhammond (approved):
thanks!

PR #2694, opened 2025-10-08
Use BindgenLoader for the Python pipeline

bendk:
I'm liking this BindgenLoader and want to use it in more places, this is a small step towards that.


mhammond (approved):
This looks good to me!

PR #2695, opened 2025-10-09
Store full module path in the metadata

bendk:
We don't use it yet, but I'm hoping to use it in the future and I think it's good to start distinguishing between crate and module names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 ## [[UnreleasedUniFFIVersion]] (backend crates: [[UnreleasedBackendVersion]]) - (_[[ReleaseDate]]_)
 
+### ⚠️ Breaking Changes for external bindings authors ⚠️
+- The `module_path` field now stores full module paths rather than crate names only.
+  External bindings authors will probably need to make some minor changes to work with this.
+  See https://github.com/mozilla/uniffi-rs/pull/2695/ for examples.
+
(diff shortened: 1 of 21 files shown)


uniffi_meta/src/util.rs
```diff
@@ -0,0 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+pub fn crate_name(module_path: &str) -> &str {
```
mhammond: do we really want a new module just for this?

praveenperera:
@bendk  This is causing issues with submodules, this fixes it for me: https://github.com/mozilla/uniffi-rs/pull/2701

But I don't have the full context of the initial change.

---

Original:

After this change, I had to make some changes, I was putting some impls in `mod ffi` if the impls were only ever used in the Swift/Kotlin side, but with this I was getting a conflicting types error.

```
  conflicting types:
     cur: Object { module_path: "cove::manager::coin_control_manager", name: "RustCoinControlManager", imp: Struct }
     new: Object { module_path: "cove::manager::coin_control_manager::ffi", name: "RustCoinControlManager", imp: Struct }
```

The fix was pretty simple, just get rid of the `ffi` modules: https://github.com/bitcoinppl/cove/pull/342/commits/c3ee1cb1b12ce216b357860dda360d810ffe0002

Issue #2402, opened 2025-01-14
Conflicting type names for enum variant and variant data


chris13524:
We have an enum like so:

```rust
#[cfg_attr(feature = "uniffi", derive(uniffi::Enum))]
pub enum SignOutputEnum {
    Signature(Bytes),
    SignOutput(SignOutput),
}
```

But we get this error:

```
yttrium.kt:4352:17 Type mismatch: inferred type is uniffi.yttrium.SignOutput but uniffi.yttrium.SignOutputEnum.SignOutput was expected
```

```
public object FfiConverterTypeSignOutputEnum : FfiConverterRustBuffer<SignOutputEnum>{
    override fun read(buf: ByteBuffer): SignOutputEnum {
        return when(buf.getInt()) {
            1 -> SignOutputEnum.Signature(
                FfiConverterTypeBytes.read(buf),
                )
            2 -> SignOutputEnum.SignOutput(
                FfiConverterTypeSignOutput.read(buf), // == error on this line ==
                )
            else -> throw RuntimeException("invalid enum value, something is very wrong!!")
        }
    } 
}
```

We renamed the `SignOutput` struct to `SignOutputObject` for now as a workaround, but it would be nice to not have this conflict in the first place as this is a convenient pattern to have the variant data type name match the enum variant name when there are deeply nested types.

We are currently using commit rev `e796e00ad150f8b14b61a859a2e8c6497b35074e` and have also encountered the same issue on the latest rev, `c7c4809ed0e6bb9195b58855cc7cb211eac13290`.

mhammond:
I think this is a dupe of #1137, even though that issue mentions UDL.

mhammond:
Turns out this is Kotlin specific, Python and Swift are fine with it.

chris13524:
Just ran into this for the second or third time! Curious if any plans for a fix

Thoralf-M:
Just had the same problem, to explain the problem in a different way since this might be helpful for others:
The Kotlin definition gets generated here https://github.com/mozilla/uniffi-rs/blob/73619b2043c8f6d1b9563a6843ccfa3d633f3196/uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt#L60
and for the enum posted in this issue it generates a class like this
```Kotlin
sealed class SignOutputEnum {
    
    data class Signature(
        val v1: Bytes) : SignOutputEnum() {
        companion object
    }

    data class SignOutput(
[...]

PR #2698, opened 2025-10-12
Kotlin: Generate enums bindings using fully qualified type names

praveenperera:
Closes: #2402 

Using this in my project and its working: https://github.com/bitcoinppl/cove/pull/340


fixtures/regressions/kotlin-enum-payload-clash/src/test.udl
```diff
@@ -0,0 +1,28 @@
+namespace regression_kotlin_enum_payload_clash {
```
mhammond: fwiw, this would have been fine without udl too.

praveenperera:
Added a CHANGELOG entry, should I delete the UDL files and just use proc macros?

mhammond:
> should I delete the UDL files and just use proc macros?

I'm not too bothered but less is more imo :) I'll leave that up to you.

praveenperera:
Makes sense, I removed the UDL files and just using proc-macros like in my other PR, its cleaner.

Issue #2699, opened 2025-10-13
Display trait not working for Kotlin errors + simple enums


thunderbiscuit:
After reading the changelog for the 0.30.0 release I tried to clean up a long standing issue of ours in our uniffi codebase, errors. Enabling the Display trait is a great feature because we can then expose both the fields important to the errors as well as provide a clear message back to the developers reading these errors/exceptions (we previously had to choose between having the field or having the relevant message).

However there appears to be a bug on the Kotlin side for exporting Display. I'm opening this issue to report it.

Consider the following 3 examples:

```rust
#[derive(uniffi::Error, Debug)]
#[uniffi::export(Display)]
pub enum Example1Enum {
    Blue { id: u32 },
    Green { issue: String },
}

impl Display for Example1Enum {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "Display for Example1Enum {:?}", self)
    }
}

#[derive(uniffi::Enum, Debug)]
#[uniffi::export(Display)]
pub enum Example2Enum {
    Blue,
    Green,
}

impl Display for Example2Enum {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "Display for Example2Enum {:?}", self)
    }
}

#[derive(uniffi::Enum, Debug)]
#[uniffi::export(Display)]
pub enum Example3Enum {
    Blue { id: u32 },
    Green { issue: String },
}

impl Display for Example3Enum {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "Display for Example3Enum {:?}", self)
    }
}
```

All three enums will produce the `description` field in Swift with the comment `// The local Rust Display implementation.`

Only the 3rd example, which is expressed as a Sealed Class in Kotlin, will get a `toString()` method implemented. The uniffi::Error and the simple Kotlin enum will not.

Note that in Python it's also a bit different than both Kotlin and Swift:
- Example1Enum (the error) gets a `__repr__` method but no `__str__`
- Example2Enum (the straight enum) gets nothing (like Kotlin)
- Example3Enum gets a `__str__` method

PR #2700, opened 2025-10-13
Kotlin: Fix display and other trait export generation

praveenperera:
Closes: #2699 

Tested separately in: https://github.com/bitcoinppl/cove/pull/342/files as well

### DIFF
--- a/fixtures/trait-methods/src/lib.rs
+++ b/fixtures/trait-methods/src/lib.rs
@@ -299,22 +299,14 @@ fn throw_flat_error(i: u8) -> Result<(), FlatError> {
     }
 }
 
-// mixed error with all possible exported traits
+// error with multiple exported traits
 #[derive(Debug, uniffi::Error, thiserror::Error, Clone)]
 #[uniffi::export(Debug, Display, Eq, Ord, Hash)]
 pub enum MultipleTraitError {
     #[error("MultipleTraitError::NoData")]
     NoData,
-    #[error("MultipleTraitError::WithCode({code})")]
-    WithCode { code: i32 },
-    #[error("MultipleTraitError::AnotherFlat")]
-    AnotherFlat,
-    #[error("MultipleTraitError::WithMessage({msg})")]
-    WithMessage { msg: String },
-    #[error("nested simple error: {0}")]
-    NestedSimple(FlatError),
-    #[error("nested complex error [{tag}]: {inner}")]
-    NestedComplex { inner: FlatError, tag: String },
+    #[error("nested error: {0}")]
+    Nested(FlatError),
 }
 
 // error that doesn't end in "Error" suffix to test class_name filter doesn't break this case
@@ -323,12 +315,8 @@ pub enum MultipleTraitError {
 pub enum ApiFailure {
     #[error("api network issue")]
     NetworkIssue,
-    #[error("api timeout after {duration_ms}ms")]
-    Timeout { duration_ms: u32 },
     #[error("api server down")]
     ServerDown,
-    #[error("api rate limited: {retry_after}s")]
-    RateLimited { retry_after: u32 },
 }
 
 impl Ord for ApiFailure {
@@ -336,19 +324,8 @@ impl Ord for ApiFailure {
         match (self, other) {
             (ApiFailure::NetworkIssue, ApiFailure::NetworkIssue) => std::cmp::Ordering::Equal,
             (ApiFailure::NetworkIssue, _) => std::cmp::Ordering::Less,
-            (ApiFailure::Timeout { .. }, ApiFailure::NetworkIssue) => std::cmp::Ordering::Greater,
-            (ApiFailure::Timeout { duration_ms: d1 }, ApiFailure::Timeout { duration_ms: d2 }) => {
-                d1.cmp(d2)
-            }
-            (ApiFailure::Timeout { .. }, _) => std::cmp::Ordering::Less,
-            (ApiFailure::ServerDown, ApiFailure::RateLimited { .. }) => std::cmp::Ordering::Less,
+            (ApiFailure::ServerDown, ApiFailure::NetworkIssue) => std::cmp::Ordering::Greater,
             (ApiFailure::ServerDown, ApiFailure::ServerDown) => std::cmp::Ordering::Equal,
-            (ApiFailure::ServerDown, _) => std::cmp::Ordering::Greater,
-            (
-                ApiFailure::RateLimited { retry_after: r1 },
-                ApiFailure::RateLimited { retry_after: r2 },
-            ) => r1.cmp(r2),
-            (ApiFailure::RateLimited { .. }, _) => std::cmp::Ordering::Greater,
         }
     }
 }
@@ -377,20 +354,8 @@ impl Hash for ApiFailure {
 fn throw_api_failure(i: u8) -> Result<(), ApiFailure> {
     match i {
         0 => Err(ApiFailure::NetworkIssue),
-        1 => Err(ApiFailure::Timeout { duration_ms: 5000 }),
-        2 => Err(ApiFailure::ServerDown),
-        _ => Err(ApiFailure::RateLimited { retry_after: 60 }),
+        _ => Err(ApiFailure::ServerDown),
     }
-}
-
-#[uniffi::export]
-fn throw_api_failure_timeout(duration_ms: u32) -> Result<(), ApiFailure> {
-    Err(ApiFailure::Timeout { duration_ms })
-}
-
-#[uniffi::export]
-fn throw_api_failure_rate_limited(retry_after: u32) -> Result<(), ApiFailure> {
-    Err(ApiFailure::RateLimited { retry_after })
 }
 
 impl Ord for MultipleTraitError {
@@ -398,40 +363,12 @@ impl Ord for MultipleTraitError {
         match (self, other) {
             (MultipleTraitError::NoData, MultipleTraitError::NoData) => std::cmp::Ordering::Equal,
             (MultipleTraitError::NoData, _) => std::cmp::Ordering::Less,
-            (MultipleTraitError::WithCode { .. }, MultipleTraitError::NoData) => {
+            (MultipleTraitError::Nested(_), MultipleTraitError::NoData) => {
                 std::cmp::Ordering::Greater
             }
-            (MultipleTraitError::WithCode { .. }, MultipleTraitError::WithCode { .. }) => {
+            (MultipleTraitError::Nested(_), MultipleTraitError::Nested(_)) => {
                 std::cmp::Ordering::Equal
             }
-            (MultipleTraitError::WithCode { .. }, _) => std::cmp::Ordering::Less,
-            (MultipleTraitError::AnotherFlat, MultipleTraitError::WithMessage { .. }) => {
-                std::cmp::Ordering::Less
-            }
-            (MultipleTraitError::AnotherFlat, MultipleTraitError::AnotherFlat) => {
-                std::cmp::Ordering::Equal
-            }
-            (MultipleTraitError::AnotherFlat, _) => std::cmp::Ordering::Greater,
-            (MultipleTraitError::WithMessage { .. }, MultipleTraitError::WithMessage { .. }) => {
-                std::cmp::Ordering::Equal
-            }
-            (
-                MultipleTraitError::WithMessage { .. },
-                MultipleTraitError::NestedSimple(_) | MultipleTraitError::NestedComplex { .. },
-            ) => std::cmp::Ordering::Less,
-            (MultipleTraitError::WithMessage { .. }, _) => std::cmp::Ordering::Greater,
-            (MultipleTraitError::NestedSimple(_), MultipleTraitError::NestedSimple(_)) => {
-                std::cmp::Ordering::Equal
-            }
-            (MultipleTraitError::NestedSimple(_), MultipleTraitError::NestedComplex { .. }) => {
-                std::cmp::Ordering::Less
-            }
-            (MultipleTraitError::NestedSimple(_), _) => std::cmp::Ordering::Greater,
-            (
-                MultipleTraitError::NestedComplex { .. },
-                MultipleTraitError::NestedComplex { .. },
-            ) => std::cmp::Ordering::Equal,
-            (MultipleTraitError::NestedComplex { .. }, _) => std::cmp::Ordering::Greater,
         }
     }
 }
@@ -460,16 +397,7 @@ impl Hash for MultipleTraitError {
 fn throw_multiple_trait_error(i: u8) -> Result<(), MultipleTraitError> {
     match i {
         0 => Err(MultipleTraitError::NoData),
-        1 => Err(MultipleTraitError::WithCode { code: 42 }),
-        2 => Err(MultipleTraitError::AnotherFlat),
-        3 => Err(MultipleTraitError::WithMessage {
-            msg: "test".to_string(),
-        }),
-        4 => Err(MultipleTraitError::NestedSimple(FlatError::NotFound)),
-        _ => Err(MultipleTraitError::NestedComplex {
-            inner: FlatError::Unauthorized,
-            tag: "complex".to_string(),
-        }),
+        _ => Err(MultipleTraitError::Nested(FlatError::NotFound)),
     }
 }
 
--- a/fixtures/trait-methods/tests/bindings/test.kts
+++ b/fixtures/trait-methods/tests/bindings/test.kts
@@ -61,8 +61,7 @@ try {
     assert(e.toString() == "error: not found")
 }
 
-// mixed error with all possible exported traits (Debug, Display, Eq, Ord, Hash)
-// tests flat variants, field variants, and nested Display
+// error with multiple exported traits (Debug, Display, Eq, Ord, Hash)
 try {
     throwMultipleTraitError(0u)
     throw AssertionError("should have thrown")
@@ -70,26 +69,12 @@ try {
     assert(e.toString() == "MultipleTraitError::NoData")
 }
 
+// test nested error with Display
 try {
     throwMultipleTraitError(1u)
     throw AssertionError("should have thrown")
-} catch (e: MultipleTraitException.WithCode) {
-    assert(e.toString() == "MultipleTraitError::WithCode(42)")
-}
-
-// test nested error with Display
-try {
-    throwMultipleTraitError(4u)
-    throw AssertionError("should have thrown")
-} catch (e: MultipleTraitException.NestedSimple) {
-    assert(e.toString() == "nested simple error: error: not found")
-}
-
-try {
-    throwMultipleTraitError(5u)
-    throw AssertionError("should have thrown")
-} catch (e: MultipleTraitException.NestedComplex) {
-    assert(e.toString() == "nested complex error [complex]: error: unauthorized")
+} catch (e: MultipleTraitException.Nested) {
+    assert(e.toString() == "nested error: error: not found")
 }
 
 // test Eq on flat variants
@@ -108,8 +93,8 @@ try {
 // test Ord
 var err2: MultipleTraitException? = null
 try {
-    throwMultipleTraitError(2u)
-} catch (e: MultipleTraitException.AnotherFlat) {
+    throwMultipleTraitError(1u)
+} catch (e: MultipleTraitException.Nested) {
     err2 = e
 }
 assert(err1!! < err2!!)
@@ -134,3 +119,12 @@ try {
 } catch (e: ApiFailure.NetworkIssue) {
     assert(apiErr == e)
 }
+
+// test Ord on non-Error type
+var apiErr2: ApiFailure? = null
+try {
+    throwApiFailure(1u)
+} catch (e: ApiFailure.ServerDown) {
+    apiErr2 = e
+}
+assert(apiErr!! < apiErr2!!)
--- a/uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
@@ -14,7 +14,6 @@ sealed class {{ type_name }}(message: String): kotlin.Exception(message){% if co
         class {{ variant|error_variant_name }}(message: String) : {{ type_name }}(message)
         {% endfor %}
 
-
     companion object ErrorHandler : UniffiRustCallStatusErrorHandler<{{ type_name }}> {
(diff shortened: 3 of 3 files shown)


praveenperera:
@mhammond i think this ready now, tests passing locally I'm not sure why the lint is not passing. And the test failing looks unrelated?

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
 
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.30.0...HEAD).
 
+### What's Fixed
+
+- Kotlin: Enums and Errors now support exporting trait methods (Display, Debug, Eq, Hash, Ord) via `toString()`,
```
mhammond: thanks! Maybe call out *flat enums*, remove the last sentence and a PR/issue number :)

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -4,19 +4,34 @@
...
+    {%- if contains_object_references %}Disposable{% endif -%}
+    {%- if contains_object_references && uniffi_trait_methods.ord_cmp.is_some() %}, {% endif -%}
+    {%- if uniffi_trait_methods.ord_cmp.is_some() %}Comparable<{{ type_name }}>{% endif -%}
+{% endif %} {
         {% for variant in e.variants() -%}
         {%- call kt::docstring(variant, 4) %}
         class {{ variant|error_variant_name }}(message: String) : {{ type_name }}(message)
         {% endfor %}
 
+    {%- let uniffi_trait_methods = e.uniffi_trait_methods() %}
```
mhammond: i think the decl above works here?

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -15,6 +15,10 @@ enum class {{ type_name }} {
     {%- call kt::docstring(variant, 4) %}
     {{ variant|variant_name }}{% if loop.last %};{% else %},{% endif %}
     {%- endfor %}
+
+    {%- let uniffi_trait_methods = e.uniffi_trait_methods() %}
```
mhammond, line 19: not sure about this - there are no tests for it, and I'm not sure it makes sense to support this for a numeric enum?

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -51,7 +59,13 @@ public object {{ e|ffi_converter_name }}: FfiConverterRustBuffer<{{ type_name }}
 {% else %}
 
 {%- call kt::docstring(e, 0) %}
-sealed class {{ type_name }}{% if contains_object_references %}: Disposable {% endif %} {
+sealed class {{ type_name }}
```
mhammond: I don't understand these changes and don't see test changes for non-flat enums.

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -4,19 +4,34 @@
 
 {% if e.is_flat() %}
 {%- call kt::docstring(e, 0) %}
-sealed class {{ type_name }}(message: String): kotlin.Exception(message){% if contains_object_references %}, Disposable {% endif %} {
+sealed class {{ type_name }}(message: String): kotlin.Exception(message)
+{%- let uniffi_trait_methods = e.uniffi_trait_methods() -%}
+{%- if contains_object_references || uniffi_trait_methods.ord_cmp.is_some() %},
+    {%- if contains_object_references %}Disposable{% endif -%}
+    {%- if contains_object_references && uniffi_trait_methods.ord_cmp.is_some() %}, {% endif -%}
+    {%- if uniffi_trait_methods.ord_cmp.is_some() %}Comparable<{{ type_name }}>{% endif -%}
```
mhammond: just add comparable like it adds disposable (newlines make sense tho!)?

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -4,19 +4,34 @@
...
+    {%- call kt::uniffi_trait_impls(uniffi_trait_methods) %}
+
     companion object ErrorHandler : UniffiRustCallStatusErrorHandler<{{ type_name }}> {
         override fun lift(error_buf: RustBuffer.ByValue): {{ type_name }} = {{ ffi_converter_name }}.lift(error_buf)
     }
 }
 {%- else %}
 {%- call kt::docstring(e, 0) %}
-sealed class {{ type_name }}: kotlin.Exception(){% if contains_object_references %}, Disposable {% endif %} {
+sealed class {{ type_name }}: kotlin.Exception()
```
mhammond: ditto here

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -4,19 +4,34 @@
...
+    {%- if contains_object_references && uniffi_trait_methods.ord_cmp.is_some() %}, {% endif -%}
+    {%- if uniffi_trait_methods.ord_cmp.is_some() %}Comparable<{{ type_name }}>{% endif -%}
+{% endif %} {
         {% for variant in e.variants() -%}
         {%- call kt::docstring(variant, 4) %}
         class {{ variant|error_variant_name }}(message: String) : {{ type_name }}(message)
         {% endfor %}
 
+    {%- let uniffi_trait_methods = e.uniffi_trait_methods() %}
+    {%- call kt::uniffi_trait_impls(uniffi_trait_methods) %}
```
mhammond: I wonder why enums need each variant to implement these but errors don't?

CHANGELOG.md
praveenperera: This should work with nested enums as long as the "parent" enum implements the trait, I was trying a few different things and I think I left in things I shouldn't have. But I will add tests for non flat enums and update the PR.

fixtures/trait-methods/src/lib.rs
```diff
@@ -267,4 +242,165 @@ impl Hash for UdlEnum {
...
+    );
+    assert_eq!(
+        Ord::cmp(&TraitEnum::I(0), &TraitEnum::I(1)),
+        std::cmp::Ordering::Equal
+    );
+    assert_eq!(TraitEnum::I(0), TraitEnum::I(1));
+    assert_ne!(s1, TraitEnum::I(0));
+    assert!(s1 < TraitEnum::I(0));
+}
+
```
praveenperera, line 428: moved test to the bottom

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -15,6 +15,14 @@ enum class {{ type_name }} {
     {%- call kt::docstring(variant, 4) %}
     {{ variant|variant_name }}{% if loop.last %};{% else %},{% endif %}
     {%- endfor %}
+
+    {%- let uniffi_trait_methods = e.uniffi_trait_methods() %}
+    {%- if let Some(fmt) = uniffi_trait_methods.display_fmt.or(uniffi_trait_methods.debug_fmt.clone()) %}
+    override fun toString(): String {
+        return {{ fmt.return_type().unwrap()|lift_fn }}({% call kt::to_ffi_call(fmt) %})
+    }
+    {%- endif %}
```
praveenperera: Flat enums already have all the other methods so adding just `toString()` with an override

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -51,7 +67,11 @@ public object {{ e|ffi_converter_name }}: FfiConverterRustBuffer<{{ type_name }}
 {% else %}
 
 {%- call kt::docstring(e, 0) %}
-sealed class {{ type_name }}{% if contains_object_references %}: Disposable {% endif %} {
+sealed class {{ type_name }}{% if contains_object_references %}: Disposable {% endif %}
+{%- let uniffi_trait_methods = e.uniffi_trait_methods() -%}
+{%- if uniffi_trait_methods.ord_cmp.is_some() -%}
+{% if contains_object_references %}, {% else %} : {% endif %}Comparable<{{ type_name }}>
+{%- endif %} {
```
praveenperera, line 66: Remove the unnecessary stuff but put it on a new line

CHANGELOG.md
```diff
@@ -6,18 +6,12 @@
...
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.30.0...HEAD).
 
-### What's New?
+### What's Fixed
 
-- Kotlin: Rust enums with nested payload types are generated using the inner type’s fully qualified
-  name, avoiding naming conflicts and allowing payloads to reuse the variant name ([#2698](https://
-  github.com/mozilla/uniffi-rs/pull/2698)).
+- Kotlin: Enums and errors now support exporting trait methods (Display, Debug, Eq, Hash, Ord) via `toString()`,
+  `equals()`, `hashCode()`, and `compareTo()` implementations ([#2700](https://github.com/mozilla/uniffi-rs/pull/2700)).
```
praveenperera: We have tests for non flat enums now so hoping this is okay now

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
praveenperera: added a test with numeric enum that has `Display` implemented on it

CHANGELOG.md
mhammond: We always did though, right?

praveenperera: Yes, the test I added was for a flat enum that traits exported (58973af7e) in which case we have to generate just the `toString()` method as an override.

But with that I don't think we need to mention flat vs not, because its all handled?

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
mhammond: I might have got things back to front in my last comment - this case is the non-numeric enum. Why wouldn't this support all uniffi traits?

For numeric enums I'm still not sure it makes sense - what use-case do you see which isn't covered simply by knowing the enum's numeric value?

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
mhammond: I don't get this either - what makes `display` magic or special here?

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
```diff
@@ -93,6 +113,9 @@ sealed class {{ type_name }}{% if contains_object_references %}: Disposable {% e
         }.let { /* this makes the `when` an expression, which ensures it is exhaustive */ }
     }
     {% endif %}
+
+    {%- call kt::uniffi_trait_impls(uniffi_trait_methods) %}
```
mhammond, line 109: I think this is wrong - the existing call a few lines above should handle this.

CHANGELOG.md
mhammond: As I noted, we already have those tests. I haven't spent much time trying to understand this, so can you please explain exactly what problem for non-flat enums you are fixing here?

praveenperera:
@mhammond Flat enums get generated as kotlin `enum` in which all the methods expect `toString()` are marked final and can't be overwritten, which is what makes `toString()` special. Non flat enums (with payloads) are generated are `sealed class` which doesn't have the same restriction.

The test I added catches this and without the special `toString()` handling fails with:

```
trait_methods.kt:2765:5: error: 'hashCode' in 'NumericEnum' is final and cannot be overridden.
    override fun hashCode(): Int = value.hashCode()
```

Though since we aren't exposing the actual `hash`, `ord` and `equal` impls we always fallback to the ones provided by kotlin, I don't know what the best path forward is.

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
praveenperera: this is for the methods in the outer `Enum`, the one above is exporting the functions for the payloads.

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
praveenperera: I think it makes sense to support `toString()` even in these cases because you can derive `Display` on it, for example if they want to turn a enum variant into a helpful message, but more importantly, since its working for all other enums, users would assume it would work for numeric enums too, since I think most people think of all of them simplify as enums.

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
praveenperera: @mhammond added a `MixedEnum` test (https://github.com/mozilla/uniffi-rs/pull/2700/commits/ad67e97f91f595c881eeed751624bb7eb83a7e0a)  to hopefully show more clearly why this is needed. Without this line you get.

Right now the tests reference particular line numbers, I will get rid of those, just wanted to add that for clarity, or I can get rid of the test all together.

```
/Users/praveen/code/uniffi-rs/target/tmp/uniffi_trait_methods-de816aef8d9dd269/uniffi/trait_methods/trait_methods.kt:2230:5: error: object 'MixedEnum.NoFields' is not abstract and does not implement abstract member:
fun compareTo(other: T): Int
    object NoFields : MixedEnum()
    ^^^^^^^^^^^^^^^
/Users/praveen/code/uniffi-rs/target/tmp/uniffi_trait_methods-de816aef8d9dd269/uniffi/trait_methods/trait_methods.kt:2280:5: error: object 'MixedEnum.AnotherNoFields' is not abstract and does not implement abstract member:
fun compareTo(other: T): Int
    object AnotherNoFields : MixedEnum()
    ^^^^^^^^^^^^^^^^^^^^^^
test uniffi_foreign_language_testcase_test_kts ... FAILED

failures:

---- uniffi_foreign_language_testcase_test_kts stdout ----
Creating testing out_dir: /Users/praveen/code/uniffi-rs/target/tmp/uniffi_trait_methods-de816aef8d9dd269
Error: running `kotlinc` failed
```

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
praveenperera: you're right, I've removed it

praveenperera:
I’m going to put this back into draft 

### Todo:

- [x] Add mixed flat and with fields error test case like with Enum
- [x] Handle trait methods other than `toString()` on not flat errors not working because of inconsistent  renaming of `Error` —> `Exception` (completed in https://github.com/mozilla/uniffi-rs/commit/bd1d545372d96abcb9ca1215e25818628f7d6c18)

### Unanswered question

~~How to handle flat enums that can’t override any trait method other than `toString()`?~~

**Went with option 3 in: c68d17f06**

**options**

1. Only generate `toString()` and let kotlin fallback to its own generated funcs for the rest. (current option, but could cause logic errors to users)
2. Don’t generate any trait methods for flat enums 
3. Panic if anything other than `toString()` is trying to be generated

praveenperera:
@mhammond this is ready to review now

> Unanswered question
> How to handle flat enums that can’t override any trait method other than toString()?

Decided to go with option 3 and panic if a flat error tries to export traits other than `Display` when creating kotlin bindings.

Let me know if you want me to squash some commits. Also let me know if I should add a test for the last commit.

fixtures/trait-methods/src/lib.rs
```diff
@@ -267,4 +242,462 @@ impl Hash for UdlEnum {
...
+            count: 0,
+        })),
+    }
+}
+
+// mixed enum with object variants (no fields) and data class variants (with fields)
+// this tests that parent-level trait implementations work for object variants
+#[derive(uniffi::Enum, Debug, Clone)]
+#[uniffi::export(Debug, Display, Eq, Ord, Hash)]
+pub enum MixedEnum {
```
mhammond: I'd rather see one new variant added to `TraitEnum` than a completely new enum here. There's a lot of duplicated testing here which makes it difficult to see exactly what doesn't work today and makes things more maintainable over time.

fixtures/trait-methods/tests/bindings/test.kts
```diff
@@ -28,13 +28,373 @@ assert(TraitRecord(s="hi", i=2) != TraitRecord(s="yo", i=3))
...
+
+// mixed enum with object variants (requires parent-level trait impls - line 117)
+// object variants have no fields, so they can't have their own trait impl bodies
+// they MUST inherit from the sealed parent's implementations
+val objectVariant1 = getMixedEnum(0u) // NoFields - object variant
+val objectVariant2 = getMixedEnum(2u) // AnotherNoFields - object variant
+val dataVariant1 = getMixedEnum(1u)   // WithString - data class variant
+val dataVariant2 = getMixedEnum(3u)   // WithNumber - data class variant
+
+// test toString on object variants - will fail without line 117
```
mhammond: "line 17" isn't going to age well (and isn't even great today - line 17 of what exactly? Of this file? It looks like you haven't touched this)

mhammond (commented):
It's going to take me some time to decouple all of this - it would be helpful if we can trim the tests down to only things that don't work, even even then, try and repro in existing code if you can - eg, if the problem is an enum variant with zero fields, just add a new variant.

fixtures/trait-methods/tests/bindings/test.kts
praveenperera: yes will remove all references to specific lines, I thought it might help your review.

line 117 is referring to the line refed in this comment: https://github.com/mozilla/uniffi-rs/pull/2700#discussion_r2429877033

fixtures/trait-methods/src/lib.rs
praveenperera: that makes sense, ive removed MixedEnum and added the case to TraitEnum

praveenperera:
@mhammond I removed a bunch of redundant tests, hopefully its easier to review now

fixtures/trait-methods/src/lib.rs
```diff
@@ -267,4 +249,253 @@ impl Hash for UdlEnum {
...
+    AnotherFlat,
+    #[error("MultipleTraitError::WithMessage({msg})")]
+    WithMessage { msg: String },
+    #[error("nested simple error: {0}")]
+    NestedSimple(FlatError),
+    #[error("nested complex error [{tag}]: {inner}")]
+    NestedComplex { inner: FlatError, tag: String },
+}
+
+// error that doesn't end in "Error" suffix to test class_name filter doesn't break this case
```
mhammond: this seems unnecessary in this test - what value does MultipleTraitError add that this does not? Maybe merged into ApiError (or something concise if you really need 2, with the difference obvious)?

fixtures/trait-methods/src/lib.rs
```diff
@@ -267,4 +249,253 @@ impl Hash for UdlEnum {
...
+
+#[uniffi::export]
+fn throw_api_failure_rate_limited(retry_after: u32) -> Result<(), ApiFailure> {
+    Err(ApiFailure::RateLimited { retry_after })
+}
+
+impl Ord for MultipleTraitError {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match (self, other) {
+            (MultipleTraitError::NoData, MultipleTraitError::NoData) => std::cmp::Ordering::Equal,
```
mhammond, line 364: this is all a lot of code to verify some fairly simple checks that need to be made.

fixtures/trait-methods/src/lib.rs
```diff
@@ -267,4 +249,253 @@ impl Hash for UdlEnum {
...
+#[uniffi::export]
+fn throw_flat_error(i: u8) -> Result<(), FlatError> {
+    match i {
+        0 => Err(FlatError::NotFound),
+        1 => Err(FlatError::Unauthorized),
+        _ => Err(FlatError::InternalError),
+    }
+}
+
+// mixed error with all possible exported traits
```
mhammond: I don't think we need such rich error tests - can we get away with 1 error type and fewer variants?

uniffi_bindgen/src/bindings/kotlin/templates/EnumTemplate.kt
mhammond: Ah yeah - I missed that the enum itself is comparable now - previously it was just the variants. This makes more sense to me now!

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
```diff
@@ -4,19 +4,28 @@
...
+{%- let uniffi_trait_methods = e.uniffi_trait_methods() -%}
+{%- if uniffi_trait_methods.ord_cmp.is_some() -%}
+{% if contains_object_references %}, {% else %}, {% endif %}Comparable<{{ type_name }}>
+{%- endif %} {
         {% for variant in e.variants() -%}
         {%- call kt::docstring(variant, 4) %}
         class {{ variant|error_variant_name }}(message: String) : {{ type_name }}(message)
         {% endfor %}
 
+
```
mhammond: nit: we don't really need this?

mhammond (commented):
Sorry for the delay and I think I've got my head around it. I don't think we need quite as comprehensive testing of every possible variant, we just need to be sure the correct rust fn is being called. So if you are ok with it, I still think it can be quite a bit smaller without reducing actual test coverage. But thanks for the fixes!

uniffi_bindgen/src/bindings/kotlin/templates/ErrorTemplate.kt
praveenperera: Will remove

fixtures/trait-methods/src/lib.rs
praveenperera: Ya that’s fair I’ll see how I can get rid of most that tomorrow.

praveenperera:
@mhammond simplified the tests, I still left `ApiFailure` because it doesn't end in `Error` so it doesn't get transformed to `...Exception` but I can remove it completely if you prefer.

mhammond (approved):
Thanks for your patience :)

PR #2701, opened 2025-10-14
Fallback to crate name look up if full module path fails

praveenperera:


praveenperera:
done in https://github.com/mozilla/uniffi-rs/pull/2702

PR #2702, opened 2025-10-14
Make namespace_for_module_path use the crate name

bendk:
This should have been changed as part of #2695, but we didn't notice it because it didn't seem to cause any issues.  However, @praveenperera pointed out that it was causing issues for him in https://github.com/mozilla/uniffi-rs/pull/2695#issuecomment-3402475701.

I believe we can fix it by calculating crate_name from the module_path. That makes the assert! unneeded.


praveenperera:
awesome thanks @bendk I'll close my draft PR

badboy (approved):

PR #2705, opened 2025-10-21
Benchmark suite improvements

bendk:
Added more the benchmark test cases, rather than have 3 cases [there's now around a dozen](https://github.com/mozilla/uniffi-rs/blob/7b753b7188b0b39f071ee71d961faba5908c438b/fixtures/benchmarks/src/lib.rs#L87-L160).  Each case targets passing a particular kind of data across the FFI, both as arguments and return values.  Theses allow us to test performance changes with much more specificity.  The one case that we no longer include is passing args, but returning void.  I don't think that one was particularly important.

Also added a simple Android app to run the benchmarks so that we can:
- Double-check performance gains with real devices
- Run benchmarks under a profiler on real devices to see how closely they match real-world scenarios.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1668,6 +1668,8 @@ version = "0.22.0"
 dependencies = [
  "clap",
  "criterion",
+ "regex",
+ "thiserror",
  "uniffi",
  "uniffi_bindgen",
 ]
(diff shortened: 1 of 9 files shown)


mhammond (approved):
Bit of a rubber stamp, but lgtm

badboy (approved):

bendk:
Updating the benchmarks so that all inputs are referenced.  I'm slightly worried that compilers could optimize out lifting these values if they realize they're never used.

Issue #2704, opened 2025-10-17
Support for defining methods on records and enums


anubhavm11:
Hello, I was looking through the changelog for `0.30.0` and saw this line:

> Enums and Records can have methods, so the Method now carries self_type instead of the object name. In the templates, for Callable.takes_self() is replaced with Callable.self_type().

So I tried defining methods on a record:

```rust
#[derive(uniffi::Enum)]
pub enum Weekday {
    Sunday,
    Monday,
    Tuesday,
    Wednesday,
    Thursday,
    Friday,
    Saturday,
}

#[uniffi::export]
impl Weekday {
    #[uniffi::method]
    fn abbreviation(&self) -> String {
        let abbr = match self {
            Weekday::Sunday => "S",
            Weekday::Monday => "M",
            Weekday::Tuesday => "T",
            Weekday::Wednesday => "W",
            Weekday::Thursday => "Th",
            Weekday::Friday => "F",
            Weekday::Saturday => "Sa",
        };

        abbr.to_string()
    }
}
```

I found that although the above code compiles without issues, generating bindings errors out with the following:

`add_method_meta: object Weekday not found`

This code does not compile in `0.29.2`, the previous version I was using, giving the following error:

```
the trait bound `schedule::event::Weekday: FfiConverterArc<UniFfiTag>` is not satisfied
the following other types implement trait `FfiConverterArc<UT>`:
  ...
required for `Arc<schedule::event::Weekday>` to implement `FfiConverter<UniFfiTag>`
required for `Arc<schedule::event::Weekday>` to implement `Lift<UniFfiTag>
```

I looked in [the error site](https://github.com/mozilla/uniffi-rs/blob/bfe625c01e7037773854ea67ea000d8bdbf184c3/uniffi_bindgen/src/interface/mod.rs#L967) in `uniffi_bindgen/src/interface/mod.rs` and it looks like defining methods is still not available for non-objects, so I was wondering if this is a bug. As I have not previously run into a situation where my code compiles but the bindings generation errors out.

mhammond:
Sorry, that changelog could be clearer - this is in the section for external bindings authors, so for them, it is indeed true that enums and records can have methods. However, the only methods currently supported for them are the "uniffi traits" - ie, exposing the `Eq`, `Cmp` etc traits. It's not possible (yet?) to have user-defined methods on these types, even though the underlying metadata and internal types would support it (and therefore external binding authors need to take that into account)

anubhavm11:
[...]

Issue #1935, opened 2024-01-08
Constructors for records


Sajjon:
The `#[uniffi::constructor]` macro *can* be attached to a function on a `uniffi::Recrod` struct's function returning `Self` today, however, it does not generate and constructor in the generated Swift code. It would be nice to support constructors for records.

[Follows up thread on Matrix](https://matrix.to/#/!tsNDDGHvNGVNmvOOzi:mozilla.org/$mBzEIb41NBunAfCBo4TLSA6Fpsd0a3lX89Ex38ho3bY?via=mozilla.org&via=matrix.org&via=element.io), @bendk asked me to create a GH issue.

jplatte:
`#[uniffi::constructor]` only has an effect inside of `#[uniffi::export]`-ed `impl` blocks.

So I would re-frame this feature request as supporting exported impl blocks for non-object types (records, but also enums). I think constructors are only one of the ways in which UniFFI-generated enum and record types could feel more native in some target languages, support for methods would be a big improvement as well.

mhammond:
> The #[uniffi::constructor] macro can be attached to a function on a uniffi::Recrod

As implied by @jplatte, the actual bug here is that the attribute is allowed there.

Regarding the feature request though: IIUC, records do get constructors, with args exactly matching the fields. So the idea of supporting secondary constructors and methods makes sense, but I'm less clear about the primary constructor. Can you describe your use-case some more? Do you really want to avoid consumers being able to specify each field when creating the record?

Sajjon:
Yes it is manifold:
1) there is a bug, where I'm allowed to attach `#[uniffi::constructor]` to a Record without compilation error, but it is not exposed.
2) Yes I would like expose other than memberwise init in Swift (and Kotlin), i.e. constructors returning `Result` (throwing init in Swift).

(And yes ultimately I want methods on Records as well, but I thought that was harder to achieve than constructors)

mhammond:
> 2\. I would like expose other than memberwise init in Swift

Does that mean "instead of the default constructor" or "as well as the default ctor"?

Sajjon:
@mhammond alongside of default constructor - which I also would like to make internal instead of public as mentioned in another Matrix thread

mhammond:
The entire "constructor" concept here is a foreign one - so a "private primary constructor" doesn't really make sense - from the bindings POV, this would just mean "no constructor". So it sounds like you *do* want your constructor to be the primary and only constructor for the bindings?

[...]

Issue #1470, opened 2023-02-04
Let `#[Enum]` interfaces have methods


linabutler:
One thing that I've found super helpful when working with enums with associated data is being able to define getters and predicates on them. Here's an example inspired by https://github.com/mozilla/application-services/pull/5346#discussion_r1095272665:

```webidl
dictionary SuggestionDetails {
    string title;
    string url;
    string? icon;
};

// Each suggestion type has a common set of details, and additional fields specific to that type.
#[Enum]
interface Suggestion {
  Shopping(SuggestionDetails details, string full_keyword, string advertiser);
  Education(SuggestionDetails details, string full_keyword);
  Wikipedia(SuggestionDetails details);
  Weather(SuggestionDetails details, CurrentConditions current_conditions, Forecast forecast);
};
```

I'd love to be able to define methods on `Suggestion` like:
* `SuggestionDetails details()`, a getter that extracts the details from each variant.
* `boolean is_sponsored()`, a predicate that checks the suggestion type and returns whether it's sponsored (`Shopping`) or not (`Education`, `Wikipedia`, `Weather`).

UniFFI doesn't allow doing that yet, but @mhammond's follow-up comment (https://github.com/mozilla/application-services/pull/5346#discussion_r1095317470) inspired me to file a ticket to see if it could! 😊 As a workaround, we can add extensions in hand-written Kotlin and Swift files, but I wonder if we could push this down into UniFFI.

(It would be even nicer to make these `readonly attribute`s instead of methods—I think all our target languages support that? Python has `@property`, Kotlin has getters, Swift has computed properties, Ruby's getters are regular methods—but that might be too tricky).

Sajjon:
@linabutler nice! I also had the idea id of generaring computed properties in the land of Swift, but we probably need to be able to chose if a no arg method should be exported as computed property or method, since at least in Swift there is an informal rule that computed properties SHOULD be constant time.

PR #2706, opened 2025-10-22
Capture metadata for constructors and methods on Records and Enums.

mhammond:
Without this, methods on data classes fail to generate any bindings with obscure errors. With this patch the metadata for these items is collected, but not necessarily used.

This also implements methods, but not constructors, for Python, and nothing at all for Swift and Kotlin. But now the blockers to implementing all of this is in the templates and not in the core.

Towards #2704, #1935, #1470.

### DIFF
--- a/uniffi_bindgen/src/pipeline/initial/from_uniffi_meta.rs
+++ b/uniffi_bindgen/src/pipeline/initial/from_uniffi_meta.rs
@@ -211,6 +211,12 @@ impl UniffiMetaConverter {
                     list.into_values()
                         .map(|mut r| {
                             let key = (module_path.clone(), r.name.clone());
+                            if let Some(methods) = self.methods.remove(&key) {
+                                r.methods.extend(methods.into_values());
+                            }
+                            if let Some(constructors) = self.constructors.remove(&key) {
+                                r.constructors.extend(constructors.into_values())
+                            }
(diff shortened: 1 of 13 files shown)


uniffi_bindgen/src/pipeline/initial/from_uniffi_meta.rs
```diff
@@ -211,6 +211,19 @@ impl UniffiMetaConverter {
...
+                            if let Some(methods) = self.methods.remove(&key) {
+                                if self.trait_methods.contains_key(&key) {
+                                    bail!("{} contains both methods and trait methods", r.name)
+                                }
+                                r.methods.extend(methods.into_values());
+                            } else if let Some(trait_methods) = self.trait_methods.remove(&key) {
+                                r.methods.extend(Self::convert_trait_methods(
+                                    trait_methods.into_values().collect(),
+                                ));
+                            }
```
bendk, line 216: Should this case and the enum one be an error?  I think `trait_methods` should only be used for methods on a trait interface.

bendk (approved):
Awesome!  I'm excited for the new functionality.

uniffi_bindgen/src/pipeline/initial/from_uniffi_meta.rs
mhammond: yeah, oops - I just removed trait_methods entirely here rather than making it an error case

PR #2709, opened 2025-10-25
meta: Bump siphasher version.

emilio:
No API change but this prevents us from duplicating the dependency in mozilla-central.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1299,9 +1299,9 @@ dependencies = [
 
 [[package]]
 name = "siphasher"
-version = "0.3.11"
+version = "1.0.1"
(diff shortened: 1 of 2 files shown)


mhammond (approved):
lgtm - I assume you are yet to bump m-c though? Also, I'm not sure when we will do a new release for this, so it might not be that useful to you for a while. We can also merge it to 0.30 when times comes to put it into m-c, probably as 0.30.1. There are a few other dependency challenges there (at least toml, maybe more :( )

emilio:
https://phabricator.services.mozilla.com/D270085 is up, it patches siphasher 0.3 into 1.0 to avoid the duplication for now. When this gets to m-c you'll have to remove build/rust/siphasher

PR #2710, opened 2025-10-27
kotlin: use `Throwable.stackTraceToString` to show more info

flisky:


bendk:
This isn't a bad idea, but I'm slightly concerned about the extra string length.  How big to the strings get with all the stack trace info in them?  Also, what do you plan to do with this extra information?

flisky:
We're logging `UnexpectedUniFFICallbackError` while transforming it into our business error. Ultimately, the error logging will be sent to our APM system. It's hard to find root cause without backtrace. And we don't worry about the size too much, because developer experience is top priority, and backtrace is usually not-so huge.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
```diff
@@ -103,7 +103,7 @@ internal inline fun<T> uniffiTraitInterfaceCall(
         writeReturn(makeCall())
     } catch(e: kotlin.Exception) {
         callStatus.code = UNIFFI_CALL_UNEXPECTED_ERROR
-        callStatus.error_buf = {{ Type::String.borrow()|lower_fn }}(e.toString())
+        callStatus.error_buf = {{ Type::String.borrow()|lower_fn }}(e.stackTraceToString())
```
bendk: Can you wrap this call and the one below in another try..catch to handle the case where `stackTraceToString()` throws?  If that throws, then just store an empty `RustBuffer()` in `error_buf`.

bendk (changes requested):
Sounds good to me, let's just put a bit more defensive programming around `stackTraceToString()`.  It's a more complicated method, so it might be more likely to throw.

uniffi_bindgen/src/bindings/kotlin/templates/Helpers.kt
flisky: I don't think it's really necessary. `stackTraceToString` has no exception declaration, and in fact, it's just a simple utility method in [five lines]( https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/src/kotlin/util/Exceptions.kt). Kotlin stdlib wraps it for common.

bendk: I don't know, that function eventually calls [printStackTrace](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#printStackTrace--), which seems fairly complicated to me.  This code path shouldn't be hit often and any unexpected exception would probably cause a nasty crash when trying to unwind into the Rust portion of the stack.  It seems worth it to be extra cautious.  In hindsight,  I think we probably should have wrapped `toString()`  with a try...catch.

Issue #2624, opened 2025-08-22
Android Kotlin bindings throws "null RustBuffer had non-zero capacity"


typfel:
We've seen that our Kotlin bindings on Android sometimes throw "null RustBuffer had non-zero capacity" when calling an async function. Unfortunately we haven't found a way yet to reproduce the error since the stack trace doesn't give any hints on which code path we are executing:

```
Exception D.m: null RustBuffer had non-zero capacity
  at com.wire.crypto.uniffi.InternalException.<init> (InternalException.java:260)
  at com.wire.crypto.uniffi.Core_crypto_ffiKt.uniffiCheckCallStatus (core_crypto_ffi.kt:294)
  at com.wire.crypto.uniffi.Core_crypto_ffiKt.access$uniffiCheckCallStatus (core_crypto_ffi.kt:1)
  at com.wire.crypto.uniffi.Core_crypto_ffiKt.uniffiRustCallWithError (core_crypto_ffi.kt:279)
  at com.wire.crypto.uniffi.Core_crypto_ffiKt.uniffiRustCallAsync (core_crypto_ffi.kt:2169)
  at com.wire.crypto.uniffi.Core_crypto_ffiKt$uniffiRustCallAsync$1.invokeSuspend (core_crypto_ffi.kt:17)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:33)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:101)
  at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run (LimitedDispatcher.java:113)
  at kotlinx.coroutines.scheduling.TaskImpl.run (Tasks.kt:89)
  at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely (CoroutineScheduler.java:589)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask (CoroutineScheduler.kt:823)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker (CoroutineScheduler.kt:720)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run (CoroutineScheduler.kt:707)
```

We'll continue trying to find steps to reproduce this but maybe you have some hints how it's even possible to end up with a RustBuffer which is in this inconsistent state?

We are currently running uniffi version 0.29.2

bendk:
[...]

PR #2711, opened 2025-10-28
Kotlin: Call `write()` during `uniffiHandleError`

Patryk27:
Spotted over at
https://github.com/mozilla/uniffi-rs/issues/2624#issuecomment-3451575950.

I'm 50/50 on whether we actually need that call or not (not a JNA expert either), but we're already doing that for `uniffiHandleSuccess` and it cannot hurt, so.

Note that tests (`./docker/cargo-docker.sh test`) fail with:

```
test uniffi_foreign_language_testcase_test_py ... ok
/mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6/unary_result_alias.swift:360:46: error: unknown attribute 'unchecked'
fileprivate final class UniffiHandleMap<T>: @unchecked Sendable {
                                             ^
/mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6/unary_result_alias.swift:360:56: error: use of undeclared type 'Sendable'
fileprivate final class UniffiHandleMap<T>: @unchecked Sendable {
                                                       ^~~~~~~~
test uniffi_foreign_language_testcase_test_swift ... FAILED
test uniffi_foreign_language_testcase_test_kts ... ok

failures:

---- uniffi_foreign_language_testcase_test_swift stdout ----
Creating testing out_dir: /mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6
Error: running `swiftc` to compile bindings failed (cd "/mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6" && "swiftc" "-emit-module" "-module-name" "unary_result_alias" "-o" "libtestmod_unary_result_alias.so" "-emit-library" "-swift-version" "5" "-Xcc" "-fmodule-map-file=/mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6/unary_result_aliasFFI.modulemap" "-I" "/mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6" "-L" "/mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6" "-luniffi_unary_result_alias" "/mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6/unary_result_alias.swift")


failures:
    uniffi_foreign_language_testcase_test_swift
```

... but this doesn't seem related to the changes here 👀

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceImpl.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/CallbackInterfaceImpl.kt
@@ -56,15 +56,14 @@ internal object {{ trait_impl }} {
                 uniffiFutureCallback.callback(uniffiCallbackData, uniffiResult)
             }
             val uniffiHandleError = { callStatus: UniffiRustCallStatus.ByValue ->
-                uniffiFutureCallback.callback(
-                    uniffiCallbackData,
-                    {{ meth.foreign_future_ffi_result_struct().name()|ffi_struct_name }}.UniffiByValue(
-                        {%- if let Some(return_type) = meth.return_type() %}
-                        {{ return_type.into()|ffi_default_value }},
-                        {%- endif %}
-                        callStatus,
-                    ),
-                )
+                val uniffiResult = {{ meth.foreign_future_ffi_result_struct().name()|ffi_struct_name }}.UniffiByValue(
+                    {%- if let Some(return_type) = meth.return_type() %}
+                    {{ return_type.into()|ffi_default_value }},
+                    {%- endif %}
+                    callStatus,
+                );
+                uniffiResult.write()
+                uniffiFutureCallback.callback(uniffiCallbackData, uniffiResult)
             }
 
             {%- match meth.throws_type() %}


badboy:
> Note that tests (`./docker/cargo-docker.sh test`) fail with:

It seems we just never actually updated that script to use the newer Docker container.
CI is fine on this.
So yeah, unrelated to the changes.

bendk:
The code looks right to me, thanks.

Do you have any way to test these changes and see how it affects the error rate?  Ideally, I'd like to test them before we merged.  But I'd be okay with merging and then testing as well.  The thing I'd really like to avoid is merging this now, not knowing if it affects anything or not and forgetting about the work altogether.  A big reason for that is that if this does fix a bug, then we should audit the code and find other instances of missing `write()` calls.

Patryk27:
Makes sense -- I'll ask around if we can do A/B testing, I'll let you know tomorrow!

Patryk27:
Status report: got sick and didn't have chance to ask yet - we'll see next week 😅 🤒

Patryk27:
Status: we unfortunately can't carry out A/B testings yet, but we've rewritten all problematic APIs to be sync and - judging by Sentry logs - those don't crash anymore at all.

Since the codegen'ed sync APIs don't call `.write()` either (seems so?), I think the chances of that "missing" `.write()` being the culprit in async APIs are quite low, so I think it's better to close this - looks like a red herring more than a solution.

Issue #2712, opened 2025-10-29
Kotlin bindings can't be used in JVM unit tests on v0.30.0 when Cleaner is enabled


jmartinesp:
We just upgraded to `v0.30.0` in our project since we wanted to be able to use JNA direct mapping, and we just ran into this issue: as soon as we try to use any class/interface from the bindings, the `UniffiLib.init` function is executed, which tries to load the native library and we get a crash like:

```
Could not initialize class foo.Bar
java.lang.NoClassDefFoundError: Could not initialize class foo.UniffiLib
	at foo.Bar.<init>
    ....
```

I think this is related to having the JNA Cleaner enabled, as this cleaner is now implemented as:

```kotlin
internal object UniffiLib {
    
    // The Cleaner for the whole library
    internal val CLEANER: UniffiCleaner by lazy {
        UniffiCleaner.create()
    }
    ...
}
```

And when instantiating a class with a `NoHandle` value I see this being generated:

```kotlin
@Suppress("UNUSED_PARAMETER")
constructor(noHandle: NoHandle) {
    this.handle = 0
    this.cleanable = UniffiLib.CLEANER.register(this, UniffiCleanAction(handle))
}
```

Which will initialize the `UniffiLib` object.

We use these classes with their `NoHandle` constructor variant to be able to create fakes for those, and unit test classes that have references to them in our project.

The alternatives I see to fix this are:

1. Keep it as is, so nothing that has a reference to an FFI class can be unit tested, supporting only instrumentation tests.
2. Make `cleanable` optional, so we don't initialize it there and so we avoid accidentally initializing the `UniffiLib` object - I'm not sure this is the only point where this is happening though. Either that or having the lazy `Cleaner` instance outside the `UniffiLib`.
3. Come back to https://github.com/mozilla/uniffi-rs/issues/1352 and use interfaces for params and return types so there's no chance at all this could happen.

I hope the answer is not option 1 😅 .

jmartinesp:
In case you want a bit more context of how these fake FFI objects using `NoPointer/NoHandle` are used, [here](https://github.com/element-hq/element-x-android/pull/5625) is the PR that displayed the issue when trying to integrate some bindings built using `0.30.0`. All the ignored tests in there are affected because they use some `FakeFfi...` variant of a class in the bindings, which extends them using the `NoHandle` constructor.

mhammond:
[...]

PR #2713, opened 2025-10-29
Kotlin: make the `cleanable` field optional

jmartinesp:
This is needed so we can set it to `null` in the `NoHandle` constructors.

Works around the issue seen in https://github.com/mozilla/uniffi-rs/issues/2712 in my local tests, since it avoids initialising the native library through `UniffiLib`.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
@@ -139,7 +139,7 @@ open class {{ impl_class_name }}: Disposable, AutoCloseable, {{ interface_name }
     @Suppress("UNUSED_PARAMETER")
     constructor(noHandle: NoHandle) {
         this.handle = 0
-        this.cleanable = UniffiLib.CLEANER.register(this, UniffiCleanAction(handle))
+        this.cleanable = null
     }
 
     {%- match obj.primary_constructor() %}
@@ -155,7 +155,7 @@ open class {{ impl_class_name }}: Disposable, AutoCloseable, {{ interface_name }
     {%- endmatch %}
 
     protected val handle: Long
-    protected val cleanable: UniffiCleaner.Cleanable
+    protected val cleanable: UniffiCleaner.Cleanable?
 
     private val wasDestroyed = AtomicBoolean(false)
     private val callCounter = AtomicLong(1)
@@ -166,7 +166,7 @@ open class {{ impl_class_name }}: Disposable, AutoCloseable, {{ interface_name }
         if (this.wasDestroyed.compareAndSet(false, true)) {
             // This decrement always matches the initial count of 1 given at creation time.
             if (this.callCounter.decrementAndGet() == 0L) {
-                cleanable.clean()
+                cleanable?.clean()
             }
         }
     }
@@ -194,7 +194,7 @@ open class {{ impl_class_name }}: Disposable, AutoCloseable, {{ interface_name }
         } finally {
             // This decrement always matches the increment we performed above.
             if (this.callCounter.decrementAndGet() == 0L) {
-                cleanable.clean()
+                cleanable?.clean()
             }
         }
     }


mhammond:
This seems fine to me - @paxbun do you have any thoughts?

paxbun:
I also think it makes sense not to instantiate the cleaner for objects created with `NoHandle`, but I can't see any reason why `UniffiCleaner.create()` fails under that circumstance at a glance. There should be another exception logged under the `java.lang.NoClassDefFoundError: Could not initialize class foo.UniffiLib` error mentioned in #2712.

@jmartinesp By "JVM unit tests", did you mean unit tests that runs on the host machine? I see you mentioned instrumentation tests, and IIRC, there was a fallback logic that uses `UniffiJnaCleaner` if creating the Android cleaner fails.

jmartinesp:
> I also think it makes sense not to instantiate the cleaner for objects created with `NoHandle`, but I can't see any reason why `UniffiCleaner.create()` fails under that circumstance at a glance. 

`UniffiCleaner` doesn't fail, it's the `UniffiLib` object indirectly instantiated (since it's referenced) that fails, since it tries to load the native library in an environment where it's not available (JVM unit tests).

> @jmartinesp By "JVM unit tests", did you mean unit tests that runs on the host machine? I see you mentioned instrumentation tests, and IIRC, there was a fallback logic that uses `UniffiJnaCleaner` if creating the Android cleaner fails.

Yes, those. But again, the problem is not the cleaner itself, but it being part of an object that will fail to be initialized in this environment. Another possible solution would be to extract the cleaner object outside of `UniffiLib`.

paxbun:
Sorry for the confusion, I've just found gobley/gobley#82, and it seems that I haven't contributed this change to this repo as well. I still think not creating an object cleaner for objects created with `NoHandle` is okay, but how about adding a logic that creates a `UniffiJnaHandler` on Android when creating `AndroidSystemCleaner` still fails?

jmartinesp:
> Sorry for the confusion, I've just found [gobley/gobley#82](https://github.com/gobley/gobley/pull/82), and it seems that I haven't contributed this change to this repo as well. I still think not creating an object cleaner for objects created with `NoHandle` is okay, but how about adding a logic that creates a `UniffiJnaHandler` on Android when creating `AndroidSystemCleaner` still fails?

As I said above it's not the cleaner that fails to initialise. The problem is, calling this in the `NoHandle` constructor of all FFI classes:

```kotlin
constructor(noHandle: NoHandle) {
    this.handle = 0
    this.cleanable = UniffiLib.CLEANER.register(this, UniffiCleanAction(handle))
}
```

Will instantiate `object UniffiLib` since it was referenced, and that now does:

```kotlin
internal object UniffiLib {
    
    // The Cleaner for the whole library
    internal val CLEANER: UniffiCleaner by lazy {
        UniffiCleaner.create()
    }
    
    init {
        Native.register(UniffiLib::class.java, findLibraryName(componentName = "matrix_sdk_ffi"))
        ...
    }
}
```

And that `Native.register` call will fail because the library does not include the bindings for the host machine for these tests (and it shouldn't, AFAICT).

So the options are:

1. Not initializing the cleaner, which makes sense because it's not needed in a pure JVM context, and also prevents `UniffiLib` from calling `register`.
2. Moving `CLEANER` outside of `UniffiLib` so it's not initialized by simply being referenced. The cleanable would still be initialized, which isn't necessary, but the error is also fixed.

Or doing both. In this PR I implemented the first option.

paxbun:
Oh now I can see the exact reason, sorry for the confusion again (We're putting the desktop shared object to unit tests using `testImplementation(...)`, I overlooked it). Then not accessing `UniFFILib.CLEANER` by making the cleaner field nullable would be sufficient.

jmartinesp:
> Oh now I can see the exact reason, sorry for the confusion again (We're putting the desktop shared object to unit tests using `testImplementation(...)`, I overlooked it). Then not accessing `UniFFILib.CLEANER` by making the cleaner field nullable would be sufficient.

Ah, I see, in gobley you do have the desktop bindings, but in our case we only build them for Android, no wonder we didn't see the problem under the same perspective 😅 .

mhammond (approved):

PR #2715, opened 2025-11-03
TOML based renaming in the bindings.

### LINKED ISSUE #1426: Allow things to be renamed in the bindings config

### LINKED ISSUE #2427: [Feature Request] Proc macro to rename generated variant name
Running into issue where I create a color name in the Swift AssetCatalog and create the same color name in Rust, xcode spits out a unambiguous enum variant error because of the way swift enums can be initialized with just he variant name. 

I fixed it why just prefixing the variant name with a character. Ex:

``rust
#[derive(Debug, Copy, Clone, Serialize, Deserialize, Hash, Eq, PartialEq, uniffi::Enum)]
pub enum WalletColor {
    Red,
    Blue,
    Green,
    Yellow,
    Orange,
    Purple,
    Pink,
    CoolGray,
    Custom {
        r: u8,
        g: u8,
        b: u8,
    },

    // prefixed with W to avoid conflicts with swift asset color name
    WAlmostGray,
    WAlmostWhite,
    WBeige,
    WPastelBlue,
    WPastelNavy,
    WPastelRed,
    WPastelYellow,
    WLightMint,
    WPastelTeal,
    WLightPastelYellow,
}
```

But if i could get a rename attribute like `#[serde(rename]` I could do

```rust
#[derive(Debug, Copy, Clone, Serialize, Deserialize, Hash, Eq, PartialEq, uniffi::Enum)]
pub enum WalletColor {
    Red,
    Blue,
    Green,
    Yellow,
    Orange,
    Purple,
    Pink,
    CoolGray,
    Custom {
        r: u8,
        g: u8,
        b: u8,
    },

    // prefixed with W to avoid conflicts with swift asset color name
    #[uniffi(rename = "almostGray_")]
    AlmostGray,
    #[uniffi(rename = "almostWhite_")]
    AlmostWhite,
...
```

### LINKED ISSUE #2502: Support customising argument label in Swift
Is it possible to support custom argument label for traits on Swift? I am using a foreign trait and would like to customise the argument label for a particular trait that was exported. Thanks!

mhammond:
All builtin bindings support renaming almost all of the interface (types, args, items, variants, etc)
via TOML definitions.

Fixes #2427, fixes #1426, fixes #2502


uniffi_bindgen/src/pipeline/general/nodes.rs
```diff
@@ -112,15 +112,12 @@ pub enum CallableKind {
     /// Interface/Trait interface method
     Method { self_type: TypeNode },
     /// Interface constructor
-    Constructor {
-        interface_name: String,
-        primary: bool,
-    },
+    Constructor { self_type: TypeNode, primary: bool },
```
mhammond, line 115: I thought I needed these `self_type` changes - turns out I don't but I think it's an improvement wrt consistency?

docs/manual/src/renaming.md
```diff
@@ -0,0 +1,40 @@
+# Renaming
+
+UniFFI provides two ways to customize names in generated bindings:
+
+1. **Proc-macro attributes** - Use variants of `#[uniffi(name = "...")]` to rename items directly in Rust code. See [proc-macro renaming](proc_macro/renaming.md).
+2. **TOML configuration** - Define language-specific renames in `uniffi.toml`.
+
+## TOML-based Renaming
+
+You can rename types, functions, methods, constructors, and their members (fields, variants, arguments) on a per-language basis using `uniffi.toml`.
```
bendk, line 10: How does this work with primary constructors, which are often special cased and don't have names?  After skimming the code, I think it might not have any effect but I'm not sure.  Any behavior is fine with me for this corner case, but let's document it.

mkdocs.yml
```diff
@@ -104,6 +105,7 @@ nav:
 - Internals:
   - ./internals/design_principles.md
   - ./internals/crates.md
+  - ./glossary.md
```
bendk, line 108: Thanks, I think I forgot to add this

uniffi_bindgen/src/bindings/swift/mod.rs
```diff
@@ -75,6 +79,22 @@ impl BindingGenerator for SwiftBindingGenerator {
...
+                module_renames.insert(module_path, c.config.rename.clone());
+            }
+        }
+
+        // Apply rename configurations to all components
+        if !module_renames.is_empty() {
+            for c in &mut *components {
+                rename(&mut c.ci, &module_renames);
+            }
+        }
```
bendk: Do you need to update the `generate_swift_bindings` function below as well?  That's the one used for `uniffi-bindgen-swift` rather than `uniffi-bindgen -l swift`.

This makes me want to finish my work and update the code that uses `generate_external_bindings()` + `BindingGenerator` to use `BindgenLoader` instead.  It feels a bit weird to have this code in `update_component_configs`, which was supposed to be about mutating the `Config` objects, not the `ComponentInterfaces`.  I totally support you coding it this way, it just feels like one more instance where we're trying to add functionality to the process and the `BindingGenerator` trait is adding extra complications.

mhammond: >  It feels a bit weird to have this code in update_component_configs, which was supposed to be about mutating the Config objects

yeah, agreed. And we have `all_component_interfaces` - it's all messy, and I thought about cleaning it up, and ended up deciding to live with it :)

PR #2716, opened 2025-11-03
Be more selective about the tests we run with the uniffi/ffi-trace feature

mhammond:
Running the entire test suite takes a long time in CI with this feature.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,7 +101,7 @@ jobs:
                 --target wasm32-unknown-unknown \
                 --manifest-path "$manifest_path"
             done
-      - run: cargo test --features=uniffi/ffi-trace
+      - run: cargo test --features=uniffi/ffi-trace -p uniffi-fixture-coverall -p uniffi-fixture-futures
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -101,7 +101,7 @@ jobs:
                 --target wasm32-unknown-unknown \
                 --manifest-path "$manifest_path"
             done
-      - run: cargo test --features=uniffi/ffi-trace
+      - run: cargo test --features=uniffi/ffi-trace -p uniffi-fixture-coverall -p uniffi-fixture-futures
```
bendk, line 104: I thought this line was supposed to run all the tests.  I'd be fine removing the `ffi-trace` feature if that's slowing things down.

mhammond:
oh, oops, right - for some reason I thought this was *in addition* to the regular tests. And now that I look more closely, I don't really see the tests with this feature take a significantly longer time - a minute or so. For some reason I convinced myself these were significantly slower. So yeah, let's just kill this.

Issue #2717, opened 2025-11-03
Swift objects created with `init(noHandle:)` crash on deinit.


pixlwave:
Hello, we’ve just updated the Matrix Rust SDK to use 0.30.0 (from 0.28.0) however we’re having an issue using the Swift release of the SDK: We updated all of our mocks with the new `init(noHandle:)` (replacing `init(noPointer:)`) but they now crash with `EXC_BAD_ACCESS (code=1, address=0xfffffffffffffff0)` on deinit. It can be easily reproduced with the following snippet:

```
var mockInstance: Any? = SomeRustClass(noHandle: .init())
mockInstance = nil
```

<img width="918" height="78" alt="Image" src="https://github.com/user-attachments/assets/eea53abf-664c-4980-8603-212f96ce7477" />

Digging into the issue, we opened a PR that prevented this for `init(noPointer:)` (#2001) but it looks like the `guard let pointer` was removed in #2569 when switching to this new `handle` based system, resulting in no check before calling `uniffi_library_fn_free_object`.

@bendk: Hope you don't mind the ping but it appears you probably have the history on this to understand what the solution should be :)

mhammond:
Fixed by #2720.

PR #2720, opened 2025-11-03
Fix Swift mock object deinit and add a test for them (#2717)

bendk:


bendk:
@pixlwave sorry, I didn't understand what I was deleting when I updated this code.  Does this look right to you?

fixtures/swift-mock-objects/tests/bindings/test_mock_objects.swift
```diff
@@ -0,0 +1,33 @@
...
+    }
+}
+
+var mocked: TestObjectMock? = TestObjectMock()
+
+// Test that our mock object can override the functions of the real object
+assert(mocked!.isMock())
+
+// Test that we don't crash when deinitializing the mock object
+mocked = nil
```
pixlwave, line 33: If you're on Xcode 26, it might be nice to wrap this in an exit test assertion, so that your CI continues with a test failure if it does crash: https://developer.apple.com/documentation/testing/exit-testing

I think something like this would do it – but caveat, I haven't had the opportunity to try this myself yet.

```swift
#expect(processExitsWith: .success) {
    var mocked: TestObjectMock? = TestObjectMock()
    
    // Test that our mock object can override the functions of the real object
    assert(mocked!.isMock())

    // Test that we don't crash when deinitializing the mock object
    mocked = nil
}
```

bendk: Yeah, that would be nice, but I don't think we're on a new enough version of swift.  We're not even using XCode, just the CLI.  I think we're still using Swift 5 and if not we still want to support it, so for now let's stick with this version.

pixlwave: Ah my bad, I interpreted the lack of `XCTestCase` as meaning you were using the new Swift Testing framework, now I look again I realise this is using plain old `assert`s 🤦‍♂️

Sorry for the noise.

bendk: Yeah, we're in the stone age with these tests.  There's a bunch of comments saying we should switch to a real testing framework, but it's been working well enough that no one's put in the time.  No problem for the comment and thanks for the bug report.

PR #2723, opened 2025-11-07
Checksum refactor

bendk:
Added `calc_checksum_from_metadata` methods to several types. Removed `self_type` from the checksum calculation.  This simplifies the `BindgenLoader` code a bit.

Note: this commit should only be included `0.31.0` not a patch release, since the new checksums won't match the old ones.


mhammond:
probably want a CHANGELOG too

PR #2724, opened 2025-11-10
Methods on records for Swift and Kotlin.

mhammond:


.circleci/config.yml
```diff
@@ -52,7 +52,7 @@ jobs:
   Lint Rust with clippy:
     docker:
       - image: cimg/rust:1.77.1
-    resource_class: small
+    resource_class: large
```
mhammond, line 55: clippy in CI kept failing with SIGKILL, which presumably means OOM - let's see if this fixes it.

bendk: I've noticed this too, seems like a good fix.

PR #2726, opened 2025-11-12
Add `uniffi::generate` to generate bindings

bendk:
This is essentially a way to run `uniffi-bindgen generate` programmatically.

It uses `BindingsLoader` rather than `generate_bindings` and `BindingGenerator`. I think the new system is easier to extend, see https://github.com/mozilla/uniffi-rs/pull/2648 for details.


bendk:
I'm concerned that the new build model will fail for some users, so I tested this with the Mozilla application-services repo and everything worked okay (except for an unrelated issue with the code in `main`).  I don't know a good way to test this more, so I'm hitting the merge button.

Issue #2728, opened 2025-11-12
Mixing proc-macros and UDL can cause metadata conflicts.


bendk:
After https://github.com/mozilla/uniffi-rs/pull/2695, mixing proc macros and UDL can cause failures from metadata conflicts.  This happens because `uniffi_udl` uses `ComponentInterface::module_path()` for the module path, but that only contains the first component of the module path ([example](https://github.com/mozilla/uniffi-rs/blob/2f4dc25a7d2298da000820e3e32f91cbbb8db8e6/uniffi_udl/src/converters/mod.rs#L101))

I found two ways to break this, there may be more:
- Define a proc-macro type in a submodule and a struct in UDL that contains the proc-macro type
- Define a custom type in the UDL and put the `custom_type!` macro in a submodule.

PR #2730, opened 2025-11-13
Check for duplicate metadata in the pipeline code

bendk:
Fail when we see 2 metadata definitions with the same name, but different definitions. 

I made this while trying to repro #2728, before fully understanding the problem.  It doesn't help there, but I still think it's a good change.

### DIFF
--- a/uniffi_bindgen/src/pipeline/initial/from_uniffi_meta.rs
+++ b/uniffi_bindgen/src/pipeline/initial/from_uniffi_meta.rs
@@ -6,7 +6,10 @@
 
 use anyhow::{anyhow, bail, Context, Result};
 use camino::Utf8Path;
-use std::{collections::BTreeMap, fs};
+use std::{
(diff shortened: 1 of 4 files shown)


uniffi_bindgen/src/pipeline/initial/from_uniffi_meta.rs
```diff
@@ -43,6 +46,35 @@ pub struct UniffiMetaConverter {
     trait_impls: BTreeMap<(String, String), BTreeMap<uniffi_meta::Type, ObjectTraitImpl>>,
 }
 
+/// Utility trait used to insert metadata items into a BTreeMap, but bail on duplicates
+trait InsertIfEmpty<K, V> {
```
mhammond: name has slightly confusing semantics - maybe `InsertUnique`. Or even `Add`. But I don't care, whatever you decide.

PR #2731, opened 2025-11-13
Normalize module paths for `TypeUniverse`

### LINKED ISSUE #2728: Mixing proc-macros and UDL can cause metadata conflicts.

bendk:
If a type was referenced in both the UDL and proc-macro code, then the `TypeUniverse` code could fail because the module paths didn't match up. See the proc-macro fixture changes for an example of this.

To fix things, I made the TypeUniverse code normalize the `module_path` to be just the crate name.  This is unfortunate, but at least we can use full module paths in the pipeline code.

Fixes #2728.

### DIFF
--- a/fixtures/proc-macro/src/lib.rs
+++ b/fixtures/proc-macro/src/lib.rs
@@ -8,17 +8,21 @@ mod callback_interface;
 
 use callback_interface::TestCallbackInterface;
 
-#[derive(uniffi::Record)]
-pub struct One {
-    inner: i32,
-}
+// Test proc-macro definitions in a submodule (#2728)
+mod one {
+    #[derive(uniffi::Record)]
+    pub struct One {
+        pub inner: i32,
+    }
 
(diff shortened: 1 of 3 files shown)


bendk:
I tested this with application-services and it makes the builds work again.

PR #2733, opened 2025-11-14
Prevent completing foreign futures twice

bendk:
We've been seeing segfaults related to freeing the oneshot Sender instance on older mobile phones in Firefox Android. I'm not sure why this is, but my best guess is that we're calling the complete function twice, which will attempt to free the same arc reference twice.

My theory on how this happens is we call `handleSuccess` then an exception is thrown and we also call `handleError`. I haven't been able to reproduce this on my machine, except when I add manual throw statements.  However, it seems like it's theoretically possible and
maybe on older phones we're running into this.   For example the
following lines could throw after invoking the complete callback:

https://github.com/java-native-access/jna/blob/0deb54b46dc04f655e3c1d46e848fd26bf47c09a/src/com/sun/jna/Function.java#L489-L492

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/Async.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/Async.kt
@@ -52,9 +52,17 @@ internal inline fun<T> uniffiTraitInterfaceCallAsync(
     // Uniffi does its best to support structured concurrency across the FFI.
     // If the Rust future is dropped, `uniffiForeignFutureDroppedCallbackImpl` is called, which will cancel the Kotlin coroutine if it's still running.
     @OptIn(DelicateCoroutinesApi::class)
-    val job = GlobalScope.launch {
-        try {
-            handleSuccess(makeCall())
+    val job = GlobalScope.launch coroutineBlock@ {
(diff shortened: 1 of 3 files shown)


uniffi_bindgen/src/bindings/kotlin/templates/Async.kt
```diff
@@ -52,17 +52,27 @@ internal inline fun<T> uniffiTraitInterfaceCallAsync(
     // Uniffi does its best to support structured concurrency across the FFI.
     // If the Rust future is dropped, `uniffiForeignFutureDroppedCallbackImpl` is called, which will cancel the Kotlin coroutine if it's still running.
     @OptIn(DelicateCoroutinesApi::class)
-    val job = GlobalScope.launch {
-        try {
-            handleSuccess(makeCall())
+    val job = GlobalScope.launch coroutineBlock@ {
+        // Note: it's important we call either `handleSuccess` or `handleError` exactly once.  Each
+        // call consumes an Arc reference, which means there should be no possibility of a double
+        // call.  The following code is structured so that will will never call both `handleSuccess`
```
cyrilmottier:
```suggestion
        // call.  The following code is structured so that we will never call both `handleSuccess`
```

PR #2734, opened 2025-11-14
Prevent completing foreign futures twice

bendk:
Backporting this to the 0.29.x branch since this is what the Firefox code is currently using.  The plan is to make a 0.29.5 release and test if that reduces the crashes.

We've been seeing segfaults related to freeing the oneshot Sender instance on older mobile phones in Firefox Android. I'm not sure why this is, but my best guess is that we're calling the complete function twice, which will attempt to free the same arc reference twice.

My theory on how this happens is we call `handleSuccess` then an exception is thrown and we also call `handleError`. I haven't been able to reproduce this on my machine, except when I add manual throw statements.  However, it seems like it's theoretically possible and
maybe on older phones we're running into this.   For example the
following lines could throw after invoking the complete callback:

https://github.com/java-native-access/jna/blob/0deb54b46dc04f655e3c1d46e848fd26bf47c09a/src/com/sun/jna/Function.java#L489-L492

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ## [[UnreleasedUniFFIVersion]] (backend crates: [[UnreleasedBackendVersion]]) - (_[[ReleaseDate]]_)
 
+### What's fixed?
+
+- Prevented a potential segfault when completing foreign futures ([#2733](https://github.com/mozilla/uniffi-rs/pull/2733))
+
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.29.4...HEAD).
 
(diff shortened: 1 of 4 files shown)


bendk:
Should we also make a 0.30.1 release for this?  I'm on the fence.

bendk:
I also updated the CHANGELOG, which I should have done in the original PR.

mhammond:
> Should we also make a 0.30.1 release for this? I'm on the fence.

I'm a little on the fence too - it's never been reported, and I think we still hope to bump a-s and m-c to a new uniffi soon, and we said that would just jump to 0.31 there. If we think we might do that in the next few weeks I think we could skip it?

mhammond:
(not to mention that at this stage it is quite speculative!)

PR #2738, opened 2025-11-18
Reimplement support for --config on Python

### LINKED ISSUE #2703: Python: the additional config is not respected
**Steps to reproduce**

Run `uniffi-bindgen` for Python, supplying the path to an additional config file that does not exist.

```
uniffi-bindgen generate --language python --out-dir foo --config /path/to/the/file/that/doesnt/exist.toml --lib-file lib_mylib.so
```

**Expected result**

The `uniffi-bindgen` command fails.

**Actual result**

No errors are reported; the bindings are generated.

**Additional info**

The config file is not an input for the Python pipeline: https://github.com/mozilla/uniffi-rs/blob/7783937ffb028d07520f4c11d59b9689991d141d/uniffi/src/cli/uniffi_bindgen.rs#L340

bendk:
This flag's behavior is a bit weird in the current world of library mode and multiple crates, since it overrides the config for all crates.  We should probably update how it works at some point, but for now we can at least restore how it worked before.

Refactored the UDL/config path handling to help enable this.  I wanted to do this now to help enable future config flags, like one to specify different paths by crate.  One assumption the new code makes is that UDL data lives in a file, but I think that's safe to make.

Fixes #2703

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
 - Removed the `[Swift|Kotlin|Python|Ruby]BindingGenerator` types.  Use `uniffi::generate` instead
   to generate these bindings.
 
+### What's Deprecated?
+- `BindgenCrateConfigSupplier`.  Use the new `BindgenPaths` type instead.
+
 ### What's New?
 
 - Added the `uniffi::generate` function.  This implements the `uniffi-bindgen generate` command and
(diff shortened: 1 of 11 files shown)


bendk:
Tell me if I went too far with the config path changes.  I figure this is a good time to make them, but I could also implement using the existing system.

mhammond (approved):
so, um, you made a pipeline for the config? :) This is certainly an improvement, thanks.

bendk:
> so, um, you made a pipeline for the config? :) This is certainly an improvement, thanks.

Heh I guess so.  I was mostly trying to copy the tracing-subscriber API for this one, which is why I used "layer" for the name.

PR #2739, opened 2025-11-19
Methods on enums.

mhammond:


mhammond:
> Looks like non-flat python enums aren't supported. Was that an omission? Can you add tests for those as well?

heh, great question :) tl;dr - non-flat Python enums already work and have tests.

So in #2706 , I *thought* I added enum support for Python then realized it was *only* for non-flat enums - so existing tests are [here](https://github.com/mozilla/uniffi-rs/blob/main/fixtures/proc-macro/tests/bindings/test_proc_macro.py#L180). In #2724 I changed the changelog to avoid mentioning enum methods at all, then finally came back to this which fills in those pieces.

bendk (approved):
:facepalm: Oh yeah, I remember you mentioning that.  In that case, this one is perfect.

PR #2750, opened 2025-12-10
Apply the offset when picking data from a multi-arch file

badboy:
In a multi-arch file multiple archs are in the file back-to-back The Mach header tells us at which offset they start, but offsets _within_ that part are not global, but for that part only. Thus we need to pass on the file data from that offset on, otherwise we're looking at random other data.

Currently a Mach-O multi-arch file fails to properly parse:

    Error: finding components in 'your_library.dylib'

    Caused by:
        0: extracting metadata for '_UNIFFI_META_GLEAN_CORE_CUSTOM_TYPE_YOUR_TYPE'
        1: Unexpected metadata code: 209

I decided to also add a test that compiles enough C code on-the-fly to generate a library that UniFFI can parse. The test combines two of these libraries into a multi-arch file for parsing and calls the API.

---

The bug showed up as a build failure for Glean when we were trying to publish the universal2 file: https://bugzilla.mozilla.org/show_bug.cgi?id=2004934

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1867,6 +1867,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "uniffi-fixture-multi-arch-extract"
+version = "0.22.0"
+dependencies = [
+ "tempfile",
+ "uniffi_bindgen",
+ "xshell",
+]
+
+[[package]]
 name = "uniffi-fixture-proc-macro"
 version = "0.22.0"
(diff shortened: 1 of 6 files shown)


badboy:
I just realized that my test never actually runs on CI as we're not running on macOS.
I could instead add the compiled binary (with instructions how to reproduce it) to the repository directly. Thoughts?

PR #2753, opened 2025-12-14
Kotlin: Support record & enum serialization

### LINKED ISSUE #2751: Support for @Serializable in Kotlin
The Swift configuration supports `Codable` with [`generate_codable_conformance`](https://mozilla.github.io/uniffi-rs/next/swift/configuration.html) however the same for Kotlin's `Serializable` doesn't seem to exist. This is helpful to make JSON serializing easier.

Currently the generated bindings must [manually be patched](https://github.com/reown-com/yttrium/blob/4a82c029e00f473d459b9685acd9d7751a87ad58/generate_kotlin_locally.sh#L135) to add this annotation.

Would it be easy to add support for this?

cc @quetool

paxbun:
Fixes #2751.

To-Dos:

- [ ] Rename `fixtures/swift-codable` to `fixtures/serialization`?
- [ ] More options to customize class discriminator for enum variants?
- [ ] Any additional CI errors? What about `Dockerfile-build.checksum`?

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -49,8 +49,10 @@ test suite you will need:
 * Kotlin:
   * `kotlinc`, the [Kotlin command-line compiler](https://kotlinlang.org/docs/command-line.html).
   * `ktlint`, the [Kotlin linter used to format the generated bindings](https://ktlint.github.io/).
-  * The [Java Native Access](https://github.com/java-native-access/jna#download) JAR downloaded and its path
-    added to your `$CLASSPATH` environment variable.
+  * Several JARs downloaded and their path added to your `$CLASSPATH` environment variable:
+    * [Java Native Access](https://github.com/java-native-access/jna#download)
+    * [KotlinX Serialization Core runtime](https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-serialization-core-jvm/1.6.3/kotlinx-serialization-core-jvm-1.6.3.jar)
+    * [KotlinX Serialization JSON runtime](https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-serialization-json-jvm/1.6.3/kotlinx-serialization-json-jvm-1.6.3.jar)
 * Swift:
   * `swift` and `swiftc`, the [Swift command-line tools](https://swift.org/download/).
   * The Swift `Foundation` package.
--- a/fixtures/swift-codable/tests/bindings/test_codable.kts
+++ b/fixtures/swift-codable/tests/bindings/test_codable.kts
@@ -1,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import kotlinx.serialization.*
+import kotlinx.serialization.json.*
+import uniffi.codable_test.*
+
+val simpleRecord = SimpleRecord(
+    string = "Test",
+    boolean = true,
+    integer = 1,
+    floatVar = 1.1,
+    vec = listOf(true),
+)
+
+val jsonSimpleRecord = Json.encodeToString(simpleRecord)
+val deserializedSimpleRecord = Json.decodeFromString<SimpleRecord>(jsonSimpleRecord)
+assert(deserializedSimpleRecord.string == "Test")
+assert(deserializedSimpleRecord.boolean)
+assert(deserializedSimpleRecord.integer == 1)
+assert(deserializedSimpleRecord.floatVar == 1.1)
+assert(deserializedSimpleRecord.vec == listOf(true))
+
+// MultiLayerRecord
+val multilayer = MultiLayerRecord(
+    simpleEnum = SimpleEnum.ONE,
+    reprU8 = ReprU8.TWO,
+    simpleRecord = simpleRecord,
+)
+
+val jsonMultiLayerRecord = Json.encodeToString(multilayer)
+val deserializedMultiLayerRecord = Json.decodeFromString<MultiLayerRecord>(jsonMultiLayerRecord)
+assert(deserializedMultiLayerRecord.simpleEnum == SimpleEnum.ONE)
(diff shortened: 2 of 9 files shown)


fixtures/swift-codable/tests/bindings/test_codable.kts
```diff
@@ -0,0 +1,62 @@
...
+    floatVar = null,
+    vec = listOf("A", "B"),
+)
+
+val implicitNulls = Json { explicitNulls = false }
+val jsonRecordWithOptionals = implicitNulls.encodeToString(optionals)
+assert(jsonRecordWithOptionals == """{"vec":["A","B"]}""")
+
+// ComplexEnum
+val withClassDiscriminator = Json { classDiscriminator = "#class" }
```
paxbun, line 52: This will embed `"#class": "ComplexEnum.String"` into the resulting JSON string, but some may want to customize the value of that property as well. We can provide this via the `@kotlinx.serialization.SerialName` annotation, but I'm not sure there is an elegant way to customize it via the configuration file. I'm thinking of something like the custom type configuration, e.g.:

```toml
[bindings.kotlin.serialization.ComplexEnum.String]
serial_name = "complex_enum_string"
```

which will put `"#class": "complex_enum_string"` to the resulting string.

docs/contributing.md
```diff
@@ -49,8 +49,10 @@ test suite you will need:
 * Kotlin:
   * `kotlinc`, the [Kotlin command-line compiler](https://kotlinlang.org/docs/command-line.html).
   * `ktlint`, the [Kotlin linter used to format the generated bindings](https://ktlint.github.io/).
-  * The [Java Native Access](https://github.com/java-native-access/jna#download) JAR downloaded and its path
-    added to your `$CLASSPATH` environment variable.
+  * Several JARs downloaded and their path added to your `$CLASSPATH` environment variable:
+    * [Java Native Access](https://github.com/java-native-access/jna#download)
+    * [KotlinX Serialization Core runtime](https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-serialization-core-jvm/1.6.3/kotlinx-serialization-core-jvm-1.6.3.jar)
```
paxbun, line 54: The latest version of KotlinX Serialization is currently 1.9.0. Since the CI is also using KotlinX Coroutines 1.6.3, whose latest version is also 1.10.2, I chose a version released at the time, when Kotlin 2.0 had not been released yet.

Should we fix the Kotlin version and upgrade the dependencies here?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -86,6 +86,8 @@ pub struct Config {
     disable_java_cleaner: bool,
     #[serde(default)]
     pub(super) rename: toml::Table,
+    #[serde(default)]
+    generate_serializable_types: bool,
```
paxbun, line 90: Are there any better names for this?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -903,6 +905,78 @@ mod filters {
         let spaces = usize::try_from(*spaces).unwrap_or_default();
         Ok(textwrap::indent(&wrapped, &" ".repeat(spaces)))
     }
+
+    fn serializable_type(ty: &Type, ci: &ComponentInterface) -> Result<bool, askama::Error> {
```
paxbun, line 909: Would it be okay to make `ComponentInterface::iter_types_in_item` public and use it here?

docs/contributing.md
mhammond: That makes sense to me. Also, here I think it might be worth mentioning that the serialization libs are optional - eg, something like "Optionally, if you choose to use the `generate_serializable_types` option so that serializable types are generated, you will need ..."?

fixtures/swift-codable/tests/bindings/test_codable.kts
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
mhammond, line 1: I think it makes sense to rename this parent directory now? maybe something like "generated_serialization" or similar - this each test having a unique name would be ok (eg, "test_codable" for swift and "test_serialization" for kotlin.

 I'm not too bothered if you don't want to do this here or would prefer it be done in a followup etc.

(The name "codable" always gets me - I don't immediately think "serialization" when I see it :)

fixtures/swift-codable/tests/bindings/test_codable.kts
mhammond: Assuming you don't intend blocking this PR on that tweak, it might be worth opening a PR where we can discuss this, then add a short comment in the fixture itself which notes what you said and links to the PR, so people exploring this fixture also discovers this.

Re the idea for customizing, that seems good - another alternative might be a format similar to what we did for renaming - https://mozilla.github.io/uniffi-rs/next/renaming.html - which would be more like:

```
[bindings.kotlin.serialization.serial_names]
"ComplexEnum.String" = "complex_enum_string"
```

(I've no idea if that example makes actual sense in this context, but some degree of consistency makes some sense)

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
mhammond: maybe `generate_serializable_annotation`? That seems to better reflect what actually changes here (ie, it doesn't change how we render the type itself)?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
mhammond: yeah, that would be fine and better than copying it

mhammond (commented):
This is fantastic, thanks!

vvolkgang:
@paxbun was looking into what would be required to add kotlinx.serialization annotations and was surprised to see this PR open, you're always one step ahead! Anything we can do to help moving it forward?

PR #2754, opened 2025-12-15
Update default Rust version to 1.90.0

mhammond:
This caused quite a few warnings for `large_enum_variant` - these aren't entirely trivial to fix, but seem worth avoiding in the future, so I whitelisted all individually.


badboy:
> Rust version 0.90.0

ehehehe, good ol' times.

PR #2762, opened 2025-12-19
Update BindgenPathsLayer so it supplies the TOML directly rather than…

mhammond:
… a path

As Ben and I discussed - at least I think this is what we said 😅 

This has kinda made the name `BindgenPathsLayer` not ideal, but I think it's OK - the intent is still that it abstracts away the need for uniffi to know all the paths, even if this method isn't working directly with paths. Renaming might make sense though if you think?

This seems to work fine with gecko, although I've another issue there I think is unrelated.

### DIFF
--- a/uniffi_bindgen/src/bindgen_paths.rs
+++ b/uniffi_bindgen/src/bindgen_paths.rs
@@ -44,25 +44,14 @@ impl BindgenPaths {
         self.layers.push(Box::new(layer));
     }
 
-    /// Get the config path for a crate
-    pub fn get_config_path(&self, crate_name: &str) -> Option<Utf8PathBuf> {
-        self.layers
-            .iter()
-            .find_map(|l| l.get_config_path(crate_name))
-    }
-
     /// Get the config table for a crate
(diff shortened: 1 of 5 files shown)


uniffi_bindgen/src/bindgen_paths.rs
```diff
@@ -83,11 +72,11 @@ impl BindgenPaths {
 
 /// Trait for finding UDL and config paths
 pub trait BindgenPathsLayer {
-    /// Lookup the config TOML file path.
+    /// Lookup and load the config TOML for a crate.
     ///
-    /// This is usually `[crate-root]/uniffi.toml`
-    fn get_config_path(&self, _crate_name: &str) -> Option<Utf8PathBuf> {
-        None
+    /// This is usually loaded from `[crate-root]/uniffi.toml`
```
bendk: How about adding a comment here mentioning how flexible this is?  Something like: "However, layers are free to load the TOML table however they want.  For example, there could be a shared TOML file for all crates and this method could return one of the child values.

Issue #2752, opened 2025-12-11
Kotlin pointer-only FFI experiment


bendk:
We've been discussing an "FFI v2" for a few months now with an eye towards better performance on non-C languages, like Kotlin, Python, and external JS bindings.  Performance was the main goal, but we also speculated that it might improve stability on Kotlin by side-stepping some JNA issues (either in JNA itself or how we're using JNA).

However, we've recently started noticing a lot of Android crashes, especially when using the newer async code.  There's also #2740.  Before we implement the FFI v2, we should probably try to address the stability issues first.  That way we can design the FFI v2 around the same principles and have good confidence that it will also improve stability.  We should putting in a lot of work to implement a new FFI that's shiny and fast, but still has stability issues on Kotlin

As a starting point, I'd like to experiment with a pointer-only FFI.  Pointers seem like one of the safer types to pass with JNA since they're always exactly 1 machine word.  We can use them as a buffer to pack in other value types.  Here's what I'm thinking:

* Rename the `scaffolding-ffi-buffer-fns` feature `ffi-pointer` and use that for the user function scaffolding on Kotlin. `scaffolding-ffi-buffer-fns` was an experiment at a new FFI that was never adopted, but it seems great for this purpose.  I think `ffi-pointer` is a nicer name it has the same form as `ffi-v2`, which is the feature flag I'm hoping to use for the later one.
* Implement other functions, like `rust_buffer_alloc` using the pointer FFI.
* Prefix all the new FFI functions with `uniffi_pointer_`.

We probably shouldn't merge or release this code since the goal is experimentation.  Instead we can create a branch off of `0.27.3`, update the Firefox Android/application-services code to use that branch, then monitor our crash reports.  We can also re-introduce our async viaduct code, which we had to back out last time because of crashes, and see if it's better with this new system.  If this doesn't address crashes, then I guess we're back to the drawing board.  Hopefully we learn something at least.

If other teams also want to test out this code, that's great.  I'm not sure how hard it would be to rebase the changes on top of a different version, but it should be doable.

ncalexan:
Speaking for Firefox for Android, this seems like a sensible path forward.  In particular, I strongly support point b) [here](https://github.com/mozilla/uniffi-rs/issues/2740#issuecomment-3642505186), namely moving away from JNA.

[...]

PR #2765, opened 2025-12-20
Kotlin pointer ffi

bendk:
I tried to split this up into smaller commits, see them for notes on the changes and design decisions.

This is experimental code and we shouldn't merge this.  Instead, the plan is:
- Make sure CI passes
- Get the code approved
- Switch Firefox Android to use this branch directly (any other projects that want to try this can too).
- Send it out to the nightly channel and monitor the crash rates
- If we feel good about that, consider turning back on the async code that was causing high crashes before

If everything goes perfectly, this will eliminate the crashes and this can form the basis of a new FFI.  If not, then we'll have to tweak more until we do eliminate the crashes.  See #2752 for details.

### DIFF
--- a/uniffi/Cargo.toml
+++ b/uniffi/Cargo.toml
@@ -29,7 +29,7 @@ clap = { version = "4", default-features = false, features = [ "cargo", "derive"
 trybuild = "1"
 
 [features]
-default = ["cargo-metadata"]
+default = ["cargo-metadata", "pointer-ffi"]
 # Printout tracing information on FFI calls.  Useful for debugging issues with the bindings code.
 ffi-trace = ["uniffi_core/ffi-trace", "uniffi_bindgen?/ffi-trace"]
 # Support for features needed by the `build.rs` script. Enable this in your
--- a/uniffi_bindgen/Cargo.toml
+++ b/uniffi_bindgen/Cargo.toml
@@ -11,7 +11,7 @@ keywords = ["ffi", "bindgen"]
 readme = "../README.md"
 
 [features]
-default = ["cargo-metadata"]
+default = ["cargo-metadata", "pointer-ffi"]
 cargo-metadata = ["dep:cargo_metadata"]
 bindgen-tests = ["cargo-metadata", "dep:uniffi_testing"]
 ffi-trace = ["uniffi_testing?/ffi-trace"]
--- a/uniffi_core/Cargo.toml
+++ b/uniffi_core/Cargo.toml
@@ -19,7 +19,7 @@ once_cell = "1.10.0"
 static_assertions = "1.1.0"
 
 [features]
-default = []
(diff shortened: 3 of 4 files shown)


badboy:
> Switch Firefox Android to use this branch directly (any other projects that want to try this can too).

Doe this mean also switching over Glean to it (because Glean is shipped as part of Gecko)?

bendk:
> Doe this mean also switching over Glean to it (because Glean is shipped as part of Gecko)?

I'm hoping not.  Glean gets shipped in `libxul` and app-services is in `libmegazord`.  I think it should be okay for the 2 libraries to have different UniFFI code in them, but I'm not totally sure.  Can you think of any way this would go wrong?

bendk:
There's a ton of code in this one, but most of it is fairly mechanical.  However, there were a couple decisions that I wasn't sure about:

- Should I have tried harder to minimize the LOC changed?  I think I could have done that by implementing this more as a wrapper on top of the existing code.  For example, having the pointer-ffi functions call into the existing legacy FFI functions.  I figured we're going to want make the changes this sooner-or-later so we might as well do it with this PR.  However, if there's a bug in the new code that causes a crash, then it could make interpreting the crash reports harder.
- Passing the callback functions as a separate argument.  I'm pretty sure we could cast them to a `u64` and pack them into a buffer instead.  Actually, maybe we would always pack two `u64`s, one for the function pointer and one for the data argument.   I avoided this because I thought passing the function pointer as an extra argument would work better once we moved to JNI, but I'm not sure about that anymore.  We could easily define a JNI function than inputted a JVM method, and returned a pair of `u64`s for the method.  The other reason I avoided it was that casting function pointers to `u64`s felt weirder than casting data pointers, but I'm not sure how important that is in practice.

badboy:
> > Doe this mean also switching over Glean to it (because Glean is shipped as part of Gecko)?
> 
> I'm hoping not. Glean gets shipped in `libxul` and app-services is in `libmegazord`. I think it should be okay for the 2 libraries to have different UniFFI code in them, but I'm not totally sure. Can you think of any way this would go wrong?

Right. Then probably there's not much of an issue indeed.

bendk:
Updated this one to use the FFI buffers more:
  - Callbacks are now packed into the buffers
  - Instead of sending pointers to VTables, we pack the callbacks into the buffer and send that.  (I think the two actually have the same memory representation, but the code changed a bit).

This simplifies the FFI a lot because we're back to all FFI functions having the exact same signature `(*mut u8) -> ()`.

I also think this is better for the overall plan.  If all goes well and this does fix the stability issues, then I think we're all set to move to the new FFI with confidence.  With the last version, maybe we'd need to make these changes and push it out as a second iteration/experiment.  I could definitely be wrong about this, but my feeling is this is the right amount of changes for our first experiment.

PR #2766, opened 2025-12-23
HandleKind::Interface now carries an ObjectImpl, because JS cares wha…

mhammond:
…t the impl is

@bendk I think this is what we discussed - as you'll see in the patch, I think that either using `ObjectImpl` is wrong, or what this patch needs more work so that it always gets the precise variant in the places where today it doesn't know what it is.

Seems to work with js, although I've not actually tested it 😅

### DIFF
--- a/uniffi_bindgen/src/bindings/python/pipeline/nodes.rs
+++ b/uniffi_bindgen/src/bindings/python/pipeline/nodes.rs
@@ -617,7 +617,11 @@ pub enum HandleKind {
     RustFuture,
     ForeignFuture,
     ForeignFutureCallbackData,
-    Interface {
+    StructInterface {
+        module_name: String,
+        interface_name: String,
+    },
+    TraitInterface {
         module_name: String,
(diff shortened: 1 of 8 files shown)


mhammond:
> For that reason I like using `ObjectImpl` rather than only distinguishing interface vs trait.

But it seems like at least a couple of places in callback_interfaces.rs doesn't actually know whether it is a `::Trait` or a `CallbackTrait` - so languages which care about the distinction might get it wrong?

(At least It *think* both these variants use vtables in the same way?)

bendk:
> But it seems like at least a couple of places in callback_interfaces.rs doesn't actually know whether it is a `::Trait` or a `CallbackTrait` - so languages which care about the distinction might get it wrong?

Ahh, i didn't realize that.  In that case, I'm for a boolean.

mhammond:
> Ahh, i didn't realize that. In that case, I'm for a boolean.

I ended up just adding a new variant - wdyt?

PR #2767, opened 2025-12-23
Remove redundant toLong() conversion method call

rvandermeulen:
Fixes a new warning introduced in Kotlin 2.3. I haven't managed to test this locally, but AFAICT this addresses the code tripping up the warning.


bendk (commented):
Looks great to me, let's merge this.

PR #2777, opened 2026-01-08
Ignore any other doc annotation when trying to parse doc comments

### LINKED ISSUE #2776: Can't use `#[doc(hidden)]` with  #[uniffi::export(async_runtime = "tokio")]
The following error is emitted:

```
error: expected `=`
  --> crates/nym-vpn-lib-uniffi/src/environment.rs:47:10
   |
47 |     #[doc(hidden)]
   |          ^
```

Source:

```rs
impl Environment {
    // some code omitted for brevity

    #[doc(hidden)]
    pub async fn __stub_to_keep_compiler_happy(&self) {
        // Stub method to please compiler
    }
}
```

badboy:
Fixes #2776

---

This fixes the immediate issue.
Looking at the code we maybe might be able to change it further and completely remove the error case there.
I don't think it can even occur still?

### DIFF
--- a/uniffi_macros/src/util.rs
+++ b/uniffi_macros/src/util.rs
@@ -291,14 +291,19 @@ pub(crate) fn extract_docstring(attrs: &[Attribute]) -> syn::Result<String> {
     attrs
         .iter()
         .filter(|attr| attr.path().is_ident("doc"))
-        .map(|attr| {
-            let name_value = attr.meta.require_name_value()?;
+        .filter_map(|attr| {
+            let Ok(name_value) = attr.meta.require_name_value() else {
+                return None;
+            };
             if let Expr::Lit(expr) = &name_value.value {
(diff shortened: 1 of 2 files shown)


uniffi_macros/src/util.rs
```diff
@@ -291,14 +291,19 @@ pub(crate) fn extract_docstring(attrs: &[Attribute]) -> syn::Result<String> {
     attrs
         .iter()
         .filter(|attr| attr.path().is_ident("doc"))
-        .map(|attr| {
-            let name_value = attr.meta.require_name_value()?;
+        .flat_map(|attr| {
```
bendk: Nit: Would `filter_map` match this code better?

badboy: I'm like 95% sure that my brain was _reading_ this as filter map (hey, I wrote this code first thing in the morning!)

PR #2778, opened 2026-01-08
Count an async constructor as an async method for an async-exported impl

### LINKED ISSUE #2775: no async methods in this impl block
Creating a method with async constructors but no other async methods breaks `#[uniffi::export(async_runtime = "tokio")]` macros:

```
15 | #[uniffi::export(async_runtime = "tokio")]
   |                                  ^^^^^^^

error: could not compile `nym-vpn-lib-uniffi` (lib) due to 1 previous error

```

Something along those lines fails to compile:

```rs
#[derive(uniffi::Object)]
pub struct TestObject {
    value: u16
}

impl TestObject {
    #[uniffi::constructor]
    pub async fn new() -> Self {
      Self { value: 1 }
    }
}
```

Adding a blank async stub methods seems to please the compiler:

```rs
impl TestObject {
    #[uniffi::constructor]
    pub async fn new() -> Self {
      Self { value: 1 }
    }

    pub async fn __stub(&self) {
        // Stub method to please compiler
    }
}
```

badboy:
Fixes #2775

### DIFF
--- a/fixtures/futures/src/lib.rs
+++ b/fixtures/futures/src/lib.rs
@@ -499,4 +499,16 @@ async fn cancel_delay_using_trait(obj: Arc<dyn AsyncParser>, delay_ms: i32) {
     assert_eq!(future.await, Err(Aborted));
 }
 
+/// A test object with no (async) methods.
+#[derive(uniffi::Object)]
+pub struct TestObject;
+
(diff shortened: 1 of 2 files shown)


pronebird:
You're on the roll

PR #2783, opened 2026-01-11
Support renaming enum variant

Iskander508:
Allow renaming enum variants, example:
```
#[derive(uniffi::Enum)]
pub enum Enum {
    #[uniffi(name = "RenamedVariant")]
    VariantA,
    VariantB(Data),
}
```

### DIFF
--- a/fixtures/rename/src/lib.rs
+++ b/fixtures/rename/src/lib.rs
@@ -25,6 +25,7 @@ pub enum Enum {
 #[uniffi(name = "RenamedError")]
 pub enum MyError {
     #[error("Simple error")]
+    #[uniffi(name = "RenamedErrorVariant")]
     Simple,
 }
 
--- a/fixtures/rename/tests/test_rename.py
+++ b/fixtures/rename/tests/test_rename.py
@@ -26,7 +26,7 @@ class TestRename(unittest.TestCase):
         self.assertEqual(obj.renamed_method(), 123)
 
         # Test renamed error type exists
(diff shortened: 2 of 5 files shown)


Iskander508:
@mhammond Sorry, could you please help me a little?: Why the automated tests failed with my changes? I haven't touched the `uniffi-fixture-futures` tests. Thank you

Iskander508:
> That LGTM, thanks. The test failure is one that's known to be flaky and we really should fix it - I think it's unrelated and is likely to pass when run again.
> 
> Could you please add a CHANGELOG entry and some docs to https://github.com/mozilla/uniffi-rs/blob/main/docs/manual/src/proc_macro/renaming.md?

OK, thanks. I've added an example in the markdown. I hope it is self-explanatory like that.

mhammond:
It looks like you need to run `TRYBUILD=overwrite cargo test -p uniffi_uitests` to handle a change in the error messages for invalid attributes etc - it's not immediately obvious to me what the difference is, but it's probably expected and should become clear once you run that and commit the change it creates.

Iskander508:
> It looks like you need to run `TRYBUILD=overwrite cargo test -p uniffi_uitests` to handle a change in the error messages for invalid attributes etc - it's not immediately obvious to me what the difference is, but it's probably expected and should become clear once you run that and commit the change it creates.

Thank you. It pointed out a problem: A similar feature can apply to errors. So I also implemented it for error variants.

PR #2787, opened 2026-01-13
Improved pipeline mappings

### LINKED ISSUE #2780: Improve pipeline type system
The IR pipeline code has been working pretty well, but there are a few warts with how the types work:

- When fields are added, we initialize them to the default value then expect the pipeline code to overwrite this value with the "real" value. We're essentially moving back to weak typing for these with all the associated issues: what if you forget to set a field? What if you read a field in a pass before it's set? etc
- Removing fields is difficult, especially the use-case where you want to field data in the conversion pass, but then discard it.
- We need to define new versions of each node type, even if there's no changes between the IRs.  For example `ObjectImpl` doesn't change, but we need to define a 4 versions of it.
- Some of it is feels a bit magical. A simple rename isn't that bad, but changing an optional field to a non-optional field gets pretty weird.

We should update the pipeline system to address these issues.  One way to do that would be avoid trying to create our own dynamic typeing system and lean more into regular Rust types.  I believe we can do that by leveraging the following trait:

```
  pub trait MapNode<Output, Context> {
      fn map_node(self, context: &mut Context) -> Result<Output>
      where
          Self: Sized;
  }
```

`MapNode` can be used to map the root node from one IR to the root node of the next IR.  It also can be implemented on the child types then everything can be mapped recursively.  `map_node` inputs a `Context` type, which can be used to send data to descendant nodes when recursively mapping.  For example, it can store the current module name, self type, etc.

Since `MapNode` is a more-or-less normal Rust trait that operates on normal Rust types, we fix the issues listed above:
- No need to initialize fields to placeholder values
- Removing fields is no problem.  `map_node` will have access to the removed fields from the `Self` type.
- We can re-use node types from previous IRs like `uniffi_meta::ObjectImpl` if they haven't changed at all.
- The level of magic feels very low to me compared to all the other attempts.

Finally, I think `Node::visit()` is still going to be useful to visit all child nodes and I want to keep it, but let's make it do some type-checking.  If I call visit with a node type that can't possibly be reached, for example trying to visit `general::Callable` nodes from `initial::Root`, then it should return an error.

(See #2761 for an older attempt to solve this problem).

bendk:
The big idea is to make the mappings use the normal Rust typing system, rather then the semi-dynamic-typing system we were using before. See #2780 for a longer discussion of the reasoning behind this.

* We can now re-use types from previous IRs if they haven't changed. `initial::Type` is re-used by the general/python IRs and the `uniffi_meta::Radix` type is re-used for all of them.  Because of this, I added more methods to these types since we don't need to keep re-defining them.
* It's a lot easier to remove fields, so there are a few places where we do it earlier when it feels more natural.  For example, we now remove the `Function.arguments` field in the `general` pass, since it's redundant with `Callable.arguments`.
* The general pipeline no longer adds the `LiteralNode` and `FfiTypeNode` types.  These types didn't actually add any information, they were just done because future passes might want to use the transformation.  With the new system this is no longer needed, since we can always map a type later and add/remove fields without the issues we had before.
* Renamed the Python node `Namespace` to `Module`, which makes more sense with Python
* Kept the `Node` trait and the `visit()` methods.  At first I thought we would get rid of these, but they were still very useful.

Fixes #2780


bendk:
This is a huge PR, sorry for that but I don't know how to make a change like with without re-implementing the entire bindgen process.  For what it's worth, the line count is very large but a good deal of the changes are minor refactors.

I'd recommend starting with the documentation changes and focusing on the overall flow of the new system.

bendk:
I'll also just note for fun that this is moving us away from a "nano-pass" based pipeline to more of a query-based pipeline.  The next step is adding [salsa](https://crates.io/crates/salsa) to the mix to get incremental compiles and LSP support :rofl:

docs/manual/src/internals/bindings_ir_pipeline.md
```diff
@@ -20,161 +20,194 @@ Types inside an IR are called "nodes" and they must derive a `Node` trait impl.
...
+* `Node::has_descendant()` tests if a predicate is true for any descendant.
 
-## Node conversions
+These are useful when you want to populate a node field using it's descendants.  For example,
+building up the FFI definitions by visiting all functions/methods inside a namespace.
 
-`Node::try_from_node` attempts to convert a node from one IR to the corresponding node from another IR.
-For example, `from_ir::Node` will be converted to `into_ir::Node` using the following rules:
+Node methods input a closure that inputs a node type, for example `root_node.visit(|func: Function| ...)`.
+`visit()` will walk the node tree and call that closure with all matching nodes.  The methods panic
```
mhammond, line 32: I haven't looked at the code yet, but the sentence about panicing seems a little odd - it implies trying to visit all enums in a namespace would panic if no enums were used?

docs/manual/src/internals/bindings_ir_pipeline.md
```diff
@@ -20,161 +20,194 @@ Types inside an IR are called "nodes" and they must derive a `Node` trait impl.
...
-    name_from_previous_pass: Option<String>
-    /// This field will be a non-optional version of `SourceNode::name`.
-    /// It will be initialized to an empty string, then a pipeline pass will populate it using
-    /// `name_from_previous_pass` combined with logic to handle the `None` case.
+#[derive(Node, MapNode)]
+// Use `#[map_node(from([type_name]))]` to declare the type we're mapping from
+#[map_node(from(prev_ir::Namespace))]
+// Use `#[map_node(update_context([expr]))]` update the context for descendant `map_node()` calls.
+#[map_node(update_context(context.set_namespace_name(&self.name)))]
+pub struct Namespace {
```
mhammond, line 81: again, yet to look at the code, but at face value, it's not clear this magic is worthwhile, and that your explicitly spelled out example above is clearer and doesn't add too many extra lines of code. I guess I'll see as I look at the impl though.

Well, thought I'd see for myself - I implemented it for `Record` and ended up with:
```
impl MapNode<Record, Context> for initial::Record {
    fn map_node(self, context: &Context) -> Result<Record> {
        let mut context: Context = context.clone();
        context.update_from_record(&self)?;
        Ok(Record {
            fields_kind: records::fields_kind(&self.fields),
            self_type: context.self_type()?,
            name: rename::type_(&context.namespace_name()?, self.name, &context)?,
            uniffi_trait_methods: self.uniffi_traits.map_node(&context)?,
            fields: self.fields.into_iter().map(|f| f.map_node(&context)).collect::<Result<_>>()?,
            constructors: self.constructors.into_iter().map(|c| c.map_node(&context)).collect::<Result<_>>()?,
            methods: self.methods.into_iter().map(|m| m.map_node(&context)).collect::<Result<_>>()?,
            docstring: self.docstring,
        })
    }
}
```

which isn't *too* bad, but I must concede it is tedious. So I think what you came up with is good and anyone can choose to use it or not.

docs/manual/src/internals/bindings_ir_pipeline.md
```diff
@@ -20,161 +20,194 @@ Types inside an IR are called "nodes" and they must derive a `Node` trait impl.
...
+    // Use `#[map_node(from)]` on variants/fields when they've been renamed from the previous IR
+    #[map_node(from(Object))]
+    Interface {
+        #[map_node(from(prev_name_field))]
+        name: String,
+        // ...
+    },
+    // Use `#[map_node(added)]` for variants that have been added in this IR.  The generated code
+    // will not try to map these variants since they didn't exist in the previous IR
+    #[map_node(added)]
```
mhammond: kinda like my comment above, it would be interesting to know how this would look if not automatically derived.

docs/manual/src/internals/bindings_ir_pipeline.md
```diff
@@ -20,161 +20,194 @@ Types inside an IR are called "nodes" and they must derive a `Node` trait impl.
...
+for the next pass. This works because the `MapNode` impl is specific to the `Context` type for the
+pass.
+
+## Module structure
+
+Each IR will typically have a module dedicated to them with the following structure:
+
+* `mod.rs` -- Top-level module.
+* `nodes.rs` -- Node definitions with `MapNode` derives.
+* `context.rs` -- Defines the `Context` struct
```
mhammond, line 180: Until I got here, I assumed there might be many Context structs per IR, some dedicated to specific mapping requirements. Maybe somewhere above might mention there's typically one `Context`? (Still not yet at the code, so maybe this is wrong!)

uniffi_bindgen/src/pipeline/general/nodes.rs
```diff
@@ -39,63 +47,57 @@ pub struct Namespace {
...
 }
 
-#[derive(Debug, Clone, Node)]
+#[derive(Debug, Clone, Node, MapNode)]
+#[map_node(from(initial::Function))]
 pub struct Function {
-    pub name: String,
+    #[map_node(callable::function_callable(&self, context)?)]
     pub callable: Callable,
-    pub is_async: bool,
```
mhammond, line 46: huh, I like the fact all these redundant fields are gone! But why does docstring not live in callable?

uniffi_bindgen/src/pipeline/general/namespaces.rs
```diff
@@ -0,0 +1,49 @@
...
+                callback_interfaces::ffi_definitions(&namespace, context)?,
+                rust_future::ffi_definitions(&namespace)?,
+            ]
+            .into_iter()
+            .flatten(),
+        )),
+        checksums: checksums::checksums(&namespace)?,
+        type_definitions: sort::sort_type_definitions(
+            [
+                type_definitions_from_api::type_definitions(&namespace, context)?,
```
mhammond: I don't really understand this, and it relates to my comment above about the magic I don't quite grok. On one hand, it appears that `type_definitions_from_api::type_definitions` is doing the mapping of the `TypeDefinition`s - the comment there even says "// Map types to type definitions to add" and does add all types - which makes sense, but I don't understand why we *also* call `map_node()` on them here too - ie, I don't really understand why `TypeDefinition` needs to implement `MapNode` given what `type_definitions()` does.

uniffi_bindgen/src/pipeline/general/mod.rs
```diff
@@ -4,60 +4,37 @@
...
+pub use context::Context;
+pub use indexmap::{IndexMap, IndexSet};
 pub use nodes::*;
-use uniffi_pipeline::{new_pipeline, Node, Pipeline};
+use uniffi_pipeline::{new_pipeline, use_prev_node, MapNode, Node, Pipeline};
 
 /// General IR pipeline
 ///
 /// This is the shared beginning for all bindings pipelines.
 /// Bindings generators will add language-specific passes to this.
```
mhammond: this comment seems like it should maybe be updated? It's more like "they define their own pipeline based on the output of this one" (which I think was true before, right?)

docs/manual/src/internals/bindings_ir_pipeline.md
```diff
@@ -20,161 +20,194 @@ Types inside an IR are called "nodes" and they must derive a `Node` trait impl.
...
-    /// `into_ir::Node` wraps `from_ir::WrappedNode`.
-    /// All `from_ir::WrappedNode` values will be replaced with `Node { wrapped: wrapped_node }` values.
-    /// Any other fields will be initialized to their default values.
-    ///
-    /// This is mostly used as a way to add fields to enum types.
-    /// Wrap the enum with a struct, then add fields to that struct.
-    /// For example, `TypeNode` wraps the `Type` enum from `uniffi_meta`.
-    #[node(wraps)]
-    wrapped: WrappedNode,
+impl MapNode<Namespace, Context> for prev_ir::Namespace {
```
mhammond: How is this going to work when `prev_ir` is in a different crate?

mhammond, line 53: I don't get how this will work for pipelines in a different crate, eg, js - isn't this requiring the js crate to implement a trait defined externally on a type defined externally?

docs/manual/src/internals/bindings_ir_pipeline.md
mhammond: oops, also meant to note that it wasn't obvious every map_node call ends up with its own cloned context. I'll maybe try and suggest an edit here to make all that clearer.

docs/manual/src/internals/bindings_ir_pipeline.md
bendk: It checks the types not the values, so if you have `Vec<Enum>` the check will pass even if the vec is empty.  I added some more comments to make this clear.

docs/manual/src/internals/bindings_ir_pipeline.md
bendk: I went back and forth on this one.  The thing I like most about this system is that you can read `nodes.rs` and generally see how things gets mapped.  I had one version that had a `map_node.rs` module that did everything and it felt a bit harder to understand what's going on.  I could be convinced to go the manual road though.

docs/manual/src/internals/bindings_ir_pipeline.md
bendk: Edit's are welcome.  FYI, we only clone the context if there's an `update_context` function.

uniffi_bindgen/src/pipeline/general/nodes.rs
bendk: It probably should, maybe we can do that in another pass?  I was trying to minimize how much this PR changed the IR structs though.

uniffi_bindgen/src/pipeline/general/namespaces.rs
bendk: It's a weird distinction.  I feel like the module/function names could be improved, but I'm not sure what names to use.  In the meantime, I added comments to explain things.

docs/manual/src/internals/bindings_ir_pipeline.md
bendk: It's actually allowed by the orphan rule.  If `prev_ir` doesn't define an impl for `prev_ir::Namespace` then the crate that defines `Namespace` can.  I'm pretty sure this is already working and tested, since we're mapping the `uniffi_meta` types in `uniffi_bindgen`.

mhammond: huh!

docs/manual/src/internals/bindings_ir_pipeline.md
mhammond: yeah, I think this is fine and the correct tradeoff. I like that the docs spell it out like this though.

bendk:
Looks like we've got general approval for this one, I'm going to merge it.

PR #2789, opened 2026-01-15
Mutable byte slice support

### LINKED ISSUE #1974: Ability to share data across FFI without a copy
We've been exploring using UniFFI for our updated version of a core library to make things easier on us than using straight C FFI code & hand written wrappers. Problem is we have some parts of the code where we'd want to minimize overhead due to low latency video use cases. So ideally when sharing something like an object (interface struct with a `Vec<u8>` inside for example) we'd want to be able to get the ptr/len to let the foreign swift/etc code do the final copy, avoiding the copy happening at FFI level as well.

ladvoc:
Still WIP and incomplete.

Closes #1974


skeet70:
Was this superseded by  https://github.com/mozilla/uniffi-rs/issues/2864 and its PRs?

mhammond:
> Was this superseded by https://github.com/mozilla/uniffi-rs/issues/2864 and its PRs?

Yeah, I think we might as well close this one.

PR #2792, opened 2026-01-16
Fixes for renaming types inside submodules

bendk:


mhammond:
oh, and seeing you need to fix a merge conflict, maybe a changelog would be good too!

Issue #2686, opened 2025-10-04
New bindgen test fixtures


bendk:
Our test fixtures are nice because they're fairly complete, but I think there's a better version to be had.  I've been experimenting with some [new fixtures in uniffi-bindgen-gecko-js](https://searchfox.org/firefox-main/source/toolkit/components/uniffi-bindgen-gecko-js/uniffi_bindings_tests) and I think we should move those into the main tree.

The new tests exist in 2 crates: a main testing crate and a second crate to test external types.  The main crate has a bunch of features that bindings authors can enable one-by-one.  Each feature enables a single module that focuses on a specific UniFFI feature.  Features start very basic, like [calling a function](https://searchfox.org/firefox-main/source/toolkit/components/uniffi-bindgen-gecko-js/uniffi_bindings_tests/src/simple_fns.rs) and [handling primitive types](https://searchfox.org/firefox-main/source/toolkit/components/uniffi-bindgen-gecko-js/uniffi_bindings_tests/src/primitive_types.rs) and extend to complex features like [trait interfaces](https://searchfox.org/firefox-main/source/toolkit/components/uniffi-bindgen-gecko-js/uniffi_bindings_tests/src/trait_interfaces.rs).

Tests are not intended to represent real-world scenarios.  Instead they try to exercise  different requirements in a logical order with the least number of cases needed.  For example, the [records test](https://searchfox.org/firefox-main/source/toolkit/components/uniffi-bindgen-gecko-js/uniffi_bindings_tests/src/records.rs) has a case for with lifting/lowering records with primitive values, then a case for lifting/lowering nested records.

These tests aren't perfect, but I like the overall structure.  It found nice for developing uniffi-bindgen-gecko-js for a few reasons:
  - You can build up your bindgen code step-by-step.  Start by enabling the `simple_fns` feature and test that you can make rust calls.  Then move to primitive types, records, and keep going until you have all the features covered.
  - It's more obvious why things fail.  Since each test is very specific, you can more easily know what failed.
  - There's less crates involved.  I don't need a dozen dev-dependencies (and we don't need to publish that many).

If we switched to this system, I don't think these replace all our fixtures/examples.  We'd probably still want:
 - Fixtures to test the scaffolding code (maybe the existing fixtures could fill this role)
 - Examples, since they serve as a form of documentation and a form of end-to-end testing
 - Language specific tests to target particular bugs or features like `swift-omit-labels`

PR #2793, opened 2026-01-16
New bindgen test fixtures

bendk:
Added new test fixture library as described in #2686 and implemented Swift tests for it.  We've been using these tests for a while for uniffi-bindgen-gecko-js and they've been working pretty well.

Moved Swift-specific tests into subdirectories.  I like this organization better for a few reasons.  The biggest one is that it matches how an external bindings generator would have them.

This only adds tests for Swift and leaves the old example/fixture tests in place.  If we decide to go on this path, I think the plan should be:
* Add tests for the other languages
* Add newly created tests to `bindgen-tests` rather than the fixtures/examples.
* Continue to use the fixtures and examples, since they may cover things not covered by these tests.  If one of those tests picks up an issue that these miss, we should update `bindgen-tests` to cover it.
* Eventually, we can decide that the fixtures/examples aren't adding value and delete them.


bendk:
How do others feel about merging these and following the plan from the description?  I think the tests are good, the main question is if that plan to move away from the current fixtures and towards these tests is the right way to do it.

@mhammond @badboy

mhammond:
I agree with the approach in general, so I'm fine with that. My only real fear is that we never actually move away from fixtures, so now we've just got one extra thing and no fewer things, but that's on us to be aggressive.

badboy:
I don't _fully_ understand what makes this better over the current structure (so far it's mostly moving files?), howerver I'm not blocking it.
I would like to see some docs (possibly in the readme of that folder), which describes

1) when tests should go in there vs a fixture
2) how to run the tests / how to run an individual test

I'm ok if this lands in a followup though.

bendk:
Thanks for the feedback, I definitely agree.  I added a README with some more explanation.  I also agree we don't want to get stuck in limbo for too long, I can commit to trying to keep moving this forward.

bindgen-tests/README.md
```diff
@@ -0,0 +1,59 @@
...
+* Language-specific tests get their own directory.
+  For example, the Swift-specific tests all live in the `swift/` subdirectory.
+  If we decide to split `uniffi-bindgen-swift` into it's own crate,
+  this will make it easier for the tests to move there.
+
+## Running the tests
+
+* Go to one of the language directories (for example `bindgen-tests/swift`)
+* `./cargo test`
+* ...or something like `./cargo test test_records` to test a specific feature
```
badboy: This should need `./`


```suggestion
* `cargo test`
* ...or something like `cargo test test_records` to test a specific feature
```

badboy (approved):

PR #2795, opened 2026-01-17
Replace `java.time` with `kotlin.time` for duration and instant

yukibtc:
Switch from `java.time.Duration` and `java.time.Instant` to `kotlin.time.Duration` and `kotlin.time.Instant` to avoid requiring Android API 26+.

Ref https://github.com/rust-nostr/nostr-sdk-ffi/issues/66

### DIFF
--- a/fixtures/uniffi-fixture-time/tests/bindings/test_chronological.kts
+++ b/fixtures/uniffi-fixture-time/tests/bindings/test_chronological.kts
@@ -1,36 +1,37 @@
 import uniffi.chronological.*;
-import java.time.Duration
-import java.time.Instant
-import java.time.DateTimeException
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.nanoseconds
+import kotlin.time.Duration.Companion.seconds
+import kotlin.time.Instant
 
 // Test passing timestamp and duration while returning timestamp
-assert(add(Instant.ofEpochSecond(100, 100), Duration.ofSeconds(1, 1))
-        .equals(Instant.ofEpochSecond(101, 101)))
+assert(add(Instant.fromEpochSeconds(100, 100), 1.seconds + 1.nanoseconds)
+        .equals(Instant.fromEpochSeconds(101, 101)))
 
 // Test passing timestamp while returning duration
-assert(diff(Instant.ofEpochSecond(101, 101), Instant.ofEpochSecond(100, 100))
-        .equals(Duration.ofSeconds(1, 1)))
+assert(diff(Instant.fromEpochSeconds(101, 101), Instant.fromEpochSeconds(100, 100))
+        .equals(1.seconds + 1.nanoseconds))
 
 // Test pre-epoch timestamps
-assert(add(Instant.parse("1955-11-05T00:06:00.283000001Z"), Duration.ofSeconds(1, 1))
+assert(add(Instant.parse("1955-11-05T00:06:00.283000001Z"), 1.seconds + 1.nanoseconds)
         .equals(Instant.parse("1955-11-05T00:06:01.283000002Z")))
 
 // Test exceptions are propagated
 try {
-        diff(Instant.ofEpochSecond(100), Instant.ofEpochSecond(101))
+        diff(Instant.fromEpochSeconds(100), Instant.fromEpochSeconds(101))
         throw RuntimeException("Should have thrown a TimeDiffError exception!")
 } catch (e: ChronologicalException) {
         // It's okay!
 }
 
 // Test max Instant upper bound
-assert(add(Instant.MAX, Duration.ofSeconds(0)).equals(Instant.MAX))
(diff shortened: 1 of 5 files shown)


paxbun:
Hi! Thanks for the contribution. Since `kotlin.time.Instant` requires Kotlin 2.3+, we at least need to add a bindgen option to control whether to use it.

Plus, the CI still uses an older version of Kotlin; we need to update it.

@mhammond Could you have a look at this?

mhammond:
> requires Kotlin 2.3+, we at least need to add a bindgen option to control whether to use it.

Another option might be to just declare we no longer support old versions of Kotlin - how likely is it that people might be on the latest uniffi but be stuck on a kotlin that old?

(I'm not advocating for that particularly, but in general I don't see a problem with bumping the min supported versions of things over time)

mhammond:
> Another option might be to just declare we no longer support old versions of Kotlin

Now I look at this, what I wrote might be fine in the general case, but it looks like this change would break existing kotlin code, which we typically try hard to avoid - so I think a new option would be good.

OTOH though, I've actually been thinking of deprecating duration and instant - custom types offer far more flexibility and offers the same capability - is there a reason to not do that instead? ie, why wouldn't you just use a custom type?

paxbun:
> how likely is it that people might be on the latest uniffi but be stuck on a kotlin that old?

Kotlin 2.3 is the latest release currently, and even a Mozilla project like [mozilla/application-services](https://github.com/mozilla/application-services/blob/af728a890e7f5fd53ae2bc9b06c1a7ff3e688afa/gradle/libs.versions.toml#L15) still uses Kotlin 2.2. In my experience, upgrading from Kotlin 2.x to 2.y doesn't cause any difficulties, but those stuck on 1.9 will encounter many issues when migrating to Kotlin 2.x. Using a library built with a higher version of Kotlin is typically unavailable in a lower version of Kotlin due to metadata incompatibility. They sometimes need to update the IDE as well, due to Android Gradle plugin issues.

Spring Boot 4 was released in Nov 2025 and requires Kotlin 2.2, but Spring Boot 3 needs Kotlin 1.7.

So I think there might be some projects with many reasons not to upgrade to Kotlin 2.x. If we're going to bump the minimum supported version of Kotlin, I believe that shouldn't be higher than 1.9.

> is there a reason to not do that instead? ie, why wouldn't you just use a custom type?

I believe many Java and Kotlin programmers want to mix their Rust code with Java's locale and timezone-aware Calendar and DateTime APIs, where the corresponding libraries in the Rust world are not mature enough.

mhammond:
> > is there a reason to not do that instead? ie, why wouldn't you just use a custom type?
> 
> I believe many Java and Kotlin programmers want to mix their Rust code with Java's locale and timezone-aware Calendar and DateTime APIs, where the corresponding libraries in the Rust world are not mature enough.

I'm suggesting a custom type would allow you do exactly that - iirc, currently a duration is basically a tuple of (u64, u32), which both sides convert appropriately. I think a custom type could do exactly that - but you'd have more control over the semantics rather than using the semantics we chose.

To put it another way, I don't think we'd add timestamp and duration today given we have custom types - instead, we'd be more likely to document a complete solution using custom types.

paxbun:
> I'm suggesting a custom type would allow you do exactly that - iirc, currently a duration is basically a tuple of (u64, u32),

Oh, now I get the point. You're right that a duration is just a tuple of a long and an int value.

I actually just read the code changes, not rust-nostr/nostr-sdk-ffi#66; now I've just read it, and I think that shouldn't be fixed with this.

@yukibtc AGP provides a feature called [desugaring](https://developer.android.com/studio/write/java8-support-table), and it provides a compatibility library, [google/desugar_jdk_libs](https://github.com/google/desugar_jdk_libs), which contains an implementation of older Java APIs that is not natively supported by the OS. It also provides `java.time.Duration`, so the issue rust-nostr/nostr-sdk-ffi#66 will be okay with it.

yukibtc:
Sorry for the late reply.

> Since kotlin.time.Instant requires Kotlin 2.3+

I missed that requirement, thanks for pointing it out.

> custom types offer far more flexibility and offers the same capability - is there a reason to not do that instead? ie, why wouldn't you just use a custom type?

I initially started with a custom type, but when I noticed the use of `java.time` instead of `kotlin.time`, I thought it might make sense to align with that, but I wasn’t aware of the Kotlin 2.3+ requirement.

> AGP provides a feature called [desugaring](https://developer.android.com/studio/write/java8-support-table), and it provides a compatibility library, [google/desugar_jdk_libs](https://github.com/google/desugar_jdk_libs), which contains an implementation of older Java APIs that is not natively supported by the OS. It also provides java.time.Duration, so the issue https://github.com/rust-nostr/nostr-sdk-ffi/issues/66 will be okay with it.

Awesome, thank you!

---

Thanks again for the feedback, I’m going to close this PR.

Issue #2653, opened 2025-09-15
Bundling UniFFI in a local Swift Package Manager package - need to import the module declared in the modulemap?


bes:
Hello!

I am doing a multi-step process of creating a local SwiftPM package using https://github.com/ianthetechie/uniffi-starter (but with some tweaks to better suit my setup).

To get the whole thing to compile I need to modify the swift files generated by UniFFI

```swift
// This file was autogenerated by some hot garbage in the `uniffi` crate.
// Trust me, you don't want to mess with it!

// swiftlint:disable all
import Foundation

// Added by me
import MyModuleMapModuleName
```

But, as the message says: "you don't want to mess with it!". And I don't, but it doesn't compile without changing it.

Am I doing something wrong?

lougeniaC64:
Hey @bes, is `MyModuleMapModuleName` the name of your uniffi module or another module you'd like to import? cc @bendk

bes:
~It's the uniffi module created by xcodebuild~

EDIT: the actual statement should be:
It's the name of the module in the module file created by `uniffi_bindgen_swift`, and then I create an xcframework using xcodebuild with that module (and headers) as a input.

mhammond:
You should find your generated code includes code generated by https://github.com/mozilla/uniffi-rs/blob/2d1912a57f731f9fc4936d0cc5270c3f416a1044/uniffi_bindgen/src/bindings/swift/templates/wrapper.swift#L17-L19 - we *think* it is this block which does the import you need. What is the name of the module as written to that block?

bes:
It does include that, but it's not enough (at least not in my case)

# Commands

UniFFI
```shell
cargo run --bin uniffi-swift \
          -- \
          "./target/$build_arch/$build_target/libmy_ios.a" "./target/generated/uniffi" \
          --swift-sources \
          --headers \
          --modulemap \
          --module-name MyModuleMapModuleName \
          --modulemap-filename module.modulemap
```

Then I move the .swift files out of the `generated` directory into `sources/uniffi` (see file tree below) and finally

```shell
    xcodebuild -create-xcframework \
      -library "./target/aarch64-apple-ios/debug/libmy_ios.a" -headers "./target/generated/uniffi/" \
      -library "./target/aarch64-apple-ios-sim/debug/libmy_ios.a" -headers "./target/generated/uniffi/" \
      -output "$SWIFT_PM_DIR/my_ios.xcframework"
```

# Resulting files

[...]

Issue #2763, opened 2025-12-20
System for mixing and matching binding generators


bendk:
Right now we have several builtin bindings generators and maybe a dozen external bindings generators.  We've long wanted to integrate everything together so the external bindings generators weren't second class.  I think we can do this by leaning into the fact that we already encourage users to [define their own uniffi-bindgen binary](https://mozilla.github.io/uniffi-rs/latest/tutorial/foreign_language_bindings.html).

Here's the proposal:
  - Bindings generators define a `clap` Command struct for themselves.
  - They implement the `uniffi::RunCommand` trait for those types, which defines a single `run()` method.
  - Users assemble those commands to create their `uniffi-bindgen` binary
 
For example:

```
#[derive(Parser)]
#[command(version, about, long_about = None)]
struct Cli {
  command: Commands
}

#[derive(Subcommand)]
enum Commands {
   // Builtin commands come from `uniffi`.  We could also moving these to their own crates
   Kotlin(uniffi::KotlinCommand),
   Python(uniffi::PythonCommand),
  // External bindings
   Js(uniffi_bindgen_react_native::Command),
   Cs(uniffi_bindgen_cs::Command),
   // ... define as many as you link
}

fn main() {
  let cli = Cli::parse();
  cli.command.run();
}
```

This system is simple but flexible.  It puts external bindings on the same footing as bulitin bindings and gives users a simple way to run any of them using the same CLI.

bendk:
One of my motivations for this is that I want to experiment with generating Rust code from the metadata and building everything together into a final `.so` file.  For example, if we want to use JNI for the Kotlin/Java bindings, then we need a way of including the Rust JNI code in the final library.  I think we could do that by creating `uniffi_bindgen_jni` which users would invoke my running something like `uniffi-bindgen jni [path-to-rlib-file] --out-dir [path-to-build-directory]`.  This would:
  - Generate a `.rs` file with the JNI code
  - Run `rustc` to compile that file together with the rlib.  This would generate the `.so` file that Kotlin/Java loads in the build directory.

Another use-case for is to generate Rust code that needs to see the entire API.  For example, if we implement the buffer-based FFI described in #2752, then maybe we could generate a single FFI function that could call any of the scaffolding functions (something like `extern "C" fn dispatch_call(call_id: u64, ffi_buffer: *mut u8)`).  That's something that you can only do if you know all the functions in the API, which we don't when we're inside a proc-macro.

bendk:
[...]

PR #2798, opened 2026-01-19
Config file support improvements

mhammond:
* For swift, allow `--config` to be used to override all config options, like currently supported for all other languages.
* For all bindings, allow a new `--crate-metadata` option, which takes a path to a TOML file where you can specify the location of crates. In theory, this would allow you to avoid using cargo-metadata at the cost of maintaining yet another TOML file with crate locations.

As described in [this comment](https://github.com/mozilla/uniffi-rs/issues/2653#issuecomment-3765884106). Needs docs, but thought I'd get it up to see what people thought.

### DIFF
--- a/uniffi_bindgen/src/crate_metadata.rs
+++ b/uniffi_bindgen/src/crate_metadata.rs
@@ -1,0 +1,113 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! Helpers which allow metadata for crate locations to be specified in TOML.
+//! Returns what `cargo_metadata` does, but useful in environments where `cargo_metadata`
+//! isn't available or doesn't work.
+//!
+//! For example, you might have TOML with:
+//! ```toml
+//! [my_crate]
+//! crate_root = "path/to/my_crate"
+//!
+//! [other_crate]
+//! config = "path/to/other_crate/other.toml"
+//! "other_crate.udl" = "path/to/other_crate/src/other_crate.udl"
+//! ```
+//! In the first example, `uniffi.toml` and any requested UDL files will be located relative to the `crate_root`
+//! In the second example, the explicit paths to files are specified.
+//! In all cases, paths can be relative to the TOML file itself, or absolute.
+
+use anyhow::{anyhow, Context, Result};
+use camino::Utf8PathBuf;
+use std::fs;
+
+use crate::BindgenPathsLayer;
+
+#[derive(Debug, Clone, Default)]
+pub struct TomlCrateConfigSupplier {
+    // any relative paths in the TOML will be considered relative to this.
+    root: Utf8PathBuf,
+    // the content of the .toml file we loaded.
+    toml: toml::Table,
+}
+
+impl TomlCrateConfigSupplier {
+    pub fn new(path: &Utf8PathBuf) -> Result<Self> {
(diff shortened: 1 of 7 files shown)


uniffi_bindgen/src/crate_metadata.rs
```diff
@@ -0,0 +1,113 @@
...
+use std::fs;
+
+use crate::BindgenPathsLayer;
+
+#[derive(Debug, Clone, Default)]
+pub struct TomlCrateConfigSupplier {
+    // any relative paths in the TOML will be considered relative to this.
+    root: Utf8PathBuf,
+    // the content of the .toml file we loaded.
+    toml: toml::Table,
```
bendk, line 33: Would it be worth it to define a struct for the config data?  If we decide to call this the "global config", then I think we should.  If not, I could go either way.

bendk (commented):
This is a great improvement.

The one thing I wonder about is how this will work alongside the concept of a "global config" file that we've talked about, which also came up in #2763.  Is this a good time to start iterating on that?  What if we called this the "global config" file and said that right now it's only used to specify UDL/config paths.  Then we could extend it by adding more fields without needing to add new config flags.  I'm not sure on this one, but it seems like it could work out nicely.

bendk:
BTW, a global config file is another way to solve #2653.  You could specify the `ffi_module_name` there and it would apply to all crates.  This makes a more sense than a crate-specific setting, since they all need to be the same value or else building will fail as described in that comment.

mhammond:
Yeah, we should think about these config semantics holistically. 

> BTW, a global config file is another way to solve https://github.com/mozilla/uniffi-rs/issues/2653. You could specify the ffi_module_name there and it would apply to all crates.

It would depend on the semantics, because while `ffi_module_name` *must* be the same for all crates, `ffi_module_filename` *must not* be the same for any crate. I think this is one reason why the existing config override mechanism is broken - it can give us a way to use the same value for all crates, but not a way to override them per crate (right?).

I think any improved system would want to support a "config hierarchy" - ie, a crate-level config, then a workspace config, etc? I also think it's important that work in a non-cargo-workspace world - ie, that we continue to allow uniffi.toml in each crate *and* provide a reasonable mechanism to find those files that doesn't rely on cargo_metadata. Beyond that, I'm not sure what the semantics should be.

ie, naively, I was thinking something like:
```
[crate_one]
# Even though we've lots of config stuff here, we still need this to 
# locate the UDL and per-crate uniffi.toml if we aren't in a cargo workspace
crate_root = "path/to/crate"
[crate_one.bindings.swift]
ffi_module_filename = "crate_one_ffi"
[crate_one.bindings.swift.rename]
...
```
and I guess this could *also* have top-level entries like:
```
[bindings.swift] 
# applies to all crates.
ffi_module_name = "my_library"
[bindings.swift.rename] # <-- doesn't really make sense here, but maybe that's OK?

```
Is that what you had in mind? Or something else? (I think we'd probably want a different table for the "global" options though?)

mhammond:
(Also, looking at #2763, there seems to be a suggestion that the config could be used for more than uniffi - so do we want a top-level `[uniffi]` table?)

bendk:
> Is that what you had in mind? Or something else? (I think we'd probably want a different table for the "global" options though?)

Your suggestion of a hierarchy makes sense to me.  I agree that only some config values make sense to have as globals.  

Honestly, I don't have much in mind other than I think it's a good idea to start exploring with.  I guess start slow and build up?  I'd be fine with only supporting `config` and `crate_metadata` for now, but just calling it the "global config" and continuing to iterate.

PR #2800, opened 2026-01-20
Fix broken documentation links

yamamel:
## Summary
- Fixed 6 broken documentation links across 4 files
- Updated URLs from old `/udl/` structure to new `/latest/` structure
- All links have been verified to be working

## Changes
Updated documentation URLs with the following mapping:
- `/udl/remote_ext_types.html` → `/latest/types/remote_ext_types.html`
- `/udl/ext_types_wrapped.html` → `/latest/types/custom_types.html` (wrapped types were renamed to custom types)
- `examples/log-formatter` → `examples/remote-types` (example was renamed)

## Modified Files
- `CHANGELOG.md` (2 links - only the broken ext_types_wrapped links)
- `docs/manual/src/types/remote_ext_types.md` (1 link)
- `examples/remote-types/README.md` (1 link)
- `uniffi_macros/src/lib.rs` (2 links)

All updated links have been verified to ensure they point to valid pages in the new documentation structure.

Note: I preserved historical documentation links in CHANGELOG.md that were valid at the time of their respective releases, only fixing genuinely broken links.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -796,7 +796,7 @@ Significant patches to UniFFI's builtin bindings which you will need to port inc
 
 - Error handling when converting custom types has been updated. If your `wrap()`
   function returns an `Err`, in some cases it now [may not panic but instead
-  return the error declared by the function](https://mozilla.github.io/uniffi-rs/latest/types/custom_types.html#error-handling-during-conversion).
+  return the error declared by the function](https://mozilla.github.io/uniffi-rs/0.27/udl/custom_types.html#error-handling-during-conversion).
 
 ### What's Changed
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -820,7 +820,7 @@ Significant patches to UniFFI's builtin bindings which you will need to port inc
   constructor, you must add one explicitly.
 
 ### What's Changed
-- Kotlin and Swift, like Python, now support [simple "wrapped" types](https://mozilla.github.io/uniffi-rs/udl/ext_types_wrapped.html).
+- Kotlin and Swift, like Python, now support [simple "wrapped" types](https://mozilla.github.io/uniffi-rs/latest/types/custom_types.html).
```
bendk: Making an older changelog entry point to the latest docs doesn't feel right to me, but neither does a broken link.  As a compromise, how about pointing it to the oldest version we're still making docs for: https://mozilla.github.io/uniffi-rs/0.27/?

bendk (changes requested):
Thanks for the link fixes.  They look good to me with the exception of the changelog links.  I think we should pin them to a version so they don't get outdated again as we rework the documentation.

CHANGELOG.md
yamamel: Great, I changed the two broken links in changelog to v0.27 documents. Thanks for your advice. I see some other links using [https://mozilla.github.io/uniffi-rs/next/](https://mozilla.github.io/uniffi-rs/next/), should we use links like that.

bendk: The changes look great, thanks this fix.

PR #2809, opened 2026-02-03
Disable `clippy::exhaustive_structs` on `UniFfiTag`

Johennes:
This exempts the `UniFfiTag` struct from `clippy::exhaustive_structs`. This avoids clippy errors in downstream projects which cannot add the exemption themselves because the struct is hidden behind the macro.

See also https://github.com/ruma/ruma/pull/2338#issuecomment-3835061395.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### What's Fixed
 
 - Fixed bug that sometimes prevented renaming items inside a submodule [#2792](https://github.com/mozilla/uniffi-rs/pull/2792)
+- Exempted `UniFfiTag` from `clippy::exhaustive_structs` since downstream projects may depend on it [#2809](https://github.com/mozilla/uniffi-rs/pull/2809)
 
 ### What's New?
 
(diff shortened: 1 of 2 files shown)


bendk:
I'm confused about this clippy warning.  Isn't adding a field to a struct normally a breaking change?

That said, I don't see anything wrong with adding this clippy attribute if it helps downstream crates.  We could also flag the struct `non-exhaustive`.

Johennes:
> I'm confused about this clippy warning. Isn't adding a field to a struct normally a breaking change?

Yes, it is. I think the point is that `#[non_exhaustive]` adds a few [restrictions](https://doc.rust-lang.org/reference/attributes/type_system.html) on consuming code that limit those breakages, such as requiring a default arm in `match`. I assume this is why ruma enforces it.

jplatte:
Since it's just a type tag, adding `#[non_exhaustive]` and thus making it not constructible (outside the defining crate) seems reasonable. It's technically a breaking change though.

bendk:
I'm fine with merging this as-is, it's a bit weird but I can't see any downside.  I guess it's technically a breaking change, but I would file this in the non-breaking category since I feel like `UniffiTag` is part of our internal API even though it needs to be `pub` to make things work.  I'd like to keep the option of a `0.31.1` release after we merge this.

Does anyone else have concerns with this one?

jplatte:
Adding non-exhaustive would be a breaking change, silencing the lint most certainly isn't. No concerns either way from me.

bendk:
Good point, let's just merge this one as-is then.

PR #2812, opened 2026-02-06
Fix list in async-overview.md

mgeisler:
The lack of a newline meant that the bullet points ran together into a single paragraph.


bendk (approved):
Awesome, thanks for the fix!

Issue #2804, opened 2026-01-23
Multiple constructors are not available on Android; default constructor is lost!


pronebird:
```rs
#[uniffi::export(async_runtime = "tokio")]
impl NymEnvironment {
    #[uniffi::constructor]
    pub async fn new(
        cache_dir: PathBuf,
        network_name: &str,
    ) -> Result<Self, VpnError> {
        todo!()
    }

    #[uniffi::constructor]
    pub fn new_with_mainnet_fallback() -> Result<Self, VpnError> {
        todo!()
    }

    pub async fn __stub_to_keep_compiler_happy(&self) {
        // todo: remove after updating to uniffi 0.31
    }
}
```

When trying to create object using `new` initializer, I get nasty argument mismatch issue.

```kt
val env = NymEnvironment(storagePath, "some")
```

The default initailizer looks like some internal uniffi method that must not be exposed:

```kt
constructor(withHandle: UniffiWithHandle, handle: Long) {
```

bendk:
I don't think this is an issue with multiple constructors, it's that the constructor is `async` which isn't supported on Kotlin.

I also think you might have issues with that `PathBuf` argument.  UniFFI doesn't support that directly I usually just use `String` to pass paths across the FFI.  Maybe the only reason you're not seeing those yet is because we're not generating the constructor yet.

pronebird:
> I also think you might have issues with that `PathBuf` argument. UniFFI doesn't support that directly I usually just use `String` to pass paths across the FFI. Maybe the only reason you're not seeing those yet is because we're not generating the constructor yet.

I have custom type implemented. We use UTF-8 in Kotlin/Swift so that should work just fine.

```rs
uniffi::custom_type!(PathBuf, String, {
    remote,
    try_lift: |val| Ok(PathBuf::from(val)),
    lower: |val| val.display().to_string()
});
```

pronebird:
> I don't think this is an issue with multiple constructors, it's that the constructor is `async` which isn't supported on Kotlin.

This does not seem to be correct. It works fine as long as the method is not called `new` but something like `newSomething`, then it appears in the output.

bendk:
> > I don't think this is an issue with multiple constructors, it's that the constructor is `async` which isn't supported on Kotlin.
> 
> This does not seem to be correct. It works fine as long as the method is not called `new` but something like `newSomething`, then it appears in the output.

[...]

PR #2813, opened 2026-02-06
Kotlin: define constructors that throw for async functions

bendk:
These constructors always just throw an exception, since Kotlin doesn't supprot async constructors.  This is less confusing for users than just skipping the codegen (#2804).

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
@@ -141,17 +141,19 @@ open class {{ impl_class_name }}: Disposable, AutoCloseable, {{ interface_name }
         this.cleanable = null
     }
 
-    {%- match obj.primary_constructor() %}
-    {%- when Some(cons) %}
+    {%- if let Some(cons) = obj.primary_constructor() %}
     {%-     if cons.is_async() %}
+    {%- call kt::docstring(cons, 4) %}
+    constructor({% call kt::arg_list(cons, true) -%}) {
+        throw InternalException("async primary constructors are not supported on Kotlin");
+    }
     // Note no constructor generated for this object as it is async.
     {%-     else %}
     {%- call kt::docstring(cons, 4) %}
     constructor({% call kt::arg_list(cons, true) -%}) :
         this(UniffiWithHandle, {% call kt::to_ffi_call(cons) %})
     {%-     endif %}
-    {%- when None %}
-    {%- endmatch %}
+    {%- endif %}
 
     protected val handle: Long
     protected val cleanable: UniffiCleaner.Cleanable?


paxbun:
How about adding `@Deprecated` or opt-in annotations to such constructors? With this, we can notify users at compile time that async constructors are not supported in Kotlin as well.

bendk:
In #2804 we discussed making the codegen fail at build time, however I went with this approach for a couple reasons:

* It matches how we handle things in [Python](https://github.com/mozilla/uniffi-rs/blob/a70f45a5f7cf39283c3b9eb8c66031dab3617c03/uniffi_bindgen/src/bindings/python/templates/InterfaceTemplate.py#L13)
* I could see a use-case for a library that defined an async constructor knowing that it only worked on some bindings and providing an alternative method for other bindings.  This is how `fixtures/futures` works actually.

I'm not totally sure this is the right way though, a build-time failure would be nicer in many cases compared to generating code that's just going to throw when you use it.  However, that second bullet feels important to me.  Maybe we could extend the language-specific config code that allows function/method renaming to also allow skipping codegen for functions/methods.  That would give those users a decent path forward: they just need to add the constructor to that skip configuration.

paxbun:
I'm not sure I understood the comment correctly, but `@Deprecated` will just warn the users, not cause compilation errors. If the users know this async constructor behavior, they can explicitly suppress the deprecation using the `@Suppress` annotation.

bendk:
My comment wasn't a response to yours, we just happened to post at the same time.  It was more me thinking out-loud on other ways to handle this.

Adding `@Deprecated` would be okay with me.  I think that might be an improvement over the current code.  However, I'm still wondering if a build error at codegen time would be the best behavior.

paxbun:
I think this will be confusing to many users, but we can simulate async constructors in Kotlin by adding `suspend operator fun invoke(...)` to the companion object. I've never seen an actual production-level code that does this, though.

```kotlin
class Foo private constructor(public val v: String) {
  override fun toString() = "Foo($v)"
  companion object {
    suspend operator fun invoke(v: Int): Foo {
      return Foo(v.toString())
    }
  }
}

fun main() = runBlocking {
  println(Foo(15)) // invoking Foo.Companion.invoke(15)
}
```

paxbun:
https://github.com/search?q=%22suspend+operator+fun+invoke%22+%22companion+object%22+language%3AKotlin&type=code

image

There seem to be several libraries that actually do this. Maybe it would be okay to implement async constructors like this.

pronebird:
If I understand it right this will result in runtime error. I am not particularly fond of this solution since CI checks will pass and give a false sense of safety.

Rust supports conditional compilation so it's possible to handle differences for android/ios by providing alternative constructors. As such build error isn't a problem but would help catch issues early on during CI pass

bendk:
> If I understand it right this will result in runtime error. I am not particularly fond of this solution since CI checks will pass and give a false sense of safety.

I agree with this and made https://github.com/mozilla/uniffi-rs/pull/2822 as a way to move us towards build-time errors.

uniffi_bindgen/src/bindings/kotlin/templates/ObjectTemplate.kt
```diff
@@ -141,17 +141,19 @@ open class {{ impl_class_name }}: Disposable, AutoCloseable, {{ interface_name }
...
 
-    {%- match obj.primary_constructor() %}
-    {%- when Some(cons) %}
+    {%- if let Some(cons) = obj.primary_constructor() %}
     {%-     if cons.is_async() %}
+    {%- call kt::docstring(cons, 4) %}
+    constructor({% call kt::arg_list(cons, true) -%}) {
+        throw InternalException("async primary constructors are not supported on Kotlin");
+    }
     // Note no constructor generated for this object as it is async.
```
mhammond, line 150: this comment is now wrong?

mhammond (commented):
IIUC, you don't want to do this now, but would rather throw at bindgen time?

bendk:
> IIUC, you don't want to do this now, but would rather throw at bindgen time?

Yup, that's how I'm feeling.  I implemented the "throw at bindgen time" option in https://github.com/mozilla/uniffi-rs/pull/2827, I think we should merge that one and close this one.

bendk:
I just merged #2827 instead of this one.

PR #2815, opened 2026-02-06
Fail when we see no metadata

bendk:
We recently had issues because a library had its symbols stripped, which made metadata generation fail.  It would have been easier to spot these issues if `uniffi-bindgen` failed loudly rather than returning `0` but with no output.  I tested this code with `cargo run -p uniffi-bindgen-cli` and it does indeed fail with code `1` when run against a library with no UniFFI metadata or with a crate filter that doesn't match any crates.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/mod.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/mod.rs
@@ -6,6 +6,7 @@ use crate::{
     bindings::GenerateOptions, interface::rename, BindgenLoader, Component, ComponentInterface,
     Result,
 };
+use anyhow::bail;
 use camino::{Utf8Path, Utf8PathBuf};
 use fs_err as fs;
 use std::collections::HashMap;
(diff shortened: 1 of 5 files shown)


uniffi_bindgen/src/bindings/kotlin/mod.rs
```diff
@@ -19,6 +20,12 @@ pub mod test;
 /// Generate Kotlin bindings
 pub fn generate(loader: &BindgenLoader, options: GenerateOptions) -> Result<()> {
     let metadata = loader.load_metadata(&options.source)?;
+    if let Some(crate_filter) = &options.crate_filter {
+        if !metadata.contains_key(crate_filter) {
+            bail!("No UniFFI metedata found for crate {crate_filter}");
```
mhammond: typo - "metadata" - which you copy-pasted everywhere this message appears.

mhammond (approved):

PR #2816, opened 2026-02-09
ADR: Buffer FFI

bendk:
We've been talking about this for a while and I think it's a good time to start the discussion for real.

### DIFF
--- a/docs/adr/0010-buffer-ffi.md
+++ b/docs/adr/0010-buffer-ffi.md
@@ -1,0 +1,440 @@
+# Buffer FFI
+
+* Status: proposed
+* Deciders:
+* Date: 2024-05-24
+
+Discussion and approval: [PR 2816](https://github.com/mozilla/uniffi-rs/pull/2816)
+
+## Context and Problem Statement
+
+Our current FFI is based on passing arguments/return values using the C ABI.
+This forces languages like Kotlin, Python, and JS that can't make C calls directly
+to use an intermediate layer to make these calls.
+Usually this means a libffi-based library, like ctypes or JNA.
+However, there are several issues with this approach:
+
+* Performance can be poor especially when using JNA.
+* JNA is has been a persistent source of issues on Kotlin.
+  There are a couple current issues without clear solutions: #2740, #2624.
+* Limitations in these libraries limit how we can design the FFI.
+  For example, callback methods can't return values directly because of https://bugs.python.org/issue5710.
+
+We want to design a new FFI that avoids these issues as much as possible.
+
+## Scope
+
+This document discussion a new FFI for languages like Kotlin, Python, and JS that can't make C calls directly.
+This hopefully helps us move toward a stable "1.0" FFI,
+but this document is not proposing that we freeze the API at this point.
+
+Languages like Swift that can make C calls directly are not considered.
+Maybe we will continue to use the current FFI for those languages or maybe we'll develop another new FFI.
+
+The exact mechanism for making these calls is also out-of-scope for this ADR.
+For example, Kotlin and Java may use JNA or JNI.
+
+## Decision Drivers
+
+## Decisions
+
+This document is organized around several independent decisions rather than a single one.
+Hopefully this makes each decision clearer.
+
+### Using a buffer to pass FFI values
+
+The main proposal is to change the general form for FFI calls to:
+
+```
+extern "C" uniffi_buffer_ffi_function_name(ffi_buffer: *u8) {
+    // ... code here
+}
+```
+
+Each FFI call inputs a single buffer that's used for both input arguments and return values.
+Let's name the FFI the "Buffer FFI" and the function argument the "FFI buffer".
+To avoid conflicts with the current FFI, all FFI functions should be prefix with `uniffi_buffer_`
+
+Callees should:
+
+ * Read all arguments from the buffer
+ * Lift the argument data into high-level types
+ * Call the exported function using the lifted arguments
+ * Lower the return value into an FFI type
+ * Write the result the buffer:
+    * For successful calls, write `0` followed by the return value.
+    * For expected errors, write `1` followed by the error value
+    * For unexpected errors, write `2` followed by a `RustBuffer` containing a error message.
+
+#### Packing primitive values to the FFI buffer
+
+* Ints and floats are packed in native-endian format.
+* Pointers are casted to `u64` values then packed into the buffer.
+  Function pointers are handled the same way.
+* All items are aligned to 64-bit addresses.
+
+#### Allocating the buffer
+
+Callers must ensure the buffer is large enough to hold all arguments as well as the return value
+(one or the other, not both at once).
+Callers may use different strategies to allocate/free the FFI buffer, including:
+
+* Allocating a new buffer for each call
+* Creating an array on the stack
+* Statically allocating a single buffer for each thread in a thread-local variable
+
+In order to support all of these strategies, callers must read all data from the buffer immediately
+before there's any chance of another call across the FFI.
+
+As long as no dynamically sized values are present in the argument list or return value,
+then the buffer will have a fixed and relatively small size.
+Buffers for dynamically sized values are discussed in the "RustBuffer and heap data" section below.
+
+#### Performance
+
+The performance of the buffer FFI has been well tested on Kotlin 
+with benchmarks showing that it performs much faster than the current FFI.
+See the appendix below for details,
+the TLDR is that it speeds up most calls by a factor of 100x or so.
+
+### Using JNI/pyo3 to pass FFI values
+
+The main alternative to the buffer FFI that was considered
+was using a language-specific bindings layer like JNI or pyo3.
+Defining a Python module using the C API is another potential path.
+We mostly focused on JNI and this section will reflect that.
+However, it's expected that the same logic applies to pyo3 and other systems.
+
+This would mean generating Rust code that looked like this:
+
+```
+#[allow(clippy::missing_safety_doc)]
+#[unsafe(no_mangle)]
+pub unsafe extern "system" fn Java_some_package_name_UniffiLibrary_rustFunc(
+    mut env: JNIEnv,
+    _class: JClass,
+    int_arg: i32,
+    vec_of_ints: JObject,
+    struct_arg: JObject,
+    vec_of_structs: JObject,
+) -> f64 {
+    // Note: no lifting needed for `int_arg`
+
+    // Lift the vec_of_ints arg
+    let vec_of_ints_elements = env.get_array_elements_critical(vec_of_ints, ReleaseMode::NoCopyBack).unwrap();
+    let vec_of_ints_lifted = vec_elements.iter().collect();
+
+    // Lift the struct arg
+    let field_ids = STRUCT_FIELD_IDS.get().unwrap();
+    let struct_lifted = TheStruct {
+        field_a: env.get_field_unchecked(
+            &struct_arg, 
+            field_ids.0,
+            ReturnType::Primitive(Primitive::Int),
+        ).unwrap().i().unwrap(),
+        field_b: env.get_field_unchecked(
+            &struct_arg, 
+            field_ids.1,
+            ReturnType::Primitive(Primitive::Double),
+        ).unwrap().d().unwrap(),
+    };
+
+    // Lift the vec_of_structs arg
+    // Unfortunately, we need to get each element one at a time rather in this case.
+    let vec_of_structs_length = env.get_array_length(vec_of_structs).unwrap();
+    let vec_of_structs_lifted = (0..vec_of_structs_length).iter()
+        .map(|i| {
+            let struct = env.get_object_array_element(vec_of_structs, i);
+            TheStruct {
+                field_a: env.get_field_unchecked(
+                    &struct, 
+                    field_ids.0,
+                    ReturnType::Primitive(Primitive::Int),
+                ).unwrap().i().unwrap(),
+                field_b: env.get_field_unchecked(
+                    &struct, 
+                    field_ids.1,
+                    ReturnType::Primitive(Primitive::Double),
+                ).unwrap().d().unwrap(),
+            }
+        })
+        .collect::<Vec<_>>();
+
+    // Call the Rust function
+    let result = rust_func(int_arg, vec_of_ints_lifted, struct_lifted, vec_of_structs_lifted);
+
+    // ...lower and return the return value, handle errors, etc.
+}
+```
+
+The corresponding generated Kotlin code would be fairly simple,
+since a lot of the lowering is happening in the JNI layer.
+
+JNI code performs better than buffers in some cases:
+
+* **Primitive values** (ints, floats, bools, etc).  Benchmarks show about a 20% speedup.
+* **Arrays of primitives**.
+  This has not been tested, but it seems safe to assume the JNI approach is faster.
+
+However, buffers perform faster in other cases:
+
+* **Structs**.
+  It's slightly faster to read elements from a buffer than to make a JNI function call for each field.
+* **Enums**.
+  In addition to needing JNI calls per field, but you also need JNI calls to figure out the enum variant.
+  Calling `IsInstanceOf` for each variant seems very slow (although this has not been tested).
+* **Nested data** (vecs of structs, hash maps, structs with enum fields, etc).
+  At this point buffers start to significantly out-perform JNI
+  since you need to make more than 1 JNI call per item.
+  For example, with a vec of structs you need to make an extra JNI call per item in addition to the
+  JNI calls for each field.
+
+See https://github.com/mozilla/uniffi-rs/issues/2672 for further discussion and the rough
+benchmarks.
+
+JNI supports passing nio buffers to the C functions which we could use to speed up these cases.
+We could use the nio buffers for cases from the second list
+and "normal JNI" for cases in the first list.
+
+However, if the only cases where you don't want a buffer are primitives and arrays of primitives,
+you start to wonder why not use a buffer for everything?
+The performance difference will often be negligible
+since primitive values are already quite fast relative to structs/enums,
+and the code generation would be simplified.
+This would essentially mean we're back to the buffer FFI approach, but using JNI to implement it.
+
+### Using the C ABI to pass FFI values
+
+Another option would be to double-down on the C-ABI and use it to pass structs and enums as well.
+Structs would be passed as the `repr(C)` version of the struct.
+Enums would be passed as tagged unions: one `u64` field for the variant discriminant
+and a `repr(C)` union for the variant data.
+
+This would be difficult to implement in JNA and would likely be slower than the existing code,
+since it would require more complex JNA logic to read and write the structs.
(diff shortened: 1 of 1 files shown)


bendk:
We're currently testing some experimental Kotlin code that implements an FFI somewhat along these lines, but solely focused on trying to reduce the crashes we're seeing (https://github.com/mozilla/uniffi-rs/pull/2765).  I don't think we should approve this before that's done, but I also don't think that experiment blocks the discussion here.

I think we can be confident this will work on Kotlin and improve performance there.  Should we try to implement this on another platform as well to confirm that it can be done and to check the benchmark differences?  uniffi-bindgen-gecko-js would be my pick.

docs/adr/0010-buffer-ffi.md
```diff
@@ -0,0 +1,269 @@
...
+
+This document discussion a new FFI for languages like Kotlin, Python, and JS that can't make C calls directly.
+This hopefully helps us move toward a stable "1.0" FFI,
+but this document is not proposing that we freeze the API at this point.
+
+Languages like Swift that can make C calls directly are not considered.
+Maybe we will continue to use the current FFI for those languages or maybe we'll develop another new FFI.
+
+The exact mechanism for making these calls is also out-of-scope for this ADR.
+For example, Kotlin and Java may use JNA or JNA.
```
jmartinesp: Tough choice 😀 !

bendk:
Updated the ADR to consider the JNI alternative more seriously (and fix that funny typo).

I should also mention that the ADR is written as if this decision has already been made, but that's just because it's the style we'll want it in eventually.  This is still up-for-debate.

bendk:
Pushed an update that:
- Adds a comparison to boltffi, since that's been gaining attention lately
- Adds a discussion of going all-in on a C-ABI, which is how I think would want to handle it for languages like Swift if we had the time.
- Updates the main decision to say that a couple of the options have potential and we may want to check them out in the future.  I don't want to commit to doing that, but I don't want it to seem like this ADR blocks exploring those options either.

docs/adr/0010-buffer-ffi.md
```diff
@@ -0,0 +1,440 @@
...
+
+## Context and Problem Statement
+
+Our current FFI is based on passing arguments/return values using the C ABI.
+This forces languages like Kotlin, Python, and JS that can't make C calls directly
+to use an intermediate layer to make these calls.
+Usually this means a libffi-based library, like ctypes or JNA.
+However, there are several issues with this approach:
+
+* Performance can be poor especially when using JNA.
```
mhammond, line 17: I think it's worth noting here that non-native types have no easy solution in this world which contributes to performance challenges, because the C ABI often can't represent these types (eg, IIUC, some enums, probably Arc's etc, can't just be "copy bits and you are done"). I'm not sure if the first point should incorporate that, or as a new point - I guess it depends on whether JNA would still have poor performance if only native types existed?

docs/adr/0010-buffer-ffi.md
```diff
@@ -0,0 +1,440 @@
...
+* JNA is has been a persistent source of issues on Kotlin.
+  There are a couple current issues without clear solutions: #2740, #2624.
+* Limitations in these libraries limit how we can design the FFI.
+  For example, callback methods can't return values directly because of https://bugs.python.org/issue5710.
+
+We want to design a new FFI that avoids these issues as much as possible.
```
mhammond, line 23: this raises an interesting point that we've discussed in person but doesn't seem covered here - is part of the problem that UniFFI decided to use a single FFI to cover all bindings? IOW, it seems bad a Python bug forces us to make compromises which impact the languages we actually care about. Similarly, it doesn't scale as other languages covered by external bindings may have similar issues we are unaware of?

IOW, is the fact we are saying "***a*** new FFI" just compounding that original sin?

you kinda touch on this below in a few places, (eg, talking about where you say there might be 2 FFIs) but I'm consolidating my comments here - should this FFI go further?

I'm thinking of working towards specifically *allowing* (but not necessarily *having* in the first instance) language-specific FFIs. Explicitly allowing Swift, Python and Kotlin to have their own, custom targeted FFIs. Then, once we allow that, this becomes purely about the Kotlin FFI. For pragmatic reasons, we may initially have Python use the same FFI, but later an O3-based Python specific API might be contributed. Later, the swift-specific FFI would keep some of the existing FFI, but maybe could in-fact use repr(C) structs directly in some cases etc.

docs/adr/0010-buffer-ffi.md
```diff
@@ -0,0 +1,440 @@
...
+
+## Scope
+
+This document discussion a new FFI for languages like Kotlin, Python, and JS that can't make C calls directly.
+This hopefully helps us move toward a stable "1.0" FFI,
+but this document is not proposing that we freeze the API at this point.
+
+Languages like Swift that can make C calls directly are not considered.
+Maybe we will continue to use the current FFI for those languages or maybe we'll develop another new FFI.
```
mhammond, line 32: I agree it's beyond the scope of this doc, but I think that if we accept this, we should promise to revisit our existing FFI. eg, we note above that our current FFI isn't as performant for Swift as it could be because it tries to work with Kotlin/Python etc, so this ADR removing the restriction offers the ability to extract more perf from Swift like languages. (OTOH though, this becomes moot if we allow *many* FFIs)

docs/adr/0010-buffer-ffi.md
```diff
@@ -0,0 +1,440 @@
...
+
+The performance of the buffer FFI has been well tested on Kotlin 
+with benchmarks showing that it performs much faster than the current FFI.
+See the appendix below for details,
+the TLDR is that it speeds up most calls by a factor of 100x or so.
+
+### Using JNI/pyo3 to pass FFI values
+
+The main alternative to the buffer FFI that was considered
+was using a language-specific bindings layer like JNI or pyo3.
```
mhammond, line 103: This is roughly what I'm advocating we *allow*, even if we choose to take a slower path to get there and choose to use a new middle-ground FFI for Kotlin *and* Python in the first instance.

docs/adr/0010-buffer-ffi.md
```diff
@@ -0,0 +1,440 @@
...
+     These could call any UniFFI scaffolding function regardless of the signature of the underlying Rust function.
+   * Bad, because we need to cast/serialize data pointers and function pointers to the buffer.
+     This makes pointer providence trickier and could cause issues on exotic platforms
+     where the pointer width is greater than 64 bits.
+
+* [A2] Current FFI
+   * Good, because we've already implemented it.
+ 
+* [A3] Language-specific FFIs, like JNA/pyo3
+   * Bad, because it performs poorly for structs/enums/nested data.
```
mhammond, line 244: So yeah :)

I don't understand this first bullet - language specific FFIs are only as bad as the author makes them - they can be as fast as possible given they only need to consider a single language?

The second point is valid though, but not clear how bad that actually is - it's probably "just" a complexity cost and not a size cost, as most binaries will have a single target?

docs/adr/0010-buffer-ffi.md
```diff
@@ -0,0 +1,440 @@
...
+  * `rustbuffer_alloc` can return a pointer since the caller knows the length/capacity
+  * `rustbuffer_free` can input the pointer/length/capacity as separate arguments, rather than the struct.
+
+## Appendix
+
+### Performance testing
+
+When designing this FFI, we did a lot of Kotlin performance testing to compare the different possibilities.
+Kotlin was chosen because it has the worst performance of all builtin bindings.
+The buffer FFI should also improve performance for Python and other languages,
```
mhammond, line 376: this mildly contradicts above where you say "other" languages  are out of scope (ie, you explicitly mention Python and Kotlin here, so "other" must include Swift here?). ie, I'm not necessarily convinced this would improve performance for Swift functions which don't fail and only use native types.

I think I'm suggesting that even if we can't take "every language has its own FFI" on board, should we commit to "there are 2 FFIs, swift would never use the one we describe here"? JS is also largely absent from this doc.

docs/adr/0010-buffer-ffi.md
```diff
@@ -0,0 +1,440 @@
...
+* The `TodoEntry` struct and `Vec<TodoEntry>` are passed using a buffer (the `e`, `e2`, `l2` params and the return value)
+  boltffi calls this the "WireProtocol" and defines `WireReader` and `WireWriter` types.
+  boltffi uses a separate buffer for each of these values.
+
+This is very close to [A4]:
+  * We also pass Object handles as longs
+  * It uses a mixed strategy of sometimes passing JVM values directly and sometimes packing them in a buffer.
+    We may decide to use a different strategy for particular types, but the general strategy is the same
+  * The main difference is that boltffi allocates and passes multiple buffers,
+    while this ADR recommends using a single buffer for all arguments.
```
mhammond, line 440: This seems to be saying their worst-case is close to our worst-case. Is it worth mentioning how their best-case is so much more performant than our current story, and how that might change with the proposals here?

docs/adr/0010-buffer-ffi.md
```diff
@@ -0,0 +1,440 @@
...
+  * Bad, because only works when we know the field layout.
+    In practice this means it only works for Rust -> FFI calls.
+    Furthermore, we'd have to handle `#[repr]` attributes somehow.
+  * Bad, because it adds significant complexity to the FFI.
+  
+#### Decision Outcome
+
+Option [B2] Serialize structs/enums fields into the FFI buffer
+
+### RustBuffer and heap data
```
mhammond, line 309: this section seems like it would be an important driver here. I guess I don't see any hard-blockers here, but it would seem nice to have a high degree of confidence we aren't thrashing the heap and/or using more buffers then we assume we can get away with.

mhammond (commented):
Thanks for the update - I hope I'm not distracting from the issue with my thoughts and I'm certainly not trying to prevent motion here!

docs/adr/0010-buffer-ffi.md
bendk: I'm thinking we should frame this as non-native types represent inherent complexity that we need to deal with and using JNA on top of that adds extra problems.

docs/adr/0010-buffer-ffi.md
bendk: I totally agree and this made me think maybe the whole ADR should have a different structure.  Rather than saying "This is the new FFI" I'd like to:

* Lay out several options for passing FFI values and discuss the pros and cons
* Propose a couple concrete FFIs that we can definitely support (one aimed at Kotlin/Python/JS and one aimed at Swift).
* Discuss and encourage bindings to create language-specific extensions on the FFIs

It's a big change though, so I didn't want to force push it over this PR.  Here's my draft, how do others feel about it?

https://github.com/bendk/uniffi-rs/blob/push-muwspntysyoy/docs/adr/0010-buffer-ffi.md

bendk: I also agree with all your other comments below here and I think that new draft addresses them, but tell me if there's anything that I missed.

bendk:
If you squint enough while picturing JNI, then primitive arguments look like registers and the FFI buffer looks like the stack.  What if we leaned into that?  For example:

```rust
// If want to export this API:

pub struct Point {
    coord_x: f64,
    coord_y: f64,
}

pub struct Line {
    start: Point,
    end: Point,
}

pub fn gradient(ln: Line) -> f64 {
    // ...
}


// I'm pretty sure the fastest way would be with this JNI function
pub unsafe extern "system" fn Java_org_mozilla_geometry_UniffiLibrary_gradient(
    mut env: JNIEnv,
    _class: JClass,
    a1: f64,
    a2: f64,
    a3: f64,
    a4: f64,
) -> f64 {
    let result = gradient(Line {
        start: Point { coord_x: a1, coord_y: a2 },
        end: Point { coord_x: a3, coord_y: a4 },
    };
    // then return the `f64` value, which is slightly trickier than it seems
}
```

The general idea is that we deconstruct each struct into it's primitive values, pass those values as primitive arguments, then reconstruct the values on the other side. After some limit (maybe 32 args?), then the primitive values "spill over" and we start packing them to the FFI buffer.

Enums would work roughly the same way.  We'd need to reserve enough arguments to handle the largest variant + the discriminant.  This means some arguments could be unused and we'd just pass default values in that case.  I think this is faster than packing into the FFI buffer, but we could benchmark it.

Return values are interesting.  We can only return 1 primitive value, so we only have 1 register. Also, return values are essentially an enum since we want to catch Rust panics and return an unexpected error value in that case.  I'm thinking the return value will be the `call_status` value AKA the discriminant of that enum.  For simple functions like gradient, we'll usually return `0` for the call status and pack the return value at the start of the FFI buffer.

BTW, where's the FFI buffer argument?  I'm wondering if we can avoid passing that altogether by implementing a JNI interface that looks like this:

```rust
// Get an 8-bit value from the FFI buffer at a specific offset
//
// The FFI buffer is a thread-local byte array that Rust manages
pub unsafe extern "system" fn Java_org_mozilla_uniffi_FfiBuffer_getByte(
    mut env: JNIEnv,
    _class: JClass,
    offset: i64,
) -> i8 { }

// Repeat for all other primitive types and also setter methods
```

Having Rust manage the buffer provides some nice benefits:
* A single buffer for both sides of the FFI
* You don't have to pass around the buffer pointer arg
* The Kotlin side doesn't need any special thread local code and doesn't need to make an allocation per FFI call.

I can't see any real downside.  As discussed in the ADR, if both sides of the FFI read all FFI arguments immediately, and this is thread-local, then the buffer can be safely shared.

Zooming out and thinking about this ADR, I think [the new direction I drafted up](https://github.com/bendk/uniffi-rs/blob/push-muwspntysyoy/docs/adr/0010-buffer-ffi.md) is the way to go.  It seems too strong to say all languages should use the buffer FFI. I'd rather say that the buffer FFI is a decent baseline that's fairly easy for non-C based languages to adopt, but those languages are also free to experiment with language-specific optimizations.

Also, this was inspired by #2841 and the IR pipeline code.  If you put those together, you can generate both the Rust scaffolding and foreign language bindings from the same initial pipeline pass, which makes it easier to implement complex argument passing schemes.  For example, that's where we would decide how primitive value are passed: which JNI argument is it allocated too? / what's it's offset in the FFI buffer?  Trying to line everything up when the code is split between `uniffi_macros` and `uniffi_bindgen` seems much harder.

bendk:
Re-reading this ADR, I'm not sure there's really a decision to be had here.  The context/summary describes our current problems, but the rest of the ADR mostly boils down to "languages should make their own FFI choices".  There's also the "no-brainer" section and the section proposing that we allocate one RustBuffer instead of many, which feels like a no-brainer to me as well.

I feel like there's some good discussion here, but maybe it's not really a ADR.  I could see some of it going into a design doc or even just a docstring at the top of a module.  The benchmarking numbers maybe could go into a doc somewhere, but I'm not sure they're going to be useful after we've implemented the new Kotlin FFI.  I could also see some smaller / highly focused ADRs, but again I'm not convinced it's worth it.

How do others feel?

docs/adr/0010-buffer-ffi.md
paxbun: @bendk "Language-specific FFIs" here refers to JNI/pyo3, right?

@mhammond I think the first bullet means:

- JNI and pyo3 only support passing primitive numeric values (such as `int` or `float`) and `jobject` or `PyObject*`.
- There's no feasible way to pass C structures, which requires calculating the memory layout of the structure and determining whether to pass them as a value or a reference (due to the size of registers).
- So we need to pass structs/enums/nested data using Java or Python classes, which forces the Rust side to access the foreign side data using reflection-like ways.
- Which results in poor performance.

mhammond: right - but I don't think that's relevant here. What I'm trying to say is that "language specific FFIs" are able to do *the best possible thing* for that language. The fact that pyo3 isn't able to use C structs remains true for *all* these options, so mentioning that as "bad" here doesn't make sense to me - every other option would be either just as bad, or worse. There's no option which can change that fact.

So maybe I'm just missing what "language-specific FFIs" here actually means? Because to my mind, [A4] also doesn't make sense, because in my meaning, [A3] would already encourage using a buffer for performance where that makes sense for that language.

mhammond:
I like your new version! However, as we discussed async, it seems that for us to support "language specific FFIs" that work with external bindings, we would need something like #2841, and external bindings would be forced to implement something using a technique like that so that a build.rs script could generate the Rust side of that language-specific API? Otherwise, they could just choose to use the "builtin" FFI, losing the opportunity for some optimizations, but otherwise deciding that's a "good enough" outcome for them just to avoid forcing all their consumers into that model.

In which case I'm now back to wondering whether a better outcome is where you started - with 2 "blessed" FFIs which we (and external authors) do our best with.

I agree that we don't seem in a place to make a decision here yet, and also unclear if it makes sense as an ADR at all.

bendk:
I'm still not sure how the ADR should look, but I have more thoughts after doing some work on the new Kotlin FFI.

I'm thinking we should combine the idea of a "FFI buffer" and a "Rust Buffer".  The current ADR uses the first term for a small/fixed-sized buffer and uses the second term for a dynamically sized buffer.  However, I think we can have both:

- Let's call it an "FFI Buffer"
- The entire FFI Buffer is a linked list of individual buffers
- The first buffer has some fixed size (maybe 256 bytes)
- Subsequent buffers are twice the size of the previous one
- The pointer to the next buffer is stored as the last 8 bytes of the buffer (null if there's no next buffer)

This has a lot of nice features:
- Using a single buffer type simplifies the code
- You can still optimize for small buffers.  When a call is finished and it's time to free the buffers, you skip freeing the first buffer and instead store it in a thread local.  That way you can always get a 256 byte buffer without needing to perform an allocation.
   - We should benchmark that it's actually faster to use a thread local than do the allocation.  I like the fact that the code won't need to change much to test this.
   - If it is faster, then we get a form of the "small vec optimization" for free since the same optimization will also apply to calls that include small dynamically sized values.
- You only need a single pointer to pass the FFI buffer around, which avoids some issues.  For example, it's not so easy to return a (data, size) tuple from a JNI function.
- ...but returning a buffer value isn't really an issue anyway because you can always re-use the input FFI buffer for the return value.

Maybe we can reframe this ADR as "an efficient buffer for FFI values".  You start with the general case, then discuss how small buffers can be optimized.   We could hand-wave about language-specific optimizations to pass values via arguments/return values rather than a buffer.  I'm not sure that's needed, but I do sort-of like picturing the FFI buffer as the canonical way to pass values and every language-specific change is just an optimization on top of that.

BTW, even for the Swift case, I think we'll always need buffers for some types, like hash maps or vecs of enums.

bendk:
I don't think this is a good solution anymore.  After actually trying to implement this I realized two things:

* The idea I proposed in [this comment](https://github.com/mozilla/uniffi-rs/pull/2816#issuecomment-4117958058) is actually very general and can work for many types.  The only types that need a buffer are Vec, HashMap, and Set.
* For those types, it's better to allocate multiple fixed sized buffers than a single buffer.  Yes you have to pay the cost for the allocation, but in return you don't checks in your code to make sure everything is aligned and doesn't overflow.  When I benchmarked things the multiple buffers were faster and the code was simpler.

PR #2820, opened 2026-02-12
Support exporting traits on Records

### LINKED ISSUE #2819: Enable exported foreign trait implementations on Records
With https://github.com/mozilla/uniffi-rs/pull/2724 users can now export methods on Records. It would be nice if we could take that one step further and have a Record implement a foreign trait. 

For example, if I export `impl MyForeignTrait for MyRecord` I would like the generated swift to have: `extension MyRecord: MyForeignTrait {}`. 

For some context, I want  this feature on Records instead of just using an Object because I specifically want the memory model of Records to avoid garbage collection issues in Kotlin but still have the UX of passing around foreign traits. The main downside is the performance hit of raising/lowering into a buffer, but that is acceptable for my use case and presumably others.

joe-p:
This PR adds support for records implementing trait interfaces (previously only objects could). On the Kotlin side this is straightforward. The record's data class declaration includes the trait interfaces. On the Swift side, the main design decision is around `AnyObject`. Existing `...Protocol` types all extend `AnyObject`, which prevents value types (structs) from conforming. To solve this without breaking existing code, a new `Any...Protocol` is generated without the `AnyObject` constraint, and the existing `...Protocol` inherits from both `AnyObject` and `Any...Protocol`. Records then conform to `Any...Protocol`.

It should be noted that I am not super familiar with Swift, Kotlin, or the uniffi codebase, so I did use Claude for some of the work in this PR. 

Closes #2819

TODO: 

- [ ] Update docs
- [ ] Make sure kotlin is passing tests (don't have it setup locally)
- [ ] Scope out any work needed for python


mhammond:
fwiw, I'd probably approve this when it's in good shape - it's not unlike some of the odd-but-sane things I'm responsible for :)

PR #2822, opened 2026-02-13
Allow bindings to exclude items

bendk:
The motivation for this is avoiding async constructors (https://github.com/mozilla/uniffi-rs/pull/2813).

We'd like things to fail at build time if you have an async constructor and are generating bindings for a language that doesn't support them. However, this prevents a valid use case: I want the async constructor on languages that do support them but not on ones that don't.

This change allows us to support that by allowing users to exclude the constructor on languages that don't support it.

Actually, now that think about it I don't think any built-in language supports async constructors, but maybe external binding generators could or this could be useful for some other feature.

### DIFF
--- a/uniffi_bindgen/src/interface/exclude.rs
+++ b/uniffi_bindgen/src/interface/exclude.rs
@@ -1,0 +1,79 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! Exclude CI items based on the TOML configuration.
+
+use std::collections::HashSet;
+
+use crate::interface::{AsType, ComponentInterface, Constructor, Function, Method, TypeUniverse};
+
+pub fn apply_exclusions(ci: &mut ComponentInterface, exclusions: &[String]) {
+    let exclusions: HashSet<&str> = exclusions.iter().map(String::as_str).collect();
+
+    remove_functions(&exclusions, &mut ci.functions);
+    remove_type_definitions(&exclusions, &mut ci.records, &mut ci.types);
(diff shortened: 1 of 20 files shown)


pronebird:
P.s: I think Swift supports async init()

uniffi_bindgen/src/pipeline/general/namespaces.rs
```diff
@@ -29,8 +29,19 @@ pub fn map_namespace(namespace: initial::Namespace, context: &Context) -> Result
...
+                    .into_iter()
+                    .filter_map(|type_def| {
+                        match exclude::should_exclude_toplevel_item(type_def.name(), context) {
+                            Err(e) => Some(Err(e)),
+                            Ok(true) => None,
+                            Ok(false) => Some(type_def.map_node(context)),
+                        }
+                    })
+                    .collect::<Result<Vec<_>>>()?,
+                // Add new type definitions for types found by walking the IR and fitype_nnding all types
```
mhammond: typo in comment - `fitype_nnding`

uniffi_bindgen/src/interface/exclude.rs
```diff
@@ -0,0 +1,79 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! Exclude CI items based on the TOML configuration.
+
+use std::collections::HashSet;
+
+use crate::interface::{AsType, ComponentInterface, Constructor, Function, Method, TypeUniverse};
+
+pub fn exclude(ci: &mut ComponentInterface, exclude: &[String]) {
```
mhammond: total nit, but maybe `apply_exclusions` would be more consistent with the renaming and even a little more descriptive?

mhammond (approved):
This makes sense and is consistent with per-binding renaming.

PR #2824, opened 2026-02-16
Performance improvement by fixing python loop over bytes

Hollyqui:
Improving performance significantly by not looping over bytes individually but just allocating buffer directly. For detailed description see here: https://github.com/n0-computer/iroh-ffi/issues/229, I noticed this when working with iroh where the loop reduces transfer speeds significantly.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/templates/RustBufferTemplate.py
+++ b/uniffi_bindgen/src/bindings/python/templates/RustBufferTemplate.py
@@ -168,7 +168,7 @@ class _UniffiRustBufferBuilder:
             self.rbuf = _UniffiRustBuffer.reserve(self.rbuf, num_bytes)
         yield None
         self.rbuf.len += num_bytes
-    
+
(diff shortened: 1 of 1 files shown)


uniffi_bindgen/src/bindings/python/templates/RustBufferTemplate.py
```diff
@@ -166,17 +168,18 @@ def _reserve(self, num_bytes):
             self.rbuf = _UniffiRustBuffer.reserve(self.rbuf, num_bytes)
         yield None
         self.rbuf.len += num_bytes
-
+
```
mhammond: could you please remove this extra whitespace.

mhammond (commented):
this seems great, thanks!

uniffi_bindgen/src/bindings/python/templates/RustBufferTemplate.py
Hollyqui: removed

PR #2825, opened 2026-02-19
Kotlin: expose `Object.uniffiIsDestroyed`

jmartinesp:
This can help us understand when an object has been destroyed so we can avoid using it, instead of try-catching its usage and recovering from any errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Record fields can now be renamed with the proc-macro `name = "new_field_name"` attribute ([#2794](https://github.com/mozilla/uniffi-rs/pull/2794))
 - Items can be excluded from the generated bindings using `uniffi.toml`.
 - Added `mutable_records` configuration option to allow specific records to remain mutable even when `generate_immutable_records` is enabled (Kotlin and Swift).
+- Kotlin objects now have an `uniffiIsDestroyed` property that returns `true` if the Rust reference no longer exists ([#2825](https://github.com/mozilla/uniffi-rs/pull/2825))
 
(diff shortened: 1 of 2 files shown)


jhugman:
> instead of try-catching its usage and recovering from any errors.

My understanding of the `destroy` method is that it shouldn't error. If it errors then either: 

- it's a bug in uniffi's cleanable/uniffi_free machinery OR
- it's a bug in the client Rust `drop` method.

> f we're exposing the `destroy` method then `isDestroyed` also makes sense

I would like to propose a naming convention which discourages use, and reduces collisions. `uniffiDestroy` and `uniffiIsDestroyed` might the straw fox to start with.

bendk:
> My understanding of the destroy method is that it shouldn't error. If it errors then either:

IIUC, the error would happen when calling other methods on the object after it's already been destroyed.  Is that correct @jmartinesp?

> I would like to propose a naming convention which discourages use, and reduces collisions. uniffiDestroy and uniffiIsDestroyed might the straw fox to start with.

`destroy` comes from one `AutoClosable`  so I don't think think we can change that.  I like `uniffiIsDestroyed` though -- especially since this is a new method.  Just to double check: there's no method for this in one of the standard interfaces right?  I didn't see one in `AutoClosable` or `Disposable`, but maybe I missed something.

jmartinesp:
> IIUC, the error would happen when calling other methods on the object after it's already been destroyed. Is that correct @jmartinesp?

Yes, I think `destroy` would just no-op, but calling any other method would throw an exception.

I don't mind renaming this if you think `uniffiIsDestroyed` makes it less likely to have conflicts with existing methods.

bendk:
This is looking great to me.  Can you resolve that conflict on the changelog?  I'll merge as soon as that's ready.

PR #2827, opened 2026-02-23
Fail when we can't generate async constructors

### LINKED ISSUE #2804: Multiple constructors are not available on Android; default constructor is lost!

bendk:
Rather than silently succeeding, fail to generate code when a constructor is async and the language doesn't support it.  Users can exclude these constructors from the bindings if they want the old behavior (see #2822).

Fixes #2804.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 ## [[UnreleasedUniFFIVersion]] (backend crates: [[UnreleasedBackendVersion]]) - (_[[ReleaseDate]]_)
 
+### ⚠️ Breaking Changes ⚠️
+- Kotlin and Python now fail to generate bindings when there are async primary constructors.
+  Previously these languages skipped the constructor in this case or generated a constructor that always threw.
+  You can get similar behavior by adding the primary constructor to the uniffi.toml excludes list in uniffi.toml (e.g. `excludes = ["MyObject.new"])
+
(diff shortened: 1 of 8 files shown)


bendk:
I "tested" this by running the fixtures/futures test cases.  The Python/Swift tests did fail at codegen time with the expected message before I added the exclusions.

bendk:
> lgtm. I expect a ui-test could be made to work here if we cared enough?

That'd be nice, but we can't just plug this into `trybuild` because it requires running uniffi-bindgen.  I'm feeling like we can just leave this as-is for now.

PR #2831, opened 2026-02-26
Add a new `mutable_records` configuration

crazytonyli:
Allow specific records to remain mutable even when `generate_immutable_records` is enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - `Box<T>` now automatically implements FFI traits when `T` implements them, allowing direct use in enum variants and function parameters without NewType wrappers ([#2808](https://github.com/mozilla/uniffi-rs/pull/2808))
 - Record fields can now be renamed with the proc-macro `name = "new_field_name"` attribute ([#2794](https://github.com/mozilla/uniffi-rs/pull/2794))
 - Items can be excluded from the generated bindings using `uniffi.toml`.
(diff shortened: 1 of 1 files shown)


uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -202,10 +204,17 @@ impl Config {
...
-    pub fn generate_immutable_records(&self) -> bool {
+    fn generate_immutable_records(&self) -> bool {
         self.generate_immutable_records.unwrap_or(false)
     }
 
+    /// Whether a specific record should be generated with immutable fields.
+    /// A record is immutable only if `generate_immutable_records` is enabled
+    /// and the record is not listed in `mutable_records`.
+    pub fn is_record_immutable(&self, name: &str) -> bool {
+        self.generate_immutable_records() && !self.mutable_records.contains(name)
```
bendk, line 215: I think this is going to fail if the record is renamed via the renaming config, but maybe we can just live with this for now.  I think it would be easier to solve if we moved these bindings to using the pipeline code.

I'm going to add a note to the renaming doc that explains this, since it doesn't just affect this feature it also affects any config that targets a type name.

bendk (changes requested):
This looks awesome.  Can you add a CHANGELOG entry?  Other than that it seems great.

bendk:
Here's the PR about renaming: https://github.com/mozilla/uniffi-rs/pull/2832.  It focuses on the custom type configuration, but it's written in a general way that also applies to this new config option.

PR #2808, opened 2026-01-31
feat: Implement Lift/Lower traits for Box<T>

WaterWhisperer:
Automatically implement FFI traits for `Box<T>` when `T` implements them. This allows using `Box<T>` directly in enum variants and function parameters without needing NewType wrappers.

Fixes #2665

PR #2834, opened 2026-03-01
Add support for recursive enums

### LINKED ISSUE #396: Add support for recursive enums

jhugman:
Fixes #396 

Partly motivated by wanting to understand the new pipeline architecture, and partly because I want to be able to: 

- represent JSON without serializing/deserializing, or
- send a render tree to a Jetpack Compose or Swift UI player.

This PR adds support for recursive enums.

Things I've learned:

- there's a general pipeline, and language specific pipelines.
- sharing between the `ComponentInterface` and pipeline is possible, but they only share the `Type` enum. Pretty much everything is different.
- Adding things to UDL and procmacros is not worth it, if you're calculating what the right thing to do is.

I had considered adding a `recursive = true` procmacro, but:
- validating mutually recursive enums wasn't possible in the front end.
- there had to be a UDL version too.
- validating that the user had explicitly marked all their enums correctly meant calculating the single unique arrangement that was possible. Given only one possible labeling was possible, then auto-labelling as recursive stops being "magical", and more like "just remove this friction".

The design I settled on: 

- calculate a graph of enums that depend on one another. Graphs have enum types as nodes, and edges are "contained in via Box, Vec, Map, Record".
- detect cycles in that graph; so we can do self-recursive and mutually recursive enums.
- then mark the enums in those cycles as "recursive".
- the templates can then render the correct code accordingly.

I've shared the "detect cycles", `find_recursive_enum_names` between the two backends, and re-implemented the `enum_dep_graph` and `mark_recursive_enums` for each of `interface` and `pipeline`.

Unit tests, method structure should all look the same between the two implementations.

For integration tests I've loaded as many positive cases in to the Python, Swift and Kotlin cases as I can think of.

For the templates:
- For Swift, we just need to add `indirect` for recursive enums. 
- For Python, we need to switch to using Forward Refs instead of type names, i.e. `"Enum"` and `typing.Optional("Enum")`.
- For Kotlin, only a change to help kotlinc infer a return type more easily was necessary.
- I haven't looked at the Ruby version, but I can't imagine it will need any extra work.

### DIFF
--- a/uniffi_bindgen/src/bindings/python/pipeline/context.rs
+++ b/uniffi_bindgen/src/bindings/python/pipeline/context.rs
@@ -11,8 +11,7 @@ pub struct Context {
     pub cdylib: Option<String>,
     pub current_config: Option<PythonConfig>,
     pub module_namespace: Option<String>,
-    pub recursive_enum_names: HashSet<String>,
-    pub recursive_record_names: HashSet<String>,
+    pub recursive_type_names: HashSet<String>,
 }
 
 impl Context {
@@ -26,45 +25,23 @@ impl Context {
             None => PythonConfig::default(),
         });
         self.module_namespace = Some(namespace.name.clone());
-        self.recursive_enum_names = namespace
-            .type_definitions
-            .iter()
-            .filter_map(|td| {
-                if let general::TypeDefinition::Enum(e) = td {
-                    if e.recursive {
-                        Some(e.name.clone())
-                    } else {
-                        None
-                    }
-                } else {
-                    None
-                }
-            })
-            .collect();
-        self.recursive_record_names = namespace
-            .type_definitions
-            .iter()
-            .filter_map(|td| {
-                if let general::TypeDefinition::Record(r) = td {
-                    if r.recursive {
-                        Some(r.name.clone())
-                    } else {
-                        None
-                    }
-                } else {
-                    None
-                }
-            })
-            .collect();
+        let mut recursive_type_names = HashSet::new();
+        namespace.visit(|e: &general::Enum| {
+            if e.recursive {
+                recursive_type_names.insert(e.name.clone());
+            }
+        });
+        namespace.visit(|r: &general::Record| {
+            if r.recursive {
+                recursive_type_names.insert(r.name.clone());
+            }
+        });
+        self.recursive_type_names = recursive_type_names;
         Ok(())
     }
 
-    pub fn is_recursive_enum(&self, name: &str) -> bool {
-        self.recursive_enum_names.contains(name)
-    }
-
-    pub fn is_recursive_record(&self, name: &str) -> bool {
-        self.recursive_record_names.contains(name)
+    pub fn is_recursive(&self, name: &str) -> bool {
+        self.recursive_type_names.contains(name)
     }
 
     pub fn module_namespace(&self) -> Result<&str> {
--- a/uniffi_bindgen/src/bindings/python/pipeline/types.rs
+++ b/uniffi_bindgen/src/bindings/python/pipeline/types.rs
@@ -93,21 +93,8 @@ pub fn type_name(ty: &Type, context: &Context) -> Result<String> {
 /// body (needed for Python < 3.10 which evaluates annotations eagerly).
 pub fn type_annotation(ty: &Type, context: &Context) -> Result<String> {
     match ty {
-        Type::Enum { name, .. } => {
-            let base = type_name(ty, context)?;
-            if context.is_recursive_enum(name) {
-                Ok(format!("\"{}\"", base))
-            } else {
-                Ok(base)
-            }
-        }
-        Type::Record { name, .. } => {
-            let base = type_name(ty, context)?;
-            if context.is_recursive_record(name) {
-                Ok(format!("\"{}\"", base))
-            } else {
-                Ok(base)
-            }
+        Type::Enum { name, .. } | Type::Record { name, .. } if context.is_recursive(name) => {
+            Ok(format!("\"{}\"", type_name(ty, context)?))
         }
         Type::Optional { inner_type } => Ok(format!(
             "typing.Optional[{}]",


fixtures/recursive-enums/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+        Trie::Leaf(v) => v,
+        Trie::Branch { children } => children.into_values().map(|t| trie_sum(*t)).sum(),
+    }
+}
+
+// Cycle through a record: RoseTree → RoseData → Vec<RoseTree>
+#[derive(uniffi::Enum)]
+pub enum RoseTree {
+    Leaf(i32),
+    Branch(Box<RoseData>),
```
bendk: Does this need to be a box?  I thought the `Vec` in `RoseData.children` would be enough to break the cycle.

uniffi_bindgen/src/bindings/python/pipeline/nodes.rs
```diff
@@ -405,6 +406,8 @@ pub struct ExternalType {
 pub struct TypeNode {
     #[map_node(types::type_name(&self.ty, context)?)]
     pub type_name: String,
+    #[map_node(types::forward_ref_type_name(&self.ty, context)?)]
+    pub forward_ref_type_name: String,
```
bendk: Maybe this could be live the `Field` struct, maybe named something like `Field::type_annotation`.  Right now it feels like it could be hard to know when you should use this vs `type_name`.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1179,6 +1180,42 @@ impl ComponentInterface {
...
+        let recursive =
+            crate::pipeline::general::infer_recursive_enums::find_recursive_enum_names(&deps);
+        self.mark_recursive_enums(&recursive);
+    }
+
+    /// Build the structural-containment graph for `find_recursive_enum_names`.
+    ///
+    /// Each key is an enum name; its value is the set of enum names reachable
+    /// through its variant fields (unwrapping `Optional`/`Sequence`/`Map` wrappers
+    /// and crossing `Record` boundaries).
```
bendk: At least for Rust, going through `Map` or `Sequence` breaks the cycle in the sense that you don't need to box anything since `HashMap` and `Vec` are already doing a heap allocation.  How does this work for other languages?  My worry here is that different languages deal with it differently, so auto-detection won't work.

Also, doesn't only 1 of the types need to be marked as recursive to break the cycle?  I think that raises a similar question.

bendk: Maybe we can take conservative approach that works for all languages.  It's probably not that big a deal if more fields are marked `indirect` than need to be.  That said, I think we should note this and have some sort of plan for improvement.

jhugman: Yeah, I think the consequences of a false-positive (labeling something `indirect` when it isn't) is Swift puts the enums on the heap when it doesn't need to be; but a false-negative is compile time failure.

jhugman:
> In Rust, if 2 types refer to each other either type can be boxed and the other type doesn't have to. 

Ooh. TIL.

> I'm not sure about auto-detecting these cycles, since it seems like there are human decisions to be made. 

Yeah, I thought you wouldn't like auto-detection. I did consider the everything explicit approach, but adding keywords to UDL _and_ procmacros, as well as two backends didn't seem worth it for 

> I also wonder if different languages have different requirements, so there's some language-specific configuration that's needed here. I'm not really sure on either of those points though, WDYT?

Yes, that's a good idea.

Recursive enums types are a non-issue for Kotlin, Ruby, Typescript, Go, Java or Dart.

I think there's a need for some kind of cycle breaking (automatic or manual) in Rust, Swift and C++.

Forward references are only needed for Python: none of the other languages need them.

For a cycle of enum types like: Rock -> Paper -> Scissors -> Rock:

- In Rust, we need to put the associated value of a variant on the heap to break the cycle—with a `box`, or `Option`, `Vec`, `HashMap` etc. It would be a whole lot easier if we `Box` was part of the IR/`ComponentInterface`.
- In Swift, we can label any one of these _types_ `indirect` to break the cycle. I think you can do variant level labelling too.
- For C++, we need to break the cycle at the variant _field_ level, with a `std::shared_ptr` or similar. But we _also_ need to do the same for `Record`s.

So, I'm thinking:

- if we can label `Record`, then we can help [the C++ backend](https://github.com/NordSecurity/uniffi-bindgen-cpp/pull/59).
- if uniffi can detect and record where the `Box<T>` happens (which variant), OR we get the user to double label the variant (see below), then we can generate the exact same shape as Rust.
- if we can't persuade uniffi to track which is boxed and which is not, OR we don't do any procmacro/UDL work, then we label everything in the cycle as recursive. It would mean slightly more than necessary on the foreign heap, but conservative and guaranteed correct.

```rust
#[derive(uniffi::Enum)]
enum Rock {
  #[uniffi(indirect)]
  Beats(Box<Scissors>)
  LosesTo(Paper)
}

#[derive(uniffi::Enum)]
enum Paper {
  Beats(Rock)
  #[uniffi(indirect)]
  LosesTo(Box<Scissors>)
}

#[derive(uniffi::Enum)]
enum Scissors {
  Beats(Paper)
  LosesTo(Rock
}
```

Given that cycle detection is deterministic, and the consequence for over-labeling is low to zero, I'm inclined to leave any work on the UDL/procmacros for another PR.

I'd also be really interested in [@larkdoo](https://github.com/Larkooo)'s P.O.V.

WDYT

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1179,6 +1180,42 @@ impl ComponentInterface {
...
+                    .collect();
+                (e.name().to_string(), deps)
+            })
+            .collect()
+    }
+
+    fn mark_recursive_enums(&mut self, recursive: &HashSet<String>) {
+        for e in &mut self.enums {
+            if recursive.contains(e.name()) {
+                e.recursive = true;
```
mhammond: I don't really think the bool on enum metadata is correct - the thing creating metadata never supplies it, but instead it's just used internally here iiuc. ie, should the interface just own the hashset?

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -1341,6 +1378,57 @@ fn throws_name(throws: &Option<Type>) -> Option<&str> {
...
+        Type::Map {
+            key_type,
+            value_type,
+        } => {
+            enum_names_in_type_inner(key_type, records, visited_records);
+            enum_names_in_type_inner(value_type, records, visited_records)
+        }
+        Type::Record { name, .. } => {
+            // Guard against mutually-recursive records.
+            if !visited_records.insert(name.clone()) {
```
mhammond: what is the plan for records?

jhugman:
> Seeing we now have direct support for boxed types, I'm kinda surprised there's not a simpler approach we could take here. Sadly though I've not spent much time thinking about it, so I could be way off base.

Yeah, my read of [#2808](https://github.com/mozilla/uniffi-rs/pull/2808/changes#diff-94063caa1d5fea5835eeb0c519f8f812034e3ac4c76ea0be0aa93dc60786adbcR323-R325) was that the Box is completely erased in the IR.

uniffi_bindgen/src/interface/mod.rs
jhugman: Having worked through [the example of C++](https://github.com/mozilla/uniffi-rs/pull/2834#issuecomment-3987718999), I think I'm inclined to mark them as recursive too.

bendk:
I'm fine with the "conservative approach" of making any type involved in a cycle indirect.  This seems like a fine starting point that we can improve in the future.

Maybe we could do this by allowing the bindings to see which types were boxed.  I think it would work great on Swift to make any boxed type indirect.  I wonder if all bindings will want to follow the same rules though.  Maybe we could have binding-specific configuration for that.

uniffi_bindgen/src/bindings/python/pipeline/nodes.rs
jhugman: oh, nice!

uniffi_bindgen/src/bindings/python/pipeline/context.rs
```diff
@@ -2,13 +2,17 @@
...
+
 use super::*;
 
 #[derive(Default, Clone)]
 pub struct Context {
     pub cdylib: Option<String>,
     pub current_config: Option<PythonConfig>,
     pub module_namespace: Option<String>,
+    pub recursive_enum_names: HashSet<String>,
+    pub recursive_record_names: HashSet<String>,
```
bendk: This could just be recursive names, since type names need to be unique.

uniffi_bindgen/src/interface/mod.rs
bendk: I'm not sure if `interface::Enum` needs to match `uniffi_meta::EnumMetadata`, but I do like storing a hash set.  There's already the `errors` and `callback_interface_throws_types` sets, so it feels more consistent.

uniffi_bindgen/src/bindings/python/pipeline/context.rs
```diff
@@ -22,9 +26,47 @@ impl Context {
...
+                    if e.recursive {
+                        Some(e.name.clone())
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                }
+            })
+            .collect();
```
bendk: I think the visit method could work great here:

```
namespace.visit(|e: Enum| {
  if e.recursive {
     self.recursive_enum_names.push(e.name.clone());
   }
});
```

jhugman: Ooh, I **like** this.

jhugman:
@bendk ready for your re-review.

mhammond:
I still haven't dug in detail, but I don't see a need to given all the discussions above, so lgtm

@bendk I'm mildly surprised we went with a new "fixture" rather than the new "bindgen-tests" mechanism - for my better understanding of that new capability, is there a reason for that other than momentum?

bendk:
> @bendk I'm mildly surprised we went with a new "fixture" rather than the new "bindgen-tests" mechanism - for my better understanding of that new capability, is there a reason for that other than momentum?

That's a great point and I flaked on this one.  One issue is that we need start the Kotlin/Python code.  We don't need to implement everything, but at least enough to run the test.  I'm going to try to see if I can get this move over.  The good news is I don't think the test itself needs any refactoring to move over to those tests.

PR #2841, opened 2026-03-18
Parse metadata from Rust source directly

bendk:
Added a new crate that generates the metadata by parsing the Rust source with syn.  Updated `bindgen-tests` to use the new system. Right now only bindings generation is supported, but I'm hoping to also use this to generate scaffolding.

This brings several nice features and some drawbacks compared to the current procmacro code:

* Good, because we can generate things without building the interface
* Good, because we could generate the scaffolding using the same pipeline as the bindings which allows for more shared logic.
* Good, because scaffolding generation can see the entire interface:
  * We could go back to a single checksum function again
  * I believe this removes the need to special case external/remote types with macros like `use_remote_type!`.  Maybe we could parse dependency crates and avoid the need to duplicate type definitions in `#[remote]`.
  * This could simplify uniffi-bindgen-gecko-js and similar bindgens, where we need to generate a single entrypoint function that dispatches to the real function.
* Good, because it can enable language-specific scaffolding generation. It's not clear how an external bindings generator could hook into the `uniffi_macros` code.  With this change, they can use `uniffi_parse_rs` to parse the Rust sources and generate language-specific code from there.
* Bad, because we don't evaluate macros.  Users would need to define all their UniFFI types in "regular Rust".
* Bad, because parsing Rust from scratch adds complexity.

See the README.md and DESIGN.md docs for details.

Parsing is pretty quick at this point.  I was able to parse all the UniFFI fixtures/examples in 170ms.  Parsing just `uniffi-bindgen-tests` and `uniffi-bindgen-tests-external-types-source` took 15ms.


bendk:
I started with the intentions of making a POC, but then I kept running this against real-world code and fixing the parse errors that I saw.  `uniffi_parse_rs` is huge, but it's mostly straightforward macro parsing code and tests.  The main interesting modules are the ones that deal with type resolution (`paths.rs` and `types.rs` and `use_.rs`).  If we want to merge this, I still need to write a CHANGELOG entry and update the manual.

uniffi_bindgen/src/bindgen_paths.rs
```diff
@@ -68,6 +80,40 @@ impl BindgenPaths {
...
+    ///
+    /// This is the entrypoint `.rs` file, typically `[crate-root]/src/lib.rs.
+    pub fn get_rust_source_path(&self, crate_name: &str) -> Option<Utf8PathBuf> {
+        self.layers
+            .iter()
+            .find_map(|l| l.get_rust_source_path(crate_name))
+    }
+
+    /// Get a list of enabled features for a crate.
+    pub fn get_crate_features(&self, crate_name: &str) -> Option<Vec<String>> {
```
bendk: This doesn't makes so much sense in a type named `BindgenPaths`.  Maybe we should rename this to something like `BindgenInputs`.

uniffi_parse_rs/src/expect/error_through_type_alias.txt
```diff
@@ -0,0 +1,15 @@
...
+   │                            ---------------------- while resolving type
+   ·
+10 │     pub type Result<T> = Result2<T>;
+   │                          ---------- while resolving type
+11 │ 
+12 │     pub type Result2<T, E=Error> = std::result::Result<T, E>;
+   │                           ^^^^^ Unknown item
+   │
+   = Do you need to add a UniFFI derive or uniffi::export?
+   = Does the item come from a non-UniFFI crate?
```
bendk, line 14: I used the `codespan-reporting` crate for error messages and it generates some really nice ones that came in handy while debugging things.

examples/remote-types/src/lib.rs
```diff
@@ -45,7 +45,7 @@ impl LogSink {
     }
 
     /// Log a message to our file
-    pub fn log(&self, _level: log::Level, _msg: String) {
+    pub fn log(&self, _level: LogLevel, _msg: String) {
```
bendk, line 48: This was the only code change I needed to get `uniffi_parse_rs` to parse all of our examples/fixtures.  Since the type is named `LogLevel` in the `#[remote_type]` macro, we need to use that same name in the code.  I also tested parsing a bunch of crates in application-services and these were the only [changes](https://github.com/mozilla/application-services/commit/f62837252a7d2a3096fe4d83d86324f1ced60857) I needed to make there as well.

bindgen-tests/lib/uniffi.toml
```diff
@@ -1,3 +1,5 @@
+[bindings.swift]
+omit_checksums = true
```
bendk: The parser doesn't generate checksums, since those are tied to the macro code.  I think it's possible but I'd rather not do it.  Instead, I'd rather investigate using this for scaffolding generation.  That would allow us to improve the checksum system by just generating a single checksum function for the entire interface.

bendk:
This PR was really hard to understand because it was huge and everything was squash in together.  I split it up into a separate logical commits, which should be easier to review.  It's still a lot of code, but hopefully it's more approachable.

uniffi_parse_rs/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+pub use items::{BuiltinItem, Item};
+pub use modules::Module;
+pub use objects::{Constructor, Method, Object};
+pub use records::{Field, Record};
+pub use traits::Trait;
+
+pub use anyhow::Result;
+pub type MetadataGroupMap = HashMap<String, MetadataGroup>;
+
+/// Top-level API for `rust_parse_rs`
```
badboy:
```suggestion
/// Top-level API for `uniffi_parse_rs`
```

uniffi_parse_rs/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+///
+/// This parses Rust sources and creating a `MetadataGroupMap` from them.
```
badboy:
```suggestion
/// This parses Rust sources and creates a `MetadataGroupMap` from them.
```

uniffi_parse_rs/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+/// All failable methods return `anyhow::Result`
+/// since that integrates better with the rest of the UniFFI.
```
badboy:
```suggestion
/// to integrate better with the rest of the UniFFI.
```

uniffi_parse_rs/src/ir.rs
```diff
@@ -0,0 +1,100 @@
...
+        self.crate_roots.insert(ident.clone(), Item::Module(module));
+        match self.crate_roots.get(&ident) {
+            Some(Item::Module(m)) => Ok(m),
+            _ => unreachable!(),
+        }
+    }
+
+    pub fn crate_root(&self, ident: &Ident) -> Option<&Module> {
+        self.crate_roots.get(ident).map(|item| match item {
+            Item::Module(item) => item,
```
badboy: Just a note for my understanding.
I wondered why we even wrap `Module` in `Item::Module`, when the code path above is the only one inserting it. 
But I guess just using `Module` causes problems further down the line where other code handles the general `Item` case.

uniffi_parse_rs/src/objects.rs
```diff
@@ -0,0 +1,139 @@
...
+        Ok(Self { attrs, ident })
+    }
+
+    pub fn name(&self) -> String {
+        self.ident.unraw().to_string()
+    }
+
+    pub fn obj_metadata(&self) -> Result<uniffi_meta::ObjectMetadata> {
+        Ok(uniffi_meta::ObjectMetadata {
+            module_path: "".into(), // TODO
```
badboy: Elaborate on this TODO. Where would `module_path` coming from?

uniffi_parse_rs/src/enums.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+impl Variant {
+    pub fn parse(
+        enum_attrs: &EnumAttributes,
+        attrs: VariantAttributes,
+        v: syn::Variant,
+    ) -> syn::Result<Self> {
+        let discr = Self::parse_discriminent(&v, enum_attrs)?;
```
badboy:
```suggestion
        let discr = Self::parse_discriminant(&v, enum_attrs)?;
```

uniffi_parse_rs/src/enums.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        Ok(Self {
+            attrs,
+            ident: v.ident,
+            discr,
+            fields,
+        })
+    }
+
+    fn parse_discriminent(
```
badboy:
```suggestion
    fn parse_discriminant(
```

uniffi_parse_rs/src/enums.rs
```diff
@@ -0,0 +1,157 @@
...
+                    lit.base10_parse()?,
+                    uniffi_meta::Radix::Decimal,
+                    discr_type.clone(),
+                ))),
+                Lit::Int(lit) => Ok(Some(LiteralMetadata::UInt(
+                    lit.base10_parse()?,
+                    uniffi_meta::Radix::Decimal,
+                    discr_type.clone(),
+                ))),
+                _ => Err(syn::Error::new(discr.span(), "Invalid discriminent")),
```
badboy:
```suggestion
                _ => Err(syn::Error::new(discr.span(), "Invalid discriminant")),
```

uniffi_parse_rs/src/enums.rs
```diff
@@ -0,0 +1,157 @@
...
+                            -(lit.base10_parse()?),
+                            uniffi_meta::Radix::Decimal,
+                            discr_type.clone(),
+                        ))),
+                        _ => Err(syn::Error::new(discr.span(), "Invalid discriminent")),
+                    },
+                    _ => Err(syn::Error::new(discr.span(), "Invalid discriminent")),
+                }
+            }
+            _ => Err(syn::Error::new(discr.span(), "Invalid discriminent")),
```
badboy:
```suggestion
                        _ => Err(syn::Error::new(discr.span(), "Invalid discriminant")),
                    },
                    _ => Err(syn::Error::new(discr.span(), "Invalid discriminant")),
                }
            }
            _ => Err(syn::Error::new(discr.span(), "Invalid discriminant")),
```

uniffi_parse_rs/src/errors.rs
```diff
@@ -0,0 +1,220 @@
...
+    InvalidArgType,
+    #[error("Enum and Error derives are mutually exclusive")]
+    MultipleEnumDerives,
+    #[error("Invalid attribute")]
+    InvalidAttr,
+    #[error("Invalid repr type")]
+    InvalidRepr,
+    #[error("Invalid discriminant")]
+    InvalidDiscr,
+    #[error("invalid file path: {0}")]
```
badboy:
```suggestion
    #[error("Invalid file path: {0}")]
```

you used uppercase for all others.

uniffi_parse_rs/src/test_src/results.rs
```diff
@@ -0,0 +1 @@
+
```
badboy: What's the reason this is empty?
If there is one maybe it should be here in a comment `/* intentionally left empty because ... */`

uniffi_parse_rs/src/enums.rs
```diff
@@ -46,10 +47,15 @@ impl Enum {
...
+    pub fn enum_metadata<'ir>(
+        &self,
+        ir: &'ir Ir,
+        cache: &mut LookupCache<'ir>,
+        path: &RPath<'ir>,
+    ) -> Result<uniffi_meta::EnumMetadata> {
         let is_flat = matches!(self.attrs.shape, EnumShape::Error { flat: true });
         Ok(uniffi_meta::EnumMetadata {
-            module_path: "".into(), // TODO
+            module_path: path.path_string(),
```
badboy: Guess that solves any of my previous "what does this TODO mean?" comments

uniffi_parse_rs/src/impls.rs
```diff
@@ -38,14 +41,55 @@ impl Impl {
         })
     }
 
-    pub fn impl_metadata(&self) -> Result<Vec<uniffi_meta::Metadata>> {
-        let self_name = "TraitName"; // TODO
+    pub fn self_type_and_name<'ir>(
```
badboy: I don't see any use of this function across the code base.

uniffi_parse_rs/src/paths.rs
```diff
@@ -0,0 +1,788 @@
...
+        Self {
+            items: vec![crate_root],
+        }
+    }
+
+    /// Get the item this path points to
+    pub fn item(&self) -> &'ir Item {
+        self.items
+            .last()
+            .unwrap_or_else(|| panic!("UniFFI internal error in Path::item(): items is empty"))
```
badboy: that's just expect, isn't it?

```suggestion
            .expect("UniFFI internal error in Path::item(): items is empty"))
```

uniffi_parse_rs/src/paths.rs
```diff
@@ -0,0 +1,788 @@
...
+
+    pub fn module(&self) -> &'ir Module {
+        self.items
+            .iter()
+            .rev()
+            .find_map(|i| match i {
+                Item::Module(m) => Some(m),
+                _ => None,
+            })
+            .unwrap_or_else(|| panic!("UniFFI internal error in Path::module: no modules found"))
```
badboy: as above: expect

uniffi_parse_rs/src/paths.rs
```diff
@@ -0,0 +1,788 @@
...
+    }
+
+    pub fn crate_root(&self) -> &'ir Module {
+        match self.items.first() {
+            Some(Item::Module(m)) => m,
+            i => panic!(
```
badboy: guess here expect wouldn't work, because we can't interpolate `i` then

uniffi_parse_rs/src/paths.rs
```diff
@@ -0,0 +1,788 @@
...
+            fn span(&self) -> Span {
+                match self {
+                    Self::Use(item_use, _) => item_use.span(),
+                    Self::Item(ident, _) => ident.span(),
+                }
+            }
+            fn matches_use(&self, path: RPath<'_>) -> bool {
+                match self {
+                    Self::Use(_, p) => {
+                        println!("{p} - {path} {}", *p == path);
```
badboy: stray println?

uniffi_parse_rs/src/test_src/name_conflicts.rs
```diff
@@ -40,3 +40,14 @@ mod mod3 {
 
 use mod3::*;
 use mod3::mod4::*;
+
+// Complete corner case: record that shares the same name with the custom_type! macro.
+//
+// This is not an error in Rust, since macros live in a different namespace.  It's doesn't need to
```
badboy:
```suggestion
// This is not an error in Rust, since macros live in a different namespace.  It doesn't need to
```

uniffi_parse_rs/src/errors.rs
```diff
@@ -47,6 +47,8 @@ pub enum ErrorKind {
     TraitWithoutDyn,
     #[error("Expected trait")]
     ExpectedTrait,
+    #[error("Invalid builtin type")]
```
badboy, line 50: should this error message say anything about being from custom types or will that be clear from the full error message?

uniffi_parse_rs/DESIGN.md
```diff
@@ -0,0 +1,42 @@
+# Passes
+
+`uniffi_parse_rs` is designed to parse Rust source code in 2 passes:
+
+The first pass focusing on parsing the Rust syntax and only applies local reasoning.
```
badboy:
```suggestion
The first pass focuses on parsing the Rust syntax and only applies local reasoning.
```

uniffi_parse_rs/DESIGN.md
```diff
@@ -0,0 +1,42 @@
...
+For example, `syn::Type` is not parsed, since that would require resolving the type path
+to find the struct/enum with a UniFFI derive.
+Failable functions for this pass generally return `syn::Result`,
+which simplifies things since we're often calling syn functions at this point.
+
+The second pass converts the IR from the first pass into `uniffi_meta` types.
```
badboy, line 12: The above paragraph should mention what it outputs, namely an IR (Intermediate Representation) of the somewhat-parsed code.

uniffi_parse_rs/DESIGN.md
```diff
@@ -0,0 +1,42 @@
...
+which requires resolving the type paths (e.g. determining that `foo::bar::Baz` is a Record).
+Failable functions for this pass generally return `crate::Result`,
+which allows us to report better errors by tracking path resolution context.
+For example, if a type alias is involved, then we want to also show the user
+that the error happened when evaluating that type aliases.
+
+This split has several advantages:
+  * It's easier to resolve types after performing the first pass.
+  * The code for the first pass can be integrated with `uniffi_macros`
+    since it only parses one item at a time and it uses `syn::Result`.
```
badboy: For proper formatting a newline might be required, I think.

```suggestion
This split has several advantages:

* It's easier to resolve types after performing the first pass.
* The code for the first pass can be integrated with `uniffi_macros`
  since it only parses one item at a time and it uses `syn::Result`.
```

uniffi_bindgen/src/bindings/swift/mod.rs
```diff
@@ -145,8 +145,12 @@ pub fn generate_swift_bindings(options: SwiftBindingsOptions) -> Result<()> {
     #[cfg(feature = "cargo-metadata")]
     let mut paths = BindgenPaths::default();
 
+    // TODO: handle cargo target/ features
```
badboy: Not sure I fully understand what this todo is about. For future problem-solving it might be worth slightly expanding on it.

badboy (commented):
Amazing work!
Code-wise this looks ok (I had to skip through the last commit, the final uniffi integration, a bit due to time)
It's a whole lot of code, but it's the verbose syn-to-uniffi translation that blows it up so much. That's all very mechanical.
Thanks for splitting it up in commits, that made the review tractable.
Note that I reviewed and commented commit-by-commit, so it might be that some comments are invalid with later changes.

With that out of the way I do have some more thoughts on this.

With this we now have another place that we need to keep in sync with the macros or other changes.
I'm not sure how much maintenance overhead that is.

That said, it is yet another mode. Will we support both this and the library parsing? 
That's a lot of different things we have to support for effectively the same outcome.

Obviously this could use some docs eventually, but hold off on that until we decide to land it.

I would like to take this for a spin with Glean, I can get to that in like 2 weeks when I'm back.
Meanwhile others definitely should have a look at it (cc @mhammond, @gruberb), not necessarily for a full code review, but for input on the approach.

badboy:
Oh maybe missing from my above comment: I think I am in favor of this approach, given it probably supports all we need for Glean. The library mode has its own challenges, like the upfront library build of the crate, which parsing source code doesn't have.

uniffi_parse_rs/src/ir.rs
bendk: It's for the `RPath` type which wants to treat a path as a vec of Items `foo::bar::Baz` becomes something like `[&Item::Module(_), &Item::Module(_), &Item::Record(_)]`.  The only way I could make that work without cloning everything was to wrap all the modules in `Item`.

uniffi_parse_rs/src/errors.rs
bendk: It will highlight the type and I was thinking the user should be able to figure it out from there.  I could add a reason field though, WDYT?

bendk:
> With this we now have another place that we need to keep in sync with the macros or other changes.
I'm not sure how much maintenance overhead that is.
> .... Will we support both this and the library parsing?
> That's a lot of different things we have to support for effectively the same outcome.

This is a really good point, I think it's the key question.

Maybe we could support both.   `uniffi_macros` could be refactored to use the first-pass code and that would reduce the amount of duplication.

If we only have to support one, I'm kind of in favor of this approach.  It comes with a lot of benefits, in particular I like the ability to see the entire interface when generating the scaffolding, being able to generate it without building the library, and handling external/remote types better.  OTOH, this can't handle macros like the current system can and maybe we'll find out about other issues.

I'm not sure what the right path is.  Maybe we call this "experimental", ask users to try it, and see what issues come up for them.

bendk:
I forgot one of the biggest benefits to this approach: it enables language-specific bindgens.  You can have a `uniffi-bindgen-kotlin` that parses the Rust code and generates JNI scaffolding.  I can't think of a good way to do that with the current macro system.

bendk:
Pushed a bunch of changes, most of them are in https://github.com/mozilla/uniffi-rs/pull/2841/changes/cd92bb484a36716e3b8e96532fb5923414170c10.  `uniffi_parse_rs` now find the path to your types, so you no longer need to re-export them in the top-level module.

bendk:
Pushed some more changes to handle corner cases with `orig_name`, public paths, and renames via `use` statements.  This is now based on #2868 since we needed that to fix these corner cases.

bindgen-tests/lib/uniffi.toml
mhammond: fair enough - but why omit them for all languages? Should this comment go into that file? And finally, re-using `kotlin` I wonder if it is a bit limiting for people looking to use both (eg, while transitoning or just exploring?) OTOH, I guess the "global config" stuff would make that easier (ie, they could just use an entirely different toml file with the new bindgen)

examples/remote-types/src/lib.rs
mhammond: I'm having trouble keeping up with these PRs, but is that limitation explained in the docs?

bindgen-tests/lib/src/lib.rs
```diff
@@ -132,13 +132,19 @@ pub mod test_util {
...
     }
 
-    pub fn generate_sources(library_dir: &Utf8Path, language: uniffi::TargetLanguage) {
-        let library_filename = library_filename();
+    pub fn generate_sources(tempdir: &Utf8Path, language: uniffi::TargetLanguage) {
+        let source = match language {
+            uniffi::TargetLanguage::Kotlin => "src:uniffi-bindgen-tests-kotlin",
+            uniffi::TargetLanguage::Python => "src:uniffi-bindgen-tests-python",
+            uniffi::TargetLanguage::Swift => "src:uniffi-bindgen-tests-swift",
+            uniffi::TargetLanguage::Ruby => "src:uniffi-bindgen-tests-ruby",
```
mhammond, line 140: how do we ensure we maintain test coverage for library mode?

uniffi/src/cli/uniffi_bindgen.rs
```diff
@@ -100,6 +105,22 @@ enum Commands {
         /// all sub-dependencies causes obscure platform specific problems.
         #[clap(long)]
         metadata_no_deps: bool,
+
```
mhammond, line 108: kinda wondering if we should just support a new CLI entirely? Some of these flags only working for some ways of creating the bindgen vs some only making sense with the rust parsing one seems like it's not going to scale very well. I really haven't thought that through though, so have no idea of what that would involve and other tradeoffs it might cause.

OTOH, I see these are also used by cargo-metadata, which is something we should have been doing anyway, right?

uniffi_parse_rs/src/expect/invalid_type.txt
```diff
@@ -0,0 +1,12 @@
+error: Unknown item
```
mhammond, line 1: I'm yet to discover exactly what these are for, but assuming tests, so having then under `src/` seems a little odd?

uniffi_parse_rs/src/test_src/feature_detection.rs
```diff
@@ -0,0 +1,49 @@
+#[cfg(feature = "feature1")]
```
mhammond, line 1: similarly, while the directory does include `test`, does it need to be under `src`?

mhammond (commented):
This is a massive patch, and while I'll look at it again soon, I'm never going to be able to give it a very thorough review - but that's OK as the key stuff here is in a new crate, and it's clearly done with thought and tests (even if they seem to be in an odd location). I can't see any reason we can't merge this soon.

examples/remote-types/src/lib.rs
bendk: Good catch, I updated the README to reflect this.

bindgen-tests/lib/uniffi.toml
bendk: Yeah very good points.  I updated it so that we always skip checksum checks when using `uniffi_parse_rs` rather than force the user to put it in their config file.

My hope is that we can re-introduce checksums using a single global checksum function, like we had before (see https://github.com/mozilla/uniffi-rs/issues/2932).  I did a bit of prep work for that in the pipeline code by introducing the `ChecksumMode` enum.  I didn't do that for the component interface side, because I'm kind of hoping we move Swift and Kotlin over to the pipeline first.  Maybe that's optimistic, if so we'll need to introduce a similar enum over there.

uniffi_bindgen/src/bindings/python/templates/NamespaceLibraryTemplate.py
```diff
@@ -93,5 +77,9 @@ class {{ ffi_struct.name.0 }}(ctypes.Structure):
 {%- endfor %}
 
 {# Ensure to call the contract verification only after we defined all functions. -#}
-_uniffi_check_contract_api_version(_UniffiLib)
-# _uniffi_check_api_checksums(_UniffiLib)
```
bendk, line 97: Side note: when adding the checksum skipping code, I noticed that we already were silently skipping checksums in several places.  I tried to re-enable the checks where I could, mostly so the bindgen tests would be red before I implemented the checksum skipping code.

mhammond (approved):

PR #2854, opened 2026-03-30
fix: Resolve memory leak in rust_future_free by properly consuming handle

### LINKED ISSUE #2758: Memory leaks when Rust's async function called from Swift
Apple's profiling tool Instruments is reporting memory leaks on each async call from Swift to Rust.

I've tried investigating this but couldn't find the cause. Could anyone help take a look? 

## Minimal example:

``` rust
// rust side
#[uniffi::export]
async fn rust_side_function() -> u32 { 42 }
```

```swift
// swift side
Task {
    _ = await rustSideFunction()
}
```

## Observed Behavior

Each call to rustSideFunction() leaks 64 + 64 + 128 bytes (256 bytes total) on my machine (M4 Mac).

I've created [this repository](https://github.com/b01o/mre-uniffi-future-leak-swift) to demonstrate the example.

songhuaixu:
### Problem
When calling Rust async functions from Swift, memory leaks occur on each invocation (256 bytes per call on M4 Mac). This issue was first introduced in UniFFI 0.30.0.

Fixes #2758

### Root Cause
In `uniffi_core/src/ffi/rustfuture/mod.rs`, the `rust_future_free` function incorrectly uses `Handle::into_arc_borrowed` instead of `Handle::into_arc`.

- `into_arc_borrowed` is for APIs where the handle stays valid after the call: it calls `Arc::increment_strong_count` on the handle’s pointer, then `Arc::from_raw`. When the temporary `Arc` is dropped at the end of the call, the count returns to what it was, so the foreign side still holds the last strong reference.
- `rust_future_free` must **fully consume** that last reference. Using `into_arc_borrowed` here leaves **one extra strong count** on the allocation after `free` finishes. Nothing else drops it, so the `Arc` (and the whole `RustFuture`) is never deallocated.
- `into_arc` only does `Arc::from_raw` (no bump). It takes ownership of the handle’s single strong reference; dropping the `Arc` at the end of `free` drives the count to zero.

Because `RustFuture` owns two `Mutex` fields (and the rest of the future state), each leaked handle keeps that entire allocation alive—matching the steady per-call growth seen from Swift.

### Fix
Change the method call in `rust_future_free` to use `into_arc`, so the handle’s ownership is transferred and the allocation is dropped when the function returns.

```rust
// Before (0.30.0+)
pub unsafe fn rust_future_free<FfiType>(handle: Handle) {
    trace!("rust_future_free: {handle:?}");
    Handle::into_arc_borrowed::<RustFuture<FfiType>>(handle).free()
}

// After
pub unsafe fn rust_future_free<FfiType>(handle: Handle) {
    trace!("rust_future_free: {handle:?}");
    Handle::into_arc::<RustFuture<FfiType>>(handle).free()
}

### DIFF
--- a/uniffi_core/src/ffi/rustfuture/mod.rs
+++ b/uniffi_core/src/ffi/rustfuture/mod.rs
@@ -180,5 +180,5 @@ where
 /// The [Handle] must not previously have been passed to [rust_future_free]
 pub unsafe fn rust_future_free<FfiType>(handle: Handle) {
     trace!("rust_future_free: {handle:?}");
-    Handle::into_arc_borrowed::<RustFuture<FfiType>>(handle).free()
+    Handle::into_arc::<RustFuture<FfiType>>(handle).free()
 }


mhammond:
makes sense to me, but @bendk should review this.

QuentinPerez:
Makes sense, it looks like it was re-introduced during a refactoring, since it had already been fixed 2 years ago.
https://github.com/mozilla/uniffi-rs/pull/2056/changes

bendk (approved):
Thanks for the fix and sorry for the dumb mistake.

There's even a [test for this](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_core/src/ffi/rustfuture/tests.rs), but it only tests the code one level lower.  We should move that up one level.  I was actually planning on doing some refactoring soon, I'll try to update the tests then.

Qizot:
@bendk can we get a patch release with this fix?

bendk:
Yes, this seems like a good reason for a patch.  I'll start that now.

bendk:
To close the loop here, we released 0.31.1 on Monday with this fix.  Thanks again for the patch!

PR #2859, opened 2026-04-07
WIP: uniffi-bindgen-kotlin-jni

bendk:
This is a JNI-based Kotlin bindgen based around [uniffi_parse_rs](https://github.com/mozilla/uniffi-rs/pull/2841) and some new `FfiBuffer` code.  I'm in the middle of working through this, but I figure it's worth a share.  There's a few new ideas that I'm exploring here:

## Using `uniffi_parse_rs` to parse the Rust source directly. 

This means the macros are essentially no-ops, they're just used as markers so that `uniffi_parse_rs` can find the items to export.  Instead, the scaffolding is generated in a build script, like we used to do for UDL.

This allows us to use the pipeline code to generate the JNI, which feels nicer to me than doing this with the macros.  The main reason though is I wanted to find a way for external bindings generators to also generate language-specific scaffolding like this, and I couldn't think of a good way for them to hook into the macro code.  If this works out, I want to use the same system for `uniffi-bindgen-gecko-js`.

One exception to the "macros are no-ops" rule is custom types.  For those we really do need to generate code at the macro call site so that the expressions work correctly.

## FFI Buffers 

The FFI is currently implemented using a buffer to pass all arguments/return values.  See the module for details on how this works, but this feels like a good general system for passing values across the FFI.  One thing I like is that we always create exactly 1 buffer and use it for everything, which means cleanup is simply freeing that buffer in a `finally` block.

As discussed in #2841, it's not always the fastest way to pass values though.  Passing values using primitive arguments is going to out-perform this system.  My plan is to implement everything using buffers, then come back and try to find cases where we can improve performance using primitive arguments.  I think there should be a lot of opportunities and hopefully the optimized code is still fairly clean.

## No RustCallStatus

This code doesn't use `RustCallStatus` at all.  If we're using JNI, we can just create and throw exceptions directly, so let's use that instead.

## Single shared Kotlin package

All of the UniFFI internal code gets placed in the `uniffi` package.  This has some nice advantages:

- No UniFFI internal code in the consumer-facing packages
- Less duplication since the code for different Rust crates gets shared
- No need to worry about type errors because we define different `RustBuffer` type for each Rust crate.

## Less reliance on traits for scaffolding generation

Scaffolding generation doesn't use traits like `uniffi::FfiConverter`.  Instead it just generates functions and calls them.  This leads to some nice wins:

- Constructors and other functions that return interfaces can return either `Arc<T>` or `T` values.  We currently handle it by defining the `LowerReturn` trait and adding an [extra impl](https://github.com/mozilla/uniffi-rs/blob/d358fab0565116d90633ad89c98cb887d7367f23/uniffi_macros/src/object.rs#L226-L232) for the non-arc case.  In this branch, we define the `ArcOrOwned` trait and [input an `impl ArcOrOwned` value](https://github.com/mozilla/uniffi-rs/blob/3d826c71278b7faaf788729fb8cd5e9f668ae4a1/uniffi-bindgen-kotlin-jni/src/scaffolding/templates/class.rs#L21) to handle this.  I like the new system better, but more importantly I like that our FFI functions don't need to have exactly the same signature which is what enables this.
- Callback errors are currently handled using multiple traits plus [auto-ref specialization](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_core/src/ffi/callbackinterface.rs#L179) which is pretty weird and hard to understand.  In this branch, we know if an error type implement `From<UnexpectedUniFFIError>` or not, which lets us do the [specialization](https://github.com/mozilla/uniffi-rs/blob/3d826c71278b7faaf788729fb8cd5e9f668ae4a1/uniffi-bindgen-kotlin-jni/src/scaffolding/templates/callback_interface.rs#L92-L104) in a much more straightforward way.

(Re-reading this, I'm not sure if it totally makes sense.  I guess the main point is I didn't feel like I needed to come up with complex type/trait arrangements so much with this approach).


bendk:
Bunch of updates for this one:

* Implemented a bunch of features, now the `bindgen-tests` pass
* Factored out some prep work and created separate PRs for them (#2868, #2869, #2870, #2871)


My next steps are:
* Implement the last bits of functionality and update `bindgen-tests` to test it
* Start testing out optimizations

uniffi-bindgen-kotlin-jni/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Run your executable and use the `bindings` subcommand to generate bindings.
+For example:
+
+`cargo run -p [bindgen-binary-crate-name] bindings src:[library-crate-name] [out-dir]`
+
+## Extra requriments compared to the normal macro code
+
+* All exported items must be `pub`.  This includes types, fields, functions, etc.
+* The top-level Rust crate must directly depend on all UniFFI crates
```
bendk, line 64: How do others feel about these requirements?  They don't seem too bad to me, but they are extra requirements compared to how `uniffi-bindgen-kotlin` currently works.

We could maybe work around them by putting some extra smarts into `uniffi_parse_rs` and the macro code, but it doesn't seem so easy to me.

~BTW, there's also currently another requirement: items need to be re-exported from the top-level `lib.rs` file.  However, I think we can work around that one with just `uniffi_parse_rs` and not messing with the macro system.  I think the requirement will just be that they need to be somehow reachable by outside crates (i.e. it's reachable by following a chain of `pub` modules from the root module).~

bendk:
Couple updates:
- Rebased this one top of #2874 and added functionality to make the tests work.  I think this is pretty close to feature complete
- Added multiple options for `Cleaner` implementations, like `uniffi-bindgen` currently has.

The next step is trying to optimize this.

bendk:
More updates:
* Hooked up the benchmarks to the new code
* Switch to using raw `jni-sys` instead of the `jni` crate, which lead to significant benchmark improvements. I tried to achieve the same with `jni`, but I felt like I was working against the library by not using their normal API. A side benefit is lighter dependencies for consumer crates

The bechmarks are looking pretty good and I haven't started really optimizing things yet:

```
                                                   main                 jni             jni-sys
-----------------------------------------------------------------------------------------------
kotlin-rust-call-only                            3.51us       262ns (13.4x)        54ns (65.0x)
kotlin-rust-enums                               22.48us       479ns (46.9x)       254ns (88.5x)
kotlin-rust-errors                               9.78us      2.75us  (3.6x)      2.20us  (4.4x)
kotlin-rust-hash-maps                           23.55us      2.60us  (9.0x)      2.35us (10.0x)
kotlin-rust-interfaces                          11.92us       613ns (19.5x)       354ns (33.7x)
kotlin-rust-large-strings                       28.67us     18.13us  (1.6x)     18.61us  (1.5x)
kotlin-rust-nested-data                         23.85us     15.05us  (1.6x)      8.39us  (2.8x)
kotlin-rust-primitives                           4.39us       377ns (11.6x)       131ns (33.5x)
kotlin-rust-records                             22.53us       544ns (41.4x)       336ns (67.0x)
kotlin-rust-strings                             22.26us      2.44us  (9.1x)       750ns (29.7x)
kotlin-rust-trait-interfaces                    12.51us       652ns (19.2x)       373ns (33.5x)
kotlin-rust-vecs                                23.06us       952ns (24.2x)       706ns (32.7x)
rust-kotlin-call-only                            7.06us       578ns (12.2x)       156ns (45.2x)
rust-kotlin-enums                               24.04us       815ns (29.5x)       374ns (64.3x)
rust-kotlin-errors                              11.84us      4.35us  (2.7x)      3.57us  (3.3x)
rust-kotlin-hash-maps                           24.73us      3.12us  (7.9x)      2.62us  (9.4x)
rust-kotlin-interfaces                          13.76us      1.38us  (9.9x)       490ns (28.1x)
rust-kotlin-large-strings                       29.79us     18.90us  (1.6x)     18.37us  (1.6x)
rust-kotlin-nested-data                         26.68us     12.38us  (2.2x)      7.59us  (3.5x)
rust-kotlin-primitives                           4.91us       677ns  (7.2x)       249ns (19.7x)
rust-kotlin-records                             23.94us       859ns (27.9x)       419ns (57.1x)
rust-kotlin-strings                             23.72us      2.78us  (8.5x)      1.10us (21.5x)
rust-kotlin-trait-interfaces                    13.79us      1.46us  (9.4x)       604ns (22.8x)
rust-kotlin-vecs                                24.22us      1.28us (18.9x)       808ns (30.0x)
```

bendk:
I'm going to close this one in favor of #2911.  That one is faster, simpler and the commits are cleaner.

> My plan is to implement everything using buffers, then come back and try to find cases where we can improve performance using primitive arguments.

It turns out that we can use this technique for all types except dynamically sized ones like vecs/hashmaps/sets.  This is faster and also simplifies the code since there's less different cases.

PR #2869, opened 2026-04-13
Adding uniffi_meta::Type::Box

bendk:
Remember store metadata about which types were boxed.  This is needed if we want to generate scaffolding using the pipeline.  Bindings can mostly ignore it.

### DIFF
--- a/fixtures/metadata/src/tests.rs
+++ b/fixtures/metadata/src/tests.rs
@@ -15,7 +15,7 @@ mod person {
         name: String,
         #[uniffi(default)]
         preferred_name: String,
-        age: u16,
+        age: Box<u16>,
     }
 }
 
@@ -180,6 +180,9 @@ mod test_type_ids {
         check_type_id::<Option<u8>>(Type::Optional {
             inner_type: Box::new(Type::UInt8),
         });
+        check_type_id::<Box<u8>>(Type::Box {
+            inner_type: Box::new(Type::UInt8),
+        });
         check_type_id::<Vec<u8>>(Type::Bytes);
         check_type_id::<Vec<u16>>(Type::Sequence {
             inner_type: Box::new(Type::UInt16),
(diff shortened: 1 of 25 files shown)


fixtures/metadata/src/tests.rs
```diff
@@ -180,6 +180,9 @@ mod test_type_ids {
         check_type_id::<Option<u8>>(Type::Optional {
             inner_type: Box::new(Type::UInt8),
         });
+        check_type_id::<Box<u8>>(Type::Box {
```
mhammond, line 183: Is it worth adding a boxed type to one of the structs above and explicitly checking the metadata?

mhammond (commented):
It seems like this might want more tests - eg, the bindings are actually changes, particularly Python, which implies something should make sure that's exercised.

And tbqh, I don't quite understand why this is necessary - I thought `Box<>` already worked? eg, `fixtures/recursive-enums` uses them extensively

bendk:
> It seems like this might want more tests - eg, the bindings are actually changes, particularly Python, which implies something should make sure that's exercised.
> 
> And tbqh, I don't quite understand why this is necessary - I thought `Box<>` already worked? eg, `fixtures/recursive-enums` uses them extensively

Yes, this was working and has tests (there's also `recursive-types` in `bindgen-tests`).   The reason for the change to eventually enable generating scaffolding using the pipeline (or `ComponentInterface`).  In that case, the distinction between generating code for a `Box<T>` vs a `T` becomes important.

fixtures/metadata/src/tests.rs
mhammond: makes sense to me, thanks! But I think there's still a test we could add here (ie, in this test, not exactly here) with a `Box<>` in a struct, but I'm fine if I'm just missing something :) :shipit:

mhammond (approved):

bendk:
> makes sense to me, thanks! But I think there's still a test we could add here (ie, in this test, not exactly here) with a Box<> in a struct, but I'm fine if I'm just missing something :) :shipit:

I made `person.age` use a box, was that what you were thinking?  If not, I can add the test in another pass.

mhammond:
> I made person.age use a box, was that what you were thinking? If not, I can add the test in another pass.

Yeah, that's exactly what I meant, thanks.

PR #2870, opened 2026-04-13
bindgen-tests updates

bendk:
* Made export items `pub`.  This is needed for the `uniffi-bindgen-kotlin-jni` code I'm working on. It also seems more correct, since we're calling these functions from across the FFI.
* Order the `futures` tests before `trait_interfaces`.  This is more logical, since `trait_interfaces` depends on the async code.
* Added a `From<uniffi::UnexpectedUniFFICallbackError>` impl to the callback tests.
* Re-export all items in the top-level crate.  This is currently needed to for `uniffi-bindgen-kotlin-jni`.

### DIFF
--- a/bindgen-tests/external-types/Cargo.toml
+++ b/bindgen-tests/external-types/Cargo.toml
@@ -8,4 +8,4 @@ crate-type = ["lib", "cdylib", "staticlib"]
 
 [dependencies]
 thiserror = "2"
-uniffi = { workspace = true }
+uniffi = { path = "../../uniffi", default-features = false }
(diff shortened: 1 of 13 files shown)


mhammond (approved):
Requiring exported items to be pub is a significant breaking change, even if in practice it turns out to only be necessary for kotlin in the future. How should we communicate that? I'd be fine with documenting it in the kotlin work, but not sure how we wont forget?

bendk:
> Requiring exported items to be pub is a significant breaking change, even if in practice it turns out to only be necessary for kotlin in the future. How should we communicate that? I'd be fine with documenting it in the kotlin work, but not sure how we wont forget?

Agreed that it's an important change that we should discuss and document.  I made a note in the README in that branch and started a thread in that PR.

Also, requiring items to be in the re-exported in the top-level crate feels like a more questionable change to me.  I removed that change from this PR and I'm going to try to fix the `uniffi_parse_rs` code so that we don't need that.

PR #2873, opened 2026-04-16
fix: sort uniffi_traits to ensure reproducible builds

ethicnology:
`HashSet` iteration order is non-deterministic in Rust due to randomized hash seeds. When multiple traits are passed to `#[uniffi::export(Debug, Eq, Hash)]`, they are collected into a `Vec` in arbitrary order, causing generated scaffolding code and embedded metadata to differ between builds.

This breaks reproducible builds, the same source produces different binary outputs.


uniffi_macros/src/export/item.rs
```diff
@@ -13,6 +13,20 @@ use super::attributes::{
 use crate::util::extract_docstring;
 use uniffi_meta::UniffiTraitDiscriminants;
 
+/// Collect and sort traits deterministically.
+/// HashSet iteration order is random per process; sorting ensures reproducible builds.
+fn sorted_traits(args: ExportStructArgs) -> Vec<UniffiTraitDiscriminants> {
+    let mut traits: Vec<UniffiTraitDiscriminants> = args.traits.into_iter().collect();
+    traits.sort_by_key(|t| match t {
```
bendk: Could this be `t as isize`?  I think that's safe and won't need updating if we add new variants.  I don't have a strong opinion either way though, if you want to keep the current code that's fine with me.

bendk (changes requested):
Thanks for the PR.  It looks good to me, but I had a suggestion for getting the key.

uniffi_macros/src/export/item.rs
ethicnology: Thanks for your review @bendk 

https://github.com/mozilla/uniffi-rs/pull/2873/commits/57118e3f081e949be986cba90fda92be9767c163 switched to `*t as isize` and added `Clone, Copy` derives to `UniffiTraitDiscriminants` since `sort_by_key` dereference requires `Copy`

bendk (approved):
Thanks for the update, let's merge this.

PR #2875, opened 2026-04-19
Generate CallableKind::VTableMethod again

bendk:
I noticed that we were no longer using `CallableKind::VTableMethod` for callback interface methods.  The JS code relies on this, so I updated the code to fix that.


mhammond (approved):
Surprised we need the python change here - is that because it's tricky to remove the enum field? Regardless, the general changes make sense.

bendk:
> Surprised we need the python change here - is that because it's tricky to remove the enum field? Regardless, the general changes make sense.

Yeah, it's harder because the macros generate an enum match statement that lists all the fields.  I haven't figured out a way for them to gnore the field like we can for a struct.

Issue #2864, opened 2026-04-08
Implement zero copy byte buffers for `&[u8]` and `&mut [u8]`


jhugman:
Following discussion in #2847 this issue is the MVP of zero copy buffers, with foreign owned byte arrays over synchronous calls.

The frontends already detect:
- `[ByRef] bytes` in UDL
- `&[u8]` in proc-macros.

The foreign language bindings represent this as:
- `Data` for Swift
- `ByteBuffer` for Kotlin — this is a change in behaviour.
- `bytes` for Python

For arguments in async calls, this would revert to copying semantics.

For `&mut [u8]`, this is slightly more involved:

- Introduce a `[ByMutRef] bytes` annotation in UDL
- Detect mutability in the proc macro, and error if async.
- Introduce a `LiftMutRef` trait to correspond to the `LiftRef` trait.
- alter the foreign language bindings:
  - `inout Data` for Swift
  - `bytearray` for Python

Without mutability, this would be more minimal, but non-viable.

Issue #2263, opened 2024-10-10
Support references in foreign traits somehow


mhammond:
Discussed a little with @linabutler and @bendk about this papercut. It exists for functions too, but seems more of a PITA for traits.

Eg, you can't use [this as a foreign trait](https://github.com/mozilla/uniffi-rs/blob/0b05ee1ab599445388956de1135cc80cfdc929fb/fixtures/proc-macro/src/lib.rs#L55) 
```
55 | #[uniffi::export(with_foreign)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   | |
   | expected `&str`, found `String`
   | help: change the parameter type to match the trait: `&str`
```

which seems unfortunate, even though it's tricky. I can't see an issue for this, so here's one!

linabutler:
I promised I'd file an issue about this...and didn't, oops! 😬 Thanks for starting the discussion!

References in traits (and functions) feel tricky: they're definitely more ergonomic for Rust callers, but they hide a "surprise copy" for foreign callers. I think most folks would expect that a `&T` argument would avoid a copy, but the scaffolding actually creates an owned `T`, passes `&T` as an argument, and then destroys `T`.

(If your function or trait method wanted to conditionally take ownership of the `&T`, it would need to do another copy, leading to a surprising incentive for such a function to take an owned `T` instead of `&T`).

I wonder if we should think about:

1. Disallowing `&T` arguments in UniFFI functions and traits, _but_...
2. Alowing `Cow<'_, T>` arguments. (I also really liked @bendk's suggestion to allow `impl Into<Cow<'_, T>>`).

That way, the scaffolding could always pass a `Cow::Owned(T)`, and a Rust caller could pass a `Cow::Borrowed(T)`. `Cow` doesn't have a blanket `From<&'a T>` implementation, but it does implement `From` for owned and borrowed strings, paths, and slices, so a Rust caller would be able to write, say, `func("foo".into())` instead of `func(Cow::Borrowed("foo"))`.

I'm not sure that we can fix this "for real", and make cross-language references work properly 😔 Foreign objects—especially in the languages we're targeting—can't uphold Rust's "aliasing XOR mutability" guarantee, in either direction (C / C++ could, maybe, if we were _very careful_?)...and if we're passing a dictionary or a list, we need to serialize it, anyway.

WDYT?

mhammond:
[...]

PR #2878, opened 2026-04-22
Add support for `&[u8]` / `[ByRef] bytes` to travel from Foreign to Rust

jhugman:
This is part 1 of 2, for #2864.

This is just shipping `&[u8]` from the foreign side to Rust, in the argument position.

I've put tests in `bindgen-tests` ~~AND in `fixtures/`~~, mainly because there are ~~UDL parts and~~ proc macro parts to this.


uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -898,6 +898,37 @@ mod filters {
...
+    /// pointer to. Other args take the per-Type `type_name` path.
+    #[askama::filter_fn]
+    pub(super) fn type_name_for_arg(
+        arg: &Argument,
+        _: &dyn askama::Values,
+        ci: &ComponentInterface,
+    ) -> Result<String, askama::Error> {
+        if arg.is_borrowed_bytes() {
+            Ok("java.nio.ByteBuffer".to_string())
+        } else {
```
bendk, line 914: I've been working on the JNI-based bindings and I'm thinking of doing something very similar there -- special-casing some types to be passed as native arrays and also special-casing argument passing.

Would this would work for a callback argument where Rust is passing data to Kotlin?   If not, what about naming it something like `lower_type_name_for_arg`?  I feel like the more precise the better for the special cases.

fixtures/zero-copy-bytes/src/zero_copy_bytes.udl
```diff
@@ -0,0 +1,3 @@
+namespace zero_copy_bytes {
+    u32 sum_bytes_udl([ByRef] bytes buf);
+};
```
bendk: Do we need a separate fixture to for this?  I'd hoping to migrate all the tests to `bindgen-tests`.  UDL and macros share enough code that I don't feel like we need to test both cases.

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferTemplate.kt
```diff
@@ -65,3 +65,40 @@ internal open class ForeignBytes : Structure() {
...
+// the FFI call; the Rust side treats it as a borrow.
+internal object FfiConverterByRefBytes : FfiConverter<java.nio.ByteBuffer, ForeignBytes.ByValue> {
+    override fun lower(value: java.nio.ByteBuffer): ForeignBytes.ByValue {
+        require(value.isDirect) { "UniFFI zero-copy &[u8] requires a direct ByteBuffer. Use ByteBuffer.allocateDirect()." }
+        val remaining = value.remaining()
+        val fb = ForeignBytes.ByValue()
+        fb.len = remaining
+        // Zero-length direct buffers: skip getDirectBufferPointer (platform-variable behavior)
+        // and pass null. The Rust side treats (null, 0) as &[].
+        fb.data = if (remaining == 0) null else com.sun.jna.Native.getDirectBufferPointer(value)
```
bendk, line 89: I think this is the right way to go for this PR, but I'll just note that in my testing creating a struct and passing it over the FFI slows things down significantly.  It would be faster to pass this as 2 primitive `data`/ `length` arguments.  However, that would require a significant rewrite of the code.  Also, the performance penalty is fixed so the bigger the buffer the less it matters.

uniffi_bindgen/src/bindings/swift/templates/macros.swift
```diff
@@ -11,6 +11,19 @@
...
+        `$0` shorthand. #}
+    {%- if func|has_borrowed_bytes_arg %} uniffiCallStatus in{% endif %}
+    {#- Open nested `FfiConverterByRefBytes.lower` scopes for any `&[u8]`
+        args. The `ForeignBytes` value is only guaranteed valid inside the
+        scope, so the full FFI call runs inside the innermost closure. #}
+    {%- for arg in func.arguments() -%}
+    {%-     if arg|is_borrowed_bytes %}
+        FfiConverterByRefBytes.lower({{ arg.name()|var_name }}) { {{ arg.name()|var_name }}Fb in
+    {%-     endif %}
+    {%- endfor %}
```
bendk, line 25: Generating all these nested closures is pretty wild and I like it.  What about always naming the `RustCallStatus` arg to keep the code below more consistent?

uniffi_core/src/ffi/ffiserialize.rs
```diff
@@ -220,6 +220,21 @@ impl FfiSerialize for RustBuffer {
...
+        // Safety: the foreign bindings are responsible for sending us the correct data.
+        let (len, data) = unsafe { (buf[0].i32, buf[1].ptr as *const u8) };
+        unsafe { ForeignBytes::from_raw_parts(data, len) }
+    }
+
+    fn put(buf: &mut [FfiBufferElement], value: Self) {
+        buf[0].i32 = value.len;
+        buf[1].ptr = value.data as *const std::ffi::c_void;
+    }
+}
```
bendk, line 236: Sorry for making you implement this one, we should probably get rid of this trait.  The experiment taught us something, but I don't think we're going to use this exact FFI.

uniffi_macros/src/fnsig.rs
```diff
@@ -12,6 +12,15 @@ use proc_macro2::{Span, TokenStream};
 use quote::quote;
 use syn::{spanned::Spanned, FnArg, Ident, Pat, Receiver, ReturnType, Type};
 
+fn is_u8_slice(ty: &Type) -> bool {
+    if let Type::Slice(s) = ty {
+        if let Type::Path(p) = &*s.elem {
+            return p.path.get_ident().is_some_and(|i| i == "u8");
```
bendk: - I think `path.is_ident` could simplify this code.
- This is going to fail under pathological conditions, like users creating a type alias named `u8` for their own types.  Maybe the most likely case is a user writing out something `&[::std::core::u8]` which I don't think this would correctly identify.  None of this feels like a problem in practice, but could you a short doc saying something about this?

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
...
   You can get similar behavior by adding the primary constructor to the uniffi.toml excludes list in uniffi.toml (e.g. `excludes = ["MyObject.new"])
+- **Kotlin:** Functions declared with `[ByRef] bytes` in UDL or `&[u8]` in proc-macros now accept `java.nio.ByteBuffer` (direct) instead of `ByteArray`. To migrate, wrap a `ByteArray` as `ByteBuffer.allocateDirect(arr.size).put(arr).flip()`. This enables zero-copy transfer of byte buffers from Kotlin to Rust ([#2864](https://github.com/mozilla/uniffi-rs/issues/2864)).
+- **Ruby:** Functions declared with `[ByRef] bytes` or `&[u8]` are not supported by the Ruby binding in this release. Using them will cause codegen to fail or produce code that breaks at runtime ([#2864](https://github.com/mozilla/uniffi-rs/issues/2864)).
+- **Async `&[u8]` / `[ByRef] bytes` is not supported this release.** Swift fails at compile time; Kotlin and Python may build but have unverified lifetime guarantees across suspension. Use owned `Vec<u8>` / `bytes` for async functions until a dedicated follow-up lands ([#2864](https://github.com/mozilla/uniffi-rs/issues/2864)).
```
bendk: This is a breaking change for external bindings right?  I wonder if there's a way to prevent/mitigate that with a feature flag that users would have to set to opt-in to the new behavior.  If the feature flag caused `is_borrowed_bytes` to always return `false` would work?

bendk (changes requested):
This looks really great.  I left a few comments, but the main thing I care about is making it so this doesn't have to break things for external bindings.  What do you think about the feature flag idea?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
jhugman: Good idea. I've renamed as suggested.

fixtures/zero-copy-bytes/src/zero_copy_bytes.udl
jhugman: Ok; I'm not sure what `[ByRef] bytes` was supposed to do previously, but it works already, and no new code is added in this PR that needs testing.

I'll remove this test.

CHANGELOG.md
jhugman: I think in the face of a feature flag, I'm going to row back on the breakiness of this change: Since moving `&[u8]` across the FFI has never been possible, and `[ByRef] bytes` was mapped to `&Vec<u8>`, but is documented as behaving like `&[u8]`.

So, I think in a narrow sense: it's a breaking change, but in the way a bug fix is a breaking change. I don't think it needs a feature flag.

It definitely needs support from third party bindgen authors however, and that may in and of itself be a cause to increment the version number.

I would hold off on the version bump until the `&mut [u8]` variant lands.

WDYT?

jhugman:
Thanks for your review, @bdk! I think I've addressed your feedback. Please may I have another?

CHANGELOG.md
bendk: I still think this is a breaking change and should be called out as such in the CHANGELOG.  I think I get what you're saying about the old behavior being buggy, but there's another way to look at it:  supporting `&[u8]` was not about a performance guarantee but more about supporting a particular Rust signature.  I've pictured it as a minor performance win, since the Rust -> Rust call could be efficient, even if Foreign -> Rust call still needed a copy.

However, I don't think you need a feature flag for this.  I re-reviewed the Kotlin/Swift/Python changes and they didn't look that difficult to me.  I've definitely introduced some bigger breaking changes.

bendk (changes requested):
Still looking good, the one change I'd like to see is that this is called out as a "breaking change for external bindings" in the changelog, with a link to this PR.  That way external bindings authors can see Kotlin/Swift/Python changes and get a feel for what they should do.

uniffi_bindgen/src/bindings/kotlin/templates/RustBufferTemplate.kt
```diff
@@ -65,3 +65,40 @@ internal open class ForeignBytes : Structure() {
...
+    }
+
+    override fun lift(value: ForeignBytes.ByValue): java.nio.ByteBuffer =
+        error("ByRef bytes cannot be lifted: zero-copy &[u8] only flows foreign->Rust")
+
+    override fun read(buf: java.nio.ByteBuffer): java.nio.ByteBuffer =
+        error("ByRef bytes cannot be read from a buffer: zero-copy &[u8] only flows foreign->Rust")
+
+    override fun write(value: java.nio.ByteBuffer, buf: java.nio.ByteBuffer): Unit =
+        error("ByRef bytes cannot be written to a buffer: zero-copy &[u8] only flows foreign->Rust")
```
mhammond: If I'm not mistaken, this is the "only in argument position" part of the leading comment - so the message for this one isn't about the direction. Not that it's a material distinction as we don't expect to see this in any case, but maybe making the message more accurate is worthwhile?

If I'm not confused and this makes sense, it applies to the other bindings below too.

CHANGELOG.md
mhammond: I sometimes think supporting byref in that way is a bit of a mis-feature as it might imply things which aren't true to some consumers, and it's not universally supported - eg, #2263 - but I agree this should be called out as a breaking change.

mhammond (commented):
This seems like a great feature! It needs doc though.

mhammond:
> I sometimes think supporting byref in that way is a bit of a mis-feature

Maybe I actually think proc-macros supporting it by default was the mistake? You should opt-in to this sugar such as a uniffi attribute or something, but a plain `export` accepting a reference can be misleading. But that ship has probably sailed.

CHANGELOG.md
bendk: Side note, since this discussion is interesting:  If we were staring from scratch, I'd consider supporting both `AsRef<T>` and `&T` and make the distinction that we only implement `&T` when we can actually do zero-copy (or maybe a less-than-normal-copy, since I think the current code copies things like 3 times).

bendk (approved):
I checked out the CHANGELOG changes and they look good to me.  I think the only remaining item is Mark's documentation suggestion.

jhugman:
Thanks @bendk ; @mhammond: is 241d814d5 the right direction?

mhammond:
Those docs are definitely in claude's voice 😉 but they are fine, thanks

PR #2879, opened 2026-04-25
Implement a "global config".

### LINKED ISSUE #2653: Bundling UniFFI in a local Swift Package Manager package - need to import the module declared in the modulemap?

### LINKED ISSUE #2866: Unified Project Configuration
I spent quite some time thinking about @bendk's comment in #2798 that we should support a better "global config", and thinking about my comments there which have a strawperson's idea for what that might look like. So I asked claude to help come up with a plan which covers the usecase in that PR, and also how we use `--config` in the gecko-js bindings. Below is what we came up with, but before that, it suggested some discussion points. My takes:

1. I think we should just break backwards compatibility with old "config override" files - the burden of shifting to this isn't that bad, and they are almost certainly going to want to do that eventually to take advantage of this.
2. no real opinion, but "project" seems quite sane.
3. is talking about legacy code we deprecated quite a few versions ago, but remains in-place in-case other external generators are still using it. If it gets too ugly I think we should remove it, but keep it if the cost is low.
4. We do want it.

WDYT?

# Discussion points:
The key insight in the design: Rather than changing BindgenPaths semantics (which would risk breaking external generators), ProjectConfigLayer does all three-level merging internally and presents a single pre-merged answer to the existing first-match-wins system. This makes it a purely additive change.

Open questions the team may want to discuss:

1. Old-format --config detection: Auto-detecting old vs. new format by checking for [defaults]/[crates.*]/[crate-roots] keys works but is a heuristic. An alternative is a semver-bump that changes the flag semantics cleanly, or accepting that old-format users will get a parse error that directs them to migrate.
  2. Naming: The plan uses "project config" as the concept. The PR discussion uses "global config". Either works — the team may prefer different terminology.
  3. The legacy BindgenCrateConfigSupplier path: The plan leaves this alone, but it has different merge semantics (it DOES merge). It's used by generate_external_bindings() which external generators call. Whether to unify it with the new system is a separate decision.
  4. Swift CLI: uniffi-bindgen-swift has no --config flag today. Do we want to add one there too? (The PR #2798 discussion touched on this.)

# Plan: Unified Project Configuration for UniFFI

## Context

UniFFI's configuration system has grown organically and has several problems:

1. **The `--config` override is broken for multi-crate projects.** It applies the same config to ALL crates (ignoring crate name), so you can set `ffi_module_name` (must be same for all) but not `ffi_module_filename` (must differ per crate).

2. **No way to express global defaults.** Settings like `ffi_module_name` that must be identical across crates can only be set by either duplicating them in every crate's `uniffi.toml` or using the broken override.

3. **No way to locate crates without cargo-metadata.** Non-cargo build systems (Firefox mach, Bazel, Buck2) can't use `cargo metadata` to discover crate roots.

4. **Two parallel config systems.** The legacy `BindgenCrateConfigSupplier` does deep merging; the newer `BindgenPaths` does first-match-wins. Confusing and inconsistent.

PR [#2798](https://github.com/mozilla/uniffi-rs/pull/2798) attempts to fix (3) with a `--crate-metadata` option, but the discussion concluded we should solve these problems holistically.

**Goal:** A single "project config" file that replaces `--config`, supports global defaults, per-crate overrides, and crate location — forming a clear, layered configuration system.

### File Format

The `--config` flag accepts a project config file with three optional sections:

```toml
# Locate crates without cargo-metadata (optional)
# Paths are relative to this file's directory
[crate-roots]
my_crate = "crates/my_crate"
other_crate = "vendor/other_crate"

# Defaults applied to ALL crates (optional)
# Same structure as a per-crate uniffi.toml
[defaults.bindings.swift]
ffi_module_name = "MyLibraryFFI"
generate_immutable_records = true

[defaults.bindings.kotlin]
android = true

# Per-crate overrides (optional)
# Same structure as uniffi.toml, nested under crate name
[crates.my_crate.bindings.swift]
ffi_module_filename = "my_crate_ffi"

[crates.other_crate.bindings.swift]
ffi_module_filename = "other_crate_ffi"

[crates.other_crate.bindings.kotlin]
package_name = "com.example.other"
```

### Config Resolution Order (per crate)

For each crate, the effective config is produced by deep-merging these layers (later wins):

1. **`[defaults]`** from project config — lowest priority
2. **Per-crate `uniffi.toml`** — found via `[crate-roots]` or cargo-metadata
3. **`[crates.<name>]`** from project config — highest priority

Deep merge uses the existing `merge_toml()` function: tables merge recursively, scalars/arrays replace.

### How This Solves Each Problem

| Problem | Solution |
|---------|----------|
| Override applies same config to all crates | `[crates.<name>]` sections allow per-crate overrides |
| No global defaults | `[defaults]` applies to all crates |
| Can't locate crates without cargo-metadata | `[crate-roots]` maps crate names to paths |
| Config override vs. uniffi.toml are all-or-nothing | Three-level merge gives fine-grained control |

### Example: The Swift `ffi_module_name` / `ffi_module_filename` Problem

```toml
# ffi_module_name MUST be the same for all crates
[defaults.bindings.swift]
ffi_module_name = "MyAppFFI"

# ffi_module_filename MUST differ per crate
[crates.auth.bindings.swift]
ffi_module_filename = "auth_ffi"

[crates.storage.bindings.swift]
ffi_module_filename = "storage_ffi"
```

### Example: Non-cargo Environment (Firefox-style)

```toml
[crate-roots]
suggest = "../application-services/components/suggest"
tabs = "../application-services/components/tabs"
remote_settings = "../application-services/components/remote_settings"

# No [defaults] or [crates.*] needed if per-crate uniffi.toml files
# at those roots are sufficient
```

### Architecture: `ProjectConfigLayer`

Rather than changing `BindgenPaths` semantics (which would break external generators like gecko-js), we create a single `ProjectConfigLayer` that implements `BindgenPathsLayer` and handles all merging internally.

When `--config` points to a project config file, `ProjectConfigLayer` becomes the **only** layer in `BindgenPaths`. It wraps the cargo-metadata layer internally as a fallback for crate discovery. From `BindgenPaths`' perspective, it's just one layer returning one pre-merged table per crate — fully compatible with first-match-wins.

```
ProjectConfigLayer
├── defaults (from [defaults])
├── crate discovery
│   ├── [crate-roots] entries (tried first)
│   └── cargo-metadata (fallback for crates not in [crate-roots])
└── overrides (from [crates.<name>])
```

### Key Implementation Steps

1. **New module `uniffi_bindgen/src/project_config.rs`**
   - `ProjectConfigFile` struct (serde-deserializable from the TOML)
   - `ProjectConfigLayer` implementing `BindgenPathsLayer`
   - `get_config()`: merge defaults → crate uniffi.toml → per-crate overrides
   - `get_udl_path()`: resolve from crate-roots, fall back to inner layer

2. **Make `merge_toml` public** in `uniffi_bindgen/src/lib.rs` (line 564, currently private)

3. **Add `add_project_config_layer()` to `BindgenPaths`** in `bindgen_paths.rs`

4. **Update `generate_with_bindgen_paths()`** in `bindings/mod.rs`:
   - When `config_override` is set, detect if it's a project config (has `[defaults]`, `[crates.*]`, or `[crate-roots]` keys) vs. old-style override
   - For project config: create `ProjectConfigLayer` wrapping cargo-metadata as inner layer
   - For old-style override: use existing `ConfigOverrideLayer` (with deprecation warning)

5. **CLI remains `--config`** in `uniffi/src/cli/uniffi_bindgen.rs` — same flag, new capability

### Files to Modify

| File | Change |
|------|--------|
| `uniffi_bindgen/src/project_config.rs` | **New**: `ProjectConfigLayer` impl |
| `uniffi_bindgen/src/lib.rs` | Make `merge_toml` public, add `mod project_config` |
| `uniffi_bindgen/src/bindgen_paths.rs` | Add `add_project_config_layer()` method |
| `uniffi_bindgen/src/bindings/mod.rs` | Update `generate_with_bindgen_paths()` to handle project config |
| `uniffi/src/cli/uniffi_bindgen.rs` | Update `--config` help text |

### Backward Compatibility

- **Per-crate `uniffi.toml` files**: Unchanged. Continue working exactly as before.
- **`--config` with old-style files**: Auto-detected and handled with deprecation warning. Old-style files lack `[defaults]`, `[crates.*]`, and `[crate-roots]` keys.
- **`BindgenPathsLayer` trait**: Unchanged. External generators (gecko-js, react-native) keep working.
- **`BindgenPaths` semantics**: Unchanged (first-match-wins). The `ProjectConfigLayer` does merging internally.
- **Legacy `BindgenCrateConfigSupplier`**: Not touched in this change (separate cleanup).

### External Binding Generators

The system is **opaque to language-specific config structures**. It does TOML table merging without parsing `bindings.swift.*` or `bindings.kotlin.*` — language-specific deserialization happens downstream, unchanged.

External generators can:
1. **Ignore this entirely** — continue using custom `BindgenPathsLayer` implementations (gecko-js pattern)
2. **Use `ProjectConfigLayer` directly** — call `paths.add_project_config_layer()` when constructing `BindgenPaths` programmatically
3. **Benefit transparently** — if users pass `--config` with a project config file, the generator gets pre-merged config from `get_config()` without any code changes

## User-Facing Description

> ### Project Configuration
>
> For projects using multiple UniFFI crates, a project config file gives you centralized control over configuration.
>
> ```
> uniffi-bindgen generate --config project.toml -l swift -o out/ libmylibrary.dylib
> ```
>
> The file has three optional sections:
>
> - **`[crate-roots]`** — Maps crate names to their root directories. Use this when cargo-metadata is unavailable or to override crate locations.
>
> - **`[defaults]`** — Configuration applied to every crate, using the same format as `uniffi.toml`. These are the lowest priority — a crate's own `uniffi.toml` overrides them.
>
> - **`[crates.<name>]`** — Per-crate overrides with the highest priority. These override both `[defaults]` and the crate's own `uniffi.toml`.
>
> **Priority** (lowest to highest):
> 1. `[defaults]` from project config
> 2. The crate's `uniffi.toml`
> 3. `[crates.<name>]` from project config
>
> All sections use the same format as `uniffi.toml` and are merged using deep table merging (tables merge recursively, scalar values replace).

## Verification

1. **Unit tests** for `ProjectConfigLayer`:
   - Three-level merge produces correct results
   - `crate-roots` resolution with relative paths
   - Fallback to cargo-metadata for unlisted crates
   - Empty/partial project config (defaults-only, overrides-only, roots-only)

2. **Integration test** using existing `ext-types` fixture:
   - Create a project config with defaults + per-crate overrides
   - Generate bindings and verify the config was applied correctly

3. **Old-format detection test**: Verify old-style `--config` files trigger deprecation warning and continue working

4. **Firefox gecko-js smoke test**: Verify gecko-js builds are unaffected (its custom `BindgenPathsLayer` is unchanged)

mhammond:
As described in #2866. Fixes #2866, closes #2798, closes #2653

### DIFF
--- a/docs/manual/src/bindings.md
+++ b/docs/manual/src/bindings.md
@@ -1,19 +1,85 @@
 # Generating bindings
 
-Bindings is the term used for the code generates for foreign languages which integrate
-with Rust crates - that is, the generated Python, Swift or Kotlin code which drives the
-examples.
+"Bindings" is the term for the non-Rust side of the environment - the generated Python, Swift or Kotlin code which drives the examples.
 
 UniFFI comes with a `uniffi_bindgen` which generates these bindings. For introductory
-information, see [Foreign Language Bindings in the tutorial](./tutorial/foreign_language_bindings.md)
+information, see [Foreign Language Bindings in the tutorial](./tutorial/foreign_language_bindings.md).
+There are also external bindings which don't live in this repo.
 
-# Customizing the binding generation.
+# Customizing the binding generation
 
-Each of the bindings reads a file `uniffi.toml` in the root of a crate which supports
-various options which influence how the bindings are generated. Default options will be used
-if this file is missing.
+Each binding generator has their own unique configuration options to customize their output.
+There's a toml-based capability to specify these various options.
 
-`--config` option can be used to specify additional uniffi config file. This config is merged with
-the `uniffi.toml` config present in each crate, with its values taking precedence.
+Binding generation can be configured with a `uniffi.toml` in the root of each crate,
+and by specifying the path to a toml file which specifies global configuration.
 
-Each binding supports different options, so please see the documentation for each binding language.
+For the builtin bindings, see [Kotlin](./kotlin/configuration.md), [Python](./python/configuration.md) and [Swift](./swift/configuration.md).
+
+# Global configuration
+
+For projects with multiple UniFFI crates, a global config file lets you set shared defaults and
+per-crate overrides in one place. For example:
+
+```
+uniffi-bindgen generate --config global.toml -l swift -o out/ libmylibrary.dylib
+```
+
+The file has three optional sections:
+
+- **`[crate-roots]`** — Maps crate names to their root directories. Use this when
+  `cargo metadata` is unavailable or to override crate locations. Paths are relative
+  to the config file's directory.
+
+- **`[defaults]`** — Configuration applied to every crate, using the same format as
+  `uniffi.toml`. These are the lowest priority and are overridden by a crate's own
+  `uniffi.toml`.
+
+- **`[crates.<name>]`** — Per-crate overrides with the highest priority. These override
+  both `[defaults]` and the crate's own `uniffi.toml`.
+
+**Priority** (lowest to highest): `[defaults]` → crate's `uniffi.toml` → `[crates.<name>]`
+
+All sections use the same format as `uniffi.toml`. Tables/maps are merged recursively; scalar
+values are replaced.
+
+### Example: shared FFI module name with per-crate filenames
+
+```toml
+# ffi_module_name must be the same for all crates
+[defaults.bindings.swift]
+ffi_module_name = "MyAppFFI"
+
+# ffi_module_filename must differ per crate
+[crates.auth.bindings.swift]
+ffi_module_filename = "auth_ffi"
+
+[crates.storage.bindings.swift]
+ffi_module_filename = "storage_ffi"
+```
+
+# Crate locations
+
+UniFFI needs to locate UDL files and `uniffi.toml` files for each referenced crate.
+It uses the same mechanism in both places - either via the `cargo-metadata` feature,
+and/or via global configuration.
+See the discussion about `[crate-roots]` above for more info.
+If there's no `[crate-roots]` entry for a crate, we'll try and fall back to using `cargo-metadata`
+if it's enabled, otherwise we'll use the default config for the crate (and give an error if we need a UDL).
+
+### Example: locating crates without cargo-metadata
+
+```toml
+[crate-roots]
+# These entries also tell UniFFI where to find `suggest.udl` or `uniffi.toml` files,
+# avoiding cargo-metadata. Path is relative to the directory of this config file.
+suggest = "./components/suggest"
+tabs = "./components/tabs"
+
+# This same file can also have `[defaults]` or overrides, like:
+[defaults.bindings.swift]
+some_global_option = true # default for every crate
+
+[crates.suggest.bindings.swift]
+some_option = true # specific override for the `suggest` crate.
+```
--- a/uniffi_bindgen/src/global_config.rs
+++ b/uniffi_bindgen/src/global_config.rs
@@ -1,0 +1,404 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use std::{collections::HashMap, fs};
+
+use anyhow::{Context, Result};
+use camino::{Utf8Path, Utf8PathBuf};
+use serde::Deserialize;
+
+use crate::{merge_toml, BindgenPaths, BindgenPathsLayer};
+
+/// A [BindgenPathsLayer] backed by an explicit `[crate-roots]` mapping.
+pub struct CrateRootsLayer {
+    roots: HashMap<String, Utf8PathBuf>,
+}
+
+impl BindgenPathsLayer for CrateRootsLayer {
+    fn get_crate_root(&self, crate_name: &str) -> Option<Utf8PathBuf> {
+        self.roots.get(crate_name).cloned()
+    }
+}
+
+/// Deserialized representation of a global config TOML file.
+#[derive(Deserialize, Default)]
+struct GlobalConfigFile {
+    #[serde(rename = "crate-roots", default)]
+    crate_roots: HashMap<String, String>,
+    #[serde(default)]
+    defaults: toml::value::Table,
+    #[serde(default)]
+    crates: HashMap<String, toml::value::Table>,
+}
+
+/// Global configuration for UniFFI.
+///
+/// Holds defaults and per-crate overrides from a global config file.
+/// Config resolution requires a `&BindgenPaths` to locate per-crate `uniffi.toml` files.
+#[derive(Default)]
+pub struct GlobalConfig {
+    defaults: toml::value::Table,
+    crate_overrides: HashMap<String, toml::value::Table>,
+}
+
+impl GlobalConfig {
+    /// Parse a global config file.
+    ///
+    /// Returns the `GlobalConfig` and an optional `CrateRootsLayer`. If a
+    /// `CrateRootsLayer` is returned, add it to your `BindgenPaths` before
+    /// calling `get_config`.
+    pub fn from_file(path: &Utf8Path) -> Result<(Self, Option<CrateRootsLayer>)> {
+        let contents =
+            fs::read_to_string(path).with_context(|| format!("read file: {:?}", path))?;
+
+        // Check for old-style flat config files before full parse
+        let raw: toml::value::Table =
+            toml::de::from_str(&contents).with_context(|| format!("parse toml: {:?}", path))?;
+
+        let has_global_config_keys = raw.contains_key("crate-roots")
+            || raw.contains_key("defaults")
+            || raw.contains_key("crates");
+
+        if !has_global_config_keys && !raw.is_empty() {
+            eprintln!(
+                "warning: {path} looks like an old-style --config override file. \
+                The --config flag now expects a global config file with [defaults], \
+                [crates.<name>], and/or [crate-roots] sections. \
+                Old-style flat config files are no longer supported."
+            );
+            return Ok((Self::default(), None));
+        }
+
+        let file: GlobalConfigFile =
+            toml::de::from_str(&contents).with_context(|| format!("parse toml: {:?}", path))?;
+
+        let crate_roots_layer = if file.crate_roots.is_empty() {
+            None
+        } else {
+            let base_dir = path.parent().unwrap_or(Utf8Path::new("."));
+            let roots = file
+                .crate_roots
+                .into_iter()
+                .map(|(name, rel_path)| (name, base_dir.join(rel_path)))
+                .collect();
+            Some(CrateRootsLayer { roots })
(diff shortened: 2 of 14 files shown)


docs/manual/src/bindings.md
```diff
@@ -1,19 +1,86 @@
...
+There's a toml-based capability to specify these various options.
 
-`--config` option can be used to specify additional uniffi config file. This config is merged with
-the `uniffi.toml` config present in each crate, with its values taking precedence.
+In the simple default configuration, the bindings read a file `uniffi.toml` in the root of each crate.
+There's also a global configuration option, useful when multiple crates need to share some settings.
+This configuration has the same options as available in `uniffi.toml`.
 
-Each binding supports different options, so please see the documentation for each binding language.
+Of the builtin bindings, see [Kotlin](./kotlin/configuration.md), [Python](./python/configuration.md) and [Swift](./swift/configuration.md).
```
bendk: This reads weirdly to me, what about "For the builtin bindings configuration options, ...."?

docs/manual/src/bindings.md
```diff
@@ -1,19 +1,86 @@
...
 
-Each of the bindings reads a file `uniffi.toml` in the root of a crate which supports
-various options which influence how the bindings are generated. Default options will be used
-if this file is missing.
+Each binding generator has their own unique configuration options to customize their output.
+There's a toml-based capability to specify these various options.
 
-`--config` option can be used to specify additional uniffi config file. This config is merged with
-the `uniffi.toml` config present in each crate, with its values taking precedence.
+In the simple default configuration, the bindings read a file `uniffi.toml` in the root of each crate.
```
bendk: Is it simpler to have a `uniffi.toml` for each crate?  Maybe this should be in a more neutral tone like "bindings can be configured by a `uniffi.toml` file in the root of each crate and/or a global `uniffi.toml`file that covers all crates".  I don't have a strong opinion about this though, what's your opinion?

docs/manual/src/bindings.md
```diff
@@ -1,19 +1,86 @@
...
+- **`[defaults]`** — Configuration applied to every crate, using the same format as
+  `uniffi.toml`. These are the lowest priority and are overridden by a crate's own
+  `uniffi.toml`.
+
+- **`[crates.<name>]`** — Per-crate overrides with the highest priority. These override
+  both `[defaults]` and the crate's own `uniffi.toml`.
+
+**Priority** (lowest to highest): `[defaults]` → crate's `uniffi.toml` → `[crates.<name>]`
+
+All sections use the same format as `uniffi.toml`. Tables are merged recursively; scalar
```
bendk: Maybe "Table" should be replaced with "Tables/maps".  It took me a bit to realize that that applied to things like the external package map.

This also makes think of the `custom_types` config, which doesn't seem to make sense to in the `[defaults]` section.  Does it makes sense to have a list of config options that can't have defaults?  I'm not sure, maybe it's better to keep things consistent even if it allows weird configurations.

docs/manual/src/bindings.md
```diff
@@ -1,19 +1,86 @@
...
+if it's enabled, otherwise we'll use the default config for the crate (and give an error if we need a UDL).
+
+### Example: locating crates without cargo-metadata
+
+```toml
+[crate-roots]
+# These entries also tell UniFFI where to find `suggest.udl` or `uniffi.toml` files,
+# avoiding cargo-metadata.
+suggest = "./components/suggest"
+tabs = "./components/tabs"
```
bendk, line 77: I love this feature and I think it will help `uniffi_parse_rs` which currently needs to lean heavily on `cargo-metadata` to find the source roots.  For example, I wanted to test out the new code with the boltffi benchmarks and had to hack things because they chose to exclude the benchmark crates from the cargo workspace.  It would be great to have a global `uniffi.toml` file that could manually configure the crate roots.

Where are the paths relative to?

uniffi_bindgen/src/global_config.rs
```diff
@@ -0,0 +1,391 @@
...
+            || raw.contains_key("defaults")
+            || raw.contains_key("crates");
+
+        if !has_global_config_keys && !raw.is_empty() {
+            eprintln!(
+                "warning: {path} looks like an old-style --config override file. \
+                The --config flag now expects a global config file with [defaults], \
+                [crates.<name>], and/or [crate-roots] sections. \
+                Old-style flat config files are no longer supported."
+            );
```
bendk, line 69: Very nice.  I think this will save some users a bunch of time.

uniffi_bindgen/src/lib.rs
```diff
@@ -561,7 +563,7 @@ fn overridden_config_value(
     Ok(toml::Value::from(config))
 }
 
-fn merge_toml(a: &mut toml::value::Table, b: toml::value::Table) {
+pub fn merge_toml(a: &mut toml::value::Table, b: toml::value::Table) {
```
bendk: Should we add some extra error checking to this?  For example, I think we should return an error if one side has a table and the other side has a scalar for some key.

bendk (changes requested):
This looks great overall, there's just some details to iron out.  I'm picking "Request changes" because I have several questions, but I don't really have strong opinions one way or another.  Maybe the answer to all of them is "the current code is right" and we could just merge as-is.

docs/manual/src/bindings.md
mhammond: I'd have thought `custom_types` might make sense in `[defaults]` in some cases - eg, some of the examples around using `Url` in Kotlin etc?

Regardless though, I do think we should avoid trying to validate too much here - there are many other config options which don't make sense and which we don't really try to guard against.

docs/manual/src/bindings.md
mhammond: Relative to the directory with the global config file - I mentioned that above here, but also added a note here too.

uniffi_bindgen/src/lib.rs
mhammond: yeah, good idea!

docs/manual/src/bindings.md
mhammond: yeah, I agree this isn't great and still not quite happy with the changes I made, love to get further ideas.

bendk (approved):
The changes look great to me and I agree with all the decisions.

PR #2883, opened 2026-04-30
Expose standard Rust traits for ruby

saks:
**This PR implements standard Rust traits for the Ruby bindings.**

It adds idiomatic Ruby support for common Rust traits by mapping them to their natural Ruby equivalents:

- `std::fmt::Debug` → `#inspect` 
- `std::fmt::Display` → `#to_s`
- `PartialEq` / `Eq` → `#==` and `#eql?`
- `PartialOrd` / `Ord` → `#<=>`, `#<`, `#<=`, `#>`, `#>=`
- `Hash` → `#hash`

### Key Benefits
- More idiomatic and Ruby-like API
- Better interoperability with Ruby's standard library and ecosystem
- Enables use of common Ruby idioms (e.g. `pp`, `Set`, `Hash` keys, sorting, etc.)

**No breaking changes** — all existing functionality continues to work as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 - Kotlin objects now have an `uniffiIsDestroyed` property that returns `true` if the Rust reference no longer exists ([#2825](https://github.com/mozilla/uniffi-rs/pull/2825))
 - Updated `askama` version to `0.15.6`
 - Custom Types can have docstrings in some languages ([#2853](https://github.com/mozilla/uniffi-rs/pull/2853))
+- Ruby: Expose standard Rust traits for generated ruby code (#2883)
 
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.31.1...HEAD).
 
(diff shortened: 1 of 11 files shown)


bendk (approved):

My only request is that it would be great to get Ruby hooked up to the newer [bindgen-tests](https://github.com/mozilla/uniffi-rs/blob/main/bindgen-tests/).  I'm hoping to migrate away from using the older fixtures to using that crate.  There's a bunch of improvements there IMO and you could start by just enabling the `rust-traits` feature to only have those tests.  I think the main work would be porting the existing test runner to the new system ([here's the Python version](https://github.com/mozilla/uniffi-rs/blob/d7d4ff44f042925be392f9840aed898927f40b1f/bindgen-tests/python/src/lib.rs#L113-L153)]

Would you'd be interested in doing that, either in this PR or a follow-up?  I don't want to block this PR on that work, I just wanted to ask before merging.

saks:
Thank you @bendk! 

If possible, I'd like to do a migration to [bindgen-tests](https://github.com/mozilla/uniffi-rs/blob/main/bindgen-tests/) in a follow-up PR. I didn't know about plans to migrate to it. I'm happy to do it for ruby, thank for letting me know!

bendk:
That sounds good to me, let's merge this.

PR #2884, opened 2026-04-30
Update `askama` version to `0.16.0`

GuillaumeGomez:
This new version brings a lot of nice new features. You can see the list [here](https://github.com/askama-rs/askama/releases/tag/v0.16.0).

Not sure if any will be useful to you but in any case, they're available. :)


bendk (commented):
Seems like a nice set of new features and there's no breaking changes for us.  Let's merge.

bendk (approved):
Seems like a nice set of new features and there's no breaking changes for us.  Let's merge.

Issue #2458, opened 2025-02-28
Async code doesn't conform to Swift 6 Sendable


mhammond:
Executing `UNIFFI_TEST_SWIFT_VERSION=6 cargo test -p uniffi-fixture-futures` will fail with various Sendable conformance issues.

chris-marrero:
### `@Sendable` Solution
I think I've come up with a preliminary solution to this problem. The error primarily lies in the fact that we are passing a closure to a `Task` that is not marked as `@Sendable`. This is clearly incorrect as Task takes a `sending` closure.

So, we simply mark the closures to be `@Sendable`. 
```swift 
// Async.swift
{%- if ci.has_async_callback_interface_definition() %}
private func uniffiTraitInterfaceCallAsync<T>(
    makeCall: @Sendable @escaping () async throws -> T,
    handleSuccess: @Sendable @escaping (T) -> (),
    handleError: @Sendable @escaping (Int8, RustBuffer) -> ()
) -> UniffiForeignFuture {
    ...
}
```
```swift
// CallbackInterfaceImpl.swift
static let vtable: ... (
       ...
            let makeCall = {
                {% if meth.is_async() %}@Sendable {% endif %}() {% if meth.is_async() %}async {% endif %}throws -> ... in
                ...
            }
)
```
Unfortunately, in the `makeCall` closure, we capture the parameters in the async function defined by the user. That means every parameter must be `Sendable`. All of the numerical primitives are `Sendable`, but all of the other types are not. Specifically:
- `RustArcPtr` which is converted into `UnsafeMutableRawPointer`. This should be fine as Arc should imply `Sendable`
- `RustBuffer`
- `RustCallStatus`
- `ForeignBytes`
- potentially `Callback`
- `Reference` which is converted to `UnsafePointer`
- `MutReference` which is converted to `UnsafeMutablePointer`
- `Struct` for Records as they are not marked as `Sendable`

Some of these are pretty simple. `RustBuffer`, `RustCallStatus`, `ForeignBytes` can all easily be marked as `Sendable` in the bridging header. This is assuming these types are actually guaranteed to be `Send + Sync` by UniFFI. (One weird thing is that the Docker container's clang test fails on the below because it sees the attribute is invalid. But, the clang test works on my system. I wonder if it needs to be updated)
```swift
typedef struct __attribute__((swift_attr("@Sendable"))) RustBuffer
{
    ...
} RustBuffer;
```
[...]

Issue #2448, opened 2025-02-21
Sendable errors with Swift 6


mhammond:
If we run our tests against Swift 6 we continue to see errors related to `Sendable`. This is true even after #2318 and #2415.

In particular, traits and callbacks are problematic. Traits are probably easy (they are really just a variation on plain Objects, which were already fixed by the above PRs). Callbacks may be trickier because we can not simply annotate that user-implemented objects are sendable.

This PR is to track both fixing these and a process so our CI can ensure compliance continues while continuing to support Swift 5.

Issue #2046, opened 2024-03-24
UniFFI generated Swift code fails to compile "Swift 6" mode - due to mutation of shared state.


Sajjon:
Whenever we land #2045 (or other PR enabling `Sendable` types) we should try to ensure that UniFFI generated Swift code compiles (and works...) under "Swift 6" mode, that is with `enableExperimentalFeature("StrictConcurrency")`

<img width="1624" alt="Screenshot 2024-03-24 at 07 25 34" src="https://github.com/mozilla/uniffi-rs/assets/864410/d209954f-1660-4f7c-ba84-361b659ab189">

We can already now "emulate Swift 6" mode, by adding:

```swift
let swiftSettings: [SwiftSetting] = [
	.enableExperimentalFeature("StrictConcurrency"),
	.unsafeFlags(["-warnings-as-errors"])
]
```
And then use `swiftSettings: swiftSettings` in our Swift SPM targets, [like so](https://github.com/radixdlt/sargon/commit/8a4f6eb656fcf17d0a2de610ace47b23170fbe4a)

Doing this shows some issues with how UniFFI sets up Callbacks, specifically with `vtable` in `UniffiCallbackInterface<MY_TRAIT>` and `handleMap` in `FfiConverterType<MY_TRAIT>`.

The problem is they use static mutable properties, which is verboten in Swift 6.

There are a few ways to solve this, I think perhaps using `@TaskLocal` is the simplest way forward. See PointFree's episode of ShardState, in this code they have a very similar scenario, a global shared stated inside an enum namespace:

[`enum SharedLocals { @TaskLocal static var isAsserting = false; }`](https://github.com/pointfreeco/episode-code-samples/blob/0e202d8b3cbf81ec823d51ed3f362cf4c3753869/0271-shared-state-pt4/swift-composable-architecture/Sources/ComposableArchitecture/Shared.swift#L4)

Here is the documentation about [`@TaskLocal`](https://developer.apple.com/documentation/swift/tasklocal)

We could need to change how `uniffiCallbackInit<MY_TRAIT>` works
```swift
private func uniffiCallbackInit<MY_TRAIT>() {
    uniffi_<MY_PROJECT>_fn_init_callback_vtable_<MY_TRAIT>(&UniffiCallbackInterface<MY_TRAIT>.vtable)
}
```
[...]

PR #2886, opened 2026-05-04
swift: nonisolated(unsafe) on callback vtablePtr for Swift 6 strict concurrency

hideselfview:
Under Swift 6 strict concurrency (`SWIFT_VERSION = "6"` or higher), the generated `vtablePtr` static in `CallbackInterfaceImpl.swift` fails to compile:

```
error: static property 'vtablePtr' is not concurrency-safe because non-'Sendable' type
'UnsafePointer<UniffiVTableCallbackInterface…>' may have shared mutable state
```

A non-isolated `static let` whose type isn't `Sendable` is rejected because static storage is reachable from any thread. `UnsafePointer<UniffiVTableCallbackInterface…>` doesn't satisfy `Sendable` automatically: the pointee is the FFI-imported callback vtable struct, and Swift can't auto-derive `Sendable` for it.

The access pattern is genuinely safe:

- The pointer is initialized exactly once during static init and never mutated.
- The pointee struct's fields are C function pointers — their addresses don't change, and pointer-sized reads are atomic on every platform Swift targets.
- Those function pointers reach uniffi-generated trampolines that lock `UniffiHandleMap`'s internal `NSLock` before any mutation. Designed for concurrent invocation.

`nonisolated(unsafe)` is the standard escape hatch for that case — same pattern uniffi already uses adjacent to this code: `UniffiHandleMap` is `@unchecked Sendable` ("All mutation happens with this lock held, which is why we implement @unchecked Sendable"), generated object classes are `@unchecked Sendable`. The new `vtablePtr` static is the same kind of value — write-once shared storage, thread-safe by construction — and was missed when the 0.31 template restructured the vtable from a 1-element array literal to an explicit heap allocation. The array form happened to be strict-concurrency-clean because each subscript access produced a value-copy; the pointer form announces the shared storage to the type checker.

The annotation is wrapped in `#if swift(>=6)` because the Swift 5 language-mode parser rejects `nonisolated(unsafe)` on stored properties (it was added in SE-0420 for Swift 5.10+ and is gated to Swift 6 mode). Under Swift 5 the strict-concurrency check that needs the annotation isn't enforced anyway, so the unannotated form compiles cleanly there.

Surfaces under the existing Swift-6-tracking issues: #2448 (umbrella), #2046, #2458.

### Note on CI coverage

I tried adding a `UNIFFI_TEST_SWIFT_VERSION=6 cargo test -p uniffi-bindgen-tests-swift` step to `.circleci/config.yml` to lock in the `vtablePtr` fix, but it surfaces other pre-existing Swift 6 errors that aren't in scope for this PR. Specifically, `uniffiTraitInterfaceCallAsync` / `uniffiTraitInterfaceCallAsyncWithError` in `Async.swift` fail with:

```
uniffi_bindgen_tests.swift: error: passing closure as a 'sending' parameter risks
causing data races between code in the current task and concurrent execution of
the closure
   |
   |     let task = Task {
   |                `- error: ...
```

— the `Task { ... }` closure in those helpers captures non-`Sendable` callback parameters (`makeCall`, `handleSuccess`, `handleError`). Independent template fix; ideally a follow-up against #2448. Once that lands, adding the Swift 6 CI step makes sense as a separate PR.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
@@ -113,17 +113,9 @@ fileprivate struct {{ trait_impl }} {
         {%- endfor %}
     )
 
-    // Rust stores this pointer for future callback invocations, so it must live
-    // for the process lifetime (not just for the init function call).
-    //
-    // `nonisolated(unsafe)` is needed under Swift 6 strict concurrency: a
-    // non-isolated `static let` of a non-Sendable type is rejected, and the
-    // FFI-imported pointee struct can't be auto-derived `Sendable`. The
-    // pointee is initialized once during static init and never mutated; its
-    // fields are C function pointers (addresses, atomic reads) whose targets
-    // are uniffi-generated trampolines designed for concurrent invocation.
-    // Same escape hatch the adjacent `UniffiHandleMap`/generated object
-    // templates already use.
+    // `nonisolated(unsafe)` is needed under Swift 6 strict concurrency.
+    // This is safe because the pointee is initialized once during static init
+    // and never mutated by either side of the FFI.  Its fields are C function pointers.
     nonisolated(unsafe) static let vtablePtr: UnsafePointer<{{ vtable|ffi_type_name }}> = {
         let ptr = UnsafeMutablePointer<{{ vtable|ffi_type_name }}>.allocate(capacity: 1)
         ptr.initialize(to: vtable)
--- a/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
@@ -115,11 +115,33 @@ fileprivate struct {{ trait_impl }} {
 
     // Rust stores this pointer for future callback invocations, so it must live
     // for the process lifetime (not just for the init function call).
+    //
+    // `nonisolated(unsafe)` is needed under Swift 6 strict concurrency: a
+    // non-isolated `static let` of a non-Sendable type is rejected, and the
+    // FFI-imported pointee struct can't be auto-derived `Sendable`. The
+    // pointee is initialized once during static init and never mutated; its
+    // fields are C function pointers (addresses, atomic reads) whose targets
+    // are uniffi-generated trampolines designed for concurrent invocation.
+    // Same escape hatch the adjacent `UniffiHandleMap`/generated object
+    // templates already use.
+    //
+    // The annotation requires Swift 6 language mode. In Swift 5 mode the
+    // parser rejects `nonisolated(unsafe)` on stored properties, but the
+    // strict-concurrency check that needs the annotation isn't enforced
+    // there either, so the unannotated form compiles cleanly.
+#if swift(>=6)
+    nonisolated(unsafe) static let vtablePtr: UnsafePointer<{{ vtable|ffi_type_name }}> = {
+        let ptr = UnsafeMutablePointer<{{ vtable|ffi_type_name }}>.allocate(capacity: 1)
+        ptr.initialize(to: vtable)
+        return UnsafePointer(ptr)
+    }()
+#else
     static let vtablePtr: UnsafePointer<{{ vtable|ffi_type_name }}> = {
         let ptr = UnsafeMutablePointer<{{ vtable|ffi_type_name }}>.allocate(capacity: 1)
         ptr.initialize(to: vtable)
         return UnsafePointer(ptr)
     }()
+#endif
 }
 
 private func {{ callback_init }}() {
--- a/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
@@ -124,24 +124,11 @@ fileprivate struct {{ trait_impl }} {
     // are uniffi-generated trampolines designed for concurrent invocation.
     // Same escape hatch the adjacent `UniffiHandleMap`/generated object
     // templates already use.
-    //
-    // The annotation requires Swift 6 language mode. In Swift 5 mode the
-    // parser rejects `nonisolated(unsafe)` on stored properties, but the
-    // strict-concurrency check that needs the annotation isn't enforced
-    // there either, so the unannotated form compiles cleanly.
-#if swift(>=6)
     nonisolated(unsafe) static let vtablePtr: UnsafePointer<{{ vtable|ffi_type_name }}> = {
         let ptr = UnsafeMutablePointer<{{ vtable|ffi_type_name }}>.allocate(capacity: 1)
         ptr.initialize(to: vtable)
         return UnsafePointer(ptr)
     }()
-#else
-    static let vtablePtr: UnsafePointer<{{ vtable|ffi_type_name }}> = {
-        let ptr = UnsafeMutablePointer<{{ vtable|ffi_type_name }}>.allocate(capacity: 1)
-        ptr.initialize(to: vtable)
-        return UnsafePointer(ptr)
-    }()
-#endif
 }
 
 private func {{ callback_init }}() {
--- a/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
@@ -113,6 +113,9 @@ fileprivate struct {{ trait_impl }} {
         {%- endfor %}
     )
 
+    // Rust stores this pointer for future callback invocations, so it must live
+    // for the process lifetime (not just for the init function call).
+    //
     // `nonisolated(unsafe)` is needed under Swift 6 strict concurrency.
     // This is safe because the pointee is initialized once during static init
     // and never mutated by either side of the FFI.  Its fields are C function pointers.


bendk (commented):
This is awesome, it's nice to have Swift experts helping us out through these details.  I think this should get a CHANGELOG entry, but other than than I'm happy to merge as-is.  The PR is flagged as a draft though, is there something else you think is needed before this is ready to merge?

hideselfview:
> it's nice to have Swift experts helping us out through these details.

:embarassed: one of the reasons these PRs are in draft is because I (@dminkovsky) am not a Swift expert and this (and my other recent PR) is Claude Code output. I couldn't figure out why it opened these PRs without signing off as CC like it does in my personal projects like 100 times a day, and wanted to add that disclaimer! My apologies.

In any case! I played with this output, thought about it to the degree I am qualified (meaning, not much), have tested it in the context of my app. But I am not a Swift expert...

Edit: my other recent PR https://github.com/mozilla/uniffi-rs/pull/2887

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
```diff
@@ -115,11 +115,33 @@ fileprivate struct {{ trait_impl }} {
...
     // for the process lifetime (not just for the init function call).
+    //
+    // `nonisolated(unsafe)` is needed under Swift 6 strict concurrency: a
+    // non-isolated `static let` of a non-Sendable type is rejected, and the
+    // FFI-imported pointee struct can't be auto-derived `Sendable`. The
+    // pointee is initialized once during static init and never mutated; its
+    // fields are C function pointers (addresses, atomic reads) whose targets
+    // are uniffi-generated trampolines designed for concurrent invocation.
+    // Same escape hatch the adjacent `UniffiHandleMap`/generated object
+    // templates already use.
```
bendk: I'm not sure if we need all of these words.  Maybe this could be shorted to:

```
// `nonisolated(unsafe)` is needed under Swift 6 strict concurrency.
// This is safe because the pointee is initialized once during static init
// and never mutated by either side of the FFI.   Its fields are C function pointers.
```

bendk:
No problems.  As you might have guessed, we're not Swift experts either even though we're trying to generate Swift code.  It would be great to get a green light from someone who really knows though, I'll see if I can get someone to check it over.

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
```diff
@@ -115,11 +115,33 @@ fileprivate struct {{ trait_impl }} {
...
+    //
+    // The annotation requires Swift 6 language mode. In Swift 5 mode the
+    // parser rejects `nonisolated(unsafe)` on stored properties, but the
+    // strict-concurrency check that needs the annotation isn't enforced
+    // there either, so the unannotated form compiles cleanly.
+#if swift(>=6)
```
Cramsden: Are you sure we need this version check? AFAIK nonisolated(unsafe) should compile just fine in swift 5

ih-codes: Agreed, when I tested in a quick demo project, `nonisolated` causes the described warning, but `nonisolated(unsafe)` disables compiler checks and emits no warnings in Swift 5. So I would verify whether this diverging code is even necessary. 🤔 

- Demo Swift 5 language mode with `nonisolated` (produces warning):

Image


- Demo Swift 5 language mode with `nonisolated(unsafe)` (no warnings):

Image

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
ih-codes: Just piggybacking on this to say, nice documentation. It's good practice to explain why you use `nonisolated(unsafe)` or `@unchecked Sendable` like done in this PR, because basically that's communicating "I'm opting out of compiler checks and promising to manually make this type concurrency safe".

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
dminkovsky: Thanks, great question, I will double check.

dminkovsky: Yes, thank you all, you are correct:

---


```
$ swift --version
swift-driver version: 1.127.15 Apple Swift version 6.2.4 (swiftlang-6.2.4.1.4 clang-1700.6.4.2)
Target: arm64-apple-macosx26.0
```

### Annotated form

`/tmp/test_nonisolated.swift`:

```swift
struct VTable {
    var fn: @convention(c) () -> Void
}

fileprivate struct Holder {
    nonisolated(unsafe) static let vtablePtr: UnsafePointer<VTable> = {
        let ptr = UnsafeMutablePointer<VTable>.allocate(capacity: 1)
        ptr.initialize(to: VTable(fn: { }))
        return UnsafePointer(ptr)
    }()
}

print(Holder.vtablePtr)
```

```
$ cd /tmp && swiftc -swift-version 5 test_nonisolated.swift -o test_swift5 2>&1; echo "---exit: $?---"
---exit: 0---

$ cd /tmp && swiftc -swift-version 5 -strict-concurrency=complete test_nonisolated.swift -o test_swift5_strict 2>&1; echo "---exit: $?---"
---exit: 0---

$ cd /tmp && swiftc -swift-version 6 test_nonisolated.swift -o test_swift6 2>&1; echo "---exit: $?---"
---exit: 0---
```

| Compile flags | Result |
|---|---|
| `-swift-version 5` | clean |
| `-swift-version 5 -strict-concurrency=complete` | clean |
| `-swift-version 6` | clean |

### Unannotated form

`/tmp/test_unannotated.swift`:

```swift
struct VTable {
    var fn: @convention(c) () -> Void
}

fileprivate struct Holder {
    static let vtablePtr: UnsafePointer<VTable> = {
        let ptr = UnsafeMutablePointer<VTable>.allocate(capacity: 1)
        ptr.initialize(to: VTable(fn: { }))
        return UnsafePointer(ptr)
    }()
}

print(Holder.vtablePtr)
```

```
$ cd /tmp && swiftc -swift-version 5 test_unannotated.swift -o test_un_5 2>&1; echo "---exit: $?---"
---exit: 0---

$ cd /tmp && swiftc -swift-version 5 -strict-concurrency=complete test_unannotated.swift -o test_un_5_strict 2>&1; echo "---exit: $?---"
test_unannotated.swift:6:16: warning: static property 'vtablePtr' is not concurrency-safe because non-'Sendable' type 'UnsafePointer<VTable>' may have shared mutable state; this is an error in the Swift 6 language mode [#MutableGlobalVariable]
 4 |
 5 | fileprivate struct Holder {
 6 |     static let vtablePtr: UnsafePointer<VTable> = {
   |                |- warning: static property 'vtablePtr' is not concurrency-safe because non-'Sendable' type 'UnsafePointer<VTable>' may have shared mutable state; this is an error in the Swift 6 language mode [#MutableGlobalVariable]
   |                |- note: add '@MainActor' to make static property 'vtablePtr' part of global actor 'MainActor'
   |                `- note: disable concurrency-safety checks if accesses are protected by an external synchronization mechanism
 7 |         let ptr = UnsafeMutablePointer<VTable>.allocate(capacity: 1)
 8 |         ptr.initialize(to: VTable(fn: { }))

Swift.UnsafePointer:1:23: note: generic struct 'UnsafePointer' does not conform to the 'Sendable' protocol
1 | @frozen public struct UnsafePointer<Pointee> : Copyable where Pointee : ~Copyable {
  |                       `- note: generic struct 'UnsafePointer' does not conform to the 'Sendable' protocol
2 |     public let _rawValue: Builtin.RawPointer
3 |     public init(_ _rawValue: Builtin.RawPointer)

[#MutableGlobalVariable]: <https://docs.swift.org/compiler/documentation/diagnostics/mutable-global-variable>
---exit: 0---

$ cd /tmp && swiftc -swift-version 6 test_unannotated.swift -o test_un_6 2>&1; echo "---exit: $?---"
test_unannotated.swift:6:16: error: static property 'vtablePtr' is not concurrency-safe because non-'Sendable' type 'UnsafePointer<VTable>' may have shared mutable state [#MutableGlobalVariable]
 4 |
 5 | fileprivate struct Holder {
 6 |     static let vtablePtr: UnsafePointer<VTable> = {
   |                |- error: static property 'vtablePtr' is not concurrency-safe because non-'Sendable' type 'UnsafePointer<VTable>' may have shared mutable state [#MutableGlobalVariable]
   |                |- note: add '@MainActor' to make static property 'vtablePtr' part of global actor 'MainActor'
   |                `- note: disable concurrency-safety checks if accesses are protected by an external synchronization mechanism
 7 |         let ptr = UnsafeMutablePointer<VTable>.allocate(capacity: 1)
 8 |         ptr.initialize(to: VTable(fn: { }))

Swift.UnsafePointer:1:23: note: generic struct 'UnsafePointer' does not conform to the 'Sendable' protocol
1 | @frozen public struct UnsafePointer<Pointee> : Copyable where Pointee : ~Copyable {
  |                       `- note: generic struct 'UnsafePointer' does not conform to the 'Sendable' protocol
2 |     public let _rawValue: Builtin.RawPointer
3 |     public init(_ _rawValue: Builtin.RawPointer)

[#MutableGlobalVariable]: <https://docs.swift.org/compiler/documentation/diagnostics/mutable-global-variable>
---exit: 1---
```

| Compile flags | Result |
|---|---|
| `-swift-version 5` | clean |
| `-swift-version 5 -strict-concurrency=complete` | **warning** — "this is an error in the Swift 6 language mode" |
| `-swift-version 6` | **error** — `not concurrency-safe` |

dminkovsky: I pushed 94976860, thank you!

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
dminkovsky: Totally, I pushed @bendk's version verbatim 👍 .

dminkovsky:
Screenshot 2026-05-09 at 7 43 07 AM

I was thinking of marking the PR "Ready" but I can't.

badboy:
> Screenshot 2026-05-09 at 7 43 07 AM
> 
> I was thinking of marking the PR "Ready" but I can't.

did that for you.

badboy:
Looks like this at least needs a rebase against `main`

hideselfview:
> > I was thinking of marking the PR "Ready" but I can't.
> 
> did that for you.

Thanks!

> Looks like this at least needs a rebase against `main`

Done.

bendk (approved):
This looks great to me, let's merge.

Issue #2771, opened 2025-12-29
Support Cancelling/Dropping Rust Futures


TheBlueMatt:
#1768 tried to add support for making Futures from Swift cancellable on the Rust side, which has some thorny API questions so was ultimately abandoned however, Rust's native Futures are always allowed to be Dropped without any kind if cancellation logic. Thus, making the native Rust Futures passed to Swift cancellable should be doable without any kind of strange API on the Rust side.

This would make mapping Rust APIs to Swift a bit more "native", as Rust libraries may expect callers to simply Drop Futures they no longer need whereas Swift clients cannot do that.

https://github.com/mozilla/uniffi-rs/pull/1768#issuecomment-3696776499 pointed to specific code that would need to be updated.

I probably won't have time for this immediately but might at some point.

pronebird:
Just make an object wrapping tokio cancellation token perhaps and feed it into the future for graceful shutdown.

bes:
I too wanted "automatic cancellation" on the Swift side, but then I did as @pronebird said, and honestly it feels good enough

```swift
    // Wraps a tokio::CancellationToken
    let cancelToken = CancelToken.init()
    await withTaskCancellationHandler {
        await myRustAsyncFn(cancelToken: cancelToken)
    } onCancel: {
        cancelToken.cancel()
    }
```

Then I need to spawn a tokio task on the Rust side. This doesn't poll using uniffi's built-in stuff, but I still use tokio everywhere internally so it isn't an issue for me...

TheBlueMatt:
At least for my use-case I just did a trivial `sed` pass on the generated swift bindings to have native cancellation. The only issue is it doesn't make sense to do it universal because then all of your async methods get a `throws` an can throw a `CancellationError`. To upstream this we'd either have to be okay with every async method getting two copies (one that can throw and be cancelled and one that cannot) which is a bit weird but I suppose doable, or somehow allow annotations as to which methods should be cancellable.

```
#!/bin/sh
perl -i -0pe 's/private let UNIFFI_RUST_FUTURE_POLL_MAYBE_READY: Int8 = 1/private let UNIFFI_RUST_FUTURE_POLL_MAYBE_READY: Int8 = 1
private let UNIFFI_RUST_FUTURE_POLL_CANCELLED: Int8 = 2

fileprivate func uniffiRustCallAsyncCancellable<F, T>(
    rustFutureFunc: () -> UInt64,
[...]

PR #2887, opened 2026-05-04
swift: propagate Task.cancel() to Rust drop

hideselfview:
Wires Swift `Task.cancel()` through `withTaskCancellationHandler` to call the auto-emitted `ffi_*_rust_future_cancel_*` symbols, then throws `CancellationError` on `CALL_CANCELLED`. Gated behind a `cancellable` opt-in so existing async API surfaces don't gain a `throws` clause.

Today, the Swift `uniffiRustCallAsync` template uses plain `withUnsafeContinuation` for its poll loop. When the calling Swift `Task` is cancelled, `Task.isCancelled` becomes true but the await keeps suspending; the Rust future continues being polled and runs to natural completion. Drop-time guards (`AbortOnDrop`, `CancelOnDrop`, sqlx connection release, reqwest abort) on the Rust side never fire on cancel.

## Opt-in

The author of the Rust function attests that the future is drop-safe by adding `cancellable`:

```rust
#[uniffi::export(cancellable)]
pub async fn long_running() -> Foo { ... }

// also valid on an impl block, applies to every async fn in it
#[uniffi::export(cancellable)]
impl SomeObject { ... async fn ... }
```

In UDL: `[Async, Cancellable] void long_running();`

The annotation rides on top of the existing `#[uniffi::export]` surface — same kw-arg style as `async_runtime`. It is rejected on traits (foreign callback futures awaited by Rust use a different `UniffiForeignFutureTask` mechanism, not in scope here).

## Mechanism

Three template-level changes:

- `uniffiRustCallAsync` keeps its existing signature (no `cancelFunc`, no cancel-handler wrap). A new `uniffiRustCallAsyncCancellable` is emitted alongside whenever any callable in the namespace opts in — same body but with `cancelFunc` and the poll loop wrapped in `withTaskCancellationHandler { … } onCancel: { cancelHandle.cancel(cancelHandle.handle) }`. The cancel function pointer + handle are captured in a small `@unchecked Sendable` holder so the `onCancel` closure satisfies the `@Sendable` requirement.
- `Helpers.swift`'s `case CALL_CANCELLED: fatalError(...)` becomes `throw CancellationError()` (deallocating the error buffer first). This is strictly more correct and only fires on cancellable async paths; sync paths and non-cancellable async paths can never reach it.
- `call_async` macro picks `uniffiRustCallAsyncCancellable` and emits `cancelFunc: <ffi_rust_future_cancel symbol>` only at cancellable callsites. The `throws` / `is_try` macros add `throws` / `try` when the callable is cancellable, the same way they already do when it `throws` a Rust-side error.

After this, when a Swift Task running a cancellable callable is cancelled the Rust scheduler synchronously fires the registered continuation with `Ready`, the poll loop exits, `complete()` runs and drops the inner Rust future (firing all the Drop guards), and Swift sees `CALL_CANCELLED` and throws `CancellationError`. Non-cancellable callables behave exactly like before this PR.

The `is_cancellable` bit flows through metadata: `FnMetadata` / `MethodMetadata` / `ConstructorMetadata` / `TraitMethodMetadata` each grow a bool, contract version bumped 30 → 31, readers and writers updated. UDL parses `[Cancellable]` and rejects it without `[Async]`.

## Tests

`fixtures/futures` gains a `swift_cancel_hang` async function (`#[uniffi::export(cancellable)]`) that hangs forever holding a drop guard, plus drop and poll counters. The Swift test cancels a Task running it, asserts `CancellationError` is thrown, and asserts the drop counter incremented exactly once. The test spins on the poll counter to wait for the first poll deterministically rather than sleeping (otherwise cancel can race the first poll, the scheduler goes Empty → Cancelled with no callback fired, the next poll short-circuits on `is_cancelled()` before running the body, and the drop guard is never constructed).

Also fixes the existing `useSharedResource` cancel test that previously had a `FIXME` saying cancel didn't actually cancel — `use_shared_resource` is now `cancellable`, and the test does a `do/catch` for `CancellationError`.

## Notes

Tracks #2771. Closed PR #1768 from 2023 carried roughly the same approach (plus opt-in plumbing) and was closed for lack of energy, not technical reasons.

All commits in this PR were authored with [Claude Code](https://claude.com/claude-code) (Opus 4.7) — `Co-Authored-By` trailer on each commit.

### DIFF
--- a/fixtures/futures/src/lib.rs
+++ b/fixtures/futures/src/lib.rs
@@ -162,7 +162,7 @@ impl Drop for SwiftCancelDropGuard {
     }
 }
 
-#[uniffi::export]
+#[uniffi::export(cancellable)]
 pub async fn swift_cancel_hang() -> Result<(), MyError> {
     let _guard = SwiftCancelDropGuard;
     SWIFT_CANCEL_POLL_COUNT.fetch_add(1, Ordering::SeqCst);
@@ -412,7 +412,7 @@ pub enum AsyncError {
     Timeout,
 }
 
-#[uniffi::export(async_runtime = "tokio")]
+#[uniffi::export(async_runtime = "tokio", cancellable)]
 pub async fn use_shared_resource(options: SharedResourceOptions) -> Result<(), AsyncError> {
     use once_cell::sync::Lazy;
     use tokio::{
--- a/fixtures/futures/src/lib.rs
+++ b/fixtures/futures/src/lib.rs
@@ -5,7 +5,10 @@
 use std::{
     future::Future,
     pin::Pin,
-    sync::{Arc, Mutex, MutexGuard},
+    sync::{
+        atomic::{AtomicU32, Ordering},
+        Arc, Mutex, MutexGuard,
+    },
     task::{Context, Poll, Waker},
     thread,
     time::Duration,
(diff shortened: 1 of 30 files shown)


bendk:
If I remember correctly, the issue here is that this makes all async functions throw on Swift, even for users that don't ever call `Task.cancel`. I remember quite a bit of pushback on this.  I think this has to be opt-in somehow and the main question is how should users opt-in to it?  Globally, per-function, some other way?

dminkovsky:
> If I remember correctly, the issue here is that this makes all async functions throw on Swift, even for users that don't ever call `Task.cancel`. I remember quite a bit of pushback on this. I think this has to be opt-in somehow and the main question is how should users opt-in to it? Globally, per-function, some other way?

oh yeah, duh great call! Sorry I was very focused on my use-case. I added a `cancellable` option to the `export` which seems to be a good fit! But I haven't played with this yet, or tried it end-to-end in my codebase. I guess the tests here speak for themselves to some extent, but I need to read them :)

fixtures/futures/src/lib.rs
```diff
@@ -352,7 +406,7 @@ pub enum AsyncError {
     Timeout,
 }
 
-#[uniffi::export(async_runtime = "tokio")]
+#[uniffi::export(async_runtime = "tokio", cancellable)]
```
bendk, line 415: I like the test changes, but could you remove the "swift_" prefix?  I think other languages might want to use the same mechanism.

Also, even better would be adding a cancellation test to `bindgen-tests/lib/src/futures.rs`, since I'm hoping to move/consolidate the tests there.  It's not a blocker or anything though, since there's no cancellation test at the moment.

uniffi_bindgen/src/bindings/swift/templates/Helpers.swift
```diff
@@ -102,7 +102,8 @@ private func uniffiCheckCallStatus<E: Swift.Error>(
             }
 
         case CALL_CANCELLED:
-            fatalError("Cancellation not supported yet")
+            callStatus.errorBuf.deallocate()
```
bendk, line 105: Does Rust allocate the buffer in this case?  I think it doesn't and this could be removed, but it's not a big deal either way.

bendk (changes requested):
Sorry for taking forever to review this one, I got distracted with some other work.

The main code looks pretty good to me, I just had some minor comments.  My main question is still how to configure this.  Using a macro attribute doesn't feel right to me since it feels very swift-specific.  For example, Kotlin always supports cancellation and I think it would just ignore that attribute.  There might be other external bindings that work the same way.

Because of that, maybe this could be done in a Swift-specific section the TOML file using syntax similar to the rename config:

```
[bindings.swift.async_cancellable]
# Flag these functions/methods as cancellable
"my_function" = true
"MyObject.method" = true
# Functions/methods are non-cancellable by default, but you can explicitly specify that if you want
"MyObject.method2" = false
```

Again, sorry for taking so long to respond on this one.  I feel bad that I'm suggesting you throw away all that macro code, but I do think this configuration will improve things for our users.

hideselfview:
hey no worries, thanks so much for the feedback! I was just thinking I need to get back around to this myself. I expect to get back to this via upstream work in 2-4 weeks. Best!

PR #2895, opened 2026-05-07
Re-do the Docker image, now based on the Swift image

### LINKED ISSUE #2892: All docker docs and utilities are way out of date.
Alot of our docs and shell scripts refer to a very old `rfkelly` owned docker image. However, in practice our CI is using https://hub.docker.com/r/janerik/uniffi-ci-test.

It looks like the stuff in `./docker` wasn't used with the new image - should we update that docker directory to point at the new image with the idea it might still be partially useful to someone, or just delete it entirely and update the docs, or something else - @badboy wdyt?

via #2801

badboy:
This is now significantly bigger (~~7.6~~ 7.0 GB), but that's unavoidable with
the new way Swift is shipped and all the other tooling we need.

Fixes #2892

---

fwiw I'm trying to figure out if we can push that image to ghcr.io/mozilla/uniffi to make it more "official"

### DIFF
--- a/fixtures/rename/uniffi.toml
+++ b/fixtures/rename/uniffi.toml
@@ -223,9 +223,9 @@ lift = "urllib.parse.urlsplit({})"
 lower = "urllib.parse.urlunsplit({})"
 
 [bindings.kotlin.custom_types.RenamedUrl]
-type_name = "URL"
-imports = [ "java.net.URL" ]
-lift = "URL({})"
+type_name = "URI"
+imports = [ "java.net.URI" ]
+lift = "URI({})"
 lower = "{}.toString()"
 
 [bindings.swift.custom_types.RenamedUrl]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,7 +76,7 @@ jobs:
       - build-api-docs
   Rust and Foreign Language tests:
     docker:
-      - image: janerik/uniffi-ci-test:latest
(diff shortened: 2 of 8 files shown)


fixtures/rename/uniffi.toml
```diff
@@ -223,9 +223,9 @@ lift = "urllib.parse.urlsplit({})"
 lower = "urllib.parse.urlunsplit({})"
 
 [bindings.kotlin.custom_types.RenamedUrl]
-type_name = "URL"
-imports = [ "java.net.URL" ]
-lift = "URL({})"
+type_name = "URI"
```
mhammond, line 226: mildly surprised by this, but I guess 🤷

mhammond (approved):
Thanks heaps for doing this!

fixtures/rename/uniffi.toml
mhammond: actually, this change is what's broken CI?

badboy: 🤔 oh, I think because I need to keep the `type_name`?
Regardless the `java.net.URL` constructor is deprecated: https://docs.oracle.com/en/java/javase//20/docs/api/java.base/java/net/URL.html#constructor-deprecation
And that fails tests then because we set `-Werror`

I'll double-check this again.

badboy: Ah, I simply forgot to commit the part of the code that was also needed.

Issue #2740, opened 2025-11-19
Kotlin: JNA direct mapping puts unsigned integer values to signed `Short` during checksum checks on Arm32 Android devices


paxbun:
JNA version: 5.18.1

We're experiencing a weird API checksum check failure that occurs only in release mode Kotlin on Arm32 Android devices, which doesn't happen in debug mode or on Aarch64 devices. This is the first version of our app that uses the direct-mapped JNA introduced in #2229.

This is the actual Firebase Crashlytics log; the part that throws the exception is as follows.

<img width="821" height="358" alt="Image" src="https://github.com/user-attachments/assets/c711cad8-9dc0-4b66-91a4-350b2ed9f804" />

```kotlin
if (uniffi_library_name_checksum_func_api_name() != 62517.toShort()) {
    throw RuntimeException("UniFFI API checksum mismatch: try cleaning and rebuilding your project")
}
```

It appears that this issue occurs only in release mode Kotlin when the checksum is greater than 32767 (2^15 - 1) (which should be negative on the Kotlin side), regardless of the Cargo profile used on the Rust side.

As shown in the image above, the released version doesn't log the actual value returned from the checksum functions, so I ran the example app project and logged the value:

<img width="1111" height="79" alt="Image" src="https://github.com/user-attachments/assets/9557eae2-e124-4b54-91e8-15e3652e7dbd" />

<img width="1268" height="344" alt="Image" src="https://github.com/user-attachments/assets/2dd7bac5-df2b-4d92-817d-b5447b71d56d" />

And surprisingly, the returned **signed** short value contained 53053, which is the reason why the checksum check is failing.

This is the disassembly of `uniffi_gobley_example_todolist_checksum_func_create_entry_with`:

```
> llvm-objdump -d ./libgobley_example_todolist.so | grep -A 3 uniffi_gobley_example_todolist_checksum_func_create_entry_with
00015e08 <uniffi_gobley_example_todolist_checksum_func_create_entry_with>:
   15e08: f64c 703d     movw    r0, #0xcf3d
   15e0c: 4770          bx      lr
```

You can see that it doesn't perform an additional sign extension, so the upper 2 bytes of the `r0` register are zero-padded, in contrast to `func1` in the following code.

<img width="1309" height="808" alt="Image" src="https://github.com/user-attachments/assets/5fed2a06-a17a-4bc6-8576-f6ec3e641b05" />

[...]

PR #2897, opened 2026-05-07
Kotlin: widen unsigned small integer direct return carriers

lwz23:
## Summary

This is a candidate fix for #2740. It changes the Kotlin/JNA direct return carrier for unsigned sub-word FFI return values so that `FfiType::UInt8` and `FfiType::UInt16` are declared as returning `Int` from direct JNA functions, rather than `Byte` / `Short`.

The public Kotlin API remains unchanged: `u8` still lifts to `UByte`, and `u16` still lifts to `UShort`.

## Motivation

The issue discussion identified that ARM32 returns signed and unsigned fundamental data types smaller than 4 bytes differently: signed values are sign-extended, while unsigned values are zero-extended. If Kotlin/JNA declares a Rust `u16` / C `uint16_t` FFI function as returning `Short`, that declaration can disagree with the actual unsigned return ABI.

This patch avoids making the fix checksum-specific. It treats checksum functions and regular user functions consistently by separating the Kotlin/JNA direct return carrier mapping from the existing general by-value FFI mapping.

## Changes

- Add a Kotlin generator helper/filter for direct FFI return carrier names.
- Map direct returns for `FfiType::UInt8` and `FfiType::UInt16` to `Int`.
- Keep parameters, struct fields, and buffer serialization on the existing `Byte` / `Short` carriers.
- Update checksum comparison to compare the `Int` return value directly.
- Add `Int` overloads to `FfiConverterUByte.lift` and `FfiConverterUShort.lift` so widened direct returns still lift to the public Kotlin unsigned types.
- Extend Kotlin primitive tests to cover unsigned values outside the signed sub-word range:
  - `output_u8() == 255.toUByte()`
  - `output_u16() == 65535.toUShort()`
  - boundary roundtrips for `255.toUByte()` and `65535.toUShort()`

## Verification

Run locally on macOS/JVM:

- `cargo check -p uniffi_bindgen`
- `cargo test test_primitive_types -- --nocapture` from `bindgen-tests/kotlin`
- `cargo test -p uniffi-bindgen-tests-kotlin -- --nocapture`

The full Kotlin bindgen test suite passed locally: 18 passed, 0 failed.

I have not yet validated this on an Android ARM32 device, so I am opening this as a draft/candidate fix for review.

## Note

I am a PhD student currently exploring pain points that Firefox and related Mozilla projects encounter while adopting Rust. I would like to contribute to Firefox's Rustification effort where I can. If this approach is not the right direction for UniFFI, feedback would be very welcome, and I am happy to keep iterating on the fix.


bendk:
Wow, this is pretty interesting, but I wonder about this statement:

> declaration can disagree with the actual unsigned return ABI.

Is there documentation for this?  IIUC, it works because the ABI returns the value using a machine word in all cases so .  However, how does this give us extra guarantees compared to the current code?  TBH, I keep forgetting why the sign/zero extension causes the issue in the first place, so I'm probably just missing something.

I'll say again that my preference is to move away from JNA and towards JNI and I'm working through a branch that does just that.  However, I'm fine with merging short-term fixes in the meantime.

paxbun:
> Is there documentation for this? IIUC, it works because the ABI returns the value using a machine word in all cases so .

IIUC the documentation says that the user can use `Short` or `Byte` to retrieve `UShort` or `UByte` value returned from functions, but the actual JNA code shows that they're regarding such values as `Int` internally, as I mentioned in the following comment.

https://github.com/mozilla/uniffi-rs/issues/2740#issuecomment-3557517041

This fix implemented a potential fix I suggested as "but we need to put something like ... would it be okay to do this? Or do we have to introduce another function like ffi_type_label_return_type in gen_kotlin/mod.rs?", so this fix will work on all kinds of Android ABIs (not sure about other CPU architectures supported by Linux), but AFAIK, there's no documentation for this.

mhammond:
in #2740 @bendk asked "However, doesn't this same issue affect all user functions that return a u16 or u8 as well?" - but these new tests all pass on main. What's special about the checksums we aren't capturing?

lwz23:
Thanks, that's a good point. The primitive runtime tests prove that the widened return carrier still works for normal Kotlin APIs, but they don't reproduce the architecture-sensitive checksum failure on CI. The reason is that normal exported functions go through the public FfiConverter lift path, while checksum validation compares the raw JNA direct return carrier directly.

So I agree these tests are not a full regression for the Android ARM32 checksum case. They are mostly coverage to ensure the new direct-return carrier mapping does not break normal u8/u16 returns. I can tighten this PR by adding generator-level coverage for the checksum declaration/comparison, or a checksum-specific Kotlin fixture if that's preferred.


> in #2740 @bendk asked "However, doesn't this same issue affect all user functions that return a u16 or u8 as well?" - but these new tests all pass on main. What's special about the checksums we aren't capturing?

mhammond:
Still trying to get my head around this, so please excuse silly/repeated questions :)

> but they don't reproduce the architecture-sensitive checksum failure on CI

CI obvs doesn't have this obscure platform - I wonder if that's possible? But do you believe this new tests *would* fail on those platforms?

> normal exported functions go through the ... lift path, while checksum validation compares the raw JNA direct return carrier directly.

These patches touch `lift` though?

lwz23:
> CI obvs doesn't have this obscure platform - I wonder if that's possible? But do you believe this new tests _would_ fail on those platforms?

On the tests: I don't think the new generator-level checksum test would reproduce the original ARM32 runtime failure by itself. It only verifies the generated Kotlin shape: the checksum FFI declaration now uses an `Int` direct-return carrier, and the checksum comparison no longer converts the expected value to `Short`.

The primitive Kotlin runtime tests have a different purpose. They check that normal exported `u8`/`u16` functions still work after widening the direct-return carrier to `Int`, by lifting that `Int` back into the public Kotlin `UByte`/`UShort` type. Those are the tests I would expect to be meaningful on an affected platform for normal exported functions returning values above the signed range, but they are not a full substitute for an Android ARM32 checksum runtime test.

> These patches touch `lift` though?

Yes, the patch touches `lift` intentionally. Checksum validation itself does not go through `lift`; it calls the integrity-checking FFI functions and compares the raw direct-return carrier. But normal exported functions do go through `lift`, and after this patch their raw `u8`/`u16` direct-return carrier is `Int`, so the converters need `lift(Int)` overloads to preserve the public Kotlin API type (`UByte`/`UShort`).

So the patch has two related parts:
1. use `Int` as the JNA direct-return carrier for unsigned small integer returns;
2. handle the two consumers of that carrier: normal functions via `lift(Int)`, and checksum validation via direct `Int` comparison.

jkmassel:
Dropping in to note that we independently arrived at similar conclusions after reproducing this from scratch on our own – I only discovered this PR (and the associated issue) as I was about to open my own 😅. 

I can confirm it's a bug in AOSP, and can reproduce the issue on-device (using a Galaxy S9 and forcing the app to run in 32-bit mode). See https://github.com/jkmassel/uniffi-armv7-aot-checksum-bug if you're interested. Running this PR against the test cases in that repo resolved them on-device.

Unless you can run 32-bit Android emulators in CI, it'll probably be tricky to put this under test in an automated way. If I can help at all though, please let me know!

jkmassel (changes requested):
A few thoughts, hope they're helpful:

1. IMHO this appears to be the correct approach – widening the type (as opposed to bit masking) is a simple + effective fix with little runtime overhead.
2. Signed types don't need widening because Android correctly handles them.
3. Should the old `lift` implementations be removed?
4. The changes for `java.net.URL => java.net.URI` and the CI config seem unrelated to this PR. I wonder if they could be pushed to a future PR?

lwz23:
Thanks, this is very helpful. The independent on-device confirmation is especially useful here, since it covers exactly the Android 32-bit runtime case that CI does not currently exercise.

I pushed a small cleanup commit that removes the unrelated CI and `java.net.URL`/`URI` test changes from this PR, so the diff should now be scoped to the unsigned small integer return-carrier fix and its tests.

On the old `lift` implementations: I kept `lift(Byte)` / `lift(Short)` because they are still used by the existing `FfiConverter<UByte, Byte>` / `FfiConverter<UShort, Short>` paths, including buffer reads and non-return-carrier uses. The new `lift(Int)` overloads are only for direct FFI returns after widening the JNA carrier to `Int`.

mhammond:
That's all very helpful, thanks!

CI failing here does seem odd, I'll try and dig into that some more. To clarify some of my earlier comments, I'm not as bothered by not having this odd architecture in CI (obviously would take it if it was easy though!) but my poorly phrased questions were really about trying to get a good sense of if it would hypothetically fail - I was really just trying to ensure what parts of this were actually guesses. This extra clarity is great, thank you both/all!

I will however leave the final approval to @bendk as he knows this stuff way way better than I (ie, I'm still not 100% sure about those lift changes, but I'm sure that's just a me problem!)

bendk:
I'm sorry, but I'm still not following this.  Can someone post details about exactly what's the difference between before and after?  I'm hoping for something like:

- Rust returns A
- The ARM ABI sign-extends it, resulting in B
- JNA does it's own sign-extension, resulting in C
- Kotlin compares C to D and succeeds/fails

My other question, is how can we know this is safe on other platforms.  It seems like this only succeeds because we're matching the quirks of ARM-32.

jkmassel:
@bendk – here's the "before this PR" flow for a signature check.

1. Rust returns `A = 0xD2FF` (16 bits, unsigned).
2. ARM ABI stores it as `0x0000D2FF` (32 bits, callee-zero-extended because the source type is unsigned).
3. JNA passes the value to ART (Android Runtime), which is supposed to narrow + sign-extend the 16-bit value into a Java int. ART’s armv7 AOT [1] path skips that step and uses the raw register value as the int. `C = 0x0000D2FF = 54015`.
4. Kotlin compares against `D = -11521` (= 54015.toShort(), precomputed at compile time). `C != D`, throws.

[1] There are [a few different ART compilation modes](https://source.android.com/docs/core/runtime/configure#compilation_options), and only the `speed` one is affected. So this might work fine for a debug build then crash in production, depending on the build settings (I hit this while trying to track down this problem).

Here's the after - the first two steps are identical:

3. Kotlin now declares the return as `Int`, so JNA reads a 32-bit value. `C = 0x0000D2FF = 54015`. Same on every platform.
4. Kotlin compares against `D = 54015` (which is now an `Int` constant, no `.toShort()` needed). `C == D`, we're good.

> how can we know this is safe on other platforms

Yeah, good question – the checksum value is a 16-bit value going in a 32 (arm32) or 64-bit (arm64) register. The ARM ABI ensures it's padded correctly on every arch. Declaring the return as Int just reads those bits as-is instead of asking JNA/ART to narrow and sign-extend. There are no ARM quirks specific to this and there's actually _less_ conversion work.

---

Hope that helps, happy to answer further questions if helpful.

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -504,6 +504,17 @@ impl KotlinCodeOracle {
...
+    fn ffi_type_label_for_direct_return(
+        &self,
+        ffi_type: &FfiType,
+        ci: &ComponentInterface,
+    ) -> String {
+        match ffi_type {
+            FfiType::UInt8 | FfiType::UInt16 => "Int".to_string(),
+            _ => self.ffi_type_label_by_value(ffi_type, ci),
+        }
+    }
```
bendk, line 519: Can you add a docstring here discussing what's happening.  Maybe just a few words plus a link to this PR and the issue would be great.

bendk (changes requested):
That's super helpful, thanks very much.  I'm still a bit hesitant to take this, since it feels like it fixes the issue in this case but I'm still not feeling totally confident it will work in other cases.  That said, we know the current behavior is buggy so why not?  Also, I'm hoping to move to JNI anyways.

I think the only thing needed is documentation.  Can you add a CHANGELOG entry in addition to the docstring?

mhammond:
I can't see what's wrong with CI here, and other later PRs aren't reproducing it. Maybe try rebasing of main and re-pushing?

lwz23:
Thanks, I followed that suggestion and refreshed the PR branch against current `main`, then pushed again.

While I was there I also addressed `bendk`'s remaining documentation request: this now has both a `CHANGELOG` entry and a short doc comment explaining why Kotlin/JNA widens unsigned `u8`/`u16` direct return carriers to `Int`.

So the current CI run should reflect the patch on top of the latest `main`, rather than the older diverged branch state.

bendk (approved):
Tests are passing and the docs look good to me.  Thanks for the patch and working it through with us!

bendk (approved):

bendk (changes requested):
Oops, it looks like this doesn't merge cleanly.  Can you try rebasing one more time?  Also, can you squash this down to 1 commit?

lwz23:
I rebased this one more time against current `main` and squashed it down to a single commit.

It should now merge cleanly: the PR is `ahead_by: 1`, `behind_by: 0`, and the new CI run is underway on that rebased single-commit branch.

bendk (approved):
Thanks for the rebase/squash.  I think we can merge now.

jkmassel:
@bendk (and/or @mhammond?) – I'm curious if y'all are planning a release that includes this fix soon? No worries if not, we're just trying to decide whether to wait for a release or pin `main` in our project.

Thanks!! 🙇

bendk:
I want to publish a new release, but there are a few loose ends to tie up.  I'm hoping we can do it by the end of the week.

bendk:
A bit late, but the process is starting now: https://github.com/mozilla/uniffi-rs/pull/2922

Issue #2675, opened 2025-09-29
Support async functions that return external types in Kotlin.


bendk:
In #2659, @joe-p pointed out that we had a regression which made Python not work.  I'm going to fix the regression in that issue, but we should also make this work on Kotlin.  I think moving to the pipeline system would make this much easier.

PR #2898, opened 2026-05-12
Fix Kotlin async external RustBuffer returns

### LINKED ISSUE #2630: Kotlin bindings do not compile for a type alias as async return value
## Issue

The Kotlin bindings fail to compile for this example

```rust
use core::Example;

uniffi::setup_scaffolding!();

// Setup the type alias
mod uniffi_setup {
    type InternalExample = core::Example;
    type ExternalExample = types::ExternalExample;
    uniffi::custom_type!(InternalExample, ExternalExample, { remote });
}

// Use the alias as the return value in an async method
#[uniffi::export]
pub async fn example() -> Example {
    Example { value: 0 }
}
```

## Minimal reproduction sample

https://github.com/antonWetzel/uniffi-remote-async

## Rust Setup

- `types` (uniffi)
    - Define the rust type `ExternalExample` with uniffi
- `core` (rust only)
    - Define the rust only type `Example`, which is used in the code
    - Define the `Into` and `From` between `Example` and `ExternalExample`
- `consumer` (uniffi)
    - Define the custom type for Kotlin `ExternalExample` and rust `Example`
    - Define uniffi method `async example() -> Example`

## Result

1. Rust compiles
1. Bindings are created
1. **ERROR:** Bindings don't compile

```
.../bindings/uniffi/consumer/consumer.kt:1256:48
Argument type mismatch: actual type is 'uniffi.consumer.RustBuffer.ByValue',
but 'uniffi.types.RustBuffer.ByValue' was expected.
```

## Why this setup

Only types and consumer crates must be uniffi, the real logic can be split into multiple rust crates

## Ideas

1. Mark a kotlin package as the base package
    - Only the base package contains the `RustBuffer`, `...` glue
    - Other packages import the glue

lwz23:
## Summary

Fix Kotlin bindings for async functions that return custom/external types whose FFI carrier is an external `RustBuffer`.

For these async return values, Kotlin's `rust_future_complete_rust_buffer` helper returns the local package's `RustBuffer.ByValue`, but the return type's lift converter can expect an external package's `RustBuffer.ByValue`. That causes the compile error reported in #2630.

This changes the Kotlin async-complete generation to inspect the return `FfiType`, not only whether the high-level return type is external. If the return carrier is `FfiType::RustBuffer(Some(external_meta))` from another crate, the complete result is converted into that external package's `RustBuffer` before it is passed to the lift converter.

## Tests

- `cargo fmt --check`
- `git diff --check`
- `CARGO_INCREMENTAL=0 cargo check -p uniffi_bindgen`
- `JAVA_HOME=/opt/homebrew/opt/openjdk/libexec/openjdk.jdk/Contents/Home PATH=/opt/homebrew/opt/openjdk/bin:$PATH CLASSPATH=/Users/liaowenzhao/Desktop/UniFFI/test-deps/jna-5.18.1.jar:/opt/homebrew/Cellar/kotlin/2.3.21/libexec/lib/kotlinx-coroutines-core-jvm.jar CARGO_NET_OFFLINE=true CARGO_INCREMENTAL=0 cargo test -p uniffi-fixture-ext-types test_imported_types_kts -- --nocapture`

Generated Kotlin now includes conversions like:

```kotlin
RustBufferGuid.create(it.capacity.toULong(), it.len.toULong(), it.data)
RustBufferOuid.create(it.capacity.toULong(), it.len.toULong(), it.data)
```

before calling the lift converters for the async return values.

Fixes #2630.
Related to #2675.

### DIFF
--- a/fixtures/ext-types/lib/src/lib.rs
+++ b/fixtures/ext-types/lib/src/lib.rs
@@ -143,6 +143,36 @@ fn get_nested_external_ouid(ouid: Option<NestedExternalOuid>) -> NestedExternalO
     ouid.unwrap_or_else(|| NestedExternalOuid(Ouid("nested-external-ouid".to_string())))
 }
 
+#[uniffi::export]
+async fn get_nested_external_ouid_async(ouid: Option<NestedExternalOuid>) -> NestedExternalOuid {
+    get_nested_external_ouid(ouid)
+}
+
+pub struct LocalExternalGuid {
+    pub value: String,
+}
+
+impl From<Guid> for LocalExternalGuid {
+    fn from(value: Guid) -> Self {
+        Self { value: value.0 }
+    }
+}
+
+impl From<LocalExternalGuid> for Guid {
+    fn from(value: LocalExternalGuid) -> Self {
+        Guid(value.value)
+    }
+}
+
+uniffi::custom_type!(LocalExternalGuid, Guid, { remote });
+
+#[uniffi::export]
+async fn get_local_external_guid_async() -> LocalExternalGuid {
+    LocalExternalGuid {
+        value: "local-external-guid".to_string(),
+    }
+}
+
 // A struct
 fn get_uniffi_one_type(t: UniffiOneType) -> UniffiOneType {
     t
--- a/fixtures/ext-types/lib/tests/bindings/test_imported_types.kts
+++ b/fixtures/ext-types/lib/tests/bindings/test_imported_types.kts
@@ -6,6 +6,8 @@ import uniffi.imported_types_lib.*
 import uniffi.imported_types_sublib.*
 import uniffi.uniffi_one_ns.*
 import uniffi.ext_types_custom.*
+import java.net.URI
+import kotlinx.coroutines.runBlocking
 
 // First step: implement a trait from an external crate in Kotlin and pass it to a function from this
 // crate.  This tests #2343 -- the codegen for this module needs to initialize the vtable from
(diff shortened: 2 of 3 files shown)


bendk:
> This changes the Kotlin async-complete generation to inspect the return `FfiType`, not only whether the high-level return type is external.

Can you explain why this fixes things?  It seems like if the FfiType is external then the high-level type would also be external.

lwz23:
Good question. I think my wording was a bit imprecise: the important part isn't that the whole `FfiType` is external, but that the `RustBuffer` carrier has external metadata.

The failing case is a local custom type whose builtin/underlying type is external. For example, in the maintainer repro:

```rust
pub struct NestedExternalOuid(pub Ouid);
uniffi::custom_newtype!(NestedExternalOuid, Ouid);
```

`NestedExternalOuid` is local to `uniffi_ext_types_lib`, so `ci.is_external(return_type)` is false. But `FfiType::from(return_type)` recurses through the custom type's builtin. The builtin is external `Ouid`, whose carrier is:

```rust
FfiType::RustBuffer(Some(ExternalFfiMetadata { name: "Ouid", module_path: "ext_types_custom::..." }))
```

That shows up in generated Kotlin as:

```kotlin
typealias NestedExternalOuid = Ouid
typealias FfiConverterTypeNestedExternalOuid = FfiConverterTypeOuid
```

so the lift converter is really the external `Ouid` converter, which expects `uniffi.ext_types_custom.RustBuffer.ByValue`.

The async complete helper, however, is declared in the local package and returns the local `uniffi.imported_types_lib.RustBuffer.ByValue`. The old `ci.is_external(return_type)` guard skipped the conversion because the high-level type was local, then Kotlin passed the local RustBuffer directly to the external lift converter.

So the contract I'm trying to encode is: for async complete, choose the RustBuffer package based on the return carrier metadata that the lift converter expects, not only on whether the high-level API type itself is external.

Direct external returns were already covered by the old guard; this patch is for local custom/alias types that project through an external RustBuffer carrier.

bendk (approved):
Wow, thanks for the explanation and the patch.  I guess it is possible for the high-level type to be local, but the FFI type to be external.

PR #2899, opened 2026-05-16
Allow async_runtime on trait exports

JssDWt:
Accept #[uniffi::export(async_runtime = "tokio")] on trait exports. Trait method futures now get the same async_compat::Compat wrap that inherent impls and free functions already use to enter a tokio runtime context on every poll.

### DIFF
--- a/fixtures/futures/src/lib.rs
+++ b/fixtures/futures/src/lib.rs
@@ -432,6 +432,32 @@ fn get_say_after_udl_traits() -> Vec<Arc<dyn SayAfterUdlTrait>> {
     vec![Arc::new(SayAfterImpl1), Arc::new(SayAfterImpl2)]
 }
 
+// Example of an async trait whose Rust implementation uses tokio resources.
+#[uniffi::export(with_foreign, async_runtime = "tokio")]
+#[async_trait::async_trait]
+pub trait SayAfterTraitTokio: Send + Sync {
+    async fn say_after(&self, ms: u16, who: String) -> String;
+}
+
+#[async_trait::async_trait]
+impl SayAfterTraitTokio for SayAfterImpl1 {
+    async fn say_after(&self, ms: u16, who: String) -> String {
+        say_after_with_tokio(ms, who).await
+    }
+}
+
+#[async_trait::async_trait]
+impl SayAfterTraitTokio for SayAfterImpl2 {
+    async fn say_after(&self, ms: u16, who: String) -> String {
+        say_after_with_tokio(ms, who).await
+    }
+}
+
+#[uniffi::export]
+fn get_say_after_tokio_traits() -> Vec<Arc<dyn SayAfterTraitTokio>> {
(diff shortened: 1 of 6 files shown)


bendk:
I had a bit of a hard time following the commit message.  It correctly describes how async functions work, but it didn't help me understand what this patch was aiming to do.  For me, the most relevant part was the last paragraph:

> Honor args.async_runtime in trait scaffolding the same way inherent impls do. The change is opt-in: trait exports without the attribute behave as before. Foreign implementations of the trait are unaffected because they dispatch through callback_interface::foreign_async_call, which is not part of this path.

How about putting that one first?  In fact, you could shorten it to just that paragraph IMO.

fixtures/futures/src/lib.rs
```diff
@@ -432,6 +432,32 @@ fn get_say_after_udl_traits() -> Vec<Arc<dyn SayAfterUdlTrait>> {
...
+    async fn say_after(&self, ms: u16, who: String) -> String {
+        say_after_with_tokio(ms, who).await
+    }
+}
+
+#[uniffi::export]
+fn get_say_after_tokio_traits() -> Vec<Arc<dyn SayAfterTraitTokio>> {
+    vec![Arc::new(SayAfterImpl1), Arc::new(SayAfterImpl2)]
+}
+
```
bendk, line 460: Side note:  I was going to suggest moving this to the [bindgen-tests](https://github.com/mozilla/uniffi-rs/blob/main/bindgen-tests/lib/src/futures.rs), since that's generally where new fixtures should go.  However, I realize this isn't really testing the "bindgens", it's testing the Rust scaffolding.  I think this is as good a place as any for these types of tests, but maybe we should eventually make a `scaffolding-tests` crate that's the mirror of `bindgen-tests` in that it tries to directly test the scaffolding code, without any attempt to make it look like a real-world example.  One difference would be that it wouldn't need to test multiple bindings languages, it would just need one.  Or maybe zero and it could just test the FFI code directly from Rust.

bendk (changes requested):
The code looks great to me.  How do yo feel about my commit message suggestion?

Also, please add a note to the CHANGELOG mentioning this feature.

JssDWt:
Reworked the commit message per your suggestion, dropped `with_foreign` from the title (the patch applies to all trait exports), and added a CHANGELOG entry. Thanks for the review!

> I had a bit of a hard time following the commit message

Sorry about that. I had some help from Claude drafting this PR. I had been fightning Claude about the commit message already, but you're right it was still not understandable.

fixtures/futures/src/lib.rs
JssDWt: For this PR specifically, you think I should move it into `scaffolding-tests` now?

bendk: No, I was just thinking outloud.  We don't have a scaffolding tests yet, so using this fixture seems good to me.

bendk (approved):
Thanks for the update, this is looking great!

Issue #2801, opened 2026-01-21
Add built-in support for HashSet


bes:
Hello!

I just discovered that there is [no out-of-the-box support](https://mozilla.github.io/uniffi-rs/latest/types/builtin_types.html) for HashSet in uniffi.

According to [this matrix thread](https://matrix.to/#/!tsNDDGHvNGVNmvOOzi:mozilla.org/$-PEvrJNc4k8N7Ka-A1l3NzI6_yta8BzBiPocbMHWa0E?via=mozilla.org&via=matrix.org&via=element.io) it might be because of UDL, but that it could be supported by macros.

Would it be possible to add? Thank you!

(Edit: Removed the "Please" from the title to make it a bit more readable, so I will write it here instead: Please add this feature ❤️)

bes:
We are really sorely missing this feature; can I help out in any way to get it done?

mhammond:
we'd take a good patch 😄

bes:
@mhammond I am working on a PR, but I am having some troubles :)

* `rfkelly/uniffi-ci:latest` is really really old
* `docker/Dockerfile-build` is outdated, the swift toolchain file does not exist anymore
    * Updated to `cimg/rust:1.90.0`
    * Updated to swift 6.3 (static) but get a bunch of test failures anyway
```
failures:

---- uniffi_foreign_language_testcase_test_swift stdout ----
Creating testing out_dir: /mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6
Error: running `swiftc` to compile bindings failed (cd "/mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6" && "swiftc" "-emit-module" "-module-name" "unary_result_alias" "-o" "libtestmod_unary_result_alias.so" "-emit-library" "-swift-version" "5" "-Xcc" "-fmodule-map-file=/mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6/unary_result_aliasFFI.modulemap" "-I" "/mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6" "-L" "/mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6" "-luniffi_unary_result_alias" "/mounted_workdir/target/tmp/uniffi_unary_result_alias-81c252357d8743a6/unary_result_alias.swift")
```

* Running tests locally on macOS is also challenging
[...]

PR #2901, opened 2026-05-17
Add support for HashSet in uniffi proc macros

bes:
Solves https://github.com/mozilla/uniffi-rs/issues/2801

### DIFF
--- a/uniffi_bindgen/src/bindings/ruby/templates/wrapper.rb
+++ b/uniffi_bindgen/src/bindings/ruby/templates/wrapper.rb
@@ -14,6 +14,7 @@
 # helpers directly inline like we're doing here.
 
 require 'ffi'
+require 'set'
 
 
 module {{ ci.namespace()|class_name_rb }}
--- a/uniffi_bindgen/src/interface/mod.rs
+++ b/uniffi_bindgen/src/interface/mod.rs
@@ -439,6 +439,13 @@ impl ComponentInterface {
             .any(|t| matches!(t, Type::Map { .. }))
     }
 
+    /// Check whether the interface contains any set types
+    pub fn contains_set_types(&self) -> bool {
+        self.types
+            .iter_local_types()
+            .any(|t| matches!(t, Type::Set { .. }))
+    }
+
     /// Check whether the interface contains any object types
     pub fn contains_object_types(&self) -> bool {
         self.types
@@ -1419,7 +1426,8 @@ fn type_names_in_type(ty: &Type) -> Vec<String> {
         Type::Enum { name, .. } | Type::Record { name, .. } => vec![name.clone()],
         Type::Box { inner_type }
         | Type::Optional { inner_type }
-        | Type::Sequence { inner_type } => type_names_in_type(inner_type),
+        | Type::Sequence { inner_type }
+        | Type::Set { inner_type } => type_names_in_type(inner_type),
         Type::Map {
             key_type,
             value_type,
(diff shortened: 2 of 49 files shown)


uniffi_bindgen/src/bindings/ruby/templates/wrapper.rb
```diff
@@ -14,6 +14,7 @@
 # helpers directly inline like we're doing here.
 
 require 'ffi'
+require 'set'
```
bes, line 17: Unsure about this one, not a ruby expert. Maybe it could be improved somehow?

mhammond: @saks do you have any thoughts here?

mhammond (commented):
Thanks! This seems very thorough and looks great to me. It does however needs docs and a CHANGELOG.

mhammond:
Some of your new tests are failing though...

bes:
I hade trouble running all the tests in the new docker image locally. As I wrote, I'm not a ruby (or python, or kotlin) expert; I mainly work with Rust, wasm, and Swift.

With a little help of some search engines I managed to figure out what was wrong. Pushing fixes shortly.

A problem I have is that I can't run all the tests locally, it fails after (a long) while with this

```
failures:

---- uniffi_foreign_language_testcase_test_futures_swift stdout ----
Creating testing out_dir: /mounted_workdir/target/tmp/uniffi_futures-33456a6570cdb80a
Error: running `swift` to run test script failed (cd "/mounted_workdir/target/tmp/uniffi_futures-33456a6570cdb80a" && "swift" "-I" "/mounted_workdir/target/tmp/uniffi_futures-33456a6570cdb80a" "-L" "/mounted_workdir/target/tmp/uniffi_futures-33456a6570cdb80a" "-ltestmod_futures" "-luniffi_futures" "-swift-version" "5" "-Xcc" "-fmodule-map-file=/mounted_workdir/target/tmp/uniffi_futures-33456a6570cdb80a/futuresFFI.modulemap" "/mounted_workdir/fixtures/futures/tests/bindings/test_futures.swift")
```

Anyway...

I have also added a line to the changelog and tried to poke at the documentation, although there I am unsure if I did too little or too much, happy to take some direction. My main concern is that I don't want to give the impression that this is supported in UDL, but I also did not want to touch the UDL documentation just to add a random "HashSet is not supported" thingy.

mhammond:
It looks like you might need to run `TRYBUILD=overwrite cargo test -p uniffi_uitests` to get CI happy

bes:
Thanks. I finally got tests to work on macOS (had to do some changes to `bindgen-tests/swift/src/lib.rs` - not included in this PR). Ran the command you suggested.

uniffi_bindgen/src/interface/mod.rs
```diff
@@ -439,6 +439,13 @@ impl ComponentInterface {
             .any(|t| matches!(t, Type::Map { .. }))
     }
 
+    /// Check whether the interface contains any set types
```
mhammond, line 442: This is unused - but apparently so are many of the existing ones, so this is fine, but I do wonder if we should remove the others?

mhammond (approved):
This is great, thanks!

uniffi_bindgen/src/bindings/ruby/templates/wrapper.rb
saks: Thanks for adding the ruby part. Let me look into it 👍

PR #2902, opened 2026-05-17
Allow foreign-only trait implementations

mhammond:
Add `#[uniffi::export(foreign)]` for traits where only foreign code crosses the FFI boundary, and `#[uniffi::export(rust, foreign)]` as the canonical replacement for the now deprecated `with_foreign flag`. Internally, `ObjectImpl` is restructured to carry a `TraitKind` discriminant.

Gets callback interfaces even closer to being less of a special case too - but that can be a followup.

@bendk as we kinda discussed in slack. After going around quite a few circles, I think this syntax seems good - but entirely open to others. No hurry at all here - wdyt?


uniffi_bindgen/src/bindings/python/pipeline/interfaces.rs
```diff
@@ -18,7 +19,7 @@ pub fn protocol(int: &general::Interface, context: &Context) -> Result<Protocol>
                 docstring: int.docstring.clone(),
             }
         }
-        ObjectImpl::CallbackTrait => {
+        ObjectImpl::Trait(TraitKind::Both | TraitKind::ForeignOnly) => {
             // Trait interface that can be implemented in Rust or Python
```
bendk:
```suggestion
            // Trait interface that can be implemented in Python
```

uniffi_meta/src/types.rs
```diff
@@ -44,18 +54,37 @@ impl ObjectImpl {
...
 
     pub fn has_struct(&self) -> bool {
         matches!(self, Self::Struct)
     }
 }
 
+// TraitKind passes through the pipeline unchanged (it's a plain enum with no node fields)
+impl<C> MapNode<TraitKind, C> for TraitKind {
+    fn map_node(self, _context: &C) -> Result<TraitKind> {
+        Ok(self)
```
bendk: I think you can avoid this impl with `#[derive(MapNode)]` on the definition and `uniffi_pipeline::use_prev_node(TraitKind)` in the pipeline module.  I kind-of remember writing code like this and deciding against it.  IIRC, the error messages are clearer with `use_prev_node`.  But I could just be misremembering.

bendk (commented):
Code looks good and the feature makes sense to me.  I'm interested to hear what others think, but I'm a +1 on this.

bendk (approved):
This still looks good to me, I think it's ready to land (after a rebase).

PR #2905, opened 2026-05-19
Adding more optional/vec tests

bendk:
I'm planning on special casing these in for uniffi-bindgen-kotlin-jni, and it seems likely that other languages will want to do the same. Let's make sure we're testing as many different cases here.


mhammond (approved):
lgtm, no idea what's going on with that CI failure - the other PR which saw that was fixed by a rebase, and your branch does look a little old, so maybe that will fix it?

bendk:
Looks like the rebased fixed things, let's merge.

PR #2908, opened 2026-05-24
Improve the global configuration docs

mhammond:
In particular, create a top-level `configuration` docs endpoint.


mkdocs.yml
```diff
@@ -49,6 +49,7 @@ nav:
   - ./types/builtin_types.md
   - ./types/enumerations.md
   - ./types/records.md
+  - ./types/bytes.md
```
badboy, line 52: Oh, so this never showed up? Oops I guess.

badboy (approved):

PR #2909, opened 2026-05-26
Document that constructors on enums and records aren't supported.

### LINKED ISSUE #2907: Unable to generate bindings if Enum or Record type has associated #[uniffi::constructor]
Thanks to https://github.com/mozilla/uniffi-rs/pull/2706, we capture the metadata needed for methods and associated constructor functions on `Record` and `Enum` types, and thanks to https://github.com/mozilla/uniffi-rs/pull/2724, we use the captured metadata to generate bindings for methods and constructors on `Record` and `Enum` types. 

However, while the methods work correctly, the constructors cause the bindgen to error out with a misleading message (as of `uniffi@92c213de`).

With the following simple example (exported method on `Record` type), the bindgen works perfectly:

```rust
uniffi::setup_scaffolding!();

#[derive(uniffi::Record)]
pub struct Foo {
    pub bar: String,
}

#[uniffi::export]
impl Foo {
    pub fn get_bar(&self) -> String {
        self.bar.clone()
    }
}
```

Similarly, we can export a free function that constructs the `Record`; the bindgen still works perfectly:

```rust
// ... as above ...

#[uniffi::export]
pub fn new_foo(bar: String) -> Foo {
    Foo { bar }
}
```

But as soon as we add an exported constructor, the bindgen fails:

```rust
// ... as above ...

#[uniffi::export]
impl Foo {
    #[uniffi::constructor]
    pub fn new(bar: String) -> Self {
        Self { bar }
    }
}
```

The error message is misleading (the return type *is* `Self`, and `Arc<Self>` doesn't compile):

```
$ cargo build && cargo run --bin uniffi-bindgen generate ...
extracting metadata for '_UNIFFI_META_UNIFFI_TEST_CONSTRUCTOR_FOO_NEW'

Caused by:
    Constructor return type must be Self or Arc<Self>
```

Perhaps this is simply not supported yet, but the fact that https://github.com/mozilla/uniffi-rs/pull/2724 explicitly added *both* methods and constructors to `derive_ffi_funcs` makes me wonder if this is a bug/oversight:

```diff
diff --git a/uniffi_bindgen/src/interface/record.rs b/uniffi_bindgen/src/interface/record.rs
index 30f652b66..f368a9961 100644
--- a/uniffi_bindgen/src/interface/record.rs
+++ b/uniffi_bindgen/src/interface/record.rs
@@ -123,6 +123,12 @@ impl Record {
     }
 
     pub fn derive_ffi_funcs(&mut self) -> Result<()> {
+        for c in self.constructors.iter_mut() {
+            c.derive_ffi_func();
+        }
+        for m in self.methods.iter_mut() {
+            m.derive_ffi_func()?;
+        }
         for ut in self.uniffi_traits.iter_mut() {
             ut.derive_ffi_func()?;
         }
```

mhammond:
Also gives a better error message if you try, and adds bindgen tests for regular methods on them.

Fixes #2907

This made me remember why we don't support it - it's tricky because in many languages the normal way to create records and enums is either via a contructor (eg, kotlin), or not compatible with enums (eg, python). This isn't to say we couldn't do something better in the future, but we might as well document the status quo.

### DIFF
--- a/bindgen-tests/kotlin/tests/enums.kts
+++ b/bindgen-tests/kotlin/tests/enums.kts
@@ -31,3 +31,5 @@ assert(ExplicitValuedEnum.THIRTEENTH.value == 13.toUByte())
 assert(GappedEnum.ONE.value == 10.toUByte())
 assert(GappedEnum.TWO.value == 11.toUByte()) // Sequential value after ONE (10+1)
 assert(GappedEnum.THREE.value == 14.toUByte()) // Explicit value again
+// Enum methods
+assert(EnumWithData.A(20.toUByte(), 40.toUShort()).roundtrip() == EnumWithData.A(20.toUByte(), 40.toUShort()))
--- a/bindgen-tests/kotlin/tests/records.kts
+++ b/bindgen-tests/kotlin/tests/records.kts
@@ -33,3 +33,5 @@ assert(
     fieldRec=SimpleRec(a=42.toUByte())
   )
 )
+// methods
+assert(SimpleRec(a=42.toUByte()).roundtrip() == SimpleRec(a=42.toUByte()))
--- a/bindgen-tests/lib/src/enums.rs
+++ b/bindgen-tests/lib/src/enums.rs
@@ -11,11 +11,18 @@ pub enum EnumNoData {
     C,
 }
 
-#[derive(uniffi::Enum)]
+#[derive(uniffi::Enum, Clone)]
 pub enum EnumWithData {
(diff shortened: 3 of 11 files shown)


bendk (approved):
Looks good to me.

Are enum methods only supported on python at this point?

mhammond:
> Are enum methods only supported on python at this point?

Nope, supported in Swift and Kotlin too, I was just lazy. I just added the same tests for them too.

PR #2911, opened 2026-06-02
uniffi-bindgen-kotlin-jni

bendk:
# Overview and Status

This PR implements a new Kotlin bindgen based on JNI instead of JNA and using a new FFI instead of the existing scaffolding code.  There's two main benefits to this:

* **Stability**.  By avoiding JNA, we avoid the bugs that come from that library like [#2624](https://github.com/mozilla/uniffi-rs/issues/2624) and [#2740](https://github.com/mozilla/uniffi-rs/issues/2740). In fairness to JNA, our usage is not at all typical, so it's not surprising that we're triggering bugs.  For Firefox Android, we've seen much less crashes while testing out the [Kotlin pointer FFI branch](https://github.com/mozilla/uniffi-rs/pull/2765).  That branch still uses JNA, but tries to avoid the areas we suspected bugs were coming from.  I think this is an improved version of that one.  Of course, introducing a bunch of new code means we're likely introducing some new bugs too, but they should be more tractable.  There's a lot of magic happening behind the scenes to for JNA to present it's simplified interface and it's very hard to debug bugs in that layer.
* **Performance**.  This runs significantly faster than the current Kotlin bindings.  Part of that is JNI, but it's more because the FFI has been reworked from scratch with an eye on performance and a benchmark suite measure different approaches. Our benchmark suite shows [significant gains compared to the current Kotlin bindings](https://gist.github.com/bendk/7bc8886cb33b8bd5d257336dc33d54a5) with many benchmarks improving by over 100x. This code also [compares well against BoltFFI 0.24.1](https://gist.github.com/bendk/5054d1477361125ef430e2ccabfd5aed). The table there comes from their benchmarks and a negative speedup means that the UniFFI generated code ran faster. We're faster than BoltFFI in many of their benchmarks, I think their main advantage at this point is the primitive array optimization that I discuss below.  (BoltFFI development is speeding along and already on 0.25.2, but their benchmarks require patching to work with the new bindgen and I didn't have time to redo that).

I think this code is complete enough to loose the "WIP" label, but it's not ready to be merged yet either since it's still dependant on other PRs like #2841

This is implemented as a separate bindgen so we could merge this, call it "experimental" and continue to support `uniffi-bindgen -l kotlin` until we're ready to tell users to move to `uniffi-bindgen-kotlin-jni`.  Implementing it as a separate crate also means it's a POC that external bindings generators can also generate language-specific Rust scaffolding.

# The new FFI

The new FFI was designed from scratch.  It doesn't use the existing scaffolding functions or the FFI traits like `Lift` and `Lower`.  I'll discuss the main parts of the FFI here, there's more details in the `DESIGN.md` doc.

## Primitives

We can pass integers, floats, booleans, strings, and arrays of these types directly via JNI.

I kept the "lift" and "lower" terminology for this.  `String` lowers to a `jni_sys::string` and `u32` values are lifted from an `i32`.  Like the current scaffolding, `jni_sys::string` and `i32` are the "FFI types".  The main difference is that high-level types can lower to multiple FFI types. This is how we handle records, enums and options.

## Records/enums/options

These types are deconstructed into multiple FFI types when passed across the FFI.  For records this usually means passing each field as a separate FFI value.

When a high-level type is an argument we can simply pass that as multiple arguments to the scaffolding function.  Returning values is trickier, since JNI only supports 1 return value. For Kotlin -> Rust sync calls, we handle it by having Rust call the Kotlin lift function via JNI, passing the FFI values.  Rust gets a `jobject` back and returns that the JNI function. Here's some [example code](https://gist.github.com/bendk/5b1120bb0488ec954c46d142a8ca75f2) that shows all of this in action.  See DESIGN.md for more details.

Enums are also deconstructed into multiple FFI types, however a given variant may only use a subset of those values.  The others are passed using default values and ignored by the other side of the FFI ([example code](https://gist.github.com/bendk/ab015cab45116790b51705dd243ff0aa)). One potential issue with this is that it can theoretically result in a very large number of arguments for a single enum if the variants have different types.  My guess is this won't be an issue in practice, but if it does we can take steps to handle it.

Finally, options are deconstructed into FFI types.  In general, we can treat them like enums and pass them as a boolean, plus FFI types for the inner type.  However, we can also special-case them if we can find a niche value to use as `None`.  Strings can use `null`, floats can use a NaN value, and ints smaller than 64-bits can be casted to an `i64` and use `i64::MAX`.

When one of these types lowers to a single FFI type (for example Option<String> or a record with 1 field), then we say the type "lowers to a primitive" and it's passed across the FFI just like a primitive (i.e. the JNI function can return that primitive directly rather than making the JNI
call described above).

Overall, I think this is a great way to pass values across the FFI and provides some of the largest speedups compared to the current FFI.  Unfortunately we can't do this for dynamically-sized types.  Let's discuss those now.

## Vecs/HashMaps/buffers

I tested a bunch of ways to pass these across the FFI and ended up with allocating a `java.nio.ByteBuffer` for each vec/hashmap/dynamically sized value and passing that byte buffer across the FFI.

I tried to only allocate a single buffer for the entire call as discussed in #2816. However, that created a lot of complexity in the bindings generator and was slower for large buffers since it meant we needed a lot more bounds/alignment checking.  For example, my initial attempted called [this function](https://github.com/bendk/uniffi-rs/blob/178d8011ae0afc4ba3b8bacebb52f71d8af2a3fa/uniffi-bindgen-kotlin-jni/src/bindings/templates/shared/FfiBuffer.kt#L24-L37) for each value packed in the buffer.  Those checks weren't getting optimized out by the JVM and they added up when you had a large buffer.  I think we could avoid the alignment checks in some cases, but that would introduce more complexity into the bindings and we'd still need the bounds checks. Everything becomes simpler and faster if we allocate a separate buffer for each vec ([example](https://gist.github.com/bendk/fcefee5e1d3460481f7463d16f1f3651)).

I also tried to calculate the sizes ahead of time to avoid the bounds-check, which worked until I got to custom types.  Custom types need to be lowered first before you can calculate their size, but then where do you store the lowered type in Rust?  There's not really a way to do this, especially when you consider things like a vec of custom types.  BTW, I realized the current [Kotlin bindings](https://github.com/mozilla/uniffi-rs/blob/f5d73a60e119798fb35b019f7699a56af02b3061/uniffi_bindgen/src/bindings/kotlin/templates/CustomTypeTemplate.kt#L48-L50) cheat by lowering the type, calculating the size, throwing away the lowered value and
lowering again later on.

### Primitive arrays

Types like `Vec<i32>` are optimized by using `IntArray` which faster to pass than `java.nio.ByteBuffer`.  We could speed this up even more if the generated Kotlin functions used `IntArray` instead of `List<Int>`.  The Array <> List conversions represented a majority of time for the total operation. Maybe this is something we could look into in the future, possibly behind a config option.  I didn't think the speedup was work the breaking change at this point though.

### Small buffers

One issue with the current code is that it's not ideal for small buffers.  This is because the JNI `NewDirectByteBuffer` call needed to create the object is relatively slow.  When I measured it was something like 5x slower than making a Kotlin -> Rust JNI call.  This means that for small buffers it's faster to do everything in Rust: Rust allocates the buffer, returns the data pointer casted to a `Long`, then Kotlin makes a JNI call to fetch each value.  This approach doesn't scale well, since the JNI calls to fetch each value add up, but it was the fastest for small buffers in my testing.

Maybe we could address this somehow down the road.  One idea I had would be to keep an already allocated and constructed `java.nio.ByteBuffer` around in thread-local data.  If the data fit in that buffer we could use that and avoid the cost of constructing a new JVM object. However, I think the current system is fine for now and it's more important to optimize large buffers than small buffers.

### Future direction: one-copy buffers

One thing I really like about this approach is that it brings us closer to the ultimate solution for vecs:  calling `Vec::into_raw_parts` and creating the NIO buffer directly from that without any copying.  This is not exactly zero-copy, since we'd still copy the buffer data into the Kotlin data class, but a single copy seems optimal for non-primitive vecs.

The only reason we can't do that is that we don't know the layout of the types inside the buffer and knowing that layout is hard because Rust  doesn't guarantee a stable type layout.  BoltFFI handles this by adding `repr(C)` to types, but this isn't a very general solution since you can't add `repr(C)` to `Option<>` or remote types like `log::Level`.

The solution I'd like to try at some point is creating a one-time initialization process where Rust calculates the field offsets, passes them to Kotlin, then Kotlin uses those offsets instead of hard-coded values.  Since this is all happening inside the running process, the offsets won't change. The [field_offset](https://docs.rs/field-offset/latest/field_offset/) macro will be great for this, but unfortunately it doesn't handle enum fields yet.  Maybe we'd need to wait for that feature to be stabilized or maybe we could get our hands dirty and measure it ourselves with some pointer arithmetic.

## Exceptions

Instead of mutating a `RustCallStatus` struct to track errors/exceptions, the new code throws exceptions directly via JNI.  The Rust code constructs the exception object, similar to how it constructs a record object, then causes the current method to throw.  This seems simpler overall.

## Special-casing certain arguments

The `ArgFfi::Custom` type is used to special case the FFI for certain arguments rather than use the normal FFI for that type.  I originally added this to optimize object references in general and receiver arguments in particular.  However, it also worked for the zero-copy `&[u8]` strategy that was introduced after I started this work.  In general, this seems like a decent general purpose way to optimize certain arguments.  OTOH, maybe it wouldn't be needed if `uniffi_meta::Type` had a more variants like `Ref` and `Slice`.

# Code sharing

If we create other language-specific bindgens, will each one need to write everything from scratch or can we share some of the code? Here's my current view, take it with a grain of salt.  I think we'll learn a lot more once we actually try to generalize things.

The FFI buffer layout code should be fairly easy to share.  We should also be able to share the code that maps high-level types to FFI types.  One question I have there is if the [standard FFI type mapping](https://github.com/bendk/uniffi-rs/blob/a01603f1af087029b62619834b6863654af49f41/uniffi-bindgen-kotlin-jni/src/pipeline/ffi_types.rs#L7-L67) would be the same for all languages.  Maybe it would be or maybe we'd have to make the general pipeline somehow generic over FfiType enum plus a trait that includes the `standard_ffi_type_mapping` function.  I'm pretty sure we could share some template code as well, although there are some issues that we'd need to figure out there as well.

The JNI code that exports Rust functions to Kotlin and calls Kotlin functions from Rust definitely can't be shared.

If you combine that all together, you can sort-of picture a future where the type lowering/lifting/reading/writing logic is shared, the Rust code to do that is mostly shared, and the job of the bindings generator is:

* Implementing the FFI calls and wiring up the JNI/pyo3/whatever code to the lift/lowering logic.
* Generating the type lift/lower/read/write functions in the foreign language
* Generating the type definitions in the foreign languages (e.g. defining data classes for records).

## uniffi-bindgen-java

For this case, we could actually share the Rust JNI code and we would only need to generate Java bindings instead of Kotlin ones.  Overall this seems very straight-forward to me, but some parts, like async, will probably require changes.

# uniffi_parse_rs and the generated code layout

This is based on using `uniffi_parse_rs` to parse the interface, run it through the pipeline, then generate the Rust JNI code from that rather than  generating the Rust code in the macros. The main reason is that I couldn't think of a good way for external bindings to hook into the macro code and I wanted them to be able to do the same thing if they wanted.  I think the pipeline made the implementation easier overall.

The generated code is structured a bit differently than our current system:
  * Single Rust module for all crates
  * Single Kotlin with shared FFI code
  * Kotlin module for each Rust crate that defines the public interface

This provides a several advantages: There's less code duplication with a shared Kotlin module (#1896). Issues with external types pretty much go away since all the FFI code is in one module. Finally, it feels cleaner to separate the FFI code from the exported interface.  For example, we're currently exporting our helper code in the user's generated modules (#1425).  With the new system, the helper code is still public, but now it's contained in the `uniffi.Uniffi` module which users don't import.

# Smaller changes that we might want to port back to the general pipeline

## ids
All types have an `id` field, which lets us uniquely identify them and avoid collisions when generating functions (e.g. `lift_type_{id}`).  I like this better than `canonical_name` since it's guaranteed to be unique.  With canonical_name I always worry about issues when we mash the strings together and I was worried about it even more since we're combining FFI functions from multiple source crates into a single generated module.  For example, if the user tries to export both `foo::bar_baz()` and `foo_bar::baz()` will the generated function names be unique?

## CallableResult

I added the `CallableResult` type, which is a return and throws type together.  This was useful when I needed to generate a function for each combination, for example to complete an async call.  We do something very similar for JS as well.

## Less reliance on traits

`uniffi-bindgen-kotlin-jni` relies a lot less on traits like `Lift` and `Lower`.  Instead it generates functions and knows how to reference those function names.  This gives some extra flexibility, for example lift functions can now input any number of arguments which is fine since they don't have to implement any trait.  When generate code calls a lift function it needs to make sure it also generates the right number of arguments, which was totally fine in practice.


uniffi/Cargo.toml
```diff
@@ -49,6 +49,12 @@ bindgen-tests = [ "dep:uniffi_bindgen", "uniffi_bindgen?/bindgen-tests" ]
 # Enable support for Tokio's futures.
 # This must still be opted into on a per-function basis using `#[uniffi::export(async_runtime = "tokio")]`.
 tokio = ["uniffi_core/tokio"]
+# Generate scaffolding in the UniFFI macros
+#
+# This is a default feature that should always be enabled,
```
mhammond: nit, this should be softened and less scary - eg, "except for experimental bindgens like `uniffi-bindgen-kotlin-jni` which generate the necessary scaffolding"...

uniffi_bindgen/src/bindings/python/test.rs
mhammond, line 47: here and elsewhere, why not `Default::default`?

mhammond (commented):
This looks good to me - it needs a rebase as there's a lot from other PRs remaining, but seeing it's mostly an entire new crate the risk seems very low.

uniffi_bindgen/src/bindings/python/test.rs
bendk: Not sure really, I guess I like spelling out the type to make it what we're passing in more explicit.  I don't have strong feelings either way though.

mhammond (approved):
This is purely additive, and an amazing amount of work offering (hopefully 😅 ) great performance improvements. I deleted my minor "nit" comments as they don't add anything to this - thanks!

bendk:
I just merged the uniffi_parse_rs PR, next up is this one once I get CI working again.  There's certainly bugs to be fixed and details to be worked out, but the risk isn't so high since this is a totally new bindings generator (`uniffi-bindgen -l kotlin` will continue to work the same).

PR #2912, opened 2026-06-03
More bindgen tests updates

bendk:
These were based my experience implementing uniffi-bindgen-kotlin-jni:

* Move options test into it's own fixture, I wanted to test those separately from the vec/hashmap tests.  Renamed the `compound_types` fixture to `collections`.
* Added test for optional interfaces
* Rust-only trait interfaces
* Test different arg/return types with trait interfaces
* Added test for errors with no fields (but also not flat)

I had claude help me generate a lot of these, it seemed like a good fit since these were mostly cut and paste.


mhammond (approved):
bit of a rubber stamp, but lgtm. I wonder if we can ask claude to identify fixtures which are becoming entirely redundant now?

bendk:
> I wonder if we can ask claude to identify fixtures which are becoming entirely redundant now?

Interesting, it's worth a shot.  Maybe I'll try that out next week.

PR #2915, opened 2026-06-08
Pipeline IDs

bendk:
Add `Callable.id` and `Type.id`.  These are useful when you need a unique ID for one of these, for example because you want to populate a hash map or generate a unique function name.

I prefer using a unique ID over canonical_name for that because it avoids name collisions in weird corner cases.  For example `HashMap<Foo, TypeBar>` and `HashMap<FooType, Bar>` both map to the same canonical name.  Also, the ID is unique over all crates.  `canonical_name` would also need to incorporate the namespace to make that true, which would increase the number of corner cases.

We're generating these in the Gecko JS pipeline and I also used them for uniffi-bindgen-kotlin-jni and I think other languages will want them as well.


bendk:
This got weird because I changed my jj push bookmark template.  Closing this in favor of https://github.com/mozilla/uniffi-rs/pull/2919

PR #2916, opened 2026-06-09
Add sync callbacks support for Ruby

saks:
Hello Uniffi Team!

The main thing that this PR tries to accomplish is: **add support for sync callbacks and trait interfaces for ruby**. No async calls for now. PR adding both will be too large. An additional goal was to keep improving `bindgen-tests` for ruby. 

I also found that:
- few tests needed cleanup (copy-paste issues from my previous PR)
- ruby classes for enums and records couldn't have methods
- HashMap can only have string keys
- there were lots of places in code where class/method names were assumed to match expectations from another file

Since I was discovering (and addressing) these while implementing callbacks and trait interfaces, I didn't split changes into multiple PRs, but I will if PR volume becomes an issue for code review (please let me know).

I've also added tests into "examples" and "fixtures", please let me know if any of those are undesirable because of introduction of `bindgen-tests`(which I like much more TBH :) ).

I've tested these changes on my projects using `uniffi` and I cannot spot any issues so far.
Please let me know if you have any suggestions or concerns.
Thank you!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 - Updated `askama` version to `0.15.6`
 - Custom Types can have docstrings in some languages ([#2853](https://github.com/mozilla/uniffi-rs/pull/2853))
 - Ruby: Expose standard Rust traits for generated ruby code (#2883)
+- Ruby: Add support for sync foreign traits ([#2916](https://github.com/mozilla/uniffi-rs/pull/2916))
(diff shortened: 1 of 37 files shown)


bendk (approved):
This looks great to me.  I didn't review in that much detail, since it seems very similar to the other implementations.  Is there anything in particular that you think is risky?  If not, I guess we can merge this.

saks:
> This looks great to me. I didn't review in that much detail, since it seems very similar to the other implementations. Is there anything in particular that you think is risky? If not, I guess we can merge this.

Thank you for having a look! I can't think of anything that poses high risks here, please feel free to merge it when you get a chance 👍

PR #2923, opened 2026-06-16
Add async support for ruby

### LINKED ISSUE #1871: [Async] functions in Ruby
Async functions are available for the kotlin and swift and python bindings but the support is lacking for ruby. Is there any plans to work on this?

saks:
Hello UniFFI team!

This PR focuses on support of async calls for ruby bindings (fixes #1871) for objects, top level functions and callback interfaces.

#### Foreign async callbacks are synchronous blocking calls in a thread
I decided to run foreign trait methods synchronously in `Thread.new`. This matches what Ruby can do without `async def`, but it differs from Python/Kotlin/Swift where foreign methods are truly async. Implementers must block until done. True async Ruby (e.g. `async` gem) would need internal blocking/waiting (I really wanted to keep dependencies minimal, at least for now).

#### Rust-side polling
The pipe + timed `wait_readable` approach seems reasonable MRI-specific way to release the `GVL`(Global VM Lock) and integrate with `Fiber::Scheduler#io_wait`. It allows for parallel execution in threads or using `Fiber::Scheduler`.

#### Deadlock?
While implementing `fixtures/futures/tests/bindings/test_futures.rb`, I've noticed that tests would randomly get stuck. That rabbit hole got me into looking into how "futures", "scheduler" and "oneshot" work. 

I'm pretty sure that one (if not the main) reason is how Ruby is interfacing with Rust (because Python tests are fine). It seems like a common approach to release locks as soon as possible. Does this align with the intended design, or is there a historical reason the lock is held during the callback?  (FYI: `scheduler.rs` has more methods that lock during the callback). Would be really nice if someone can look at it closely.

Without these changes (in `uniffi_core`), I can reliably reproduce this problemt locally by:
```sh
cd fixtures/futures
for i in $(seq 1 100); do echo "=== Run $i ==="; timeout 30 cargo test _rb --features uniffi/ffi-trace -- --nocapture > /tmp/futures_test_$i.txt 2>&1; ret=$?; if [[ $ret -ne 0 ]]; then echo "FAILED (exit $ret) - output in /tmp/futures_test_$i.txt"; tail -40 /tmp/futures_test_$i.txt; break; else echo "PASSED"; fi; done
```
I'm attaching one of the files with debug messages, just in case.
[futures_test_8.txt](https://github.com/user-attachments/files/28861049/futures_test_8.txt)

#### Async methods for records/enums
I was going to also add support for [async methods for records and enums](https://github.com/saks/uniffi-rs/pull/4/changes/769058b3fa12af64bdd72900edcc52745ef98580), however that made some of the [kotlin tests fail](https://app.circleci.com/pipelines/github/saks/uniffi-rs/173/workflows/33596a0c-d459-4bc7-a74a-38ee5b466a72/jobs/816). Please let me know if I'm missing something here.

### Docs
With async being added in this PR, I believe that ruby is quite on par with the rest of the batch and I thought that it would be a good opportunity to update docs and tutorials with ruby-specific details. Sure easy to move it to another PR, please let me know if this is needed.

Please let me know if you have any questions or concerns.
Thank you!

### DIFF
--- a/bindgen-tests/README.md
+++ b/bindgen-tests/README.md
@@ -40,14 +40,14 @@ This has a few advantages:
     * Swift
         * Add a Swift test case file in `bindgen-tests/swift/tests`
         * Add a test function for the new test case in `bindgen-tests/swift/src/lib.rs`
-    * Kotlin: TODO
-    * Python: TODO
-    * Ruby: TODO
+    * Kotlin: TODO (see `bindgen-tests/kotlin/` — partially implemented)
+    * Python: see `bindgen-tests/python/tests/`
+    * Ruby: see `bindgen-tests/ruby/tests/`
 
 ## Migration plan
 
-The tests are only currently supported for Swift, the next step is making them work for the other
-bindings languages.
+Swift was the first language migrated; Ruby and Python suites also exist under `bindgen-tests/`.
+Kotlin support is still being ported.
 
 After that, we'll want to keep the existing fixtures for a while, since these tests may not be testing everything.
 New tests should be added here if possible.
--- a/docs/manual/src/futures.md
+++ b/docs/manual/src/futures.md
@@ -1,6 +1,6 @@
 # Async/Future support
 
-UniFFI supports exposing async Rust functions over the FFI. It can convert a Rust `Future`/`async fn` to and from foreign native futures (`async`/`await` in Python/Swift, `suspend fun` in Kotlin etc.)
+UniFFI supports exposing async Rust functions over the FFI. It can convert a Rust `Future`/`async fn` to and from foreign native futures (`async`/`await` in Python/Swift/Ruby, `suspend fun` in Kotlin etc.)
 
 Check out the [examples](https://github.com/mozilla/uniffi-rs/tree/main/examples/futures) or the more terse and thorough [fixtures](https://github.com/mozilla/uniffi-rs/tree/main/fixtures/futures).
 
@@ -96,6 +96,12 @@ Use `uniffi_set_event_loop()` to handle this case.
 It should be called before the Rust code makes the async call and passed an eventloop to use.
 
 Note that `uniffi_set_event_loop` cannot be glob-imported because it's not part of the library's `__all__`.
+
+### Ruby: thread-based async
+
+Unlike Python and Swift, Ruby has no native async/await. When Rust calls an async Ruby callback
+method, UniFFI spawns a new OS thread for each invocation. There is no event loop requirement,
+though a Fiber scheduler is supported for the Ruby -> Rust polling direction.
 
 ## Cancelling async code.
 
--- a/uniffi_bindgen/src/bindings/ruby/templates/Async.rb
+++ b/uniffi_bindgen/src/bindings/ruby/templates/Async.rb
@@ -1,0 +1,184 @@
+# RustFuture poll codes
+UNIFFI_RUST_FUTURE_POLL_READY = 0
+UNIFFI_RUST_FUTURE_POLL_WAKE = 1
+
+# Handle map for storing write-end IO objects used by the continuation callbacks.
+UNIFFI_ASYNC_HANDLE_MAP = UniffiHandleMap.new
+
+# Continuation callback for async functions.
+# Called by Rust when the future is ready to make progress.
+# Writes the poll code to the pipe so the waiting thread/fiber can continue.
+#
+# Exceptions must never escape this Proc.
+# Rust invokes it directly via FFI, so any unhandled Ruby exception would be caught
+# by Ruby-FFI, which swallows it (prints a warning) and returns a garbage value to Rust.
+# The poll code Rust receives would be invalid, causing the Ruby polling loop to
+# never see POLL_READY and block forever.
+UNIFFI_CONTINUATION_CALLBACK = Proc.new do |data, poll_code|
+  begin
+    wr = UNIFFI_ASYNC_HANDLE_MAP.get data
+
+    # This can only be blocking if pipe if full. Rust will never try to call continuation
+    # callback more than once (so will not fill it up), hence this call should never block.
+    wr.putc poll_code
+  rescue Exception
+    # Swallow exception. A leak or a hang is better than a hard VM segfault.
+  end
+end
+
+# Poll a Rust future to completion.
+#
+# This works both with and without a Fiber::Scheduler:
+# - Without scheduler: wait_readable blocks with a timeout, releasing the GVL
+#   so the callback can fire from foreign threads (works around MRI limitation
+#   where rb_thread_call_with_gvl cannot wake up threads in indefinite sleep).
+# - With scheduler: wait_readable hooks into io_wait, yielding the fiber.
+#
+# cancel_fn is called in the ensure block when exception interrupts an in-flight poll.
+# This guarantees Rust fires the continuation callback so the handle-map entry is released
+# and the pipe is drained before we free the future.
+def self.uniffi_rust_call_async(rust_future, poll_fn, cancel_fn, complete_fn, free_fn, lift_func, error_ffi_converter)
+  rd = wr = nil
+  handle = nil
+  poll_in_flight = false
+
+  begin
+    rd, wr = IO.pipe
+    wr.sync = true # avoid buffering and delayed read for rd.wait_readable
+    handle = UNIFFI_ASYNC_HANDLE_MAP.insert wr
+
+    loop do
+      poll_in_flight = true
+      UniFFILib.public_send(poll_fn, rust_future, UNIFFI_CONTINUATION_CALLBACK, handle)
+
+      # Blocks until the continuation callback writes to the pipe.
+      # Releases the GVL so the callback can fire from foreign threads.
+      # With a Fiber::Scheduler, this hooks into io_wait for non-blocking concurrency.
+      rd.wait_readable
+      poll_code = rd.getbyte
+      poll_in_flight = false
+
+      break if poll_code == UNIFFI_RUST_FUTURE_POLL_READY
+    end
+
+    result = if error_ffi_converter.nil?
+      ::{{ ci.namespace()|class_name_rb }}.rust_call(complete_fn, rust_future)
+    else
+      ::{{ ci.namespace()|class_name_rb }}.rust_call_with_error(error_ffi_converter, complete_fn, rust_future)
+    end
+
+    lift_func.call(result)
+  ensure
+    # Defer any further Thread#raise / Timeout during cleanup so all steps execute
+    # automatically. Without this, a second raise during wait_readable(0.5) would skip
+    # handle-map removal, pipe close, and free_fn - leaking FDs and Rust memory.
+    Thread.handle_interrupt(Exception => :never) do
+      if poll_in_flight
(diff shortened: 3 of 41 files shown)


bindgen-tests/README.md
```diff
@@ -40,14 +40,14 @@ This has a few advantages:
     * Swift
         * Add a Swift test case file in `bindgen-tests/swift/tests`
         * Add a test function for the new test case in `bindgen-tests/swift/src/lib.rs`
-    * Kotlin: TODO
-    * Python: TODO
-    * Ruby: TODO
+    * Kotlin: TODO (see `bindgen-tests/kotlin/` — partially implemented)
+    * Python: see `bindgen-tests/python/tests/`
+    * Ruby: see `bindgen-tests/ruby/tests/`
```
bendk, line 45: Thanks for updating these, I totally forgot about them.

docs/manual/src/futures.md
```diff
@@ -1,6 +1,6 @@
 # Async/Future support
 
-UniFFI supports exposing async Rust functions over the FFI. It can convert a Rust `Future`/`async fn` to and from foreign native futures (`async`/`await` in Python/Swift, `suspend fun` in Kotlin etc.)
+UniFFI supports exposing async Rust functions over the FFI. It can convert a Rust `Future`/`async fn` to and from foreign native futures (`async`/`await` in Python/Swift/Ruby, `suspend fun` in Kotlin etc.)
```
bendk, line 3: Can you add a note on how Ruby async differs from the rest?  IIUC, the main difference is that async calls from Rust -> Ruby will spawn a new thread.

uniffi_bindgen/src/bindings/ruby/templates/Async.rb
```diff
@@ -0,0 +1,184 @@
...
+  thread.raise(UniffiInternalCancelled, 'Future was canceled') if once.claim! && thread&.alive?
+end
+
+# Execute a foreign async callback method in a background thread.
+# Enforces the at-most-once guarantee on handle_success / handle_error: whichever
+# fires first (normal completion or Rust-side drop) suppresses the other.
+def self.uniffi_trait_interface_call_async(make_call, uniffi_out_dropped_callback, handle_success, handle_error, error_type = nil, lower_error = nil)
+  once = UniffiOnceFlag.new
+
+  thread = Thread.new do
```
bendk, line 139: Starting a new thread seems okay to me, but out of curiosity is there any way to make this a fiber when `Fiber::Scheduler` is present?

uniffi_core/src/ffi/rustfuture/future.rs
```diff
@@ -262,15 +262,28 @@ where
 
     unsafe fn waker_wake(ptr: *const ()) {
         trace!("RustFuture::waker_wake called ({ptr:?})");
-        Self::recreate_arc(ptr).scheduler.lock().unwrap().wake();
+        let arc = Self::recreate_arc(ptr);
+        // Take the callback out under the lock, then invoke it AFTER releasing
+        // the lock. The callback may re-enter foreign code that blocks on a
+        // runtime lock, which could deadlock if we still held the scheduler lock.
```
bendk: Could you elaborate on what you mean by "The callback may re-enter foreign code that blocks on a runtime lock"?  I think this means that the waker callback might poll the future directly, rather then scheduling a poll call for later.

a) I'm not sure how this would happen, can you walk me through it?
b) I think this might be an issue with other async code.  I can't find clear documentation about this and I might be misremembering, but I believe many futures assume that the `poll()` won't be called inside the `wake()` call and will deadlock if it is.

bendk (changes requested):
This is really great.  My only concern is the locking change and that might just because I'm missing something.  Every time there's async work I feel like I need to re-learn what I figured out last time because I forgot it.

I'll check out that kotlin issue and see what's up.

bendk:
https://github.com/mozilla/uniffi-rs/pull/2925 should fix the issues with async methods on records, thanks for catching that one.

saks:
> #2925 should fix the issues with async methods on records, thanks for catching that one.

Great, once it is merged, I'll return my changes for these kinds of methods, thank you!

uniffi_bindgen/src/bindings/ruby/templates/Async.rb
saks: Thank you for the question, let me think about my options here.

docs/manual/src/futures.md
saks: Good call! I've added more details below in the same file, please have a look.

uniffi_core/src/ffi/rustfuture/future.rs
saks: I've updated comments, hopefully these make more sense now. Please let me know if those are too... hm... ruby-focused, I can try to clean it up.

Let me try to explain my understanding of the scenario that I think I understand well. I'm describing the logic in `main` branch, not after my changes.

### Scheduler
The sequence starts with ruby calling async rust function.
1. Ruby enters pool loop and calls `rust_future_poll(callback)`
2. Ruby releases GVL when calls `rd.wait_readable`
3. Meanwhile: Rust polls inner future -> Pending
4. Rust async work completes and `waker_wake` fires
5.   it acquires scheduler lock and calls callback (that needs GVL for Ruby bindings)
6. On Ruby side, exception/timeout arrives and ruby acquires GVL and enters `ensure` block 
7.   here Ruby calls `rust_future_cancel` which wants scheduler lock
8. Meanwhile Rust is waiting for GVL that is held by `ensure` block.

I believe it is only an issue for Ruby, since it is, as far as I understand, the only language that needs to acquire global runtime lock in order to execute callbacks.

As for the `oneshot.rs`. I don't fully understand how deadlock happens. It is rather an empirical observation at this point. I'll try to dig deeper if you like.

uniffi_core/src/ffi/rustfuture/future.rs
```diff
@@ -262,15 +262,31 @@ where
...
         trace!("RustFuture::waker_wake called ({ptr:?})");
-        Self::recreate_arc(ptr).scheduler.lock().unwrap().wake();
+        let arc = Self::recreate_arc(ptr);
+        // Take the callback out under the lock, then invoke it AFTER releasing
+        // the lock. The callback crosses the FFI into a foreign runtime,
+        // which may require acquiring a runtime lock (e.g. Ruby's GVL).
+        // Meanwhile, the calling Ruby thread holds the GVL and may enter
+        // cancel/free (which acquires the scheduler lock) - an ABBA deadlock
+        // between scheduler lock and the GVL.
+        let callback = arc.scheduler.lock().unwrap().wake();
```
bendk: Like I mentioned in that comment, I think this makes sense in general, but I'd like to keep it so `wake()` calls the callback.  It feels a bit safer/cleaner to me, since we can verify that the scheduler always calls the callback exactly once without needing to check the call sites.

What if you moved the Mutex inside the scheduler?  Then I think you could do a similar dance inside the `wake()` method.

uniffi_core/src/ffi/rustfuture/future.rs
bendk: Interesting, I guess this could be an issue for Python as well.  Maybe the way Python handles the timeout/exceptions avoids that problem or maybe we just didn't test it enough.  This does seem like a different issue that the one I was thinking of and it justifies the changes to `uniffi_core`.

uniffi_core/src/ffi/rustfuture/future.rs
saks: Thank you, let me look into it!

saks: I've moved `Mutex` inside the scheduler. This allowed me to update all methods to execute callback after releasing the lock with minimal changes overall. All logical changes are now in scheduler module only, all callback are called there.

I've also added one more Ruby test that makes it super easy to trigger a deadlock.

Please have a look and let me know if you have any suggestions.

uniffi_bindgen/src/bindings/ruby/templates/Async.rb
saks: There are several issues that made me abandon the idea of using Fibers here:
1. Cancellation protocol is much harder to implement, need some cooperative polling.
2. It is not very common for Ruby application to use fibers (I don't see too many uses in the field I'm working)
3. Even if I figure out how Rust can execute code in Ruby thread with attached scheduler, we still need ruby code in the callback to be carefully written not to block. I do not expect this to be a reasonable expectation. And if callback ruby code is blocking - we're back to a regular blocking thread but with much complexity added.

Take this, however, with a grain of salt, I don't work with async Ruby much :)

bendk: If Ruby users are expected to start a new thread for each operation, then what's the benefit of supporting `async` functions?  For other languages you normally want it so that your thread pool of N threads can execute more than N operations at once.  Is this just so that Ruby users have some way of using existing async Rust libraries?  That seems a good enough of a reason to do this, I'm just wondering if that's all.  For example, if I was designing a Rust library from scratch and Ruby was my main consumer is there any motivation to use async functions in that library?

Side note: I think you could ignore cancellation.  In general, it's hard to support native cancellation across all languages and we have other languages like Swift that don't.  IMO, the best way to handle this is for a library to expose a `cancel()` method or something similar.

bendk (approved):
Thanks for those changes to `uniffi_core` this looks great to me.  I think the code is ready to merge but I'm still curious about [that question I asked you](https://github.com/mozilla/uniffi-rs/pull/2923#discussion_r3428698998) so I'm going to wait a bit before hitting the button.

uniffi_bindgen/src/bindings/ruby/templates/Async.rb
saks: > If Ruby users are expected to start a new thread for each operation, then what's the benefit of supporting async functions?

It is only true for foreign callbacks. For an async Rust function/method called by Ruby, we have a good enough support (no new thread for each call, current thread is blocked if no sceduler used, if Fibers::Scheduler is present, it automatically hooks into non-blocking wait). I do expect foreign callbacks to be used less often than regular async Rust functions. 

> Side note: I think you could ignore cancelation

👍

saks:
> Thanks for those changes to `uniffi_core` this looks great to me. I think the code is ready to merge but I'm still curious about [that question I asked you](https://github.com/mozilla/uniffi-rs/pull/2923#discussion_r3428698998) so I'm going to wait a bit before hitting the button.

Sounds good, thank you. Could you please give me ~2 hours to add async methods back for records/non-flat enums?

uniffi_bindgen/src/bindings/ruby/templates/Async.rb
bendk: Ahh I see, that makes sense to me.

bendk:
> Sounds good, thank you. Could you please give me ~2 hours to add async methods back for records/non-flat enums?

Sorry I hit the merge button too fast.  I guess we'll need a new PR for that.

saks:
> > Sounds good, thank you. Could you please give me ~2 hours to add async methods back for records/non-flat enums?
> 
> Sorry I hit the merge button too fast.  I guess we'll need a new PR for that.

No worries 👍 Thank you for your help!

PR #2926, opened 2026-06-17
Allow literal number suffix for `uniffi::Enum`

flisky:
`Buffa` generate enum declaration with `i32` [suffix](https://github.com/anthropics/buffa/blob/6b35c31d3ef89afff9047d8db5eab828a7a51b65/buffa-codegen/src/enumeration.rs#L175). uniffi macros goes error with `integer literals with suffix not supported by UniFFI here`, and I think it could be relaxed?

BTW, there're two more `literals with suffix` checking for integer/flaut default value , and maybe it could be relaxed, too?

https://github.com/mozilla/uniffi-rs/blob/ed38554ca5da78411f04bfd2c0a5e8a339e2eac6/uniffi_macros/src/default.rs#L73-L75


fixtures/proc-macro/src/lib.rs
```diff
@@ -292,6 +292,7 @@ fn get_mixed_enum(v: Option<MixedEnum>) -> MixedEnum {
 #[derive(uniffi::Enum)]
 pub enum ReprU8 {
     One = 1,
+    Two = 2u8,
```
mhammond: it's possible `Two` doesn't exist to check enum value skipping, so you mind adding this as `Four` (and maybe even adding a comment that `Two` is explicitly skipped?)

mhammond:
Test is failing at https://github.com/mozilla/uniffi-rs/blob/418ab380b02d83bbdbf9db10802f85629e14109a/fixtures/proc-macro/tests/bindings/test_proc_macro.swift#L191. Also, while you are making that change, adding a new CHANGELOG entry would be appreciated too, thanks!

mhammond (approved):
thanks!

Issue #2881, opened 2026-04-29
Method name collision when implementing a non-exported trait for a UniFFI-exported  trait


threema-theo:
Apparently there is a naming collision after expanding traits exported as `#[uniffi:export(with_foreign)]`

# Expected behavior

I expect that the following code compiles:

```rust
use std::sync::Arc;

uniffi::setup_scaffolding!();

pub trait Foo {
    fn foo(&self) -> String;
}

#[uniffi::export(with_foreign)]
pub trait FooUniffi: Send + Sync {
    fn foo(&self) -> String;
}
impl<T: FooUniffi + ?Sized> Foo for Arc<T> {
    fn foo(&self) -> String {
        FooUniffi::foo(self.as_ref())
    }
}

pub trait Bar {
    fn bar(&self) -> Arc<dyn Foo>;
}

#[uniffi::export(with_foreign)]
pub trait BarUniffi: Send + Sync {
    fn bar(&self) -> Arc<dyn FooUniffi>;
}
impl<T: BarUniffi + ?Sized> Bar for Arc<T> {
    fn bar(&self) -> Arc<dyn Foo> {
        Arc::new(BarUniffi::bar(self.as_ref()))
    }
}
```

# Observed Behavior

The code above does not compile, clippy throws the following error:
```
error[E0308]: mismatched types
   --> src/lib.rs:23:1
    |
 23 | #[uniffi::export(with_foreign)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | |
    | expected trait `FooUniffi`, found trait `Foo`
    | arguments to this function are incorrect
    |
    = note: expected struct `std::sync::Arc<dyn FooUniffi>`
               found struct `std::sync::Arc<dyn Foo>`
help: the return type of this call is `std::sync::Arc<dyn Foo>` due to the type of the argument passed
   --> src/lib.rs:23:1
    |
 23 | #[uniffi::export(with_foreign)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this argument influences the return type of `lower_return`
note: associated function defined here
   --> /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/uniffi_core-0.31.1/src/ffi_converter_traits.rs:296:8
    |
296 |     fn lower_return(v: Self) -> Result<Self::ReturnType, RustCallError>;
    |        ^^^^^^^^^^^^
[...]

PR #2934, opened 2026-06-29
Scaffolding tests (and #2881)

bendk:
Starting the scaffolding-tests crate.  This is like bindgen-tests, but it targets the generated scaffolding code without round-tripping it through a foreign implementation.

Used https://github.com/mozilla/uniffi-rs/issues/2881 as the first test case and fixed it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2273,6 +2273,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "uniffi-scaffolding-tests"
+version = "1.0.0"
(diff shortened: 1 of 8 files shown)


bendk:
I thought about this more and I don't really like the idea of the `scaffolding-tests` crate anymore.  We may be moving to a place where bindings are generating their own scaffolding ([2911](https://github.com/mozilla/uniffi-rs/pull/2911), which makes me want to put all the tests in `bindings-tests`.

PR #2935, opened 2026-06-29
fix(kotlin): Make sure uniffiCheckApiChecksums compares against the rust generated u16.

o0Ignition0o:
This changeset is a followup to https://github.com/mozilla/uniffi-rs/issues/2740.

On AArch64, the unused upper bits of the return register are unspecified, and in rust (at the moment) they use the sign-extension of the 16-bit value (see https://github.com/rust-lang/rust/issues/97463#issuecomment-1140210248 for more information).
This means any checksum value greater than or equal to 0x8000 would carry 0xFFFF.
Checksum comparisons would thus always fail comparing a negative value with the expected valid one.
This changeset adds a 16 bits mask to checksums so we only compare the 16 relevant bits, and the check doesn't fail on AArch64 anymore.

I'm not 100% sure, but it looks like the project's CI matrix doesn't contain arm architectures, otherwise the Kotlin path `*/test_generated_bindings.rs` integration tests may have caught it.

If there's anything I can do to improve the fix or add more tests, please let me know!

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
+++ b/uniffi_bindgen/src/bindings/kotlin/templates/NamespaceLibraryTemplate.kt
@@ -120,7 +120,8 @@ private fun uniffiCheckContractApiVersion(lib: IntegrityCheckingUniffiLib) {
 @Suppress("UNUSED_PARAMETER")
 private fun uniffiCheckApiChecksums(lib: IntegrityCheckingUniffiLib) {
     {%- for (name, expected_checksum) in ci.iter_checksums() %}
-    if (lib.{{ name }}() != {{ expected_checksum }}) {
+    {#- please don't delete the mask: https://github.com/mozilla/uniffi-rs/pull/2935 #}
+    if ((lib.{{ name }}() and 0xFFFF) != {{ expected_checksum }}) {
         throw RuntimeException("UniFFI API checksum mismatch: try cleaning and rebuilding your project")
     }
     {%- endfor %}


uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -1126,4 +1126,43 @@ mod test {
...
+        let checksum_checks = bindings
+            .split("private fun uniffiCheckApiChecksums")
+            .nth(1)
+            .expect("generated checksum checks")
+            .split("/**")
+            .next()
+            .expect("end of generated checksum checks");
+
+        assert!(
+            checksum_checks.contains("and 0xFFFF) !="),
```
bendk: I get wanting to add a test for this, but I'm not sure how much value it's adding.  IIUC, the only way this would fail is if someone deleted the `and 0xFFFF` part from the code.  I think a comment saying "please don't delete this and pointing to the PR" would do the same thing.

bendk (commented):
Seems like this came from https://github.com/mozilla/uniffi-rs/pull/2897, which tried to fix the checksums on arm32 arches.  The fix seems fine with me, but I'd like to see someone with more knowledge chime in.  @lwz23, @jkmassel, @paxbun  what do you think?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
o0Ignition0o: True, I'll update the pr

edit: fixed in 965691b

bendk (commented):
Hmm, no comments.  I can't see any reason for this to backfire though, the most that could happen is it has no effect.  How do you feel about merging it @o0Ignition0o?

mhammond (approved):
I'm with Ben and think we should just merge this.

jplatte:
This might be worth a patch release.

PR #2937, opened 2026-06-30
Handle ambiguous method names (#2881)

bendk:
Also updated the `trait_interfaces` test to test the `self: Arc<Self>` codepath.


mhammond (approved):
changelog seems appropriate!

mhammond (approved):
oops, meant to hit "approve" when I mentioned a changelog entry should be added
[edit: apparently I can't read, because I did hit approve 😅 ]

PR #2940, opened 2026-07-07
Add mutable zero-copy byte arrays

### LINKED ISSUE #2864: Implement zero copy byte buffers for `&[u8]` and `&mut [u8]`

jhugman:
Fixes #2864 .

This is the second part of the #2864. The first was implementing non-mutable `&[u8]`. This is for `&mut [u8]`.


docs/manual/src/types/bytes.md
```diff
@@ -57,3 +57,44 @@ Each binding maps `[ByRef] bytes` to a natural zero-copy type:
...
+- **Pass a writable buffer.** Each binding needs a mutable buffer type and
+  rejects a read-only one.
+
+Each binding maps a mutable argument to a writable type:
+
+| Binding | Foreign type                 | Note                                                |
+|---------|------------------------------|-----------------------------------------------------|
+| Swift   | `inout Data`                 | The call runs inside `Data.withUnsafeMutableBytes`. |
+| Python  | `bytearray`                  | Pass `bytes` and it fails — `bytes` is immutable.   |
+| Kotlin  | direct `java.nio.ByteBuffer` | The same direct buffer as `[ByRef]`; writes land in it. |
```
badboy: Might make sense to keep the same order as the table above (Kotlin, Swift, Python)

badboy (approved):
This is surprisingly straight forward. Great job!

docs/manual/src/types/bytes.md
```diff
@@ -57,3 +57,44 @@ Each binding maps `[ByRef] bytes` to a natural zero-copy type:
...
+```
+
+Every `&[u8]` rule above still holds. Two more:
+
+- **Synchronous only.** `&mut [u8]` / `[ByMutRef]` in an `async` function
+  fails to compile (proc-macro) or generate (UDL). An async call can
+  resume on another thread after the caller has moved on and freed the
+  buffer.
+- **Pass a writable buffer.** Each binding needs a mutable buffer type and
+  rejects a read-only one.
```
bendk, line 89: I think there should be another one that says something like:

- Callers must follow Rust's aliasing rules, for example:
   -  Don't pass the same byte array, or overlapping slices, to Rust from separate threads at the same time.
   - Don't mutate or free the byte array while it's passed to a Rust function in another thread.

fixtures/mut-bytes/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! End-to-end UDL fixture for zero-copy `&mut [u8]` arguments declared with
+//! `[ByMutRef] bytes` in the UDL. The scaffolding generator lowers such an
+//! argument to `&mut [u8]`, which flows through the proc-macro machinery so
+//! that Rust writes land in the caller's buffer in place.
```
bendk: Do we need a new fixture for this?  I think the bindgen-tests might be good enough.

docs/manual/src/types/bytes.md
```diff
@@ -57,3 +57,44 @@ Each binding maps `[ByRef] bytes` to a natural zero-copy type:
...
+    void fill([ByMutRef] bytes buf);
+};
+```
+
+Every `&[u8]` rule above still holds. Two more:
+
+- **Synchronous only.** `&mut [u8]` / `[ByMutRef]` in an `async` function
+  fails to compile (proc-macro) or generate (UDL). An async call can
+  resume on another thread after the caller has moved on and freed the
+  buffer.
```
bendk, line 87: I'm fine with this restriction for now, but does this always have to be there?  Even with sync functions there are ways to abuse the system and break Rust's rules.

uniffi_bindgen/src/bindings/python/pipeline/nodes.rs
```diff
@@ -188,24 +188,38 @@ pub struct Argument {
     pub name: String,
     pub ty: TypeNode,
     pub by_ref: bool,
+    pub by_mut_ref: bool,
```
bendk, line 198: We're starting to accumulate these flags.  At some point maybe we should introduce an `ArgKind` enum or maybe move this to the type system (i.e. add `uniffi_meta::Type::Ref` and `uniffi_meta::Type::MutRef` variants).  No need to change anything for this PR, I'm just thinking out loud.

docs/manual/src/types/bytes.md
```diff
@@ -57,3 +57,44 @@ Each binding maps `[ByRef] bytes` to a natural zero-copy type:
...
+- **Pass a writable buffer.** Each binding needs a mutable buffer type and
+  rejects a read-only one.
+
+Each binding maps a mutable argument to a writable type:
+
+| Binding | Foreign type                 | Note                                                |
+|---------|------------------------------|-----------------------------------------------------|
+| Swift   | `inout Data`                 | The call runs inside `Data.withUnsafeMutableBytes`. |
+| Python  | `bytearray`                  | Pass `bytes` and it fails — `bytes` is immutable.   |
```
bendk: Should we use `byterray` for both cases?  Otherwise it's awkward for an API where some functions input `&[u8]` and others input `&mut [u8]`.

bendk (commented):
This looks really nice.  I left some high-level comments, but didn't look through all the internals yet.

docs/manual/src/types/bytes.md
jhugman: Now supports bytearray for both, but `bytes` in the immutable case.

jhugman:
@bendk I have an approval from @badboy , I've addressed your comments, and fixed JNI following a rebase. Is there anything else you need before this can be landed?

docs/manual/src/types/bytes.md
bendk: Very nice, that's great.

bindgen-tests/lib/src/bytes.rs
```diff
@@ -16,3 +16,20 @@ pub fn sum_bytes_procmacro(buf: &[u8]) -> u32 {
 pub fn first_byte_procmacro(buf: &[u8]) -> Option<u8> {
     buf.first().copied()
 }
+
+/// Zero-copy `&mut [u8]` — proc-macro path. Overwrites `buf` with a known
+/// pattern (`buf[i] = i as u8`). The caller observes the writes in place.
+#[uniffi::export]
+pub fn fill_bytes_procmacro(buf: &mut [u8]) {
```
bendk: nit: can you remove the `_procmacro` suffix?  AFAICT, there's only the proc-macro path now.

uniffi-bindgen-kotlin-jni/runtime/src/lift_lower.rs
```diff
@@ -539,3 +539,24 @@ pub unsafe fn lift_bytes_ref<'a>(env: *mut JNIEnv, byte_buffer: jobject) -> Resu
...
+    env: *mut JNIEnv,
+    byte_buffer: jobject,
+) -> Result<&'a mut [u8]> {
+    let ptr = ((**env).v1_4.GetDirectBufferAddress)(env, byte_buffer).cast::<u8>();
+    if ptr.is_null() {
+        bail!("lift_bytes_mut_ref: GetDirectBufferAddress failed");
+    }
+    let capacity = ((**env).v1_4.GetDirectBufferCapacity)(env, byte_buffer) as usize;
+    Ok(std::slice::from_raw_parts_mut(ptr, capacity))
+}
```
bendk, line 585: Nice!  Thanks for updating `uniffi-bindgen-kotlin-jni`.

uniffi-bindgen-kotlin-jni/runtime/src/lift_lower.rs
```diff
@@ -539,3 +539,24 @@ pub unsafe fn lift_bytes_ref<'a>(env: *mut JNIEnv, byte_buffer: jobject) -> Resu
...
+/// # Safety
+/// env must point to a valid JNIEnv
+pub unsafe fn lift_bytes_mut_ref<'a>(
+    env: *mut JNIEnv,
+    byte_buffer: jobject,
+) -> Result<&'a mut [u8]> {
+    let ptr = ((**env).v1_4.GetDirectBufferAddress)(env, byte_buffer).cast::<u8>();
+    if ptr.is_null() {
+        bail!("lift_bytes_mut_ref: GetDirectBufferAddress failed");
+    }
```
bendk: I notice the empty buffer test case doesn't work here.  Could you set ptr to [NonNull::dangling()](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.dangling) for when there's a null pointer and 0 capacity to make that pass?  I'm not that concerned either way, but it could be nice.

bendk (changes requested):
This is looking really nice.  I have a couple minor requests, but I think we can merge this very soon.

uniffi-bindgen-kotlin-jni/runtime/src/lift_lower.rs
jhugman: I added a helper `direct_buffer_ptr`, then used it from `lift_bytes_mut_ref` and `lift_bytes_ref`.

It validates the ptr as a non null pointer or zero capacity.

jhugman:
Cool, thanks @bendk ! I have addressed your comments.

bendk:
Did you forgot to push?  I don't see any new changes.

docs/manual/src/types/bytes.md
mhammond: nano-nit, I think this would read better as "`bytes` can't be used as a mutable argument" - for some odd reason this confused me.

docs/manual/src/types/bytes.md
```diff
@@ -51,9 +51,56 @@ Each binding maps `[ByRef] bytes` to a natural zero-copy type:
 |---------|--------------------------|---------------------------------------------------------------|
 | Kotlin  | `java.nio.ByteBuffer`    | Must be a *direct* buffer (`ByteBuffer.allocateDirect(...)`). |
 | Swift   | `Data`                   | The call runs inside `Data.withUnsafeBytes`.                  |
-| Python  | `bytes`                  | The `bytes` object's stable internal pointer is used.         |
+| Python  | `bytes` or `bytearray`   | Zero-copy either way; a `bytes` object's buffer is immutable and stable. |
```
mhammond: similarly I find this jarring - "The object's stable internal pointer is used in both cases"?

docs/manual/src/types/bytes.md
```diff
@@ -51,9 +51,56 @@ Each binding maps `[ByRef] bytes` to a natural zero-copy type:
...
+## Mutable buffers (`&mut [u8]` / `[ByMutRef] bytes`)
+
+A **synchronous** function can take `&mut [u8]` (or `[ByMutRef] bytes` in
+UDL) to borrow a foreign-owned buffer and write to it. The writes land in
+the caller's buffer — no copy in or out.
+
+```rust
+#[uniffi::export]
+pub fn fill(buf: &mut [u8]) {
+    // The caller sees these writes once the call returns.
```
mhammond: This comment almost implies the buffer isn't updated until then - I'd suggest just remove it?

uniffi_bindgen/src/pipeline/general/nodes.rs
```diff
@@ -185,6 +185,7 @@ pub struct Argument {
     pub orig_name: String,
     pub ty: TypeNode,
     pub by_ref: bool,
+    pub by_mut_ref: bool,
```
mhammond, line 188: not a big deal, but why not an enum here?

mhammond (commented):
This is fantastic, thanks!

jhugman:
> Did you forgot to push? I don't see any new changes.

Gah, I think you're right. Pushed now.

uniffi_bindgen/src/pipeline/general/nodes.rs
jhugman: In all honesty, expediency: trying to keep a relatively simple PR getting bigger; and the desire to not make another breaking change. 

I can file a follow-up issue, or fix it here. Which would you prefer?

bendk (approved):
This looks great to me.  I think the only issue left is the question is if we should switch to an enum for `by_ref` and `by_but_ref` in this PR or in a followup.

mhammond:
I believe it should be an enum, as currently it's possible to represent an "impossible" state iiuc. I'm not too bothered if it happens in this or a followup, but I'm really trying to avoid a followup being filed and never implemented. Your agent can probably do this easily and an additional commit for ease of review might allay the reviewability concern. otoh, if you commit to doing this in a fast follow I'm perfectly fine.

jhugman:
Let's do a fast follow. Have filed #2954, ~~and assigned to it myself~~. Please assign it to me.

mhammond:
fantastic, thank you!

mhammond:
I do wonder how you feel about suggestion to use `Span` and `RawSpan` on Swift. Naively it seems to make sense and it will be difficult to change once released.

snej:
Could Spans be a configurable option?

Issue #2929, opened 2026-06-18
Swift 6 strict concurrency: async foreign-trait codegen emits #SendingClosureRisksDataRace hard error (0.31.2)


losrodriguez17:
### Summary

Under Swift 6 strict concurrency (language mode `.v6`), the **async foreign-trait** Swift bindings generated by 0.31.2 fail to compile with a hard error:

```
error: passing closure as a 'sending' parameter risks causing data races between
code in the current task and concurrent execution of the closure
[#SendingClosureRisksDataRace]
```

This is the async analogue of the sync-callback issue fixed in 0.31.2 ("Prevent strict concurrency warning with callback interface tables"). The sync path is now clean, but the **async foreign-trait dispatch path is still uncovered**, so any `#[uniffi::export(with_foreign)]` trait that has `async` methods cannot be consumed from a Swift target built in language mode `.v6`.

### Versions

- uniffi: **0.31.2** (latest on crates.io)
- Swift: **6.3.2** (`swiftlang-6.3.2.1.108`), target `arm64-apple-macosx26.0` (Xcode 26.5)
- Swift package: `swiftLanguageModes: [.v6]`

### Where it fails

The shared async dispatch helpers run the foreign call inside `let task = Task { ... }`:

```swift
private func uniffiTraitInterfaceCallAsync<T>(
    makeCall: @escaping () async throws -> T,
    handleSuccess: @escaping (T) -> (),
    handleError: @escaping (Int8, RustBuffer) -> (),
    droppedCallback: UnsafeMutablePointer<UniffiForeignFutureDroppedCallbackStruct>
) {
    let task = Task {          // <-- #SendingClosureRisksDataRace
        var callResult: T
        do {
            callResult = try await makeCall()
        } catch {
            handleError(CALL_UNEXPECTED_ERROR, FfiConverterString.lower(String(describing: error)))
            return
        }
        handleSuccess(callResult)
    }
    ...
}
```

(`uniffiTraitInterfaceCallAsyncWithError<T, E>` has the identical shape and fails the same way.)

### Root cause

[...]

PR #2943, opened 2026-07-13
Fixed swift concurrency error (#2929)

bendk:
I saw the error when running the bindgen-tests with Swift 6.3 and UNIFFI_TEST_SWIFT_VERSION=6. Fixed it by marking all closures @Sendable and also forcing the RustBuffer values to be Sendable via `nonisolated(unsafe)` which I got from https://github.com/swiftlang/swift/blob/dc600dc96ae7e992a879f679714c55d7dc8a0dd6/userdocs/diagnostics/sendable-closure-captures.md.

### DIFF
--- a/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
+++ b/uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
@@ -31,7 +31,16 @@ fileprivate struct {{ trait_impl }} {
             uniffiCallStatus: UnsafeMutablePointer<RustCallStatus>
             {%- endif %}
         ) in
-            let makeCall = {
+            {%- for arg in ffi_callback.arguments() %}
+            {%- if let FfiType::RustBuffer(_) = arg.type_() %}
+            {# RustBuffer is not a Sendable type because it stores a pointer.
+             # However, we coordinate with the Rust side of the FFI to ensure it's used in a thread-safe manner.
+             -#}
+            nonisolated(unsafe) let {{ arg.name()|var_name }} = {{ arg.name()|var_name }}
+            {%- endif %}
+            {%- endfor %}
+
+            let makeCall: @Sendable () {% if meth.is_async() %}async {% endif %}throws -> {{ meth.return_type()|return_type_name }} = {
                 () {% if meth.is_async() %}async {% endif %}throws -> {% match meth.return_type() %}{% when Some(t) %}{{ t|type_name }}{% when None %}(){% endmatch %} in
                 guard let uniffiObj = try? {{ ffi_converter_name }}.handleMap.get(handle: uniffiHandle) else {
                     throw UniffiInternalError.unexpectedStaleHandle
@@ -68,7 +77,7 @@ fileprivate struct {{ trait_impl }} {
             {%- endmatch %}
             {%- else %}
 
-            let uniffiHandleSuccess = { (returnValue: {{ meth.return_type()|return_type_name }}) in
+            let uniffiHandleSuccess: @Sendable ({{ meth.return_type()|return_type_name }}) -> () = { (returnValue) in
                 uniffiFutureCallback(
                     uniffiCallbackData,
                     {{ meth.foreign_future_ffi_result_struct().name()|ffi_struct_name }}(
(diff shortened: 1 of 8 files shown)


bendk:
Here's the generated output before/after and a diff.

[uniffi_bindgen_tests.swift.zip](https://github.com/user-attachments/files/29983141/uniffi_bindgen_tests.swift.zip)

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
```diff
@@ -31,7 +31,16 @@ fileprivate struct {{ trait_impl }} {
             uniffiCallStatus: UnsafeMutablePointer<RustCallStatus>
             {%- endif %}
         ) in
-            let makeCall = {
+            {%- for arg in ffi_callback.arguments() %}
+            {%- if let FfiType::RustBuffer(_) = arg.type_() %}
+            {# RustBuffer is not a Sendable type because it stores a pointer.
+             # However, we coordinate with the Rust side of the FFI to ensure it's used in a thread-safe manner.
+             #}
+            nonisolated(unsafe) let {{ arg.name()|var_name }} = {{ arg.name()|var_name }}
```
bendk, line 39: I don't feel super confident about this one, but it seems okay because we do  make sure that we use our C data in a thread-safe manner.

bendk:
With this change I see warnings when running in Swift-5 mode:

```
uniffi_bindgen_tests.swift:1284:29: warning: converting non-Sendable function value to '@Sendable (TestError) -> RustBuffer' may introduce data races
1282 |                 handleSuccess: uniffiHandleSuccess,
1283 |                 handleError: uniffiHandleError,
1284 |                 lowerError: FfiConverterTypeTestError_lower,
     |                             `- warning: converting non-Sendable function value to '@Sendable (TestError) -> RustBuffer' may introduce data races
1285 |                 droppedCallback: uniffiOutDroppedCallback
1286 |             )
```

That seems okay to me, but I wonder what others think.

uniffi_bindgen/src/bindings/swift/templates/CallbackInterfaceImpl.swift
ih-codes: As long as you are assuming full responsibility for proper thread safety on the `RustBuffer`, this is fine. Like I mentioned in slack, `nonisolated(unsafe)` is opting this type out of compiler checks (so we won't get errors/warnings). You would then have to ensure the `RustBuffer` type is protected with its own runtime synchronization methods (locks, queues, etc.). Which it sounds like you're doing. 👍

ih-codes (approved):
Hi @bendk ! I took a quick look and checked the diffs. No concerns from my end with adding `@Sendable` since you get warnings/errors if your types are not actually Sendable. As for the `nonisolated(unsafe)`, I added a note but I'm not concerned here either since you say you have threading guarantees elsewhere.

Regarding your comment:
> With this change I see warnings when running in Swift-5 mode:
> 
> ```
> uniffi_bindgen_tests.swift:1284:29: warning: converting non-Sendable function value to '@Sendable (TestError) -> RustBuffer' may introduce data races
> 1282 |                 handleSuccess: uniffiHandleSuccess,
> 1283 |                 handleError: uniffiHandleError,
> 1284 |                 lowerError: FfiConverterTypeTestError_lower,
>      |                             `- warning: converting non-Sendable function value to '@Sendable (TestError) -> RustBuffer' may introduce data races
> 1285 |                 droppedCallback: uniffiOutDroppedCallback
> 1286 |             )
> ```
> 
> That seems okay to me, but I wonder what others think.

Could you try adding `@Sendable` before the access control modifier of `FfiConverterTypeTestError_lower` to see if that fixes the warning? 🤔  I would've expected the compiler to infer that but maybe it can't here for some reason.
```
@Sendable public func FfiConverterTypeTestError_lower(_ value: TestError) -> RustBuffer {
    return FfiConverterTypeTestError.lower(value)
}
```

adudenamedruby (approved):

bendk:
> Could you try adding @Sendable before the access control modifier of FfiConverterTypeTestError_lower to see if that fixes the warning? 🤔 I would've expected the compiler to infer that but maybe it can't here for some reason.

That worked great, thanks!

> As long as you are assuming full responsibility for proper thread safety on the RustBuffer, this is fine.

Yeah, this has always been the UniFFI model: we need to manually ensure thread safety ourselves.  I don't think we have an alternative either, since we're passing these functions across the FFI so there's no way a compiler on either side can verify things themselves.

ih-codes:
> That worked great, thanks!

Oh great, very good! 👌 

> Yeah, this has always been the UniFFI model: we need to manually ensure thread safety ourselves. I don't think we have an alternative either, since we're passing these functions across the FFI so there's no way a compiler on either side can verify things themselves.

Yes, I agree with this (about no alternatives). Like you said the compiler can't really help us here anyway so it's in the hands of developers to ensure thread safety one way or another!

mhammond (approved):

fenhl:
Hi, any plans for a new release? I ran into #2929 in a project of mine.

PR #2944, opened 2026-07-16
Swift concurrency fixes for bindgen-tests

bendk:
Add `nonisolated(unsafe)` to global variables and put them behind a lock just in case there really was a concurrency issue with how we were accessing them.

Don't use DispatchGroups/Tasks to schedule the async code anymore, instead just call the async functions directly. The old code hangs Swift 6 It's not completely clear to me, but I think it has to do with the fact that the way we were creating tasks was not compatible with whatever async context Swift creates for the script.

This makes the tests pass for me in both swift 5 and swift 6 mode.

### DIFF
--- a/bindgen-tests/swift/tests/callback_interfaces.swift
+++ b/bindgen-tests/swift/tests/callback_interfaces.swift
@@ -2,20 +2,30 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+import Foundation
 import uniffi_bindgen_tests
 
-var callbackRefCount = 0;
+// We want to mutate callbackRefCount from async swift code.
+// The simplest way to do that seems to be to mark it `nonisolated(unsafe)`.
+nonisolated(unsafe) var callbackRefCount = 0;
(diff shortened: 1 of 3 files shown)


bindgen-tests/swift/tests/callback_interfaces.swift
```diff
@@ -2,20 +2,30 @@
...
+import Foundation
 import uniffi_bindgen_tests
 
-var callbackRefCount = 0;
+// We want to mutate callbackRefCount from async swift code.
+// The simplest way to do that seems to be to mark it `nonisolated(unsafe)`.
+nonisolated(unsafe) var callbackRefCount = 0;
+// Let's use this lock whenever we mutate callbackRefCount, just to make sure that everything is
+// thread-safe.
+nonisolated(unsafe) var callbackRefCountLock = NSLock()
```
mhammond, line 13: this all seems good, but wondering why swift atomics weren't used?

mhammond (approved):

bindgen-tests/swift/tests/callback_interfaces.swift
bendk: Don't they require an extra package?  I can never figure out how to use them with my basic setup.

Issue #2903, opened 2026-05-18
UnexpectedUniFFICallbackError: misleading reason when Kotlin callback throws Throwable or Error


threema-torsten:
# UnexpectedUniFFICallbackError shows misleading reason when Kotlin callback throws `Throwable` or `Error`
 
When a Kotlin callback throws a `Throwable` that does not extend `java.lang.Exception`, the `reason` field of `UnexpectedUniFFICallbackError` shows a confusing internal message instead of the actual error.
 
## Expected behavior
The reason should reflect the thrown error, e.g. `ThrowableClass` or its message.
 
## Actual behavior
The reason is always `not enough bytes remaining in buffer (0 < 4)`, regardless of what was thrown.
 
## Reproduction
 
Throwing an instance of this class works correctly:
```kotlin
class ExceptionClass : Exception()
```
 
But throwing any of these produces the misleading reason:
```kotlin
class ThrowableClass : Throwable()
class ErrorClass : java.lang.Error()
```
 
## Root cause (suspected)
UniFFI's callback error handling appears to only serialize `Exception` subclasses. Anything higher in the `Throwable` hierarchy causes a serialization failure, which surfaces as the buffer error instead of the original exception.
 
## Impact
This makes debugging callback errors very difficult, since the misleading message gives no indication of what actually went wrong.

mhammond:
Sounds plausible - @paxbun?

mhammond:
fixed by #2945

PR #2945, opened 2026-07-17
Fix Kotlin Throwable handling (#2903)

bendk:
Catch kotlin.Throwable rather than kotlin.Exception, which allows us to get the message correctly.  Also updated the tests for uniffi-bindgen-kotlin-jni, but we were actually already handling it right there.


mhammond (commented):
probably worth a changelog, particularly if it is the complete fix for the referenced issue.

mhammond (approved):

Issue #2295, opened 2024-10-31
Allow calling from more than one language at the same time


jhugman:
I'd like to be able to initialize and use Rust from multiple languages at the same time [in the same process].

Example use case:

From React-Native Javascript, call into Rust which then calls out to Kotlin.

The C ABI should remain the same for the foreign language, so calling _in_ to Rust would need no modifications.

I _think_ the only thing to need to change would be adding an index parameter to callback registration and `UniffiRustFutureContinuationCallback`, so that the callback knows which `vtable` to call into. This would require work on both foreign and Rust sides.

I don't think this would be on Mozilla's roadmap, so am quite happy to do the work.

Is this a feature you'd be willing to land?

mhammond:
It sounds interesting and not too intrusive, so I don't see why not!

jhugman:
Ok, so I got callbacks working. This should be just introducing a `lang_index`, and passing it as another value in the `FfiConverterCallback`, making it an `FfiConverterRustBuffer`.

Futures doesn't look too complicated: it looks like it should just work unchanged.

Foreign traits are looking a bit more problematic, and may need some wider changes: because they're implemented in Rust or foreign side, and you can call them like an object, they need to be able to call `free` and `clone_pointer`. Two possible approaches come to mind: 

1. add a variant of the `FfiType` to account for the differences between a `RustArcPtr` and `RustArcPtr` with a source language.
2. modify the associated value for `FfiType::RustArcPtr(_)` from a seemingly unused `String` to a new `ArcType` enum with variants `Raw` and `WithSource`.
3. generate extra constructors, `clonePointer` and `free` methods that accept `RustBuffers` instead of Pointers.

My preference I think would be 2, but I'm wondering if this may be pulling on a ball of string to something "too intrusive". What do you think? Should I keep going? Any advice @mhammond ?

mhammond:
Can you put up a draft PR to help make it concrete? @bendk understands all this stuff the best by far though, so I'm happy to look but his feedback will be more valuable.

bendk:
I think a draft PR would be great.  One thing we've been talking about is trying to improve the internals docs, which are really out of date at this point.  Maybe you could add something there that explains the issue and how you want to solve it.  Bonus points for mermaid diagrams (BTW, I'm hoping to update the lift/lower ones soon and I'm planning on stealing the diagram from your recent JS talk for that).

[...]

PR #2958, opened 2026-08-07
feat: scope callback vtables per realm

### LINKED ISSUE #2957: Callbacks run in the wrong realm when several share one cdylib
Related to #2295, which is about several languages in one process. Same root cause, no multi-language involved: one cdylib, one language, several realms of it.

**Setup**

A Node `worker_threads` pool sharing one cdylib. Vitest and Jest use one by default, so this reaches people who never asked for threads.

**Example**

A callback interface, and a function that calls it:

```rust
pub trait Source: Send + Sync {
    fn read_at(&self, offset: u64, length: u32) -> Vec<u8>;
}

#[uniffi::export]
pub fn first_byte(source: Arc<dyn Source>) -> u8 {
    source.read_at(0, 1)[0]
}
```

Each worker implements it over its own data:

```js
// worker.js, started twice: once with 0xAA, once with 0xBB
const { firstByte } = require("./generated");

class Source {
  constructor(byte) { this.byte = byte; }
  readAt(offset, length) { return new Uint8Array([this.byte]).buffer; }
}

console.log(firstByte(new Source(myByte)));
```

Each worker should print the byte it was given. Both print the same one, whichever worker registered last, and no error is raised. In my case 0 of 12 worker realms were correct.

**Why**

Both workers register a vtable for `Source`. `UniffiForeignPointerCell` is one `AtomicPtr` per cdylib and `set` stores unconditionally, so the last registration owns the callback for the whole process.

Handles hide it. `handle.rs` mints foreign handles from a map of odd values and every realm's map starts at the same place, so the handle from worker A finds a real object in worker B's map. A stale handle would be caught; a colliding one is not.

**A language index would not fix this**

#2295 is heading towards encoding a language index in the handle. Two Node worker realms are both JS, so they get the same index and still collide. Scoping by foreign runtime instance would work, and also separates two languages. Scoping by language does not separate two realms of one language.

**Workaround, and what it costs**

Each realm loads its own byte-for-byte copy of the cdylib, so each gets its own statics. That takes 0 of 12 realms to 12 of 12. It has to be a real copy: `dlopen` dedupes by device and inode, so a hard link or a symlink returns the existing image with the existing statics.

The copy defeats page sharing. Distinct files are distinct inodes, so text and rodata are resident once per realm, about 19 MB each for my release build, scaling with the worker count. It also depends on loader behaviour that uniffi does not document relying on.

**Fix**

#2958 gives the cell one slot per realm, keyed by an identifier the bindings tag into the top bits of the handles they mint. That space is free: only the lowest bit is reserved. Backward compatible, and no bindings need regenerating.

That is the Rust half. Tagging the handles is the bindings' side, opted into separately. I have the Node side working locally and can follow up with it.

stupside:
Fixes #2957. Related to #2295, which asks for several languages in one process; this is the same root cause reached with one language and several realms of it.

## The problem

Several instances of a foreign runtime can load one cdylib in a single process: Node worker threads, JVM classloaders, Python sub-interpreters. Each registers its own callback vtable.

`UniffiForeignPointerCell` held exactly one pointer and `set` is an unconditional store, so the last realm to register owned every callback in the process.

Foreign handles then made it silent instead of loud. Each realm mints handles from its own map starting at the same value, so realm A's handle 3 and realm B's handle 3 are the same number. A call dispatched into the wrong realm resolves to a real object there rather than failing, so callers get wrong data instead of an error. I measured 0 of 12 Node worker realms behaving correctly.

## The change

The cell holds one registration per realm. A realm is identified by bits the foreign bindings tag into the top of every handle they mint, which is free space: only the lowest bit is reserved, to distinguish a foreign handle from a Rust pointer. The tag travels with the handle back into Rust, so `get_for_handle` routes a call to the realm that owns the object.

Every lookup site in the generated trait impl already has `self.handle` in scope, so the codegen change is `get()` to `get_for_handle(self.handle)` at four sites.

The realm is carried as a `u64` end to end and registrations live in a `Vec`, so there is no ceiling on how many realms may share an image and no width conversion anywhere on the path. Registration happens once per realm at startup, so the list is tiny and effectively write-once; a linear scan costs less than hashing and is dwarfed by the foreign call each lookup precedes.

## Backward compatibility

Nothing existing has to change, in any language.

- `set` and `get` keep their signatures and mean realm 0.
- An untagged handle reads as realm 0, and a realm with no registration falls back to realm 0, so bindings that know nothing about realms behave exactly as before.
- Registration gains an additive `..._for_realm` symbol rather than changing the signature of the existing one.
- `uniffi_meta` and `uniffi_bindgen` are untouched, so no checksum moves and no bindings need regenerating.

I checked that last point by running bindings generated against stock 0.31.0, without regenerating them, on a library built with this patch: a checksum change would have thrown `ApiChecksumMismatch` at import. They loaded and passed, including a four-worker-realm test.

## Tests

Five unit tests on the cell: realm extraction from the top bits, untagged handles reading as realm 0, two realms reading back their own registration rather than overwriting each other, re-registration replacing in place, fallback to realm 0 for an unregistered realm, and a realm at the top of the range to show there is no fixed bound.

`cargo test -p uniffi_core` is green (40 + 2), `cargo clippy -p uniffi_core --all-targets` is clean, and `cargo fmt` has been run.

## What this does not do

It gives each realm its own vtable registration. It does not tag handles: that is the foreign bindings' side of the contract, and each set of bindings has to opt in by minting tagged handles and calling `..._for_realm`. Bindings that do nothing keep working as realm 0.

I have the Node side of this working against a local build and can follow up with it, or with a second language, if that would help review.

One design note for #2295: the identity that matters is the foreign runtime instance rather than the language. Two Node worker realms are both JS, so a language index would give them the same value and they would still collide. Scoping by runtime instance also separates two languages, so it is the strictly finer of the two.

### DIFF
--- a/uniffi_core/src/ffi/foreigncallbacks.rs
+++ b/uniffi_core/src/ffi/foreigncallbacks.rs
@@ -9,29 +9,81 @@
 //! storing the callback.
 
 use std::{
-    ptr::{null_mut, NonNull},
-    sync::atomic::{AtomicPtr, Ordering},
+    ptr::NonNull,
+    sync::{PoisonError, RwLock},
 };
 
-// Cell type that stores any NonNull<T>
+/// Bit position of the realm identifier inside a foreign handle.
+///
+/// A realm is one instance of a foreign runtime: a Node worker thread, a JVM classloader, a
+/// Python sub-interpreter. Several can load the same cdylib in one process, and each registers
+/// its own callback vtable pointing at its own runtime.
+///
+/// Foreign bindings reserve only the lowest bit of a handle, which must be set so a foreign
+/// handle is distinguishable from a Rust pointer. The top bits are free, so a realm tags every
+/// handle it mints and the tag travels back into Rust with the handle. This split leaves each
+/// realm 2^48 handles and allows 2^16 realms.
+pub const UNIFFI_REALM_SHIFT: u32 = 48;
+
+/// Identifier of the realm that minted `handle`.
+///
+/// An untagged handle yields 0, so bindings that do not tag keep working unchanged: they
+/// register into realm 0 and read back from realm 0.
+pub const fn uniffi_realm_of_handle(handle: u64) -> u64 {
+    handle >> UNIFFI_REALM_SHIFT
+}
+
+/// Cell type that stores one `NonNull<T>` per realm.
+///
+/// Registration happens once per realm at startup, so the realm list is tiny and effectively
+/// write-once. A linear scan of it costs less than hashing, and is dwarfed in any case by the
+/// foreign call that every lookup precedes.
 #[doc(hidden)]
-pub struct UniffiForeignPointerCell<T>(AtomicPtr<T>);
+pub struct UniffiForeignPointerCell<T>(RwLock<Vec<(u64, NonNull<T>)>>);
 
 impl<T> UniffiForeignPointerCell<T> {
     pub const fn new() -> Self {
-        Self(AtomicPtr::new(null_mut()))
+        Self(RwLock::new(Vec::new()))
+    }
+
+    /// Store the pointer registered by `realm`, replacing any previous registration for it.
+    pub fn set_for_realm(&self, realm: u64, callback: NonNull<T>) {
+        let mut realms = self.0.write().unwrap_or_else(PoisonError::into_inner);
+        match realms.iter_mut().find(|(id, _)| *id == realm) {
+            Some((_, registered)) => *registered = callback,
+            None => realms.push((realm, callback)),
+        }
     }
 
     pub fn set(&self, callback: NonNull<T>) {
-        self.0.store(callback.as_ptr(), Ordering::Relaxed);
+        self.set_for_realm(0, callback);
+    }
+
+    /// Look up the pointer registered by the realm that minted `handle`.
+    ///
+    /// Falls back to realm 0 when that realm registered nothing, which is what a handle minted
+    /// by bindings that do not tag looks like.
+    pub fn get_for_handle(&self, handle: u64) -> &T {
+        let realm = uniffi_realm_of_handle(handle);
+        let realms = self.0.read().unwrap_or_else(PoisonError::into_inner);
+        let registered = realms
+            .iter()
+            .find(|(id, _)| *id == realm)
+            .or_else(|| realms.iter().find(|(id, _)| *id == 0))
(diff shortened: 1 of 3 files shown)


uniffi_core/src/ffi/foreigncallbacks.rs
```diff
@@ -9,30 +9,82 @@
...
+/// register into realm 0 and read back from realm 0.
+pub const fn uniffi_realm_of_handle(handle: u64) -> u64 {
+    handle >> UNIFFI_REALM_SHIFT
+}
+
+/// Cell type that stores one `NonNull<T>` per realm.
+///
+/// Registration happens once per realm at startup, so the realm list is tiny and effectively
+/// write-once. A linear scan of it costs less than hashing, and is dwarfed in any case by the
+/// foreign call that every lookup precedes.
```
bendk, line 40: How do you know this is true?  The cost of a foreign call differs widely by language.

uniffi_core/src/ffi/foreigncallbacks.rs
```diff
@@ -9,30 +9,82 @@
...
     }
 
-    pub fn get(&self) -> &T {
+    /// Look up the pointer registered by the realm that minted `handle`.
+    ///
+    /// Falls back to realm 0 when that realm registered nothing, which is what a handle minted
+    /// by bindings that do not tag looks like.
+    pub fn get_for_handle(&self, handle: u64) -> &T {
+        let realm = uniffi_realm_of_handle(handle);
+        let realms = self.0.read().unwrap_or_else(PoisonError::into_inner);
```
bendk, line 68: What does `PoisenError::into_inner` do here?

uniffi_core/src/ffi/foreigncallbacks.rs
```diff
@@ -9,30 +9,82 @@
...
+        let registered = realms
+            .iter()
+            .find(|(id, _)| *id == realm)
+            .or_else(|| realms.iter().find(|(id, _)| *id == 0))
```
bendk, line 72: Why this fallback?

bendk (changes requested):
This seems like a nice idea to me, but I'm confused about the details.  Also, how would this work on the bindings side?  For example, how would they know which realm to send?

Also, I find it a bit weird that we would be using both the lower and upper bits to tag the handles.  Would it be possible to just one one or the other?

PR #2962, opened 2026-08-09
fix(python): prefer the running event loop

### LINKED ISSUE #2961: Python: a call from a second event loop gets a Future from the wrong loop
**What I was doing**

Calling a uniffi library from a process that runs more than one asyncio loop, having called `uniffi_set_event_loop` as the docs say.

**What happens**

The first call made from any loop other than the one passed to `uniffi_set_event_loop` fails:

```
RuntimeError: Task <...> got Future <Future pending> attached to a different loop
```

**Why**

`_uniffi_get_event_loop` prefers the global over the loop actually running the call:

```python
def _uniffi_get_event_loop():
    if _UNIFFI_GLOBAL_EVENT_LOOP is not None:
        return _UNIFFI_GLOBAL_EVENT_LOOP
    else:
        return asyncio.get_running_loop()
```

`_uniffi_rust_call_async` uses that to `create_future()` and to receive the poll continuations, so the future is built on the loop that was set and awaited from the loop that made the call.

```mermaid
sequenceDiagram
    participant B as Loop B (making the call)
    participant Lookup as _uniffi_get_event_loop
    participant A as Loop A (passed to uniffi_set_event_loop)

    Note over A: uniffi_set_event_loop(A)
    B->>Lookup: which loop?
    Lookup-->>B: A, because the global is set
    B->>A: create_future()
    A-->>B: a Future belonging to A
    B->>B: await that Future
    Note over B: RuntimeError: got Future attached to a different loop
```

`uniffi_set_event_loop` exists so a thread with no loop of its own has one to schedule a callback on. `_uniffi_rust_call_async` is a coroutine, so it always has a running loop and never needs that fallback.

**Repro**

```python
import asyncio, futures

first = asyncio.new_event_loop()
futures.uniffi_set_event_loop(first)

async def main():
    await futures.always_ready()

asyncio.run(main())   # a second loop, and this raises
```

**Fix**

Prefer the running loop, keep the global as the fallback for threads that have none. PR to follow.

Affects 0.31.0 through 0.32.0 and `main`; `_uniffi_get_event_loop` is unchanged across all of them.

stupside:
Fixes #2961.

**What I was doing**

Calling a uniffi library from a process running more than one asyncio loop, having called `uniffi_set_event_loop` as the docs say.

**What happens**

Any call made from a loop other than the one passed to `uniffi_set_event_loop` raises `got Future attached to a different loop`.

**Why**

`_uniffi_get_event_loop` prefers the global over the loop running the call, and `_uniffi_rust_call_async` uses it both to `create_future()` and to receive the poll continuations.

```mermaid
sequenceDiagram
    participant B as Loop B (making the call)
    participant Lookup as _uniffi_get_event_loop
    participant A as Loop A (set globally)

    B->>Lookup: which loop?
    Lookup-->>B: A, because the global is set
    B->>A: create_future()
    A-->>B: a Future belonging to A
    B->>B: await it
    Note over B: RuntimeError
```

**The fix**

```diff
 def _uniffi_get_event_loop():
-    if _UNIFFI_GLOBAL_EVENT_LOOP is not None:
-        return _UNIFFI_GLOBAL_EVENT_LOOP
-    else:
-        return asyncio.get_running_loop()
+    try:
+        return asyncio.get_running_loop()
+    except RuntimeError:
+        if _UNIFFI_GLOBAL_EVENT_LOOP is None:
+            raise
+        return _UNIFFI_GLOBAL_EVENT_LOOP
```

The global is the fallback for threads with no loop of their own, which is what it was added for. `_uniffi_rust_call_async` is a coroutine, so it always has a running loop and never needs it. Nothing changes for a single loop process, where the running loop is the global.

**Test**

`test_call_from_a_second_event_loop` in the futures fixture. Checked both ways: it fails on this branch with the template change reverted, and passes with it.

Python and Swift fixtures pass here. Kotlin and Ruby I could not run locally (no `kotlinc`, no `ffi` gem); this touches only the Python templates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@
 - Kotlin: Fixed messages for error classes that inherit `Throwable`, but not `Exception`.
 - Fix UDL remote enums, now allowing `[Enum, Remote] interface { ... }`
   (via [#2823](https://github.com/mozilla/uniffi-rs/issues/2823)).
+- Python: event-loop lookup now prefers the calling thread's running loop, falling back to the one
+  passed to `uniffi_set_event_loop`.
 
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.32.0...HEAD).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
   (via [#2823](https://github.com/mozilla/uniffi-rs/issues/2823)).
 - Python: event-loop lookup now prefers the calling thread's running loop, falling back to the one
   passed to `uniffi_set_event_loop`.
+- Python: rename `uniffi_set_event_loop` to `uniffi_set_default_event_loop`
 
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.32.0...HEAD).
 
--- a/uniffi_bindgen/src/bindings/python/templates/Async.py
+++ b/uniffi_bindgen/src/bindings/python/templates/Async.py
@@ -24,10 +24,15 @@ def uniffi_set_event_loop(eventloop: asyncio.BaseEventLoop):
     _UNIFFI_GLOBAL_EVENT_LOOP = eventloop
 
 def _uniffi_get_event_loop():
-    if _UNIFFI_GLOBAL_EVENT_LOOP is not None:
-        return _UNIFFI_GLOBAL_EVENT_LOOP
-    else:
+    # A loop running on this thread owns the futures the call awaits, so it wins. The global is
+    # the fallback for threads that have none, which is what it was added for: preferring it
+    # everywhere hands one loop another loop's future as soon as a process runs two.
+    try:
         return asyncio.get_running_loop()
+    except RuntimeError:
+        if _UNIFFI_GLOBAL_EVENT_LOOP is not None:
+            return _UNIFFI_GLOBAL_EVENT_LOOP
+        raise
 
 # Continuation callback for async functions
 # lift the return value or error and resolve the future, causing the async function to resume.
--- a/uniffi_bindgen/src/bindings/python/templates/Async.py
+++ b/uniffi_bindgen/src/bindings/python/templates/Async.py
@@ -17,9 +17,9 @@ This is needed if some async functions run outside of the eventloop, for example
       like `pollster` to block on the async call.
 
 In this case, we need an event loop to run the Python async function, but there's no eventloop set
-for the thread.  Use `uniffi_set_event_loop` to force an eventloop to be used in this case.
+for the thread.  Use `uniffi_set_default_event_loop` to force an eventloop to be used in this case.
 """
-def uniffi_set_event_loop(eventloop: asyncio.BaseEventLoop):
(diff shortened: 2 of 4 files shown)


mhammond:
This makes sense to be, but I think @bendk has the most context around this.

stupside:
> This makes sense to be, but I think @bendk has the most context around this.

Thank you, happy to help, waiting for more context from @bendk 

Right now the python code we get from Uniffi hangs indefinitely, so we had to apply a patch on our side.

uniffi_bindgen/src/bindings/python/templates/Async.py
```diff
@@ -24,10 +24,15 @@ def uniffi_set_event_loop(eventloop: asyncio.BaseEventLoop):
...
-    else:
+    # A loop running on this thread owns the futures the call awaits, so it wins. The global is
+    # the fallback for threads that have none, which is what it was added for: preferring it
+    # everywhere hands one loop another loop's future as soon as a process runs two.
+    try:
         return asyncio.get_running_loop()
+    except RuntimeError:
+        if _UNIFFI_GLOBAL_EVENT_LOOP is not None:
+            return _UNIFFI_GLOBAL_EVENT_LOOP
+        raise
```
bendk: This logic makes sense to me, but I think a couple other things should change:

- Rename `uniffi_set_event_loop` so that it's clear it's only used for callbacks called from threads without an event loop, i.e. a Rust-managed thread.  I think `uniffi_set_default_event_loop` seems good to me.
- Update the docs.  I think the wording there is fine for the new behavior, maybe the only thing that needs to be updated is the function name.
- Update the exception message to describe what's going on better.  Something like "UniFFI error: no running event loop and uniffi_set_default_event_loop not called".

CHANGELOG.md
```diff
@@ -21,6 +21,11 @@
 - Kotlin: Fixed messages for error classes that inherit `Throwable`, but not `Exception`.
 - Fix UDL remote enums, now allowing `[Enum, Remote] interface { ... }`
   (via [#2823](https://github.com/mozilla/uniffi-rs/issues/2823)).
+- Python: a call made from a second event loop is no longer awaited on the loop passed to
+  `uniffi_set_event_loop`. The lookup now prefers a loop running on the calling thread and falls
+  back to the global one, which is what it was added for; preferring the global everywhere built
+  the poll future on one loop and awaited it from another, raising "got Future attached to a
+  different loop".
```
bendk: Could you make this more concise?

bendk (changes requested):
Thanks for the contribution.  This seems right to me, but I have some change requests for the docs / naming.

stupside:
> Thanks for the contribution. This seems right to me, but I have some change requests for the docs / naming.

Thanks for the review @bendk, I will check it out when I'm back from work.

uniffi_bindgen/src/bindings/python/templates/Async.py
stupside: Thanks for the feedback, I applied you suggestions and renamed `uniffi_set_event_loop` to `uniffi_set_default_event_loop`.

CHANGELOG.md
stupside: Sure, please check it out and let me know if it's better now.

bendk (approved):
Thanks for the update, this looks great to me.

PR #2963, opened 2026-08-10
Fix const_evaluatable_unchecked warning in pointer-ffi async callback scaffolding

mhammond:
Note this is against the kotlin-pointer-ffi branch, and helps explain why app-services is hitting errors trying to bump to rust 1.94. app-services isn't actually on this branch (and earlier version of it maybe?), so it's not clear how easily a re-vendor for this branch will be - but it seems that we are going to want this regardless.

The pointer-FFI scaffolding for async callback interface methods used `ffi_buffer_size!` directly as an array length inside the generated `async` block. `ffi_buffer_size!` expands to an inline `const` block, and in array-length position inside a coroutine body that const inherits the coroutine's generics, so rustc 1.92+ fires `const_evaluatable_unchecked` (rust#76200) even though the size is fully concrete.

Hoist the length into a nested `const` item, which cannot reference the enclosing body's generics and so is not treated as generic-dependent.

### DIFF
--- a/fixtures/futures/src/lib.rs
+++ b/fixtures/futures/src/lib.rs
@@ -2,6 +2,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+// Regression test for the async callback interface scaffolding; we want to avoid
+// needing `const_evaluatable_unchecked`, especially given that capability
+// is going to be removed by rustc eventually - see rust#76200.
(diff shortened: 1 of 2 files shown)


bendk (approved):

bendk:
Nice catch, thanks for the patch.  I'm not sure why CI is red all of a sudden though, any thoughts?

mhammond:
I was a little confused, this branch should die and will be able to one we bump uniffi in a-s.

PR #2964, opened 2026-08-10
Refactor by_ref and by_mut_ref into a single PassBy enum

### LINKED ISSUE #2954: Replace by_mut_ref and by_ref with an enum
#2940 added a second bool to several structs (e.g. `FnParamMetadata`, `*::ArgType`, `*::Argument`), so four states can be represented, but only three of these are legal.

This issue is to replace both bools with a three variant enum.

> I believe it should be an enum, as currently it's possible to represent an "impossible" state iiuc. I'm not too bothered if it happens in this or a followup, but I'm really trying to avoid a followup being filed and never implemented. Your agent can probably do this easily and an additional commit for ease of review might allay the reviewability concern. otoh, if you commit to doing this in a fast follow I'm perfectly fine.

_Originally posted by @mhammond in https://github.com/mozilla/uniffi-rs/issues/2940#issuecomment-5186093188_

jhugman:
Fixes #2954.


mhammond:
That looks great, thanks! CI failures are odd and very noisy, but it looks like a failure is:

```
======================================================================
FAIL: test_annotations (__main__.Annotations.test_annotations)
----------------------------------------------------------------------
  File "/home/circleci/project/fixtures/coverall/tests/bindings/test_coverall.py", line 527, in test_annotations
    self.assertEqual(get_type_hints(println), {"text": str, "return": type(None)})
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: {'text': <class 'coverall._UniffiFfiConverterString'>, 're[21 chars]pe'>} != {'text': <class 'str'>, 'return': <class 'NoneType'>}
+ {'return': <class 'NoneType'>, 'text': <class 'str'>}
- {'return': <class 'NoneType'>,
-  'text': <class 'coverall._UniffiFfiConverterString'>}

----------------------------------------------------------------------
Ran 25 tests in 0.170s

FAILED (failures=1)
test uniffi_foreign_language_testcase_test_coverall_py ... FAILED

```

uniffi_bindgen/src/bindings/python/pipeline/nodes.rs
```diff
@@ -222,12 +224,12 @@ impl Argument {
...
-        } else if self.by_ref && matches!(self.ty.ty, Type::Bytes) {
-            "typing.Union[bytes, bytearray]".to_string()
-        } else {
-            self.ty.type_name.clone()
+        match self.pass_by {
+            PassBy::MutRef if matches!(self.ty.ty, Type::Bytes) => "bytearray".to_string(),
+            PassBy::Ref if matches!(self.ty.ty, Type::Bytes) => {
+                "typing.Union[bytes, bytearray]".to_string()
+            }
+            _ => self.ty.type_name.clone(),
```
jhugman, line 232: CI failed here. Fixed now.

mhammond (approved):
that's great, thanks!

PR #2965, opened 2026-08-15
Remove library mode.

mhammond:
All deprecated in v0.31.0 and all unused internally. Removes an entire module named literally 'library_mode` and a chunk of other confusing code and public exports we moved away from a long time ago.

Doesn't remove the (already ignored) `--library` cli option.


docs/manual/src/proc_macro/index.md
```diff
@@ -23,7 +23,7 @@ true for all of UniFFI, so proceed with caution and the knowledge that things ma
 
 ## Build workflow
 
-Be sure to use library mode when using UniFFI proc-macros (See the [Foreign language bindings docs](../tutorial/foreign_language_bindings.md) for more info).
+Be sure to pass a cdylib path when generating bindings for crates that use UniFFI proc-macros (See the [Foreign language bindings docs](../tutorial/foreign_language_bindings.md) for more info).
```
bendk: uniffi_parse_rs also works and technically so does a static library.  Maybe this could be something like "pass `path-to-library-file` or `src:[crate-name]` when generating bindings for crates that use UniFFI proc-macros.  Passing a path to a UDL file won't work".

docs/manual/src/tutorial/foreign_language_bindings.md
```diff
@@ -47,22 +47,20 @@ Then you can run `uniffi-bindgen` from any crate in your project using `cargo ru
 
 ## Running uniffi-bindgen using a library file
 
-Use `generate --library` to generate foreign bindings by using a cdylib file built for your library.
-This can be more convenient than specifying the UDL file -- especially when multiple UniFFI-ed crates are built together in one library.
-This should be used where possible - some "external type" features don't work otherwise.
+Pass a cdylib path directly to `generate` to generate foreign bindings from a compiled library.
```
bendk: I think "library path" is more accurate than "cdylib path" here too.  I don't think we need to mention `uniffi_parse_rs` though here until that's more mature though.

bendk (approved):
Thanks for the cleanup, I'm happy to see this go.  I just had a couple docs suggestions.

PR #2966, opened 2026-08-17
Escape /* and */ in Kotlin and Swift docstrings

### LINKED ISSUE #2411: Escape `/*` from rustdoc.
```rs
/// This is a rustdoc that contains /* in it
struct S;
```
That will open a new comment block in the generated languages (swift, kotlin) thus generating incorrect code.

liamiepops:
Fixes #2411.

Docstrings are emitted verbatim into a `/** ... */` block, so comment markers in the text change the structure of the generated code. A `*/` closes the block early and the rest of the line is parsed as code. A `/*` opens a block that is never closed, which swallows everything after it. Both cases produce Kotlin and Swift that doesn't compile.

Nested `/* ... */` is fine, since both languages allow nested block comments. Only the unbalanced cases break. Python is unaffected because it uses `"""`.

This escapes `*/` as `*\/` and `/*` as `/\*` in the Kotlin and Swift docstring filters.

Adds two functions to the `docstring` and `docstring-proc-macro` fixtures, one with a `/*` in the docstring and one with a `*/`. Without the fix, `kotlinc` and `swiftc` both reject the generated bindings. The Rust assertions on docstring content still pass in that state, since the text is present, just in the wrong place syntactically.

The TypeScript backend in `uniffi-bindgen-react-native` uses the same filter, where any `*/` breaks the output since TypeScript has no nested block comments. Fixed there in jhugman/uniffi-bindgen-react-native#438.

Ran the full suite in the Docker image. 226 test result lines, all passing.

### DIFF
--- a/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
+++ b/uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
@@ -1058,7 +1058,8 @@ mod filters {
         _: &dyn askama::Values,
         spaces: &i32,
     ) -> Result<String, askama::Error> {
-        let middle = textwrap::indent(&textwrap::dedent(docstring.as_ref()), " * ");
+        let escaped = docstring.as_ref().replace("*/", "*\\/").replace("/*", "/\\*");
+        let middle = textwrap::indent(&textwrap::dedent(&escaped), " * ");
         let wrapped = format!("/**\n{middle}\n */");
 
         let spaces = usize::try_from(*spaces).unwrap_or_default();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@
 - Python: event-loop lookup now prefers the calling thread's running loop, falling back to the one
   passed to `uniffi_set_event_loop`.
 - Python: rename `uniffi_set_event_loop` to `uniffi_set_default_event_loop`
+- Kotlin and Swift: escape `/*` and `*/` in docstrings, which previously generated code that would not compile
+  (via [#2411](https://github.com/mozilla/uniffi-rs/issues/2411)).
 
 [All changes in [[UnreleasedUniFFIVersion]]](https://github.com/mozilla/uniffi-rs/compare/v0.32.0...HEAD).
 
--- a/fixtures/docstring-proc-macro/src/lib.rs
+++ b/fixtures/docstring-proc-macro/src/lib.rs
@@ -127,4 +127,12 @@ uniffi::custom_type!(
 #[uniffi::export]
 pub fn test_long_docstring() {}
 
+/// <docstring-comment-open> contains a /* in the text
+#[uniffi::export]
+pub fn test_comment_open() {}
+
+/// <docstring-comment-close> contains a */ in the text
+#[uniffi::export]
+pub fn test_comment_close() {}
+
 uniffi::include_scaffolding!("docstring-proc-macro");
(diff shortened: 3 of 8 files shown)


bendk:
> Nested /* ... */ is fine, since both languages allow nested block comments. Only the unbalanced cases break. Python is unaffected because it uses """.

Side note since this isn't part of the commit, but I'm not sure any of these claims are true.  Is this LLM-generated?

uniffi_bindgen/src/bindings/kotlin/gen_kotlin/mod.rs
```diff
@@ -1058,7 +1058,8 @@ mod filters {
         _: &dyn askama::Values,
         spaces: &i32,
     ) -> Result<String, askama::Error> {
-        let middle = textwrap::indent(&textwrap::dedent(docstring.as_ref()), " * ");
+        let escaped = docstring.as_ref().replace("*/", "*\\/").replace("/*", "/\\*");
```
bendk, line 1061: Why escape the second character?  It reads messy to me, but I don't have a great suggestion either.  Maybe the empty string or adding a space in-between the two chars.

bendk (commented):
The changes look good to me.  Maybe we could bikeshed a bit on what the best replacement is, but I could easily be convinced to go with the current solution.  What's your thoughts on the best replacement?

bendk:
> > Nested /* ... */ is fine, since both languages allow nested block comments. Only the unbalanced cases break. Python is unaffected because it uses """.
> 
> Side note since this isn't part of the commit, but I'm not sure any of these claims are true. Is this LLM-generated?

Thinking this through more, isn't the only issue the `*/` sequence?  Maybe that's the only one that needs to be escaped.

liamiepops:
> > > Nested /* ... */ is fine, since both languages allow nested block comments. Only the unbalanced cases break. Python is unaffected because it uses """.
> > 
> > 
> > Side note since this isn't part of the commit, but I'm not sure any of these claims are true. Is this LLM-generated?
> 
> Thinking this through more, isn't the only issue the `*/` sequence? Maybe that's the only one that needs to be escaped.

Tested this. Escaping only `*/` isn't enough. The `/*` case still fails in both languages.

Swift:

```
uniffi_docstring_proc_macro.swift:1488:24: error: unterminated '/*' comment
1387 | /**
     | `- note: comment started here
...
1488 | // swiftlint:enable all
     |                        `- error: unterminated '/*' comment
```

Kotlin:

```
uniffi_docstring_proc_macro.kt:2008:1: error: syntax error: Unclosed comment.
```

That also answers your first question. The `/*` in the docstring opens a nested comment, the closing `*/` the formatter emits closes that one instead of the outer block, and the outer block runs to the end of the file. It only behaves that way because both languages allow nested block comments.

On the replacement, I don't have a strong preference. The Python filter already escapes `"""` as `\"\"\"`, so backslash-escaping seemed like the consistent choice, but a space between the characters would work well too. I'm happy to change it to whatever you'd prefer.

On the LLM question, yes, I use them in my toolchain. In this case I've confirmed the claims hold.

bendk:
Thanks for the explanation.  TIL about nested comments in Swift and Kotlin.  I think I was confused by the "Python is unaffected..." part.  When I first read that I thought you were saying Python doesn't need escaping.  Re-reading it now, I think I should have taken that as it's not part of this patch because it uses a different comment style, which makes sense.

Slashes seem okay to me and I realize that of course you need put the slash before the second character.  You're not escaping in the traditional sense, you just need to put some character in-between the `/` and `*`.

bendk (approved):
Thanks for the patch and explanation, let's merge.

Issue #2823, opened 2026-02-16
Remote `interface`s don't work in UDL


tnull:
In our code we have a few instances of types that are defined in remote crates which do not know about UniFFI at all. So far, we simply defined these types as everything else in our local UDL file and used `UniffiCustomTypeConverter`s for the implementation part which worked seamlessly.

We're now trying to upgrade from 0.28 to 0.28, need to apply the new `uniffi::custom_type!` approach, migrating away from `UniffiCustomTypeConverter`, which however doesn't seem to work for remote `Enum` `interfaces`  and `Traits`. For example, these definitions where we simply added `Remote` (as in other places) don't work for us:

```
[Remote, Enum]
interface ClosureReason {
...
```
and
```
[Remote, Trait]
interface VssHeaderProvider {
...
}
```

respectively leading to

```
  Caused by:
      conflicting attributes on interface definition
```

and 

```
error[E0210]: type parameter `T` must be used as the type parameter for some local type (e.g., `MyStruct<T>`)
    --> /home/tnull/workspace/ldk-node/target/debug/build/ldk-node-c4e52de90d03d02f/out/ldk_node.uniffi.rs:3123:1
     |
3123 | #[::uniffi::export_for_udl]
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^ type parameter `T` must be used as the type parameter for some local type
     |
     = note: implementing a foreign trait is only possible if at least one of the types for which it is implemented is local
     = note: only traits defined in the current crate can be implemented for a type parameter
     = note: this error originates in the attribute macro `::uniffi::export_for_udl` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0210`.
error: could not compile `ldk-node` (lib) due to 11 previous errors
exit 101
```

We had already reported this a few weeks back on Matrix and there it was concluded that this likely is a bug we should open an issue for (excuse the delay here!).

While we eventually def. want to upgrade to newer versions, it would be great if a fix could be backported to v0.29 as we need to take the updates step-by-step and v0.29 is the one we're currently stuck on. :)

mhammond:
[...]

PR #2967, opened 2026-08-19
Support rust-only remote traits.

mhammond:
indirectly via #2823, fixes a `// TODO` and fits in fine.

rust-only limitation could probably be overcome later if someone cared enough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
   Benchmarks show large improvements in performance and we believe this will be more stable than the
   current Kotlin bindings.  However, the code is very fresh so use at your own risk.
   See [#2911](https://github.com/mozilla/uniffi-rs/pull/2911).
+- Added support for remote trait interfaces - ie, traits defined in a crate which doesn't use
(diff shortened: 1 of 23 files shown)


bendk (approved):
This is very cool.  I hadn't thought about this at all but I can definitely see it being used.

I wonder if it would be possible to support foreign impls at some point.  I think it would mean doing some unsafe/messy stuff with the pointer metadata, but maybe it's possible.

mhammond:
> I wonder if it would be possible to support foreign impls at some point.

I think we really just would need to stop adding `uniffi_foreign_handle()` to the trait, replacing it with a per-trait global function maybe?

bendk:
> > I wonder if it would be possible to support foreign impls at some point.
> 
> I think we really just would need to stop adding `uniffi_foreign_handle()` to the trait, replacing it with a per-trait global function maybe?

Yes something like that, but that global function would have to do some dirty stuff to make the distinction.  It inputs an `Arc<dyn Trait>`, how can it tell the difference between a regular Rust impl and the impl we generate for the foreign side?  I guess it would have to split the wide pointer into a pointer and vtable, check if the vtable matches the one for our the struct we generate for foreign impls, then look into the pointer to get the foreign handle.  Every step there feels scary to me, but maybe it's possible or would be if Rust stabilized some more stuff around wide pointers.

PR #2969, opened 2026-08-19
Remove some unused pipeline code

bendk:
In the spirit of https://github.com/mozilla/uniffi-rs/pull/2965, let's also remove this code and leverage `BindgenLoader` instead.  I don't think this needs a changelog entry because the pipeline stuff is still semi-experimental.


mhammond (approved):
🎉

PR #2972, opened 2026-08-27
fix source-parsing pipeline and kotlin-jni bindgen for large real-world crate graphs

flisky:
## What's fixed

Pointing the new Rust-source parsing pipeline (`uniffi_parse_rs`) and the
experimental `uniffi-bindgen-kotlin-jni` at a large existing workspace
(two UniFFI crates, many modules, several non-UniFFI dependencies) hit a
series of issues. Each commit fixes one of them, with tests.

### `uniffi_parse_rs` — path/type resolution

- **Leading `self::` path segments didn't resolve.** `RPath::child()`
  special-cased `super` and `crate` but not `self`, so common
  disambiguating imports like `pub use self::http::*;` (where a module
  shares its name with a crate) failed with "Unknown item".
- **`dyn Trait` references carried the wrong module path.** The
  referencing module's path was used instead of the trait's own public
  path, so a trait object referenced from another module or crate matched
  no definition ("type module path not found").
- **Custom types now resolve by the type they cover, not only by name in
  the module that invoked `custom_type!`.** The macro flow is type-keyed
  (the generated impls apply wherever the type is named), but source
  parsing resolved names lexically, so a type alias in another module, a
  non-UniFFI type, or a direct import from an unparsed crate
  (`use rust_decimal::Decimal` covered by
  `type Decimal = rust_decimal::Decimal;` + `custom_type!(Decimal, String)`)
  failed with "Unknown item". A registry, built once per `LookupCache`,
  maps each custom type's underlying type back to the custom type.
- **`use_remote_type!(implementing_crate::Type)` no longer requires
  `Type` to live at that path.** Matching the macro's semantics (the type
  resolves in the invoking module's scope; the first segment only names
  the crate), resolution falls through to the module's own items instead
  of failing — and shadowing a covered alias no longer reports a spurious
  name conflict.
- **Generic containers exposed through hand-written FFI impls resolve.**
  `custom_type!` can't cover generics; crates expose e.g.
  `indexmap::IndexSet<T>` with hand-written trait impls whose
  `TYPE_ID_META` declares the wire shape. The parser now reads that
  declaration and resolves the target like the equivalent builtin
  container (`TYPE_VEC` / `TYPE_OPTION` / `TYPE_HASH_SET` / `TYPE_HASH_MAP`),
  skipping impls whose `concat` order doesn't match the type's parameters.

### `uniffi_bindgen`

- **The dependency sort silently dropped same-named definitions.** When
  sorting type definitions across namespaces, a type used outside its
  defining namespace has both its real definition and an `External`
  marker under the same canonical name; keying by name alone kept only
  one, breaking downstream consumers (e.g. the kotlin-jni FFI type oracle
  failing with "Can't find FFI types for Custom { … }").

### `uniffi-bindgen-kotlin-jni`

- **`Vec<u8>` fields inside FFI buffers** ("No FFI buffer Layout for
  Bytes"): packed like `String` as (data, length, capacity), with the
  matching JNI bridges and Kotlin wrappers.
- **`Timestamp` / `Duration` fields inside FFI buffers** ("No FFI buffer
  Layout for Duration"): the layout entry matching what the templates
  already generate (8-byte seconds + 4-byte nanoseconds).
- **Container scaffolding is generic over the container type**, so types
  like `IndexSet<T>` declared via `TYPE_ID_META` (see above) compile:
  lowering takes `impl IntoIterator<Item = T>`, lifting returns
  `C: FromIterator<T>`; every call site has a concrete type to infer from.
- **Generated callback impls derive `Clone, Debug`**, required when an exported
  foreign trait has a `Clone`/`Debug` supertrait.

---

Disclaimer: CC Fable model helps me to port our 4-years-old project to use the newest kotlin jni binding.
Please let me know if you need to split this into small PRs.

### DIFF
--- a/uniffi_parse_rs/src/types.rs
+++ b/uniffi_parse_rs/src/types.rs
@@ -385,11 +385,12 @@ impl<'ir> RPath<'ir> {
                     Ok(path) => path,
                     Err(e) if e.is_not_found() => {
                         // The path doesn't name a UniFFI item, but it may name a non-UniFFI
-                        // type that a `uniffi::custom_type!` elsewhere covers.
+                        // type that a `uniffi::custom_type!` or a hand-written `TypeId` impl
+                        // elsewhere covers.
                         match self
                             .resolve(ir, cache, &ty_path.path, Namespace::NonUniffiType)
                             .ok()
-                            .and_then(|path| cache.registered_custom_type(&path))
+                            .and_then(|path| cache.registered_type_mapping(&path))
                         {
                             Some(custom_type_path) => custom_type_path,
                             None => return Err(e),
@@ -402,7 +403,7 @@ impl<'ir> RPath<'ir> {
                 // registered for the item this path resolves to (e.g. a type alias in another
                 // module), resolve to the custom type instead.
                 if matches!(path_to_type.item()?, Item::Type(_) | Item::NonUniffi(..)) {
-                    if let Some(custom_type_path) = cache.registered_custom_type(&path_to_type) {
+                    if let Some(custom_type_path) = cache.registered_type_mapping(&path_to_type) {
                         path_to_type = custom_type_path;
                     }
                 }
@@ -1093,6 +1094,51 @@ pub mod tests {
                 &mut cache,
                 "custom_type_paths",
                 "external_crate2::Other"
+            ),
+            Err(ErrorKind::NotFound),
+        );
+    }
+
+    #[test]
+    fn test_resolve_types_with_hand_written_type_id_impls() {
+        let ir = Ir::new_for_test(&["type_id_impls"]);
+        let mut cache = LookupCache::new(&ir);
+
+        // `IndexSet<T>` declares `TYPE_HASH_SET` in its hand-written `TypeId` impl
+        assert_eq!(
+            run_resolve_type(&ir, &mut cache, "type_id_impls::usage", "IndexSet<u32>"),
+            Ok(Type::HashSet(Box::new(Type::UInt32))),
+        );
+        // `IndexMap<K, V>` declares `TYPE_HASH_MAP`
+        assert_eq!(
+            run_resolve_type(
+                &ir,
+                &mut cache,
+                "type_id_impls::usage",
+                "IndexMap<String, u32>"
+            ),
+            Ok(Type::HashMap(
+                Box::new(Type::String),
+                Box::new(Type::UInt32)
+            )),
+        );
+        // The same types, named by full path
+        assert_eq!(
+            run_resolve_type(
+                &ir,
+                &mut cache,
+                "type_id_impls",
+                "fake_indexmap::IndexSet<u32>"
+            ),
+            Ok(Type::HashSet(Box::new(Type::UInt32))),
+        );
+        // `Reversed<K, V>` concatenates its params in the wrong order: not registered
+        assert_eq!(
+            run_resolve_type(
+                &ir,
+                &mut cache,
+                "type_id_impls",
+                "fake_indexmap::Reversed<String, u32>"
             ),
             Err(ErrorKind::NotFound),
         );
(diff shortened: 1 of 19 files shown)


uniffi_parse_rs/src/types.rs
```diff
@@ -516,11 +516,17 @@ impl<'ir> RPath<'ir> {
...
+                    Item::Trait(tr) => {
+                        // Use the trait's own public path, not the referencing module's —
+                        // the reference must match the metadata generated for the trait's
+                        // definition, which may live in another module or crate.
+                        let names = trait_path.public_path_to_item(ir, cache)?;
+                        Ok(Type::Trait {
+                            module_path: names.module_path,
+                            name: names.name,
+                            export_ty: tr.attrs.export_ty,
+                        })
```
bendk, line 528: This change looks good, but I'm not sure about the comment.  It felt pretty wordy to me and I needed to read the test examples to figure out the point of this.  I notice that the other usages don't have any comments, maybe we could just remove this?

uniffi_parse_rs/src/types.rs
```diff
@@ -1127,6 +1133,16 @@ pub mod tests {
                 export_ty: TraitExportType::TraitInterface(TraitKind::RustOnly),
             })
         );
+        // Referenced from another module, the type must carry the trait's own module path,
+        // matching the metadata generated for the trait's definition
```
bendk, line 1137: This one could also probably be deleted.  Introducing concepts like the "referencing module" and "carrying a module path" feels counterproductive to me.

IMO, the best way to explain this would be to name the params somehow.  Maybe it could look something like this:

```
ResolveTypeForTests {
   source_module: "types",
   type_string: "dyn mod1::TraitInterface",
   expected: Ok(Type::Trait {
         module_path: "types::mod1".into(),
         name: "TraitInterface".into(),
          export_ty: TraitExportType::TraitInterface(TraitKind::RustOnly),
   }),
   ir: &ir,
  cache: &mut cache,
}.run();
```

But that's probably a separate PR.

uniffi_parse_rs/src/types.rs
```diff
@@ -1594,7 +1610,8 @@ pub mod tests {
                 None
             ),
             Ok(uniffi_meta::Type::Object {
-                module_path: "types".into(),
+                // The trait's own module path, not the referencing module's
```
bendk, line 1613:
```suggestion
```

uniffi_parse_rs/src/types.rs
```diff
@@ -1650,7 +1667,8 @@ pub mod tests {
             run_resolve_arg(&ir, &mut cache, "types", "&dyn mod1::TraitInterface", None),
             Ok(ArgType {
                 ty: uniffi_meta::Type::Object {
-                    module_path: "types".into(),
+                    // The trait's own module path, not the referencing module's
```
bendk, line 1670:
```suggestion
```

bendk (changes requested):
Sorry it took me a while to review this patch, I've was on PTO for a bit then playing catch up.  Also, thanks for generating this, there's a lot of good stuff here.

However, I found it hard to review it because there was just so much there and a lot of it was AI generated. This made it hard for me to follow -- what does "covering a type" mean?  I went through the first 2 commits, but when I got to the 3rd one on custom types I started loosing the thread.

I'd love to get this merged, but the first step is probably getting a better workflow for these changes.  Could you split this into 10 PRs, one for each commit?  That way we can go through the changes 1 by 1 and merge the non-controversial things earlier.

Please try to make each commit have a clear and specific example that it's fixing.  For example, I'm pretty sure [this example](https://github.com/mozilla/uniffi-rs/pull/2972/changes/4be148e5b2345626153abada63c930815e18da11#diff-604f20d387089a24d75de73cb6030ab66e3463ce5ff9d68809f89f0ee1d7b638R18-R22) works already which makes me unclear on which ones the commit is actually fixing.  If possible, find an existing test module and add new test there that show what the commit is fixing.

Lastly, the comments tended to confuse me more then anything.  I think it's better to avoid LLM-generated ones in most cases.  Even better, I'd love a comment or two in your own words that explains what the code is doing.

flisky:
Huge thanks to @bendk, and it only remains custom types #3000 & container types #3002 to us, and we can accept the restriction about the remote types.

PR #2975, opened 2026-08-29
Add regression test

jplatte:
Hey @bendk, can you take a look at this? I found a regression in a commit you wrote (d49730506ae93ac0b89ff6cc98f8bdfe15c14305).

### DIFF
--- a/bindgen-tests/lib/src/custom_type_arc.rs
+++ b/bindgen-tests/lib/src/custom_type_arc.rs
@@ -1,0 +1,18 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Regression test for a custom type pattern that works with 0.32.0, but no
+// longer works as of d49730506ae93ac0b89ff6cc98f8bdfe15c14305.
+
+use std::sync::Arc;
+
+#[derive(uniffi::Object)]
+pub struct Foo(u32);
+
+pub struct Bar(u32);
+
+uniffi::custom_type!(Bar, Arc<Foo>, {
+    try_lift: |val| Ok(Bar(val.0)),
+    lower: |custom| Arc::new(Foo(custom.0)),
+});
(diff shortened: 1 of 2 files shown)


bendk:
I can repro this one too and it seems like an issue with the Rust trait solver.  The first error is `the trait bound 'Foo: FfiConverterArc<UniFfiTag>' is not satisfied`, but when you expand the `#[derive(uniffi::Object)]` macro for `Foo`, you do see an impl:

```
unsafe impl<UT> ::uniffi::FfiConverterArc<UT> for Foo { 
 ...
}
```

It seems like the generics plus the blanket impl is somehow confusing rustc.

I don't know of a great short term solution.  It's possible that we just need to fiddle with the bounds or something and it'll work.  I feel like I've been saying this a bunch, but this is one more reason that I like the uniffi-bindgen-kotlin-jni approach.  Instead of implementing traits, we [define the lift/lower functions](https://github.com/mozilla/uniffi-rs/blob/7638c169b2c663cbceebc87dc8f41f391fe6c8f7/uniffi-bindgen-kotlin-jni/src/scaffolding/templates/class.rs#L9-L15) we need then [generate calls to those functions directly](https://github.com/mozilla/uniffi-rs/blob/7638c169b2c663cbceebc87dc8f41f391fe6c8f7/uniffi-bindgen-kotlin-jni/src/scaffolding/templates/custom.rs#L9).  I'm pretty we wouldn't have that same issue here and if we did it would be more tractable.

jplatte:
I tried `cargo expand` and I did **not** see that impl. In fact I saw no output for the derive at all, which had me confused for a sec. But then I dug a bit and I'm pretty sure I found the issue.

This is specifically about what happens when the `scaffolding` cargo feature is not enabled. It used to be that the derives *and* the `custom_type!` macro would output nothing without it. So you could write `custom_type!` invocations that depend on the output of the proc-macros. Your commit broke that with [this](https://github.com/mozilla/uniffi-rs/commit/d49730506ae93ac0b89ff6cc98f8bdfe15c14305#diff-7edf8005ee0acbfc72d5a08cc22369caf4574726f34df0f90b400a39d0ebd196) change.

bendk:
Sorry about that, you're right.  I used LSP expand macro and didn't get the right results.  I was able to repro and made https://github.com/mozilla/uniffi-rs/pull/3003 for this.

jplatte:
Thanks! Closing this then :)

PR #2976, opened 2026-09-03
Adding CHANGELOG entry for #2935

bendk:


bendk:
Actually, nevermind about this one.  I'm hoping to release a 0.32.1 with this fix, so we'll get the changelog entry when we merge the changes back from that branch into main.

PR #2980, opened 2026-09-04
Ruby: add external/remote types support

saks:
Library-mode only (`generate --library`). Single-UDL with an external type is a bindgen error -- we don't have the other crates' metadata, so the generated require would be nonsense.

Python/Kotlin already have this implementation, and it's small there because they prefix an existing `FfiConverter`. Ruby didn't have such infra. Nested `read/write` were methods on one builder (`write_TypeFoo`), mixed in with include, so two crates that both export `Foo` clobber each other and we couldn't even point at the defining crate. `lift/lower` were `Askama` filters with no crate context. That's why this PR looks like a huge rewrite.

What we generate now:
- `require` the other crate's .rb
- `RustBuffer.alloc/free` stay on the consumer cdylib
- `read_*`/`write_*` are module functions on the defining crate
- custom types call that crate's `uniffi_lift_*` / `uniffi_lower_*` (`Url` → `URI`, identity newtypes still go through `uniffi_in_range`)
- errors: uses method object for the defining crate's reader, so a corrupt buffer raises their `InternalError`

`[bindings.ruby.module_name]` sets the generated module name. Library mode fills `[bindings.ruby.external_packages]` from each peer's module_name; a mismatch is a bindgen error. Keys are crate names (`uniffi_one`, not `uniffi_one_ns`); hyphens and underscores are the same. `require` still uses the namespace / filename.

Most of the template churn (`ObjectTemplate`, `Async`, `Helpers`, etc) is just `self.lift_rb` / `self.module_name` after those moved onto `RubyWrapper`. The actual serialization change is `RustBuffer*.rb` + `wrapper.rb` + `CustomTypeTemplate.rb` + `macros.rb`. `gen_ruby/mod.rs` has the owner-policy comment if the four different "which crate?" helpers are confusing.

`fixtures/ext-types/lib` is the happy path. `transitive-lib` is the one that matters for mixin dispatch: it only depends on sub-lib, so it can't accidentally `require` `uniffi-one`. `bindgen-tests/ruby/tests/external_types.rb` checks imported identity customs still `range-check`. `gen_ruby/tests.rs` is mostly “did we include a foreign mixin again.”

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@
   See [#2911](https://github.com/mozilla/uniffi-rs/pull/2911).
 - Added support for remote trait interfaces - ie, traits defined in a crate which doesn't use
   UniFFI. Use `#[uniffi::export(remote)]` or `[Trait, Remote]` in UDL. Foreign implementations are not supported, see the docs for more.
+- Ruby: Add `bindings.ruby.module_name` to set the generated module name. In library mode,
+  `[bindings.ruby.external_packages]` is auto-filled from each peer's `module_name`; a consumer
+  override that does not match is a bindgen error (via [#2980](https://github.com/mozilla/uniffi-rs/pull/2980)).
+- Ruby: Add external types support (via [#2980](https://github.com/mozilla/uniffi-rs/pull/2980)).
 
 ### What's Fixed
 - Kotlin: Fixed messages for error classes that inherit `Throwable`, but not `Exception`.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1900,6 +1900,7 @@ dependencies = [
  "thiserror",
  "uniffi",
  "uniffi-bindgen-tests-external-types-source",
+ "uniffi-bindgen-tests-mid-types",
  "url",
 ]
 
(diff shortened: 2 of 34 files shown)


bendk:
This looks very cool and also huge.  It's mostly only changing the Ruby codegen though, so I don't feel like I need to review that thoroughly.  In my mind Ruby is basically an external bindgen that happens to be in tree because it was merged early.  That said, I'm also happy to review that code if you're looking for that.  What's you're preference here?

One thing I noticed is that you updated the `fixtures/ext-types` tests.  Would it be possible to move those changes to `bindgen-tests`?  We're hoping to move over to those and ditch most of the fixture tests.

saks:
Thank you @bendk, I'll need some time to see how to move new tests to `bindgen-tests` without breaking `uniffi-bindgen-kotlin-jni` and hopefully reduce the PR side.

saks:
@bendk please have another look, I've removed all the changes from `fixtures` dir and reduced PR size as much as I can. 
As to what kind of review this PR can benefit from: my preference is a light look at the shared pieces — `bindgen-tests` (including the new `mid-types` crate) and the `remote`/`external-types` docs. The Ruby codegen is large because the backend had no FfiConverter to hang imports on; I don’t think it’s a good use of your time to line-review that. Happy to walk through anything that looks off.

PR #2983, opened 2026-09-07
kotlin-jni: support Bytes fields inside FFI buffers

flisky:
fixes `No FFI buffer Layout for Bytes` and also for `Duration`/`Timestamp`.

parts of #2972

### DIFF
--- a/uniffi-bindgen-kotlin-jni/src/pipeline/ffi_buffer.rs
+++ b/uniffi-bindgen-kotlin-jni/src/pipeline/ffi_buffer.rs
@@ -63,6 +63,9 @@ impl FfiBufferLayoutOracle {
             Type::UInt64 | Type::Int64 | Type::Float64 => Layout::from_size_align(8, 8)?,
             // One 8-byte handle
             Type::Interface { .. } => Layout::from_size_align(8, 8)?,
+            // 8-byte seconds at offset 0, 4-byte nanoseconds at offset 8
+            // (see the timestamp/duration scaffolding templates)
+            Type::Timestamp | Type::Duration => Layout::from_size_align(12, 8)?,
             // (data, length, capacity) fields
             Type::String | Type::Bytes => Layout::from_size_align(24, 8)?,
             // (data, size) fields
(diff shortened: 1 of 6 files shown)


uniffi-bindgen-kotlin-jni/src/pipeline/ffi_buffer.rs
```diff
@@ -63,8 +63,11 @@ impl FfiBufferLayoutOracle {
             Type::UInt64 | Type::Int64 | Type::Float64 => Layout::from_size_align(8, 8)?,
             // One 8-byte handle
             Type::Interface { .. } => Layout::from_size_align(8, 8)?,
+            // 8-byte seconds at offset 0, 4-byte nanoseconds at offset 8
+            // (see the timestamp/duration scaffolding templates)
+            Type::Timestamp | Type::Duration => Layout::from_size_align(12, 8)?,
```
bendk, line 68: This feels like it could have been in a separate PR, but it looks accurate so I guess we might as well merge it alongside the rest of the changes.

uniffi-bindgen-kotlin-jni/src/scaffolding/templates/ffi_buffer.rs
```diff
@@ -78,6 +78,42 @@ unsafe extern "system" fn Java_uniffi_Scaffolding_ffiBufferWriteString(
...
+#[unsafe(no_mangle)]
+unsafe extern "system" fn Java_uniffi_Scaffolding_ffiBufferReadBytes(
+    env: *mut uniffi_jni::JNIEnv,
+    _: *mut uniffi_jni::jclass,
+    buf: uniffi_jni::jobject,
+    offset: ::std::primitive::i32,
+) -> uniffi_jni::jbyteArray {
+    unsafe {
+        uniffi_jni::rust_call_with_env(env, |env| {
+            let (ptr, capacity) = uniffi_jni::lift_buffer(env, buf)?;
```
bendk, line 90: I don't love how this reads the capacity then throws it away, but that's the same as how the string packing works so no need to fix it in this PR.  I made https://github.com/mozilla/uniffi-rs/issues/2987 for that.

bendk (approved):
This is great thanks, I'm going to merge it now.

PR #2984, opened 2026-09-07
uniffi_parse_rs: honor use_remote_type!'s local-scope semantics

flisky:
When the `use_remote_type!` path doesn't resolve, fall through to the module's regular items instead of propagating the failure, matching the macro's own semantics: the last segment names the type, resolved in the invoking module's scope, and the first segment only names the crate implementing the FFI traits.

Also skip special items in the regular-item scan — they're handled by `child_special_item` before it runs, and matching them again reported a spurious name conflict between the `use_remote_type!` invocation and the type alias it covers.

part of #2972


flisky:
rebased onto #2990, and will be reopened after the base merged.

PR #2985, opened 2026-09-07
uniffi_bindgen: don't drop same-named definitions in the dependency sort

flisky:
part of #2972

### DIFF
--- a/uniffi_bindgen/src/pipeline/general/sort.rs
+++ b/uniffi_bindgen/src/pipeline/general/sort.rs
@@ -130,6 +130,15 @@ impl FfiDefinitionDependencyLogic {
 
 struct TypeDefinitionDependencyLogic;
 
+impl TypeDefinitionDependencyLogic {
+    fn type_name(ty: &TypeNode) -> String {
+        match ty.ty.namespace() {
+            Some(namespace) => format!("{namespace}.{}", ty.canonical_name),
+            None => ty.canonical_name.clone(),
+        }
+    }
+}
+
 impl DependencyLogic for TypeDefinitionDependencyLogic {
     type Item = TypeDefinition;
 
@@ -145,9 +154,9 @@ impl DependencyLogic for TypeDefinitionDependencyLogic {
             | TypeDefinition::Enum(Enum { self_type, .. })
             | TypeDefinition::Interface(Interface { self_type, .. })
             | TypeDefinition::CallbackInterface(CallbackInterface { self_type, .. })
-            | TypeDefinition::Custom(CustomType { self_type, .. })
-            | TypeDefinition::External(ExternalType { self_type, .. }) => {
-                self_type.canonical_name.clone()
+            | TypeDefinition::Custom(CustomType { self_type, .. }) => Self::type_name(self_type),
+            TypeDefinition::External(ExternalType { self_type, .. }) => {
+                format!("External:{}", Self::type_name(self_type))
             }
         }
     }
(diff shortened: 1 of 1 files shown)


uniffi_bindgen/src/pipeline/general/sort.rs
```diff
@@ -56,16 +60,20 @@ impl<L: DependencyLogic> DependencySorter<L> {
     }
 
     fn recurse(&mut self, current_name: String) {
-        let Some(current_item) = self.unsorted.shift_remove(&current_name) else {
+        let Some(current_items) = self.unsorted.shift_remove(&current_name) else {
```
bendk: I don't think this is correct, because it means we recurse through all definitions that share the name.  For example:

- Type `A` depends on `B`
- There's also an external type named `B`

When processing `A`, this will recurse through both `B` items even though only one is a dependency of `A`.

Instead, I think the fix is to make `L::Item` include both the crate name and item name so that it's unique in this case of the external item.  Ideally it would store the entire module path.  You can do this by either making it a tuple or by keeping it a string and joining all parts together.  If possible, it would be great to have a bindgen-test for this.

bendk (changes requested):

uniffi_bindgen/src/pipeline/general/sort.rs
```diff
@@ -145,9 +154,9 @@ impl DependencyLogic for TypeDefinitionDependencyLogic {
             | TypeDefinition::Enum(Enum { self_type, .. })
             | TypeDefinition::Interface(Interface { self_type, .. })
             | TypeDefinition::CallbackInterface(CallbackInterface { self_type, .. })
-            | TypeDefinition::Custom(CustomType { self_type, .. })
-            | TypeDefinition::External(ExternalType { self_type, .. }) => {
-                self_type.canonical_name.clone()
+            | TypeDefinition::Custom(CustomType { self_type, .. }) => Self::type_name(self_type),
+            TypeDefinition::External(ExternalType { self_type, .. }) => {
+                format!("External:{}", Self::type_name(self_type))
```
bendk, line 159: Do we need the `External:` prefix?  I would have thought the code in `type_name` would be enough to create a unique name.  However, then I looked at the struct and realized that I'm not sure if the `External::namespace` field refers to the source namespace or the namespace of the module that's using it.  External types are really weird.

Can you check if the test passes if this is changed to just call `Self::type_name(self_type),` like the others.  If not, can you write up a comment explaining why we need this?

bendk (changes requested):
This is looking very good, I just wonder why external types need to be treated differently than all the other types.

uniffi_bindgen/src/pipeline/general/sort.rs
flisky: yes, we need this to distinguish external & custom types, and which marker doesn't matter.

There is a paragraph about this in the description generated by LLM:

```
A type used outside its defining namespace also contributes an External
marker alongside its real definition. Those two describe the same type
in the same namespace, so no name derived from the type can separate
them — the marker is named as the marker it is instead. They only meet
when the definitions of several namespaces are sorted together, which
uniffi-bindgen-kotlin-jni does to build its FFI type oracles; there the
two collided and whichever arrived last won. When the marker won, the
definition consumers actually need was gone, and generating bindings
failed with "Can't find FFI types for Custom { ... }" for any custom type
referenced from another crate.
```

flisky: The test `test_external_markers_do_not_replace_real_definitions` covers this behaivor, and doesn't cover the reason.

flisky: Take another way to handle external types. Hope it more clear to follow.

uniffi_bindgen/src/pipeline/general/sort.rs
```diff
@@ -12,8 +12,18 @@ use super::*;
...
     type_definitions: impl IntoIterator<Item = TypeDefinition>,
 ) -> Vec<TypeDefinition> {
-    let type_sorter = DependencySorter::new(type_definitions, TypeDefinitionDependencyLogic);
-    type_sorter.sort()
+    // `TypeDefinition::External` says a type is defined in another namespace, and carries
+    // no definition of its own.  It depends on nothing, and dependents name the type
+    // itself -- which is the real definition's name -- so it adds nothing to the
+    // ordering.  What it would add is a collision: it shares that name, so once the
+    // definitions of several namespaces are sorted together it can overwrite the
+    // definition consumers need.
```
bendk: I liked the previous version better.  [This line](https://github.com/flisky/uniffi-rs/blob/eadccb2a7ffc494f93018d9f3fb1416ddd140256/uniffi_bindgen/src/pipeline/general/sort.rs#L232) communicates that we treat the external type as having no dependencies and it feels easier to maintain than adding this new extra process here.

I think my only concern was the `External:` prefix, which I thought wasn't needed.  If it is indeed needed, then I'd rather take that code over this one.

bendk (changes requested):

uniffi_bindgen/src/pipeline/general/sort.rs
flisky: Restored.

bendk (approved):
Thanks for working this through with me.  I'm still not sure why the `External:` prefix is needed, but it's not a big deal either way since the type will always have 0 dependencies in any case.  Let's merge this one.

PR #2986, opened 2026-09-07
uniffi_parse_rs: use the trait's own path for trait object references

flisky:
A `dyn Trait` reference resolved to a `Type::Trait` carrying the *referencing* module's path, while the metadata generated for the trait's definition carries the trait's own public path.  A reference from another module or crate (e.g. `Arc<dyn some_crate::module::Callback>` in a downstream binding crate) produced a type that matched no definition, failing downstream with "type module path not found".

Use `public_path_to_item` on the resolved trait, the same way the Record/Enum/Object arms already do.

part of #2972

### DIFF
--- a/uniffi_parse_rs/src/types.rs
+++ b/uniffi_parse_rs/src/types.rs
@@ -516,11 +516,17 @@ impl<'ir> RPath<'ir> {
                 }
                 let trait_path = self.resolve(ir, cache, trait_bounds[0], Namespace::Type)?;
                 match trait_path.item()? {
-                    Item::Trait(tr) => Ok(Type::Trait {
-                        module_path: self.path_string(),
-                        name: tr.ident.unraw().to_string(),
-                        export_ty: tr.attrs.export_ty,
-                    }),
+                    Item::Trait(tr) => {
+                        // Use the trait's own public path, not the referencing module's —
+                        // the reference must match the metadata generated for the trait's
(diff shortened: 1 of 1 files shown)


uniffi_parse_rs/src/types.rs
```diff
@@ -1127,6 +1133,16 @@ pub mod tests {
                 export_ty: TraitExportType::TraitInterface(TraitKind::RustOnly),
             })
         );
+        // Referenced from another module, the type must carry the trait's own module path,
+        // matching the metadata generated for the trait's definition
```
bendk: Again, I don't know "the type must carry the trait's own module path" means.  I think I understand "Referenced from another module," but this terminology is not used anywhere else and I'd like to avoid it.

bendk (changes requested):
The code looks great, I don't feel like the comments are useful.  In fact, they seem more confusing then helping them.  I'm assuming this was LLM-generated, tell me if that's not right.  How do you feel about removing the comments?

uniffi_parse_rs/src/types.rs
flisky: Yes, the LLM tends to be wordy.  Actually, it was more wordy in the first generation.

bendk (approved):
Thanks for the quick update, this is looking good now.

PR #2990, opened 2026-09-09
uniffi_parse_rs: resolve custom types by the type they cover

flisky:
Make the routes from the previous commit resolve, by keying custom types
on the type they cover rather than on the module that named them.

Build a crate-graph-wide registry mapping each custom type's underlying
type to the custom type item. For every `custom_type!`, the ident's
non-special binding in its module is resolved (following non-generic type
alias chains) and each step is registered by canonical item path; if the
chain ends in an unparsed crate, the final path is registered textually.

It's called `TypeMappingRegistry` rather than naming custom types: what a
mapping holds is the item that resolving the covered type should reach,
and nothing about the lookup is specific to custom types — a hand-written
`TypeId` impl declares the same kind of mapping, onto a builtin container.
Custom types are the only kind here.

Resolution consults the registry in two places: a type path that only
resolves in the (previously unused) NonUniffiType namespace is accepted
when a custom type covers it, and a global path into an unparsed crate
falls back to the textual registry before erroring. Between them these
cover every route, since the registry holds each step of the alias chain:
a local endpoint resolves through the first, an external one through the
second.

The registry is built eagerly by `LookupCache::new`, which replaces the
derived `Default` — the cache now needs the IR. `Ir::resolve_items` uses
`LookupCache::empty()` instead: the registry can only be built once the
items it maps are resolved, which is what that pass is doing. Building it
resolves paths itself, against a scratch cache whose registry is empty, so
those lookups see no registered custom types and nothing they memoize
survives into the caller's cache.

part of #2972

### DIFF
--- a/uniffi_parse_rs/src/test_src/custom_type_paths.rs
+++ b/uniffi_parse_rs/src/test_src/custom_type_paths.rs
@@ -1,0 +1,52 @@
+// Custom types resolved by the type they cover, not the module the macro was invoked in.
+//
+// The `custom_type!` invocations live in `registrations`; other modules name the
+// underlying types directly (through aliases, non-UniFFI structs, or paths into
+// unparsed crates) and must still resolve to the custom type.
+
+mod types {
+    // A non-UniFFI type covered by a custom type registered in another module
+    pub struct Concrete { }
+
+    // An alias to a type from an unparsed crate, covered by a custom type in another module
+    pub type ExternalAlias = external_crate::ExternalType;
+}
+
+mod registrations {
+    use crate::types::{Concrete, ExternalAlias};
+
+    uniffi::custom_type!(Concrete, String, {
+        remote,
+        into: |obj| obj.to_string(),
+        try_from: |s| s.parse(),
+    });
+
+    uniffi::custom_type!(ExternalAlias, String, {
+        remote,
+        into: |obj| obj.to_string(),
+        try_from: |s| s.parse(),
+    });
+
+    // A custom type over a local alias to an unparsed crate.  Other modules import the
+    // external type directly (`use external_crate2::Direct`) and never see this alias.
+    type Direct = external_crate2::Direct;
+
+    uniffi::custom_type!(Direct, String, {
+        remote,
+        into: |obj| obj.to_string(),
+        try_from: |s| s.parse(),
+    });
+}
+
+mod usage {
+    use external_crate2::Direct;
+
+    use crate::types::{Concrete, ExternalAlias};
+
+    #[derive(uniffi::Record)]
+    pub struct UsesCustomTypes {
+        pub concrete: Concrete,
+        pub aliased: ExternalAlias,
+        pub direct: Direct,
+    }
+}
--- a/uniffi_parse_rs/src/test_src/custom_type_paths2.rs
+++ b/uniffi_parse_rs/src/test_src/custom_type_paths2.rs
@@ -1,0 +1,12 @@
+// Mirrors a binding crate that reuses custom types implemented by `custom_type_paths`:
+// it names the underlying type straight from the unparsed crate, and that must resolve
+// to the custom type registered over in `custom_type_paths`.
+
+mod usage2 {
+    use external_crate2::Direct;
+
+    #[derive(uniffi::Record)]
+    pub struct BindingRecord {
+        pub direct: Direct,
+    }
+}
--- a/uniffi_parse_rs/src/ir.rs
+++ b/uniffi_parse_rs/src/ir.rs
@@ -151,7 +151,7 @@ impl Ir {
     }
 
     pub fn into_metadata_group_map(self) -> Result<MetadataGroupMap> {
-        let mut cache = LookupCache::default();
+        let mut cache = LookupCache::new(&self);
         self.crate_roots_and_paths()
             .map(|(mut module_path, module)| {
                 let mut group = MetadataGroup {
@@ -182,7 +182,9 @@ impl Ir {
 
         // Resolve the items
         let mut resolved_map = HashMap::new();
-        let mut cache = LookupCache::default();
+        // `LookupCache::empty()`, not `new()`: the type mapping registry can only be
+        // built once the items it maps are resolved, which is what's happening here.
+        let mut cache = LookupCache::empty();
         for crate_root in self.crate_roots.values() {
             crate_root.try_visit_modules_and_paths(|module, rpath| {
                 let Some(unresolved) = unresolved_map.remove(&module.id) else {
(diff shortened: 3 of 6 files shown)


uniffi_parse_rs/src/test_src/custom_type_paths.rs
```diff
@@ -0,0 +1,52 @@
+// Custom types resolved by the type they cover, not the module the macro was invoked in.
+//
+// The `custom_type!` invocations live in `registrations`; other modules name the
+// underlying types directly (through aliases, non-UniFFI structs, or paths into
+// unparsed crates) and must still resolve to the custom type.
```
bendk, line 5:
```suggestion
// Test custom types across different module paths. 
```

uniffi_parse_rs/src/test_src/custom_type_paths.rs
```diff
@@ -0,0 +1,52 @@
...
+
+mod types {
```
bendk, line 7:
```suggestion
mod type_definitions {
```

uniffi_parse_rs/src/test_src/custom_type_paths.rs
```diff
@@ -0,0 +1,52 @@
...
+    // A non-UniFFI type covered by a custom type registered in another module
+    pub struct Concrete { }
+
+    // An alias to a type from an unparsed crate, covered by a custom type in another module
+    pub type ExternalAlias = external_crate::ExternalType;
+}
+
+mod registrations {
```
bendk, line 15:
```suggestion
mod macro_calls {
```

uniffi_parse_rs/src/test_src/custom_type_paths.rs
```diff
@@ -0,0 +1,52 @@
+// Custom types resolved by the type they cover, not the module the macro was invoked in.
+//
+// The `custom_type!` invocations live in `registrations`; other modules name the
+// underlying types directly (through aliases, non-UniFFI structs, or paths into
+// unparsed crates) and must still resolve to the custom type.
+
+mod types {
+    // A non-UniFFI type covered by a custom type registered in another module
```
bendk, line 8:
```suggestion
```

uniffi_parse_rs/src/test_src/custom_type_paths.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub struct Concrete { }
+
+    // An alias to a type from an unparsed crate, covered by a custom type in another module
+    pub type ExternalAlias = external_crate::ExternalType;
```
bendk, line 12: Let's call this a "remote type" rather than an "external type".  It's a subtle and annoying distinction, but external types come from other UniFFI-enabled crates, probably ones in your workspace.  These crates will be parsed.  Remote types come from non-UniFFI crates which won't be parsed.  I think the comment can be removed, since the meaning of "remote type" is documented in several places already.

uniffi_parse_rs/src/test_src/custom_type_paths.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    uniffi::custom_type!(ExternalAlias, String, {
+        remote,
+        into: |obj| obj.to_string(),
+        try_from: |s| s.parse(),
+    });
+
+    // A custom type over a local alias to an unparsed crate.  Other modules import the
+    // external type directly (`use external_crate2::Direct`) and never see this alias.
+    type Direct = external_crate2::Direct;
```
bendk, line 32: Isn't this the same case as `ExternalAlias`/`RemoteAlias`?  I think it can be removed.

uniffi_parse_rs/src/test_src/custom_type_paths2.rs
```diff
@@ -0,0 +1,12 @@
+// Mirrors a binding crate that reuses custom types implemented by `custom_type_paths`:
+// it names the underlying type straight from the unparsed crate, and that must resolve
+// to the custom type registered over in `custom_type_paths`.
```
bendk, line 3: Let's remove this and add a comment on the field itself.

uniffi_parse_rs/src/test_src/custom_type_paths2.rs
```diff
@@ -0,0 +1,12 @@
...
+
+mod usage2 {
+    use external_crate2::Direct;
+
+    #[derive(uniffi::Record)]
+    pub struct BindingRecord {
```
bendk, line 9:
```suggestion
    pub struct BindingRecord {
        // Custom type defined in the `custom_type_paths` crate
```

uniffi_parse_rs/src/test_src/custom_type_paths.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    // An alias to a type from an unparsed crate, covered by a custom type in another module
+    pub type ExternalAlias = external_crate::ExternalType;
+}
+
+mod registrations {
+    use crate::types::{Concrete, ExternalAlias};
+
+    uniffi::custom_type!(Concrete, String, {
+        remote,
```
bendk, line 19: This shouldn't have the `remote` attribute, since it's not a remote type.  Maybe `uniffi_parse_rs` should error out it if sees that attribute for local types.

bendk (changes requested):
Thanks for this one, I'm finally getting the issue here.

## The big picture

I think we should fix the non-remote case, however [the remote case presents some tricky questions](https://github.com/mozilla/uniffi-rs/issues/2994).  I'm not sure what the right solution is there and I'm kind of thinking the current behavior might actually be what we want.

How do you feel about addressing the local case now and leaving the remote case for later?  This means that you could import the `Concrete` type however you want, but when using the `ExternalType` and `Direct`, you would need to import from the module that contained the `custom_type!` macro call.

## The code

The examples are pretty good, but again the comments got to me.  I suggestions a handful of changes that make things clearer in my mind.

I feel like the code itself is overly complicated, maybe because it was addressing the remote case.  If we're only dealing with local items, I'm pretty sure this can all be handled by tweaking the `resolve_items()` code to add the custom type item to the module where the type was defined rather than the module with the `custom_type!` macro call.

How would you feel about me commandeering this PR and making the changes myself?  I'm only suggesting this because it's LLM code, which I think means that you're attached to it at all.  I'm not sure if that's the right way to handle this though, if you want to continue working on the code that's fine with me.

flisky:
I'm very glad if you would take over on this. The LLM could generate several solutions and bunches of code, and it's hard to me to dertermine which is better.

Sorry for the late response.

bendk:
> I'm very glad if you would take over on this. The LLM could generate several solutions and bunches of code, and it's hard to me to dertermine which is better.

I made https://github.com/mozilla/uniffi-rs/pull/3000 for this

> Sorry for the late response.

No problem at all, thanks for testing this and creating the failing test cases.  It's really useful to have examples.

Issue #2404, opened 2025-01-16
`[derive(uniffi::Record)]` should support `#[uniffi::skip]` attribute to skip fields


dev-ardi:
Right now the only way to pass the struct through uniffi with different fields is to create a "close but not quite" struct with the missing fields.

Serde has lots of custom attributes to control deserialization, uniffi could support a few of those. `#[uniffi::skip]` is almost trivial to implement.

bendk:
How would this work when the record was passed back from the foreign language -> Rust?  At that point, we'd need to fill in some value for the field.  This is kind of similar to the situation we found ourselves in with "flat" errors.

oriongonza:
It's Default value sound like the reasonable approach

bendk:
> Its Default value sounds like the reasonable approach.
>  In my opinion this should be used for Rust -> FFI, not backwards

Those 2 approaches are not exactly compatible though.  If we go with the default approach, that means we need to generate a code to lift the types, using `field_name: Default::default()` for any missing fields.  This means that things will fail to compile if the field doesn't implement default -- even if you only pass the value one way across the FFI.

To me, the second option seems more useful -- I want to define a struct that is only passed from Rust -> the foreign language and I want to skip certain fields.

The first option would also be possible, but it seems kind of confusing given that we already have a `#[uniffi::default]` attribute that works kind-of like that but it subtly different.

ladvoc:
>  This means that things will fail to compile if the field doesn't implement default -- even if you only pass the value one way across the FFI.

It might be a good idea to call this `#[uniffi::skip_with_default]` to make the fact it uses the type's default implementation explicit.

PR #2991, opened 2026-09-10
uniffi_macros: add #[uniffi(skip)] for enum and error variants

mgeisler:
Exported enums sometimes need an internal variant that only Rust code will construct, carrying data that cannot pass the FFI boundary.

This adds a new `#[uniffi(skip)]` attribute which can be added to enum and error variants. It drops the variant from both the macro-generated metadata and the `uniffi_parse_rs` source-level extractor, so its fields never need to satisfy any FFI trait.

Lowering a skipped variant across the FFI panics, since no wire tag is ever reserved for it. Skipping every variant is a compile-time error.

This feature is related to #2404, but not the same: that issue is about skipping _fields_ in a record – this is about skipping _variants_ in an enum. Because the foreign code cannot create the skipped variants, the question of what to do with "missing" data does not come up.

---

The code here was written with Claude Code and I'm not very familiar with it. The ask to Claude was:

> Please add a new feature: I would like a `#[uniffi(skip)]` attribute for enum variants (not struct (record) fields). It should:
>
> - not generate foreign code for the variant
> - allow arbitrary payloads for this variant (types that cannot cross the FFI boundary)
>
> Please add this, write tests for it (follow the project conventions), and update user facing documentation.

I later explained Claude that these variants are used for

> We have Web, Android, and iOS apps that send events to a Rust core. They only ever constructs a subset of variants, while some variants are Rust-internal. Since the internal events are never sent to foreign code, we will never run into the runtime panics when trying to lower these types.

I asked it about other cases in UniFFI where runtime panics are accepted, and it said

> Existing precedent for "compiles fine, panics on the wrong operation":
>
> - `#[non_exhaustive]` remote enums panic if an upstream crate adds a variant your bindings don't know about
> - `#[uniffi(flat_error)]` without `with_try_read` where lifting always panics, by design.

### DIFF
--- a/fixtures/enum-types/src/lib.rs
+++ b/fixtures/enum-types/src/lib.rs
@@ -142,16 +142,46 @@ fn roundtrip_boxed_record(boxed: Box<BoxedContent>) -> Box<BoxedContent> {
     boxed
 }
 
+// A variant marked #[uniffi(skip)] is invisible to bindgen, so its
+// payload can hold a type that can't cross the FFI boundary.
+#[derive(uniffi::Enum)]
+pub enum EnumWithSkippedVariant {
+    Visible,
+    #[uniffi(skip)]
+    Hidden(std::sync::mpsc::Receiver<()>),
+}
+
+#[uniffi::export]
+fn get_visible_enum_value() -> EnumWithSkippedVariant {
+    EnumWithSkippedVariant::Visible
+}
+
 uniffi::include_scaffolding!("enum_types");
 
 #[cfg(test)]
 mod test {
-    use crate::AnimalSignedInt;
+    use crate::{AnimalSignedInt, EnumWithSkippedVariant};
 
     #[test]
     fn check_signed() {
         assert_eq!(AnimalSignedInt::Koala as i8, -1);
         assert_eq!(AnimalSignedInt::Wallaby as i8, 0);
         assert_eq!(AnimalSignedInt::Wombat as i8, 1);
+    }
+
+    #[test]
+    fn skipped_variant_is_constructible_in_rust() {
+        let (_tx, rx) = std::sync::mpsc::channel::<()>();
+        let value = EnumWithSkippedVariant::Hidden(rx);
+        assert!(matches!(value, EnumWithSkippedVariant::Hidden(_)));
+    }
+
+    #[test]
+    #[should_panic(expected = "#[uniffi(skip)]")]
+    fn skipped_variant_cannot_cross_the_ffi() {
+        let (_tx, rx) = std::sync::mpsc::channel::<()>();
+        let _ = <EnumWithSkippedVariant as uniffi::Lower<crate::UniFfiTag>>::lower(
+            EnumWithSkippedVariant::Hidden(rx),
+        );
     }
 }
--- a/uniffi_macros/src/error.rs
+++ b/uniffi_macros/src/error.rs
@@ -4,7 +4,7 @@ use syn::{DeriveInput, Index};
 use uniffi_meta::EnumShape;
 
 use crate::{
-    enum_::{rich_error_ffi_converter_impl, variant_metadata, EnumItem, VariantAttr},
+    enum_::{rich_error_ffi_converter_impl, variant_attrs, variant_metadata, EnumItem},
     ffiops,
     util::{
         create_metadata_items, extract_docstring, ident_to_string, orig_name_metadata,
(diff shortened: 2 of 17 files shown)


mgeisler:
I know the PR here comes out of the blue – but since Claude could generate the code quickly, I did it this way.

The need for this is because we're migrating from https://github.com/redbadger/facet-generate to UniFFI, and `facet-generate` gets a [`skip` feature from Facet](https://facet.rs/reference/field-attributes/#skip).

fixtures/enum-types/src/lib.rs
```diff
@@ -142,16 +142,46 @@ fn roundtrip_boxed_record(boxed: Box<BoxedContent>) -> Box<BoxedContent> {
...
 }
 
+// A variant marked #[uniffi(skip)] is invisible to bindgen, so its
+// payload can hold a type that can't cross the FFI boundary.
+#[derive(uniffi::Enum)]
+pub enum EnumWithSkippedVariant {
+    Visible,
+    #[uniffi(skip)]
+    Hidden(std::sync::mpsc::Receiver<()>),
+}
```
bendk, line 152: We're hoping to move tests like these to `bindgen-tests` and remove these fixtures in the near future.  Could you ask Claude to move the new test there rather than this fixture?  If you did that, I think you could remove the tests in `fixtures/proc-macro` as well.  The `fixtures/uitests` changes are great though, don't remove those.

The test itself looks good, I like that we're testing a type that definitely can't be sent over the FFI.

uniffi_macros/src/error.rs
bendk, line 135: Could this also iterate over skipped and panic with the a message like "Attempted to read a #[uniffi(skip)] enum variant across the FFI", similar to the `lower` code?

uniffi_parse_rs/src/attrs/enums.rs
```diff
@@ -144,6 +145,9 @@ impl VariantAttributes {
                             let name: LitStr = meta.input.parse()?;
                             parsed.name = Some(name.value());
                             Ok(())
+                        } else if meta.path.is_ident("skip") {
+                            skip = true;
```
bendk, line 149: I think this should work and it would be fine for the initial PR, but it would be better if we could generate the error messages like `uniffi_macros` does.  I think this means:
- Adding a `VariantAttributes::skipped` field
- Setting that field here rather than returning `Ok(None)` below
- Updating the [scaffolding code](https://github.com/mozilla/uniffi-rs/blob/main/uniffi-bindgen-kotlin-jni/src/scaffolding/templates/enum.rs#L12) to `bail!` if it sees one of the skipped variants
- Tracking the `skip` in the [pipeline code](https://github.com/mozilla/uniffi-rs/tree/main/uniffi_bindgen/src/pipeline).
- Update the [bindings code](https://github.com/mozilla/uniffi-rs/blob/0d732ee489c129fdd4baf82355649f453df7f9fc/uniffi-bindgen-kotlin-jni/src/bindings/templates/Enum.kt#L31) to skip over the skipped variants.

Now that I write it out, this is a fairly long list.  If Claude can't figure it out easily, then this would be fine as a follow-up.

bendk (changes requested):
This makes sense as a feature and the changes look mostly good to me.  There were just a few minor issues that I saw.

I also what others think about the new syntax.  It makes sense to me, but maybe others can think of a better way to express this.  I'm thinking we can wait to merge this for a week or so to see if anyone has thoughts.

mhammond:
This does seem like a footgun and not really the same as the other panics - this is more dynamic, where a struct holding another struct holding this enum will panic because some other Rust code far away from here used a "forbidden" variant. OTOH, a reasonable objection to that objection is "well don't use this capability". But I really am interested in other's opinions here.

(Assuming we take this, `skip` seems fine to me and other alternatives aren't immediately obvious)

bendk:
> This does seem like a footgun and not really the same as the other panics - this is more dynamic, where a struct holding another struct holding this enum will panic because some other Rust code far away from here used a "forbidden" variant. OTOH, a reasonable objection to that objection is "well don't use this capability". But I really am interested in other's opinions here.

I wonder what other's think too.

I'm thinking of this like a custom type that maps the enum with all the variants to one with the non-skipped variants and panics if it sees a skipped variant.  That's what I'd recommend someone do with UniFFI today and this is just a more convenient way to do it.  But it is somewhat of a footgun and there's an argument to be had that we shouldn't make it too easy to do.

It also make me think that we should make the method `try_lower` instead of `lower` which would be the first step to allowing users to handle these types of errors.

Issue #2981, opened 2026-09-04
Remove redundant fixture tests


bendk:
The [bindgen-tests crate](https://github.com/mozilla/uniffi-rs/issues/2686) was created to replace the fixture tests and has several advantages to them.  We should go through the fixture tests and find tests that are redundant with the bindgen tests.  For example, `fixtures/ext-libs` is covered by the `external-types` module.

For each of those tests we should find any functionality tested by the fixture test that's not tested by the bindgen test, update the bindgen tests to cover that, and delete the fixture.

PR #2993, opened 2026-09-11
Detect desugared async functions as async

jmartinesp:
# Changes

- Function signatures in `uniffi_parse_rs` and `uniffi_macros` now are also used to check the function's return type when the `async` modifier is missing to double check if it should be considered an async function. Each crate checks this value on its own, but under the hood a shared helper code to check the asyncness of the function using the return type is shared.
- The helper code lives in a new `uniffi_syn_utils` crate since I wasn't sure `uniffi_meta` would be the right crate for it and it should be shared between the 2 crates to have the exact same results.
- Added tests to check the behaviour of desugared `async` functions and traits.

# Motivation

We recently had a PR in our project where a contributor changed a normal function in a callback interface into an async one, like this:

```rust
#[uniffi::export(callback_interface)]
#[async_trait::async_trait]
pub trait Foo: Send + Sync {
   async fn bar() -> Bar
}
```

However, our project is already slow to build and we found `async_trait` in the past slowed down the build time significantly so we mostly forbid its usage, but when I suggested using the desugared version, like this:

```rust
#[uniffi::export(callback_interface)]
pub trait Foo: Send + Sync {
  fn bar() -> Pin<Box<Future<Output = Bar> + Send + Sync>>>
}
```

This contributor pointed out this wouldn't work with exported FFI code because UniFFI relies on the `async` keyword in a function to detect if it's asynchronous, and as I later checked, they were right about it.

So with some assistance from AI I tried to have this desugared async format detected as a valid async function in UniFFI, and it turns out it was simpler than I expected at first, although I'm not sure if I might have missed some corner cases or constraints. Also, I was quite hesitant to add the new crate, since only the added shared code would live there.


mhammond (commented):
Channelling Ben, I think it would probably be better if the tests for this were actually in bindgen-tests/lib/src/futures.rs.

uniffi_syn_utils/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "uniffi_syn_utils"
```
mhammond, line 2: yeah, this crate really is unfortunate. Hopefully Ben has some idea here.

bendk:
> Channelling Ben, I think it would probably be better if the tests for this were actually in bindgen-tests/lib/src/futures.rs.

Sorry about this, I really need to take on https://github.com/mozilla/uniffi-rs/issues/2981 soon.

uniffi_macros/src/fnsig.rs
```diff
@@ -120,11 +120,28 @@ impl FnSignature {
...
-        let output = match sig.output {
-            ReturnType::Default => quote! { () },
-            ReturnType::Type(_, ty) => quote! { #ty },
+        // A function is async if it is declared `async fn`, or if it manually returns a
+        // boxed future such as `Pin<Box<dyn Future<Output = T> + Send>>` — the shape
+        // `#[async_trait]` expands `async fn` into. Dropping the `#[async_trait]` macro and
+        // the `async` modifier while returning that future by hand must still be treated as
+        // an async function: the scaffolding awaits the returned value either way.
+        //
+        // For a real `async fn`, `syn` gives us the unsugared `Output` type. To treat the
```
bendk, line 129:
```suggestion
        // For a real `async fn`, `syn` gives us the desugared `Output` type. To treat the
```

bendk: nit: I think "desuraged" is more standard, can you change it here and in the other comments?

uniffi_parse_rs/src/functions.rs
```diff
@@ -143,6 +145,33 @@ impl ReturnType {
...
+    /// A function is async when it is declared `async fn` (`declared_async`), or when it
+    /// manually returns a boxed future such as `Pin<Box<dyn Future<Output = T> + Send>>` —
+    /// the shape `#[async_trait]` expands `async fn` into. Removing the `#[async_trait]`
+    /// macro and the `async` modifier while returning that future by hand is still an async
+    /// function as far as the FFI is concerned.
+    ///
+    /// For a real `async fn`, `syn` hands us the unsugared `Output` type directly. To treat
+    /// the manual future the same way, we unwrap the future here and use its `Output` as the
+    /// effective return type.
+    pub fn parse_async(declared_async: bool, output: syn::ReturnType) -> syn::Result<(bool, Self)> {
```
bendk, line 159: `parse_async` doesn't feel like the right name to me.  I read it as either parsing a return type, but asynchronously, or parsing the `async` keyword.  I don't have a great suggestion though.

uniffi_syn_utils/src/lib.rs
```diff
@@ -0,0 +1,128 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+//! `syn` AST helpers shared across the uniffi crates.
+//!
+//! These utilities operate purely on `syn` types so they can be reused by both the source
+//! parser (`uniffi_parse_rs`) and the proc-macros (`uniffi_macros`) without either crate
+//! depending on the other.
```
bendk: TODO

uniffi_syn_utils/Cargo.toml
bendk: I think we should have separate implementations for `uniffi_macros` and `uniffi_parse_rs`.

For `uniffi_macros`, the current implementation makes a lot of sense and I think we should just move the code from this crate to there.

For `uniffi_parse_rs`, I think we can do better by adding a `BuiltinItem::Future` and `BuiltinItem::Pin` variant, then connect it to the type resolution code.  This would allow handling this even through type aliases and weird imports that end up renaming `Pin`, `Box` and/or `Future` (and prevent false positives when other types are renamed to those names).  It's a bit niche, but I could see a user creating a type alias so they don't need to type out `Pin<Box<dyn Future<Output = T>>>>`.

The code to do this is mostly already there, for example the code to handle `Box` is [here](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_parse_rs/src/paths.rs#L721) and [here](https://github.com/mozilla/uniffi-rs/blob/main/uniffi_parse_rs/src/types.rs#L469-L471).  However, I think the generic handling code would need to be updated to handle associated types.  The `dyn Future` part might also break something.

If you wanted to look into this for this PR that's great.  If not, then just do the `uniffi_macros` implementation and leave `uniffi_parse_rs` as a todo.  Once you migrate the tests to `bindgen-tests`, you could add the test case to `bindgen-tests/kotlin` and add a commented out TODO test item to `uniffi-bindgen-kotlin-jni/tests`.

Long-term, I think that eventually we would want `uniffi_macros` to depend on `uniffi_parse_rs` and we should be able to share a good deal of code there.  Unfortunately however, I don't think we can share code like this that depends on type resolution, since `uniffi_macros` only parses locally.  It sees the current item that's being wrapped with a macro, but it doesn't see the entire crate and therefore can't handle things like type aliases.

bendk (changes requested):
The changes look good to me, but there's some refactoring needed: moving the tests to `bindgen-tests` and moving the code to `uniffi_macros` only.

Like I said in the comment, if you want to tackle the `uniffi_parse_rs` work, that would be awesome, but not necessary for this PR.

jmartinesp:
> For uniffi_parse_rs, I think we can do better by adding a BuiltinItem::Future and BuiltinItem::Pin variant, then connect it to the type resolution code. This would allow handling this even through type aliases and weird imports that end up renaming Pin, Box and/or Future (and prevent false positives when other types are renamed to those names). It's a bit niche, but I could see a user creating a type alias so they don't need to type out Pin<Box<dyn Future<Output = T>>>>.

I tried just deleting the code in `uniffi_parse_rs` and keeping only the changes in `uniffi_macros`, then added a callback interface with a desugared async funtion to `bindgen-tests/lib`. However, as soon as I tried to run the tests, the JNI bindings were complaining the wrong format was used for the callback interface and refused to run.

At first I tried to fix this manually, but in the end I got nowhere since I'm not familiar enough with the code and had an agent try to fix this, which it did. The code seems sensible, AFAICT it follows your suggestions and adds the minimal changes, but even like that it changes quite a lot of files, and I'm not super comfortable with that. 

If you don't mind the size and reach of the PR growing quite a bit, and most of these new changes being vibe-coded, I can upload them, since they seem to work fine.

bendk:
Let's see the vibe-coded changes.  I could see them being pretty good actually, since it might just need to copy the existing examples.

fixtures/metadata/src/tests.rs
```diff
@@ -957,6 +959,7 @@ mod test_function_metadata {
                 name: "async_sub".into(),
                 orig_name: None,
                 is_async: true,
+                desugared_async: false,
```
bendk: Do we need this?

uniffi-bindgen-kotlin-jni/src/scaffolding/templates/callback_interface.rs
```diff
@@ -5,10 +5,27 @@ struct {{ cbi.impl_struct_rs() }} {
...
+        {%- if receiver.by_ref %}
+        &self,
+        {%- else %}
+        self: ::std::sync::Arc<Self>,
+        {%- endif %}
+        {%- for a in meth.callable.arguments %}
+        {{ a.name_rs() }}: {{ a.ty.type_rs }},
+        {%- endfor %}
+    ) -> ::std::pin::Pin<::std::boxed::Box<dyn ::std::future::Future<Output = {{ meth.callable.result.return_type_rs() }}> + ::std::marker::Send + {% if receiver.by_ref %}'_{% else %}'static{% endif %}>> {
+        ::std::boxed::Box::pin(async move {
```
bendk, line 27: Could we use this version for the `async_trait` case too?  It would be nice to only have 1 code path here and I think it might mean we didn't have to pass around `desugared_async` through the pipeline.  However,  I'm not sure if it's compatible with the `async_trait` magic.

bendk (commented):
I have a few thoughts about this one, but one main question.  Let's get that figured out first before I ask those.

PR #2996, opened 2026-09-15
Updating bindgen-tests

bendk:
Used claude to audit the callbacks and enums test fixtures to find things they test that `bindgen-tests` don't.  Updated `bindgen-tests` based on those findings.

Updated `uitests` to not depend on the trait-methods UDL.  I think `uitests` should stay around long-term, but `trait-methods` can be folded in to bindgen-tests


bendk:
I think this is the first step towards cleaning up the test fixtures.  Hopefully it will make it so contributors add tests to `bindgen-tests` instead of here.  There's still a bunch of leftover tests that we probably should deal with.  For example, we can migrate the `keywords` tests to `bindgen-tests`.

mhammond (commented):
Are we sure all these are redundant? coveralls in particular has a number of things I don't think are likely to be in the new tests?

bendk:
I was definitely overeager with coverall.  I went through the code and also asked claude and came up with https://github.com/mozilla/uniffi-rs/issues/2997.  I'm going to undo the coverall deletion for this PR.

bendk:
...and Claude actually find missing parts with the other tests as well.  I changed this to only focus on the `callbacks` and `enum-types` tests.

bindgen-tests/python/tests/enums.py
```diff
@@ -12,6 +20,9 @@ def test_simple_enums(self):
         self.assertEqual(
             roundtrip_enum_with_data(EnumWithData.C()) ,
             EnumWithData.C())
+        # passing in an int value rather than an enum should fail
+        with self.assertRaises(ValueError):
+            roundtrip_enum_with_data(1)
```
bendk, line 25: This catches a new bug in the python binding: passing the wrong type to a function that expects an enum object.  `fixtures/callback-interfaces` tested the flat enum case, but not the case where the enum stored data and was therefore implemented as a Python class.

bendk:
~I realize we can't actually remove these fixtures yet, since other fixtures depend on them (I think coverall?).  So now this is just about adding test cases to `bindgen-tests`.~

Actually we can remove the fixtures, I got confused by the uniffi-bindgen-kotlin-jni errors.

bindgen-tests/lib/src/callback_interfaces.rs
```diff
@@ -45,6 +45,13 @@ pub fn invoke_test_callback_interface_set_value(cbi: Box<dyn TestCallbackInterfa
     cbi.set_value(value)
 }
 
+#[uniffi::export]
+pub fn test_optional_callback_interface(
+    cbi: Option<Box<dyn TestCallbackInterface>>,
+) -> Option<u32> {
+    cbi.map(|cbi| cbi.get_value())
+}
+
```
bendk, line 54: Adding this test exposed a real bug in `uniffi-bindgen-kotlin-jni`, thanks far noticing this one Claude!

PR #2949, opened 2026-07-26
Kotlin: stop emitting unused expressions in generated bindings

ahochheiden:
The default warnings changed in Kotlin 2.4, which Firefox is using, but `uniffi-rs` seemingly isn't yet. Regardless, we should fix the warnings.

This came up during the `application-services` monorepo work in Firefox. Full stack on Phabricator here: https://phabricator.services.mozilla.com/D313774

PR #2998, opened 2026-09-16
Kotlin: stop emitting unused expressions in generated bindings

bendk:
Pull in https://github.com/mozilla/uniffi-rs/pull/2949/ to the 0.32.x release branch.

## Original message

The default warnings changed in Kotlin 2.4, which Firefox is using, but `uniffi-rs` seemingly isn't yet. Regardless, we should fix the warnings.

This came up during the `application-services` monorepo work in Firefox. Full stack on Phabricator here: https://phabricator.services.mozilla.com/D313774


bendk:
The main motivation here is to help the FF build team since the warnings are creating issues for them.  My plan is to merge this to the branch, then let them use the branch for a while.  I'll also be using a branch to test out `uniffi-bindgen-kotlin-js` and `uniffi_parse_rs`.  Once that seems to be working, then I'll merge those into the 0.32.x branch and we can to a 0.32.2 release with those changes (plus whatever anyone else wants).

mhammond (approved):

Issue #2994, opened 2026-09-12
uniffi_parse_rs: dealing with remote custom type paths


bendk:
`uniffi_parse_rs` has issues dealing with remote custom types.  For example:

```
/// Setup a custom type for a remote type
use url::Url;
custom_type!(Url, String, {
   // ...
});

/// Create a record that contains that type
#[derive(uniffi::Record)]
pub struct MyRecord {
   // Works: using the exact same item
   a: Url,
   // Fails: using the full path from the remote crate
   b: url::Url,
   // Fails: using a different import from this crate (type aliases also fail)
   c: crate::other_module::Url,
   // Fails: using an re-export/type alias from a third crate
   d: ::some_other_crate::Url,
}
```

Furthermore, remote custom type definitions only apply to the current crate, which creates some really weird corner cases.  For example:

- crate `a` defines a custom type for `Url`
- crate `b` defines a separate custom type for `Url`
- crate `c` depends on both `a` and `b`.
- crate `c` exports a type that contains `Url`, which custom type impl should we use?

bendk:
Because of the corner case above, maybe the current status quo is the best for now.  It sort-of makes sense conceptually if you think about it like this: the `custom_type!` macro call defines a new `Url` type with the custom type logic.  If you want to use that custom type, you need to import from the module with the `custom_type!` call, not the module where the original `Url` type was defined.

I feel like we could improve the ergonomics at some point, but I don't feel like I know enough to say what the right way is now.

PR #3000, opened 2026-09-17
uniffi_parse_rs: Fix custom types path resolution

bendk:
Handle the case where a type is defined in module A, then the custom_type! macro is in module B.  Before, this would fail with item not found errors.

Track which custom types are remote or not.  Remote custom types don't get the same treatment.


mhammond (approved):
this looks correct to me

flisky:
Thanks for picking this up — the relocation approach is much simpler.

There are still two corner cases for our codebase. The following is generated by LLM, and hope it helps.

---

## 1. A local `type` alias aborts the whole crate

```rust
pub mod inner {
    pub struct Counter {}
}

pub mod stock {
    pub use crate::inner::Counter;
}

pub mod registrations {
    pub type Counter = crate::stock::Counter;

    uniffi::custom_type!(Counter, String, {
        try_lift: |val| val.parse(),
        lower: |obj| obj.to_string(),
    });
}
```

On `main` this resolves from `registrations` (the documented workaround). On this branch `resolve_items()` fails outright:

```
error: Unknown item
  │     uniffi::custom_type!(Counter, String, {
  │                          ^^^^^^^ Unknown item
```

`Namespace::NonUniffiType` matches only `Item::NonUniffi(_, _)`, but a `type` alias is `Item::Type(_)`, so the phase-2 lookup fails — and because the error propagates, nothing in the crate resolves any more. A `use` import works fine; it's specifically the alias.

The narrow fix is to widen the namespace, but that only makes the lookup land on the alias item itself, so nothing relocates anyway. Falling back seems better than either — it keeps `main`'s behaviour for anything phase 2 can't place, instead of failing the crate:

```rust
// non-remote types move to the module where the type was defined.
// If that module can't be found -- the ident is a type alias, say --
// leave the item where the macro was written rather than failing the
// whole crate.
path.resolve(
    self,
    &mut cache,
    &c.ident.clone().into(),
    crate::Namespace::NonUniffiType,
)
.and_then(|path_to_type| Ok(path_to_type.module()?.id))
.unwrap_or(module.id)
```

Following the alias through to its target would be better still, but that's a real change and I didn't want to guess at what you'd want.

## 2. Relocation also moves the scope `builtin` is resolved in

```rust
pub mod api {
    pub struct ApiResponse {
        pub data: String,
    }
}

pub mod registrations {
    pub use crate::api::ApiResponse;

    #[derive(uniffi::Record)]
    pub struct HttpResponse {
        pub data: String,
    }

    uniffi::custom_type!(ApiResponse, HttpResponse, {
        try_lift: |v| Ok(v),
        lower: |o| o,
    });
}
```

`main`:

```
registrations::ApiResponse => Ok(Custom { builtin: Record { module_path: "…::registrations", name: "HttpResponse" } })
```

this branch — every route, including the one that used to work:

```
registrations::ApiResponse => Err(NotFound)
api::ApiResponse           => Err(NotFound)
usage::ApiResponse         => Err(NotFound)
```

The custom type is found; its `builtin` is not. `_resolve_type` resolves `builtin` against `path_to_type.parent()`, which relocation has just changed, so `HttpResponse` is looked up in `api` where it isn't in scope. `CustomType::custom_type_metadata` has the same problem.

This one has no workaround — the type is unusable from anywhere — and it isn't a corner case: a custom type whose FFI representation is a `#[derive(uniffi::Record)]` sitting next to the macro is a normal way to write these.

What I did locally was record the macro's module on `CustomType` and resolve `builtin` against that:

```rust
pub struct CustomType {
    pub remote: bool,
    /// Module the `custom_type!` macro was written in.
    ///
    /// The item itself can be moved to the module defining the type it covers, but `builtin` is
    /// written in terms of what's in scope here.
    pub macro_module_path: String,
    ...
}
```

set from `path.path_string()` in `maybe_resolve_macro`, with a `path_for_module(ir, &str) -> Result<RPath>` helper to map it back. Happy to open that as a PR against this branch if it's the direction you want — it's about 50 lines including the helper, and the existing tests all pass.

bendk:
Thanks for checking on this and generating those extra cases.  The first one was pretty easy, but the second required me to rework this a bit.  Out of curiosity, what model/prompts are you using?  That way I can try running them myself.

flisky:
Opus 5 is good enough at reasoning, and luckily, we have not-so-normal uniffi setup (The uniffi is just one binding for the core, and the core is loose couple with uniffi). So the prompt could be the error info when wiring up the new parser.

bendk:
Ahh I see.  How does the new version work against your codebase?

flisky:
Both cases are fixed. Thank you!

However, two things it needed on our side, both from the remote carve-out rather than from bugs.

**Remote custom types have to be named through the module holding the macro.** 

Given the alias and its macro in one place:

```rust
// core/src/ffi_types.rs
pub type Decimal = rust_decimal::Decimal;

#[cfg(feature = "uniffi")]
uniffi::custom_type!(Decimal, String, {
    remote,
    try_lift: |v| Ok(v.parse()?),
    lower: |o| o.to_string(),
});
```

every exported signature elsewhere in the library has to be spelled through it:

```rust
// core/src/model/watchlist.rs
-use rust_decimal::Decimal;
+use crate::ffi_types::Decimal;

 #[derive(uniffi::Record)]
 pub struct Quote {
     pub price: Decimal,
 }
```

Both forms are the same Rust type and both compile; only the second resolves:

```
model :: Decimal  (use rust_decimal::Decimal)     => Err(NotFound)
model :: Decimal  (use crate::ffi_types::Decimal) => Ok(Custom { .. bridge_type: String })
```

So the constraint isn't really "import the custom type from the macro's module" — inside the
library it becomes "never import this third-party type the obvious way", in files that are
otherwise unaware UniFFI exists.

That's a one-time change and the result is arguably tidier, so I'm not arguing it's a blocker.
But it is a library rearranging itself to suit a bindings generator, and I think that's the
cost worth weighing rather than the per-site renaming, which is trivial.

(Related with #2994)

**use_remote_type! needed a workaround**

```rust
uniffi::use_remote_type!(core::AdjustType);   // AdjustType is #[uniffi::remote(Enum)] in core
```

fails with `Unknown item`, because the parser resolves `core::AdjustType` literally while the
macro treats `core` as the implementing crate and resolves `AdjustType` in the invoking
module's scope — where it is in fact imported. Rust compiles it fine; only the parser objects.

We unblocked it by adding a re-export to `core`'s root that Rust has no use for:

```rust
#[cfg(feature = "uniffi")]
pub use foo::bar::AdjustType;
```

That's #2984, which I closed when it was stacked on #2990. I checked that cherry-picking
`bdac6cf48` onto this branch produces byte-identical bindings to the re-export workaround.

bendk:
> **Remote custom types have to be named through the module holding the macro.**
> (...)
> That's a one-time change and the result is arguably tidier, so I'm not arguing it's a blocker.
But it is a library rearranging itself to suit a bindings generator, and I think that's the
cost worth weighing rather than the per-site renaming, which is trivial.

I understand that this is annoying, but I'm not sure there's a trivial fix for the general issue (#2994).  What do you mean by "per-site renaming"?

I think there is one case where I think we could make things work easily.  If you define a custom type in a crate, then whenever you use that same type inside that same crate, we should be able to find the correct custom type.  As long as everything is happening in 1 crate then I think this should be fine.  Let's deal with this in a follow-up though.

> **use_remote_type! needed a workaround**

This seems like a straightforward issue/fix, but I'd also like to handle this in a followup.  This PR feels pretty big already and I'd rather not add more to it.

